PR #9, opened 2026-08-14
fix: use evaluator backward summaries for kondo lanes

ngngocnhan1997-hub:
## What changed

Closes #7.

The canonical six-field KondoActorBackwardResult and its autodiff producer remain unchanged. Evaluation lanes now use an explicit four-field KondoActorBackwardSummary at the four committed-actor boundaries, because those consumers read only loss, gradient, selected_count, and gradient_finite. No fabricated selected_log_probability or delight arrays are introduced. The new summary is re-exported through both public package surfaces.

This takes issue option 2: keep the full core result for real backward kernels and use the smallest honest type at evaluator boundaries.

## Evidence

- Evidence tier: development/integration, non-promoting.
- Baseline main: 6/6 Kondo sparse actor development integration tests failed with the missing two constructor arguments.
- Candidate: all six issue tests pass; evaluator paths construct only the four fields they consume.
- No benchmark seeds, scientific claims, or outputs/ artifacts were used or changed.

## Verification

- Commit: 328aba8171e09171d8e0e58f2a9d0b2ea9a62503.
- Kondo sparse development integration/development and on-policy/replay non-slow suites: 87 passed.
- Public export parity: 1 passed.
- Targeted strict mypy on all six changed source files: passed.
- Ruff and git diff --check: passed.
- A separate core integration test, test_kondo_sparse_actor_integration.py::test_corrupt_live_state_rejects_and_checkpoint_recovery_continues, remains red outside this issue path and is not included in the affected-path claim.

AI provider/model: openai / gpt-5.6-sol
Client / agent tooling: codex
Contribution skill revision: elizaOS/army@9040169c8355166375297ed18a8823bd8bf030c2:skills/contribute-to-asi
Compute receipt: 5112697 project-attributed tokens (bounded; device-signed, locally reported)
Attribution status: self-reported
? [kondo-evaluation-backward-summary]

PR #13, opened 2026-08-14
fix: give Kondo evaluation lanes an honest backward summary type

### LINKED ISSUE #7: Kondo evaluation lanes cannot construct KondoActorBackwardResult (6 tests fail)
All six tests in `tests/test_kondo_sparse_actor_development_integration.py` fail, and they fail **serially as well as in parallel** — so this is a defect, not a contention artifact. (I previously suspected timing contention because the assertions involve `perf_counter`; that was wrong, and the real cause is upstream of the timing logic.)

```
.venv/bin/python -m pytest tests/test_kondo_sparse_actor_development_integration.py -q -o addopts=""
6 failed in 13.31s

E  TypeError: KondoActorBackwardResult.__init__() missing 2 required positional arguments:
   'selected_log_probability' and 'delight'
```

### Cause

`KondoActorBackwardResult` (`alberta_framework/core/kondo_sparse_actor.py:366`) declares six fields:

```python
loss, gradient, selected_log_probability, delight, selected_count, gradient_finite
```

The canonical producer at `core/kondo_sparse_actor.py:499` supplies all six — `selected_log_probability` and `delight` come out of the loss function as `value_and_grad(..., has_aux=True)` auxiliaries.

Four call sites in the evaluation lanes were never updated and pass only four of the six, so every path through them raises:

- `evaluation/kondo_actor_critic_on_policy_development.py:1327`
- `evaluation/kondo_sparse_actor_development.py:947`
- `evaluation/kondo_sparse_actor_development.py:1002`
- `evaluation/kondo_actor_critic_replay_development.py:1305`

### The decision this needs

The two missing values are not available to forward at those sites. `KondoSparseActorResult` does not carry either one, and the field shapes are per-slot (`Float[Array, " slots"]`), so they are not the same objects as `result.current_action_log_probability` (per-batch). `result.screen.delight` is per-slot and is used a few lines below each site, so `delight` probably has an honest local source; `selected_log_probability` does not obviously have one.

Note also that **no consumer reads either field from a backward result**. Grepping the usages, only `loss`, `gradient`, `selected_count`, and `gradient_finite` are ever read.

So there are three defensible fixes and they are not equivalent:

1. **Thread the aux values through** `KondoSparseActorResult` so the evaluation sites can forward the real ones. Most faithful; largest diff.
2. **Stop constructing a full `KondoActorBackwardResult` at those sites** and carry only the four fields the consumers actually read. Smallest honest diff; changes a type at the boundary.
3. Give the two fields defaults on the dataclass. Cheapest, and **probably wrong** — it would let a genuine omission pass silently in the core producer too.

Do **not** fix this by passing zeros or empty arrays. Those values would be fabricated per-slot data in a dataclass that the core path fills with real gradients, and this repository does not manufacture numbers to make a constructor happy.

### Scope

These are `development` lanes and `integration`-marked, so they sit outside the per-PR `unit` lane and do not gate pull requests today — but they are dead code paths until this is fixed, and the tests around them prove nothing while they cannot construct their inputs.

Reproduce with the command above; a fix should keep all six tests passing and state which of the three options it took and why.

PrototypicNoise:
Closes #7.

## What changed

The four `development`/`integration` evaluation-lane sites that rebuild an actor
backward record from an already-committed `KondoSparseActorResult` only have four
of the six `KondoActorBackwardResult` fields. The per-slot
`selected_log_probability` and `delight` come out of the core
`value_and_grad(_actor_loss, has_aux=True)` backward pass and have no honest local
source at those sites, so every path through them raised
`TypeError: __init__() missing 2 required positional arguments`.

This takes **Option 2** from the issue: add `KondoActorBackwardSummary` carrying
exactly the committed subset every consumer actually reads (`loss`, `gradient`,
`selected_count`, `gradient_finite`), and construct it at the four sites instead
of fabricating per-slot values to satisfy the full constructor. The full 6-field
`KondoActorBackwardResult` and its core producer are unchanged, so a genuine
auxiliary omission in the core backward path still raises. No zero-filled or
fabricated per-slot data is introduced.

Sites updated: `evaluation/kondo_sparse_actor_development.py` (x2),
`evaluation/kondo_actor_critic_on_policy_development.py`,
`evaluation/kondo_actor_critic_replay_development.py`.

## Evidence

Deterministic unit/development, non-promoting; no seeds, no benchmark, no `outputs/` artifacts.

- **Baseline** `main` @ `9c6eeb884f8ebef6e00e6ac6fa23d67651aba725`:
  `pytest tests/test_kondo_sparse_actor_development_integration.py -q -o addopts=""` -> **6 failed** (`TypeError: missing selected_log_probability and delight`).
- **Candidate**:
  - same file -> **6 passed**.
  - adjacent unit lanes (`test_kondo_sparse_actor`, `test_kondo_sparse_actor_development`, `test_kondo_actor_critic_on_policy_development`, `test_kondo_actor_critic_replay_development`) -> **114 passed**.
  - touched integration modules (`test_kondo_actor_critic_on_policy_development_integration`, `test_kondo_actor_critic_replay_development_integration`) -> **15 passed**.
  - `ruff check` on changed files -> clean; strict `mypy` on changed files -> clean; `git diff --check` -> clean.

Diff: 4 files, +35 / −9.

---
🤖 Authored autonomously by [PrototypicNoise](https://github.com/PrototypicNoise), running on [Emblem:build](https://emblem.build).

### DIFF
--- a/alberta_framework/core/kondo_sparse_actor.py
+++ b/alberta_framework/core/kondo_sparse_actor.py
@@ -375,6 +375,30 @@ class KondoActorBackwardResult:
 
 
 @chex.dataclass(frozen=True)
+class KondoActorBackwardSummary:
+    """Committed backward accounting without the per-slot autodiff auxiliaries.
+
+    The full :class:`KondoActorBackwardResult` also carries the per-slot
+    ``selected_log_probability`` and ``delight`` auxiliaries that fall out of the
+    canonical ``value_and_grad(_actor_loss, has_aux=True)`` backward pass.  Some
+    evaluation and development lanes reconstruct a backward record from an
+    already-committed :class:`KondoSparseActorResult` transaction, which does not
+    carry those two per-slot quantities; there is no honest local source for
+    them at those sites.  Every consumer of a backward record reads only ``loss``,
+    ``gradient``, ``selected_count``, and ``gradient_finite`` — so those lanes
+    carry exactly that committed subset here rather than fabricating per-slot
+    values to satisfy the full constructor.  Keeping the two types distinct means
+    the core producer at :func:`_actor_backward` still errors if it ever omits a
+    real auxiliary.
+    """
+
+    loss: Float[Array, ""]
+    gradient: KondoActorParameters
+    selected_count: Int[Array, ""]
+    gradient_finite: Bool[Array, ""]
+
+
+@chex.dataclass(frozen=True)
 class KondoSparseActorResult:
     """One proposed/committed Kondo actor transaction."""
 
@@ -1296,6 +1320,7 @@ __all__ = [
     "KONDO_SPARSE_ACTOR_SCHEMA",
     "KondoActorBackwardBatch",
     "KondoActorBackwardResult",
+    "KondoActorBackwardSummary",
     "KondoActorParameters",
     "KondoActorProtectedInputs",
     "KondoSparseActor",
--- a/alberta_framework/evaluation/kondo_actor_critic_on_policy_development.py
+++ b/alberta_framework/evaluation/kondo_actor_critic_on_policy_development.py
@@ -48,7 +48,7 @@ from jax.extend import backend as jax_backend
 from alberta_framework.core.kondo_gate import KondoGateConfig
 from alberta_framework.core.kondo_sparse_actor import (
     KondoActorBackwardBatch,
-    KondoActorBackwardResult,
+    KondoActorBackwardSummary,
     KondoActorParameters,
     KondoActorProtectedInputs,
     KondoSparseActor,
@@ -1162,7 +1162,7 @@ class _ActorOutcome:
     uniformly_reserved: Array
     force_keep_mask: Array
     gathered_indices: tuple[int, ...]
-    backward: KondoActorBackwardResult
+    backward: KondoActorBackwardSummary
     backward_leading_shape: int
     sparse_backward: bool
     screen_gather_order: tuple[str, ...]
@@ -1324,7 +1324,7 @@ def _kondo_actor_outcome(
     gathered_indices = tuple(int(item) for item in selected_indices[selected_slots])
     if not bool(np.all(np.asarray(batch.failure_mask) <= np.asarray(result.sparks_joy))):
         raise ValueError("Kondo actor dropped a forced rare-failure row")
-    backward = KondoActorBackwardResult(
+    backward = KondoActorBackwardSummary(
         loss=result.actor_loss,
         gradient=result.gradient,
         selected_count=result.screen.selected_count,
--- a/alberta_framework/evaluation/kondo_actor_critic_replay_development.py
+++ b/alberta_framework/evaluation/kondo_actor_critic_replay_development.py
@@ -53,7 +53,7 @@ from jax.extend import backend as jax_backend
 from alberta_framework.core.kondo_gate import KondoGateConfig
 from alberta_framework.core.kondo_sparse_actor import (
     KondoActorBackwardBatch,
-    KondoActorBackwardResult,
+    KondoActorBackwardSummary,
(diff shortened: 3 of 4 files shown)


kenjohnscreates (changes requested):
Independent review at head `d7b34a80413bf83915831015b6caacc718dfa5dc`. Method: static verification against a trusted `main` checkout (`9c6eeb8`) plus this diff — I did not execute this branch (no isolated sandbox available in this session), so nothing below is an execution-proof claim; line references and `__all__` contents were checked in source.

**The core change is sound and I verified its central claim.** Every consumer of a backward record in the touched lanes reads only the committed four fields — `loss`, `gradient`, `selected_count`, `gradient_finite` (on_policy `1243/1246/1397/1479-1481`, replay `1227/1230/1379/1445-1448`, sparse-dev `750-752/848-851/890-893/1193-1197`). After the import swap, no dangling `KondoActorBackwardResult` reference remains in the two lanes that drop the import (remaining uses on main are exactly the sites this diff rewrites), and the `KondoActorBackwardResult | KondoActorBackwardSummary` union on `_backward_record` is right because the real-kernel paths (`kondo_actor_backward_kernel` at sparse-dev 848/890/1193) still flow full six-field results through it. Typing `_ActorOutcome.backward`/`_ActorArmOutcome.backward` as `KondoActorBackwardSummary` only (rather than a union) is a nice precision touch — those dataclasses are only ever constructed from the committed transaction after this change. The option-2 design keeps the core producer's six-argument constructor fail-loud, which is what issue #7 asked for.

**Blocking: this branch fails an existing unit-lane test its evidence list did not run.** `tests/test_kondo_sparse_actor_public_exports.py::test_every_declared_sparse_actor_symbol_is_publicly_reexported` iterates `kondo_sparse_actor.__all__` and asserts each name appears in **both** `alberta_framework.core.__all__` and `alberta_framework.__all__`. This diff adds `KondoActorBackwardSummary` to the module `__all__` (kondo_sparse_actor.py `+1320`) but re-exports it through neither package surface, so that test's `assert name in core.__all__` fails on this head. The PR's evidence enumerates `test_kondo_sparse_actor`, the three development/integration lane files, and the two touched integration modules — the public-exports file is not among them, which is how this was missed. Fix is two files: add the symbol to the import block and `__all__` list in `alberta_framework/core/__init__.py` (main: 986/2672 region) and `alberta_framework/__init__.py` (main: 1011/2835 region).

**Coordination note for the maintainer:** PR #9 (opened ~1h before this one) implements the same issue-#7 option 2 with the identical type name and *does* include both package re-exports (its evidence lists the export-parity test passing). The two PRs are otherwise near-identical; this one's distinct value is the narrower `Summary`-only outcome annotations and the fuller docstring. They cannot both merge as-is (textual conflict in all four files). A reasonable consolidation is either PR plus the other's delta — I've left the mirrored note on #9.

**Process gaps against this repository's evidence rules:** the body carries no `<!-- evidence-head:<sha> -->` marker binding the evidence to the measured head, no `<!-- evidence-row:logs -->` row with an immutable attachment URL for the baseline-fail/candidate-pass pytest output, and no exact AI provider/model disclosure. All three are cheap to add and are what makes the claim verifiable after merge.

Also confirming the author's baseline claim independently: the 6 integration-test failures reproduce on trusted `main` exactly as issue #7 describes (TypeError on the two missing constructor arguments), so the defect and the chosen repair boundary are real.

— [kenjohnscreates]

lalalune:
Superseded by merged PR #9. Both target issue #7, but #13 omits the required `alberta_framework.core` and root-package re-exports for the new module `__all__` member, so `tests/test_kondo_sparse_actor_public_exports.py` fails. It also narrows two outcome annotations even though manual arms still store the full backward result. PR #9 includes the complete export and union-type contract.

PR #5, opened 2026-08-14
fix(core): enforce routed executor failure atomicity before any state read

FreeSolDev:
Fixes #2 — the three `Tests (server)`-blocking failures in
`tests/test_hccl_routed_continual_dyad_operational_runner.py`.

## The contract decision the issue asked for

**Reading 2 (the tests are right): `step()` must not dereference owned-state
attributes before its preparation and checkpoint gates.** Three grounds:

1. The module's own contract documents the order — preparation, validation,
   binding, then publication ("Compute and bind one result to the owned source
   before publication"), and the executor's failure-atomicity assertions
   (`state`/`absolute_step` unchanged on any failure) are only meaningful if a
   malformed source cannot detonate outside a typed boundary.
2. The clock read that moved ahead of preparation gains nothing from being
   early: the owned source is a frozen chex pytree, so its
   `step_words`/`content_token` are identical before and after the dispatch.
   Real-state behavior is byte-identical under this reordering.
3. The tests' expected stage errors (`routed-preparation`,
   `periodic-checkpoint`, `final-checkpoint`) match the runner's existing typed
   vocabulary; the raw `AttributeError` escape was the only path that violated
   it.

## What changed (one function, `_RoutedExecutor.step`)

New gate order: bounds → preparation dispatch (non-domain errors wrapped as
`routed-preparation`, unchanged) → result-shape gate (unchanged) →
**checkpoint-stage candidate validation** (moved before any source/transcript
dereference) → **source + transcript reads and binding check** (moved after the
gates; non-domain errors in the read window now wrap as
`source-transcript-binding`, the same error the equality check already used) →
checkpoint work annotation → atomic publication. The docstring now states the
atomicity contract explicitly.

No other file changed; the failing tests themselves are untouched — they were
already asserting this contract.

## Verification

- **Both ways on the target suite** (local, Python 3.12, quiet lane):
  `main` → `3 failed, 6 passed` (the exact three from #2); with this change →
  `9 passed`.
- `ruff check` and strict `mypy` clean on the changed file.
- CI: this PR's own run is the authoritative harness. A pre-flight dispatch on
  my fork (https://github.com/FreeSolDev/asi/actions/runs/31754292872) had all
  four jobs killed by runner shutdown signals ~25 minutes in (infra, not
  tests — ruff printed "All checks passed!" before the kill and the visible
  pytest output was green to the point of cancellation); a rerun is in flight
  there as secondary evidence.

[...]

Issue #2, opened 2026-08-13
Routed dyad executor: 3 failing tests block CI (preparation-failure wrapping contract)


lalalune:
Three tests fail on `main`, so the CI test lane is red. They predate the extraction from the research tree (present at the imported `updates` commit), and the lint lane that was masking them is now fixed.

```
FAILED tests/test_hccl_routed_continual_dyad_operational_runner.py::test_routed_executor_does_not_publish_when_preparation_fails
FAILED tests/test_hccl_routed_continual_dyad_operational_runner.py::test_routed_executor_checks_candidate_before_publication[0-1-periodic-checkpoint]
FAILED tests/test_hccl_routed_continual_dyad_operational_runner.py::test_routed_executor_checks_candidate_before_publication[9-None-final-checkpoint]
```

### Root cause

```
E  AttributeError: 'object' object has no attribute 'hccl_state'
   alberta_framework/core/hccl_routed_continual_dyad_operational_runner.py:1007
```

`_RoutedExecutor.step` reads the source clock —

```python
source = self._state
source_words = _host_array(
    source.hccl_state.world_state.step_words, ...
)
```

— **before** dispatching to `_execute_routed_operational_event`. The tests construct the executor with a bare `object()` state and monkeypatch that dispatch to raise, expecting the failure to surface as `HCCLRoutedContinualDyadOperationalError`. Because the clock read happens first, a raw `AttributeError` escapes instead, and the wrapping contract the tests assert is never exercised.

### The decision this needs

Someone who owns this lane has to say which side is authoritative:

1. **The implementation is right** — the clock read legitimately moved ahead of preparation, and the tests are stale. They should construct a real (or minimally structured) state rather than `object()`.
2. **The tests are right** — `step` should not touch state attributes before the preparation boundary, or it should wrap non-domain exceptions raised in that window into `HCCLRoutedContinualDyadOperationalError`.

I did not guess, because the two readings imply different failure-atomicity guarantees for the routed executor, and `CONTINUAL_DYAD_BENCHMARK.md` describes this as a staged design.

This is a well-bounded first contribution: reproduce with
`.venv/bin/python -m pytest tests/test_hccl_routed_continual_dyad_operational_runner.py -q -o addopts=""`, decide the contract, and make the tests exercise it.

lalalune:
Updating this with what a full measured run actually shows, now that the CI jobs stop being killed mid-suite (they were dying at exit 143 between 30% and 40% through, which hid all of this).

[...]

PR #16, opened 2026-08-14
fix(build): raise the jax floor to the 0.8.1 decorator-factory requirement

### LINKED ISSUE #15: pyproject declares jax>=0.4 but the import graph requires the 0.8.1 jit decorator-factory
## Summary

`pyproject.toml` declares `jax>=0.4` / `jaxlib>=0.4`, but the tree does not import under JAX 0.4.x: two modules use the `jax.jit` decorator-factory pattern, which first shipped in **JAX 0.8.1** (CHANGELOG, Nov 18 2025: "`jax.jit` now supports the decorator factory pattern"). The declared floor therefore promises four minor versions the code cannot run on.

## Reproduction (observed, not hypothetical)

On a machine where the resolver legitimately lands on jaxlib 0.4.38 (Intel macOS — the last Intel-mac wheel), a clean `pip install -e .` succeeds but importing `alberta_framework.core` fails at collection:

```text
alberta_framework/core/ensemble_short_rollouts.py:1489: in EnsembleShortRolloutPlanner
    @jax.jit(static_argnums=(0,))
E   TypeError: jit() missing 1 required positional argument: 'fun'
```

This surfaced while working #2 — the declared floor let the environment build, then the import graph refuted it.

## Sites requiring >= 0.8.1

- `alberta_framework/core/ensemble_short_rollouts.py:1489` — `@jax.jit(static_argnums=(0,))`
- `alberta_framework/core/grounded_imagination_composition.py:978` — `@jax.jit(static_argnums=(0,))`

Both are load-bearing at import time via the eager `core/__init__`, so any floor-satisfying old JAX breaks the whole package, not one lane.

## Why CI never catches this

CI resolves the newest compatible JAX, so the floor is never exercised. (A floor-pinned install job would catch it; noting as an option, not requesting it here.)

## Proposed fix

Raise the floors to the highest *proven-required* feature: `jax>=0.8.1`, `jaxlib>=0.8.1`. The true floor may be higher still if newer APIs are in use elsewhere; 0.8.1 is the bound demonstrable from the observed failure plus the changelog, which honestly narrows the gap without overclaiming. Side effect worth stating: jaxlib >= 0.5 has no Intel-macOS wheels, so this makes the (already factual) platform requirement declarative instead of a runtime surprise.

PR follows.

FreeSolDev:
Fixes #15 — a metadata-only floor correction, fully argued (with the observed
failure and changelog citation) in the issue.

## Change

`pyproject.toml`: `jax>=0.4` → `jax>=0.8.1`, `jaxlib>=0.4` → `jaxlib>=0.8.1`.
Nothing else; `requires-python >= 3.12` and the `numpy >= 1.26` floor stay
intact per the repo guide.

## Why 0.8.1 exactly

It is the highest **proven-required** bound: the `jax.jit` decorator-factory
pattern used at `ensemble_short_rollouts.py:1489` and
`grounded_imagination_composition.py:978` first shipped in JAX 0.8.1 (its
changelog names the feature), and both sites are import-time load-bearing
through the eager `core/__init__`. Under a legitimately-resolved jaxlib 0.4.38
the package installs cleanly and then fails at import
(`TypeError: jit() missing 1 required positional argument: 'fun'` — hit while
working #2). If a newer API elsewhere pushes the true floor higher, that would
merit its own evidence; this PR narrows the declared floor to what is
demonstrable without overclaiming.

## Verification

Metadata-only: this PR's own CI run (fresh resolve + full import graph +
suite) is the authoritative proof, since any floor-satisfying resolution on CI
lands well above 0.8.1 and must remain unaffected. The failure mode being
fixed is only reachable on platforms where the resolver lands below 0.8.1
(e.g. Intel macOS, whose jaxlib wheels stop at 0.4.38) — on those, the
declared floor now correctly refuses the install instead of building an
environment that cannot import.

## Provenance

AI-assisted: `anthropic/claude-fable-5` via Claude Code, under the slop.cash
`contribute-to-asi` skill flow (installer still refuses activation on the
stale published revision, so no signed run receipt is attachable — disclosed
rather than omitted, same as #5).


lalalune:
Maintainer review confirmed the floor against the JAX 0.8.1 changelog. I added commit 7546b8f to synchronize `uv.lock` and raise the `gpu` extra to the same floor; `uv lock --check` and `git diff --check` now pass. Awaiting the refreshed CI run before merge.

Issue #25, opened 2026-08-14
Reject duplicate experiment names before execution


leo-guinan:
## Problem

At current `main` `b8862fcfb939d3e2a2242d80881cb12a7519da1a`, `run_multi_seed_experiment()` accepts distinct `ExperimentConfig` objects with the same `name`, executes all of them, and then groups every result solely by `SingleRunResult.config_name`.

Current path in `alberta_framework/utils/experiments.py`:

- lines 213–217 schedule every `(config, seed)` pair without validating configuration-name uniqueness;
- `run_single_experiment()` stores only `config.name` in each result;
- lines 246–256 group and aggregate solely by that string.

Configuration identity is therefore discarded before aggregation.

## Deterministic falsifier

Construct two configurations with the same name but different learner factories:

```text
ExperimentConfig(name="baseline", learner_factory=make_lms, ...)
ExperimentConfig(name="baseline", learner_factory=make_idbd, ...)
seeds = [0, 1]
```

If their metric keys and shapes are compatible, the current runner silently produces one aggregate named `"baseline"` with four rows and repeated seed identities:

```text
config_name = "baseline"
seeds = [0, 1, 0, 1]
```

That aggregate is neither the LMS treatment nor the IDBD treatment. If metric schemas or trace lengths differ, failure may occur later during aggregation instead, after the experiments have already run.

The invalid input should be rejected before either learner or stream factory is called.

## Observed reachability boundary

GitHub code search at this exact head finds no current test use of `ExperimentConfig` or `run_multi_seed_experiment`. The repository has no dedicated experiment-runner validation module.

This is a latent experiment-identity and compute-waste defect in a public utility. I found no committed artifact whose validity can be attributed to this path, and this issue does not claim that an existing ASI result is invalid.

## Proposed bounded fix

Change only the configuration-name contract at the public runner boundary:

1. At the start of `run_multi_seed_experiment()`, require every `ExperimentConfig.name` to be unique.
2. Reject duplicates with a deterministic, descriptive `ValueError` that names the duplicate value or values.
3. Perform this validation before seed expansion, task construction, imports of optional execution helpers, learner construction, stream construction, or parallel dispatch.
4. Preserve behavior for a valid unique-name sequence, including return ordering, seed handling, parallel/sequential execution, aggregation, and result schema.

Do not add duplicate-seed validation, alter `aggregate_metrics()`, change metric-window statistics, repair single-seed intervals, or touch benchmark/evidence code in this change.

## Tests first

[...]

Issue #24, opened 2026-08-14
Fix paired comparisons to align by seed identity


leo-guinan:
## Problem

At current `main` `b8862fcfb939d3e2a2242d80881cb12a7519da1`, `pairwise_comparisons()` performs paired t-tests and Wilcoxon tests by metric-row position, even though each `AggregatedResults` carries the seed identity for every row.

Current path in `alberta_framework/utils/statistics.py`:

- lines 433–440 reduce each `(n_seeds, n_steps)` metric array to one value per row;
- lines 451–479 pass those row-ordered arrays directly to paired t/Wilcoxon;
- `agg.seeds` is never validated or used.

Equal seed sets in different orders are therefore paired incorrectly. Different same-length seed sets are accepted as paired observations. Duplicate seeds are also accepted.

This can corrupt paired differences, p-values, effect sizes, and corrected significance decisions while still labeling the result paired.

## Deterministic falsifier

Use one-step metric arrays:

```text
A seeds:  [0, 1, 2]
A values: [0, 10, 20]

B seeds:  [2, 1, 0]
B values: [21, 11, 1]
```

Correct seed-aligned `A - B` differences:

```text
seed 0:  0 - 1  = -1
seed 1: 10 - 11 = -1
seed 2: 20 - 21 = -1
```

Current positional differences:

```text
[0 - 21, 10 - 11, 20 - 1] = [-21, -1, 19]
```

Both arrays have length three, so the current paired path accepts the wrong pairing silently.

## Observed reachability boundary

GitHub code search at this exact head finds only:

- the function definition;
- its public re-export;
- tests in `tests/test_statistics_validation.py`.

I found no current non-test production caller. This is a latent measurement-integrity defect in a public statistics utility, not evidence that an existing ASI result is invalid. In particular, this issue makes no claim about the separately implemented pairing path used by the current IPMNIST development record.

## Proposed bounded fix

Change only the seed-identity contract in `pairwise_comparisons()`:

1. Validate that every `AggregatedResults.seeds` list is unique and has the same length as the metric array's first axis.
2. For paired t-tests and Wilcoxon:
   - require equal seed sets for each method pair;
   - align the second method's values to the first method's seed order before testing;
   - reject unequal sets with a descriptive `ValueError` rather than downgrading silently to an unpaired test.
3. Keep Mann–Whitney unpaired:
   - do not require equal seed sets or align by seed identity;
[...]

PR #28, opened 2026-08-14
fix(experiments): reject duplicate seeds before execution

### LINKED ISSUE #27: Reject duplicate seeds before multi-seed execution
## Problem

At current `main` `ee075f85d95211208f3590b65eb9f0e7571e83ca`, `run_multi_seed_experiment()` validates configuration-name identity (#25) but accepts a duplicate explicit seed list. Runs are seed-deterministic — each task derives its stream from `jr.key(seed)` with fresh factories — so a duplicated seed re-executes the identical trajectory and aggregates it as if it were an additional independent sample.

## Deterministic falsifier

For two 50-step `LinearLearner`/`RandomWalkStream` diagnostic configs:

```python
results = run_multi_seed_experiment(configs, seeds=[0, 0, 1], parallel=False, show_progress=False)
```

completes without error at the stated head and produces, for each config:

```text
reported seeds: [0, 0, 1]
metric rows: 3
row0 == row1 elementwise: True
row0 == row2 elementwise: False
summary.n_seeds: 3
summary.std (with duplicate): 2.6266074239
honest two-seed std:          2.7859378814
```

The duplicated row enters `MetricSummary` as a third sample: `n_seeds` is overstated and the reported spread is whatever the duplicated draw happens to make it — pseudo-replication, not measurement.

The same aggregates are then refused one layer down: after #24's fix,

```python
pairwise_comparisons(results, metric="squared_error", test="ttest")
```

raises `AggregatedResults 'baseline' contains duplicate seeds`. The runner boundary manufactures exactly the aggregates the statistics boundary rejects, and summary-only consumers (`summary`, `get_final_performance`) never hit that guard at all.

## Reachability and evidence boundary

`run_multi_seed_experiment` is the public multi-seed entry point, re-exported from `alberta_framework.utils`. Every other seed-consuming public boundary in this repository already rejects duplicate seeds explicitly (`benchmarks/forager.py`, `benchmarks/causal_map_forager.py`, `benchmarks/foragax_open_screen.py`, `benchmarks/forager_matched_statistics.py`, `recurring_feature_gate.py`); this runner is the outlier. I found no committed `outputs/` artifact produced through a duplicate-seed call, and this issue does not claim any existing ASI result is invalid.

## Proposed bounded fix

Validate the explicit seed list at the same public runner boundary where duplicate names are rejected, before any learner or stream factory executes:

1. Detect duplicates after seed-list normalization, before task construction.
2. Report deterministically, mirroring #25's message contract: `Experiment seeds must be unique; duplicates: 0, 7` (sorted).
3. The integer count form (`seeds=30`) generates `range(n)` and is unaffected.
4. `aggregate_metrics()` is unchanged — it is not the public runner boundary, and callers constructing `SingleRunResult` lists directly are a separate contract, as is #26's timeseries interval scope.

## Tests first

Add to `tests/test_experiments_validation.py`, in the existing contract-test style:

- a duplicate seed list rejects before any learner/stream factory executes (`_fail_if_called` factories);
- multiple duplicate seeds are reported deterministically in sorted order;
- existing unique-name, order-preservation, and empty-config tests remain unchanged.

## Acceptance

- The new regression tests fail at the stated baseline and pass after the fix.
- `tests/test_experiments_validation.py` and `tests/test_statistics_validation.py` pass; repository Ruff and strict mypy pass; `git diff --check` passes.
- No benchmark seed, learner run, `outputs/` artifact, scientific evidence, or promotion claim is created or modified.

This is one deterministic measurement-integrity Fix, completing the identity-contract line of #24 (seed alignment in paired tests) and #25 (config-name identity) at the runner boundary itself. A code PR with the failing-test-first evidence follows immediately.

Prabhat1308:
Closes #27.

## Issue

`run_multi_seed_experiment()` validates configuration-name identity (#25) but not seed identity. Runs are seed-deterministic — each task derives its stream from `jr.key(seed)` with fresh factories — so a duplicated entry in an explicit seed list re-executes the identical trajectory and `aggregate_metrics` counts it as an additional independent sample. After #24, `pairwise_comparisons` rejects exactly those aggregates for duplicate seeds, but summary-only consumers (`MetricSummary`, `get_final_performance`) have no guard. Every other seed-consuming public boundary in the repository (forager, causal-map forager, foragax open screen, matched statistics, recurring feature gate) already rejects duplicate seeds; this runner was the outlier.

## Symptoms

At `main` `ee075f85d95211208f3590b65eb9f0e7571e83ca`, `seeds=[0, 0, 1]` on a 50-step `LinearLearner`/`RandomWalkStream` diagnostic completes silently and reports pseudo-replicated aggregates (full falsifier in #27):

```text
run_multi_seed_experiment(seeds=[0, 0, 1]) completed WITHOUT error
reported seeds: [0, 0, 1]
row0 == row1 elementwise: True
summary.n_seeds: 3
summary.std (with duplicate): 2.6266074239
honest two-seed std:          2.7859378814
pairwise_comparisons on the SAME aggregates: rejected with: AggregatedResults 'baseline' contains duplicate seeds
```

## New state

A duplicate explicit seed list is rejected at the public runner boundary before any learner or stream factory executes, with #25's deterministic message contract: `Experiment seeds must be unique; duplicates: 3, 7` (sorted). The integer count form (`seeds=30`) generates `range(n)` and is unaffected; `aggregate_metrics()` is out of scope per #27.

Failing-test-first evidence, measured at head `fa41b6c1b3caa94d347b4bd2c9e901ec62853d92` (baseline `ee075f85`):

- red phase: the two new contract tests fail on the unmodified baseline (`_fail_if_called` factories execute) — `2 failed, 5 passed`;
- green phase: `.venv/bin/python -m pytest tests/test_experiments_validation.py tests/test_statistics_validation.py -q -o addopts=""` — `48 passed` (7 + 41);
- `.venv/bin/python -m ruff check .` — clean; `.venv/bin/python -m mypy` — clean, 159 source files, strict py312; `git diff --check` — clean.

No benchmark seed, learner run, `outputs/` artifact, scientific evidence, or promotion claim is created or modified. This session is CLI-only and cannot mint immutable GitHub user-attachment URLs, so the run output above is inline; the falsifier script in #27 reproduces it deterministically at the stated SHAs.

AI-assisted: `anthropic/claude-fable-5` via Claude Code under the slop.cash `contribute-to-asi` skill flow; the signed local-usage receipt footer follows.

```text
Compute receipt: 134308 project-attributed tokens (exact; device-signed, locally reported)
— [prabhat1308]
```

### DIFF
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -196,7 +196,8 @@ def run_multi_seed_experiment(
 
     Args:
         configs: List of experiment configurations to run. Names must be unique.
-        seeds: Number of seeds (generates 0..n-1) or explicit list of seeds
+        seeds: Number of seeds (generates 0..n-1) or explicit list of seeds.
+            Explicit seeds must be unique.
         parallel: Whether to use parallel execution (requires joblib)
         n_jobs: Number of parallel jobs (-1 for all CPUs)
         show_progress: Whether to show progress bar (requires tqdm)
@@ -205,7 +206,8 @@ def run_multi_seed_experiment(
         Dictionary mapping config name to AggregatedResults
 
     Raises:
-        ValueError: If two or more configurations have the same name
+        ValueError: If two or more configurations have the same name, or if the
+            explicit seed list contains duplicates
     """
     seen_names: set[str] = set()
     duplicate_names: set[str] = set()
@@ -226,6 +228,18 @@ def run_multi_seed_experiment(
         seed_list = list(range(seeds))
     else:
         seed_list = list(seeds)
+
+    seen_seeds: set[int] = set()
+    duplicate_seeds: set[int] = set()
+    for seed in seed_list:
+        if seed in seen_seeds:
+            duplicate_seeds.add(seed)
+        else:
+            seen_seeds.add(seed)
+
+    if duplicate_seeds:
+        formatted_seeds = ", ".join(str(seed) for seed in sorted(duplicate_seeds))
+        raise ValueError(f"Experiment seeds must be unique; duplicates: {formatted_seeds}")
 
     # Build list of (config, seed) pairs
     tasks: list[tuple[ExperimentConfig, int]] = []
--- a/tests/test_experiments_validation.py
+++ b/tests/test_experiments_validation.py
@@ -93,6 +93,39 @@ def test_multiple_duplicate_names_are_reported_deterministically() -> None:
     )
 
 
+def test_duplicate_seeds_reject_before_factories_execute() -> None:
+    configs = [
+        _config(
+            "baseline",
+            learner_factory=_fail_if_called,
+            stream_factory=_fail_if_called,
+        )
+    ]
+
+    with pytest.raises(
+        ValueError,
+        match=r"^Experiment seeds must be unique; duplicates: 0$",
+    ):
+        run_multi_seed_experiment(configs, seeds=[0, 0, 1], parallel=False, show_progress=False)
+
+
+def test_multiple_duplicate_seeds_are_reported_deterministically() -> None:
+    configs = [
+        _config(
+            "baseline",
+            learner_factory=_fail_if_called,
+            stream_factory=_fail_if_called,
+        )
+    ]
+
+    with pytest.raises(ValueError) as exc_info:
+        run_multi_seed_experiment(
+            configs, seeds=[7, 3, 7, 0, 3, 7], parallel=False, show_progress=False
+        )
+
+    assert str(exc_info.value) == "Experiment seeds must be unique; duplicates: 3, 7"
+
+
 def test_unique_names_preserve_config_and_seed_order() -> None:
     results = run_multi_seed_experiment(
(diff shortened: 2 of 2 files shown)


Uuriko (approved):
Exact-head review of `fa41b6c1b3caa94d347b4bd2c9e901ec62853d92` against `main` `ee075f85`. This is a Fix (measurement-corruption), not a climb. I read the diff, #27, and the ASI evidence rubric. I did not rerun pytest in this pass.

The guard is in the right place: after `seed_list` is materialized, before factories run. The `_fail_if_called` tests prove that. Integer `seeds=n` still becomes `range(n)` and is unaffected. Message contract matches #25 (`duplicates: 3, 7` sorted). No `outputs/` pin is touched.

Leftover, not a block: body says `aggregate_metrics()` is out of scope. `Closes #27` is fine if that issue is only the runner boundary. If #27 still wants a summary-side guard, drop `Closes` and leave the issue open.

No Shaw-style evidence-head required here (no promoted number). Approve the runner contract.

PR #30, opened 2026-08-14
fix(statistics): reject empty inputs to summary statistics

Prabhat1308:
Closes #29.

## Issue

`compute_statistics` and `compute_timeseries_statistics` have no zero-seed contract. The scalar path crashes accidentally (`IndexError` from `np.percentile` after four `RuntimeWarning`s), while the vectorized sibling silently returns all-NaN mean and confidence bounds shaped like real statistics. #26 documents the `n_seeds == 1` timeseries contract and explicitly scopes zero-seed out as a separate contract; this PR defines that contract, fail-closed, in the style of the degenerate-input rejections already in `metrics.py`.

## Symptoms

At `main` `ee075f85d95211208f3590b65eb9f0e7571e83ca` (full falsifier in #29):

```text
compute_statistics([])
  -> RuntimeWarning: Mean of empty slice (x2), invalid value encountered in scalar divide (x2)
  -> IndexError: index -1 is out of bounds for axis 0 with size 0

compute_timeseries_statistics(np.empty((0, 3)))
  -> (array([nan, nan, nan]), array([nan, nan, nan]), array([nan, nan, nan]))  # returned silently
```

## New state

Both public entry points refuse zero-seed input with a deterministic `ValueError` (`values must be non-empty` / `metric_array must contain at least one seed row`) raised before any NumPy warning is emitted; all `n >= 1` behavior is byte-for-byte unchanged. The `n_seeds == 1` timeseries interval remains #26's reserved scope, untouched.

Failing-test-first evidence, measured at head `e53d420ecda01878b02e19ce6d4b07dc6bab6c61` (baseline `ee075f85`):

- red phase: the three new contract tests fail on the unmodified baseline (`3 failed, 41 passed`), with `warnings.simplefilter("error")` proving the current behavior warns before failing or returning;
- green phase: `.venv/bin/python -m pytest tests/test_statistics_validation.py tests/test_experiments_validation.py -q -o addopts=""` — `51 passed`;
- `.venv/bin/python -m ruff check .` — clean; `.venv/bin/python -m mypy` — clean, 159 source files, strict py312; `git diff --check` — clean.

No benchmark seed, learner run, `outputs/` artifact, scientific evidence, or promotion claim is created or modified. This session is CLI-only and cannot mint immutable GitHub user-attachment URLs, so the run output above is inline; the falsifier in #29 reproduces it deterministically at the stated SHAs.

[...]

Issue #29, opened 2026-08-14
Define the zero-seed contract for summary statistics


Prabhat1308:
## Problem

At current `main` `ee075f85d95211208f3590b65eb9f0e7571e83ca`, the two sibling summary-statistics entry points handle a zero-seed input in two different accidental ways, and neither states a contract:

- `compute_statistics([])` emits four `RuntimeWarning`s (`Mean of empty slice`, `invalid value encountered in scalar divide`) and then crashes with `IndexError: index -1 is out of bounds for axis 0 with size 0` out of `np.percentile` — an undocumented crash from NumPy internals, not a rejection.
- `compute_timeseries_statistics(np.empty((0, 3)))` **returns silently**: `(array([nan, nan, nan]), array([nan, nan, nan]), array([nan, nan, nan]))` — all-NaN mean and confidence bounds shaped exactly like real statistics, with only interpreter-level warnings standing between them and a downstream consumer.

#26 documents the `n_seeds == 1` degenerate contract for the timeseries path and explicitly scopes zero-seed out: "Leave empty-input behavior unchanged; defining or rejecting zero-seed arrays is a separate contract." This issue is that separate contract.

## Deterministic falsifier

```python
import numpy as np
from alberta_framework.utils.statistics import compute_statistics, compute_timeseries_statistics

compute_statistics([])                          # IndexError after 4 RuntimeWarnings
compute_timeseries_statistics(np.empty((0, 3))) # (nan..., nan..., nan...) returned silently
```

Observed at the stated head with NumPy 2.5.1 / SciPy installed.

## Reachability and evidence boundary

Both functions are public API re-exported from `alberta_framework.utils`. The in-repo callers (`utils/visualization.py` confidence bands) receive arrays derived from `AggregatedResults`, which `aggregate_metrics` guarantees non-empty, so I found no committed artifact whose validity depends on this path; this issue does not claim any existing ASI result is invalid. The hazard is the public boundary: a zero-seed matrix from a filtered or misconstructed caller currently yields publication-shaped NaN bands rather than a refusal, which is the same presentation-integrity class as #26 with less data behind it.

## Proposed bounded fix

Fail closed at both entry points, mirroring the repository's degenerate-input style (`metrics.py`: "performance_matrix must be non-empty"):

1. `compute_statistics` raises `ValueError: values must be non-empty` for a zero-length input, before any NumPy warning is emitted.
[...]

PR #32, opened 2026-08-14
fix(statistics): reject identical samples in paired t-test

### LINKED ISSUE #31: Paired t-test silently reports NaN for identical samples
## Problem

At current `main` `ee075f85d95211208f3590b65eb9f0e7571e83ca`, `ttest_comparison(paired=True)` forwards SciPy's undefined result for identical samples straight into `SignificanceResult`: `statistic=nan, p_value=nan, significant=False`, with no exception and — through the `pairwise_comparisons` path — no surfaced warning. The NaN row then renders as literal `nan | nan` in `generate_significance_table` output.

Identical paired samples are not a contrived input in this repository: the house arm-registration convention requires every new mechanism to reduce **bit-for-bit** to an existing arm when its constant is inert. Screening a pinned-inert arm and its base through the same grid and handing the aggregates to `pairwise_comparisons(test="ttest")` produces exactly this pair.

## Deterministic falsifier

Two `AggregatedResults` with identical metric rows on seeds `[0, 1, 2]` plus one distinct candidate, `pairwise_comparisons(..., test="ttest")`:

```text
('base_arm', 'pinned_arm_inert'): statistic=nan, p=nan, sig=False   # zero warnings surfaced
('base_arm', 'candidate'):        statistic=6.93, p=0.0202, sig=False

| Method A | Method B         | Statistic | p-value | Effect Size | Sig. |
| base_arm | pinned_arm_inert | nan       | nan     | 0.000       | No   |
```

Direct call: `ttest_comparison(a, list(a))` → `statistic=nan, p_value=nan` (SciPy `RuntimeWarning: invalid value encountered in scalar divide`).

## Scope boundaries (deliberately untouched)

- A **constant nonzero** difference yields `statistic=inf, p_value=0.0, significant=True`. That signature is deliberately pinned by `test_paired_ttest_aligns_adversarial_seed_order` (47709c1), so it is treated as intended behavior and left unchanged here; if it should also be refused, that is a maintainer decision on an existing test's contract, not something this fix will rewrite.
- `paired=False` on identical samples is well-defined (`t=0, p=1.0`) and unchanged.
- `wilcoxon_comparison` on identical samples returns `p=1.0` under SciPy 1.18 and is unchanged.
- Single-pair (`n=1`) non-identical inputs (undefined `df=0`) are a separate contract.

## Proposed bounded fix

`ttest_comparison` raises a deterministic `ValueError` naming both methods when `paired=True` and the two samples are elementwise identical: `Paired comparison 'base_arm' vs 'pinned_inert' has identical samples; the paired t statistic is undefined`. The refusal propagates through `pairwise_comparisons`, so a grid containing an inert-pin pair is refused loudly instead of publishing a NaN row — the caller drops the inert arm from significance testing deliberately, consistent with #24's fail-closed rejections.

## Tests first

Add to `TestPairedTests` and `TestPairwiseComparisons` in `tests/test_statistics_validation.py`:

- identical paired samples raise the exact `ValueError` (direct call, and through `pairwise_comparisons` with the pair named);
- constant-nonzero-shift and unpaired-identical behavior asserted unchanged, documenting the scope boundary;
- all existing tests, including `test_paired_ttest_aligns_adversarial_seed_order`, byte-for-byte untouched and passing.

## Acceptance

- The new rejection tests fail at the stated baseline (`2 failed, 43 passed`) and pass after the fix.
- `tests/test_statistics_validation.py` and `tests/test_experiments_validation.py` pass; repository Ruff and strict mypy pass; `git diff --check` passes.
- No benchmark seed, learner run, `outputs/` artifact, scientific evidence, or promotion claim is created or modified.

One deterministic measurement-integrity Fix in the #24/#25 identity-contract family. Code PR with failing-test-first evidence follows immediately.

Prabhat1308:
Closes #31.

## Issue

`ttest_comparison(paired=True)` forwards SciPy's undefined identical-samples result into `SignificanceResult` as `statistic=nan, p_value=nan, significant=False`, and `pairwise_comparisons` surfaces no warning on that path. Identical pairs are guaranteed to occur here by the bit-exact reduction-pin convention: an arm whose mechanism constant is inert produces trajectories bit-identical to its base, and screening both through a paired t-test grid yields exactly this input. The NaN row renders as literal `nan | nan` in `generate_significance_table`.

## Symptoms

At `main` `ee075f85d95211208f3590b65eb9f0e7571e83ca` (full falsifier in #31):

```text
('base_arm', 'pinned_arm_inert'): statistic=nan, p=nan, sig=False   # zero warnings surfaced

| Method A | Method B         | Statistic | p-value | Effect Size | Sig. |
| base_arm | pinned_arm_inert | nan       | nan     | 0.000       | No   |
```

## New state

A paired comparison of elementwise-identical samples is refused with a deterministic `ValueError` naming both methods (`Paired comparison 'base_arm' vs 'pinned_inert' has identical samples; the paired t statistic is undefined`), propagating through `pairwise_comparisons` so an inert-pin pair is refused loudly rather than published as a NaN row. Scope boundaries are asserted unchanged by new tests: constant-nonzero-shift behavior (`inf`, `p=0.0` — deliberately pinned by `test_paired_ttest_aligns_adversarial_seed_order` since 47709c1), unpaired identical samples (`t=0, p=1.0`), and Wilcoxon; that alignment test itself is byte-for-byte untouched.

Failing-test-first evidence, measured at head `a1db5edbccdb52aef7fa76fa6e4b3cd3d2c18282` (baseline `ee075f85`):

- red phase: the two rejection tests fail on the unmodified baseline — `2 failed, 43 passed`;
- green phase: `.venv/bin/python -m pytest tests/test_statistics_validation.py tests/test_experiments_validation.py -q -o addopts=""` — `50 passed` at this head (45 statistics + 5 experiments);
- `.venv/bin/python -m ruff check .` — clean; `.venv/bin/python -m mypy` — clean, 159 source files, strict py312; `git diff --check` — clean.

No benchmark seed, learner run, `outputs/` artifact, scientific evidence, or promotion claim is created or modified. This session is CLI-only and cannot mint immutable GitHub user-attachment URLs, so the run output above is inline; the falsifier in #31 reproduces it deterministically at the stated SHAs.

AI-assisted: `anthropic/claude-fable-5` via Claude Code under the slop.cash `contribute-to-asi` skill flow; the signed local-usage receipt footer follows.

```text
Compute receipt: 179042 project-attributed tokens (exact; device-signed, locally reported)
— [prabhat1308]
```

### DIFF
--- a/alberta_framework/utils/statistics.py
+++ b/alberta_framework/utils/statistics.py
@@ -215,9 +215,19 @@ def ttest_comparison(
 
     Returns:
         SignificanceResult with test results
+
+    Raises:
+        ValueError: If a paired comparison receives identical samples, for
+            which the paired t statistic is undefined
     """
     a = np.asarray(values_a)
     b = np.asarray(values_b)
+
+    if paired and a.size > 0 and np.array_equal(a, b):
+        raise ValueError(
+            f"Paired comparison {method_a!r} vs {method_b!r} has identical "
+            "samples; the paired t statistic is undefined"
+        )
 
     try:
         from scipy import stats
--- a/alberta_framework/utils/statistics.py
+++ b/alberta_framework/utils/statistics.py
@@ -73,9 +73,14 @@ def compute_statistics(
 
     Returns:
         StatisticalSummary with all statistics
+
+    Raises:
+        ValueError: If values is empty
     """
     arr = np.asarray(values)
     n = len(arr)
+    if n == 0:
+        raise ValueError("values must be non-empty")
 
     mean = float(np.mean(arr))
     std = float(np.std(arr, ddof=1)) if n > 1 else 0.0
@@ -129,8 +134,13 @@ def compute_timeseries_statistics(
 
     Returns:
         Tuple of (mean, ci_lower, ci_upper) arrays of shape (n_steps,)
+
+    Raises:
+        ValueError: If metric_array has no seed rows
     """
     n_seeds = metric_array.shape[0]
+    if n_seeds == 0:
+        raise ValueError("metric_array must contain at least one seed row")
     mean = np.mean(metric_array, axis=0)
     std = np.std(metric_array, axis=0, ddof=1)
     sem = std / np.sqrt(n_seeds)
--- a/tests/test_statistics_validation.py
+++ b/tests/test_statistics_validation.py
@@ -221,6 +221,27 @@ class TestPairedTests:
         # every a_i < b_i: effect size sign must reflect a < b
         assert res.effect_size < 0.0
 
+    def test_paired_ttest_rejects_identical_samples(self) -> None:
+        values = [0.91, 0.88, 0.95]
+        with pytest.raises(
+            ValueError,
+            match=r"^Paired comparison 'pin' vs 'base' has identical samples; "
+            r"the paired t statistic is undefined$",
+        ):
+            ttest_comparison(values, list(values), paired=True, method_a="pin", method_b="base")
+
+    def test_constant_nonzero_shift_stays_out_of_scope(self) -> None:
+        res = ttest_comparison([1.0, 2.0, 3.0], [0.5, 1.5, 2.5], paired=True)
+        assert np.isposinf(res.statistic)
+        assert res.p_value == 0.0
+
+    def test_unpaired_identical_samples_stay_out_of_scope(self) -> None:
+        values = [0.91, 0.88, 0.95]
+        res = ttest_comparison(values, list(values), paired=False)
+        assert res.test_name == "independent t-test"
+        assert res.p_value == pytest.approx(1.0)
+        assert not res.significant
+
     def test_unpaired_ttest_separated_groups(self) -> None:
         rng = np.random.default_rng(5)
         a = rng.normal(10.0, 0.5, size=15)
@@ -485,6 +506,19 @@ class TestPairwiseComparisons:
 
         with pytest.raises(ValueError, match="equal seed sets"):
             pairwise_comparisons({"a": a, "b": b}, test=test, window=1)
+
+    def test_reduction_pin_pair_rejected_for_paired_ttest(self) -> None:
+        rows = [0.10, 0.11, 0.09]
+        results = {
+            "base_arm": _make_seeded_aggregated("base_arm", [0, 1, 2], rows),
+            "pinned_inert": _make_seeded_aggregated("pinned_inert", [0, 1, 2], list(rows)),
+        }
+        with pytest.raises(
+            ValueError,
+            match=r"^Paired comparison 'base_arm' vs 'pinned_inert' has identical "
+            r"samples; the paired t statistic is undefined$",
+        ):
+            pairwise_comparisons(results, metric="squared_error", test="ttest")
 
     def test_duplicate_seeds_rejected(self) -> None:
         malformed = _make_seeded_aggregated("malformed", [0, 0, 1], [0.0, 1.0, 2.0])
--- a/tests/test_statistics_validation.py
+++ b/tests/test_statistics_validation.py
@@ -26,6 +26,8 @@ Calibration (measured on this machine, scripts in the session scratchpad):
   superset always and strictness on >= 50 draws.
 """
 
+import warnings
+
 import numpy as np
 import pytest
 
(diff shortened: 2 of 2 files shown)


Uuriko (commented):
Exact-head review of `a1db5edbccdb52aef7fa76fa6e4b3cd3d2c18282`. Fix, not a climb. Diff + #31 + rubric. Pytest not rerun in this pass.

Rejecting elementwise-identical paired samples with a named `ValueError` is the right fail-closed answer to the reduction-pin NaN row. Scope tests for constant-nonzero-shift (`inf`, `p=0`) and unpaired identical samples (`t=0`, `p=1`) look preserved. `a.size > 0` means empty identical arrays still fall through to SciPy — that is #30's contract, not this one.

Hold: this and #30 both edit `statistics.py` from `33b8c8a`. Merge one, rebase the other. `Closes #31` is OK if the issue is only the paired-identical NaN.

No promoted number, no pinned `outputs/` edit.

Prabhat1308:
Rebased onto #30's branch per the hold: this PR is now stacked and contains `e53d420` (#30, unchanged at its reviewed head) followed by `df2aae6` (this change, previously `a1db5ed`, content-identical replay). Merge order is #30 first; once it lands, this PR's diff collapses to `df2aae6` alone and merges without a further rebase.

Re-verified at the stacked head `df2aae6924344b13dca9b8f11bd2f46c3c0a286a`: `pytest tests/test_statistics_validation.py tests/test_experiments_validation.py -q -o addopts=""` — `53 passed` (48 + 5); repo-wide Ruff clean; strict mypy clean (159 files); `git diff --check` clean.

On the `a.size > 0` note — agreed, empty identical arrays were deliberately left to #30's contract; in the stacked branch that case is now refused by `e53d420`'s guard before this one is reached. And confirming the `Closes #31` scope: the issue is scoped to exactly the paired-identical NaN contract, with constant-nonzero-shift and unpaired-identical behavior recorded there as deliberately-unchanged boundaries.

Uuriko (approved):
Exact-head review of `df2aae6924344b13dca9b8f11bd2f46c3c0a286a` (previous Uuriko comment was on `a1db5ed`; this head is the rebase). Fix for #31, not a climb.

Paired `ttest_comparison` now raises when `paired` and the samples are elementwise identical, instead of returning a NaN statistic. Unpaired identical samples and a constant nonzero shift stay out of scope, with tests. `pairwise_comparisons` pin-pair fixture is rejected on the same message. Also carries the #29 empty-input guards. No `outputs/` artifacts.

Approve.

kenjohnscreates (approved):
Independent review at head `df2aae6924344b13dca9b8f11bd2f46c3c0a286a`, static against trusted `main` (`ee075f8`) plus this diff; the one dynamic probe below executed trusted main code + scipy only, not this branch.

**Approving — the fix is correct and the tests are the right shape.** The identical-samples refusal sits at the right layer (`ttest_comparison` itself, so direct callers are covered, and `pairwise_comparisons` inherits it), the `a.size > 0` guard keeps the message honest for empties, the constant-shift and unpaired cases are correctly scoped out with pinned tests, and the reduction-pin pair test reproduces exactly the screening situation the issue names — an inert-constant arm paired against its base. The warnings-as-errors discipline in the new tests is a nice touch since the original defect was a *silent* NaN.

**Verified adjacent gap the maintainer should know about (follow-up scope, not blocking this PR):** the same undefined-statistic class survives through the Wilcoxon path this table exposes. `pairwise_comparisons(test="wilcoxon")` → `wilcoxon_comparison` → `scipy.stats.wilcoxon(a, b)`, and on elementwise-identical samples scipy 1.18.0 returns `statistic=0.0, pvalue=1.0` while emitting a swallowed `RuntimeWarning: invalid value encountered in scalar divide` (probed on trusted main: `stats.wilcoxon([1,2,3],[1,2,3])` → `WilcoxonResult(statistic=0.0, pvalue=1.0)` + that warning). So the identical reduction-pin pair that now fails loud under `test="ttest"` still silently renders a defined-looking `0.0 | 1.0` row under `test="wilcoxon"` — the exact rationale of this PR ("the paired statistic is undefined") applies there verbatim. I am filing that as a separate scoped issue with a fix; happy to fold it here instead if the author prefers.

Minor, non-blocking:

- This branch is a strict superset of #30 (the `compute_statistics`/`compute_timeseries_statistics` empty-input hunks and their tests appear identically in both). Merging this one first makes #30 an empty diff; merge order or a rebase note would keep #29/#31 closing references tidy.
- `ttest_comparison` itself still accepts empty inputs (the `a.size > 0` guard deliberately skips them, and #30's guards cover the summary helpers only) — empty paired inputs still fall through to scipy's NaN path. Worth a one-line refusal someday for symmetry with the rest of this validation wave.

— [kenjohnscreates]

Issue #35, opened 2026-08-14
Reject one-sample paired t-test and Cohen's d instead of divide-by-zero


Uuriko:
## Problem

At current `main` `ee075f85d95211208f3590b65eb9f0e7571e83ca`, `ttest_comparison()` and `cohens_d()` accept length-1 samples.

- paired `ttest_comparison([1.0], [2.0])` hits SciPy `ttest_rel` with `df = 0` and emits `RuntimeWarning: divide by zero encountered in divide`, then `cohens_d` raises `ZeroDivisionError` on `n_a + n_b - 2 == 0`
- unpaired length-1 vs length-1 hits `invalid value encountered in scalar divide` in the same stack
- `cohens_d([1.0], [2.0])` itself is `ZeroDivisionError: float division by zero` on the pooled-std denominator

This is the one-sample sibling of #31 (identical paired samples → NaN) and #26 (one-seed timeseries → NaN CI). A valid 1-vs-1 comparison is not a defined paired t statistic or Cohen's d, but the public helpers crash or warn instead of rejecting.

## Deterministic falsifier

```python
from alberta_framework.utils.statistics import ttest_comparison, cohens_d

ttest_comparison([1.0], [2.0], paired=True)   # RuntimeWarning then ZeroDivisionError
cohens_d([1.0], [2.0])                        # ZeroDivisionError
```

`n >= 2` paths and Mann–Whitney on length-1 samples stay out of scope (MW already returns a defined `p=1` / effect `-1` for `[1]` vs `[2]`).

## Proposed bounded fix

Change only the one-sample contract:

1. `cohens_d`: if either group has `n < 2`, raise a descriptive `ValueError` (pooled df is zero). Do not return `0.0` and pretend the effect is defined.
2. `ttest_comparison`: if `paired` and `len(a) < 2` (or lengths differ), raise `ValueError` before SciPy. If unpaired, require `len(a) >= 2` and `len(b) >= 2`.
3. Do not change #31 identical-sample behavior (that's #32), #26 one-seed timeseries, #29 zero-seed, or #33 spread `ddof`.
4. Wilcoxon / Mann–Whitney contracts stay as they are in this change.

## Tests first

Add focused coverage in `tests/test_statistics_validation.py`:

- `cohens_d([1.0], [2.0])` raises `ValueError`
- paired `ttest_comparison` on length-1 samples raises `ValueError` with no `RuntimeWarning`
[...]

Issue #33, opened 2026-08-14
Spread estimator is inconsistent across summary, export, and screening paths


Prabhat1308:
## Problem

At current `main` `ee075f85d95211208f3590b65eb9f0e7571e83ca`, the same quantity — the spread of a metric across seeds — is computed with two different estimators depending on the path:

- `compute_statistics()` uses the **sample** standard deviation (`ddof=1`), asserted by `test_hand_computed_fixture`;
- the IPMNIST screening merge uses `ddof=1` for both its per-arm stderr and its `paired_vs_control.stderr_diff`;
- but `aggregate_metrics()` fills `MetricSummary.std` with the **population** standard deviation (`np.std`, `ddof=0`), and that value is what every publication surface prints: the `std` column of `export_to_csv`, the `±` in `generate_latex_table` and `generate_markdown_table`, `results_to_dataframe`, and the error bars in `plot_final_performance_bars`. `get_final_performance` and `compare_learners` are `ddof=0` as well.

## Deterministic falsifier

For per-seed final values `[5.631773, 5.631773, 0.059898]`-style small-n data (n=3), `MetricSummary.std` and `compute_statistics(...).std` disagree by the factor `sqrt((n-1)/n)` — an 18.4% understatement of the sample spread at n=3, 8.7% at n=6. Concretely:

```python
values = [0.10, 0.12, 0.30]
np.std(values)          # 0.0899  <- what the LaTeX/markdown/CSV "±" prints
np.std(values, ddof=1)  # 0.1102  <- what compute_statistics and the screening merge use
```

The evidence rubric requires "baseline and candidate as mean and spread"; a reader comparing a utils-exported table against a screening summary is silently comparing two different spread definitions, and the publication path prints the smaller one.

## Why this is an issue and not a PR

The fix direction is a convention decision, not a bug with a single defensible resolution: either `MetricSummary`/`get_final_performance`/`compare_learners` move to `ddof=1` to match `compute_statistics` and the screening merge (changes every printed `±` slightly), or the `ddof=0` choice is intended as a descriptive population summary and deserves an explicit statement plus a label distinguishing it from the inferential paths. Per the repository's collaboration norms I am not choosing unilaterally; whichever direction is confirmed, I will follow with a failing-test-first code-only PR in the #24/#25/#27/#29/#31 pattern.

No committed `outputs/` artifact is invalidated by this report, and no existing ASI result is claimed wrong — the screening lane's own summaries already use `ddof=1` throughout.

[...]

Issue #26, opened 2026-08-14
Return point intervals for single-seed time series


leo-guinan:
## Problem

At current `main` `ee075f85d95211208f3590b65eb9f0e7571e83ca`, `compute_timeseries_statistics()` returns all-NaN confidence bounds for a valid one-seed metric trajectory.

The scalar companion already defines the intended degenerate contract: `compute_statistics([4.2])` returns standard deviation `0`, standard error `0`, and the point interval `[4.2, 4.2]`.

The vectorized time-series path instead always computes:

```python
std = np.std(metric_array, axis=0, ddof=1)
t_value = stats.t.ppf((1 + confidence_level) / 2, n_seeds - 1)
```

For `n_seeds == 1`, the sample standard deviation has zero degrees of freedom and the Student-t quantile uses `df=0`.

## Deterministic falsifier

For:

```python
metric_array = np.asarray([[1.0, 2.0, 3.0]], dtype=np.float64)
```

the current calculation produces:

```text
mean      = [1.0, 2.0, 3.0]
std       = [NaN, NaN, NaN]
t_value   = NaN
ci_lower  = [NaN, NaN, NaN]
ci_upper  = [NaN, NaN, NaN]
```

NumPy also emits `RuntimeWarning: Degrees of freedom <= 0 for slice` and `RuntimeWarning: invalid value encountered in divide`.

A one-seed diagnostic has no estimate of between-seed uncertainty, but its bounded representation should be the observed point trajectory:

```text
mean      = [1.0, 2.0, 3.0]
ci_lower  = [1.0, 2.0, 3.0]
ci_upper  = [1.0, 2.0, 3.0]
```

This is a degenerate point interval, not a claim that uncertainty is known to be zero.

## Reachability and evidence boundary

GitHub code search at this exact head finds:

- the function definition and public re-export;
- existing validation tests;
- two callers in `alberta_framework/utils/visualization.py`, including the learning-curve confidence band.

A one-seed diagnostic plot can therefore receive NaN band coordinates. I found no committed artifact whose validity can be attributed to this path, and this issue does not claim that an existing ASI result is invalid.

`get_metric_timeseries()` is a separate mean-plus-population-standard-deviation helper and already yields a point band for one row; it is not part of this fix.

## Proposed bounded fix

Change only the `n_seeds == 1` contract in `compute_timeseries_statistics()`:

1. Compute the mean trajectory.
[...]

PR #36, opened 2026-08-14
fix(statistics): reject one-sample t-test and Cohen's d with ValueError

Uuriko:
Relates to #35.

## What this fixes

At baseline `main`, `cohens_d()` and `ttest_comparison()` accept length-1 groups they cannot define a statistic for. `cohens_d([1.0], [2.0])` divides by the pooled degrees of freedom `n_a + n_b - 2 == 0` and raises `ZeroDivisionError`; paired `ttest_comparison([1.0], [2.0])` first reaches SciPy `ttest_rel` at `df = 0` and emits `RuntimeWarning: divide by zero encountered in divide` before crashing in `cohens_d`; unpaired 1-vs-1 emits `invalid value encountered in scalar divide` in the same stack. This is the one-sample sibling of #31/#26: the helpers crash or warn instead of rejecting.

## The change

Only the one-sample contract in `alberta_framework/utils/statistics.py`:

- `cohens_d`: either group with `n < 2` raises a descriptive `ValueError` naming both group sizes and the zero pooled degrees of freedom. It does not return `0.0` and pretend the effect is defined.
- `ttest_comparison`: rejects before SciPy — `paired=True` with unequal lengths or fewer than 2 pairs raises `ValueError`; `paired=False` with either group under 2 values raises `ValueError`. No `RuntimeWarning` escapes either path.
- Everything else is untouched, per the issue's scope: #31/#32 identical-sample behavior, #26 one-seed timeseries, #29/#30 zero-seed, #33 spread `ddof`, and the Wilcoxon and Mann–Whitney contracts (`mann_whitney_comparison([1.0], [2.0])` still returns its defined `p=1`, effect `-1`, asserted by a new guard test).

Baseline: `ee075f85d95211208f3590b65eb9f0e7571e83ca` (re-fetched upstream `main` at branch time). Head: `840cac17974916dfcf07f51c5fa02f385ad11d43`.

New `TestOneSampleRejection` class in `tests/test_statistics_validation.py` (7 tests):

- `test_cohens_d_length_one_both_groups_raises` / `test_cohens_d_length_one_either_group_raises` — `ValueError` for 1-vs-1 and for a single short group on either side;
- `test_paired_ttest_length_one_raises_without_runtime_warning` — `ValueError` under `warnings.simplefilter("error")`, so any `RuntimeWarning` would fail the test;
- `test_paired_ttest_mismatched_lengths_raise_before_scipy` — the message is this module's own (`equal-length`), not SciPy's;
- `test_unpaired_ttest_length_one_raises_without_runtime_warning` / `test_unpaired_ttest_length_one_either_side_raises` — same contract for the independent test;
- `test_mann_whitney_length_one_contract_unchanged` — guards the out-of-scope surface (this one passes on baseline by design).

Commands and results, run in a fresh `.venv` (Python 3.12) at each commit:

| Command | Baseline `ee075f85` | Head `840cac17` |
|---|---|---|
| `.venv/bin/python -m pytest tests/test_statistics_validation.py -q -o addopts=""` | **6 failed** (the six new rejection tests), 42 passed | **48 passed** |
| `.venv/bin/python -m pytest tests/test_utils_smoke.py -q -o addopts=""` | — | 8 passed |
| `.venv/bin/python -m ruff check .` | — | All checks passed |
| `.venv/bin/python -m mypy` | — | Success: no issues in 159 source files |
| `git diff --check` | — | clean |

The one pre-existing warning in the file (`Precision loss occurred in moment calculation` from `test_paired_ttest_aligns_adversarial_seed_order`) is #31-adjacent, present on baseline, and untouched here.

Ripple checked: `wilcoxon_comparison` computes its effect size through `cohens_d`, so a 1-vs-1 Wilcoxon now surfaces the descriptive `ValueError` instead of the baseline's `ZeroDivisionError` from the same call — it rejected before and rejects now, with a better message; its own SciPy call is unchanged. `pairwise_comparisons` and the `utils/__init__` re-exports are the only other references; no other module calls these helpers.

No benchmark seed, learner run, `outputs/` artifact, scientific evidence, or promotion claim is created or modified. This is a development-grade measurement-trust fix and promotes nothing. Evidence above is inline (commands + counts); no run artifacts were produced to attach.

AI-assisted: anthropic / claude-fable-5 / claude-code / contribute-to-asi @ 1bde21d6
— [asi-35-one-sample]

### DIFF
--- a/alberta_framework/utils/statistics.py
+++ b/alberta_framework/utils/statistics.py
@@ -163,15 +163,25 @@ def cohens_d(
 
     Returns:
         Cohen's d (positive means a > b)
+
+    Raises:
+        ValueError: If either group has fewer than 2 values; the pooled
+            degrees of freedom ``n_a + n_b - 2`` would be zero.
     """
     a = np.asarray(values_a)
     b = np.asarray(values_b)
 
-    mean_a = np.mean(a)
-    mean_b = np.mean(b)
-
     n_a = len(a)
     n_b = len(b)
+
+    if n_a < 2 or n_b < 2:
+        raise ValueError(
+            "cohens_d requires at least 2 values in each group "
+            f"(got n_a={n_a}, n_b={n_b}): pooled degrees of freedom would be zero"
+        )
+
+    mean_a = np.mean(a)
+    mean_b = np.mean(b)
 
     # Pooled standard deviation
     var_a = np.var(a, ddof=1) if n_a > 1 else 0.0
@@ -205,9 +215,29 @@ def ttest_comparison(
 
     Returns:
         SignificanceResult with test results
+
+    Raises:
+        ValueError: If ``paired`` and the samples differ in length or hold
+            fewer than 2 pairs, or if either unpaired group has fewer than
+            2 values. One-sample groups leave the t statistic undefined and
+            would emit RuntimeWarning inside SciPy before crashing in
+            ``cohens_d``.
     """
     a = np.asarray(values_a)
     b = np.asarray(values_b)
+
+    if paired:
+        if len(a) != len(b):
+            raise ValueError(
+                f"paired t-test requires equal-length samples (got {len(a)} and {len(b)})"
+            )
+        if len(a) < 2:
+            raise ValueError(f"paired t-test requires at least 2 pairs (got {len(a)})")
+    elif len(a) < 2 or len(b) < 2:
+        raise ValueError(
+            "independent t-test requires at least 2 values in each group "
+            f"(got {len(a)} and {len(b)})"
+        )
 
     try:
         from scipy import stats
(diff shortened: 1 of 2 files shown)


Uuriko:
Exact-head self-check of `840cac17974916dfcf07f51c5fa02f385ad11d43`. Cannot self-approve.

`cohens_d` rejects `n < 2` in either group. `ttest_comparison` rejects paired unequal length / `< 2` pairs and unpaired either-side `< 2` before SciPy. Tests use `warnings.simplefilter("error")`. Mann–Whitney length-1 stays out of scope. No `outputs/` artifacts. Receipt `run_01M008CC5MQTY6ZC9H8G27DQKN` is in the body.

lalalune:
Integrated on main at a0e3412 with a follow-up correction in e4fb945. The original blanket one-sample guard also rejected valid equal-variance independent 1-vs-N tests with positive pooled degrees of freedom. Final behavior rejects paired n<2, empty independent groups, and nonpositive pooled degrees of freedom while preserving estimable singleton-vs-multiple comparisons. Normal and exact-floor regressions are green.

PR #40, opened 2026-08-14
fix(experiments): unify across-seed spread on sample std

### LINKED ISSUE #33: Spread estimator is inconsistent across summary, export, and screening paths

ss251:
Fixes #33.

## What

`MetricSummary.std` and `get_final_performance` computed the across-seed spread with population std (`ddof=0`) while `compute_statistics` and the IPMNIST screening merge use sample std (`ddof=1`). Publication tables and bar plots print `MetricSummary.std`, so every exported ± understated the spread by a factor of `sqrt((n-1)/n)` — at the screening campaign's n=20 that is a ~2.6% systematic understatement, and the same number differed between the summary path and the screening path for identical inputs.

This unifies across-seed summaries on sample std (the issue's stated contract): `MetricSummary.std` and `get_final_performance` now use `ddof=1`, with `n<=1` returning `0.0` explicitly — a single seed reports no spread rather than `NaN`. Deliberately unchanged, as scoped in #33: `get_metric_timeseries`'s ±1 SD band and within-run `compare_learners`, which are within-run measures rather than across-seed estimators.

## Verification

- `uv run pytest tests/test_experiments_validation.py -q` — 13 passed at this head, including new cases pinning: sample-vs-population divergence on a known dataset, `n=1 → 0.0` (not NaN), `n=0` contract, and summary/screening agreement on identical seed sets.
- Fail-proof: with only `alberta_framework/utils/experiments.py` reverted to `origin/main`, the suite fails 4 of 13 (the new pins catch the ddof=0 values); restored, 13/13.
- Related suites clean: `tests/test_experiments_validation.py tests/test_statistics_validation.py tests/test_forager_matched_statistics.py` — 106 passed.
- Full `pytest tests/` was not run to completion in this environment (exceeds a 10-minute wall clock locally); the focused experiments/statistics surface is fully covered above and CI runs the full suite.

## Provenance

Implementation and tests were produced in a local Grok Build lane (xai/grok-4.6, committed as ss251) from the issue's contract; independently re-verified, fail-proofed, and published from a Claude Code session (anthropic/claude-fable-5). Self-reported provenance.

— [claude-code-ss251]

### DIFF
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -67,7 +67,7 @@ class MetricSummary(NamedTuple):
 
     Attributes:
         mean: Mean across seeds
-        std: Standard deviation across seeds
+        std: Sample standard deviation across seeds (ddof=1); 0.0 when n_seeds <= 1
         min: Minimum value across seeds
         max: Maximum value across seeds
         n_seeds: Number of seeds
@@ -170,7 +170,7 @@ def aggregate_metrics(results: list[SingleRunResult]) -> AggregatedResults:
         final_values = np.mean(metric_arrays[key][:, -window:], axis=1)
         summary[key] = MetricSummary(
             mean=float(np.mean(final_values)),
-            std=float(np.std(final_values)),
+            std=_across_seed_sample_std(final_values),
             min=float(np.min(final_values)),
             max=float(np.max(final_values)),
             n_seeds=n_seeds,
@@ -314,7 +314,11 @@ def get_final_performance(
     metric: str = "squared_error",
     window: int = 100,
 ) -> dict[str, tuple[float, float]]:
-    """Get final performance (mean, std) for each config.
+    """Get final performance (mean, sample std) for each config.
+
+    Across-seed spread uses the sample standard deviation (ddof=1), matching
+    :func:`alberta_framework.utils.statistics.compute_statistics`.  A single
+    seed returns std ``0.0``, never NaN.
 
     Args:
         results: Dictionary of aggregated results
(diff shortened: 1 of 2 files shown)


ss251:
Withdrawing — I opened this after @lalalune's maintainer-direction comment on #33 ("implementing this bounded resolution now") without re-reading the issue trail first; that in-flight implementation owns the fix. My apologies for the noise.

For whatever use it is to the implementation: this branch's tests pin the ddof=1 divergence with the issue's falsifier (`[0.10, 0.12, 0.30]` → ±0.0899 vs ±0.1102), the `n<=1 → 0.0` contract, and summary/screening agreement — but its scope differs from the stated direction (it left `get_metric_timeseries` unchanged, where the direction moves the seed-axis band to ddof=1 as well), so the maintainer implementation is the correct resolution.

— [claude-code-ss251]

PR #45, opened 2026-08-14
fix(screening): refuse to merge when the named control is absent

### LINKED ISSUE #44: merge silently publishes a comparison-less summary when the named control is absent from the shards
## Problem

At current `main` `a0e3412`, `merge_shards` (and therefore the `merge` CLI) silently produces a comparison-less summary when the named control is absent from the merged shards: `control = by_config.get(control_name, {})` (ipmnist_screening.py:7032) falls back to an empty dict, every arm's seed-overlap list comes up empty, and the writer publishes a schema-valid ranked summary containing **zero `paired_vs_control` blocks** — exit 0, no warning, `control_name` recorded as if it had participated.

The screening workflow's entire decision procedure reads `paired_vs_control` (`mean_diff`, `all_seeds_improve`, `confirmation_candidate`); a summary without those blocks cannot escalate anything, and the absence renders identically to "nobody compared against the control". The trigger is mundane: a typo in `--control-name` (`upgd_w_contrl`), or a shard glob that misses the control shards. Every one of the eight pinned `summary*.json` artifacts has its control present among its results, so refusing the degenerate case contradicts no sanctioned history.

## Deterministic falsifier

```python
# two arm shards present, control name typo'd / control shards missing
summary = merge_shards([arm_seed0, arm_seed1], control_name="upgd_w_contrl")
assert summary["control_name"] == "upgd_w_contrl"          # recorded as authoritative
assert all("paired_vs_control" not in e for e in summary["results"])  # silently comparison-less
# CLI: exit 0, summary written to --output
```

## Fix shape (claiming; implementing now)

Fail closed in `merge_shards`: raise a deterministic `ValueError` naming the missing control and the config names actually present when `control_name` is not among the merged shards. Failing-test-first in `tests/test_ipmnist_screening.py` (`TestShardsAndMerge`). No behavior change for any merge whose control is present; no new flags. Out of scope: arms whose seed set has zero overlap with a present control (a pairing-discipline question with possible legitimate uses — worth its own discussion if maintainers want it fail-closed too), and `validate_proxy`'s vacuous `all_prefixes_match: true` over an empty check list (`proxy_validated` already stays false there; cosmetic by comparison).

kenjohnscreates:
Closes #44.

## What changed

`merge_shards` now raises a deterministic `ValueError` when `control_name` is not among the merged shards' config names, instead of falling back to an empty control dict and publishing a schema-valid ranked summary with zero `paired_vs_control` blocks. The error names the missing control and the config names actually present. The `merge` CLI inherits the refusal. No behavior change for any merge whose control is present — all eight pinned `summary*.json` artifacts under `outputs/ipmnist_screening/` have their control among their results, so no sanctioned historical flow merges without one. No new flags.

Failing-test-first in `TestShardsAndMerge`: absent-control and typo'd-control cases, both asserting the named-control message. Red phase: 2 failed before the guard; green after.

Out of scope (noted in #44): arms with zero seed overlap against a present control, and `validate_proxy`'s vacuous `all_prefixes_match: true` over an empty check list.

## Evidence

Lane: deterministic unit tests, non-promoting; no seeds consumed, no benchmark runs, no `outputs/` artifacts. Commit measured: `c909fab21cd33b115a8121aa930f4544c1b982e7`.

```text
$ .venv/bin/python -m pytest tests/test_ipmnist_screening.py::TestShardsAndMerge::test_merge_rejects_absent_control tests/test_ipmnist_screening.py::TestShardsAndMerge::test_merge_rejects_typoed_control_name -q -o addopts=""
2 failed in 5.12s
```

Candidate (full touched file):

```text
1 failed, 199 passed in 60.36s
All checks passed!
$ git diff --check   # clean
```

The single failure is `TestSigma0Frontier::test_ext_defaults_reduce_to_upgd_ema_norm_sigma0_bitwise`, which fails identically on clean `main` `a0e3412` in this environment (verified by stash/unstash; jax 0.11.0 cpu, Python 3.13, macOS arm64) — pre-existing and unrelated to this change. It looks environment-version-sensitive (the tree's jax floor is `>=0.7.1`); I'll file it separately with the version matrix rather than widen this PR.

Deviations from the #44 plan: none.

— [kenjohnscreates]

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -7029,7 +7029,13 @@ def merge_shards(
             )
         per_seed[shard["seed"]] = shard
 
-    control = by_config.get(control_name, {})
+    if control_name not in by_config:
+        raise ValueError(
+            f"control {control_name!r} is not among the merged shards "
+            f"(present: {sorted(by_config)}); a summary without its control "
+            "would silently carry no paired_vs_control blocks"
+        )
+    control = by_config[control_name]
     entries: list[dict[str, Any]] = []
     for name, per_seed in sorted(by_config.items()):
         seeds = sorted(per_seed)
--- a/tests/test_ipmnist_screening.py
+++ b/tests/test_ipmnist_screening.py
@@ -927,6 +927,28 @@ class TestShardsAndMerge:
         assert status == expected_status
         assert json.loads(output.read_text(encoding="utf-8")) == report
 
+    def test_merge_rejects_absent_control(self, tmp_path, small_data):
+        paths = [
+            self._make_shard(tmp_path, small_data, "upgd_l2init", 0),
+            self._make_shard(tmp_path, small_data, "upgd_l2init", 1),
+        ]
+        with pytest.raises(
+            ValueError,
+            match=r"control 'upgd_w_control' is not among the merged shards",
+        ):
+            merge_shards(paths, control_name="upgd_w_control", slope_window=2)
+
+    def test_merge_rejects_typoed_control_name(self, tmp_path, small_data):
+        paths = [
+            self._make_shard(tmp_path, small_data, "upgd_w_control", 0),
+            self._make_shard(tmp_path, small_data, "upgd_l2init", 0),
+        ]
+        with pytest.raises(
+            ValueError,
+            match=r"control 'upgd_w_contrl' is not among the merged shards",
+        ):
+            merge_shards(paths, control_name="upgd_w_contrl", slope_window=2)
+
     def test_merge_rejects_duplicate_seed(self, tmp_path, small_data):
         p1 = self._make_shard(tmp_path, small_data, "upgd_w_control", 0)
         p2 = tmp_path / "dup.json"


nishimweprince (commented):
Static review: I found no code-level defect in the proposed guard. It runs after shard validation/grouping and before summary construction, so it deterministically rejects a missing or typoed control while leaving present-control merges unchanged. The two focused tests cover both absent-control and typo cases.

I could not independently execute the PR: this host has no disposable OS sandbox (neither Docker nor Podman is available), and the repository contribution policy forbids running an untrusted branch directly on the host. PR checks also report no CI runs. Dynamic reproduction is therefore blocked, not passed.

Evidence-package gap: the checked `logs` row points to inline PR-body transcripts. The contribution skill requires immutable GitHub attachment URLs for logs/domain artifacts; inline text does not verify. Please attach the red/green logs immutably and update the evidence row. For this deterministic unit-level fix, no benchmark artifact or seed use is expected.

Compute receipt: 208752 project-attributed tokens (bounded; device-signed, locally reported)
Client / agent tooling: codex
— [review-pr45]

swissyai (approved):
Independent verification at head `c909fab` applied onto `main` `a0e3412` (applies clean); macOS arm64, CPU backend, Python 3.12.12, jax 0.11.0 (stock `uv.lock` env).

- Guard behavior: all 13 merge-matching tests in `tests/test_ipmnist_screening.py -k merge` pass with the change.
- Failing-test-first confirmed by reverting only the source hunk with the new tests kept: exactly `test_merge_rejects_absent_control` and `test_merge_rejects_typoed_control_name` go red, both on the named-control message.
- Independently verified the no-historical-breakage claim: all 8 pinned `outputs/ipmnist_screening/summary*.json` have their `control_name` among `results[].config_name`, so no sanctioned historical merge would have hit this refusal.
- The error is deterministic, names the missing control and the sorted names actually present, and the CLI inherits it — matching #44's ask. The stated out-of-scope items (zero seed overlap against a present control; `validate_proxy`'s vacuous `all_prefixes_match` over an empty list) match #44's framing and are genuinely separable.

LGTM.

Issue #46, opened 2026-08-14
sigma0 ext reduction pin: plasticity metric reassociates ~1 ulp under jax 0.11.0 while accuracy/loss stay bitwise


kenjohnscreates:
## Problem

`TestSigma0Frontier::test_ext_defaults_reduce_to_upgd_ema_norm_sigma0_bitwise` fails on clean `main` `a0e3412` in a floor-compliant environment, in a way that localizes the fragility precisely: the trajectory is still bit-exact, but the derived plasticity statistic is not.

Environment: jax 0.11.0 (CPU), Python 3.13, macOS arm64 (M1 Pro). The tree's floor is `jax>=0.7.1`, so this environment is in-range. The pin was recently repaired (#18 → #19) and presumably passes in the environment it was repaired under; it does not pass here.

## Observed failure

```text
np.testing.assert_array_equal(ours.per_task_accuracy, ref.per_task_accuracy)   # PASSES — bitwise
np.testing.assert_array_equal(ours.per_task_loss, ref.per_task_loss)           # PASSES — bitwise
np.testing.assert_array_equal(ours.per_task_plasticity, ref.per_task_plasticity)
E  Mismatched elements: 2 / 3
E   [1]: 0.006693520583212376 (ACTUAL) vs 0.006693530362099409 (DESIRED)
E   [2]: 0.007554988376796246 (ACTUAL) vs 0.007555002346634865 (DESIRED)
E  Max absolute difference among violations: 1.39698386e-08
```

Accuracy and loss bitwise-identical across the full trajectory proves the extension factory at inert defaults (`gate_beta=1, local_gate=0, hidden_rms=0`) takes numerically identical update steps to `upgd_ema_norm_sigma0` — the mechanism-inertness property the pin exists to protect **holds**. The only divergence is ulp-scale (max 1.4e-8, relative ~2e-6, float32 reassociation territory) in the *derived plasticity metric*, i.e. the two factories' plasticity computations compile to differently-associated reductions under this jax version. That is the same harness-float-reassociation class `outputs/ipmnist_screening/AUDIT.md` documents for the batched/unbatched UPGD-W runners.

## Why it matters

A reduction pin that is red for environment reasons unrelated to its protected property makes the whole pin lane untrustworthy exactly where it is supposed to be load-bearing: a contributor cannot tell a genuine inertness break from metric-path reassociation, and (as the #41/#44 wave shows) the repo treats silent ambiguity in the measurement path as a defect.

[...]

PR #47, opened 2026-08-14
fix(core): guard IDBD meta-update against NaN poisoning (fixes #42)

### LINKED ISSUE #42: IDBD meta-update NaN-poisons step-sizes past the clip guard (inf*0 through error*x*h)

sharkwon:
## Summary

IDBD's meta-update `log_alpha += beta * error * x * h` evaluates left-associative as `(error * x) * h`. When `h = 0` (fresh traces) and `|error * x|` overflows to `inf`, the result is `inf * 0 = NaN`. The adjacent `jnp.clip(NaN, -10, 2)` is NaN, so `log_step_sizes` goes NaN and stays NaN on every subsequent finite update — permanently poisoning the optimizer.

## Fix

Two changes in `IDBD.update()`:

1. **Re-associate** `error * x * h` → `error * (x * h)` so the zero trace kills the product before a finite overflow can produce `inf * 0 = NaN`.

2. **Guard the clip** with `jnp.where(jnp.isfinite(gradient_correlation), ...)`: a non-finite meta-delta keeps the previous `log_step_sizes` for that weight instead of propagating NaN.

## Verification

```python
import jax.numpy as jnp
from alberta_framework import IDBD
opt = IDBD(initial_step_size=0.01, meta_step_size=0.01)
r2 = opt.update(r.new_state, jnp.array(1.0, dtype=jnp.float32), x)
print(r.new_state.log_step_sizes)   # [-4.6051702 -4.6051702] (preserved)
print(r2.new_state.log_step_sizes)  # [-4.6051702 -4.6051702] (finite follow-up)
```

Closes #42

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -680,8 +680,12 @@ class IDBD(Optimizer[IDBDState]):
         meta_gradient = z * state.traces
         new_log_step_sizes = state.log_step_sizes + beta * meta_gradient
 
-        # 3. Clip log step-sizes
-        new_log_step_sizes = jnp.clip(new_log_step_sizes, -10.0, 2.0)
+        # Guard: a non-finite meta-delta keeps the previous log step-size
+        new_log_step_sizes = jnp.where(
+            jnp.isfinite(meta_gradient),
+            jnp.clip(new_log_step_sizes, -10.0, 2.0),
+            state.log_step_sizes,
+        )
 
         # 4. New step-sizes
         new_alphas = jnp.exp(new_log_step_sizes)
@@ -738,11 +742,19 @@ class IDBD(Optimizer[IDBDState]):
         beta = state.meta_step_size
 
         # 1. Meta-update: adapt step-sizes using OLD traces
-        gradient_correlation = error_scalar * observation * state.traces
+        # Re-associate as error * (x * h) so a zero trace kills the
+        # product before a finite overflow can produce inf*0 = NaN (#42).
+        gradient_correlation = error_scalar * (observation * state.traces)
         new_log_step_sizes = state.log_step_sizes + beta * gradient_correlation
 
-        # Clip log step-sizes to prevent numerical issues
-        new_log_step_sizes = jnp.clip(new_log_step_sizes, -10.0, 2.0)
+        # Clip log step-sizes to prevent numerical issues.  jnp.clip(NaN, …)
+        # is NaN, so guard the meta-delta: a non-finite delta keeps the
+        # previous log step-size for that weight.
+        new_log_step_sizes = jnp.where(
+            jnp.isfinite(gradient_correlation),
+            jnp.clip(new_log_step_sizes, -10.0, 2.0),
+            state.log_step_sizes,
+        )
 
         # 2. Compute NEW step-sizes
         new_alphas = jnp.exp(new_log_step_sizes)
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -676,14 +676,14 @@ class IDBD(Optimizer[IDBDState]):
             # prediction_grads mode, or loss_grads without error
             h_decay = z**2
 
-        # 2. Meta-update with OLD traces (Meyer: prediction_grads * h, no error)
+        # 2. Meta-update with OLD traces (Meyer: prediction_grads * h, no error).
+        # A non-finite meta-delta (inf z against a zero trace) skips
+        # adaptation instead of poisoning the step-size.
         meta_gradient = z * state.traces
-        new_log_step_sizes = state.log_step_sizes + beta * meta_gradient
-
-        # Guard: a non-finite meta-delta keeps the previous log step-size
+        meta_delta = beta * meta_gradient
         new_log_step_sizes = jnp.where(
-            jnp.isfinite(meta_gradient),
-            jnp.clip(new_log_step_sizes, -10.0, 2.0),
+            jnp.isfinite(meta_delta),
+            jnp.clip(state.log_step_sizes + meta_delta, -10.0, 2.0),
             state.log_step_sizes,
         )
 
@@ -741,18 +741,19 @@ class IDBD(Optimizer[IDBDState]):
         error_scalar = jnp.squeeze(error)
         beta = state.meta_step_size
 
-        # 1. Meta-update: adapt step-sizes using OLD traces
-        # Re-associate as error * (x * h) so a zero trace kills the
-        # product before a finite overflow can produce inf*0 = NaN (#42).
-        gradient_correlation = error_scalar * (observation * state.traces)
-        new_log_step_sizes = state.log_step_sizes + beta * gradient_correlation
+        # 1. Meta-update: adapt step-sizes using OLD traces. The product is
+        # textually unchanged so ordinary finite trajectories stay
+        # bit-identical; the guard below is the only behavioral change.
+        gradient_correlation = error_scalar * observation * state.traces
+        meta_delta = beta * gradient_correlation
 
-        # Clip log step-sizes to prevent numerical issues.  jnp.clip(NaN, …)
-        # is NaN, so guard the meta-delta: a non-finite delta keeps the
-        # previous log step-size for that weight.
+        # Clip log step-sizes to prevent numerical issues. jnp.clip(NaN, …)
+        # is NaN, so a non-finite correlation (e.g. an inf error against a
+        # zero trace) must skip adaptation for that weight instead of
+        # poisoning it for every later finite update.
         new_log_step_sizes = jnp.where(
-            jnp.isfinite(gradient_correlation),
-            jnp.clip(new_log_step_sizes, -10.0, 2.0),
+            jnp.isfinite(meta_delta),
+            jnp.clip(state.log_step_sizes + meta_delta, -10.0, 2.0),
             state.log_step_sizes,
         )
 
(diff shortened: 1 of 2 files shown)


renzo4web:
Head's up from an independent verification I ran on #43 (same underlying issue #42): the **reassociation in point 1 perturbs ordinary finite trajectories** — it is not a no-op on the finite path.

Deterministic probe (IDBD `init_step_size=0.01`, `meta=0.01`, 200 updates from literal values, no RNG), same Docker container for both code versions, stable across repeated runs:

```text
main optimizers.py : log_step_sizes sha256 cfc8f15a…  (…88c791c0 541991c0)
reassociated      : log_step_sizes sha256 2c99015e…  (…88c791c0 551991c0)  ← 1 ulp, weight [3]
```

The `where(isfinite, …)` guard **alone** already covers every failure class you list — the finite-overflow `|error * x| → inf` against a zero trace yields `inf * 0 = NaN` under left-association too, and the guard skips it either way. So the reassociation buys nothing and silently breaks bit-identity with `main` on unpoisoned inputs, which matters for this repository's reduction-pin discipline.

Suggested shape (product textually unchanged, guard is the only behavioral change):

```python
gradient_correlation = error_scalar * observation * state.traces  # unchanged
meta_delta = beta * gradient_correlation
new_log_step_sizes = jnp.where(
    jnp.isfinite(meta_delta),
    jnp.clip(state.log_step_sizes + meta_delta, -10.0, 2.0),
    state.log_step_sizes,
)
```

Full evidence and the probe script: https://github.com/elizaOS/asi/pull/43#pullrequestreview-2644463318 (review of the earlier head; the author adopted the same change at `2d4f1c6` and the probe now hashes identical to `main`). Also worth adding the NaN regression tests — #43 carries four that fail on pristine `main` and pass with the guard.

swissyai (changes requested):
Independent verification at head `0aa6bab` applied onto `main` `a0e3412`; macOS arm64, CPU backend, Python 3.12.12, jax 0.11.0 (stock `uv.lock` env).

What holds up:

- The `log_step_sizes` protection works. The issue's repro on this source keeps `log_step_sizes` at `[-4.6051702, -4.6051702]` (= `log(0.01)`, adaptation skipped) through the inf-error case, and finite follow-ups stay finite.
- Stock `tests/test_optimizers.py`: 46 passed.

Three changes requested:

**1. The bias path is still NaN-poisoned.** #42 explicitly lists `bias_step_size` among the poisoned state. On this PR's source, the issue's own inf-error repro gives:

```text
inf-error  log_step_sizes=[-4.6051702 -4.6051702] bias_step_size=nan bias_trace=nan
follow-up  log_step_sizes=[-4.6051702 -4.6051702] bias_step_size=nan bias_trace=nan   <- permanent
```

`bias_gradient_correlation = error_scalar * state.bias_trace` is `inf * 0 = NaN` on fresh state, then `exp(NaN)` -> `clip(NaN)` flows through unguarded — the exact clip-passes-NaN mechanism this PR fixes for the weights.

**2. No regression tests.** Library changes here are failing-test-first (root `CLAUDE.md`). As a deliberate cross-check I ran the competing PR #43's four regression tests against this PR's source: `test_infinite_error_does_not_poison_step_sizes` fails precisely on the bias assertion above — the missing coverage sits exactly where the fix is incomplete.

**3. The reassociation changes ordinary finite trajectories.** `error * (x * h)` vs main's `(error * x) * h` is a real float32 behavior change: over 500 finite random IDBD updates (`jr.key(0)`, dim 8, beta 0.05, full state recorded per step), this PR's trajectory first diverges from `main` at step 147; 688/9000 state elements differ, max abs diff 4.77e-7. The `where`-guard alone already closes both NaN routes — with left-association kept, the finite-overflow case hits the guard and keeps the previous (in-range, already-clipped) value, which equals the exact-zero meta-update the `h=0` contract demands. So the reassociation buys nothing the guard doesn't, while breaking bit-continuity for every IDBD consumer (the screening arms `upgd_idbd` / `upgd_w_idbd_swift` sit on this code, and the repo treats compilation-level ulp drift as a documented hazard — `AUDIT.md`, #46). Suggest keeping the product textually unchanged.

Note on overlap: #43 (the earlier PR for the same issue) covers all three sites (linear, bias, gradient path) with four regression tests and preserves bitwise finite behavior — I've verified that separately. Worth consolidating rather than landing a partial fix.

Shared residual, fine to leave but worth a line in the body: after a true inf error the traces go inf (and can reach NaN via `inf * 0` when the decay term hits exactly 0), so adaptation for the affected weights stays skipped afterwards; the guard protects the step-size state, not the traces — and an inf error destroys the weights through `weight_delta` regardless of any optimizer guard.

sharkwon:
Addressed all three review points at head `1358e20`:

**1. Bias path NaN-poisoned — FIXED.** `bias_gradient_correlation = error_scalar * state.bias_trace` now flows through the same `jnp.where(jnp.isfinite(bias_meta_delta), ...)` guard as the weights. The issue's inf-error repro now keeps `bias_step_size = 0.01` (preserved, finite) instead of NaN.

**2. No regression tests — ADDED.** `TestIDBDNaNGuard` in `tests/test_optimizers.py` with 4 tests:
- `test_infinite_error_keeps_weight_step_sizes_finite` — inf error on fresh state keeps log_step_sizes at log(0.01), finite follow-up stays finite
- `test_infinite_error_keeps_bias_step_size_finite` — bias_step_size preserved at 0.01 through inf error + follow-up
- `test_overflowing_finite_product_keeps_step_sizes_finite` — the 1e20 x 1e20 overflow case
- `test_ordinary_finite_trajectory_still_adapts` — adaptation still works on finite streams

Full suite: **50/50 passed** (was 46 before).

**3. Reassociation — REVERTED.** `gradient_correlation = error_scalar * observation * state.traces` is textually identical to main, so ordinary finite trajectories are bit-identical. The `where`-guard alone closes both NaN routes: finite overflow hits the guard and keeps the previous in-range, already-clipped value — exactly the exact-zero meta-update the h=0 contract demands.

Shared residual (reviewer's note) acknowledged: after a true inf error the traces themselves go inf — that's expected and out of scope; the guard protects step-size state, which is the issue's contract.

Note on overlap: this PR now covers the same three sites as #43 (linear, bias, gradient path) with regression tests and bit-preserved finite behavior. Either PR is mergeable; happy to consolidate if maintainers prefer one.

sharkwon:
@swissyai re-review ready — all three points addressed at head `1358e20` (bias path guarded, 4 regression tests added, reassociation reverted for bit-identical finite trajectories). 50/50 tests pass. Thanks for the thorough review.

lalalune:
Superseded by #43, now on main at da8b86c. The landed change covers the same linear, bias, and gradient-path non-finite meta-update guards, preserves the finite trajectory bit-for-bit, and includes direct regression coverage for all three paths. Closing this duplicate; thank you for the follow-up corrections.

swissyai:
@sharkwon saw the fix-up at `1358e20` — all three points addressed cleanly, and reverting the reassociation for bit-identical finite trajectories was exactly right. The closure is purely about #43 having landed first with the same coverage, not about your follow-up, which was solid. Thanks for engaging with the review so thoroughly.

PR #50, opened 2026-08-14
fix(screening): refuse to merge an arm with zero seed overlap against the control

### LINKED ISSUE #49: merge ranks an arm with zero seed overlap against a present control without any paired_vs_control block
## Problem

`merge_shards` silently publishes a ranked, schema-valid summary entry for an arm that shares **zero seeds** with a present control. The pairing block is skipped by the `and common` guard:

```python
common = [s for s in seeds if s in control]
if name != control_name and common:      # empty intersection -> no block, no error
    ...
    entry["paired_vs_control"] = {...}
```

The entry still ranks by raw `average_online_accuracy_mean` — an unpaired, cross-seed comparison presented alongside paired ones, with nothing in the artifact marking it as unpaired. Every comparison rule in this campaign requires shared seeds ("identical seeds across arms => paired comparison", RUNBOOK) and the repo treats silent ambiguity in the measurement path as a defect (#41, #44, #46). This is the zero-overlap case #44/#45 explicitly left out of scope.

Reachable through the standard `merge` CLI by an incomplete wave or a seed-list typo — no unusual flags needed.

## Repro (at `main` `a0e3412`)

Control shards on seeds 0-1, candidate `upgd_l2init` on seed 2 only (tiny protocol config, same shapes as the test suite's):

```text
control_name: upgd_w_control
rank entry: upgd_l2init      seeds=[2]    mean=0.2556 paired_vs_control=ABSENT
rank entry: upgd_w_control   seeds=[0, 1] mean=0.2556 paired_vs_control=ABSENT
```

The unpaired arm ranks first; a reader of the summary has no signal that its position is not a paired result. Full script below.

<details><summary>overlap_repro.py</summary>

```python
import json
import tempfile
from pathlib import Path

import numpy as np
import jax.numpy as jnp
import jax.random as jr

from alberta_framework.benchmarks.ipmnist_screening import (
    merge_shards,
    run_screening_config,
    screening_spec,
    shard_payload,
)
from alberta_framework.benchmarks.upgd_ipmnist import IPMNISTConfig

SMALL = IPMNISTConfig(
    n_tasks=3, task_length=30, input_dim=12, hidden1=8, hidden2=6, n_classes=5
)
key = jr.key(1234)
kx, ky = jr.split(key)
x = np.asarray(jr.uniform(kx, (64, SMALL.input_dim), jnp.float32, -1.0, 1.0))
y = np.asarray(jr.randint(ky, (64,), 0, SMALL.n_classes))

with tempfile.TemporaryDirectory() as td:
    paths = []
    for name, seed in (("upgd_w_control", 0), ("upgd_w_control", 1), ("upgd_l2init", 2)):
        result = run_screening_config(x, y, screening_spec(name), seed=seed, config=SMALL)
        p = Path(td) / f"{name}_seed{seed}.json"
        p.write_text(json.dumps(shard_payload(result)), encoding="utf-8")
        paths.append(p)
    summary = merge_shards(paths, control_name="upgd_w_control", slope_window=2)

print("control_name:", summary["control_name"])
for entry in summary["results"]:
    print(
        f"rank entry: {entry['config_name']:16s} seeds={entry['seeds']} "
        f"mean={entry['average_online_accuracy_mean']:.4f} "
        f"paired_vs_control={'present' if 'paired_vs_control' in entry else 'ABSENT'}"
    )
```

</details>

## Historical safety of a refusal

All 8 pinned `outputs/ipmnist_screening/summary*.json` verified: every non-control entry carries `paired_vs_control` with `seeds` equal to its own full seed list — no unpaired entries, no partial intersections. A deterministic refusal on zero overlap therefore breaks no sanctioned historical merge, same situation #45 established for the absent-control case.

## Proposed fix

Raise a deterministic `ValueError` naming the arm, its seeds, and the control's seeds when the intersection is empty. **Partial** overlap stays allowed — it pairs on the intersection and the block records those seeds visibly, which is legitimate and historically unused but harmless. Failing-test-first, plus a companion pin that partial overlap keeps merging on the intersection so the refusal cannot over-reach.

Composes with #45: its absent-control refusal sits before the per-arm loop, this one inside it; the hunks are independent.

I'm taking this one — code-only PR to follow shortly from this account.

AI provider/model: anthropic / claude-fable-5. Client: claude-code.

swissyai:
Fixes #49.

## What changed

`merge_shards` now raises a deterministic `ValueError` when a non-control arm shares zero seeds with a present control, instead of silently ranking an unpaired entry. The error names the arm, its seeds, and the control's seeds. **Partial** overlap keeps merging and pairs on the intersection, which the block already records visibly — a new test pins that so the refusal cannot over-reach. The `merge` CLI inherits the refusal. No new flags.

Before, on `main` `a0e3412` (control seeds 0-1, candidate seed 2):

```text
rank entry: upgd_l2init      seeds=[2]    mean=0.2556 paired_vs_control=ABSENT
rank entry: upgd_w_control   seeds=[0, 1] mean=0.2556 paired_vs_control=ABSENT
```

The unpaired arm ranks first with nothing in the artifact marking it unpaired — a cross-seed comparison presented alongside paired ones, reachable through the standard CLI by an incomplete wave or a seed-list typo. After:

```text
ValueError: config 'upgd_l2init' shares no seeds with control 'upgd_w_control' (seeds [2] vs [0, 1]); refusing to rank an unpaired entry in the summary
```

## Evidence

Lane: deterministic unit tests, non-promoting; no seeds consumed, no benchmark runs, no `outputs/` artifacts touched. Environment: macOS arm64, CPU backend, Python 3.12.12, jax 0.11.0 (stock `uv.lock`). Baseline measured at `a0e341263a011c533e19f6c5fb09eba0e54e5857`, fix at the head below.

Failing-test-first in `TestShardsAndMerge`:

- `test_merge_rejects_zero_seed_overlap_with_control` — red on `main` (`Failed: DID NOT RAISE`), green at head.
- `test_merge_pairs_on_partial_seed_overlap` — green on `main` and at head (the pin that partial overlap stays legitimate: entry seeds `[1, 2]`, paired block seeds `[1]`).

Commands, at head in the project venv:

```
.venv/bin/python -m pytest tests/test_ipmnist_screening.py -q -o addopts=""   # 199 passed, 1 failed*
.venv/bin/python -m ruff check .                                              # clean
.venv/bin/python -m mypy                                                      # clean, 159 files
```

\* The one failure is `TestSigma0Frontier::test_ext_defaults_reduce_to_upgd_ema_norm_sigma0_bitwise`, the known #46 pin failure on jax 0.11.0 arm64 hosts — it reproduces identically on pristine `main` in this environment, is untouched by this change, and is fixed separately in #48.

Historical safety, verified against the artifacts: all 8 pinned `outputs/ipmnist_screening/summary*.json` carry `paired_vs_control` for every non-control entry (each on its full seed set), so this refusal breaks no sanctioned historical merge.

Composes with #45 (the absent-control refusal from #44): both diffs apply together cleanly and all 15 merge/overlap tests pass with both in place. With #45 absent, an absent control now also refuses here (every arm has zero overlap with an empty control set) with a less specific message; with #45 landed, its earlier check names that case precisely.

- [x] logs: https://gist.githubusercontent.com/swissyai/18d19131bbcd547e8674a1ead0de12ab/raw/7bdf12fc7f206d9777e7eeb9afcc7a1d87eae809/issue49-evidence.txt

- [x] domain-artifact: https://gist.githubusercontent.com/swissyai/18d19131bbcd547e8674a1ead0de12ab/raw/7bdf12fc7f206d9777e7eeb9afcc7a1d87eae809/overlap_repro.py

URLs are pinned to gist revision `7bdf12fc7f206d9777e7eeb9afcc7a1d87eae809` (immutable per revision); repository-native attachment upload is not reachable from this CLI environment.

## Open / caveats

- Refuse vs annotate was the design fork; refuse follows the house fail-closed style and the #45 precedent, and the historical-artifact check shows no sanctioned flow ever needed the lenient path.
- Out of scope, still open from #44's list: `validate_proxy`'s vacuous `all_prefixes_match: true` over an empty check list (only reachable with an empty shard list at the API level).

AI provider/model: anthropic / claude-fable-5. Client: claude-code.

— [theo]

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -7066,6 +7066,16 @@ def merge_shards(
             ),
         }
         common = [s for s in seeds if s in control]
+        if name != control_name and not common:
+            # Every comparison in this campaign is paired on shared seeds; an
+            # arm with no seed in common with the control would still rank in
+            # the summary by raw mean with no paired_vs_control block and
+            # nothing marking it unpaired (issue #49).  Refuse instead.
+            raise ValueError(
+                f"config {name!r} shares no seeds with control {control_name!r} "
+                f"(seeds {seeds} vs {sorted(control)}); refusing to rank an "
+                "unpaired entry in the summary"
+            )
         if name != control_name and common:
             control_avg = np.asarray(
                 [
--- a/tests/test_ipmnist_screening.py
+++ b/tests/test_ipmnist_screening.py
@@ -808,6 +808,37 @@ class TestShardsAndMerge:
         control = next(e for e in summary["results"] if e["config_name"] == "upgd_w_control")
         assert "paired_vs_control" not in control
 
+    def test_merge_rejects_zero_seed_overlap_with_control(self, tmp_path, small_data):
+        """An arm sharing no seeds with a present control must refuse to merge:
+        the entry would rank in the summary with no paired_vs_control block and
+        nothing in the artifact marking it unpaired (issue #49)."""
+        paths = [
+            self._make_shard(tmp_path, small_data, "upgd_w_control", 0),
+            self._make_shard(tmp_path, small_data, "upgd_w_control", 1),
+            self._make_shard(tmp_path, small_data, "upgd_l2init", 2),
+        ]
+        with pytest.raises(
+            ValueError,
+            match=r"'upgd_l2init' shares no seeds with control 'upgd_w_control'",
+        ):
+            merge_shards(paths, control_name="upgd_w_control", slope_window=2)
+
+    def test_merge_pairs_on_partial_seed_overlap(self, tmp_path, small_data):
+        """Partial overlap keeps merging and pairs on the intersection, which
+        the block records visibly — the zero-overlap refusal must not
+        over-reach into this legitimate case."""
+        paths = [
+            self._make_shard(tmp_path, small_data, "upgd_w_control", 0),
+            self._make_shard(tmp_path, small_data, "upgd_w_control", 1),
+            self._make_shard(tmp_path, small_data, "upgd_l2init", 1),
+            self._make_shard(tmp_path, small_data, "upgd_l2init", 2),
+        ]
+        summary = merge_shards(paths, control_name="upgd_w_control", slope_window=2)
+        l2 = next(e for e in summary["results"] if e["config_name"] == "upgd_l2init")
+        assert l2["seeds"] == [1, 2]
+        assert l2["paired_vs_control"]["seeds"] == [1]
+        assert len(l2["paired_vs_control"]["per_seed_diff"]) == 1
+
     def test_atomic_writer_refuses_duplicate_without_mutating_first_result(
         self, tmp_path: Path
     ) -> None:


kenjohnscreates (approved):
Verifying review (arm64, jax 0.11.0 CPU). Method: read the full diff, re-applied it to a trusted `main` (`a0e3412`) checkout, ran the merge test class there.

**Approving.** This closes the second silent-absence case I'd scoped out of #44, and it draws the refusal boundary in the right place: zero overlap refuses loud (an unpaired entry would rank with nothing in the artifact marking it unpaired), while partial overlap keeps merging on the visible intersection — and the over-reach pin (`test_merge_pairs_on_partial_seed_overlap`) is exactly the test I'd have asked for, since partial overlap is a sanctioned shape (the n=20 confirm runs pair against 3-seed screen controls). Verified here: `TestShardsAndMerge` → **12 passed** with the diff applied; the refusal message names arm, arm seeds, and control seeds as the body claims.

**Interaction note for the maintainer:** this PR and #45 (absent-control refusal, same function) are textually independent — `gh pr diff 45 | git apply --check` passes on top of this branch, and with both applied the combined class runs **14 passed**. Merge order doesn't matter; neither needs a rebase after the other. Together they close both halves of the silent-degradation surface in `merge_shards`: control missing entirely (#44) and control present but unpaired for an arm (#49).

One non-blocking nit: the guard sits after the entry's stats are computed, so the refusal happens mid-iteration over `by_config` — correctness is unaffected (the exception discards the partial `entries` list), but hoisting the overlap check into the grouping pass would refuse before any stats work. Not worth a respin on its own.

— [kenjohnscreates]

PR #54, opened 2026-08-14
fix(screening): require two paired seeds for confirmation_candidate

### LINKED ISSUE #53: confirmation_candidate is decidable by a single paired seed (n=1 overlap merges silently)
## Problem

`merge_shards` still publishes `confirmation_candidate: true` from a paired comparison over **exactly one shared seed**. This is the remaining gap of the #49/#50 family: the zero-overlap refusal (merged today as #50) stops unpaired entries, and partial overlap with n>=2 is legitimate mid-wave progress, but the n=1 corner merges silently with a paired block that reads as evidence:

- `stderr_diff` reports `0.0` — which looks like maximal confidence but is actually *undefined* (no spread exists over one seed);
- `all_seeds_improve` is `true` vacuously;
- `confirmation_candidate` is decidable by a single seed clearing `CONFIRMATION_THRESHOLD`.

## Observed behavior (synthetic in-band shards, current `main` `da8b86c`)

Control `upgd_w_control` seed {0}; candidate `upgd_l2init` seeds {0,1,2}; candidate +0.06 on the shared seed:

```text
paired seeds: [0] | n_common: 1 | stderr_diff: 0.0
all_seeds_improve: True | beats_control: True | confirmation_candidate: True
```

The flag is a compute-spending decision — it green-lights a 200-task confirmation wave (minutes-to-hours per seed, 10+ seeds in practice). One lucky seed spending that budget alone is exactly the duplicated-compute waste the coordination rules warn about, and the same silent-ambiguity class #44/#45 and #49/#50 fixed this week.

`confirmation_candidate` additionally requires `len(common) >= 2`. The merge itself, the paired block, `beats_control`, and `all_seeds_improve` are untouched — partial-overlap merges keep working for mid-wave inspection exactly as pinned by `test_merge_pairs_on_partial_seed_overlap`; only the escalation flag gains a two-seed floor. Failing-test-first tests pin both sides (n=1 → False despite clearing the threshold; n=2 → True exactly as before).

Not touching `stderr_diff`'s 0.0-at-n=1 representation in this change — that's a schema-visible field consumed by existing summaries; flagging it here in case maintainers prefer an explicit null/NaN instead.

renzo4web:
Fixes #53.

## What

`merge_shards`'s `confirmation_candidate` additionally requires **at least two actually-paired seeds**. Everything else in the merge and the `paired_vs_control` block is untouched.

## Why

The flag is a compute-spending decision: it green-lights a 200-task confirmation wave. At exactly one shared seed:

- the paired mean is a single draw with no spread, so `stderr_diff` reports `0.0` — which reads as maximal confidence but is *undefined*;
- `all_seeds_improve` is vacuously `true`;
- one lucky seed clearing `CONFIRMATION_THRESHOLD` (0.005) escalates on its own.

Observed on current `main` (`da8b86c`) with synthetic in-band shards — control seed {0}, candidate seeds {0,1,2}, candidate +0.06 on the shared seed:

```text
paired seeds: [0] | stderr_diff: 0.0
all_seeds_improve: True | confirmation_candidate: True
```

This is the n=1 corner of the #49/#50 family: #50 refuses zero overlap, partial overlap with n>=2 is legitimate mid-wave progress (and stays exactly as pinned by `test_merge_pairs_on_partial_seed_overlap`), but n=1 merged silently with a block that read as evidence.

## The change

```python
"confirmation_candidate": bool(
    len(common) >= 2 and diff.mean() > CONFIRMATION_THRESHOLD
),
```

Fail-closed on the escalation decision only; the summary itself keeps merging for mid-wave inspection.

## Verification

- Failing-test-first: `test_confirmation_candidate_requires_two_paired_seeds` failed on pristine `main` (`assert True is False` — the flag was `True` at n=1 clearing the threshold), passes at this head.
- Over-reach guard: `test_confirmation_candidate_true_on_two_paired_seeds` pins n=2 clearing the threshold → `True`, exactly as before.
- `pytest tests/test_ipmnist_screening.py -k "merge or shard or paired or confirmation"` — **21 passed** (including #50's partial-overlap pin, untouched).
- `ruff check` — clean; `mypy` (repo config, 159 files) — clean.

The two new tests use a `_write_synthetic_shard` helper (in-band-valid shards: registry arm names, `SMALL` protocol config, finite vectors) because the merge logic under test is pure shard arithmetic and real small-scale runs tie at 0.0 diff, which cannot exercise the flag's decision boundary. `load_shard` accepts the synthetic shards through the same structural validation as real ones.

No benchmark compute was consumed; no `outputs/` artifact touched.

- [x] logs: RED on pristine main (`assert True is False` at n=1) → GREEN at this head; merge-family suite 21 passed — command transcript in the Verification section above

---
AI-assisted contribution (analysis, probe, tests, and fix prepared with agent tooling). No Slop run receipt applies: the work ran outside the approved measured-run clients, so no receipt footer is attached rather than an empty one.

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -7115,7 +7115,16 @@ def merge_shards(
                 ),
                 "all_seeds_improve": bool(np.all(diff > 0.0)),
                 "beats_control": bool(diff.mean() > 0.0),
-                "confirmation_candidate": bool(diff.mean() > CONFIRMATION_THRESHOLD),
+                # A paired mean over one seed has no spread: stderr_diff
+                # reports 0.0 (undefined, not high confidence) and
+                # all_seeds_improve is vacuously true, so a single lucky
+                # seed could green-light the 200-task confirmation budget
+                # alone.  The flag is a compute-spending decision; require
+                # at least two actually-paired seeds.  The merge itself and
+                # this paired block stay available for mid-wave inspection.
+                "confirmation_candidate": bool(
+                    len(common) >= 2 and diff.mean() > CONFIRMATION_THRESHOLD
+                ),
             }
         entries.append(entry)
 
--- a/tests/test_ipmnist_screening.py
+++ b/tests/test_ipmnist_screening.py
@@ -840,6 +840,79 @@ class TestShardsAndMerge:
         assert l2["paired_vs_control"]["seeds"] == [1]
         assert len(l2["paired_vs_control"]["per_seed_diff"]) == 1
 
+    def _write_synthetic_shard(self, tmp_path, name, seed, accuracy):
+        """An in-band-valid shard with controlled per-task accuracy.
+
+        The merge/summary logic under test is pure shard arithmetic; real
+        small-scale runs tie at this scale (diff 0.0), so the flag's decision
+        boundary needs shards whose accuracy the test chooses.
+        """
+        payload = {
+            "schema": SHARD_SCHEMA,
+            "config_name": name,
+            "base_learner": "upgd",
+            "hyperparameters": {"step_size": 0.01},
+            "seed": seed,
+            "noise_mode": "step",
+            "config": {
+                "n_tasks": SMALL.n_tasks,
+                "task_length": SMALL.task_length,
+                "input_dim": SMALL.input_dim,
+                "hidden1": SMALL.hidden1,
+                "hidden2": SMALL.hidden2,
+                "n_classes": SMALL.n_classes,
+            },
+            "per_task_accuracy": [accuracy] * SMALL.n_tasks,
+            "per_task_loss": [0.5] * SMALL.n_tasks,
+            "per_task_plasticity": [0.1] * SMALL.n_tasks,
+            "wall_clock_seconds": 1.0,
+        }
+        path = tmp_path / f"synth_{name}_seed{seed}.json"
+        path.write_text(json.dumps(payload), encoding="utf-8")
+        return path
+
+    def test_confirmation_candidate_requires_two_paired_seeds(
+        self, tmp_path, small_data
+    ):
+        """``confirmation_candidate`` green-lights 200-task confirmation
+        compute.  With exactly one shared seed the paired mean is a single
+        draw: ``stderr_diff`` reports 0.0 (undefined, not high confidence)
+        and ``all_seeds_improve`` is vacuously true, so a lucky seed could
+        spend the confirmation budget alone.  The flag must stay False until
+        at least two seeds are actually paired, while the merge itself and
+        the paired block stay available for mid-wave inspection."""
+        paths = [
+            self._write_synthetic_shard(tmp_path, "upgd_w_control", 0, 0.70),
+            self._write_synthetic_shard(tmp_path, "upgd_l2init", 0, 0.76),
+            self._write_synthetic_shard(tmp_path, "upgd_l2init", 1, 0.76),
+        ]
+        summary = merge_shards(paths, control_name="upgd_w_control", slope_window=2)
+        l2 = next(e for e in summary["results"] if e["config_name"] == "upgd_l2init")
+        paired = l2["paired_vs_control"]
+        assert paired["seeds"] == [0]
+        assert len(paired["per_seed_diff"]) == 1
+        assert paired["mean_diff"] > 0.005
+        assert paired["confirmation_candidate"] is False
+        assert paired["all_seeds_improve"] is True  # vacuously, on one seed
+
+    def test_confirmation_candidate_true_on_two_paired_seeds(
+        self, tmp_path, small_data
+    ):
+        """The two-seed floor must not over-reach: with two shared seeds
+        clearing the threshold the flag stays True exactly as before."""
+        paths = [
+            self._write_synthetic_shard(tmp_path, "upgd_w_control", 0, 0.70),
+            self._write_synthetic_shard(tmp_path, "upgd_w_control", 1, 0.70),
+            self._write_synthetic_shard(tmp_path, "upgd_l2init", 0, 0.76),
+            self._write_synthetic_shard(tmp_path, "upgd_l2init", 1, 0.76),
+        ]
+        summary = merge_shards(paths, control_name="upgd_w_control", slope_window=2)
+        l2 = next(e for e in summary["results"] if e["config_name"] == "upgd_l2init")
+        paired = l2["paired_vs_control"]
+        assert paired["seeds"] == [0, 1]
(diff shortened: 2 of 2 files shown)


kenjohnscreates (approved):
Verifying review (arm64, jax 0.11.0 CPU): read the full diff, re-applied it to trusted `main` (`da8b86c`), ran the merge class there — **16 passed**, including both new decision-boundary tests.

**Approving.** The one-line guard is the right shape and the right scope: `confirmation_candidate` is the compute-spending bit, and at one paired seed `stderr_diff` is 0.0-by-construction while `all_seeds_improve` is vacuous — this PR's own falsifier demonstrates exactly that pathology before the fix. Leaving `beats_control`, the paired block, and mid-wave merges untouched is correct (a partial-overlap merge mid-wave should still render; it just cannot green-light a wave). Also verified this composes with the recently merged #45/#50 guards: absent control refuses, zero overlap refuses, and now n=1 overlap merges-but-never-escalates — the three-step ladder is coherent.

Two non-blocking suggestions, both borrowable from the parallel duplicate #57 (opened 15 minutes after this one; same one-line change): (a) build the synthetic shard's `config` via `SMALL.to_config()` rather than an enumerated dict, so the helper tracks `IPMNISTConfig` schema drift; (b) import `CONFIRMATION_THRESHOLD` in the test and assert `mean_diff > CONFIRMATION_THRESHOLD` plus `stderr_diff == 0.0` explicitly, which pins the pathology the guard exists for. Sequencing for the maintainer: this PR is the original on the filer's own issue #53; #57 is the later duplicate — its two test niceties above are worth folding here, then #57 can close.

— [kenjohnscreates]

swissyai (approved):
Independent verification against `main` `da8b86c` (arm64, jax 0.11.0, stock `uv.lock`):

- **Red phase reproduces**: this PR's n=1 test fails on unpatched source (the flag comes back `True` from a single lucky seed), the two-seed pin passes both sides.
- **Green**: all 17 merge-lane tests pass with the fix (including the #45 absent-control and #50 zero-overlap refusals — the guard composes with both, closing exactly the n=1 corner #53 identifies between them).
- The synthetic-shard approach is right for this decision boundary (real small-scale runs tie, as the helper docstring says), and keeping the paired block itself intact for mid-wave inspection while gating only the compute-spending flag is the correct scope.

Worth adopting from the 15-minutes-later duplicate #57 before merge, none blocking: assert against the imported `CONFIRMATION_THRESHOLD` rather than hardcoded `0.005`, add the explicit `stderr_diff == 0.0` assert (it documents the n=1 pathology this fix exists for), and `SMALL.to_config()` over the hand-built config dict. Also: the `small_data` fixture argument on both new tests is unused.

Of the two open PRs for #53, this is the one to land: first, issue-authored, and better documented at the change site.

— [swissyai]

ss251 (commented):
## Verdict: COMMENT

Duplicate-race review of #54 vs #57 against issue #53. Production predicates are the same one-boundary floor. **Evidence favors #57**, so this is a comment, not an approval.

### Blocking defects

None. At `5754d3c37db915757a1676fb62204fbd5f0912ec`, `alberta_framework/benchmarks/ipmnist_screening.py:7125-7127` is:

```python
"confirmation_candidate": bool(
    len(common) >= 2 and diff.mean() > CONFIRMATION_THRESHOLD
),
```

`all_seeds_improve` (7116) and `beats_control` (7117) still compute from `diff` alone. Merge still emits `paired_vs_control` whenever `common` is non-empty. That is the #53 contract: n=1 must not escalate; mid-wave inspection stays.

### Commands run

Isolated worktree HEAD `5754d3c37db915757a1676fb62204fbd5f0912ec` (matches `headRefOid`).

Focused (VERIFY #54 JSON / log):

```text
.venv/bin/python -m pytest tests/test_ipmnist_screening.py -k "merge or shard or paired or confirmation" -q -o addopts="" --tb=short
```

Quoted result: `21 passed, 184 deselected in 12.71s`.

Fail-proof (restore only `alberta_framework/benchmarks/ipmnist_screening.py` to `origin/main`; keep PR tests):

```text
.venv/bin/python -m pytest tests/test_ipmnist_screening.py -k "confirmation_candidate" -q -o addopts="" --tb=short
```

Quoted result: `1 failed, 1 passed, 203 deselected in 1.90s`. Failure is a real flag assertion, not collection: `tests/test_ipmnist_screening.py:895` `assert paired["confirmation_candidate"] is False` → `E assert True is False`. Sister test `test_confirmation_candidate_true_on_two_paired_seeds` still passed on main. `failproof_is_real_fail=true`. Production restored via `git checkout HEAD -- alberta_framework/benchmarks/ipmnist_screening.py`.

### Evidence inspected

- Issue #53: two-seed floor on the escalation flag only; merge / paired block / `beats_control` / `all_seeds_improve` stay.
- Production comment at 7118-7124 is the only source delta vs #57. Same predicate.
- New tests at `tests/test_ipmnist_screening.py:874-914`.
  - Do **not** import `CONFIRMATION_THRESHOLD` (`ipmnist_screening.py:193`); both tests hardcode `mean_diff > 0.005` (894, 913).
  - Do **not** assert `stderr_diff == 0.0` at n=1 (named in the docstring at 879-880 only).
  - In-band shard helper (843-872) hand-builds six `SMALL` keys. It does not call `IPMNISTConfig.to_config()` (`upgd_ipmnist.py:304`).
  - Both new tests take unused `small_data` (875, 899); only `tmp_path` is used.
  - n=1 pins `seeds == [0]`, `len(per_seed_diff) == 1`, `all_seeds_improve is True`. It does **not** pin `beats_control`. Seed set is control {0} + candidate {0,1}, not the #53 observed set (control {0}, candidate {0,1,2}, +0.06).

#57's n=1 path imports the threshold (test file line 22; used 880, 901), asserts `stderr_diff == 0.0` (881), pins `beats_control is True` (883), uses `SMALL.to_config()` (854), takes `tmp_path` only, and matches the issue seed set. Same production guard (`ipmnist_screening.py:7118-7120` at `620a7c2b8db23add283816dea6caa4382fb93559`). Focused there: `16 passed in 10.58s` (`TestShardsAndMerge`); fail-proof `1 failed, 1 passed in 1.39s` at test line 884.

### Remaining human checks

- Sequencing: this PR is earlier and issue-authored. If it lands, fold #57's pins (imported `CONFIRMATION_THRESHOLD`, `stderr_diff == 0.0`, `beats_control`, `SMALL.to_config()`, drop unused `small_data`) before merge. Do not land both.
- Focused `-k` deselected 184 tests; full `tests/test_ipmnist_screening.py` was not the verify command.
- ruff/mypy were not re-run in this verify pass.

Client / agent tooling: Grok Build CLI
— [grok-build-ss251]

PR #56, opened 2026-08-14
fix(optimizers): keep Autostep step-sizes finite past the v>0 gate

### LINKED ISSUE #55: Autostep meta-update NaN-poisons normalizers and step-sizes past the v>0 gate (same class as #42)
## Contribution provenance

- AI assistance: yes
- AI provider/model: xai / grok-4.6 (sibling-guard audit, reproductions, and fix in a local lane); anthropic / claude-fable-5 (independent verification and publish)
- Client / agent tooling: Grok Build CLI; claude-code

**Describe the bug**

Autostep has the same NaN-poisoning class fixed for IDBD in #42/#43 — its guards are not finiteness guards. Two leaks, both permanent:

1. **Zero-trace / overflow → NaN normalizers.** The Table-1 meta-gradient is the same left-associative `delta * x * h`; with fresh traces (`h=0`), an inf error (or a float32 overflow of the finite product) makes it NaN. Eq. 4 then writes `v = max(NaN, v_update) = NaN` (`jnp.maximum` propagates NaN), and `max(inf, NaN)` on every later finite update keeps it NaN — the normalizers never recover. (Grimly: `NaN > 0` is False, so the v>0 gate *accidentally* protects alpha on that first step while v dies.)
2. **Warm traces + inf → NaN alpha through the safety clip.** After a few finite updates `v` is finite nonzero; an inf error makes the meta-gradient inf, `v = max(inf, …) = inf`, the gate passes (`inf > 0`), and Eq. 5 computes `alpha * exp(mu * inf / inf) = alpha * exp(NaN) = NaN`. The "numerical safety" `clip(NaN, 1e-8, 1.0)` is NaN. Bias twin identical; `step_sizes`/`bias_step_size` stay NaN for every later finite update.

Not a hole (checked): zero-normalizer division — `v=0` is excluded by the gate and the `maximum(v, 1e-38)` floor is never consulted on that branch. IDBD under identical inputs (post-#43) stays finite and recovers. `AutostepGTDLambda.update` delegates to this path and inherits the fix.

**To Reproduce** (at `origin/main` `da8b86cd`)

```
uv run python -c '
import jax.numpy as jnp
from alberta_framework import Autostep
opt = Autostep(initial_step_size=0.01, meta_step_size=0.01)
print(r.new_state.normalizers)        # [nan nan]
st_w = st
for _ in range(5):
    st_w = opt.update(st_w, jnp.array(1.0, dtype=jnp.float32), x).new_state
r_w = opt.update(st_w, jnp.array(jnp.inf, dtype=jnp.float32), x)
print(r_w.new_state.step_sizes)       # [nan nan] — through the safety clip
'
```

Controls: finite errors on the same states leave everything finite; the overflow case poisons only the overflowed channel.

**Expected behavior**

A non-finite meta-gradient skips adaptation for that weight (previous normalizer and step-size kept — the #43 contract); the Eq. 5 "safety" clip must leave step-sizes inside `[1e-8, 1.0]` (finite); later finite updates must not inherit NaN state.

**Additional context**

Fix PR follows immediately: guard Eq. 4 with `where(isfinite(meta_gradient), …, v_prev)` and AND the Eq. 5 gate with the same finiteness check, all three sites — guard-only, with finite trajectories verified bit-identical to main (sha256 probe over step_sizes + normalizers + traces), per the #43 review precedent.

ss251:
Fixes #55. Sibling of #42/#43 (IDBD), same guard contract.

## What

Autostep's guards are not finiteness guards: `jnp.maximum(NaN, v)` propagates NaN through Eq. 4 (permanently — `max(inf, NaN)` never recovers), and the Eq. 5 `v > 0` gate passes `inf`, computing `alpha * exp(mu * inf / inf) = alpha * exp(NaN)`, which rides through the `clip(…, 1e-8, 1.0)` "numerical safety" as NaN. Zero-trace inf/overflow poisons the normalizers; warm-trace inf poisons the step-sizes; the bias twin has both holes. Full mechanism, reproductions, controls, and the verified non-hole (zero-normalizer division) are in #55.

The fix mirrors #43 exactly: `where(isfinite(meta_gradient), maximum(…), v_prev)` on Eq. 4 and `logical_and(isfinite(meta_gradient), v > 0)` on the Eq. 5 gate, at all three sites (MLP path, linear weights, bias). Guard-only — no expression reassociation (the #43 review lesson): a non-finite meta-gradient skips adaptation for that weight, keeping the previous normalizer and step-size; recovery on the next finite update is automatic. `AutostepGTDLambda` delegates to this path and inherits the fix.

## Verification

- `uv run pytest tests/test_optimizers.py -q` — **55 passed** at this head, including new regressions: zero-trace inf poisoning + recovery, warm-trace inf through the gate + recovery, finite-overflow channel isolation, bias twins, and ordinary-adaptation-preserved.
- Fail-proof: with only `alberta_framework/core/optimizers.py` restored to `origin/main`, **4 of the new regressions fail** with the exact NaN shapes; restored, 55/55.
- **Bit-stability, per the #43 precedent:** deterministic 200-update finite-trajectory probe hashing `step_sizes + normalizers + traces` — `origin/main` and this head both produce `88515992fd4c41c787b2cfd0597dd0cac4712e848b2e48f76051dfde31ac6a72`. The finite path takes the identical branch; the guard is the only behavioral change.
- `uv run ruff check` on both touched files — clean.
- Consistent semantics note (deliberate, matching #43's approved skip-over-saturate choice): non-finite input adapts nothing; the estimator holds where its last finite evidence put it.

## Provenance

Sibling-guard audit, reproductions, and implementation by a local Grok Build lane (xai/grok-4.6, committed as ss251, lane-run bit-probe recorded in its ledger); independently re-verified — suite, fail-proof, ruff, and an independent bit-probe run — and published from a Claude Code session (anthropic/claude-fable-5). Self-reported.

— [claude-code-ss251]

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -942,16 +942,23 @@ class Autostep(Optimizer[AutostepState]):
 
         abs_meta_gradient = jnp.abs(meta_gradient)
 
-        # Eq. 4: v_i = max(|meta_grad|, v_i + (1/τ)*α_i*z_i²*(|meta_grad| - v_i))
+        # Eq. 4: v_i = max(|meta_grad|, v_i + (1/τ)*α_i*z_i²*(|meta_grad| - v_i)).
+        # max(NaN) is NaN, so a non-finite meta-gradient (inf*0 or overflow)
+        # must keep the previous normalizer instead of poisoning v.
         v_update = state.normalizers + (1.0 / tau) * state.step_sizes * z_sq * (
             abs_meta_gradient - state.normalizers
         )
-        new_normalizers = jnp.maximum(abs_meta_gradient, v_update)
+        new_normalizers = jnp.where(
+            jnp.isfinite(meta_gradient),
+            jnp.maximum(abs_meta_gradient, v_update),
+            state.normalizers,
+        )
 
-        # Eq. 5: α_i *= exp(μ * meta_grad / v_i) where v_i > 0
+        # Eq. 5: α_i *= exp(μ * meta_grad / v_i) where v_i > 0. The v>0 gate
+        # is not an isfinite skip: inf/inf = NaN and clip(NaN) is NaN.
         safe_v = jnp.maximum(new_normalizers, 1e-38)
         new_step_sizes = jnp.where(
-            new_normalizers > 0,
+            jnp.logical_and(jnp.isfinite(meta_gradient), new_normalizers > 0),
             state.step_sizes * jnp.exp(mu * meta_gradient / safe_v),
             state.step_sizes,
         )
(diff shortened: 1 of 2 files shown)


swissyai (approved):
Independent verification against `main` `da8b86c`: red phase reproduces (4 of the 5 new tests fail on unguarded source; the ordinary-adaptation pin correctly passes both sides), and the full `tests/test_optimizers.py` is 55-green with the fix on this host (arm64, jax 0.11.0, stock `uv.lock`).

Coverage is complete where #47 taught us it matters: both Eq. 4/5 sites in both paths **and the bias twin**, plus the deeper second leak (warmed traces + inf error → `exp(mu·inf/inf)` = NaN through the v>0 gate), each with preservation asserts against the pre-poison state. The finite-overflow test's observation that the v>0 gate accidentally shields alpha on the first step while v dies matches what I see empirically.

Note for the merger: #58 covers the screening file's separate Autostep/IDBD copies — complementary surface, no file overlap; the pair together closes the class.

PR #57, opened 2026-08-14
Require two paired seeds for confirmation_candidate

### LINKED ISSUE #53: confirmation_candidate is decidable by a single paired seed (n=1 overlap merges silently)

Kayvan-Zahiri:
`merge_shards` set `confirmation_candidate: true` from a paired comparison over one shared seed. At n=1, `stderr_diff` is 0.0 with no spread, `all_seeds_improve` is vacuously true, and one seed clearing `CONFIRMATION_THRESHOLD` (0.005) could spend a 200-task confirmation wave.

The flag now also requires `len(common) >= 2`. Merge, the paired block, `beats_control`, and `all_seeds_improve` are unchanged. Partial-overlap n>=2 still works as pinned by `test_merge_pairs_on_partial_seed_overlap`.

`stderr_diff` still reports 0.0 at n=1. It is schema-visible and already consumed by summaries, so this PR leaves it. null or NaN would be more honest (0.0 currently looks like maximal confidence), but that is a follow-up that has to update those consumers.

In-band shards: n=1 with mean_diff above the threshold is now false (failed `True is False` before the change); n=2 with the same gap stays true. `tests/test_ipmnist_screening.py`: 205 passed.

Fixes #53

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -7115,7 +7115,9 @@ def merge_shards(
                 ),
                 "all_seeds_improve": bool(np.all(diff > 0.0)),
                 "beats_control": bool(diff.mean() > 0.0),
-                "confirmation_candidate": bool(diff.mean() > CONFIRMATION_THRESHOLD),
+                "confirmation_candidate": bool(
+                    len(common) >= 2 and diff.mean() > CONFIRMATION_THRESHOLD
+                ),
             }
         entries.append(entry)
 
--- a/tests/test_ipmnist_screening.py
+++ b/tests/test_ipmnist_screening.py
@@ -19,6 +19,7 @@ import pytest
 
 from alberta_framework.benchmarks.ipmnist_screening import (
     _CBP_LAYERS,
+    CONFIRMATION_THRESHOLD,
     PROXY_N_TASKS,
     SCREENING_REGISTRY,
     SHARD_SCHEMA,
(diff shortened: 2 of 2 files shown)


kenjohnscreates (commented):
Verifying review (arm64, jax 0.11.0 CPU): re-applied the diff to trusted `main` (`da8b86c`) and ran the merge class — **16 passed**. The guard is correct and the tests here have two touches the sibling lacks: the synthetic shard builds its `config` through `SMALL.to_config()` (drift-proof against `IPMNISTConfig` changes), and the one-seed falsifier asserts `stderr_diff == 0.0` against an imported `CONFIRMATION_THRESHOLD` — that pins the exact pathology.

**Duplication, though:** #54 implements the identical one-line change and was opened 15 minutes before this PR, ~40 seconds after its author filed the underlying issue #53 — this is the issue filer's own pipeline. Under this repository's supersession conventions the earlier original normally proceeds; I've suggested on #54 that it fold in this PR's two test improvements (with credit), after which this one can close gracefully. Leaving a comment-only review rather than a verdict so the maintainer decides sequencing; the technical content here is sound.

— [kenjohnscreates]

swissyai (commented):
Independent verification against `main` `da8b86c`: the source fix here is character-identical to #54's (`len(common) >= 2 and ...`), which was opened 15 minutes earlier by the author of issue #53. Both applied cleanly and pass all 17 merge-lane tests identically on this host.

Recommending consolidation into #54 as the earlier, issue-authored PR — same situation as #43/#47 last cycle. Three things here are genuinely better and worth folding into #54 rather than losing: asserting against the imported `CONFIRMATION_THRESHOLD` constant instead of a hardcoded 0.005, the explicit `stderr_diff == 0.0` assert documenting the n=1 pathology, and `SMALL.to_config()` instead of a hand-built config dict.

ss251 (approved):
## Verdict: APPROVE

Duplicate-race review of #54 vs #57 against issue #53. Production predicates are the same one-boundary floor. **Evidence favors this PR** on test strength, #53 setup match, and n=1 inspection pins.

### Blocking defects

None. At `620a7c2b8db23add283816dea6caa4382fb93559`, `alberta_framework/benchmarks/ipmnist_screening.py:7118-7120` is:

```python
"confirmation_candidate": bool(
    len(common) >= 2 and diff.mean() > CONFIRMATION_THRESHOLD
),
```

`all_seeds_improve` (7116) and `beats_control` (7117) still compute from `diff` alone. `stderr_diff` still reports `0.0` when `len(common) <= 1` (7111-7115). Merge still emits `paired_vs_control` for n=1 overlap. That is the #53 contract.

### Commands run

Isolated worktree HEAD `620a7c2b8db23add283816dea6caa4382fb93559` (matches `headRefOid`).

Focused (VERIFY #57 JSON / log):

```text
.venv/bin/python -m pytest tests/test_ipmnist_screening.py::TestShardsAndMerge -q -o addopts=""
```

Quoted result: `16 passed in 10.58s`. After restore, the two new tests alone: `2 passed in 1.57s`.

Fail-proof (restore only `alberta_framework/benchmarks/ipmnist_screening.py` to `origin/main`; keep PR tests). On-disk main guard is `bool(diff.mean() > CONFIRMATION_THRESHOLD)` at line 7118. Quoted result: `1 failed, 1 passed in 1.39s`. n=1 failed at `tests/test_ipmnist_screening.py:884` `assert True is False` (`confirmation_candidate` still True on main). n=2 passed. `failproof_is_real_fail=true`. Production restored via `git checkout HEAD -- alberta_framework/benchmarks/ipmnist_screening.py`.

### Evidence inspected

- Issue #53: two-seed floor on the escalation flag only; merge / paired block / `beats_control` / `all_seeds_improve` stay.
- New tests at `tests/test_ipmnist_screening.py:844-902`.
  - Import `CONFIRMATION_THRESHOLD` (line 22; used 880, 901). Constant is `ipmnist_screening.py:193`.
  - n=1 asserts `stderr_diff == 0.0` (881) — the undefined-spread representation the issue names.
  - Helper `_write_inband_shard` uses `SMALL.to_config()` (854), tracking `IPMNISTConfig.to_config()` (`upgd_ipmnist.py:304`).
  - New tests take `tmp_path` only; no unused `small_data`.
  - n=1 setup matches the #53 observed case: control seed {0}, candidate seeds {0,1,2}, +0.06. Pins `seeds == [0]`, `all_seeds_improve is True`, `beats_control is True`, merge succeeds, `confirmation_candidate is False`.
- Production is the one-line predicate only. #54 adds a rationale comment at 7118-7124 and the same `len(common) >= 2` guard (7125-7127 at `5754d3c37db915757a1676fb62204fbd5f0912ec`), but its tests hardcode `0.005`, skip `stderr_diff == 0.0`, skip `beats_control`, hand-build six `SMALL` keys, take unused `small_data`, and use control {0} + candidate {0,1} rather than the issue seed set. Focused there: `21 passed, 184 deselected in 12.71s`; fail-proof `1 failed, 1 passed, 203 deselected in 1.90s` at test line 895.

### Remaining human checks

- Full `tests/test_ipmnist_screening.py` was not run in this verify pass (PR body claims 205 passed). Sibling #54's broader `-k` filter is the `21 passed, 184 deselected in 12.71s` figure above.
- If the maintainer prefers the earlier issue-authored PR, fold these test pins into #54 and close this one. Do not land both.
- ruff/mypy were not re-run in this verify pass.

Client / agent tooling: Grok Build CLI
— [grok-build-ss251]

Kayvan-Zahiri:
Acknowledged. Leaving this open for maintainer sequencing against #54 rather than closing it from this side. The tests here still pin `CONFIRMATION_THRESHOLD`, `stderr_diff == 0.0` on n=1, and `SMALL.to_config()`. Happy to fold those into #54 if that is the merge path.

PR #58, opened 2026-08-14
Keep screening IDBD and Autostep meta-updates finite

Kayvan-Zahiri:
The IPMNIST screening copies of IDBD, Swift-IDBD, Autostep, and the alpha-utility gate still poison per-weight step-size state on a non-finite meta-gradient. Fresh traces are zero, so `inf * h` is NaN; `clip(NaN)` and Autostep's `max(NaN, v)` stay NaN, and every later finite update inherits it.

That is the same class core IDBD closed in #42/#43. Core Autostep is #55. These screening kernels are a separate copy on the open climbing lane, so a NaN here corrupts paired campaign measurements rather than just a library optimizer.

The meta-update now skips a non-finite channel and keeps the previous `log_alpha` / `v` / `alpha`. Autostep's M-factor and traces also ignore non-finite `z^2` so the global rescale cannot re-poison a skipped channel. On the poisoned path that is a semantic choice, not only a guard: non-finite `z^2` channels are dropped from the sum that builds `M`, so a skipped channel does not inflate the network-wide normalizer. Finite trajectories are unchanged: existing meta=0 reduction pins and the Swift-vs-IDBD bit-exact pin still pass.

Failing-test-first on inf gated/raw gradients: all four kernels went NaN on pristine main, then recovered. Restored the `params = new_params` carry in `test_meta_zero_reduces_to_half_gated_step` so that pin is still a 3-step trajectory. `tests/test_ipmnist_screening.py`: 207 passed.

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -203,6 +203,20 @@ _IDBD_LOG_ALPHA_MAX = 0.0  # alpha <= 1 keeps per-weight decay factors positive
 _AUTOSTEP_ALPHA_MIN = 1e-8
 _AUTOSTEP_ALPHA_MAX = 1.0
 
+
+def _clip_finite_log_alpha(log_alpha: Array, meta_delta: Array) -> Array:
+    """Clip an IDBD log-step-size update; skip a non-finite meta-gradient.
+
+    ``clip(NaN)`` is NaN, so inf * h=0 (or a float32 overflow of a finite
+    product) would permanently poison ``log_alpha``. Core IDBD (#42/#43)
+    keeps the previous value on that channel; screening copies must too.
+    """
+    return jnp.where(
+        jnp.isfinite(meta_delta),
+        jnp.clip(log_alpha + meta_delta, _IDBD_LOG_ALPHA_MIN, _IDBD_LOG_ALPHA_MAX),
+        log_alpha,
+    )
+
 # Metrics returned by every screening step: (accuracy, loss, plasticity).
 StepMetrics = tuple[Array, Array, Array]
 ScreeningStepFn = Callable[
@@ -402,10 +416,8 @@ def upgd_idbd_update(
     for name in params:
         keep = 1.0 - gate[name]
         z = grads[name] * keep
-        log_alpha = jnp.clip(
-            state.log_alpha[name] + meta * z * state.trace[name],
-            _IDBD_LOG_ALPHA_MIN,
-            _IDBD_LOG_ALPHA_MAX,
+        log_alpha = _clip_finite_log_alpha(
+            state.log_alpha[name], meta * z * state.trace[name]
         )
         alpha = jnp.exp(log_alpha)
         new_params[name] = params[name] * (1.0 - alpha * wd) - alpha * (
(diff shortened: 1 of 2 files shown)


swissyai (changes requested):
Independent verification against `main` `da8b86c`. This PR is **complementary to #56, not a duplicate** — it guards the screening file's own IDBD/Autostep/alpha-utility copies (which the merged #43 never touched), zero file overlap with #56's core fix. Both should land.

Verified: failing-test-first holds (4 new tests red on unpatched source, all green patched; 207 total pass); the shared `_clip_finite_log_alpha` helper preserves the original operand association, so finite-path values are unchanged; guards match #56's core semantics (`isfinite & gate`).

**One change requested:** the diff removes the `params = new_params` loop-carry at the end of the existing pinned test `test_meta_zero_reduces_to_half_gated_step`, weakening a trajectory pin into a repeated single-step check. I ran the ORIGINAL pin against this PR's source: **2 passed** — the removal is unnecessary, the pin holds as-is. Please restore that line; this repo treats weakened pins as load-bearing regressions even when accidental.

Minor, non-blocking: the `effective`-sum change (excluding non-finite `z²` channels from the M-normalization sum) is a semantic decision on the poisoned path, not just a guard — worth one sentence in the body saying so.

ss251 (approved):
Independent verification at exact head `733f4d5` (isolated worktree, `uv sync --extra dev`).

- This PR's suite at head: `tests/test_ipmnist_screening.py` — **207 passed** (~75s).
- Fail-proof: with only `alberta_framework/benchmarks/ipmnist_screening.py` restored to `origin/main` and the PR's tests kept — **4 failed, 203 passed**; all four failures are the new non-finite regressions catching the unguarded screening-embedded meta-updates. Genuine pin, not tautology.

Scope check that matters for this one: these are the screening benchmark's *embedded copies* of the IDBD/Autostep meta-updates, a separate surface from the core optimizers fixed in #43 and open in #56 — no overlap or conflict with either (different file, independent state). Guard shape matches the family precedent.

Same process note as on #60: the family's bit-stability probe receipt (deterministic finite-trajectory sha256 before/after) wasn't run for the screening copies in this pass; the diff is selection-only in the same pattern verified bit-stable on the core siblings. Worth one line from the author confirming screening outputs on a finite development slice are unchanged, since this file feeds measurement artifacts.

— [claude-code-ss251]

Kayvan-Zahiri:
Restored the `params = new_params` carry in `test_meta_zero_reduces_to_half_gated_step` so that pin is a 3-step trajectory again. The PR body now states that dropping non-finite `z^2` from the M-normalization sum is a semantic choice on the poisoned path, not only a guard.

swissyai (approved):
Re-verified at head `0637cef1e` against `main` `da8b86c`: the requested change is in and is the only change (diffed the two PR revisions — the `params = new_params` carry is restored, nothing else moved). The pinned `test_meta_zero_reduces_to_half_gated_step` runs as a 3-step trajectory again and passes, all four new NaN regressions pass, and the full focused file is 207-green on this host. The body now documents the M-normalization exclusion as a semantic decision.

Previous request addressed — approving. Complementary to #56 as before: core and screening surfaces together close the #42/#55 class.

— [swissyai]

PR #59, opened 2026-08-14
Keep TD-IDBD and AutoTDIDBD meta-updates finite

Kayvan-Zahiri:
TD-IDBD and AutoTDIDBD still poison per-weight step-size state on a non-finite TD error. Fresh h-traces are zero, so `inf * h` is NaN; `clip(NaN)` and AutoStep's `max(NaN, v)` stay NaN, and every later finite update inherits it. The bias twin has the same hole. Both the semi-gradient and ordinary-gradient TD-IDBD paths were affected.

That is the same class core IDBD closed in #42/#43. Core Autostep is #55/#56. Screening copies are #58. These are the TD optimizers used by `TDLinearLearner`, so a NaN here corrupts Step 3+ measurements rather than just a supervised library update.

The meta-update now skips a non-finite channel and keeps the previous log step-size / normalizer. Non-finite TD error also holds the previous h-trace so recovery is not sitting on inf/NaN correlation. Finite trajectories are unchanged: existing semi-vs-ordinary, terminal-state, and effective-step-size pins still pass.

Failing-test-first on inf TD error against pristine traces: both optimizers went NaN on main, then recovered. `tests/test_td_optimizers.py` and `tests/test_td_learners.py`: 48 passed.

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -1508,36 +1508,47 @@ class TDIDBD(TDOptimizer[TDIDBDState]):
 
         if self._use_semi_gradient:
             gradient_correlation = delta * observation * state.h_traces
-            new_log_step_sizes = state.log_step_sizes + theta * gradient_correlation
+            meta_delta = theta * gradient_correlation
         else:
             feature_diff = gamma_scalar * next_observation - observation
             gradient_correlation = delta * feature_diff * state.h_traces
-            new_log_step_sizes = state.log_step_sizes - theta * gradient_correlation
+            meta_delta = -theta * gradient_correlation
 
-        new_log_step_sizes = jnp.clip(new_log_step_sizes, -10.0, 2.0)
+        # clip(NaN) is NaN, so inf * h=0 must skip adaptation for that weight.
+        new_log_step_sizes = jnp.where(
+            jnp.isfinite(meta_delta),
+            jnp.clip(state.log_step_sizes + meta_delta, -10.0, 2.0),
+            state.log_step_sizes,
+        )
         new_alphas = jnp.exp(new_log_step_sizes)
 
         new_eligibility_traces = gamma_scalar * lam * state.eligibility_traces + observation
         weight_delta = new_alphas * delta * new_eligibility_traces
 
+        finite_delta = jnp.isfinite(delta)
         if self._use_semi_gradient:
             h_decay = jnp.maximum(0.0, 1.0 - new_alphas * observation * new_eligibility_traces)
-            new_h_traces = state.h_traces * h_decay + new_alphas * delta * new_eligibility_traces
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Independent verification at exact head `9a8429a`, from the reviewer who ran this same audit vein: our lane's TDIDBD entry (recorded before this PR was discovered) reached the identical mechanism — left-associative `delta * phi * h` with `h=0`, `clip(NaN)=NaN` on both the semi- and ordinary-gradient paths plus the bias twin — so this review is armed with pre-existing independent reproductions, not a read of the diff.

**Verdict: APPROVE.** Executed against this head:

- All of our independent repro cases pass: inf TD error against fresh zero traces on **both** `use_semi_gradient` settings (logs, bias, h-traces finite; next finite update recovers), the finite-overflow product case (`delta=1e20, phi=[1e20, 1]` → `[-4.6051702, -4.6051702]`, i.e. `log(0.01)` preserved on both channels), and AutoTDIDBD's normalizers under the same inputs (finite + recovers).
- **Bit-stability, per the #43/#56 precedent:** a deterministic 200-update finite TD trajectory hashing `log_step_sizes + h_traces + eligibility_traces` produces `b15c96588999c5dc0440b9179c0052f83d4c656f3164e2ea2e0379ef191e4e4a` on this head and the identical hash on pristine `origin/main` — the finite path is untouched.
- Fail-proof: with only `optimizers.py` restored to `origin/main` and this PR's tests kept, **3 fail with the exact NaN shapes**; restored, 18/18 (and the full optimizer suites pass 68/68).

Two observations, neither blocking:

1. The h-trace guard (`where(finite_delta, updated_h, h_prev)`) goes beyond the minimal meta-update repair — it also prevents inf traces from freezing future adaptation via permanently-inf meta-gradients. That's a deliberate scope extension relative to the #43 pattern and I think the right one for the TD family; worth knowing it was chosen (the PR body does say so).
2. Our lane also recorded the adjacent eligibility-trace store `gamma*λ*z + phi` as a separable `0*inf` NaN hole after an inf *feature* under default `λ=0` (same shape as the AutostepGTDLambda finding in #55's family). This PR's `finite_delta` guard doesn't cover an inf observation poisoning `z`; fine as scoped here, but flagging so it lands in someone's queue rather than falling between PRs.

Provenance: AI-assisted review — independent reproductions by a local Grok Build lane (xai/grok-4.6), verification runs and this review by anthropic/claude-fable-5 via Claude Code.

— [claude-code-ss251]

Kayvan-Zahiri:
Acknowledged on the adjacent inf-observation eligibility store (`gamma * lambda * z + phi` under default `lambda=0`). Leaving that out of this PR on purpose so the meta-update guard stays the same shape as #43. It should be a follow-up on the eligibility path, not a silent extra change here.

PR #62, opened 2026-08-14
Keep feature-learner ObGD bounds from turning 0*inf into NaN

Kayvan-Zahiri:
Fixed-budget feature discovery and the compositional DAG learner both default to ObGD bounding. Inf error correctly drives `scale` to 0, then multiplies that by an inf LMS step. `0 * inf` is NaN, which poisons output weights, feature weights, and utilities for every later finite update.

The interaction-feature learner already rejects infinite targets at the door. These two copies do not. They now replace those NaNs with zeros so the bound's intended no-op holds. Finite trajectories are unchanged.

Failing-test-first: inf targets produced NaN banks on main, then recovered. `TestFixedBudgetFeatureLearner` and `TestCompositionalFeatureLearner`: 48 passed.

Sibling of #61 (library ObGD / AGC / UPGD kappa path). Interaction features already fail closed on inf, so they are left alone.

### DIFF
--- a/alberta_framework/core/compositional_features.py
+++ b/alberta_framework/core/compositional_features.py
@@ -2827,6 +2827,27 @@ class CompositionalFeatureLearner:
             theta_delta = bounding_scale * theta_delta
             candidate_output_delta = bounding_scale * candidate_output_delta
             candidate_theta_delta = bounding_scale * candidate_theta_delta
+            output_delta = jnp.where(
+                jnp.isnan(output_delta), jnp.zeros_like(output_delta), output_delta
+            )
+            output_bias_delta = jnp.where(
+                jnp.isnan(output_bias_delta),
+                jnp.zeros_like(output_bias_delta),
+                output_bias_delta,
+            )
+            theta_delta = jnp.where(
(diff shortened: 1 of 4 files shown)


ss251 (approved):
Independent verification at exact head `455d56e` (isolated worktree, `uv sync --extra dev`).

- This PR's suites at head: tests/test_compositional_features.py + tests/test_feature_discovery.py — 120 passed — all green.
- Fail-proof: with only the production files (compositional_features.py + feature_discovery.py (feature-learner ObGD bounds)) restored to `origin/main` and the PR's tests kept — **2 failed, 118 passed**; every failure is a new non-finite regression catching the unguarded `0*inf` path, so the tests genuinely pin the fix.

Same family as #42/#43, #55/#56, #59, #60: the `where(isfinite, …)` selection-only guard shape, previous state kept on non-finite input, finite paths structurally untouched. Same process note as the sibling reviews: no per-file bit-stability probe was run in this pass; the diff is selection-only in the pattern verified bit-stable on three core siblings.

— [claude-code-ss251]

PR #63, opened 2026-08-14
Keep MLPLearner from committing ObGD 0*inf NaN weights

Kayvan-Zahiri:
MLPLearner still commits weight updates after ObGD bounding. Inf error drives the bound scale to 0, then `param += error * step` is `0 * inf` = NaN, and every later finite target inherits a poisoned net.

MultiHead already refuses that: inf is not a legal target (NaN is the missing-head sentinel), and a proposed state with non-finite floats is not committed. MLPLearner now uses the same fail-closed rule. The previous finite parameters, traces, and step count stay put so a later finite target can recover.

Failing-test-first: inf target + ObGDBounding wrote NaN weights on main, then kept the previous finite state. `tests/test_mlp_learner.py`: 74 passed.

Sibling of #61 (bounder-side 0*inf). This is the apply site the bounder cannot see.

### DIFF
--- a/alberta_framework/core/learners.py
+++ b/alberta_framework/core/learners.py
@@ -774,6 +774,16 @@ def metrics_to_dicts(metrics: Array, normalized: bool = False) -> list[dict[str,
 # =============================================================================
 
 
+def _floating_tree_is_finite(tree: object) -> Array:
+    """Return whether every floating/complex persistent leaf is finite."""
+    valid = jnp.asarray(True, dtype=jnp.bool_)
+    for leaf in jax.tree.leaves(tree):
+        array = jnp.asarray(leaf)
+        if jnp.issubdtype(array.dtype, jnp.inexact):
+            valid = valid & jnp.all(jnp.isfinite(array))
+    return valid
+
+
 class MLPLearner:
     """Multi-layer perceptron with composable optimizer, bounder, and normalizer.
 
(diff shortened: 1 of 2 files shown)


deepanshu-yd (approved):
Static review of the apply-site hold on `MLPLearner`.

This is the right shape, and it is **not** the #61 `where(isnan)` sanitizer. Inf target / non-finite proposed tree keeps the previous finite parameters via `lax.cond`. MultiHead already refuses that; matching it here is correct. Tests pin weights bitwise-equal to the pre-update state and a later finite target still updates.

Non-blocking: rejected updates still compute `squared_error = error**2` from the inf target, so metrics can be non-finite even though weights are held. Fine for this PR; do not fold a diagnostic rewrite in unless you want it.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/63", "headSha": "2e3e1cf5dc910ea7e78c774e3a234d4db4f28347", "recommendation": "partial", "reproduced": false, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": ["MLPLearner update_applied hold"], "commands": [], "evidenceUrls": [], "summary": "MLPLearner refuses to commit a non-finite proposed state; no blanket NaN rewrite. Static only."}
```

— [deepanshu-yd]

PR #64, opened 2026-08-14
Keep actor-critic ObGD applies from turning 0*inf into NaN

Kayvan-Zahiri:
Actor-critic still multiplies the bounded step by the TD error after ObGD. Inf reward drives the bound scale to 0, then `td_error * step` is `0 * inf` = NaN, and both actor and critic stay poisoned. The continuous Gaussian core has the same apply site.

The previous finite weights and traces are kept. Last observation and action still advance so a scan loop does not desynchronize from the stream. A later finite reward can recover.

Failing-test-first: inf reward + ObGDBounding wrote NaN weights on main, then kept the previous finite parameters. `tests/test_actor_critic.py` and `tests/test_continuous_actor_critic.py`: 24 passed.

Sibling of #61 (bounder) and #63 (MLPLearner apply site).

### DIFF
--- a/alberta_framework/core/actor_critic.py
+++ b/alberta_framework/core/actor_critic.py
@@ -32,6 +32,16 @@ from jaxtyping import Float, Int
 from alberta_framework.core.optimizers import Bounder, bounder_from_config
 
 
+def _floating_tree_is_finite(tree: object) -> Array:
+    """Return whether every floating/complex persistent leaf is finite."""
+    valid = jnp.asarray(True, dtype=jnp.bool_)
+    for leaf in jax.tree.leaves(tree):
+        array = jnp.asarray(leaf)
+        if jnp.issubdtype(array.dtype, jnp.inexact):
+            valid = valid & jnp.all(jnp.isfinite(array))
+    return valid
+
+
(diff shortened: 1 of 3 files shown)


deepanshu-yd (approved):
Static review of the actor-critic apply-site hold.

`params_ok = inputs_valid & _floating_tree_is_finite(updated)`, then `lax.cond` keeps previous weights but still increments `step_count` and then writes `last_observation` / `last_action`. That is the scan-safe contract the PR body claimed. Discrete and continuous paths both get it. Tests pin finite actor/critic weights under inf reward + ObGD.

Not the #61 `where(isnan)` mask. No blocking finding.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/64", "headSha": "86b0eab79149327e450d8c6c0e1c50ab9979552f", "recommendation": "partial", "reproduced": false, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": ["actor-critic scan-safe hold"], "commands": [], "evidenceUrls": [], "summary": "Actor-critic holds finite weights on inf reward and still advances last_observation/action. Static only."}
```

— [deepanshu-yd]

PR #66, opened 2026-08-14
Keep Horde actor-critic ObGD applies from turning 0*inf into NaN

Kayvan-Zahiri:
Horde-backed actors still multiply the bounded step by the TD error after ObGD. Inf reward drives the bound scale to 0, then `td_error * step` is `0 * inf` = NaN, and the actor stays poisoned. The same apply site is in QHorde, NonlinearHorde, and NonlinearQHorde.

The previous finite weights and traces are kept. Last observation and action still advance so a scan loop does not desynchronize from the stream. A later finite reward can recover.

Failing-test-first: inf reward + ObGDBounding wrote NaN actor weights on main, then kept the previous finite parameters. `tests/test_horde_actor_critic.py`: 50 passed.

Sibling of #61 (bounder), #63 (MLPLearner apply site), and #64 (linear actor-critic apply site). Independent of those PRs.

### DIFF
--- a/alberta_framework/core/horde_actor_critic.py
+++ b/alberta_framework/core/horde_actor_critic.py
@@ -54,6 +54,41 @@ from alberta_framework.core.optimizers import (
 from alberta_framework.core.types import AutostepParamState, DemonType, MLPParams
 
 
+def _floating_tree_is_finite(tree: object) -> Array:
+    """Return whether every floating/complex persistent leaf is finite."""
+    valid = jnp.asarray(True, dtype=jnp.bool_)
+    for leaf in jax.tree.leaves(tree):
+        array = jnp.asarray(leaf)
+        if jnp.issubdtype(array.dtype, jnp.inexact):
+            valid = valid & jnp.all(jnp.isfinite(array))
+    return valid
+
+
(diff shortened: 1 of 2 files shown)


deepanshu-yd (approved):
Static review. `_commit_scan_safe_actor_state` is the same hold-weights / advance-clock contract as #64, applied at the Horde actor-critic sites. Inputs and `td_error` must be finite, proposed tree must be finite, otherwise previous weights are kept. No `where(isnan)` sanitizer.

No blocking finding on the raw diff. Family packaging vs #64 remains a maintainer call.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/66", "headSha": "0a3900b944dffd8ed757652e0ce4cea8d054ade2", "recommendation": "partial", "reproduced": false, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": [], "commands": [], "evidenceUrls": [], "summary": "Horde AC uses a scan-safe finite-tree hold, not a blanket NaN mask. Static only."}
```

— [deepanshu-yd]

PR #67, opened 2026-08-14
Keep IndependentDemonHorde ObGD applies from turning 0*inf into NaN

Kayvan-Zahiri:
IndependentDemonHorde still multiplies the bounded step by the TD error after ObGD. An inf cumulant drives the bound scale to 0, then `error * step` is `0 * inf` = NaN, and that demon stays poisoned. NaN cumulants already skip the update; inf does not.

The previous finite params, traces, and optimizer states are kept on that demon. A sibling with a finite cumulant still updates. Outer step_count still advances. A later finite cumulant can recover.

Failing-test-first: inf cumulant + ObGDBounding wrote NaN weights on main, then kept the previous finite parameters. `tests/test_independent_demon_horde.py`: 11 passed.

Sibling of #61 (bounder), #63 (MLPLearner apply site), and #66 (Horde actor-critic apply site). Independent of those PRs.

### DIFF
--- a/alberta_framework/core/independent_demon_horde.py
+++ b/alberta_framework/core/independent_demon_horde.py
@@ -60,6 +60,17 @@ from alberta_framework.core.types import (
     TraceMode,
 )
 
+
+def _floating_tree_is_finite(tree: object) -> Array:
+    """Return whether every floating/complex persistent leaf is finite."""
+    valid = jnp.asarray(True, dtype=jnp.bool_)
+    for leaf in jax.tree.leaves(tree):
+        array = jnp.asarray(leaf)
+        if jnp.issubdtype(array.dtype, jnp.inexact):
+            valid = valid & jnp.all(jnp.isfinite(array))
+    return valid
+
+
 # =============================================================================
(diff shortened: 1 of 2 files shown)


deepanshu-yd (approved):
Static review. `commit = active & finite(obs) & finite(error) & finite(proposed tree)` then the existing `where(commit, new, old)` slots. That refuses a 0*inf ObGD apply without rewriting NaN to 0. Tests pin finite weights after inf error.

No blocking finding.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/67", "headSha": "70113fe17472d7e9c32a8f7a500aa3e6e7b9c753", "recommendation": "partial", "reproduced": false, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": [], "commands": [], "evidenceUrls": [], "summary": "IndependentDemonHorde commit mask now requires finite inputs and a finite proposed tree. Static only."}
```

— [deepanshu-yd]

PR #68, opened 2026-08-14
Keep OffPolicyHorde ObGD applies from turning 0*inf into NaN

Kayvan-Zahiri:
OffPolicyHordeLearner still multiplies the bounded head step by the TD error after ObGD. An inf cumulant drives the bound scale to 0, then `error_i * step` is `0 * inf` = NaN, and that head stays poisoned. NaN cumulants already skip the update; inf does not. The trunk bounder uses error=1.0, so this hole is head-only.

The previous finite head params, traces, and optimizer states are kept on the inf head. A sibling with a finite cumulant still updates. Outer step_count still advances. A later finite cumulant can recover.

Failing-test-first: inf cumulant + ObGDBounding wrote NaN head weights on main, then kept the previous finite parameters. `tests/test_off_policy_horde.py`: 10 passed.

Sibling of #61 (bounder), #63 (MLPLearner apply site), and #67 (IndependentDemonHorde apply site). Independent of those PRs.

### DIFF
--- a/alberta_framework/core/off_policy_horde.py
+++ b/alberta_framework/core/off_policy_horde.py
@@ -587,18 +587,25 @@ class OffPolicyHordeLearner:
 
             new_w = head_w + error_i * w_step
             new_b = head_b + error_i * b_step
-
-            new_w = jnp.where(active_mask[i], new_w, head_w)
-            new_b = jnp.where(active_mask[i], new_b, head_b)
-            new_w_trace = jnp.where(active_mask[i], new_w_trace, old_w_trace)
-            new_b_trace = jnp.where(active_mask[i], new_b_trace, old_b_trace)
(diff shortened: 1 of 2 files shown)


deepanshu-yd (approved):
Static review. Same commit-if-finite pattern as #67: `head_ok` replaces `active_mask` so a non-finite ObGD apply does not land. No blanket `where(isnan)`.

No blocking finding.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/68", "headSha": "50710e6a3c3bb410852a6514bc17299f006b016d", "recommendation": "partial", "reproduced": false, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": [], "commands": [], "evidenceUrls": [], "summary": "OffPolicyHorde holds a head when the proposed apply is non-finite. Static only."}
```

— [deepanshu-yd]

PR #69, opened 2026-08-14
Keep RTU ObGD updates from turning 0*inf into NaN

Kayvan-Zahiri:
The RTU actor-critic has its own ObGD, not the Bounder ABC. Inf TD error makes the bound scale 0, then `step_size * signal * z` is `0 * inf` = NaN. Kappa 0 still hits `0 * inf` in the denominator product, so even the unbounded path NaNs. Adaptive ObGD also NaNs the second-moment tree when the trace is 0.

NaN products are mapped to 0 so a bounded inf signal writes a zero update. Unbounded inf (`kappa=0`) stays inf. Finite large-error contracts are unchanged.

Failing-test-first: inf signal wrote NaN updates on main, then wrote zeros (bounded) or inf (unbounded). The four ObGD contract tests in `tests/test_recurrent_trace_actor_critic.py` passed.

Sibling of #61 (Bounder-side ObGD). This copy lives in `recurrent_trace_actor_critic.py` and does not depend on that PR.

### DIFF
--- a/alberta_framework/core/recurrent_trace_actor_critic.py
+++ b/alberta_framework/core/recurrent_trace_actor_critic.py
@@ -1217,6 +1217,11 @@ def exact_rtrl_gradient(
     )
 
 
+def _replace_nan_with_zero(value: Array) -> Array:
+    """Map ``0 * inf`` NaNs to 0 so genuine infs stay inf."""
+    return jnp.where(jnp.isnan(value), jnp.zeros_like(value), value)
+
+
 def obgd_update(
     traces: Any,
     signal: Array,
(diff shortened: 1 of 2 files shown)


Kayvan-Zahiri:
Addressed the masking concern from the family review on #61.

RTU ObGD no longer does `where(isnan, 0)`. It zeros only the `0 * inf` product when a bound collapses an inf signal, or when `kappa` itself is the zero in `0 * inf`. Inactive ObGD (`kappa=0`) still returns inf. A genuine NaN signal stays NaN (new assertion on that path).

deepanshu-yd (approved):
Static review of head `3d5dc5b`. This is the one in the 0*inf series that already has the correct gate:

- `_zero_if_collapsed_inf(product, infinite_input, collapsed)` only rewrites when the bound collapsed **and** the input is inf **and** the product is NaN
- `_zero_if_zero_times_inf` only rewrites NaN when one operand is inf and the other is exactly 0
- genuine NaN is pinned to stay NaN

That is exactly the follow-up #61 required. swissyai already reproduced the poison tests red-on-main / green-on-head in a comment on #61; this is the formal review on this PR, which had none. I did not re-execute.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/69", "headSha": "3d5dc5b99234cd1cd5efdd7b4ea39cb3046a9b50", "recommendation": "partial", "reproduced": false, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": ["_zero_if_collapsed_inf", "_zero_if_zero_times_inf"], "commands": [], "evidenceUrls": ["https://github.com/elizaOS/asi/pull/61"], "summary": "RTU uses collapse-gated 0*inf helpers; genuine NaN stays NaN. Static confirm of the #61 follow-up."}
```

— [deepanshu-yd]

PR #70, opened 2026-08-14
Keep IDBD traces from turning 0*inf into NaN

Kayvan-Zahiri:
#42 already skips the IDBD meta-update when `error * x * h` is non-finite, so log step-sizes stay finite. The h-trace and weight delta still do `alpha * error * x`. A silent feature (`x=0`) against an inf error is `0 * inf` = NaN, and that channel stays poisoned through later finite updates. Nonzero features still take an unbounded inf step.

NaN products are mapped to 0. Genuine infs stay inf. The existing log-alpha guard is unchanged.

Failing-test-first: inf error with a zero feature wrote NaN traces and weight deltas on main, then wrote 0 on that channel. `tests/test_optimizers.py`: 52 passed.

Follow-on to #42 (meta-update). Independent of #56 (Autostep) and #61 (ObGD bounder).

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -699,7 +699,10 @@ class IDBD(Optimizer[IDBDState]):
         # when error is None (trunk path), z is already loss gradient direction.
         decay = jnp.maximum(0.0, 1.0 - new_alphas * h_decay)
         if error is not None:
-            new_traces = state.traces * decay - new_alphas * jnp.squeeze(error) * z
+            # 0 * inf is NaN; a silent feature must not poison h.
+            h_step = new_alphas * jnp.squeeze(error) * z
+            h_step = jnp.where(jnp.isnan(h_step), jnp.zeros_like(h_step), h_step)
+            new_traces = state.traces * decay - h_step
         else:
             new_traces = state.traces * decay + new_alphas * z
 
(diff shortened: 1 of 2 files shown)


deepanshu-yd (changes requested):
Blocking: blanket `jnp.where(jnp.isnan(h_step), 0, h_step)` (and the same on `weight_delta` / `trace_step`) in `update` and `update_from_gradient`.

This does fix `0 * inf` on a silent feature. It also rewrites a **genuine NaN error** (or NaN alpha) on a **nonzero** feature into a zero update, so the next finite step looks healthy. That is the defect 0xcii blocked on #61 and that #69 already closed with `_zero_if_zero_times_inf` (NaN **and** one operand inf, the other exactly 0).

Please copy that predicate and add a negative test: NaN error, `x = [1, 0]` → channel 0 stays NaN, not 0.

#72 and #73 edit the same `optimizers.py` from this same base; they will conflict. Maintainer sequencing, not a reason to keep the mask.

```slop-review
{"schemaVersion":"1","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/70","headSha":"89e6242521daeaeb0531b683693063fea6862497","recommendation":"reject","reproduced":false,"securityRisk":"none","duplicateRisk":"none","usefulArtifacts":[],"commands":[],"evidenceUrls":[],"summary":"Blanket where(isnan) on IDBD traces/deltas zeros genuine NaNs. Needs the 0*inf-only gate from #61/#69."}
```

— [deepanshu-yd]

lalalune:

PR #71, opened 2026-08-14
Keep baseline optimizer state from turning 0*inf into NaN

Kayvan-Zahiri:
AdaGain's meta-update is the same `inf * h=0` hole as IDBD #42: `clip(exp(NaN))` is NaN and every later gain stays poisoned. Adam and RMSprop have a second hole: an inf error makes `m / sqrt(v)` into `inf / inf` = NaN, so moments never recover. NADALINE's silent-feature apply is `alpha * inf * 0` = NaN.

AdaGain skips the meta-update when the correlation is non-finite and maps `0 * inf` products to 0. Adam/RMSprop hold that channel's moments and write a zero step when the RMS ratio is NaN. NADALINE maps a NaN delta to 0. Genuine infs stay inf. Finite contracts are unchanged.

Failing-test-first: inf error wrote NaN step-sizes (AdaGain), NaN moments (Adam/RMSprop), and a NaN silent-feature delta (NADALINE) on main. `tests/test_baseline_optimizers.py`: 29 passed.

Sibling of #42 (IDBD meta-update) and #70 (IDBD traces). Different module (`baseline_optimizers.py`).

### DIFF
--- a/alberta_framework/core/baseline_optimizers.py
+++ b/alberta_framework/core/baseline_optimizers.py
@@ -43,6 +43,12 @@ from jaxtyping import Float
 
 from alberta_framework.core.optimizers import Optimizer, OptimizerUpdate
 
+
+def _replace_nan_with_zero(value: Array) -> Array:
+    """Map ``0 * inf`` NaNs to 0 so genuine infs stay inf."""
+    return jnp.where(jnp.isnan(value), jnp.zeros_like(value), value)
+
+
 # =============================================================================
 # State dataclasses
 # =============================================================================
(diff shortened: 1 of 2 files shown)


deepanshu-yd (changes requested):
Blocking: `_replace_nan_with_zero` is `where(isnan(value), 0, value)` applied to AdaGain/Adam-style gradients and deltas.

The docstring says it maps `0 * inf` and leaves genuine infs. It does **not** leave genuine NaNs. A NaN error on a nonzero feature becomes a zero gradient, then a finite-looking Adam/AdaGain step. Same defect as the original #61 sanitizer.

Please gate the rewrite on `isnan(product) & ((isinf(a) & (b == 0)) | (isinf(b) & (a == 0)))` (or the #69 helper) and pin a genuine-NaN negative test.

The `isfinite` hold on the meta-update / moments is the right shape; only the broad NaN rewrite needs to change.

```slop-review
{"schemaVersion":"1","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/71","headSha":"40864c8cc081a50a853b8f8226659302d451b9f5","recommendation":"reject","reproduced":false,"securityRisk":"none","duplicateRisk":"none","usefulArtifacts":[],"commands":[],"evidenceUrls":[],"summary":"baseline_optimizers._replace_nan_with_zero erases genuine upstream NaNs. Needs the 0*inf-only gate."}
```

— [deepanshu-yd]

lalalune:

PR #72, opened 2026-08-14
Keep LMS silent-feature deltas from turning 0*inf into NaN

Kayvan-Zahiri:
LMS still does `alpha * error * x`. A silent feature (`x=0`) against an inf error is `0 * inf` = NaN, so that channel's delta is not a real 0 and any caller that adds it poisons the weight. Nonzero features still take an unbounded inf step.

NaN products are mapped to 0. Genuine infs stay inf. LMS has no persistent optimizer state, so a later finite error on the same silent feature is a real 0.

Failing-test-first: inf error with a zero feature wrote a NaN weight delta on main, then wrote 0. `tests/test_optimizers.py::TestLMS`: 4 passed.

Sibling of #70 (IDBD traces) and #71 (NADALINE). Independent of #61 (ObGD bounder) even though both touch `optimizers.py`.

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -504,8 +504,11 @@ class LMS(Optimizer[LMSState]):
         alpha = state.step_size
         error_scalar = jnp.squeeze(error)
 
-        # Weight update: alpha * error * x
+        # Weight update: alpha * error * x. 0 * inf is NaN; genuine infs stay inf.
         weight_delta = alpha * error_scalar * observation
+        weight_delta = jnp.where(
+            jnp.isnan(weight_delta), jnp.zeros_like(weight_delta), weight_delta
+        )
 
         # Bias update: alpha * error
         bias_delta = alpha * error_scalar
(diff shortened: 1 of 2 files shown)


deepanshu-yd (changes requested):
Blocking: LMS `weight_delta = where(isnan(alpha * error * x), 0, ...)`.

`NaN * 1.0` is NaN, so a genuine NaN error on a **nonzero** feature is rewritten to a zero delta. The test only covers `inf * x=0`. Please use the #61/#69 0*inf-only predicate and add a NaN-error / nonzero-x negative test.

This PR shares `optimizers.py` with #70 and #73 from the same base.

```slop-review
{"schemaVersion":"1","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/72","headSha":"166b9e9663c519429271da67234cfc8076af6bf9","recommendation":"reject","reproduced":false,"securityRisk":"none","duplicateRisk":"none","usefulArtifacts":[],"commands":[],"evidenceUrls":[],"summary":"LMS where(isnan) zeros a genuine NaN error on nonzero features. Needs the 0*inf-only gate."}
```

— [deepanshu-yd]

lalalune:

PR #73, opened 2026-08-14
Keep Autostep traces from turning 0*inf into NaN

Kayvan-Zahiri:
#56 already skips Autostep's Eq. 4/5 meta-update when the correlation is non-finite, so step-sizes stay finite. The h-trace and weight delta still do `alpha * error * x`. A silent feature (`x=0`) against an inf error is `0 * inf` = NaN, and that channel stays poisoned. Nonzero features still take an unbounded inf step.

NaN products are mapped to 0. Genuine infs stay inf. Normalizers and the v>0 gate are left to #56.

Failing-test-first: inf error with a zero feature wrote NaN traces and weight deltas on main, then wrote 0 on that channel. The same hole is in `update_from_gradient`. `tests/test_optimizers.py`: 52 passed.

Follow-on to #56 (Autostep meta) and sibling of #70 (IDBD traces). The changed lines sit after #56's Eq. 5 hunks.

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -970,7 +970,12 @@ class Autostep(Optimizer[AutostepState]):
         # Trace update: h_i = h_i*(1 - α_i*z_i²) + α_i*δ*z_i
         trace_decay = 1.0 - new_step_sizes * z_sq
         if error is not None:
-            new_traces = state.traces * trace_decay + new_step_sizes * error_scalar * z
+            # 0 * inf is NaN; a silent feature must not poison h.
+            trace_step = new_step_sizes * error_scalar * z
+            trace_step = jnp.where(
+                jnp.isnan(trace_step), jnp.zeros_like(trace_step), trace_step
+            )
+            new_traces = state.traces * trace_decay + trace_step
         else:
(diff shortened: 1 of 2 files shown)


deepanshu-yd (changes requested):
Blocking: Autostep `trace_step` / `weight_delta` use `where(isnan(...), 0, ...)`. Same class as #70/#72 and the original #61 sanitizer.

A genuine NaN error on a nonzero feature becomes a zero trace step and a zero weight delta. Copy `#69`'s `_zero_if_zero_times_inf` (or the equivalent predicate) and add a negative test.

Shares `optimizers.py` with #70 and #72.

```slop-review
{"schemaVersion":"1","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/73","headSha":"1462ef52cda69f179eb65a4eef670f8f0aea4ab5","recommendation":"reject","reproduced":false,"securityRisk":"none","duplicateRisk":"none","usefulArtifacts":[],"commands":[],"evidenceUrls":[],"summary":"Autostep where(isnan) zeros genuine NaN traces/deltas. Needs the 0*inf-only gate."}
```

— [deepanshu-yd]

lalalune:

PR #74, opened 2026-08-14
Refuse inf rewards before Differential TD turns 0*inf into NaN

Kayvan-Zahiri:
Differential SARSA in this file already refuses non-finite rewards. Linear Differential TD and GTD still did `alpha * delta * z`. A silent feature against an inf reward is `0 * inf` = NaN, and rbar goes inf so later finite rewards stay poisoned.

Hold the previous finite state when the observation, reward, next observation, or proposed weights are non-finite.

The shared-trunk average-reward Horde had the same hole one layer up: inf is not NaN, so an inf cumulant stayed "active", MultiHead refused the whole trunk, and rbar still took the inf TD step. Non-finite cumulants are now inactive, and rbar only moves when the nested learner commits.

Failing-test-first: inf reward wrote `[nan, inf]` weights on main. Inf cumulant on one demon wrote inf then NaN into rbar. `tests/test_average_reward.py`: 21 passed.

Sibling of the LMS/Autostep silent-feature guards. Independent of #73.

### DIFF
--- a/alberta_framework/core/average_reward.py
+++ b/alberta_framework/core/average_reward.py
@@ -846,7 +846,10 @@ class AverageRewardHordeLearner:
     ) -> AverageRewardHordeUpdateResult:
         """Apply one shared-trunk differential Horde update."""
         next_predictions = self._learner.predict(state.learner_state, next_observation)
-        active = ~jnp.isnan(cumulants)
+        # NaN is the inactive-head sentinel. Inf is not NaN, so it used to
+        # stay "active", MultiHead refused the whole trunk, and rbar still
+        # took an inf step.
+        active = jnp.isfinite(cumulants)
         targets = cumulants - state.average_rewards + next_predictions
         targets = jnp.where(active, targets, jnp.nan)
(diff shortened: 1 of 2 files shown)


deepanshu-yd (approved):
Static review. Differential TD now treats only **finite** cumulants as active (was `~isnan`, which treated inf as active) and refuses the weight/average-reward update unless inputs and the proposed tree are finite. `average_reward` / `step_count` are held when `update_applied` is false.

That is fail-closed, not a `where(isnan)` rewrite. No blocking finding.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/74", "headSha": "a2a1b0868e46f373e3663f66e701976d82b4411a", "recommendation": "partial", "reproduced": false, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": [], "commands": [], "evidenceUrls": [], "summary": "Differential TD refuses inf rewards before the 0*inf apply and holds average_reward. Static only."}
```

— [deepanshu-yd]

PR #75, opened 2026-08-14
Refuse inf rewards before off-policy TD turns 0*inf into NaN

Kayvan-Zahiri:
Off-policy TD, ETD, and Gradient-TD all do `alpha * delta * z`. A silent feature against an inf reward is `0 * inf` = NaN, and that channel stays poisoned after later finite rewards.

Hold the previous finite state when the observation, reward, next observation, gamma, rho, or proposed weights are non-finite. Same contract as Differential SARSA / the Differential TD PR.

Failing-test-first: inf reward wrote `[nan, inf]` weights on main for all three learners. `tests/test_off_policy_td.py`: 23 passed.

Independent of #74 (average-reward file).

### DIFF
--- a/alberta_framework/core/off_policy_td.py
+++ b/alberta_framework/core/off_policy_td.py
@@ -62,6 +62,17 @@ from jaxtyping import Float, Int
 
 from alberta_framework.core.types import Observation
 
+
+def _floating_tree_is_finite(tree: object) -> Array:
+    """Return whether every floating/complex persistent leaf is finite."""
+    valid = jnp.asarray(True, dtype=jnp.bool_)
+    for leaf in jax.tree.leaves(tree):
+        array = jnp.asarray(leaf)
+        if jnp.issubdtype(array.dtype, jnp.inexact):
+            valid = valid & jnp.all(jnp.isfinite(array))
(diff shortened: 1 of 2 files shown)


deepanshu-yd (approved):
Static review. Off-policy TD / ETD / GTD proposed states are committed only when inputs and the proposed tree are finite. No `where(isnan)` sanitizer.

Non-blocking: the reject path is `lambda: state`, so `step_count` and traces do not advance. That is safer than committing a poisoned trace. #64 chose to increment the clock; either is defensible here because this state has no `last_observation` to desynchronize.

No blocking finding.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/75", "headSha": "6b386cd7a986273199f2bbce6147c1f2ed7e0d25", "recommendation": "partial", "reproduced": false, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": [], "commands": [], "evidenceUrls": [], "summary": "Off-policy TD holds the previous finite state instead of committing a 0*inf apply. Static only."}
```

— [deepanshu-yd]

PR #83, opened 2026-08-15
Hold True Online TD finite on inf reward times a silent feature

Kayvan-Zahiri:
## Summary
- True Online TD(lambda) Dutch-trace updates multiply `update_scale` by traces and `alpha * correction` by the observation. Inf reward with a silent feature is `0 * inf = NaN`, which writes NaN into one weight while the other goes to inf.
- Fail-closed: if the observation, reward, next observation, gamma, or proposed state is non-finite, keep the previous finite weights, traces, and `v_old`.
- This is the same hole as LMS / off-policy TD / options, in `TrueOnlineTDLearner` only. It does not change the MLP ObGD apply site on #63; the hunks do not overlap.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_true_online_td.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/learners.py tests/test_true_online_td.py`

### DIFF
--- a/alberta_framework/core/learners.py
+++ b/alberta_framework/core/learners.py
@@ -28,7 +28,7 @@ import chex
 import jax
 import jax.numpy as jnp
 from jax import Array
-from jaxtyping import Float
+from jaxtyping import Bool, Float
 
 from alberta_framework.core.initializers import sparse_init
 from alberta_framework.core.normalizers import (
@@ -1752,6 +1752,11 @@ class TrueOnlineTDUpdateResult:
 
     ``metrics`` columns: squared TD error, TD error, mean absolute
     eligibility trace, and the stored ``v_old`` bootstrap value.
+
+    Rejected updates (non-finite input or proposed state) keep the previous
+    learner state and report a zero TD-error diagnostic. ``update_applied``
+    is False on that path so callers do not treat the sentinel as a real
+    TD error.
     """
 
     state: TrueOnlineTDState
@@ -1759,6 +1764,7 @@ class TrueOnlineTDUpdateResult:
     next_prediction: Prediction
     td_error: Float[Array, ""]
     metrics: Float[Array, " 4"]
+    update_applied: Bool[Array, ""]
 
 
 class TrueOnlineTDLearner:
@@ -1864,15 +1870,17 @@ class TrueOnlineTDLearner:
             & jnp.isfinite(proposed_state.bias_eligibility_trace)
             & jnp.isfinite(proposed_state.v_old)
         )
+        accepted = inputs_valid & proposed_finite
         new_state = jax.lax.cond(
-            inputs_valid & proposed_finite,
+            accepted,
             lambda: proposed_state,
             lambda: state,
         )
+        reported_td_error = jnp.where(accepted, td_error, jnp.zeros_like(td_error))
         metrics = jnp.array(
             [
-                td_error**2,
-                td_error,
+                reported_td_error**2,
+                reported_td_error,
                 jnp.mean(jnp.abs(new_state.eligibility_traces)),
                 new_state.v_old,
             ],
@@ -1882,8 +1890,9 @@ class TrueOnlineTDLearner:
             state=new_state,
             prediction=jnp.atleast_1d(value),
             next_prediction=jnp.atleast_1d(next_value),
-            td_error=jnp.atleast_1d(td_error),
+            td_error=jnp.atleast_1d(reported_td_error),
             metrics=metrics,
+            update_applied=accepted,
         )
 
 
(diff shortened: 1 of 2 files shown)


alberta_framework/core/learners.py
```diff
@@ -1849,12 +1849,32 @@ def update(
...
+            lambda: proposed_state,
+            lambda: state,
+        )
         metrics = jnp.array(
             [
                 td_error**2,
                 td_error,
-                jnp.mean(jnp.abs(new_traces)),
-                new_v_old,
+                jnp.mean(jnp.abs(new_state.eligibility_traces)),
```
0xcii, line 1884: [P1] Keep rejected-update diagnostics finite or explicitly invalid

Switching these two entries to the held state does not sanitize the first two metric entries: with the PR's `inf`-reward case, `td_error**2` and `td_error` are still `inf`, and `result.td_error` is also `inf`. Thus the learner state survives while downstream metric aggregation and measurement artifacts remain poisoned. The test only asserts state finiteness. Please make the whole rejected result fail closed (finite sentinel plus explicit validity/rejection signal, or another documented contract) and assert every returned metric and `result.td_error`.

0xcii (changes requested):
I independently reviewed exact head `ef047f5182e9340a05420fdea3c7d6bef5b693e5`. Holding the previous finite learner state is useful, but the returned result remains measurement-poisoning on the exact rejected input: `td_error`, `metrics[0]`, and `metrics[1]` are still non-finite. The new test only checks state fields, so this incomplete fail-closed path passes unnoticed.

I reproduced the result construction in a disposable, network-disabled, read-only JAX container (2 CPUs, 2 GiB RAM, 128 PIDs):

```json
{"state_held_finite": true, "returned_td_error_finite": false, "metrics_all_finite": false, "nonfinite_metric_indices": [0, 1]}
```

Please make the rejected update's complete observable result fail closed: return a finite/sentinel diagnostic or an explicit validity signal that consumers must handle, rather than emitting non-finite measurement values alongside a held state. Add assertions for `result.td_error`, every metric, and any documented rejection indicator.

Compute receipt: 1966041 project-attributed tokens (bounded; device-signed, locally reported)
Client / agent tooling: codex
— [true-online-td-measurement-integrity]

Kayvan-Zahiri:
Addressed the blocking finding.

A rejected True Online TD update still holds the previous finite state. It now also reports a zero TD-error diagnostic (`td_error` and `metrics[0:2]`) and sets `update_applied=False`, so measurement artifacts are not poisoned by inf. The test asserts those fields.

alberta_framework/core/learners.py
Kayvan-Zahiri: Follow-up `ec93a2f` addresses the measurement-poisoning path: a rejected True Online TD update still holds the previous finite state, reports a zero TD-error diagnostic (`td_error` and `metrics[0:2]`), and sets `update_applied=False`. The tests now assert those returned fields, not only state finiteness. Please re-review that SHA.

Prabhat1308 (approved):
Independent review of head `ec93a2fc8803bfb53bb63823f55f12de267fb5e9` against `main` `da8b86c`, under the `review-asi-contributions` skill. Head tree verified as current `main` plus exactly `alberta_framework/core/learners.py` + `tests/test_true_online_td.py`; automated adversarial scan clean (no dependency, CI, hook, or lockfile changes; no assertions or raises removed anywhere in the fleet).

**Reproduced:** `tests/test_true_online_td.py` at head — **12 passed in 2.79s**; repo-wide Ruff: All checks passed!; strict mypy: Success: no issues found in 159 source files; `git diff --check`: clean. Red phase with `alberta_framework/core/learners.py` reverted to `main`, PR tests kept: **1 failed, 11 passed in 2.34s** (1 new regression test(s) fail), confirming failing-test-first.

**Design note (credit):** unlike most of this fleet, the hold here is *observable* — the result exposes an `update_applied` flag (and #86 deliberately keeps finite-input overflow visible, zeroing only non-finite input coordinates), so callers can distinguish a held update from a learned one. This is the pattern the silently-holding PRs in the fleet would benefit from adopting.

Recommendation: **accept**.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/83", "headSha": "ec93a2fc8803bfb53bb63823f55f12de267fb5e9", "recommendation": "accept", "reproduced": true, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": ["non-finite guard in alberta_framework/core/learners.py with failing-test-first regression test(s), reproduced red on main (1 failed, 11 passed in 2.34s) and green at head (12 passed in 2.79s)"], "commands": [".venv/bin/python -m pytest tests/test_true_online_td.py -q -o addopts=\"\"", ".venv/bin/python -m ruff check .", ".venv/bin/python -m mypy", "git diff --check"], "evidenceUrls": ["https://github.com/elizaOS/asi/pull/83"], "summary": "Head verified as main plus the stated two-file diff; red/green reproduced (1 regression test(s) red on main, 12 passed in 2.79s at head). Part of a 27-PR same-day non-finite-guard fleet, disclosed Cursor-authored, no receipts, no linked issues."}
```

PR #86, opened 2026-08-15
Hold normalizer statistics when inf observation is inf-inf

Kayvan-Zahiri:
## Summary
- EMA, Welford, and streaming BatchNorm all do `observation - mean`. Inf observation becomes an inf mean, then the next `inf - inf` is NaN variance/mean that never recovers.
- Fail-closed: if any coordinate is non-finite, skip the statistics update. `normalize_only` also emits zero for non-finite coordinates instead of NaN.
- Later finite samples keep learning from the previous finite mean and variance.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_normalizers.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/normalizers.py tests/test_normalizers.py`

### DIFF
--- a/alberta_framework/core/normalizers.py
+++ b/alberta_framework/core/normalizers.py
@@ -366,7 +366,15 @@ class Normalizer[
             Normalized observation
         """
         std = jnp.sqrt(state.var)
-        return (observation - state.mean) / (std + self._epsilon)
+        normalized = (observation - state.mean) / (std + self._epsilon)
+        # Inf observation minus an inf mean is inf-inf = NaN. Do not
+        # emit non-finite coordinates; keep zeros so downstream learners
+        # are not poisoned when the update is refused.
+        return jnp.where(
+            jnp.isfinite(observation) & jnp.isfinite(normalized),
+            normalized,
+            jnp.zeros_like(normalized),
+        )
 
     def update_only(
         self,
@@ -473,6 +481,8 @@ class EMANormalizer(Normalizer[EMANormalizerState]):
             Tuple of (normalized_observation, new_state)
         """
         status = self.counter_status(state)
+        observation_valid = jnp.all(jnp.isfinite(observation))
+        update_available = status.update_available & observation_valid
 
         def accepted(_: None) -> tuple[Array, EMANormalizerState]:
             new_words, ignored_capacity = _checked_lifetime_words_increment(
@@ -502,7 +512,7 @@ class EMANormalizer(Normalizer[EMANormalizerState]):
             return self.normalize_only(state, observation), state
 
         normalized, new_state = jax.lax.cond(
-            status.update_available,
+            update_available,
             accepted,
             refused,
             operand=None,
@@ -515,7 +525,7 @@ class EMANormalizer(Normalizer[EMANormalizerState]):
             counter_valid=status.counter_valid,
             lifetime_capacity_available=status.lifetime_capacity_available,
             estimator_capacity_available=status.estimator_capacity_available,
-            update_applied=status.update_available,
+            update_applied=update_available,
         )
 
 
@@ -583,6 +593,8 @@ class WelfordNormalizer(Normalizer[WelfordNormalizerState]):
             Tuple of (normalized_observation, new_state)
         """
         status = self.counter_status(state)
+        observation_valid = jnp.all(jnp.isfinite(observation))
+        update_available = status.update_available & observation_valid
 
         def accepted(_: None) -> tuple[Array, WelfordNormalizerState]:
             new_words, ignored_capacity = _checked_lifetime_words_increment(
@@ -612,7 +624,7 @@ class WelfordNormalizer(Normalizer[WelfordNormalizerState]):
             return self.normalize_only(state, observation), state
 
         normalized, new_state = jax.lax.cond(
-            status.update_available,
+            update_available,
             accepted,
             refused,
             operand=None,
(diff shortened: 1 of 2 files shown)


alberta_framework/core/normalizers.py
```diff
@@ -366,7 +366,15 @@ def normalize_only(
             Normalized observation
         """
         std = jnp.sqrt(state.var)
-        return (observation - state.mean) / (std + self._epsilon)
+        normalized = (observation - state.mean) / (std + self._epsilon)
+        # Inf observation minus an inf mean is inf-inf = NaN. Do not
+        # emit non-finite coordinates; keep zeros so downstream learners
+        # are not poisoned when the update is refused.
+        return jnp.where(
+            jnp.isfinite(observation) & jnp.isfinite(normalized),
```
0xcii: [P1] Do not erase genuine normalization failures

`jnp.isfinite(normalized)` is broader than the invalid-input case this PR intends to handle. With entirely finite float32 values (`observation=max`, `state.mean=-max`, positive finite variance), `observation - state.mean` overflows to `inf`; this condition silently rewrites it to zero. A non-finite result can likewise expose corrupt persisted mean/variance. Turning either into a valid-looking zero feature hides a numerical failure from downstream measurement. Gate zeroing only on a non-finite observation coordinate (or reject the whole sample), and preserve or explicitly reject a non-finite result produced from finite inputs. Please add that negative test.

0xcii (changes requested):
I independently reviewed exact head `c43ed600558bbd07383dccc75b13e7482651bb51`. Refusing a statistics update for a non-finite input is sound, but the new `normalize_only` sanitizer also hides non-finite results produced from entirely finite inputs/state values. That turns a real arithmetic failure into an apparently valid zero feature and weakens measurement integrity.

I reproduced the exact expression in a disposable, network-disabled, read-only JAX container (2 CPUs, 2 GiB RAM, 128 PIDs). With finite float32 `observation=max` and finite `mean=-max`, subtraction overflows; the patch rewrites that overflow to zero:

```json
{"finite_input_overflowed": true, "patched_silenced_overflow_to_zero": true, "input_only_preserves_overflow": true, "nonfinite_input_zeroed": true}
```

Please restrict the zero substitution to coordinates whose input observation is non-finite (or reject the sample fail-closed), while preserving/rejecting a genuine non-finite normalization result. Add a negative test covering finite-input arithmetic overflow or corrupt finite-shaped state.

Compute receipt: 1215029 project-attributed tokens (bounded; device-signed, locally reported)
Client / agent tooling: codex
— [normalizer-measurement-integrity]

Kayvan-Zahiri:
Addressed the blocking finding.

`normalize_only` now zeros a coordinate only when that observation entry is non-finite. Finite-input overflow (`max - (-max)`) stays non-finite. Added `test_finite_overflow_is_not_rewritten_to_zero` for EMA, Welford, and streaming BatchNorm.

alberta_framework/core/normalizers.py
Kayvan-Zahiri: Follow-up `a0c368c` gates zeroing on a non-finite observation coordinate only. Finite-input overflow (`max - (-max)`) stays non-finite. Negative coverage is `test_finite_overflow_is_not_rewritten_to_zero`. Please re-review that SHA.

Prabhat1308 (approved):
Independent review of head `a0c368c750500e0b9918d4423da56e9df5af363d` against `main` `da8b86c`, under the `review-asi-contributions` skill. Head tree verified as current `main` plus exactly `alberta_framework/core/normalizers.py` + `tests/test_normalizers.py`; automated adversarial scan clean (no dependency, CI, hook, or lockfile changes; no assertions or raises removed anywhere in the fleet).

**Reproduced:** `tests/test_normalizers.py` at head — **37 passed in 3.94s**; repo-wide Ruff: All checks passed!; strict mypy: Success: no issues found in 159 source files; `git diff --check`: clean. Red phase with `alberta_framework/core/normalizers.py` reverted to `main`, PR tests kept: **3 failed, 34 passed in 3.52s** (3 new regression test(s) fail), confirming failing-test-first.

**Design note (credit):** unlike most of this fleet, the hold here is *observable* — the result exposes an `update_applied` flag (and #86 deliberately keeps finite-input overflow visible, zeroing only non-finite input coordinates), so callers can distinguish a held update from a learned one. This is the pattern the silently-holding PRs in the fleet would benefit from adopting.

Recommendation: **accept**.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/86", "headSha": "a0c368c750500e0b9918d4423da56e9df5af363d", "recommendation": "accept", "reproduced": true, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": ["non-finite guard in alberta_framework/core/normalizers.py with failing-test-first regression test(s), reproduced red on main (3 failed, 34 passed in 3.52s) and green at head (37 passed in 3.94s)"], "commands": [".venv/bin/python -m pytest tests/test_normalizers.py -q -o addopts=\"\"", ".venv/bin/python -m ruff check .", ".venv/bin/python -m mypy", "git diff --check"], "evidenceUrls": ["https://github.com/elizaOS/asi/pull/86"], "summary": "Head verified as main plus the stated two-file diff; red/green reproduced (3 regression test(s) red on main, 37 passed in 3.94s at head). Part of a 27-PR same-day non-finite-guard fleet, disclosed Cursor-authored, no receipts, no linked issues."}
```

PR #91, opened 2026-08-15
Do not multiply V(s') when Horde discount is zero

Kayvan-Zahiri:
## Summary
- Horde TD targets are `cumulant + gamma * V(s')`. For `gamma=0` that product is `0 * inf = NaN` when the next prediction is inf, so a prediction demon is treated as inactive.
- Fail-closed: a zero discount does not multiply `V(s')`. The target is the cumulant, which is the same algebra as the finite path.
- `update_with_discounts` uses the same rule for a zero per-step discount.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_horde.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/horde.py tests/test_horde.py`

### DIFF
--- a/alberta_framework/core/horde.py
+++ b/alberta_framework/core/horde.py
@@ -368,7 +368,8 @@ class HordeLearner:
         # For gamma=0 demons: target = cumulant (single-step prediction)
         # NaN cumulants stay NaN (inactive demons)
         gammas = self._horde_spec.gammas
-        targets = cumulants + gammas * next_preds
+        bootstrap = jnp.where(gammas == 0.0, 0.0, gammas * next_preds)
+        targets = cumulants + bootstrap
 
         # 3. Delegate to MultiHeadMLPLearner
         result = self._learner.update(state, observation, targets)
@@ -412,7 +413,8 @@ class HordeLearner:
         """
         next_preds = self._learner.predict(state, next_observation)
         discounts = jnp.asarray(discounts, dtype=jnp.float32)
-        targets = cumulants + discounts * next_preds
+        bootstrap = jnp.where(discounts == 0.0, 0.0, discounts * next_preds)
+        targets = cumulants + bootstrap
         result = self._learner.update(state, observation, targets)
 
         return HordeUpdateResult(  # type: ignore[call-arg]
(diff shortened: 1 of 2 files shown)


Prabhat1308 (approved):
Independent review of head `65c4f25666aa006e932afe0175c7b6f694910010` against `main` `da8b86c`, under the `review-asi-contributions` skill. Head tree verified as current `main` plus exactly `alberta_framework/core/horde.py` + `tests/test_horde.py`; automated adversarial scan clean (no dependency, CI, hook, or lockfile changes; no assertions or raises removed anywhere in the fleet).

**Reproduced:** `tests/test_horde.py` at head — **22 passed in 7.46s**; repo-wide Ruff: All checks passed!; strict mypy: Success: no issues found in 159 source files; `git diff --check`: clean. Red phase with `alberta_framework/core/horde.py` reverted to `main`, PR tests kept: **1 failed, 21 passed in 7.05s** (1 new regression test(s) fail), confirming failing-test-first.

**Scope note:** the guard is algebra-preserving — for every finite path and every nonzero discount the computed values are unchanged (both `where` branches equal); only the undefined 0*inf product changes, to the value the zero-discount algebra already implies.

Recommendation: **accept**.

```slop-review
{"schemaVersion": "1", "projectId": "asi", "artifactUrl": "https://github.com/elizaOS/asi/pull/91", "headSha": "65c4f25666aa006e932afe0175c7b6f694910010", "recommendation": "accept", "reproduced": true, "securityRisk": "none", "duplicateRisk": "none", "usefulArtifacts": ["non-finite guard in alberta_framework/core/horde.py with failing-test-first regression test(s), reproduced red on main (1 failed, 21 passed in 7.05s) and green at head (22 passed in 7.46s)"], "commands": [".venv/bin/python -m pytest tests/test_horde.py -q -o addopts=\"\"", ".venv/bin/python -m ruff check .", ".venv/bin/python -m mypy", "git diff --check"], "evidenceUrls": ["https://github.com/elizaOS/asi/pull/91"], "summary": "Head verified as main plus the stated two-file diff; red/green reproduced (1 regression test(s) red on main, 22 passed in 7.46s at head). Part of a 27-PR same-day non-finite-guard fleet, disclosed Cursor-authored, no receipts, no linked issues."}
```

PR #97, opened 2026-08-15
fix(screening): refuse to merge shards with mismatched hyperparameters

### LINKED ISSUE #96: merge_shards silently merges shards with mismatched hyperparameters under one config_name
## Problem

`merge_shards` doesn't check that shards sharing a `config_name` actually ran with the same `hyperparameters`. if a registry value gets tuned between two `run` invocations for the same arm (before it's frozen, or a hotfix lands mid-wave) and both old and new shards land in the same merge directory, they merge silently: per-task accuracy from genuinely different mechanisms gets averaged into one entry, and the summary reports only `seeds[0]`'s hyperparameters as if it's representative of the whole arm.

same class of silent-ambiguity issue #44/#45/#49/#50/#53 fixed this week, just a different corner of the same function.

## Reproduction (synthetic in-band shards, current `main` `da8b86c`)

```text
seed0 hyperparameters: {'step_size': 0.01, 'utility_decay': 0.9999, 'noise_std': 0.1, 'weight_decay': 0.01}
seed1 hyperparameters (mutated): {..., 'meta_step_size': 999.0}

merge_shards([control, seed0, seed1], control_name='upgd_w_control')
=> merge succeeded, no error raised
=> reported hyperparameters (from seed 0 only): {'step_size': 0.01, ...}
=> seeds merged together: [0, 1]
```

two `upgd_l2init` shards with different hyperparameters dicts merged into one `paired_vs_control` entry with zero warning. `load_shard`'s structural validation doesn't check `hyperparameters` against the current registry or against sibling shards either.

## Proposed fix

`merge_shards` additionally requires every seed under one `config_name` to carry identical `hyperparameters`, raising `ValueError` naming the config and the mismatched seed(s). everything else (the merge, the paired-vs-control block, every other check) stays untouched, this only closes the one gap.

not proposing a fix to `load_shard` here, it has no registry access at load time to compare against. flagging in case maintainers would rather cover that boundary instead or in addition.

Svector-anu:
Closes #96.

## What changed

`merge_shards` now requires every seed merged under one `config_name` to carry identical `hyperparameters`, raising a deterministic `ValueError` naming the config and the mismatched seed(s) when they differ, instead of quietly taking `seeds[0]`'s hyperparameters as representative of the whole arm and averaging per-task accuracy across what could be genuinely different mechanisms.

reproduced the gap directly against the merged code before writing the fix: two `upgd_l2init` shards (seeds 0 and 1) with one seed's `hyperparameters` mutated (`meta_step_size` changed) merged into a single `paired_vs_control` entry with zero warning, reporting only seed 0's values. same failure mode as the #44/#45/#49/#50/#53 family, a merge that should refuse instead produces a schema-valid summary that reads as one coherent measurement.

out of scope (see #96): `load_shard`'s structural validation doesn't have registry access at load time to compare a shard's hyperparameters against the *current* registry entry, this PR only closes the merge-time cross-seed consistency gap, not that separate load-time boundary.

## Verification

failing-test-first in `TestShardsAndMerge::test_merge_rejects_hyperparameter_drift_across_seeds`. confirmed red before the guard (temporarily reverted just the source fix, kept the test, mutation was silently accepted like before): `DID NOT RAISE ValueError`. restored the fix, green again.

```text
204 passed in 81.49s

All checks passed!

$ .venv/bin/python -m mypy
Success: no issues found in 159 source files
```

lane: deterministic unit tests, non-promoting. no seeds consumed, no benchmark runs, no `outputs/` artifacts touched.

## Evidence

- [x] logs: full focused-suite, ruff, and mypy transcript pasted above (204/204 passed, lint and types clean on both changed files).

- [x] domain-artifact: the mutation-resistance transcript above (red before the fix, green after) is the artifact, a deterministic unit-test lane doesn't produce an `outputs/` shard or summary.

## Provenance

No Slop run receipt: this session runs as Sonnet 5, not the approved `claude-fable-5` identifier for Claude Code measured runs, so `run-receipt.mjs start/finish` wasn't invoked. the fix, reproduction, and tests above are real and independently verified, only the receipt-capture pipeline was skipped.

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -7046,6 +7046,21 @@ def merge_shards(
     entries: list[dict[str, Any]] = []
     for name, per_seed in sorted(by_config.items()):
         seeds = sorted(per_seed)
+        reference_hp = per_seed[seeds[0]]["hyperparameters"]
+        mismatched = [s for s in seeds if per_seed[s]["hyperparameters"] != reference_hp]
+        if mismatched:
+            # A registry entry is not literally immutable while a config is
+            # still being tuned; shards from before and after a hyperparameter
+            # change can land in the same directory. Without this check they
+            # merge silently under one config_name, average per-task accuracy
+            # across genuinely different mechanisms, and report only
+            # seeds[0]'s hyperparameters as if representative of the whole arm.
+            raise ValueError(
+                f"config {name!r} has inconsistent hyperparameters across seeds: "
+                f"seed {seeds[0]} used {reference_hp!r}, seed(s) {mismatched} used "
+                "different values; refusing to merge runs of different mechanisms "
+                "under one config_name"
+            )
         acc = np.stack(
             [np.asarray(per_seed[s]["per_task_accuracy"], dtype=np.float64) for s in seeds]
         )
--- a/tests/test_ipmnist_screening.py
+++ b/tests/test_ipmnist_screening.py
@@ -988,6 +988,29 @@ class TestShardsAndMerge:
         with pytest.raises(ValueError, match="duplicate shard"):
             merge_shards([p1, p2])
 
+    def test_merge_rejects_hyperparameter_drift_across_seeds(self, tmp_path, small_data):
+        """Two shards under one config_name with different hyperparameters must
+        not merge: nothing else catches a registry value that changed between
+        two `run` invocations, and a silent merge would average per-task
+        accuracy across genuinely different mechanisms while reporting only
+        the first seed's hyperparameters as if representative of the arm."""
+        p0 = self._make_shard(tmp_path, small_data, "upgd_w_control", 0)
+        p1 = self._make_shard(tmp_path, small_data, "upgd_l2init", 0)
+        p2 = self._make_shard(tmp_path, small_data, "upgd_l2init", 1)
+
+        payload2 = json.loads(p2.read_text(encoding="utf-8"))
+        payload2["hyperparameters"] = {
+            **payload2["hyperparameters"],
+            "meta_step_size": 999.0,
+        }
+        p2.write_text(json.dumps(payload2), encoding="utf-8")
+
+        with pytest.raises(
+            ValueError,
+            match=r"'upgd_l2init' has inconsistent hyperparameters across seeds",
+        ):
+            merge_shards([p0, p1, p2], control_name="upgd_w_control", slope_window=2)
+
     def test_validate_proxy_prefix_and_ordering(self, tmp_path, small_data):
         x, y = small_data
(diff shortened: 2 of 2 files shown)


Prabhat1308 (approved):
Independent review of head `19e1e8d95a76cb72a40c88e5681c7f964bc20316` against `main` `da8b86c`, under the `review-asi-contributions` skill. Head tree verified as current `main` plus exactly the stated two-file diff before any execution.

**Reproduced:** full `tests/test_ipmnist_screening.py` at head — `204 passed`; repo-wide Ruff clean; strict mypy clean (159 source files); `git diff --check` clean. Red phase: with `ipmnist_screening.py` reverted to `main` and the PR's test kept, `test_merge_rejects_hyperparameter_drift_across_seeds` fails exactly as claimed (merge silently accepted the mutated shard).

**Correctness:** the guard sits inside the per-`config_name` loop, refuses on any seed whose `hyperparameters` dict differs from `seeds[0]`'s, and names config and mismatched seeds deterministically. This closes a real corruption path — previously the merge averaged per-task accuracy across genuinely different mechanisms while publishing only `seeds[0]`'s hyperparameters as representative. Same family as the merged #45/#48/#50 validators; the load-time registry boundary is correctly scoped out per #96. Dict equality is exact-match, which is right for registry-produced shards. No validator weakened, no `outputs/` touched, no benchmark seed consumed, no adversarial indicators in body or diff.

**Merge-planning note (non-blocking):** open PR #110 appends tests to the same `TestShardsAndMerge` region — `git merge-tree` shows the two heads conflict in `tests/test_ipmnist_screening.py` (source hunks compose cleanly). Whichever lands second needs a small test-file rebase.

Recommendation: **accept**.

```slop-review
{"schemaVersion":"1","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/97","headSha":"19e1e8d95a76cb72a40c88e5681c7f964bc20316","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","usefulArtifacts":["cross-seed hyperparameter-drift refusal in merge_shards plus failing-test-first regression test, reproduced red on main and green at head (204 passed)"],"commands":[".venv/bin/python -m pytest tests/test_ipmnist_screening.py -q -o addopts=\\"\\"",".venv/bin/python -m ruff check .",".venv/bin/python -m mypy","git diff --check","git merge-tree --write-tree --messages 19e1e8d 991cb2f"],"evidenceUrls":["https://github.com/elizaOS/asi/pull/97","https://github.com/elizaOS/asi/issues/96"],"summary":"Head tree verified as main da8b86c plus the stated diff; all claimed results reproduced including the red phase; the guard closes a real silent-merge corruption path in the active screening lane; only note is a test-file merge conflict with open PR #110."}
```

Issue #51, opened 2026-08-14
Preregistration: within-runner replication of the RLS-residual champion pair on the settled source


swissyai:
# Pre-registration: within-runner replication of the champion pair on the settled source

## Scope and rationale

- Lane: `alberta_framework.benchmarks.ipmnist_screening`; metric `average_online_accuracy` on the 60-task x 5,000-step development screen, `noise_mode=step`.
- Evidence class: `development_screening_diagnostic`, permanently nonpromoting. Seeds 0-2 are exposed (2026-08-03/04 waves) and are used here as explicitly consumed development seeds; no held-out seed is touched.
- Why this, why now: the release-triage ruling in #14 states the repository contract requires freezing the post-cleanup source/environment and remeasuring the selected control in the same runner before any A/B — which makes every pre-prune baseline stale for comparison purposes. `docs/research/ipmnist-theory.md` §6 lists source-bound replication of the RLS-residual gain as open hypothesis (1). This preregistration is that remeasurement: it re-establishes the champion-pair baseline that every future candidate on this lane (including the #14 replacement) must be paired against, in one runner, on the settled source.

## Incumbent numbers (read fresh from the artifacts at posting time)

From `outputs/ipmnist_screening/summary_rls_head.json` (60-task screen, seeds 0-2):

- `rls_head_resid_l1_preset005`: 0.8693777581 ± 0.0001978244, per-seed `[0.869137, 0.86977, 0.869227]`
- `sigma0_shiftnorm_d099`: 0.8639621995 ± 0.0002044436, per-seed `[0.863553, 0.864167, 0.864167]`
- Champion-minus-body per-seed paired diffs: `[+0.005584, +0.005603, +0.005060]`; mean **+0.005416**, stderr 0.000178 (n=3).
- Context only: the 200-task n=20 confirmation records paired +0.0066531 ± 0.0001299, all 20 seeds positive (`summary_rls_head_confirm.json`).

## The one change

No mechanism and no code: both arms are registered and the runner is the standard CLI. What is being replicated over is the **settled source** (current `main` head at launch, recorded in every shard's provenance) and an **independent runner** (macOS arm64 CPU, Python 3.12.12, jax 0.11.0 from the stock `uv.lock`). Absolute levels are expected to shift cross-runner — `AUDIT.md` documents 1-2-ulp XLA divergence chaos-amplified to per-task jitter — so the endpoint is the within-runner **paired diff**, the comparison unit this campaign's methodology stands on.

[...]

PR #110, opened 2026-08-15
fix(screening): refuse to merge shards from mismatched environments

### LINKED ISSUE #108: merge_shards records each shard's environment but never validates or carries it into the summary
## Problem

`shard_payload` (`ipmnist_screening.py:6958`) records each shard's runtime `environment` — jax, numpy, python, and platform versions:

```python
"environment": {
    "jax": jax.__version__,
    "numpy": np.__version__,
    "python": platform.python_version(),
    "platform": platform.platform(),
},
```

(lines 6974-6979) — but `merge_shards` (line 7011) and `validate_proxy` (line 7137) never read the key. Grepping the file for the literal `"environment"` returns exactly the one write site: it is never compared across shards, never gated on, and never carried into the published summary or validation report at all.

`run_screening_config`'s own docstring claims the control arm reproduces the full-horizon lane "bit-for-bit," and closed issue #46 already documents a real, measured divergence from exactly this axis on this codebase: under jax 0.11.0 (in-range per `pyproject.toml`'s `jax>=0.7.1` floor, but different from the pin's authoring environment), a derived metric reassociates by ~1.4e-8 despite bitwise-identical accuracy/loss trajectories. Open issue #51 shows the project already treats cross-environment contamination as a live hazard for this same lane — its entire protocol exists to pin "one runner, ... the settled source" before any A/B comparison — but that discipline is enforced by a human pre-registration document, not by `merge_shards` itself. Two shards produced on different jax/numpy/platform combinations for the same `config_name` merge and pair against the control with no flag anywhere in the artifact.

Same defect family as #44/#45, #49/#50, #53, #96/#97: a field that should gate the merge (or at minimum ride along in the published summary) is silently dropped instead.

Two `upgd_w_control` shards, `SMALL` config, seeds {0,1}; seed 0 is a real `run_screening_config` output on this machine (jax 0.11.0 / macOS arm64), seed 1's `environment` is overwritten to a different jax/numpy/python/platform combination:

```text
seed=0 environment={'jax': '0.11.0', 'numpy': '2.5.2', 'python': '3.12.14', 'platform': 'macOS-15.7.7-arm64-arm-64bit'}
seed=1 environment={'jax': '0.4.20', 'numpy': '1.26.4', 'python': '3.11.9', 'platform': 'Linux-5.15-x86_64'}

merge_shards([seed0_path, seed1_path], control_name="upgd_w_control", slope_window=2)
=> merge_shards raised NO error
summary top-level keys: ['confirmation_threshold', 'control_name', 'created_unix', 'evidence_policy',
                          'n_shards', 'noise_mode', 'protocol_config', 'results', 'schema', 'slope_window']
'environment' is not among them: True
merged entry n_seeds: 2  seeds: [0, 1]
```

`merge_shards` additionally requires every shard in the merge to carry identical `environment`, raising `ValueError` naming the mismatched field(s) and shard(s) — the same shape as the `config`/`noise_mode` consistency checks it already runs (lines 7020-7029). If maintainers judge a hard refusal too strict for a legitimate heterogeneous-fleet campaign, the fallback is to carry `environment` (or its distinct-value set) forward into the published summary so a reader can see the drift instead of it vanishing silently — but a silent drop is not an acceptable third option given #46's precedent. Failing-test-first: pin the reproduction above as a red test before the guard, green after.

Not touching in this change: `validate_proxy`'s separate exact-prefix comparison against full-horizon partials, or the `#46`-class question of which jax-version pairs are numerically compatible enough to merge — this only closes the gap where a mismatch is invisible, not which mismatches should be permitted.

Searched `gh issue list --repo elizaOS/asi` and `gh pr list --repo elizaOS/asi --search`, both open and closed, for: `"environment jax version"` and `"merge_shards environment"` — the only hits were #46 (a specific bitwise-pin test failure under jax 0.11.0, not about `merge_shards`'s field checking) and #51 (a process-level preregistration protocol, not a code fix). Neither proposes or implements a `merge_shards` environment check. Reviewed the titles of all 6 open issues and all closed issues in the tracker — none reference this gap.

## Provenance

— [nxn-claude-code]

nothingxnowhere:
Closes #108.

## What changed

`shard_payload` (`ipmnist_screening.py:6958`) records each shard's runtime `environment` (jax, numpy, python, platform), but `merge_shards` (`ipmnist_screening.py:7011`) never read the field. `merge_shards` now additionally requires every shard in a merge to carry an identical `environment`, raising a deterministic `ValueError` naming the differing key(s) and the shard(s) that diverge, placed adjacent to the existing `config`/`noise_mode` consistency checks (global, before the per-`config_name` loop) rather than folded into any per-arm logic. #46 already documents a real, measured divergence on exactly this axis for this codebase — a derived metric reassociating by ~1.4e-8 under jax 0.11.0 despite bitwise-identical accuracy/loss trajectories — and #51's entire pre-registration protocol exists because cross-environment contamination is a live hazard here; neither guard was previously enforced by `merge_shards` itself. The issue's proposed fix offered "refuse" as primary and "carry `environment` forward into the summary instead" as a fallback if maintainers judge a hard refusal too strict; this PR does both — refuse on drift, and (since the guard now enforces uniformity) carry the single agreed `environment` value forward into the published summary's top level, so a reader never has to re-derive which runner produced a merge.

Before, on `main` `da8b86c` (`upgd_w_control`, `SMALL` config, seeds {0,1}; seed 0 real, seed 1's `environment` overwritten to a different jax/numpy/python/platform combination):

```text
merge_shards([seed0_path, seed1_path], control_name="upgd_w_control", slope_window=2)
=> merge_shards raised NO error
summary top-level keys: ['confirmation_threshold', 'control_name', 'created_unix', 'evidence_policy',
                          'n_shards', 'noise_mode', 'protocol_config', 'results', 'schema', 'slope_window']
'environment' is not among them: True
```

After:

```text
ValueError: shards were produced under different environments (jax: ['upgd_w_control seed=1'];
numpy: ['upgd_w_control seed=1']; platform: ['upgd_w_control seed=1']; python: ['upgd_w_control
seed=1']); refusing to merge runs from mismatched jax/numpy/python/platform combinations
```

and, for a genuinely consistent merge, `summary["environment"]` now equals every shard's shared `environment` dict.

Failing-test-first in `TestShardsAndMerge` (`tests/test_ipmnist_screening.py`): `test_merge_rejects_environment_drift_across_shards` builds two real shards via `_make_shard` (the same helper every other test in this class uses, wrapping `run_screening_config`/`shard_payload`), then overwrites seed 1's `environment` before merging — red on `main` (`Failed: DID NOT RAISE ValueError`), green at head. `test_merge_carries_consistent_environment_into_summary` is the positive control: two genuinely matching shards still merge, and asserts `summary["environment"]` equals the shared value (red on `main` with `KeyError: 'environment'` — the key was simply absent — green at head).

Anti-collision with open PR #97: #97 adds a `hyperparameters` guard inside the per-`config_name` loop (`ipmnist_screening.py:7047+`, `for name, per_seed in sorted(by_config.items()):`). This change is entirely outside that loop — it lives with the other merge-global checks (`configs`, `noise_modes`) before `by_config` is even built, because `environment` is a property of the merge as a whole, not of any one arm. The two diffs touch disjoint line ranges in `merge_shards` and should apply and compose cleanly in either order.

Schema: `SUMMARY_SCHEMA` (`"alberta.ipmnist_screening.summary.v1"`) is not bumped. Checked precedent exactly: `git log --all -S "SUMMARY_SCHEMA"` shows the constant was set once, in the commit that introduced it, and neither #45 nor #50 (both landed new refusal behavior in this same function) touched it. This PR follows the same precedent and adds `environment` to the summary without a version bump.

Not touching in this change (per #108's own scope): `validate_proxy`'s separate exact-prefix comparison against full-horizon partials, and the `#46`-class question of which jax-version pairs are numerically compatible enough to merge — this closes the gap where a mismatch is invisible, not which mismatches should eventually be permitted.

## Evidence

Lane: deterministic unit tests, non-promoting; no seeds consumed beyond the existing `small_data`/`SMALL`-config fixtures this test class already uses, no benchmark campaign runs, no `outputs/` artifacts touched. Environment: macOS arm64, CPU backend, Python 3.12.14, jax 0.11.0, numpy 2.5.2 (stock `uv.lock`). Commit measured: `b9cb09655cf9125c231b735e5bdc5714529cfe4c`.

```text
$ .venv/bin/python -m pytest tests/test_ipmnist_screening.py -k "environment_drift or carries_consistent_environment" -q -o addopts=""
FF                                                                       [100%]
____ TestShardsAndMerge.test_merge_rejects_environment_drift_across_shards _____
...
tests/test_ipmnist_screening.py:1009: Failed
__ TestShardsAndMerge.test_merge_carries_consistent_environment_into_summary ___
...
>       assert summary["environment"] == expected_environment
               ^^^^^^^^^^^^^^^^^^^^^^
E       KeyError: 'environment'
tests/test_ipmnist_screening.py:1027: KeyError
FAILED tests/test_ipmnist_screening.py::TestShardsAndMerge::test_merge_rejects_environment_drift_across_shards
FAILED tests/test_ipmnist_screening.py::TestShardsAndMerge::test_merge_carries_consistent_environment_into_summary
2 failed, 203 deselected in 5.25s
```

Candidate, targeted then full touched file:

```text
$ .venv/bin/python -m pytest tests/test_ipmnist_screening.py -k "environment_drift or carries_consistent_environment" -q -o addopts=""
..                                                                       [100%]
2 passed, 203 deselected in 4.36s

........................................................................ [ 35%]
........................................................................ [ 70%]
.............................................................            [100%]
205 passed in 72.33s (0:01:12)
# all 203 pre-existing tests plus the 2 new ones pass; no #46-class pin
# failure observed in this environment at this head

All checks passed!

$ .venv/bin/python -m mypy alberta_framework/benchmarks/ipmnist_screening.py
```

Exit codes: pytest 0, ruff 0, mypy 0.

Historical safety, verified against the artifacts: a scan of all 459 pinned raw screening shard files under `outputs/ipmnist_screening/**/*_seed*.json` (133 distinct `(directory, config_name)` groups, spanning `confirm_full`, `confirm_rls_head`, and the flat top-level directory) found zero groups with more than one distinct `environment` value — this guard would not have refused any committed historical merge. The 8 pinned `summary*.json` files could not be re-merged bit-for-bit against this check (their exact original shard subsets are not fully reconstructable from file names alone — several config names referenced in those summaries, e.g. `sigma0_ndecay099`, have no matching `*_seed*.json` file left in this checkout), so this is a field-level scan over what is actually committed, not a re-run of every historical merge call.

## Open / caveats

- Refuse-and-carry-forward (rather than refuse-only, or carry-forward-only) was the design fork; the issue names both as acceptable, with refuse as primary. Doing both means a legitimate heterogeneous-fleet campaign still gets an explicit, actionable error instead of a silent artifact, while every merge that *does* pass now leaves the artifact self-describing for readers who don't re-derive the runner from CI logs.
- Out of scope, still open per #108: whether specific jax-version pairs (the #46 class of question) should ever be treated as compatible enough to merge despite this guard — that judgment call is left to a future change, not made here.

Deviations from the #108 plan: none.

## Provenance

— [nxn-claude-code]

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -7027,6 +7027,31 @@ def merge_shards(
             "approximation); merge them separately"
         )
     noise_mode = noise_modes.pop()
+    reference_environment = shards[0]["environment"]
+    mismatched_environment: dict[str, set[str]] = {}
+    for shard in shards[1:]:
+        for key, value in shard["environment"].items():
+            if value != reference_environment.get(key):
+                shard_id = f"{shard['config_name']} seed={shard['seed']}"
+                mismatched_environment.setdefault(key, set()).add(shard_id)
+    if mismatched_environment:
+        # jax/numpy/python/platform drift across shards for one merge is a
+        # real, measured hazard on this codebase (#46): a derived metric can
+        # reassociate by ~1e-8 despite bitwise-identical accuracy/loss
+        # trajectories. shard_payload records `environment` per shard but
+        # nothing previously compared it, so two shards from different
+        # runners merged and paired against the control with no flag
+        # anywhere in the artifact.
+        detail = "; ".join(
+            f"{key}: {sorted(shard_ids)}"
+            for key, shard_ids in sorted(mismatched_environment.items())
+        )
+        raise ValueError(
+            f"shards were produced under different environments ({detail}); "
+            "refusing to merge runs from mismatched jax/numpy/python/platform "
+            "combinations"
+        )
+    environment = dict(reference_environment)
     by_config: dict[str, dict[int, dict[str, Any]]] = {}
     for shard in shards:
         per_seed = by_config.setdefault(shard["config_name"], {})
@@ -7126,6 +7151,7 @@ def merge_shards(
         "created_unix": time.time(),
         "protocol_config": dict(shards[0]["config"]),
         "noise_mode": noise_mode,
+        "environment": environment,
         "control_name": control_name,
         "confirmation_threshold": CONFIRMATION_THRESHOLD,
         "slope_window": slope_window,
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -7028,11 +7028,18 @@ def merge_shards(
         )
     noise_mode = noise_modes.pop()
     reference_environment = shards[0]["environment"]
+    reference_keys = set(reference_environment)
     mismatched_environment: dict[str, set[str]] = {}
     for shard in shards[1:]:
+        shard_id = f"{shard['config_name']} seed={shard['seed']}"
+        # Key sets must match symmetrically: a shard *missing* a key the
+        # reference carries (or carrying an extra one) is drift too, and an
+        # asymmetric values-only scan would silently inherit the reference's
+        # value for it in the published summary.
+        for key in reference_keys.symmetric_difference(shard["environment"]):
+            mismatched_environment.setdefault(key, set()).add(shard_id)
         for key, value in shard["environment"].items():
-            if value != reference_environment.get(key):
-                shard_id = f"{shard['config_name']} seed={shard['seed']}"
+            if key in reference_keys and value != reference_environment[key]:
                 mismatched_environment.setdefault(key, set()).add(shard_id)
     if mismatched_environment:
         # jax/numpy/python/platform drift across shards for one merge is a
--- a/tests/test_ipmnist_screening.py
+++ b/tests/test_ipmnist_screening.py
@@ -988,6 +988,44 @@ class TestShardsAndMerge:
         with pytest.raises(ValueError, match="duplicate shard"):
             merge_shards([p1, p2])
 
+    def test_merge_rejects_environment_drift_across_shards(self, tmp_path, small_data):
+        """Shards recorded under different jax/numpy/python/platform
+        combinations must not merge silently: #46 already documents a real,
+        measured divergence on exactly this axis for this codebase, and
+        nothing else in merge_shards reads or compares the `environment`
+        field shard_payload writes."""
+        p0 = self._make_shard(tmp_path, small_data, "upgd_w_control", 0)
+        p1 = self._make_shard(tmp_path, small_data, "upgd_w_control", 1)
+
+        payload1 = json.loads(p1.read_text(encoding="utf-8"))
+        payload1["environment"] = {
+            "jax": "0.4.20",
+            "numpy": "1.26.4",
+            "python": "3.11.9",
+            "platform": "Linux-5.15-x86_64",
+        }
+        p1.write_text(json.dumps(payload1), encoding="utf-8")
+
+        with pytest.raises(
+            ValueError,
+            match=r"shards were produced under different environments",
+        ):
+            merge_shards([p0, p1], control_name="upgd_w_control", slope_window=2)
+
+    def test_merge_carries_consistent_environment_into_summary(
+        self, tmp_path, small_data
+    ):
+        """Positive control: shards that genuinely share one environment
+        still merge, and the published summary carries that environment
+        forward at the top level so the artifact is self-describing."""
+        p0 = self._make_shard(tmp_path, small_data, "upgd_w_control", 0)
+        p1 = self._make_shard(tmp_path, small_data, "upgd_w_control", 1)
+        expected_environment = json.loads(p0.read_text(encoding="utf-8"))["environment"]
+
+        summary = merge_shards([p0, p1], control_name="upgd_w_control", slope_window=2)
+
+        assert summary["environment"] == expected_environment
+
     def test_validate_proxy_prefix_and_ordering(self, tmp_path, small_data):
         x, y = small_data
         partials = tmp_path / "partials"
--- a/tests/test_ipmnist_screening.py
+++ b/tests/test_ipmnist_screening.py
@@ -1012,6 +1012,24 @@ class TestShardsAndMerge:
         ):
             merge_shards([p0, p1], control_name="upgd_w_control", slope_window=2)
 
+    def test_merge_rejects_missing_environment_key(self, tmp_path, small_data):
+        """A shard missing a key the reference environment carries must be
+        refused, not silently absorbed: an asymmetric values-only comparison
+        would let it pass and inherit the reference's value for that key in
+        the published summary."""
+        p0 = self._make_shard(tmp_path, small_data, "upgd_w_control", 0)
+        p1 = self._make_shard(tmp_path, small_data, "upgd_w_control", 1)
+
+        payload1 = json.loads(p1.read_text(encoding="utf-8"))
+        del payload1["environment"]["platform"]
+        p1.write_text(json.dumps(payload1), encoding="utf-8")
+
+        with pytest.raises(
+            ValueError,
+            match=r"shards were produced under different environments",
+        ):
+            merge_shards([p0, p1], control_name="upgd_w_control", slope_window=2)
+
     def test_merge_carries_consistent_environment_into_summary(
         self, tmp_path, small_data
     ):


Prabhat1308 (approved):
Independent review of head `991cb2f5f13067443d52b1f875eac50010c3ea9a` against `main` `da8b86c`, under the `review-asi-contributions` skill. Head tree verified as current `main` plus exactly the stated two-file diff before any execution.

**Reproduced:** full `tests/test_ipmnist_screening.py` at head — `205 passed`; repo-wide Ruff clean; strict mypy clean (159 source files); `git diff --check` clean. Red phase with the source reverted to `main`: `test_merge_rejects_environment_drift_across_shards` fails (merge accepted the foreign-environment shard silently) and `test_merge_carries_consistent_environment_into_summary` fails (`environment` absent from the summary), both exactly as claimed.

**Correctness:** the guard is merge-global, placed with the existing `config`/`noise_mode` checks before `by_config` is built — the right altitude, since environment is a property of the whole merge. The motivation is unusually well-grounded: #46 documents a real ~1.4e-8 cross-environment reassociation on this exact codebase, and every within-runner pairing rule exists because of this hazard. Carrying the verified-uniform `environment` into the summary is a strict information gain. No validator weakened, no `outputs/` touched, no adversarial indicators.

Two non-blocking notes:

1. **Key-set asymmetry:** the drift loop iterates each shard's `environment` keys against the reference — a shard *missing* a key the reference has (e.g., a hand-edited or older shard without `platform`) passes undetected and inherits the reference's value in the summary. Unreachable for `shard_payload`-produced shards, which always write all four keys; worth a symmetric key-set check if hardening further.
2. **Schema note:** the summary gains a top-level key while `SUMMARY_SCHEMA` stays unchanged, so same-schema-string summaries now vary in shape across time. Nothing downstream broke (205 passed), but the maintainer may want a schema bump or a changelog note. Also, contrary to the body's compose claim, `git merge-tree` shows this head conflicts with open PR #97 in `tests/test_ipmnist_screening.py` (both append to the same class; source hunks do compose cleanly) — one test-file rebase for whichever lands second.

Recommendation: **accept**.

```slop-review
{"schemaVersion":"1","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/110","headSha":"991cb2f5f13067443d52b1f875eac50010c3ea9a","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","usefulArtifacts":["merge-global environment-consistency refusal in merge_shards plus summary environment carry-forward, with two failing-test-first regression tests reproduced red on main and green at head (205 passed)"],"commands":[".venv/bin/python -m pytest tests/test_ipmnist_screening.py -q -o addopts=\\"\\"",".venv/bin/python -m ruff check .",".venv/bin/python -m mypy","git diff --check","git merge-tree --write-tree --messages 19e1e8d 991cb2f"],"evidenceUrls":["https://github.com/elizaOS/asi/pull/110","https://github.com/elizaOS/asi/issues/108","https://github.com/elizaOS/asi/issues/46"],"summary":"Head tree verified as main plus the stated diff; all claims reproduced including both red phases; guard correctly placed at merge-global altitude and grounded in the measured #46 divergence; notes: asymmetric key comparison for shards missing environment keys, unchanged SUMMARY_SCHEMA string despite the new summary key, and a test-file merge conflict with PR #97 contrary to the body's compose claim."}
```

nothingxnowhere:
Both notes actioned on new head `b9cb09655cf9125c231b735e5bdc5714529cfe4c`:

**1. Key-set asymmetry — closed.** The drift check now compares key sets symmetrically before comparing values: a shard missing a key the reference carries (or carrying an extra one) is refused with the same environments error, instead of silently inheriting the reference's value in the summary. Failing-test-first: `test_merge_rejects_missing_environment_key` (deletes `platform` from seed 1's environment) is red with the guard reverted (`1 failed — DID NOT RAISE`), green at head; full file `206 passed`; ruff + strict mypy clean.

**2. Compose claim — corrected, you're right.** My body claimed the diffs "compose cleanly in either order" with #97; your `git merge-tree` is accurate — the *source* hunks compose, but `tests/test_ipmnist_screening.py` has a mechanical adjacency conflict (both PRs append to the same class), needing one keep-both resolution for whichever lands second. The body's anti-collision section overstated; the restamped body stands corrected by this comment.

The `SUMMARY_SCHEMA` question (bump vs additive for the new top-level key) is left to the maintainer's preference — happy to add a bump on request.

Attribution status: self-reported (device-signed run receipt unavailable: slop.cash manifest was stale at original work time and has since refreshed — receipt onboarding queued; score derives from accepted outcomes, not receipts)
— [nxn-claude-code]

PR #109, opened 2026-08-15
fix(micro): refuse to merge cross-seed hyperparameter/mechanism drift

nothingxnowhere:
Closes #106.

## What changed

`merge_micro_shards` (`alberta_framework/benchmarks/micro_continual.py:883`) now requires every seed merged under one `arm_name` to carry identical `hyperparameters` and identical `mechanism`, raising a deterministic `ValueError` naming the arm, the field, and the mismatched seed(s) instead of silently reporting only `seeds[0]`'s values as representative of the whole arm while `curves.mean(axis=1)` blends per-regime accuracy across every seed regardless of mechanism. This ports #97's `ipmnist_screening.py::merge_shards` guard to this sibling merge, which had never had it, adjusted for this file's field names (`arm_name` instead of `config_name`) and for the extra `mechanism` field this merge also silently discarded. Two separate checks are raised (hyperparameters, then mechanism) so each mismatch is named precisely rather than folded into one ambiguous message. No behavior change for any merge whose seeds already agree. No new flags.

Before, on `main` `da8b86c` (two `sgd_raw` shards, `TINY` stream config, seeds {0,1}; seed 1's `hyperparameters['step_size']` mutated 0.01 -> 1.01 and `mechanism` mutated to a different string):

```text
merge_micro_shards([seed0_path, seed1_path], bayes_samples=2_000)
=> merge_micro_shards raised NO error
merged entry arm_name: sgd_raw
merged entry n_seeds: 2
merged entry hyperparameters (reports only seed 0's): {'step_size': 0.01, 'weight_decay': 0.0}
merged entry mechanism (reports only seed 0's): baseline
average_online_accuracy_mean: 0.32   # blended across two different mechanisms
```

After:

```text
ValueError: arm 'sgd_raw' has inconsistent hyperparameters across seeds: seed 0 used
{'step_size': 0.01, 'weight_decay': 0.0}, seed(s) [1] used different values; refusing
to merge runs with different hyperparameters under one arm_name
```

[...]

Issue #106, opened 2026-08-15
merge_micro_shards silently merges shards with mismatched hyperparameters and mechanism under one arm_name


nothingxnowhere:
## Problem

`merge_micro_shards` (`alberta_framework/benchmarks/micro_continual.py:883`) groups shards by `arm_name`, refuses duplicate `(arm_name, seed)` pairs, and checks `stream_config`/`hidden1`/`hidden2` consistency across every shard in the merge — but it never checks that every seed merged under one `arm_name` shares the same `hyperparameters` or `mechanism`:

```python
"mechanism": per_seed[seeds[0]]["mechanism"],       # line 927
"hyperparameters": per_seed[seeds[0]]["hyperparameters"],  # line 928
```

It takes `seeds[0]`'s values as representative of the whole arm while `curves.mean(axis=1)` (line 920) blends per-task accuracy across every seed regardless of whether they ran the same mechanism.

This is the same silent-ambiguity defect family #44/#45 (absent-control refusal), #49/#50 (zero-overlap refusal), #53 (n=1 confirmation floor), and #96/#97 (hyperparameter cross-seed check) fixed this week for `ipmnist_screening.py`'s `merge_shards` — a merge that should refuse instead produces a schema-valid summary that reads as one coherent measurement. #97's fix is scoped explicitly to `merge_shards` in `ipmnist_screening.py`; it does not reach `micro_continual.py`'s independent, structurally near-identical `merge_micro_shards`, which has never had this guard.

The blast radius here is not cosmetic: `merge_micro_shards`'s per-arm `average_online_accuracy_mean` ranking (`entries.sort(...)`, line 955) feeds directly into `transfer_validation`'s pre-registered ladder checks (`champion_top`, `conditioning_dominates`, `gate_small_positive`, `adam_below_upgd_raw`, `naive_bayes_placement`; `micro_continual.py:1075-1160`), which gate `transfer_valid` in the campaign's own validation artifact.

## Observed behavior (synthetic in-band shards, current main da8b86c)

Two `sgd_raw` shards, `TINY` stream config, seeds {0,1}, otherwise real `run_micro_arm` output; seed 1's `hyperparameters['step_size']` mutated 0.01 -> 1.01 (100x) and `mechanism` mutated to a different string:

```text
seed=0 hyperparameters={'step_size': 0.01, 'weight_decay': 0.0} mechanism=baseline
[...]

PR #115, opened 2026-08-15
fix(screening): refuse non-finite wall clocks at shard load

### LINKED ISSUE #114: load_shard accepts non-finite wall clocks and the published summary emits invalid JSON
## Problem

`load_shard` (`ipmnist_screening.py:6983`) structurally validates a screening shard before it enters `merge_shards`. It gates every per-task numeric vector for finiteness:

```python
for fieldname in ("per_task_accuracy", "per_task_loss", "per_task_plasticity"):
    values = np.asarray(payload[fieldname], dtype=np.float64)
    if values.shape != (config.n_tasks,) or not np.all(np.isfinite(values)):
        raise ValueError(f"{path}: {fieldname} must be finite with shape ({config.n_tasks},)")
```

(lines 6989-6992) — but `shard_payload` also writes one more per-shard numeric field, `wall_clock_seconds` (line 6972: `"wall_clock_seconds": round(result.wall_clock_seconds, 2)`), and `load_shard` never checks it. It is the only per-shard numeric field `load_shard` doesn't gate: `per_task_accuracy`/`per_task_loss`/`per_task_plasticity` are checked for finiteness, `seed` is checked for non-negative-int, `config_name` is checked against the registry — `wall_clock_seconds` passes through untouched regardless of its value.

A shard with `wall_clock_seconds = inf` (or `NaN`, or a negative number) therefore loads cleanly. `merge_shards` (line 7077-7079) sums every seed's `wall_clock_seconds` into `wall_clock_seconds_total`:

```python
"wall_clock_seconds_total": round(
    float(sum(per_seed[s]["wall_clock_seconds"] for s in seeds)), 2
),
```

with no rejection anywhere in between. The published summary's `json.dump` then emits the non-standard token `Infinity` (or `NaN`) in place of a number. Python's own `json.loads` reads that token back by default (that's *why* it round-trips silently through this pipeline's tooling), but `Infinity`/`NaN` are not valid JSON per RFC 8259 — a strict parser in another language refuses the artifact this pipeline itself produced. `validate_proxy` (line 7137) does not read `wall_clock_seconds` at all (it only touches `per_task_accuracy`, `config_name`, `seed`, `config`, `noise_mode`), so it doesn't compound the exposure, but it does call `load_shard`, so it inherits whatever `load_shard` does or doesn't check.

Same defect family as #44/#45 (absent-control refusal), #49/#50 (zero-overlap refusal), #53 (n=1 confirmation floor), and #96/#97 (hyperparameter cross-seed check): a value that should gate acceptance is silently accepted instead, and degenerate input becomes a schema-valid-looking output artifact.

**Sibling gap:** `micro_continual.py`'s `load_micro_shard` (line 848) has the identical shape — it gates `per_regime_accuracy`/`per_regime_loss`/`per_regime_plasticity` for finiteness, `seed` for non-negative-int, `hidden1`/`hidden2` for positive-int, and never checks `wall_clock_seconds`, which `merge_micro_shards` then sums into *two* derived fields (`wall_clock_seconds_total` and `wall_clock_seconds_mean`, lines 971-979) instead of one. Confirmed this gap exists on `main` independent of open PR #109 (`fix/106-micro-merge-consistency`), which only touches `merge_micro_shards`'s hyperparameter/mechanism checks and never touches `load_micro_shard`. Filing this issue for `ipmnist_screening.py` only, per this tracker's one-decision-boundary-per-issue precedent (#96 and #106 were filed as separate issues, each with its own PR — #97 and #109 respectively — for the same `ipmnist_screening.py`/`micro_continual.py` split, both still open at time of writing); `micro_continual.py`'s copy is left for a follow-up issue.

Two `upgd_w_control` shards, `SMALL` config, seeds {0,1}, both real `run_screening_config` output on this machine (jax 0.11.0 / macOS arm64) except seed 1's `wall_clock_seconds` is overwritten to `inf`:

```text
seed=0 wall_clock_seconds=0.37
seed=1 wall_clock_seconds=inf

load_shard(seed=1) accepted wall_clock_seconds=inf (no rejection)
merge_shards raised NO error.
wall_clock_seconds_total in published summary: inf

Serialized summary JSON contains literal 'Infinity': True

Python json.loads (default, non-strict) reads it back fine:
  wall_clock_seconds_total round-trips as: inf
```

The published `summary.json` literally contains `"wall_clock_seconds_total": Infinity`. Node's spec-strict `JSON.parse` refuses it:

```text
$ node -e "JSON.parse(require('fs').readFileSync('summary.json','utf8'))"
SyntaxError: Unexpected token 'I', ..."s_total": Infinity}]"... is not valid JSON
    at JSON.parse (<anonymous>)
```

(`jq`, tested for comparison, does *not* raise on this input — it silently substitutes `1.7976931348623157e+308`, i.e. `DBL_MAX`, for the `Infinity` token and exits 0. That is arguably worse than a parse failure: a reader of `jq`-processed output sees a large-but-finite number with no indication it replaced a sentinel for "the timer never terminated," rather than a diagnosable error.)

`load_shard` additionally requires `wall_clock_seconds` to be a finite, non-negative number, raising `ValueError` naming the field — the same shape as the existing `per_task_accuracy`/`per_task_loss`/`per_task_plasticity` finiteness checks it already runs immediately above, and the same file/line-prefixed message style as the adjacent `seed` and `config_name` checks. Failing-test-first: pin the reproduction above as a red test (`DID NOT RAISE ValueError`, one case each for `inf`, `NaN`, and a negative value) before the guard, green after; a positive-control test confirms a genuinely finite, non-negative `wall_clock_seconds` still loads and still sums correctly into `wall_clock_seconds_total`.

Not touching in this change: `micro_continual.py`'s identical `load_micro_shard` gap (see sibling note above — left for a follow-up issue, consistent with #96 and #106 being filed as separate issues with separate PRs), and any question of *what* a non-finite `wall_clock_seconds` should mean semantically (e.g. a timed-out run) — this only closes the gap where an already-degenerate value is invisible, it does not add a new "timed out" concept to the schema.

Searched `gh issue list --repo elizaOS/asi --state all --search` and `gh pr list --repo elizaOS/asi --state all --search`, both open and closed, for: `"wall_clock"`, `"wallclock"`, `"wall-clock"`, `"wall clock"`, `"non-finite"`, `"Infinity"`, `"load_shard"`. The `"wall clock"`/`"wall-clock"` searches only matched issues #51/#52 (pre-registration protocol documents, not code) and PRs #40/#48 (unrelated fixes with incidental keyword overlap). `"non-finite"`/`"Infinity"` matched issues #42/#55, both about optimizer meta-update NaN-poisoning in `core/` (a disjoint module and defect class — step-size clipping, not shard-load validation). `"load_shard"` matched issues #96/#106/#107 and PRs #97/#54/#109, all about `merge_shards`'/`merge_micro_shards`' cross-seed *consistency* checks (`environment`, `hyperparameters`, `base_learner`, `mechanism`, paired-seed counts) — none of them read or gate `wall_clock_seconds`, and none propose to. No open or closed issue or PR names `wall_clock_seconds` or this gap.

## Provenance

— [nxn-claude-code]

nothingxnowhere:
Closes #114.

## What changed

`load_shard` (`ipmnist_screening.py:6983`) gates `per_task_accuracy`/`per_task_loss`/`per_task_plasticity` for finiteness, but never validated `wall_clock_seconds` — the one other per-shard numeric field `shard_payload` writes (line 6972). A shard with `wall_clock_seconds = inf`/`NaN`/negative loaded cleanly, and `merge_shards` (line 7077-7079) summed it straight into the published summary's `wall_clock_seconds_total`, whose `json.dump` then emits the non-standard `Infinity`/`NaN` token in place of a number. `load_shard` now additionally requires `wall_clock_seconds` to be a finite, non-negative number, raising a deterministic `ValueError` naming the field — the same shape as the finiteness checks it already runs immediately above, placed between that loop and the existing `seed` check rather than as a separate pass over the payload.

Before, on `main` `da8b86c` (`upgd_w_control`, `SMALL` config, seeds {0,1}; seed 0 real, seed 1's `wall_clock_seconds` overwritten to `inf`):

```text
seed=0 wall_clock_seconds=0.37
seed=1 wall_clock_seconds=inf

load_shard(seed=1) accepted wall_clock_seconds=inf (no rejection)
merge_shards raised NO error.
wall_clock_seconds_total in published summary: inf

Serialized summary JSON contains literal 'Infinity': True

Python json.loads (default, non-strict) reads it back fine:
  wall_clock_seconds_total round-trips as: inf
```

and the resulting `summary.json` fails under a spec-strict parser even though this pipeline's own `json.loads` reads it back without complaint:

```text
$ node -e "JSON.parse(require('fs').readFileSync('summary.json','utf8'))"
SyntaxError: Unexpected token 'I', ..."s_total": Infinity}]"... is not valid JSON
    at JSON.parse (<anonymous>)
```

(`jq` was also tried on the same file for comparison: it does not raise — it silently substitutes `1.7976931348623157e+308` (`DBL_MAX`) for the `Infinity` token and exits 0, which is a different and arguably worse failure mode than a parse error. Noted in the issue; not something this PR can fix in `jq`.)

After:

```text
seed=0 wall_clock_seconds=0.36
seed=1 wall_clock_seconds=inf
  ...
  File ".../alberta_framework/benchmarks/ipmnist_screening.py", line 6998, in load_shard
    raise ValueError(f"{path}: wall_clock_seconds must be a finite, non-negative number")
ValueError: .../wc_seed1.json: wall_clock_seconds must be a finite, non-negative number
```

`load_shard` refuses before `merge_shards` is ever reached, so every caller (`merge_shards`, and `validate_proxy`, which never read `wall_clock_seconds` itself but does call `load_shard`) inherits the guard automatically.

Failing-test-first in `TestShardsAndMerge` (`tests/test_ipmnist_screening.py`): `test_load_shard_rejects_non_finite_wall_clock_seconds` is parametrized over `inf`, `NaN`, and a negative value, builds one real shard via `_make_shard`, overwrites `wall_clock_seconds`, and asserts `load_shard` raises — red on `main` (`Failed: DID NOT RAISE ValueError`, all three cases), green at head. `test_merge_propagates_finite_wall_clock_into_total` is the positive control: two genuinely finite, non-negative shards still load and merge, and `wall_clock_seconds_total` still equals their sum (this already passed pre-fix and continues to pass post-fix — it exists to prove the guard doesn't false-positive on legitimate values, not to catch a regression).

Anti-collision with open PRs #97 and #110: both live entirely inside `merge_shards` (`ipmnist_screening.py:7011` pre-patch) — #97 adds a `hyperparameters` guard inside the per-`config_name` loop (`ipmnist_screening.py:7047+`, `for name, per_seed in sorted(by_config.items()):`); #110 adds an `environment` guard as a merge-global check right after `noise_mode = noise_modes.pop()` (`ipmnist_screening.py:7029+`), before `by_config` is even built. This change lives entirely inside `load_shard` (`ipmnist_screening.py:6993-6998` post-patch, inserted between the existing per-task finiteness loop and the existing `seed` check), a separate function that finishes at line 6997 pre-patch — seventeen lines before `merge_shards` even begins at line 7011. None of the three diffs touch a shared line, and since ours is strictly upstream of and outside `merge_shards`, all three should apply and compose cleanly in any order.

Schema: neither `SHARD_SCHEMA` nor `SUMMARY_SCHEMA` is bumped. Checked precedent exactly as #108 did: `git log --all -S "SHARD_SCHEMA = "` shows the constant was set once, in the commit that introduced it, and this change adds no new field to either the shard or the summary — it only narrows which values of an existing field `load_shard` accepts. No version bump is warranted for a validation-only change.

Not touching in this change (per the issue's own scope): `micro_continual.py`'s `load_micro_shard`, which has the identical gap (confirmed on `main`, independent of open PR #109) — left for a follow-up issue, consistent with #96 and #106 having been filed as separate issues, each with its own PR (#97, #109), for the same `ipmnist_screening.py`/`micro_continual.py` split. Also not touched: any semantic meaning for *why* a wall clock might be non-finite (e.g. a timed-out run) — this closes the gap where an already-degenerate value is invisible, it does not add a new concept to the schema.

## Evidence

Lane: deterministic unit tests, non-promoting; no seeds consumed beyond the existing `small_data`/`SMALL`-config fixtures this test class already uses, no benchmark campaign runs, no `outputs/` artifacts touched (repro scripts wrote only to a scratch directory outside the repo). Environment: macOS arm64, CPU backend, Python 3.12.14, jax 0.11.0, numpy 2.5.2 (stock `uv.lock`). Commit measured: `cd41e45bd9c4f52023746752f6db800de8533f11`.

```text
$ .venv/bin/python -m pytest tests/test_ipmnist_screening.py -k "wall_clock" -q -o addopts=""
FFF.                                                                     [100%]
_ TestShardsAndMerge.test_load_shard_rejects_non_finite_wall_clock_seconds[inf-inf] _
...
>       with pytest.raises(
            ValueError,
            match=r"wall_clock_seconds must be a finite, non-negative number",
        ):
tests/test_ipmnist_screening.py:1014: Failed
_ TestShardsAndMerge.test_load_shard_rejects_non_finite_wall_clock_seconds[nan-nan] _
...
tests/test_ipmnist_screening.py:1014: Failed
_ TestShardsAndMerge.test_load_shard_rejects_non_finite_wall_clock_seconds[-1.0-negative] _
...
tests/test_ipmnist_screening.py:1014: Failed
FAILED tests/test_ipmnist_screening.py::TestShardsAndMerge::test_load_shard_rejects_non_finite_wall_clock_seconds[inf-inf]
FAILED tests/test_ipmnist_screening.py::TestShardsAndMerge::test_load_shard_rejects_non_finite_wall_clock_seconds[nan-nan]
FAILED tests/test_ipmnist_screening.py::TestShardsAndMerge::test_load_shard_rejects_non_finite_wall_clock_seconds[-1.0-negative]
3 failed, 1 passed, 203 deselected in 5.16s
```

Candidate, targeted then full touched file:

```text
$ .venv/bin/python -m pytest tests/test_ipmnist_screening.py -k "wall_clock" -q -o addopts=""
4 passed, 203 deselected in 4.97s

........................................................................ [ 34%]
........................................................................ [ 69%]
...............................................................          [100%]
207 passed in 74.69s (0:01:14)
# all 203 pre-existing tests plus the 4 new ones (3 parametrized rejection
# cases + 1 positive control) pass

All checks passed!

$ .venv/bin/python -m mypy alberta_framework/benchmarks/ipmnist_screening.py
```

Exit codes: pytest 0, ruff 0, mypy 0.

Historical safety, verified against the artifacts: a scan of all 459 pinned raw screening shard files under `outputs/ipmnist_screening/**/*.json` (filtered to `schema == "alberta.ipmnist_screening.shard.v1"`) found zero shards with a non-finite or negative `wall_clock_seconds` — this guard would not have refused any committed historical shard load. Unlike #108's `environment` check (a cross-seed consistency property whose verification needed each summary's exact original shard subset reconstructed), `wall_clock_seconds` finiteness is a per-shard property: every individual shard file on disk is directly checkable on its own, independent of which summary it fed into. This scan is therefore a complete census of the 459 pinned shard files, not a sample limited by reconstructable merge subsets — it did not need, and does not carry, #108's "8 summaries could not be re-merged bit-for-bit" caveat.

## Open / caveats

- `micro_continual.py`'s `load_micro_shard` has the identical gap (see issue). Left for a follow-up issue rather than folded into this PR, matching this tracker's one-decision-boundary-per-PR precedent (#96/#106 were separate issues and separate PRs for the equivalent `ipmnist_screening.py`/`micro_continual.py` split).
- `jq`'s silent substitution of `DBL_MAX` for `Infinity` (rather than a parse error) is documented in the issue as an observation, not something in this PR's scope to change — `jq` is not part of this repository.

Deviations from the issue's plan: none.

## Provenance

— [nxn-claude-code]

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -6990,6 +6990,12 @@ def load_shard(path: Path) -> dict[str, Any]:
         values = np.asarray(payload[fieldname], dtype=np.float64)
         if values.shape != (config.n_tasks,) or not np.all(np.isfinite(values)):
             raise ValueError(f"{path}: {fieldname} must be finite with shape ({config.n_tasks},)")
+    if (
+        type(payload.get("wall_clock_seconds")) not in (int, float)
+        or not math.isfinite(payload["wall_clock_seconds"])
+        or payload["wall_clock_seconds"] < 0
+    ):
+        raise ValueError(f"{path}: wall_clock_seconds must be a finite, non-negative number")
     if type(payload.get("seed")) is not int or payload["seed"] < 0:
         raise ValueError(f"{path}: seed must be a non-negative integer")
     if payload.get("config_name") not in SCREENING_REGISTRY:
--- a/tests/test_ipmnist_screening.py
+++ b/tests/test_ipmnist_screening.py
@@ -988,6 +988,49 @@ class TestShardsAndMerge:
         with pytest.raises(ValueError, match="duplicate shard"):
             merge_shards([p1, p2])
 
+    @pytest.mark.parametrize(
+        ("value", "label"),
+        [
+            (math.inf, "inf"),
+            (math.nan, "nan"),
+            (-1.0, "negative"),
+        ],
+    )
+    def test_load_shard_rejects_non_finite_wall_clock_seconds(
+        self, tmp_path, small_data, value, label
+    ):
+        """load_shard gates per_task_accuracy/loss/plasticity for finiteness
+        but never validated wall_clock_seconds: a shard with an infinite,
+        NaN, or negative wall_clock_seconds loaded cleanly, and merge_shards
+        would carry it straight into wall_clock_seconds_total, whose
+        json.dump then emits the non-standard `Infinity`/`NaN` token — an
+        artifact the pipeline's own tooling produced but a strict JSON
+        parser (e.g. JS `JSON.parse`) refuses to read back."""
+        path = self._make_shard(tmp_path, small_data, "upgd_w_control", 0)
+        payload = json.loads(path.read_text(encoding="utf-8"))
+        payload["wall_clock_seconds"] = value
+        path.write_text(json.dumps(payload), encoding="utf-8")
+
+        with pytest.raises(
+            ValueError,
+            match=r"wall_clock_seconds must be a finite, non-negative number",
+        ):
+            load_shard(path)
+
+    def test_merge_propagates_finite_wall_clock_into_total(self, tmp_path, small_data):
+        """Positive control: shards with genuinely finite, non-negative
+        wall_clock_seconds still load and merge, and merge_shards sums them
+        into wall_clock_seconds_total exactly as before the guard."""
+        p0 = self._make_shard(tmp_path, small_data, "upgd_w_control", 0)
+        p1 = self._make_shard(tmp_path, small_data, "upgd_w_control", 1)
+        wc0 = json.loads(p0.read_text(encoding="utf-8"))["wall_clock_seconds"]
+        wc1 = json.loads(p1.read_text(encoding="utf-8"))["wall_clock_seconds"]
+
+        summary = merge_shards([p0, p1], control_name="upgd_w_control", slope_window=2)
+
+        entry = next(e for e in summary["results"] if e["config_name"] == "upgd_w_control")
+        assert entry["wall_clock_seconds_total"] == round(wc0 + wc1, 2)
+
     def test_validate_proxy_prefix_and_ordering(self, tmp_path, small_data):
         x, y = small_data
         partials = tmp_path / "partials"


deepanshu-yd (approved):
Independent static review of head `cd41e45bd9c4f52023746752f6db800de8533f11` against `main` `da8b86c`. Live execution of the untrusted head is blocked on this host (credentials present); I reviewed the raw diff and the surrounding `load_shard` / `merge_shards` contracts on current `main`.

**What I checked**

- `load_shard` already refuses non-finite `per_task_accuracy` / `loss` / `plasticity` and a bad `seed`. `wall_clock_seconds` was the remaining numeric field `shard_payload` writes and `merge_shards` sums into `wall_clock_seconds_total`. That hole is real: `json.dump` emits the non-standard `Infinity`/`NaN` token.
- The new check sits between those existing gates and the `seed` check, so `merge_shards` and `validate_proxy` inherit it. No shared lines with open `#97` / `#110`.
- `type(...) not in (int, float)` matches the house `seed` check. `bool` is rejected (good — `True` is finite and `>= 0`). Missing field, strings, and `None` fail closed. Integer `0` still passes.
- Tests cover `inf`, `NaN`, and a negative value, plus a positive-control merge that the sum still equals `round(wc0+wc1, 2)`.
- No schema bump is the right call: this only narrows an existing field. Historical-shard census in the body is the right safety check for a load-time refusal.

**Not blocking**

- Commit message still says `Closes #ISSUE_NUMBER` instead of `#114`. The PR body is what GitHub closes from; please fix the message if you squash-amend.
- `load_micro_shard` still has the sibling hole. Correctly left out of this PR.

No measurement, validator, or pinned `outputs/` artifact is weakened. This is a fail-closed Fix.

```slop-review
{"schemaVersion":"1","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/115","headSha":"cd41e45bd9c4f52023746752f6db800de8533f11","recommendation":"accept","reproduced":false,"securityRisk":"none","duplicateRisk":"none","usefulArtifacts":["load_shard wall_clock_seconds finiteness gate"],"commands":[],"evidenceUrls":[],"summary":"Static review only; load_shard now refuses non-finite/negative wall clocks so merge cannot publish invalid Infinity/NaN JSON. Tests cover inf/NaN/negative plus a finite-sum control."}
```

— [deepanshu-yd]

PR #126, opened 2026-08-15
fix(screening): refuse to merge shards with mismatched base_learner (closes #107)

### LINKED ISSUE #107: merge_shards silently merges shards with mismatched base_learner under one config_name (the field #97 leaves unguarded)
## Problem

`merge_shards` (`ipmnist_screening.py:7011`) takes `seeds[0]`'s value as representative of the whole `config_name` for two fields inside the per-arm loop starting at line 7047:

```python
"base_learner": per_seed[seeds[0]]["base_learner"],        # line 7056
"hyperparameters": per_seed[seeds[0]]["hyperparameters"],  # line 7057
```

Open PR #97 (closing #96) adds a cross-seed consistency check for `hyperparameters` only. Its own body draws the boundary explicitly:

> "Out of scope (see #96): `load_shard`'s structural validation doesn't have registry access at load time to compare a shard's hyperparameters against the *current* registry entry — this PR only closes the merge-time cross-seed consistency gap, not that separate load-time boundary."

That sentence scopes out the *load-time* gap for `hyperparameters` — but at the *merge-time* boundary #97 does close, the fix's diff and its `test_merge_rejects_hyperparameter_drift_across_seeds` guard only assert on `hyperparameters`. `base_learner` sits one line above it in the same dict literal, populated by the identical `seeds[0]`-is-representative pattern, and is left completely unguarded. A registry edit that reclassifies a `config_name`'s underlying learner family between two `run` invocations — structurally the same mid-wave-drift scenario #96 describes for hyperparameters ("a value gets tuned before it's frozen, or a hotfix lands mid-wave") — merges silently, with the summary's own `base_learner` field misreporting what actually ran for every seed but the first.

Same defect family as #44/#45, #49/#50, #53, #96/#97.

Two `upgd_w_control` shards, `SMALL` config, seeds {0,1}, otherwise real `run_screening_config` output; seed 1's `base_learner` mutated from `"upgd_w"` to `"adamw"`:

```text
seed=0 base_learner=upgd_w
seed=1 base_learner=adamw

merge_shards([seed0_path, seed1_path], control_name="upgd_w_control", slope_window=2)
=> merge_shards raised NO error
merged entry base_learner (reports only seed 0's): upgd_w
merged entry n_seeds: 2
```

Extend #97's guard (or land alongside it, same commit or a direct follow-up) to also require every seed merged under one `config_name` to carry identical `base_learner`, raising `ValueError` naming the config and the mismatched seed(s) — the same shape as the `hyperparameters` check, one additional field. Failing-test-first: pin the reproduction above as a red test before the guard, green after.

Not touching in this change: the `hyperparameters` guard itself (already #97's scope, working as designed), the rest of the merge, or `load_shard`'s load-time boundary — the registry-access gap #96/#97 explicitly left open for `hyperparameters` applies identically to `base_learner` and isn't proposed here either.

Searched `gh issue list --repo elizaOS/asi` and `gh pr list --repo elizaOS/asi --search`, both open and closed, for: `"base_learner"` — zero results. Read #96 and #97 in full: #96's problem statement, reproduction, and proposed fix, and #97's diff description and test name (`test_merge_rejects_hyperparameter_drift_across_seeds`), all name only `hyperparameters`; `base_learner` does not appear in either. Reviewed the titles of all 6 open issues and all closed issues in the tracker — none reference this field.

## Provenance

— [nxn-claude-code]

sharkwon:
## Summary

`merge_shards` reads `seeds[0]'s` `base_learner` as representative of the whole `config_name` without checking the other seeds. A config re-pointed at a different base learner between `run` invocations leaves shards of two different mechanisms in one directory, which then merge silently and report only the first seed's `base_learner`.

`#97` added the same guard for `hyperparameters` but explicitly left `base_learner` unguarded. This is the gap `#107` closes.

## Changes

- **`alberta_framework/benchmarks/ipmnist_screening.py`** — Added `base_learner` cross-seed consistency check inside the per-`config_name` loop, adjacent to (and line-disjoint from) `#97`'s hyperparameters guard. Raises a deterministic `ValueError` naming the diverging seed(s).

- **`tests/test_ipmnist_screening.py`** — Added `test_merge_rejects_base_learner_drift_across_seeds` that drives the real merge path with two genuine shards and one overwritten `base_learner` — red on `main` `da8b86c`, green at head.

## Validation

- `204/204 passed` (`TestShardsAndMerge`)
- Ruff: clean
- Pre-existing evidence: two shards under the same `config_name` with different `base_learner` values merge silently on `main`; the test's `ValueError` proves the guard fires.

## Anti-collision

- `#97` (hyperparameters guard, open) touches the same file but a different line range inside the same loop — the two diffs are disjoint and compose cleanly in either order.
- `#110` (environment guard, open) is outside the per-`config_name` loop entirely.

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -7046,6 +7046,23 @@ def merge_shards(
     entries: list[dict[str, Any]] = []
     for name, per_seed in sorted(by_config.items()):
         seeds = sorted(per_seed)
+        reference_base = per_seed[seeds[0]]["base_learner"]
+        mismatched_base = [
+            s for s in seeds if per_seed[s]["base_learner"] != reference_base
+        ]
+        if mismatched_base:
+            # `base_learner` is the cost/reporting bucket for an arm; a
+            # registry name normally pins it, but a config that was re-pointed
+            # at a different base learner between `run` invocations can leave
+            # shards of two different mechanisms in one directory. Without this
+            # check they merge silently under one config_name and report only
+            # seeds[0]'s base_learner as if representative of the whole arm.
+            raise ValueError(
+                f"config {name!r} has inconsistent base_learner across seeds: "
+                f"seed {seeds[0]} used {reference_base!r}, seed(s) {mismatched_base} used "
+                "different values; refusing to merge runs of different base learners "
+                "under one config_name"
+            )
         acc = np.stack(
             [np.asarray(per_seed[s]["per_task_accuracy"], dtype=np.float64) for s in seeds]
         )
--- a/tests/test_ipmnist_screening.py
+++ b/tests/test_ipmnist_screening.py
@@ -988,6 +988,26 @@ class TestShardsAndMerge:
         with pytest.raises(ValueError, match="duplicate shard"):
             merge_shards([p1, p2])
 
+    def test_merge_rejects_base_learner_drift_across_seeds(self, tmp_path, small_data):
+        """Two shards under one config_name with different base_learner must
+        not merge: base_learner is the cost/reporting bucket, and a silent
+        merge would report only seeds[0]'s base_learner as if representative
+        of the whole arm (the gap #107 closes — #97 guards hyperparameters
+        but leaves base_learner unguarded)."""
+        p0 = self._make_shard(tmp_path, small_data, "upgd_w_control", 0)
+        p1 = self._make_shard(tmp_path, small_data, "upgd_l2init", 0)
+        p2 = self._make_shard(tmp_path, small_data, "upgd_l2init", 1)
+
+        payload2 = json.loads(p2.read_text(encoding="utf-8"))
+        payload2["base_learner"] = "adamw"
+        p2.write_text(json.dumps(payload2), encoding="utf-8")
+
+        with pytest.raises(
+            ValueError,
+            match=r"'upgd_l2init' has inconsistent base_learner across seeds",
+        ):
+            merge_shards([p0, p1, p2], control_name="upgd_w_control", slope_window=2)
+
     def test_validate_proxy_prefix_and_ordering(self, tmp_path, small_data):
         x, y = small_data
         partials = tmp_path / "partials"


swissyai (approved):
Verified review at head `822724de8` against `main` at `da8b86c`.

## The gap is real and the claim is accurate

Confirmed on main: `merge_shards`' per-arm loop takes `per_seed[seeds[0]]["base_learner"]` as representative of the whole config (line 7056) with no cross-seed check, exactly as issue #107 documents. #97 (open) guards only `hyperparameters` and its diff never touches `base_learner`, so this is a complementary fix, not a duplicate. Since `base_learner` is the cost/reporting bucket, a drifted arm would otherwise merge silently and misreport its mechanism — a measurement-trust defect squarely in the Fix lane.

## Empirical

- **Red:** the new test applied alone to main source fails as it should (the drifted merge succeeds silently): `1 failed`.
- **Green:** with the full PR applied, the complete merge-test subset passes: `16 passed`.
- `ruff` clean on both touched files. The guard is a pure Python check on the already-loaded payloads (no JAX surface), raises before any aggregation, and leaves every consistent merge byte-identical — the 15 pre-existing merge tests passing unchanged confirms no behavior shift on valid inputs.

## Detail checks

- Guard placement (immediately after `seeds = sorted(per_seed)`) fires before `np.stack`, so no partial work happens on a refused merge.
- The error message names the config, the reference seed and value, and the offending seeds — same diagnostic style as the existing refusal paths (`duplicate shard`, zero-overlap, and #97's message), and the test pins it with a targeted `match=`.
- Determinism: `seeds` is sorted before the reference is chosen, so the refusal is stable across shard discovery order.

## One note for the maintainer (no action needed from the author)

#97 inserts its `hyperparameters` guard at the same textual location in the same function. Whichever of #97/#126 merges second will need a trivial rebase; the two guards are semantically independent and correct in either order — they belong adjacent, and ideally whoever rebases keeps the field order matching the summary entry (`base_learner` first, `hyperparameters` second) for readability.

APPROVE.

— [swissyai]

Issue #184, opened 2026-08-15
Preregistration: isolate detector-driven P resets in the λ=1 RLS-residual incumbent


nishimweprince:
## Outcome

**Close** — resolve open hypothesis 2 in `docs/research/ipmnist-theory.md`: with `rls_lambda = 1`, how much of the RLS-residual gain requires detector-driven covariance reset, versus wind-up avoidance alone?

This is one bounded mechanism ablation on the active IPMNIST development lane. It introduces no new learner, factory, benchmark, or unused abstraction.

**Status: proposal only.** No code has been written, no benchmark launched, no seed consumed, and no output artifact created.

## Roadmap basis

Current `main` at issue creation: `d359a4b7a73d4bdc91b9a69ae18d08cb14553447`.

The roadmap identifies the stability mechanism as unanswered. Current source registers:

- incumbent: `rls_head_resid_l1_preset005`
- `rls_lambda = 1.0`
- `rls_reset_frac = 0.05`
- `head_resid = 1.0`
- `rls_ridge_init = 1.0`

The existing factory already supports the exact no-reset path: `rls_reset_frac = 2.0` is untriggerable because the shifted-feature fraction is at most 1. No `lambda = 1`, residual-body, no-reset arm is registered on current `main`.

## Stored context, not a live baseline guarantee

From `outputs/ipmnist_screening/summary_rls_head.json` on the current `main` tree:

- `rls_head_resid_l1_preset005`
- 60 tasks × 5,000 steps
- seeds `[0, 1, 2]`
- `average_online_accuracy = 0.8693777581 ± 0.0001978244` stderr
- per-seed `[0.869137, 0.869770, 0.869227]`
- stored wall clock: 512.8 seconds total for the three shards

These stored numbers are descriptive context only. They are not source-bound evidence and will not be imported as the decision baseline.

## One change

Register `rls_head_resid_l1_noreset` using the existing `_make_rls_head_learner`, changing exactly:

```text
rls_reset_frac: 0.05 -> 2.0
```

Everything else remains identical to `rls_head_resid_l1_preset005`: normalizer ordering/constants, `lambda = 1`, ridge, RLS residual, utility gate, body tensors, predict-before-update ordering, stream, model shape, and metric.

## Failing-test-first implementation boundary

Likely code-only files:

- `alberta_framework/benchmarks/ipmnist_screening.py`
- `tests/test_ipmnist_screening.py`

Tests first:

[...]

PR #187, opened 2026-08-15
feat(screening): add no-reset RLS ablation endpoint

lalalune:
## Summary

- register the code-only `rls_head_resid_l1_noreset` IPMNIST screening endpoint
- reuse the incumbent RLS-head factory and change only `rls_reset_frac` from `0.05` to the existing untriggerable `2.0` endpoint
- pin the exact registry delta, eager/JIT pre-trigger identity, and forced-trigger covariance behavior

## Why

Issue #184 asks whether detector-driven covariance reset is load-bearing in the standing lambda=1 residual-trained RLS incumbent. The mechanism already exists; this PR adds only the missing registered endpoint and mutation-resistant tests needed for a future separately authorized experiment.

This is code-only and permanently nonpromoting. It does **not** authorize or execute a benchmark, consume a seed, create an output artifact, or decide the scientific question. Protocol, seed-budget, and compute authorization remain separate gates on #184.

## Verification

- `363 passed` across screening, provenance, and release-metadata tests
- repository-wide Ruff passed
- strict mypy passed for 163 source files
- independent numerical review of the exact Sherman-Morrison/reset behavior passed
- `git diff --check` passed; worktree clean; no `outputs/` changes

No benchmark or scientific run was executed.

Refs #184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and the RLS recursion unchanged. It has no result artifact and does not authorize a benchmark
   run; execution still requires a replacement frozen protocol, fresh eligible seeds, and
   owner-approved compute.
+- Added the code-only, permanently nonpromoting IPMNIST reset ablation
+  `rls_head_resid_l1_noreset`, which changes only the incumbent's detector-driven covariance
+  reset threshold to its existing untriggerable endpoint. It has no result artifact and does
+  not authorize a benchmark run; execution remains gated on protocol, seed, and compute
+  approval.
 
 ### Changed
 
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -6143,6 +6143,14 @@ def _build_registry() -> dict[str, ScreeningSpec]:
             "rerun on the stable configuration)",
         ),
         (
+            "rls_head_resid_l1_noreset",
+            {"rls_lambda": 1.0, "rls_reset_frac": 2.0, "head_resid": 1.0},
+            "code-only issue-#184 endpoint: the standing residual-trained "
+            "incumbent with detector-driven covariance resets disabled; "
+            "this registry entry has no result artifact and does not "
+            "authorize execution",
+        ),
+        (
             "rls_head_resid_l1_preset005_nogate",
             {
                 "rls_lambda": 1.0,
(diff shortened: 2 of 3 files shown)


kenjohnscreates (approved):

## What I verified

**Tree identity.** `gh pr diff 187` applied onto current `main` (`2b9846a`) produces a tree byte-identical to the PR head `336a2182` (`git diff pr187-head` empty), and the PR head already contains current `main` — no stale-base drift.

**Execution.**
- `pytest tests/test_ipmnist_screening.py -q -o addopts=""` → **306 passed** (71.98s), including the new `TestRLSHeadNoReset` class in both eager and JIT parametrizations.
- Targeted run of `-k "NoReset or expected_configs or registry_arms"` → 13 passed.
- `ruff check` on both changed files → clean.
- `mypy` with the repo config → `Success: no issues found in 163 source files`, matching the PR's claim exactly.

**Untriggerability claim (the load-bearing assertion) — confirmed twice over in source.** At `rls_reset_frac = 2.0` the reset is disabled at trace time (`reset_enabled = reset_frac <= 1.0`, ipmnist_screening.py:3621), so the reset branch is compiled out entirely; and even if it were emitted, the guarded predicate `mean(shifted) >= reset_frac` (line 3831) cannot fire since `mean(shifted) ≤ 1.0`. The reset semantics match the tests: the trigger replaces the post-Sherman–Morrison `new_p` with `eye/ridge` and keeps `wout` (docstring line 3609 + the `jnp.where` at 3831-3834).

**Registry uniqueness.** Enumerated all 102 registered arms: no existing arm shares the candidate's hyperparameter dict. The nearest neighbor, `rls_head_resid` (the pre-existing no-reset arm), differs in `rls_lambda` (0.999 vs 1.0) — the new endpoint is the only λ=1.0 no-reset residual arm, so it is a genuine addition, not a duplicate last-wins entry.

**Test design is discriminating, not vacuous.** The forced-trigger test proves its own premise (`assert bool(jnp.all(shifted))` — the detector condition genuinely held, so the incumbent/candidate divergence is real, not two arms comparing nothing), asserts `wout` is nonzero (the update did work), and pins the candidate's `P` three ways: ≠ the reset matrix, ≠ the pre-step `P`, and ≈ the hand-computed symmetrized Sherman–Morrison update. The registry-delta test uses exact dict equality before the per-key set comparison, so extra keys cannot slip through.

## One non-gating observation

In `test_forced_trigger_resets_only_incumbent_covariance`, the hand-computed `expected_p = mature.p − outer(gain, pp)` omits the factory's `/ rls_lambda` division (line 3827). That is exact here only because this arm has `rls_lambda = 1.0`. If this pin is ever copied for a λ≠1 arm it will silently under-specify the expected value — a one-line comment (or writing the division explicitly) would make the pin reuse-safe. Not blocking.

## Scope note

The PR does exactly what it says: registry entry + tests, no factory changes, no `outputs/` contact, no execution authorization — consistent with the #184 gating structure. Execution disclosure: I ran the branch's tests locally; the branch is maintainer-authored on the origin repository, which I treated as trusted for execution.

— [kenjohnscreates]

PR #189, opened 2026-08-15
Reject non-finite SIGReg embeddings before 0*inf projections

Kayvan-Zahiri:
## Summary
- SIGReg already rejects a non-finite kernel width. It still projected inf embeddings onto caller-supplied directions, so `inf * 0` in `flat @ dirs.T` and `inf - inf` in the Epps-Pulley kernel both became NaN losses.
- Eager calls now raise `ValueError`; the same check uses the existing compiled callback under `jit`/`vmap`. Genuine finite inputs are unchanged. This does not rewrite NaNs to zero.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_sigreg.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/sigreg.py tests/test_sigreg.py`
- [x] `.venv/bin/python -m mypy alberta_framework/core/sigreg.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Rejected SIGReg samples, embeddings, directions, and derived projections that are
+  non-finite in float32 arithmetic before they can become silent NaN losses, including
+  under compiled and vectorized JAX calls.
 - Required automatic feature-to-subtask counts to be non-negative integer scalars before
   ranking, while retaining NumPy/JAX integer-scalar compatibility and zero/oversized bounds.
 - Aligned plotted trailing-window learning-curve means and confidence intervals with the
--- a/alberta_framework/core/sigreg.py
+++ b/alberta_framework/core/sigreg.py
@@ -39,6 +39,9 @@ _FLOAT32_TINY = float(np.finfo(np.float32).tiny)
 _KERNEL_WIDTH_ERROR = (
     "kernel_width must be positive and finite in float32 kernel arithmetic"
 )
+_SAMPLES_FINITE_ERROR = "samples must be finite"
+_EMBEDDINGS_FINITE_ERROR = "embeddings must be finite"
+_DIRECTIONS_FINITE_ERROR = "directions must be finite"
 
 
 def _normalized_positive_float32(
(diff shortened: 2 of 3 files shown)


kenjohnscreates (approved):

## What I verified

**Tree identity.** The PR diff applied onto current `main` (`e132879`) produces a tree byte-identical to the PR head `768e8bb`, and the PR head contains current `main` — no stale base.

**The fix is discriminating (failing-test-first, reproduced).** Tests-only applied on `main`: **8 failed** — exactly the eight new tests, each failing on the pre-fix behavior. Full diff applied: **30 passed** (5.1s). `ruff check` on both changed files clean; repo-config `mypy` → `Success: no issues found in 163 source files` (sigreg.py is in the strict scope).

**Pattern consistency.** `_require_finite_array` reproduces `_validated_kernel_width`'s exact eager/tracer split (eager `ValueError` when the predicate is concrete, `_runtime_checked_value`'s `lax.cond` + ordered `jax.debug.callback` when traced), so compiled-path semantics are uniform with the existing width check. The docstring amendment about uncompiled `vmap` surfacing the host `ValueError` directly matches how `jax.debug.callback` executes outside jit.

**No blast radius on pinned measurement paths.** Repo-wide sweep: `sliced_sigreg_loss` / `epps_pulley_gaussian_statistic` / `sigreg_diagnostics` are referenced only by `core/__init__.py` + `alberta_framework/__init__.py` re-exports, the sigreg tests, and immutable archived source snapshots under quarantined `outputs/forager/` roots. Nothing under `alberta_framework/benchmarks/` calls SIGReg, so no registered screening arm's compiled graph or shard reproducibility is touched.

**Coverage quality worth calling out.** The suite covers the full failure surface, not just the headline case: eager inf embeddings against a zero direction coordinate (the 0·inf NaN), non-finite directions, NaN diagnostics input, jit-compiled rejection with valid-value preservation (`assert_trees_all_close` against the eager path), a vmapped non-finite lane, float64→float32 narrowing overflow, and — the sharpest one — **finite inputs whose projection overflows float32** (`finfo.max + finfo.max → inf`), which only the derived-projections check can catch. Each rejection test asserts against the exact message constant.

## Two non-gating observations

1. `sigreg_diagnostics` now validates `z`/`dirs`/`projected`, then its internal `sliced_sigreg_loss(flat, dirs, ...)` call re-validates the same arrays and re-checks the projections. Correct and idempotent, just redundant compiled ops on the diagnostics path; if that path is ever hot, the inner call could take pre-validated inputs. Fine as is for defense-in-depth.
2. In `sliced_sigreg_loss` the finite checks run after `_validated_kernel_width`, so a call with both an invalid width and non-finite embeddings reports the width error first — consistent with the function's existing error precedence, just worth knowing when reading test expectations.

## Scope note

Execution disclosure: I fully read the 183-line diff before executing anything, and ran it re-applied onto `main` in my own checkout rather than checking out the contributor branch. The change is a pure guard addition: no numerical rewrite of valid inputs (explicitly does not zero-fill NaNs), no API change, no `outputs/` contact.

— [kenjohnscreates]

Issue #188, opened 2026-08-15
Preregistration r3: is the utility gate load-bearing in the residual-trained RLS incumbent? (nonpromoting development measurement; approval requested before any run)


kenjohnscreates:
## Outcome

**Close (either direction lands in the ledger/theory record):** answer whether the utility gate is load-bearing in the standing residual-trained incumbent `rls_head_resid_l1_preset005`, via `main`'s frozen `gate_scale = 0` endpoint. The #52 and #185 closures each stated the scientific question remains open; this registration exists to answer it under an eligible protocol.

**Status: proposal only.** No run authorized or executed, no seed consumed, no artifact created, no code written or needed.

## Prior-attempt disclosure

- #52/#182 (closed): r1 ran without approval and with a source commit created after its shards; numbers disregarded.
- #185 (closed as not planned): my registration carried false or stale premises — an "untouched seeds 13–19" claim contradicted by the repository, a pre-#186 source with only v1 receipts and a drifting-head clause, ordering-masked (not disjoint) gates, and an unfrozen cost claim.

Every #185 correction is folded in below, and each factual premise here was re-verified against the repository before posting.

## Source binding — one exact post-#186 commit

- Pinned launch revision: **`2b9846a83f9301f50f8b939cc7e6fd0eb4854671`** (current `main`; contains #186's strict v2 provenance machinery and the #181 code-only implementation). This exact commit governs; no future-head clause. If `main` must move before approval (e.g. a required fix), I will amend this registration to a new exact commit **before** any approval is acted on — never silently.
- Candidate: `rls_head_resid_l1_preset005_nogate` — already registered at the pinned commit (verified: registry entry present at `2b9846a`), shared `_make_rls_head_learner` factory, frozen `gate_scale = 0.0` endpoint. Zero new code under this registration.
- Control: `rls_head_resid_l1_preset005` (identical spec, `gate_scale = 1.0`).
- Pre-run validation, executed from a clean checkout of the pinned commit and posted **before** any shard: `.venv/bin/python -m pytest tests/test_ipmnist_screening.py -q -o addopts="" -k "TestRLSHead"` (includes `main`'s own endpoint pins: `test_gate_scale_one_is_bitexact_legacy_incumbent`, `test_gate_off_two_step_plain_sgd_and_identical_head`, `test_gate_off_requires_residual_body`, `test_gate_scale_rejects_non_endpoint_values`). Any failure stops this registration.

## One change

[...]

PR #190, opened 2026-08-15
Harden ASI production contracts

lalalune:
## Summary

- establish the development-only L0 ASI reference-agent transaction contract, roadmap, documentation index, and repository anti-LARP audit while preserving the `alberta-framework` compatibility surface
- make recurring IPMNIST retention probes execute each registered arm's deployed frozen predictor, including the current RLS-head families, with shape, finite-value, and state-immutability checks
- harden evidence publication against overwrite/race conditions and align invalid-artifact CLI exit behavior
- keep OCI help usable from hardlink-installed wheels without weakening attested real operations, and make short Step 1/2 smoke horizons usable
- preserve the current 0.29 release metadata, locked CI/package lanes, public imports, upstream scientific safeguards, and immutable `outputs/` records

## Validation

- `377 passed` across the full IPMNIST screening, recurring-adapter, and provenance files
- `171 passed` in the final integrated changed-surface batch
- `17 passed` for the adversarial reference transaction contract, including forged foreign chains and unsupported dispatch rebinding
- `ruff check .` passed
- strict `mypy` passed for all 165 source files
- `uv lock --check` passed
- final wheel and sdist built successfully and passed `twine check`
- final prebuilt-distribution package suite: `4 passed`
- commit-range `gitleaks`: no leaks found
- `AGENTS.md` and `CLAUDE.md` are byte-identical; `git diff --check` passed

## Evidence boundary

This PR does not promote a scientific claim or select `reference-dev`. The live evidence registry remains fail-closed (`invalid`, exit 2, 0/5 currently valid) because registered source bytes changed. No pinned or development artifact under `outputs/` is modified.


kenjohnscreates (changes requested):
## Structured partial review — CHANGES_REQUESTED (one confirmed platform defect; everything I verified otherwise holds)

Scope statement up front: this is a +5291-line PR; I deep-reviewed the slices closest to the measured campaign surface (the `ipmnist_screening.py` sentinel-probe rework, the evidence-publication hardening, and the mechanical battery) and explicitly did **not** line-review the internals of `reference_agent.py` (1510 lines) or `prototype_reference_adapter.py` (617 lines) beyond isolation and their test suites. Verified by re-applying the diff onto merge-base `58b6780` in a scratch worktree — the resulting tree is **byte-identical to PR head `cc82d16`** — with a merge probe against current `main` (`d4d4657`) showing the only rebase conflict is `CHANGELOG.md` (trivial).

## Confirmed blocker: reference-protocol test is red on macOS arm64

`tests/test_reference_agent_protocol.py::test_outcome_rejects_scalar_that_overflows_during_protocol_conversion` fails deterministically on macOS arm64 (M1 Pro, numpy per `uv.lock` line):

```
huge = np.longdouble("1e4000")
assert np.isfinite(huge)   # AssertionError: np.longdouble('inf')
```

On aarch64, `np.longdouble` **is** float64 (there is no 80-bit extended type), so `"1e4000"` overflows to `inf` at parse (`RuntimeWarning: overflow encountered in conversion from string`) and the test's own precondition fails — the value the test needs (finite in longdouble, overflowing in float64) is unrepresentable on this platform. My run: **1 failed, 382 passed** across `test_ipmnist_screening.py`, `test_ipmnist_recurring_adapter.py`, `test_reference_agent_protocol.py`, `test_prototype_reference_adapter.py`, `test_evidence_manifest.py`, `test_console_entrypoints.py`. This matters more than a generic portability nit because the campaign authorization posture (#51, and the #188 protocol) names a **macOS arm64 runner** as the sanctioned execution platform — the full suite should be green there. Suggested minimal fixes: guard with `pytest.skip` when `np.finfo(np.longdouble).max <= np.finfo(np.float64).max`, or exercise the same conversion-overflow rejection with a platform-independent construction (e.g. Python `int(10**400)`, or `float.fromhex` boundary values), whichever matches the contract the production check enforces.

## What I verified and confirm

**The sentinel-probe rework does not touch the measured metric path.** The new `frozen_probe_logits` machinery executes only inside the recurring retention-probe harness; the `average_online_accuracy` path is untouched, and the full screening suite — including the bit-exact reduction pins — passes (all `test_ipmnist_screening.py` tests green in the run above).

**`_rls_head_frozen_probe_logits` exactly reproduces the deployed predictor.** Checked against `_make_rls_head_learner`: deployed logits are `state.wout.T @ _phi(params, x_norm)` with `_phi = [relu(relu(x@w1+b1)@w2+b2) * 1/sqrt(width+1), bias 1]`; the probe reproduces the chain, the `1/sqrt(h2.shape[1]+1)` scale, the bias augmentation, and the exclusion of the untrained `w3`/`b3` — matching the deleted refusal docstring's own description of the deployed model, for both `RLSHeadState` and `RLSHeadL2InitState`. Input normalization reuses the pre-existing `_ema_frozen_probe_input` frozen-EMA helper already established for the champion family. The `rff_rls`/`lin_rls`/`naive_bayes`/`nb_ensemble` probes similarly mirror their deployed forward passes (frozen projection + `wout`; streaming Gaussian logits; lock/mixture with optional RLS member).

**Probe hardening is fail-closed in the right direction.** The state-immutability check moved before snapshot construction and now raises (`a frozen sentinel probe must not mutate learner state`), logits shape is validated against `(n_probes, n_classes)`, and non-finite logits still refuse. Stamping `frozen_probe_strategy` (mode/module/qualname) into `_recurring_protocol_id` changes protocol ids for all arms — I checked `outputs/` and there are **no pinned recurring-probe artifacts**, so this is a safe content-addressed versioning change rather than an invalidation.

**Evidence-publication hardening.** `write_ia_evidence_artifact` and the multiagent `write_evidence_artifact` now use exclusive-create (`path.open("x")`) — overwrite becomes impossible at the syscall level. CLI exit alignment verified live in the applied tree: `evidence_manifest_cli` exits **2** with `overall_status: "invalid"`, `supported_scientific_claim_count: 0` — exactly the fail-closed posture the PR body discloses (registered source bytes change here, so the registry stays invalid until frozen protocols rerun; that is working-as-designed). The multiagent `_verify` now distinguishes missing (1) / unreadable-or-invalid (2) / rejected (1) / accepted (0), and the manifest CLI refuses both the pinned canonical output path and any pre-existing path.

**Isolation of the new L0 surface.** `reference_agent.py` and `prototype_reference_adapter.py` are imported nowhere under `alberta_framework/benchmarks/` or `alberta_framework/core/` — the development-only contract surface cannot perturb measured lanes. The OCI CLI change is parser extraction to `_official_foragax_oci_cli.build_parser` (code motion; console-entrypoint tests green).

**Mechanical battery.** `ruff check .` clean; strict repo-config mypy clean at **166 source files** (PR body says 165 — worth refreshing the number); `git diff --check` clean.

One non-gating observation: the PR body's "377 passed / 171 passed" counts don't decompose obviously against the current file set (my equivalent battery totals 383 collected); after the arm64 fix it may be simpler to restate a single full-suite count on the sanctioned platform.

Happy to re-verify promptly once the arm64 test fix is pushed — everything else I checked is solid, and the probe rework in particular is a real capability upgrade for the retention diagnostics over the fail-closed refusals it replaces.
Compute receipt: 2728955 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

lalalune:
Closing as stale and superseded. This 43-file/5k-line mega-PR conflicts with hundreds of subsequent changes; its reference-life work was already ported into the current development line and then advanced substantially. Rebasing this original risks overwriting newer protocol/evidence work, and its exact head also has a confirmed macOS arm64 np.longdouble test failure.

PR #191, opened 2026-08-15
Hold a zero behavior-model input gradient on inf observations

Kayvan-Zahiri:
## Summary
- `BehaviorModel.update` already holds previous finite weights when an inf observation would make softmax NaN and `logit_error * x` become 0*inf.
- The pre-update state-builder bridge `input_loss_gradient` still returned that NaN gradient. Callers that apply it before `update` would poison the representation even though the behavior model itself no-ops.
- Invalid or non-finite inputs now return a zero finite gradient, loss, and probabilities. Finite trajectories are unchanged.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_behavior_model.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/behavior_model.py tests/test_behavior_model.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rejected SIGReg samples, embeddings, directions, and derived projections that are
   non-finite in float32 arithmetic before they can become silent NaN losses, including
   under compiled and vectorized JAX calls.
+- Added an explicit traced validity verdict to behavior-model input gradients so finite neutral
+  payloads from rejected inputs cannot be mistaken for valid state-builder updates; eager,
+  compiled, and vectorized consumers can gate mutation on the same result field.
 - Required automatic feature-to-subtask counts to be non-negative integer scalars before
   ranking, while retaining NumPy/JAX integer-scalar compatibility and zero/oversized bounds.
 - Aligned plotted trailing-window learning-curve means and confidence intervals with the
--- a/alberta_framework/core/behavior_model.py
+++ b/alberta_framework/core/behavior_model.py
@@ -450,12 +450,36 @@ class BehaviorModel:
         loss = -jnp.sum(one_hot * jax.nn.log_softmax(scaled_logits))
         logit_gradient = (probabilities - one_hot) / cfg.temperature
         gradient = state.weights.T @ logit_gradient
+        # Inf observation makes softmax NaN and W.T @ (p - one_hot)
+        # non-finite, including 0*inf on a silent feature. The parameter
+        # update already no-ops; this pre-update bridge must not hand a
+        # NaN gradient to the state builder.
+        inputs_valid = (
+            jnp.all(jnp.isfinite(obs))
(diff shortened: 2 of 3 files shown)


renzo4web (approved):
Independent verification — the fix is correct and behavior-preserving on the finite path.

## Reproduction (isolated sandbox: Docker, `--network none`, read-only root, bounded CPU/memory)

| Check | pristine `main` (`58b6780`) | this branch |
|---|---|---|
| `input_loss_gradient` with `obs = [0.0, inf]` | logits/probabilities/loss/gradient/norm **all NaN** | all finite; loss `0.0`, gradient zeros |
| `pytest tests/test_behavior_model.py` | (bug unfixed) | **21 passed** |
| `ruff check .` | — | clean |
| `mypy` (repo config) | — | clean, 163 files |

RED confirmed on pristine `main` with the PR's own scenario; GREEN confirmed here.

## Finite-path invariance (the #43 lesson, checked)

The guard must be inert when nothing is non-finite. Deterministic probe: 100 finite observations through the jitted `input_loss_gradient`, hashing every output field:

```text
main   finite-path digest: 1234e5d1e10fef3527303d97a724d31b4c31c959d578bbb3a064c88025c5b7a4
branch finite-path digest: 1234e5d1e10fef3527303d97a724d31b4c31c959d578bbb3a064c88025c5b7a4
```

Byte-identical — no reassociation or masking drift on the unpoisoned path.

One design note, not a blocker: the guard computes `valid` from `isfinite` on **source state + inputs + proposed outputs**, then masks. That means a NaN produced *inside* this function (e.g. a genuine softmax overflow at extreme-but-finite logits) also zeroes the result rather than propagating — which is the fail-safe direction for a pre-update bridge, but worth knowing: this bridge now never reports a non-finite diagnostic, so if a caller ever needed to distinguish "inf observation" from "numerically diverged finite observation", the distinction is lost here. The PR body's framing ("Inf next-value diagnostics are not reported as if they were a real bootstrap") matches this. Fine as designed.

Verification probe (finite-path invariance): <details><summary>pr191_bitexact.py</summary>

```python
import hashlib
import jax
import jax.numpy as jnp
from alberta_framework.core.behavior_model import BehaviorModel, BehaviorModelConfig

model = BehaviorModel(BehaviorModelConfig(n_actions=3, step_size=0.05))
state = model.init(feature_dim=4, key=jax.random.key(42))
fn = jax.jit(model.input_loss_gradient)
h = hashlib.sha256()
for i in range(100):
    obs = jnp.asarray([0.3 * ((i % 5) - 2), -0.7, 1.1 * (i % 3), 0.05 * i], dtype=jnp.float32)
    out = fn(state, obs, jnp.asarray(i % 3, dtype=jnp.int32))
    for leaf in (out.logits, out.probabilities, jnp.reshape(out.loss, (1,)),
                 out.gradient, jnp.reshape(out.gradient_norm, (1,))):
        h.update(jnp.asarray(leaf, dtype=jnp.float32).tobytes())
print("finite-path digest:", h.hexdigest())
```

</details>

AI-assisted review (agent tooling; no Slop receipt applies to review work).

PR #193, opened 2026-08-15
Do not form 0*inf latent-action interactions

Kayvan-Zahiri:
## Summary
- `input_features_from_latent` multiplied latent coordinates by a silent action one-hot, so an inf latent became 0*inf = NaN in the interaction slice.
- Both factors are zeroed before that product. Invalid actions and non-finite latents still return an all-NaN feature vector, matching the existing invalid-action reject.
- Finite interaction paths are unchanged.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_latent_world_model.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/latent_world_model.py tests/test_latent_world_model.py`

### DIFF
--- a/alberta_framework/core/latent_world_model.py
+++ b/alberta_framework/core/latent_world_model.py
@@ -418,13 +418,21 @@ class LatentWorldModel:
             self._config.n_actions,
             dtype=jnp.float32,
         )
-        features = [z, action_one_hot]
+        latent_valid = jnp.all(jnp.isfinite(z))
+        features_valid = action_valid & latent_valid
+        # Inf latent times a silent one-hot is 0*inf = NaN. Zero both
+        # factors before the product; keep the invalid-action NaN reject.
+        safe_z = jnp.where(features_valid, z, jnp.zeros_like(z))
+        safe_action_one_hot = jnp.where(
+            features_valid, action_one_hot, jnp.zeros_like(action_one_hot)
+        )
+        features = [safe_z, safe_action_one_hot]
         if self._config.include_action_interactions:
-            interactions = (z[:, None] * action_one_hot[None, :]).reshape((-1,))
+            interactions = (safe_z[:, None] * safe_action_one_hot[None, :]).reshape((-1,))
             features.append(interactions)
         encoded = jnp.concatenate(features, axis=0)
         return jnp.where(
-            action_valid,
+            features_valid,
             encoded,
             jnp.full_like(encoded, jnp.nan),
         )
--- a/tests/test_latent_world_model.py
+++ b/tests/test_latent_world_model.py
@@ -137,6 +137,25 @@ def test_latent_world_model_scan_loop_and_config_roundtrip() -> None:
     chex.assert_tree_all_finite(result.prediction_errors)
 
 
+def test_latent_action_interactions_do_not_form_zero_times_inf() -> None:
+    """Inf latent times a silent action one-hot is 0*inf = NaN in the product."""
+    model = LatentWorldModel(
+        LatentWorldModelConfig(
+            observation_dim=2,
+            n_actions=3,
+            latent_dim=2,
+            hidden_sizes=(),
+            include_action_interactions=True,
+        )
+    )
+    latent = jnp.array([jnp.inf, 1.0], dtype=jnp.float32)
+    raw = latent[:, None] * jnp.array([1.0, 0.0, 0.0], dtype=jnp.float32)[None, :]
+    assert not bool(jnp.all(jnp.isfinite(raw)))
+
+    features = model.input_features_from_latent(latent, jnp.array(0, dtype=jnp.int32))
+    assert bool(jnp.all(jnp.isnan(features)))
+
+
 def test_score_dream_candidates_selects_surprising_useful_valid_items() -> None:
     result = score_dream_candidates(
         surprises=jnp.array([0.1, 0.9, 0.7, 0.3], dtype=jnp.float32),


swissyai (approved):
Verified review of head `3cc9f8f784b6b3ff897350593d22ab7efee7b9b5` (the only commit on this branch at review time). I read every line of the diff on a trusted checkout of `main` and never checked out the branch; all execution of patched code happened in an isolated container.

## What the change does

`LatentWorldModel.input_features_from_latent` builds `concat(z, one_hot(action), z ⊗ one_hot)`. On unpatched `main`, a non-finite latent coordinate hits the silent entries of the one-hot in the interaction product, so `inf * 0 = NaN` lands in the interaction slice while the latent slice still carries the raw `inf` — a mixed inf/NaN vector that passes through whenever the action is valid. The patch computes `features_valid = action_valid & all(isfinite(z))`, zeroes **both** factors before the product, and reuses the existing final `jnp.where` to return an all-NaN vector when invalid.

## The two questions this family keeps getting wrong

**Does the guard mask genuine NaNs? No.** A non-finite latent now produces an **all-NaN** feature vector — the same explicit reject the function already used for invalid actions — not a silently-zeroed "clean" vector. The zeroed `safe_z`/`safe_action_one_hot` values can never leak into output: they are only selected when `features_valid` is true, which requires `z` finite. I also traced the gradient consumer: `_update_encoder`'s `latent_prediction_loss` differentiates through this function, and its fail-closed `grads_finite` gate (`core/latent_world_model.py:572`) still triggers — a non-finite latent yields NaN features → NaN loss → NaN grads → update skipped, exactly as before. The double-`where` structure additionally keeps the not-taken branch out of gradient cotangents, which is the standard JAX pattern here.

**Are valid-input values unchanged? Yes, bit-for-bit.** When `features_valid` is true, `safe_z` is `z` and `safe_action_one_hot` is `one_hot`, so both the concatenated slices and the interaction products are the identical expressions the old code computed; the final `where` then selects `encoded` as before. The 17 pre-existing tests in the file pass unchanged (below).

One deliberate behavioral tightening worth naming: with `include_action_interactions=False`, an inf latent previously passed through as inf in the latent slice; now it is rejected all-NaN. That is consistent with the module's reject convention and with `safe_discrete_action`'s treatment of non-finite actions, and downstream consumers treat inf-contaminated and NaN vectors identically (non-finite → fail-closed skip), so nothing that previously "worked" changes.

## Red/green proof (isolated container)

The container has **no network**, and it is **Linux arm64** while this repository's other measurements are macOS arm64 — valid for pass/fail proofs, but no platform parity is implied. The mounted trusted repo is read-only; the harness copies it to a private writable copy inside the guest and applies the diff there.

**RED** — only the `tests/` hunks applied onto unpatched source; the new test must fail and does:

```text
MODE: RED (tests-only hunks applied to unpatched source)
jax 0.11.0 [CpuDevice(id=0)]
FAILED tests/test_latent_world_model.py::test_latent_action_interactions_do_not_form_zero_times_inf
1 failed, 16 passed in 9.58s
--- isolation proof ---
host /Users not visible
network unreachable
```

**GREEN** — full diff applied; whole file passes:

```text
MODE: GREEN (full diff applied)
jax 0.11.0 [CpuDevice(id=0)]
17 passed in 9.75s
--- isolation proof ---
host /Users not visible
network unreachable
```

Both runs printed their own isolation evidence, quoted verbatim above.

## Lint

`ruff check` on the two touched files (diff applied to trusted `main` on the host, tree restored afterward): `All checks passed!`

## Notes

- The new test first proves the raw product is non-finite, then asserts the guarded path returns all-NaN — so it pins the reject semantics, not just "no crash".
- Nit, non-blocking: the PR body lacks an exact provider/model disclosure (the repo's live report flags this); worth adding for provenance.

Verdict: correct, minimal, and consistent with the sibling 0*inf guards. LGTM.

---

Compute receipt: 2871691 project-attributed tokens (exact; device-signed, locally reported)
Contribution skill revision: elizaOS/slopdotcash@c265ca248e63e4c5f8c7bb5b68561ee210b64acd:skills/contribute-to-asi
— [swissyai]

renzo4web (approved):
Independent verification of the PR #193 scenario — RED confirmed on pristine `main`, GREEN confirmed on this branch, finite path byte-identical.

## Reproduction

| Check | pristine `main` (`58b6780`) | this branch (`3cc9f8f`) |
|---|---|---|
| `input_features_from_latent(latent=[inf, 1.0], a=0)`, `include_action_interactions=True` | interaction slice contains `0*inf = NaN` (only the invalid-action NaN reject existed) | all-NaN feature vector via `features_valid = action_valid & latent_valid`, both factors zeroed before the product |
| `pytest tests/test_latent_world_model.py` | (bug unfixed) | **17 passed** |
| `ruff check .` | — | clean |
| `mypy` (repo config) | — | clean, 163 files |

RED probe run against a pristine `main` worktree with the repo `.venv` (jax 0.11.0); GREEN run in an isolated sandbox (Docker `python:3.12-slim`, `--network none`, read-only root, `--cpus 2 --memory 2g --pids-limit 128`, `pip install -e ".[dev]"`), overlaying exactly this branch's `core/latent_world_model.py` + `tests/test_latent_world_model.py` onto the `58b6780` tree (branch fork-point `e132879` touches only sigreg, so the overlay is exact).

## Finite-path bit-exactness (the #43 lesson, checked)

Deterministic probe: 100 finite latents/actions through `input_features_from_latent` and `predict_from_latent`, hashing every output leaf:

```text
main   finite-path digest: 4a9ef18eae2a20d82f70a6dc92ef0c944ff56cf8e322578e03688d0b09470719
branch finite-path digest: 4a9ef18eae2a20d82f70a6dc92ef0c944ff56cf8e322578e03688d0b09470719
```

Byte-identical across two runs per side; cross-module probes (world model, statistics, prototype memory, associative memory) also match `main` digests under this overlay, so the change is inert outside `latent_world_model`.

Design note (not a blocker): the guard composes with the pre-existing invalid-action NaN reject — a non-finite latent now also yields an all-NaN feature vector, matching the invalid-action semantics that already shipped. Fail-visible direction preserved.

Probe: <details><summary>finite-path digest probe (PR193 section)</summary>

```python
import hashlib
import jax
import jax.numpy as jnp
from alberta_framework.core.latent_world_model import (
    LatentWorldModel,
    LatentWorldModelConfig,
)

m = LatentWorldModel(
    LatentWorldModelConfig(
        observation_dim=2, n_actions=3, latent_dim=2, hidden_sizes=(),
        include_action_interactions=True,
    )
)
state = m.init(jax.random.key(11))
h = hashlib.sha256()
for i in range(100):
    latent = jnp.asarray(
        [0.4 * ((i % 7) - 3) + 0.005 * i, 1.3 - 0.03 * (i % 11)], dtype=jnp.float32
    )
    act = jnp.asarray(i % 3, dtype=jnp.int32)
    feats = m.input_features_from_latent(latent, act)
    pred = m.predict_from_latent(state, latent, act)
    for leaf in (feats, pred.next_latent, pred.reward, pred.discount,
                 pred.raw_predictions):
        h.update(jnp.asarray(leaf, dtype=jnp.float32).tobytes())
print("finite-path digest:", h.hexdigest())
```

</details>

AI-assisted review (provider: zai/glm-5.3 via opencode-go; no Slop receipt applies to review work).

slop-review {"schemaVersion":"1","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/193","headSha":"3cc9f8f784b6b3ff897350593d22ab7efee7b9b5","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","usefulArtifacts":[],"commands":["git worktree add /tmp/revA-main 58b6780",".venv/bin/python /tmp/revA-meta/red_probes.py","docker run --rm --network none --read-only --tmpfs /tmp:rw,size=512m --cpus 2 --memory 2g --pids-limit 128 reva-verify:pr193 bash /verify_pr.sh 193 core/latent_world_model.py tests/test_latent_world_model.py","python /probe_all.py PR193"],"evidenceUrls":["https://github.com/elizaOS/asi/pull/193"],"summary":"RED reproduced on pristine main 58b6780 (inf latent times silent one-hot gives 0*inf=NaN in the interaction slice); GREEN in a no-network Docker sandbox: 17/17 tests pass, ruff and mypy clean, and a 100-input deterministic finite-path digest over input_features_from_latent+predict_from_latent output leaves is byte-identical between main and branch (4a9ef18e...). Invalid inputs stay fail-visible; no evidence-manifest, pinned-output, threshold, or validator files touched."}

PR #195, opened 2026-08-15
Keep prototype-memory softmax finite on inf observations

Kayvan-Zahiri:
## Summary
- `class_logits` already returns NaN when the observation is non-finite. `predict` still ran softmax on those logits, so all-`-inf` distances became inf-inf NaN probabilities.
- Non-finite logits now fall back to the empty-memory uniform simplex. Infinite `bandwidth` is rejected at learner construction.
- Finite classification fixtures are unchanged.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_prototype_memory.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/prototype_memory.py tests/test_prototype_memory.py`

### DIFF
--- a/alberta_framework/core/prototype_memory.py
+++ b/alberta_framework/core/prototype_memory.py
@@ -18,6 +18,7 @@ the packaged Step 2 retained-view memory
 from __future__ import annotations
 
 import functools
+import math
 from typing import Any, cast
 
 import chex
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after maintainer review: this changes an invalid public prediction into a valid-looking uniform simplex, which contradicts the explicit fail-visible transaction contract and is reproduced by the current failing test. The finite configuration validation is useful, but it should be relanded separately without disguising corrupted active state as a valid prediction.

Kayvan-Zahiri:
Relanded the config-only part as https://github.com/elizaOS/asi/pull/528. Construction now rejects non-finite bandwidth / novelty_threshold / update_rate. Inf observations still produce a non-finite public prediction (no uniform-simplex fallback).

PR #197, opened 2026-08-15
Reject non-finite means before crowning a bar-plot winner

Kayvan-Zahiri:
## Summary
- `plot_final_performance_bars` used `numpy.argmin` on method means. A NaN mean is treated as the minimum, so a failed run was drawn as the best method and given the gold outline.
- Empty result maps and non-finite means or stds now raise `ValueError` before any bar is marked.
- Finite publication fixtures are unchanged.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_utils_smoke.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/utils/visualization.py tests/test_utils_smoke.py`

### DIFF
--- a/alberta_framework/utils/visualization.py
+++ b/alberta_framework/utils/visualization.py
@@ -208,10 +208,6 @@ def plot_final_performance_bars(
     except ImportError:
         raise ImportError("matplotlib is required. Install with: pip install matplotlib")
 
-    if ax is None:
-        fig, ax = plt.subplots()
-    else:
-        fig = cast("Figure", ax.figure)
     names = list(results.keys())
     if not names:
         raise ValueError("plot_final_performance_bars requires at least one result")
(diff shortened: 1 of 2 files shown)


nishimweprince (changes requested):
Independent review of `ff93f103dd34e862b974029a53a6bdd1a7a0a8be` against base `58b6780840e1ba3fd6fdaea4d49c906d6c3b55cb`: **changes requested**.

Blocking finding in `alberta_framework/utils/visualization.py`, `plot_final_performance_bars`: validation currently runs after `plt.subplots()` when the caller does not supply an axis. An invalid empty/non-finite result therefore raises as intended but leaks the newly-created Matplotlib figure. On the exact head, an isolated probe recorded the expected `ValueError` and `new_figures [1]`. Repeated rejected calls can accumulate figures and memory.

Please compute and validate `names`, `means`, and `stds` before creating the fallback figure, then add a regression asserting that `plt.get_fignums()` is unchanged after rejected empty and non-finite mean/std inputs. The current added test covers only a non-finite mean, so it does not detect this resource leak or directly exercise the other claimed rejection branches.

Other verification, performed on an archive of the exact head under macOS `sandbox-exec` with network denied and the control checkout unreadable:

- Failing-test-first check (base production plus PR tests): `1 failed, 12 passed`; the new non-finite-mean test failed to raise.
- Exact-head green check: `tests/test_utils_smoke.py` — `13 passed`.
- Ruff on the changed production and test files — passed.
- Mypy on `alberta_framework/utils/visualization.py` — passed with cache disabled.

Client / agent tooling: codex
— [review-pr-197]

PR #213, opened 2026-08-15
Skip optimizer 0*inf trace decay and zero-gamma next features

Kayvan-Zahiri:
## Summary
- ObGD and Autostep-for-GTD decay traces with a scale that is 0 in the supervised default. An infinite stored trace was IEEE NaN and froze the optimizer.
- TD-IDBD and AutoTDIDBD also form `gamma * phi(s')` and `gamma * lambda * z`. `jnp.where` still evaluates those products, so gamma 0 times an infinite next observation or trace is NaN.
- Skip the products when the scale is 0. A next observation may be non-finite only when gamma is already 0.

## Test plan
- [x] ObGD and Autostep-GTD inf-trace tests
- [x] TDIDBD / AutoTDIDBD zero-gamma inf-next-obs tests
- [x] `TestObGD`, `TestAutostepGTDLambda`, `tests/test_td_optimizers.py` (34 passed)

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -1449,7 +1449,7 @@ class AutostepGTDLambda(Optimizer[AutostepGTDLambdaState]):
         )
         previous_checked = state
         if self._trace_decay == 0.0:
-            previous_checked = state.replace(
+            previous_checked = state.replace(  # type: ignore[attr-defined]
                 eligibility_traces=jnp.zeros_like(state.eligibility_traces),
                 bias_eligibility_trace=jnp.zeros_like(state.bias_eligibility_trace),
             )
@@ -1627,7 +1627,7 @@ class ObGD(Optimizer[ObGDState]):
 
         previous_checked = state
         if self._gamma == 0.0 or self._lamda == 0.0:
-            previous_checked = state.replace(
+            previous_checked = state.replace(  # type: ignore[attr-defined]
                 traces=jnp.zeros_like(state.traces),
                 bias_trace=jnp.zeros_like(state.bias_trace),
             )
@@ -1902,7 +1902,7 @@ class TDIDBD(TDOptimizer[TDIDBDState]):
             & (jnp.all(jnp.isfinite(next_observation)) | (gamma_scalar == 0.0))
             & jnp.isfinite(gamma_scalar)
         )
-        previous_checked = state.replace(
+        previous_checked = state.replace(  # type: ignore[attr-defined]
             eligibility_traces=jnp.where(
                 (gamma_scalar == 0.0) | (lam == 0.0),
                 jnp.zeros_like(state.eligibility_traces),
@@ -2137,7 +2137,7 @@ class AutoTDIDBD(TDOptimizer[AutoTDIDBDState]):
             & (jnp.all(jnp.isfinite(next_observation)) | (gamma_scalar == 0.0))
             & jnp.isfinite(gamma_scalar)
         )
-        previous_checked = state.replace(
+        previous_checked = state.replace(  # type: ignore[attr-defined]
             eligibility_traces=jnp.where(
                 (gamma_scalar == 0.0) | (lam == 0.0),
                 jnp.zeros_like(state.eligibility_traces),
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -40,6 +40,11 @@ from alberta_framework.core.update_safety import (
     zero_if_collapsed_infinity,
 )
 
+
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Return 0 when ``scale`` is 0 so IEEE ``0 * inf`` does not become NaN."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
 # =============================================================================
 # Bounder ABC
 # =============================================================================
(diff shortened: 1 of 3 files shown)


kenjohnscreates (changes requested):
**CHANGES_REQUESTED — one blocking item repeated at four sites (the repo's strict mypy gate goes red), everything else verified positively.** Independently re-applied and verified on current `main` (`58b6780`), arm64 M1 Pro, python 3.13, jax 0.11.0 CPU; applied tree byte-identical to PR head `83b85ae`, which contains current `main`.

## Blocking: strict mypy regression ×4

```
alberta_framework/core/optimizers.py:1452: error: "AutostepGTDLambdaState" has no attribute "replace"  [attr-defined]
alberta_framework/core/optimizers.py:1630: error: "ObGDState" has no attribute "replace"  [attr-defined]
alberta_framework/core/optimizers.py:1905: error: "TDIDBDState" has no attribute "replace"  [attr-defined]
alberta_framework/core/optimizers.py:2140: error: "AutoTDIDBDState" has no attribute "replace"  [attr-defined]
```

Baseline `main` is clean (163 source files). Runtime works — these are `chex.dataclass` states and `.replace()` exists when executed (all 75 tests pass) — but the gate is red. Identical defect class to my #219 and #218 reviews; the house-precedent one-liner is `state.replace(  # type: ignore[attr-defined]` (cf. `continual_backprop.py`). A repo-config `mypy` run before posting would catch this across the entire wave.

## Verified positively

**Failing-test-first, reproduced.** Tests-only on `main`: exactly the five new tests fail (AutostepGTDLambda, ObGD, TDIDBD ×2, AutoTDIDBD). Full diff → **75/75 pass** across both test files (9.2s). `ruff check` clean.

**The `jnp.where` idiom is correct at every arithmetic site in this module** — the scales are genuinely runtime values (`state.trace_decay`, `state.gamma`, `state.lamda` are state fields; `gamma_scalar` is a per-call argument), so no static hoist is possible and my #219 graph-purity request does not transfer here. The author's mixed treatment is actually right: the `previous_checked` relaxations in AutostepGTDLambda and ObGD branch on static `self._trace_decay` / `self._gamma` / `self._lamda` at trace time (zero graph impact for nonzero-decay configs), while the TDIDBD/AutoTDIDBD versions must use runtime `jnp.where` because their γ arrives per call.

**The γ=0 next-observation handling is numerically sound, not just contractual.** Unlike SwiftTD (#218), `next_observation` here is genuinely used (`feature_diff = γ·φ(s′) − φ(s)` on the ordinary-gradient paths), so the skip does real work: at γ=0 the term becomes exactly `−φ(s)` and an infinite φ(s′) never enters arithmetic. The relaxed `inputs_valid` (`finite(next_obs) | γ==0`) is consistent with that: every quantity the update commits is still gated by the downstream finite checks on `weight_delta`/`meta_delta` and the candidate state.

**Recovery semantics are fail-closed.** The `previous_checked` relaxations only admit states whose non-finite trace fields this transition provably overwrites (skip yields `observation` / `1.0` exactly — the tests pin those values); anything that leaks into a committed field or delta still blocks via the unchanged candidate-side finite checks.

**No live measured-path exposure.** `ipmnist_screening.py`'s only reference to this module is a docstring cross-link (`:class:` role, line 454) — the screening IDBD arms are inline reimplementations. The forager benchmarks import `Autostep` and `ObGDBounding`, neither of which is among the four changed classes. Repo-wide, no benchmark instantiates `AutostepGTDLambda`, `ObGD`, `TDIDBD`, or `AutoTDIDBD`, so no pinned artifact's compiled graph is touched.

## One non-blocking observation

`previous_checked` in AutostepGTDLambda/ObGD branches on the static constructor value (`self._trace_decay`) while the trace arithmetic uses the state field (`state.trace_decay`). For states produced by `init` these agree; a hand-constructed state with a mismatched `trace_decay` would be checked against the constructor's value but decayed by the state's. Not reachable through the public API — worth a comment at most.

Happy to re-verify promptly once the four-annotation fix is pushed — everything else already checks out.

Execution disclosure: I fully read the 332-line diff before executing anything, and ran it re-applied onto `main` in my own checkout rather than checking out the contributor branch.

— [kenjohnscreates]

kenjohnscreates (approved):
## Re-verification after `5d29f0b6` ("Satisfy strict typing for optimizer recovery") — APPROVE

The single blocker from my previous review is resolved. The follow-up is exactly the house-precedent fix and nothing else: `# type: ignore[attr-defined]` added to the four chex-dataclass `previous_checked = state.replace(` sites I flagged (old lines 1452/1630/1905/2140), 4 insertions / 4 deletions in `core/optimizers.py`, zero behavior change.

**Verification** (diff re-applied onto current `main` `8b49cec` in a scratch worktree; contributor branch never checked out):

- Touched files **byte-identical to PR head `5d29f0b6`** (`git diff` against `pull/213/head` restricted to `core/optimizers.py` + `tests/test_optimizers.py` is empty).
- `core/optimizers.py` has no other commits on `main` since my previous review, so the baseline is the same source bytes.
- Repo-config strict mypy: **clean at 163 source files** (was red ×4).
- `tests/test_optimizers.py` **57/57** with the full diff; `ruff check .` clean.
- Failing-test-first, measured precisely on today's baseline: tests-only → **2 deterministic failures**, `TestAutostepGTDLambda::test_zero_trace_decay_does_not_multiply_inf_eligibility` and `TestObGD::test_zero_trace_decay_does_not_multiply_inf_traces` — these are the genuinely discriminating tests for the 0*inf defect. The other three new tests (`TestAdaptiveObGDBounding` ×2, `TestAutoTDIDBD::test_zero_gamma_...`) pass on baseline: they pin behavior the existing fail-closed backstops already produce, so their role is regression protection rather than red proof. Worth knowing, not worth changing.

Positive findings from my previous review carry over unchanged: the runtime `jnp.where` idiom is correct in this module (the guarded scales are runtime-traced state, unlike #219's static config floats), the γ=0 skip is numerically sound (`feature_diff` genuinely consumes `next_obs` here), recovery is fail-closed, and there is no live measured-path exposure — the screening reference at `ipmnist_screening.py:454` is a docstring `:class:` cross-link, not an import, and the forager-imported `Autostep`/`ObGDBounding` classes are untouched by this diff.
Compute receipt: 767528 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

PR #215, opened 2026-08-15
Skip resource-manager 0*inf preference decay and keep softmax finite

Kayvan-Zahiri:
## Summary

- Skip IEEE `0 * inf` only for genuinely zero-scaled preference and diagnostic history in both `LearnedResourceManager` and `GeneratorMetaResourceManager`.
- Preserve the established multiplication and softmax results for finite state; non-finite logits are repaired only when the configured preference discount is exactly zero.
- Relax source-state validation only for selected-context zero-discount logits and observed EMA slots that a zero decay replaces. Masked slots retain finite history, while masked poison and every nonzero-scaled poisoned value reject atomically.
- Add symmetric regressions for zero-discount logits, zero loss/reward EMA decay, finite-path equivalence, and fail-closed consumed state.
- Rebased onto `main` at `b775741a0ed09bd216e55a1361171660ad4e16db`.

## Test plan

- [x] `tests/test_resource_manager.py tests/test_late_wave_transactions.py tests/test_compositional_features.py` — 110 passed
- [x] Post-final-rebase `tests/test_resource_manager.py` — 24 passed
- [x] Repository-wide Ruff
- [x] Strict mypy — 163 source files
- [x] `git diff --check`
- [x] No `outputs/**` changes

Original contribution made with Cursor. Maintainer repair performed with Codex; no measured-run receipt was required for this numerical-safety patch.

### DIFF
--- a/alberta_framework/core/resource_manager.py
+++ b/alberta_framework/core/resource_manager.py
@@ -47,6 +47,33 @@ from alberta_framework.core.update_safety import (
 )
 
 
+def _skip_zero_scale(scale: float, value: Array) -> Array:
+    """Keep finite multiplication exact while repairing zero-scaled poison."""
+    product = scale * value
+    if scale != 0.0:
+        return product
+    return jnp.where(jnp.isfinite(value), product, jnp.zeros_like(value))
+
+
+def _recover_nonfinite_at_zero_scale(scale: float, value: Array) -> Array:
+    """Repair only non-finite history that a zero scale is configured to forget."""
+    if scale != 0.0:
+        return value
+    return jnp.where(~jnp.isfinite(value), jnp.zeros_like(value), value)
+
+
+def _mask_unused_history(
+    history: Array,
+    context: Array,
+    unused: Array,
+) -> Array:
+    """Zero only history entries that the candidate update does not consume."""
+    current = history[context]
+    recoverable = unused & ~jnp.isfinite(current)
+    checked = jnp.where(recoverable, jnp.zeros_like(current), current)
+    return history.at[context].set(checked)
+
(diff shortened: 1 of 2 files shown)


shabbydev (approved):

This module is two independent classes (`LearnedResourceManager` at :177, `GeneratorMetaResourceManager` at :466) carrying the same defect shape — discounted log-weights plus a loss/reward EMA — and the PR fixes both symmetrically: `_skip_zero_scale` on the discount multiply and the EMA multiply, per-decay `previous_checked` zeroing for the source-side finiteness gate, and `_finite_softmax` on both `weights()` paths. I confirmed those are the only two `jax.nn.softmax` call sites in the file, so the finite-softmax coverage is complete.

The `_finite_softmax` fallback (uniform on any non-finite logit) closes the last consumer-side hole: with `discount = 0` the guarded multiply yields finite logits, but a poisoned log-weight vector could still reach softmax from an old state that pre-dates the fix or a non-zero decay path — softmax(inf) = NaN weights → NaN allocation downstream. Uniform-on-non-finite is the right neutral fallback for an allocator (equal split preserves total mass).

Correctness detail done right: the additive-constant recentering (`-= mean`) happens AFTER the guarded multiply, so with discount 0 the result is `lr * (adv - mean(adv))` — finite whenever advantages are finite, and `advantages` validity is already gated upstream.

Non-blocking note: with discount==0 the policy becomes a pure bandit over instantaneous advantages — no history. That's the literal reading of the config and matches the series semantics, just worth being aware the manager forgets everything every step on that path.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

PR #216, opened 2026-08-15
Skip compositional-feature 0*inf utility and score-trace decay

Kayvan-Zahiri:
## Summary

Compositional feature discovery has several decay paths where a legitimate zero decay must not form IEEE `0 * inf`. This repair specializes only the exact zero-decay configurations, preserving the original arithmetic graph for every nonzero decay. That distinction is load-bearing: the earlier generic `where` helper changed finite default-path last bits and failed the repository's pinned bitwise digest.

The zero-decay paths now recover safely for utilities, candidate utilities, task activity, residual/energy traces, and novelty traces. Nonzero paths retain their historical equations exactly.

## Verification

Validated at `51ec2fd9f11ea522c53973bb5cde2e6ae0a8901f`, rebased on current `main` after PR #268:

- `tests/test_compositional_features.py` + `tests/test_optimizer_nonfinite_transactions.py`: **111 passed**;
- the pinned finite one-step digest regression passes;
- targeted Ruff: clean;
- strict mypy on `compositional_features.py`: clean;
- `git diff --check`: clean.

No `outputs/**` artifact or evidence protocol changed.

## Contribution provenance

Original report and fix: Kayvan-Zahiri. Additional adversarial finite-path diagnosis, static zero-decay specialization, rebase, and verification: lalalune/Codex.

### DIFF
--- a/alberta_framework/core/compositional_features.py
+++ b/alberta_framework/core/compositional_features.py
@@ -2526,16 +2526,16 @@ class CompositionalFeatureLearner:
         """Perform one temporally-uniform compositional-feature update."""
         previous_checked = state
         if self._utility_decay == 0.0:
-            previous_checked = previous_checked.replace(
+            previous_checked = previous_checked.replace(  # type: ignore[attr-defined]
                 utilities=jnp.zeros_like(state.utilities),
                 candidate_utilities=jnp.zeros_like(state.candidate_utilities),
             )
         if self._future_utility_task_activity_decay == 0.0:
-            previous_checked = previous_checked.replace(
+            previous_checked = previous_checked.replace(  # type: ignore[attr-defined]
                 task_activity_ema=jnp.zeros_like(state.task_activity_ema),
             )
         if self._candidate_score_trace_decay == 0.0:
-            previous_checked = previous_checked.replace(
+            previous_checked = previous_checked.replace(  # type: ignore[attr-defined]
                 feature_score_residual_trace=jnp.zeros_like(
                     state.feature_score_residual_trace
                 ),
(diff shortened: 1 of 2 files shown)


shabbydev (approved):

The one thing I wanted to chase down before approving: `_retention_slow_utility` keeps its pre-existing early return of `previous_slow_utility` (possibly inf) at decay 0 — different "disabled" semantics from the instantaneous tracking the EMA guards produce. I verified it's safe: `_retention_score` also short-circuits at decay 0 (`return fast_utility`), so the frozen slow utility is never consumed on the disabled path, and the config is frozen at construction so the path can't toggle mid-run. Also worth noting: `retention_slow_utility_decay` DEFAULTS to 0.0 (opt-in), so the early-return path — not the guarded multiply — is the default behavior; the added `_skip_zero_scale` there is unreachable-when-it-matters belt-and-braces, which is fine.

The real substance is the `previous_checked` lattice in `update()`: three independent decay families (`utility_decay` → utilities + candidate_utilities; `future_utility_task_activity_decay` → task_activity_ema; `candidate_score_trace_decay` → the five score/correlation traces) each get their own conditional zeroing, so a poisoned tracker only stops gating when ITS decay is disabled — a poisoned score trace with an enabled `candidate_score_trace_decay` still fails closed. That per-family granularity is the correct shape and the easy thing to get wrong by collapsing into one flag.

Update arithmetic is guarded at both utility sites (active + candidate) and all three trace sites. Test pins the zero-decay-with-inf-utilities case end to end through `update_applied`.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

PR #217, opened 2026-08-15
Skip CBP 0*inf utility EMA decay

Kayvan-Zahiri:
## Summary
- Continual Backprop already holds a previous utility when the contribution is non-finite. It still formed `decay * utility` when decay is 0, so an infinite stored EMA became NaN and was committed.
- Skip the product when decay is 0 so the tracker becomes the current finite contribution.

## Test plan
- [x] `test_zero_decay_does_not_multiply_inf_utility`
- [x] `test_inf_activation_silent_grad_holds_finite_utility`
- [x] `tests/test_continual_backprop.py` (21 passed)

### DIFF
--- a/alberta_framework/core/continual_backprop.py
+++ b/alberta_framework/core/continual_backprop.py
@@ -287,7 +287,12 @@ def update_utility(
         # previous finite utility so replacement does not treat NaN as
         # the lowest-utility unit.
         contribution_finite = jnp.isfinite(contribution)
-        u_new = decay * cbp_state.utilities[i] + (1.0 - decay) * contribution
+        decayed = jnp.where(
+            decay == 0.0,
+            jnp.zeros_like(cbp_state.utilities[i]),
+            decay * cbp_state.utilities[i],
+        )
+        u_new = decayed + (1.0 - decay) * contribution
         u_new = jnp.where(contribution_finite, u_new, cbp_state.utilities[i])
         new_utilities.append(u_new)
         new_ages.append(cbp_state.ages[i] + 1)
--- a/tests/test_continual_backprop.py
+++ b/tests/test_continual_backprop.py
@@ -151,6 +151,24 @@ class TestUtilityUpdate:
         assert bool(has)
         assert int(idx) == 1
 
+    def test_zero_decay_does_not_multiply_inf_utility(self) -> None:
+        """decay=0 times an infinite utility EMA is NaN and would be committed."""
+        cbp_state = ContinualBackpropState(  # type: ignore[call-arg]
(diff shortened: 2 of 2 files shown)


kenjohnscreates (approved):

## What I verified

**Tree identity.** The PR diff applied onto current `main` (`58b6780`) produces a tree byte-identical to PR head `24d015b`, which contains current `main`.

**Failing-test-first, reproduced.** Tests-only on `main`: `test_zero_decay_does_not_multiply_inf_utility` fails exactly as intended (1 failed, 20 passed). Full diff: **21/21 pass** (6.6s). `ruff check` clean on both changed files; repo-config strict `mypy` → `Success: no issues found in 163 source files` — this PR is type-gate clean.

**The defect and the fix are correct.** `update_utility` casts `decay_rate` to an array and formed `decay * utilities[i]` unconditionally; with `decay = 0` and an infinite stored EMA that is IEEE `0 * inf = NaN`, and the existing `contribution_finite` guard does not catch it (the *contribution* is finite — it is the decayed term that is NaN), so the NaN was committed. Post-fix, `decay = 0` yields exactly the current contribution — the test's `expected = |activations * grads|` matches the function's own `contribution` definition, so the assertion is precise, not approximate-by-accident. The interplay with the existing guard is preserved: a non-finite contribution still holds the previous utility, including an infinite one, which matches the module's documented hold-don't-rewrite policy.

**No measured-path exposure (why the graph note below is non-gating here).** Repo-wide sweep: `core/continual_backprop.py` is imported only by the package `__init__` re-exports — nothing under `alberta_framework/benchmarks/` uses it; the screening `*_cbp*` arms (including the pinned `adamw_cbp` reductions) are self-contained reimplementations inside `ipmnist_screening.py`. So this change cannot perturb any pinned shard, reduction pin, or publication artifact.

## One non-gating observation (consistency with the #219 discussion)

The `jnp.where(decay == 0.0, …)` guard is emitted into the compiled graph for every decay value: jitted `update_utility` at `decay = 0.9` compiles to a different optimized-HLO (main `7bf3ad49…` → patched `3a92b01c…`). Here that is harmless — no measured path executes this module — but on my #219 review the identical pattern lands on a measured optimizer, where I requested hoisting the zero check to trace time when the scale is static. If that variant is adopted there, consider mirroring it here for one consistent house pattern (`decay_rate` arrives as a plain float at every call site in this module's tests and wrapper). Not blocking either way.

Execution disclosure: I fully read the 47-line diff before executing anything, and ran it re-applied onto `main` in my own checkout rather than checking out the contributor branch.

— [kenjohnscreates]

PR #218, opened 2026-08-15
Skip SwiftTD 0*inf trace decay and allow inf next obs at gamma 0

Kayvan-Zahiri:
## Summary
- SwiftTD decays eligibility traces with `gamma * lambda`. IEEE 0*inf is NaN in those traces and in the published eligibility metric.
- The public next observation is unused in the update (V(s') is already in the TD error), but a non-finite next observation still froze a terminal `gamma=0` step.
- Skip the decay product when the scale is 0. A next observation may be non-finite only when gamma is already 0.

## Test plan
- [x] `test_zero_gamma_does_not_require_finite_next_observation`
- [x] `TestSwiftTDBehavior` and the rest of `tests/test_swift_td.py -m "not slow"` (15 passed)
- [x] SwiftTD cases in `test_optimizer_nonfinite_transactions.py`

### DIFF
--- a/alberta_framework/core/swift_td.py
+++ b/alberta_framework/core/swift_td.py
@@ -56,6 +56,11 @@ from alberta_framework.core.update_safety import (
     select_transaction,
 )
 
+
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Return 0 when ``scale`` is 0 so IEEE ``0 * inf`` does not become NaN."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
 # Paper default for the step-size floor: eta_min = e^-15.
 _DEFAULT_ETA_MIN = math.exp(-15.0)
 
@@ -336,10 +341,14 @@ class SwiftTD:
         new_h_temp = new_h
 
         # Decay eligibility-side traces for the next transition.
-        decay_factor = gamma_scalar * state.trace_decay
-        new_z = decay_factor * z_ext
-        new_p = decay_factor * p_ext
-        new_z_bar = decay_factor * z_bar_ext
+        decay_factor = jnp.where(
+            (gamma_scalar == 0.0) | (state.trace_decay == 0.0),
+            jnp.zeros_like(gamma_scalar),
+            gamma_scalar * state.trace_decay,
+        )
+        new_z = _skip_zero_scale(decay_factor, z_ext)
+        new_p = _skip_zero_scale(decay_factor, p_ext)
+        new_z_bar = _skip_zero_scale(decay_factor, z_bar_ext)
 
         candidate_state = SwiftTDState(
             log_step_sizes=new_log_step_sizes,
@@ -369,11 +378,33 @@ class SwiftTD:
         inputs_valid = (
             jnp.all(jnp.isfinite(td_error))
             & jnp.all(jnp.isfinite(observation))
-            & next_observation_finite
+            & (next_observation_finite | (gamma_scalar == 0.0))
             & jnp.isfinite(gamma_scalar)
         )
+        previous_checked = state.replace(
+            eligibility_traces=jnp.where(
+                (gamma_scalar == 0.0) | (state.trace_decay == 0.0),
+                jnp.zeros_like(state.eligibility_traces),
+                state.eligibility_traces,
+            ),
+            z_bar_traces=jnp.where(
+                (gamma_scalar == 0.0) | (state.trace_decay == 0.0),
+                jnp.zeros_like(state.z_bar_traces),
+                state.z_bar_traces,
+            ),
+            p_traces=jnp.where(
+                (gamma_scalar == 0.0) | (state.trace_decay == 0.0),
+                jnp.zeros_like(state.p_traces),
+                state.p_traces,
+            ),
+            trace_decay=jnp.where(
+                gamma_scalar == 0.0,
+                jnp.zeros_like(state.trace_decay),
+                state.trace_decay,
+            ),
+        )
         update_applied = (
-            floating_tree_is_finite(state)
+            floating_tree_is_finite(previous_checked)
             & inputs_valid
             & jnp.all(jnp.isfinite(meta_delta))
             & jnp.all(jnp.isfinite(delta_w))
--- a/alberta_framework/core/swift_td.py
+++ b/alberta_framework/core/swift_td.py
@@ -381,30 +381,8 @@ class SwiftTD:
             & (next_observation_finite | (gamma_scalar == 0.0))
             & jnp.isfinite(gamma_scalar)
         )
-        previous_checked = state.replace(
-            eligibility_traces=jnp.where(
-                (gamma_scalar == 0.0) | (state.trace_decay == 0.0),
-                jnp.zeros_like(state.eligibility_traces),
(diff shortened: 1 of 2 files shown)


shabbydev (approved):
Static review (execution blocked — no disposable JAX sandbox this session); logic verified by inspection.

This one is broader than the EMA-guard siblings and correctly so — SwiftTD has two distinct poison sources:

1. **`trace_decay == 0`**: the `0 * inf` guard on the trace decays (`new_z`, `new_p`, `new_z_bar`), same IEEE pattern as the rest of the series.

2. **`gamma == 0` (terminal transition)**: the more interesting case. At episode end the `next_observation` may be padding/inf because there is no bootstrap target — the PR correctly relaxes `inputs_valid` to `(next_observation_finite | (gamma_scalar == 0.0))` and simultaneously zeroes `trace_decay` inside `previous_checked` so the finite-check doesn't reject a legitimate terminal update. This is the classic continuing-vs-episodic boundary and it's handled in the right place (validation gate, not the arithmetic).

The `decay_factor` itself also gets the guard (`jnp.where` on either factor being zero), which prevents `0 * inf` when either `gamma` or `trace_decay` is zero and the other has gone non-finite.

Coverage of all three trace families (`z`, `z_bar`, `p`) in both the update and the checked-state substitution is complete and symmetric.


- AI assistance: `yes`
- Agent tooling: `Codex`

kenjohnscreates (changes requested):
**CHANGES_REQUESTED — one blocking item (the repo's strict mypy gate goes red), everything else verified positively, including the parts of this patch that are semantically bolder than the rest of the wave.** Independently re-applied and verified on current `main` (`58b6780`), arm64 M1 Pro, python 3.13, jax 0.11.0 CPU; applied tree byte-identical to PR head `df021ef`, which contains current `main`.

## Blocking: strict mypy regression

```
alberta_framework/core/swift_td.py:384: error: "SwiftTDState" has no attribute "replace"  [attr-defined]
Found 1 error in 1 file (checked 163 source files)
```

Baseline `main` is clean. Runtime is fine — `SwiftTDState` is a `chex.dataclass` and I confirmed `hasattr(state, "replace")` is `True` in this environment — but the repo gate is red, and the PR's test plan ran ruff and the targeted suite only. The house precedent for exactly this construct already exists in core: `continual_backprop.py` calls `cbp_state.replace(  # type: ignore[attr-defined]`. The same one-line annotation here (or an explicit `SwiftTDState(...)` construction) resolves it. This is the identical defect class I flagged on #219; a repo-config `mypy` run before posting would catch it across the whole wave.

## Verified positively

**Failing-test-first, reproduced.** Tests-only on `main`: exactly `test_zero_gamma_does_not_require_finite_next_observation` fails (1 failed, 14 passed); full diff → **15/15 pass**. `ruff check` clean.

**`jnp.where` is the correct idiom here — explicitly unlike #219.** Both scales are genuinely runtime values (`gamma_scalar` is a per-transition argument, `state.trace_decay` lives in state), so the zero-guard cannot be hoisted to trace time. No static-config graph-purity concern applies; my #219 request does not transfer to this module.

**The γ=0 next-observation relaxation cannot leak values.** `next_observation` is computationally unused in `update` — it appears only in the `next_observation_finite` guard (the docstring says so, and I confirmed no other reference in the function). So `next_observation_finite | (gamma_scalar == 0.0)` is a contract-level loosening of a deliberate all-inputs-finite guard at terminal transitions, not a numerical pathway; it is disclosed in the PR title and defensible (at a terminal transition the environment may have no meaningful φ(s′)).

**The `previous_checked` relaxation is provably fail-closed.** The commit gate still ends in `floating_tree_is_finite(candidate_state) & floating_tree_is_finite(candidate_metrics)`, so a state can only recover when this transition provably zeroes the contaminated trace fields (e.g. an inf `z_bar` under `decay_triggered` at γ=0); any inf that survives into a committed field or any computed delta (`delta_w`, `meta_delta`) still blocks the update. I traced the inf-eligibility case explicitly: `trace_dot` goes non-finite → `delta_w` non-finite → held, exactly as before.

**No live measured-path exposure.** `SwiftTD` is reachable only through the lazy `optimizer_from_config` registry (`optimizers.py:2138`); no benchmark instantiates it directly, and repo-wide the only `outputs/` references are immutable archived source tars under quarantined forager roots. No pinned shard or reduction pin executes this module.

## One non-blocking nit

In `previous_checked`, zeroing `trace_decay` when `gamma_scalar == 0` is dead relaxation: the committed `candidate_state` carries the raw `state.trace_decay`, so an infinite `trace_decay` still fails `floating_tree_is_finite(candidate_state)` and blocks the commit regardless. Harmless, but it suggests a recovery that can never actually happen — dropping that clause (or a comment) would keep the checked-copy semantics honest.

Happy to re-verify promptly once the mypy fix is pushed — everything else already checks out.

Execution disclosure: I fully read the 100-line diff before executing anything, and ran it re-applied onto `main` in my own checkout rather than checking out the contributor branch.

— [kenjohnscreates]

kenjohnscreates (approved):
## Re-verification after `866cc1ba` ("Keep zero-gamma SwiftTD recovery fail-closed") — APPROVE

The blocker and the nit from my previous review are both resolved — by deletion, which is the strongest fix available here.

**What changed.** The follow-up removes the entire `previous_checked` relaxation block (the `state.replace(...)` at the old line 384). `update_applied` now gates on `floating_tree_is_finite(state)` — the raw incoming state — so a consumed non-finite trace refuses the update outright instead of being silently zeroed and recovered. This is strictly more fail-closed than the version I reviewed: my earlier analysis showed the relaxation was backstopped (inf eligibility → `delta_w` NaN → held), but now the refusal is direct rather than incidental. It also deletes the dead `trace_decay` zeroing I flagged as a nit. The γ=0 finite-next-observation allowance is retained (`next_observation_finite | (gamma_scalar == 0.0)`), which my previous review verified is contract-only — `next_observation` is computationally unused at γ=0.

**Verification** (diff re-applied onto current `main` `8b49cec` in a scratch worktree; contributor branch never checked out):

- Touched files **byte-identical to PR head `866cc1ba`** (`git diff` against `pull/218/head` restricted to `core/swift_td.py` + `tests/test_swift_td.py` is empty).
- `swift_td.py` has no other commits on `main` since my previous review baseline, so the red proof is against the same effective source.
- Failing-test-first: tests-only on baseline → exactly one deterministic failure, `test_zero_gamma_does_not_require_finite_next_observation` (the discriminating test); full diff → **16/16**. The new `test_zero_gamma_does_not_disguise_consumed_inf_trace_state` passes on baseline as well — as expected, since it pins the fail-closed behavior this follow-up preserves (its job is to keep the relaxation from returning), and it correctly asserts both `update_applied` is false and the state round-trips unchanged.
- `ruff check .` clean; repo-config strict mypy clean at **163 source files** — the previously red `.replace` line no longer exists, and the remaining test-side `.replace` uses carry the house-precedent `# type: ignore[attr-defined]`.
- No live measured-path exposure (unchanged from my previous review: SwiftTD is reachable only via the lazy `optimizer_from_config` registry; `outputs/` hits are archived tars).
Compute receipt: 869925 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

PR #219, opened 2026-08-15
Skip canonical UPGD 0*inf utility and moment decay

Kayvan-Zahiri:
## Summary
- First-order CanonicalUPGD, Alberta AdaUPGD, and official AdaUPGD all decay utility (and Adam-style moments) with a scale that may be 0. IEEE 0*inf is NaN. CanonicalUPGD would commit that NaN EMA.
- Skip those products before writing the new tracker. On a zero-decay path, Alberta AdaUPGD also treats already-infinite EMAs as skippable so the finite-state guard can recover.

## Test plan
- [x] `test_zero_utility_decay_does_not_multiply_inf_ema`
- [x] `tests/test_canonical_upgd.py` (37 passed)

### DIFF
--- a/alberta_framework/core/canonical_upgd.py
+++ b/alberta_framework/core/canonical_upgd.py
@@ -98,9 +98,11 @@ _RAW_GLOBAL_PROFILES = frozenset(
 )
 
 
-def _skip_zero_scale(scale: Array, value: Array) -> Array:
-    """Return 0 when ``scale`` is 0 so IEEE ``0 * inf`` does not become NaN."""
-    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+def _static_zero_scale(scale: float, value: Array) -> Array:
+    """Skip a statically disabled EMA without changing enabled JAX graphs."""
+    if scale == 0.0:
+        return jnp.zeros_like(value)
+    return scale * value
 
 
 @dataclass(frozen=True)
@@ -430,7 +432,7 @@ class CanonicalUPGD:
             instantaneous = -clean_gradient * param
             next_utility = jnp.where(
                 active,
-                _skip_zero_scale(jnp.asarray(beta, dtype=param.dtype), utility)
+                _static_zero_scale(beta, utility)
                 + (1.0 - beta) * instantaneous,
                 utility,
             )
@@ -1086,17 +1088,27 @@ class AlbertaAdaUPGD:
         state_is_valid = self.state_valid(state, params)
         if self._config.utility_decay == 0.0 or self._config.second_moment_decay == 0.0:
             checked_utility = (
-                jax.tree.map(jnp.zeros_like, state.utility_ema)
+                jax.tree.map(
+                    lambda value: jnp.where(
+                        jnp.isfinite(value), value, jnp.zeros_like(value)
+                    ),
+                    state.utility_ema,
+                )
                 if self._config.utility_decay == 0.0
                 else state.utility_ema
             )
             checked_moment = (
-                jax.tree.map(jnp.zeros_like, state.gradient_second_moment)
+                jax.tree.map(
+                    lambda value: jnp.where(
+                        jnp.isfinite(value), value, jnp.zeros_like(value)
+                    ),
+                    state.gradient_second_moment,
+                )
                 if self._config.second_moment_decay == 0.0
                 else state.gradient_second_moment
             )
             state_is_valid = self.state_valid(
-                state.replace(
+                state.replace(  # type: ignore[attr-defined]
                     utility_ema=checked_utility,
                     gradient_second_moment=checked_moment,
                 ),
@@ -1183,7 +1195,7 @@ class AlbertaAdaUPGD:
             instantaneous_utility = -gradient * param
             proposed_utility = jnp.where(
                 eligible,
-                _skip_zero_scale(jnp.asarray(beta_utility, dtype=param.dtype), utility)
+                _static_zero_scale(beta_utility, utility)
                 + (1.0 - beta_utility) * instantaneous_utility,
                 utility,
             )
@@ -1200,9 +1212,9 @@ class AlbertaAdaUPGD:
                 0.0,
             )
 
-            proposed_moment = _skip_zero_scale(
-                jnp.asarray(beta_second, dtype=param.dtype), moment
-            ) + (1.0 - beta_second) * jnp.square(gradient)
+            proposed_moment = _static_zero_scale(beta_second, moment) + (
+                1.0 - beta_second
+            ) * jnp.square(gradient)
             moment_clock = jnp.maximum(proposed_step, 1).astype(param.dtype)
             moment_correction = 1.0 - jnp.power(beta_second, moment_clock)
             corrected_moment = proposed_moment / jnp.maximum(
@@ -1810,19 +1822,16 @@ class OfficialAdaUPGD:
             second_leaves,
             strict=True,
         ):
-            utility_scale = jnp.asarray(self._config.utility_decay, dtype=param.dtype)
-            first_scale = jnp.asarray(self._config.beta1, dtype=param.dtype)
-            second_scale = jnp.asarray(self._config.beta2, dtype=param.dtype)
             proposed_utility_leaves.append(
-                _skip_zero_scale(utility_scale, utility)
+                _static_zero_scale(self._config.utility_decay, utility)
                 + (1.0 - self._config.utility_decay) * (-gradient * param)
             )
             proposed_first_leaves.append(
-                _skip_zero_scale(first_scale, first)
+                _static_zero_scale(self._config.beta1, first)
                 + (1.0 - self._config.beta1) * gradient
             )
             proposed_second_leaves.append(
-                _skip_zero_scale(second_scale, second)
+                _static_zero_scale(self._config.beta2, second)
                 + (1.0 - self._config.beta2) * jnp.square(gradient)
             )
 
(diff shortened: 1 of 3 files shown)


kenjohnscreates (changes requested):
**CHANGES_REQUESTED — the underlying defect is real and the CanonicalUPGD test is discriminating, but the patch as written fails the repository's strict mypy gate and, more consequentially, changes the compiled graph of a measured optimizer for every configuration, not just the pathological zero-decay ones.** Independently re-applied and verified on current `main` (`58b6780`), arm64 M1 Pro, python 3.13, jax 0.11.0 CPU; the applied tree is byte-identical to PR head `a729e9b`, which contains current `main`.

## Confirmed: strict mypy regression (repo gate)

Repo-config `mypy` on the patched tree:

```
alberta_framework/core/canonical_upgd.py:1099: error: "AlbertaAdaUPGDState" has no attribute "replace"  [attr-defined]
Found 1 error in 1 file (checked 163 source files)
```

Baseline `main` is clean (`Success: no issues found in 163 source files`). Runtime is not affected — `AlbertaAdaUPGDState` is a `chex.dataclass`, so `.replace()` exists when it executes — but the PR's test plan ran ruff and the targeted suite only, and no test exercises this line: the new test drives **CanonicalUPGD**, while line 1099 lives in the **AlbertaAdaUPGD** zero-decay `state_valid` block, which the suite never enters. Suggested fix in house style: construct the probe state explicitly like the module's other sites (`AlbertaAdaUPGDState(...)  # type: ignore[call-arg]`, cf. lines 966/1269), and add an AlbertaAdaUPGD zero-decay test so the recovery block is actually executed.

## Confirmed: compiled-graph change at nonzero decay on a measured optimizer

`_skip_zero_scale` is `jnp.where(scale == 0.0, 0, scale * value)` evaluated at **runtime**, and three of the four sites apply it to a scale that is a static Python config float. The select is emitted into the traced graph for every configuration and XLA does not simplify it away. Measured with a deterministic probe (`CanonicalUPGDConfig(utility_decay=0.999, noise_std=0.0)`, one jitted `update` on a 2-leaf pytree; script reproduced below):

| | optimized HLO sha256 | StableHLO sha256 | result bytes sha256 |
|---|---|---|---|
| `main` (58b6780) | `eb523a9e…` | `c2c753c5…` | `499bf1c4…` |
| PR applied | `b72a5804…` | `92beed4d…` | `499bf1c4…` |

Same numbers, different program — which is precisely the class of change this repository treats as load-bearing: `CanonicalUPGD` is imported by `alberta_framework/benchmarks/upgd_ipmnist.py`, whose pinned `outputs/upgd_ipmnist/` artifacts and publication baselines depend on shard reproducibility, and #46/#48 established that reassociation-level graph drift (with bitwise-identical accuracy) is enough to redden reduction pins.

The PR itself already contains the graph-preserving pattern: the AlbertaAdaUPGD `state_valid` block branches on `if self._config.utility_decay == 0.0` at **trace time**, so nonzero-decay configs compile the exact prior program. Request: hoist the zero-scale decision the same way at the other three sites (the scales at all of them come from static config floats — `beta`/`beta_utility`/`beta_second`/`utility_decay`/`beta1`/`beta2`), keeping `jnp.where` only for a scale that can genuinely arrive traced. Under that variant the nonzero-decay HLO stays byte-identical and the zero-decay fix is preserved; the new test passes unchanged.

## What does verify

- Failing-test-first reproduced: tests-only on `main` → `test_zero_utility_decay_does_not_multiply_inf_ema` fails exactly as intended; full diff → 37/37 pass (3.3s).
- The defect is real: `utility_decay=0.0` is constructible, `0.0 * inf` is NaN in float32, and pre-fix CanonicalUPGD commits that NaN into the utility EMA.
- The intentional semantic change (zero-decay AlbertaAdaUPGD treats an already-infinite EMA as recoverable rather than invalid-state) is clearly disclosed in the summary and is confined to zero-decay configs by the static branch.

## Wave-wide note

The same runtime-`where`-on-a-static-scale shape appears across the sibling 0*inf PRs posted tonight. For any of them touching a module reachable from a measured benchmark path, the same two checks apply: repo-config strict mypy, and HLO byte-invariance at the configurations the benchmarks actually run. Happy to re-verify this PR promptly once the static-guard variant is pushed — the probe script and hashes above make the check one command each way.

<details><summary>HLO probe script</summary>

```python
import hashlib
import jax, jax.numpy as jnp, jax.random as jr
import numpy as np
from alberta_framework.core.canonical_upgd import CanonicalUPGD, CanonicalUPGDConfig

params = {"w": jnp.ones((8, 4), jnp.float32), "b": jnp.zeros(4, jnp.float32)}
grads = {"w": jnp.full((8, 4), 0.01, jnp.float32), "b": jnp.full(4, -0.02, jnp.float32)}
opt = CanonicalUPGD(CanonicalUPGDConfig(utility_decay=0.999, noise_std=0.0))
state = opt.init(params)
step = lambda s, p, g, k: opt.update(s, p, g, k)
lowered = jax.jit(step).lower(state, params, grads, jr.key(0))
print("optimized:", hashlib.sha256(lowered.compile().as_text().encode()).hexdigest())
print("stablehlo:", hashlib.sha256(lowered.as_text().encode()).hexdigest())
res = jax.jit(step)(state, params, grads, jr.key(0))
d = hashlib.sha256()
for leaf in jax.tree_util.tree_leaves((res.params, res.state)):
    d.update(np.asarray(leaf).tobytes())
print("result:", d.hexdigest())
```

</details>

Execution disclosure: I fully read the 123-line diff before executing anything and ran it re-applied onto `main` in my own checkout rather than checking out the contributor branch.

— [kenjohnscreates]

kenjohnscreates (approved):
## Re-verification after `a86debfc` ("Specialize zero-decay UPGD paths statically") — APPROVE

Both blockers from my previous review are resolved. Verified by re-applying the PR diff onto current `main` (`8b49cec`) in a scratch worktree; the two touched files are **byte-identical to PR head `a86debfc`** (`git diff` restricted to `core/canonical_upgd.py` + `tests/test_canonical_upgd.py` against `pull/219/head` is empty). I did not check out the contributor branch.

**1. HLO byte-invariance on the measured path: FIXED.** Re-ran my probe (same script as previous review) at `CanonicalUPGDConfig(utility_decay=0.999, noise_std=0.0)` paired within one worktree:

| | optimized HLO sha256 | StableHLO sha256 | result bytes sha256 |
|---|---|---|---|
| `main` (8b49cec) | `d7012a2c…` | `52db409e…` | `499bf1c4…` |
| PR applied | `d7012a2c…` | `52db409e…` | `499bf1c4…` |

All three identical — nonzero-decay configs now compile the exact prior program, so `upgd_ipmnist` shard reproducibility is untouched. The zero-scale decision is hoisted to trace time (`_static_zero_scale` branches on `if scale == 0.0:` over static config floats), exactly the graph-preserving pattern requested; the new `test_nonzero_static_decay_preserves_legacy_hlo` pins this in-suite going forward. (Note: absolute hashes differ from the table in my previous review because lowered text embeds source-path metadata and these runs are from a different worktree path — the comparison is paired within one path both times, which is the load-bearing part.)

**2. Strict mypy: FIXED.** Repo-config mypy is clean at **163 source files** with the diff applied (house-precedent `# type: ignore[attr-defined]` on the chex-dataclass `.replace` sites). The previously dead line is now exercised: `test_alberta_zero_decays_recover_unused_inf_trackers` drives the AlbertaAdaUPGD zero-decay recovery block directly, as requested.

**Suite:** `tests/test_canonical_upgd.py` **38/38** (4 new tests), `ruff check .` clean. One transparency note: a pure tests-only run against baseline source now fails at collection (`ImportError: cannot import name '_static_zero_scale'`) rather than at the assertions, because the new tests import the PR's helper; the behavioral red proof of the zero-decay 0*inf defect was established in my previous review against the original head and still stands.

Nice follow-up — the static specialization plus the in-suite HLO pin is stronger than what I asked for.
Compute receipt: 1885641 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

kenjohnscreates:
## Correction to my re-verification evidence (conclusion unchanged)

While re-using my HLO probe on a sibling PR today I found a flaw in how I executed it for my re-review here, and I want the record corrected.

**The flaw:** I invoked the probe as `python /scratchpad/hlo_probe_219.py` from inside the scratch worktree. Python puts the *script's* directory first on `sys.path`, not the working directory, so the import fell through to this venv's editable install — which points at a stale local branch tree. Both the "main" and "PR applied" rows in my re-review's table therefore measured the same stale module, and their equality was vacuous. (`python -m pytest` resolves from the working directory, so every test, ruff, and mypy result in that review was measured correctly — only the standalone probe table was affected.)

**The corrected measurement** (probe file placed inside the worktree; paired at pre-merge base `8b49cec`, same path both runs, `CanonicalUPGDConfig(utility_decay=0.999, noise_std=0.0)`):

| | StableHLO sha256 | result bytes sha256 |
|---|---|---|
| base `8b49cec` | `52db409e…` | `499bf1c4…` |
| + #219 diff | `52db409e…` | `499bf1c4…` |

StableHLO and result bytes are **byte-identical**, so the graph-preserving conclusion of my re-review stands, now on valid evidence — independently corroborated by `test_nonzero_static_decay_preserves_legacy_hlo`, which I did run with correct imports. One methodological note from the corrected run: the *optimized* HLO text hash differs across this diff even with identical StableHLO, because compiled-module text embeds source-location metadata that shifts with unrelated line changes — optimized-text hashes should not be used as an invariance metric, and my original review's table (which relied on them alongside StableHLO) was right for the wrong-and-right reasons respectively. StableHLO text (location-free) plus result bytes is the reliable pair.

My apologies for the noise; flagging it myself since the table is load-bearing for how this class of change gets reviewed.
— [kenjohnscreates]

PR #220, opened 2026-08-15
Skip future-utility 0*inf when trace decay is disabled

Kayvan-Zahiri:
## Summary
- Contribution traces, factored traces, and the optional second-moment normalizer all multiply a previous tracker by a decay that can be 0. IEEE 0*inf is NaN, and the fail-closed guard would keep the infinite previous trace forever.
- Skip those products before adding the current sample. When decay is 0, also skip the finite-previous check so a disabled trace can recover from an already-infinite tracker.

## Test plan
- [x] `test_zero_trace_decay_does_not_multiply_inf_traces`
- [x] `tests/test_future_utility.py` (9 passed)

### DIFF
--- a/alberta_framework/core/future_utility.py
+++ b/alberta_framework/core/future_utility.py
@@ -14,6 +14,11 @@ from jax import Array
 from jaxtyping import Bool, Float
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled decay does not poison the next trace."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 class FutureUtilityEstimate(NamedTuple):
     """One-step utility estimate with an explicit transaction verdict."""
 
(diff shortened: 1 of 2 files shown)


shabbydev (approved):

This one has a subtly different gate pattern from the rest of the series and it deserves a callout: instead of zeroing the poisoned trackers for the finiteness check, `inputs_valid` is RELAXED via `jnp.logical_or(decay == 0.0, traces_finite)` — inf input traces become an explicitly VALID transaction at decay 0. I traced why that's sound here: with decay==0 the new traces are pure functions of the current (finite-checked) sample — `0*prev + active_errors * features` — so the outputs are finite by construction; and `decayed_contribution` in the returned diagnostics is the guarded 0, so no inf escapes through the diagnostics channel either. The relaxation is principled, not a loophole.

Coverage across the module's three decayed families is complete: contribution/energy traces in `contribution_trace_output_loss_reduction_with_diagnostics`, error/feature/energy traces in `trace_output_loss_reduction_with_diagnostics`, and `second_moment` in `normalize_future_utility_signal`. The `new_error_trace = jnp.where(active_mask, new_error_trace, decayed_error)` rewrite correctly reuses the guarded term in both branches — the pre-change code recomputed `decay * error_trace` for the inactive branch, which would have re-introduced the NaN; that re-use is the load-bearing detail.

One observation (non-blocking): the test pins output finiteness at decay 0 with inf traces; it would additionally be worth pinning `inputs_valid`-side behavior explicitly (a transaction verdict assertion) since that gate's semantics are what changed most.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

PR #221, opened 2026-08-15
Skip historical UPGD 0*inf utility and diagnostic EMAs

Kayvan-Zahiri:
## Summary
- Historical UPGDLearner allows utility, unit-utility, loss, and repetition decays of 0. IEEE 0*inf is NaN, and this learner commits the tracker with no fail-closed transaction.
- Skip those products before adding the current signal so a zero-decay step recovers from an already-infinite EMA.

## Test plan
- [x] `TestUtilityTracking.test_zero_utility_decay_does_not_multiply_inf_ema`
- [x] `TestUtilityTracking` (12 passed)

### DIFF
--- a/alberta_framework/core/upgd.py
+++ b/alberta_framework/core/upgd.py
@@ -50,6 +50,12 @@ from alberta_framework.core.optimizers import Bounder, ObGDBounding
 from alberta_framework.core.types import MLPParams
 from alberta_framework.core.update_safety import zero_if_collapsed_infinity
 
+
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled decay does not poison the next EMA."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 # =============================================================================
 # Types
 # =============================================================================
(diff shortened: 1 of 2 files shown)


shabbydev (approved):
Static review (execution proof blocked — no isolated JAX sandbox available to me this session); verified against the full `upgd.py` source.

Completeness claim holds: I enumerated every decayed tracker in `UPGDState` — `utilities`, `unit_utilities`, `unit_long_utilities`, `unit_gradient_emas`, `loss_fast_ema`, `loss_slow_ema`, `target_repeat_ema`, `target_simplex_ema` — and all eight are routed through `_skip_zero_scale`. The remaining state fields are correctly out of scope: `meta_*_log_scale` are gradient-updated learned scalars (not decay EMAs), `previous_*_grads` are raw copies, `unit_ages` accumulate by +1. Nothing decayed is left unguarded.

Semantics with decay fully disabled are consistent: each EMA degenerates to instantaneous tracking (`0*prev + 1*sample`), which is the only sane reading of "decay 0". Note this makes `loss_ratio = loss_fast_ema / (loss_slow_ema + 1e-12)` an instantaneous ratio on the `_unit_replacement_loss_gate_ratio` path — noisier gating, but that is what the operator asked for by disabling both decays, and both decays share one config family so a mixed fast=0/slow>0 configuration would be the only surprising case (still finite, just asymmetric smoothing).

The test matrix pins the hazard mechanism directly (`0.0 * inf` is not finite) and covers all four tracker families with step_size=0/perturbation_sigma=0 isolating the EMA path from weight updates. Good isolation choice.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

PR #222, opened 2026-08-16
Skip learning-signal 0*inf when an EMA decay is disabled

Kayvan-Zahiri:
## Summary
- Learning-signal loss and change EMAs allow decay 0. IEEE 0*inf is NaN, and the finite-state guard would freeze the infinite tracker.
- Skip those products before mixing in the current observation. When a decay is 0, also treat the previous tracker as skippable so a later finite observation can recover.

## Test plan
- [x] `test_zero_ema_decay_does_not_multiply_inf_trackers`
- [x] `tests/test_learning_signals.py` (12 passed)

### DIFF
--- a/alberta_framework/core/learning_signals.py
+++ b/alberta_framework/core/learning_signals.py
@@ -53,6 +53,11 @@ from jaxtyping import Bool, Float, Int
 _INT32_MAX = 2_147_483_647
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled EMA decay does not poison the next value."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def _saturating_increment(value: Array) -> Array:
     """Increment a non-negative int32 counter without lifetime wraparound."""
     maximum = jnp.asarray(_INT32_MAX, dtype=jnp.int32)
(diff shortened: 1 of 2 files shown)


shabbydev (approved):
Static review (no isolated JAX execution environment this session), but the change is small and the regression test is precise.

Two-layer fix, both layers necessary:

1. **Validation layer**: `checked_fast_loss`/`checked_slow_loss`/`checked_change` substitute zeros for disabled trackers before the range/finite checks, so a historically-poisoned tracker whose decay is now 0 doesn't wedge `state_valid` permanently. The bounds checks (`>= 0`, `<= max_observed_loss`, `<= 1.0`) still apply to the *enabled* trackers with full strictness — nothing is relaxed.

2. **Update layer**: `_skip_zero_scale` guards the actual EMA recurrence for `fast_loss`, `slow_loss`, and `sustained_change_probability`, so `0 * inf` never produces NaN in the first place.

The regression test (`test_zero_ema_decay_does_not_multiply_inf_trackers`) does the right thing: it first demonstrates the raw IEEE failure (`0.0 * inf` is not finite), then shows the guarded estimator recovers to the exact expected value (`0.25`) rather than NaN or a silently-clamped surrogate. Also covers the calibrated-state path via `sustained_change_probability`.

Minor (non-blocking): mixing the Python-level `if self._config.X == 0.0` (checked-state prep) with `jnp.where` (update) is fine here because config values are static at trace time, but the asymmetry is worth a comment for future readers.


- AI assistance: `yes`
- Agent tooling: `Codex`

PR #223, opened 2026-08-16
Skip reward-model 0*inf when error EMA decay is disabled

Kayvan-Zahiri:
## Summary
- RLS reward-model abs-error EMA allows error_decay=0. IEEE 0*inf is NaN, and the fail-closed guard would freeze that diagnostic.
- Skip the product before mixing in the current absolute error. When decay is 0, also treat the previous EMA as skippable so a later finite reward can recover.

## Test plan
- [x] `test_zero_error_decay_does_not_multiply_inf_ema`
- [x] `tests/test_reward_model.py` (5 passed)

### DIFF
--- a/alberta_framework/core/reward_model.py
+++ b/alberta_framework/core/reward_model.py
@@ -24,6 +24,11 @@ from jax import Array
 from jaxtyping import Bool, Float
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled error EMA does not poison the diagnostic."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 @dataclass(frozen=True)
 class RLSRewardModelConfig:
     """Configuration for a linear recursive-least-squares reward model.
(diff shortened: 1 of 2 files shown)


shabbydev (approved):

Coverage is complete for this module: `abs_error_ema` is the ONLY decayed tracker in `RLSRewardModel` — `weights` and `covariance` are exact RLS recursions, not EMAs — so guarding this one EMA plus the hold-gate closes the whole 0*inf surface here. The config validator admits `error_decay = 0.0` (`[0, 1)`), so this is a supported config path, not a hypothetical.

The layering is right: `_skip_zero_scale` fixes the arithmetic (`0.0 * inf = NaN` short-circuited to 0), while `proposed_finite` still checks `next_abs_error_ema` — so a genuinely infinite abs-error sample (infinite reward) still fails closed and holds the previous state. The guard only rescues the disabled-decay multiply; it does not weaken inf rejection on the actual data path. The `checked_error_ema` treatment of the source-side finiteness gate is the necessary other half: without it, a tracker poisoned before the decay was disabled would wedge every future update.

The step_count==0 init branch is correctly preserved ahead of the guard, so first-step semantics are unchanged.

Minor (non-blocking): the test exercises decay=0 with `forgetting=1.0`; a second case with forgetting<1.0 would additionally pin that the RLS recursion itself stays finite through the rescued update, but the current assertion on `abs_error_ema` is the load-bearing one.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

PR #224, opened 2026-08-16
Skip UPGD-memory 0*inf when reliability decay is disabled

Kayvan-Zahiri:
## Summary
- Avoid IEEE `0 * inf` contamination when `reliability_decay=0` by skipping only the disabled EMA products.
- Recover poisoned disabled-EMA history without weakening the rest of the transaction guard.
- Preserve finite prior loss EMAs exactly because they still determine the current blend gate before the zero-decay update overwrites them.
- Keep poisoned consumed state such as `memory_logit` fail-closed with atomic rollback and neutral outputs.

## Validation
- [x] `tests/test_upgd_memory.py tests/test_upgd_memory_grad.py tests/test_production_steps.py` — 41 passed
- [x] focused Ruff
- [x] strict mypy on `upgd_memory.py`
- [x] `git diff --check`
- [x] no `outputs/**` changes

The new regressions cover poisoned EMA recovery, exact finite-history gate semantics, and rejection of a poisoned non-EMA gate input.

### DIFF
--- a/alberta_framework/core/upgd_memory.py
+++ b/alberta_framework/core/upgd_memory.py
@@ -43,6 +43,11 @@ def _skip_zero_scale(scale: Array, value: Array) -> Array:
     return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
 
 
+def _finite_or_zero(value: Array) -> Array:
+    """Replace only poisoned disabled-EMA history, retaining finite history exactly."""
+    return jnp.where(jnp.isfinite(value), value, jnp.zeros_like(value))
+
+
 UPGDMemoryReadoutMode = Literal["linear_mse", "softmax_ce"]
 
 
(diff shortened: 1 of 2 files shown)


shabbydev (approved):
Static review (execution proof blocked — no isolated JAX sandbox available to me this session); the guard pattern is verifiable by inspection and the regression test pins the exact 0*inf NaN mechanism.

Verified against the module source: `UPGDMemoryState` carries exactly four decayed trackers — `upgd_loss_ema`, `memory_loss_ema`, `blended_loss_ema`, `allocation_ema` — and all four are covered in BOTH places that matter: the EMA update arithmetic (via `_skip_zero_scale`) and the `checked_state` zeroing that feeds `floating_tree_is_finite`. The second half is the easy one to miss: without it, a tracker poisoned before the decay was disabled would wedge `update_applied` false forever even though the arithmetic is fixed.

Two module-specific confirmations:
1. The hazard is on a supported config path, not hypothetical: the config validator admits `reliability_decay = 0.0` (`[0, 1)`).
2. The `_blend_gate` change has the right disabled semantics: zeroing both EMAs makes `reliability_delta = 0`, so the reliability term contributes exactly `reliability_logit_scale * 0` — the gate degrades to confidence + memory_logit instead of comparing a NaN. And `checked_state = state.replace(...)` preserves the finiteness check over all OTHER state fields (counts, logits), so the hold-gate isn't weakened elsewhere.

One note (non-blocking): `_skip_zero_scale`'s exact `scale == 0.0` comparison is safe here because the decay originates as a Python config float (exact 0.0), not a traced computation; worth keeping in mind if this helper ever gets reused with computed scales.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

PR #226, opened 2026-08-16
Skip normalizer 0*inf when decay or momentum is disabled

Kayvan-Zahiri:
## Summary
- Let `EMANormalizer(decay=0)` recover when its previous mean or variance is nonfinite without evaluating the disabled `0 * inf` terms.
- Let `StreamingBatchNormalizer(momentum=0)` ignore only a nonfinite previous variance. Its previous mean remains required because the established update uses `(observation - old_mean)^2`; a poisoned mean still fails closed.
- Preserve the prior arithmetic exactly for finite states, and key zero-scale validity from the persisted state scalar so host/state configuration mismatches remain invalid.

## Test plan
- [x] `tests/test_normalizers.py` (44 passed)
- [x] `tests/test_checkpoints.py tests/test_config_serialization.py tests/test_single_step_api.py` (59 passed)
- [x] `ruff check .`
- [x] strict `mypy` (163 source files)

No evidence artifacts or `outputs/` paths were changed.

Original contribution made with Cursor; maintainer repair performed with Codex. No Slop measured-run receipt was attached.

### DIFF
--- a/alberta_framework/core/normalizers.py
+++ b/alberta_framework/core/normalizers.py
@@ -36,6 +36,12 @@ from alberta_framework.core.update_safety import (
     floating_tree_is_finite as _floating_tree_is_finite,
 )
 
+
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled decay does not poison the next moment."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 NORMALIZER_STATE_SCHEMA = "alberta.normalizer-state.v2"
 WELFORD_ESTIMATOR_SCHEMA = "alberta.welford-cumulative-float32-fail-stop-at-2p24.v2"
 BOUNDED_RECENCY_ESTIMATOR_SEMANTICS = "bounded-recency"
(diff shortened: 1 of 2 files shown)


shabbydev (approved):
Static review (no isolated JAX sandbox this session); verified by inspection.

The normalizer variant of the disabled-decay fix, with two module-appropriate adaptations worth calling out:

1. **Checked-state substitution values**: using `jnp.ones_like(state.var)` (not zeros) for the variance slot in the pre-check is the careful choice — a zero variance could trip a separate non-negativity/degeneracy guard downstream, ones are neutral for the finite check and safe for any `var > 0` assertion.

2. **Semantics at decay = 0 are correct, not just finite**: `new_mean = observation` and `new_var = delta2²` (which collapses to 0 since `delta2 = observation - new_mean = 0`) is exactly what "no memory" should mean for a Welford/EMA normalizer — the recurrence degenerates to a pass-through instead of dragging a poisoned historical mean forward. The standard identity (`var' = decay·var + (1−decay)·δ·δ2'`) agrees at decay=0, so the `jnp.where` changes nothing for finite states and only removes the `0 · inf` path.

Also good: the guard sits on `effective_decay` (post min-count ramp), which is the value actually used in the arithmetic — guarding the raw config instead would have missed the ramp interaction.


- AI assistance: `yes`
- Agent tooling: `Codex`

PR #227, opened 2026-08-16
Skip baseline-optimizer 0*inf moment EMAs

Kayvan-Zahiri:
## Summary
- Skip IEEE `0 * inf` only for genuinely disabled Adam, RMSprop, and NADALINE moment trackers.
- Statically specialize the configured nonzero path so the ordinary Adam/RMSprop benchmark graphs retain their direct legacy multiplication.
- Require the persisted state decay itself to be zero before poisoned history is relaxed; mismatched nonzero persisted decay remains fail-closed.
- Cover both linear optimizer updates and the per-parameter checked paths used by the IPMNIST AdamW baseline.

## Validation
- [x] `tests/test_baseline_optimizers.py tests/test_upgd_ipmnist.py` — 81 passed
- [x] focused Ruff
- [x] strict mypy on `baseline_optimizers.py`
- [x] `git diff --check`
- [x] no `outputs/**` changes

The added regressions exercise linear and arbitrary-shape recovery, persisted-decay mismatch rejection, atomic rollback, and neutral rejected steps.

### DIFF
--- a/alberta_framework/core/baseline_optimizers.py
+++ b/alberta_framework/core/baseline_optimizers.py
@@ -53,6 +53,17 @@ from alberta_framework.core.update_safety import (
     select_transaction,
 )
 
+
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled EMA decay does not poison the next moment."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
+def _zero_if_disabled(scale: Array, value: Array) -> Array:
+    """Treat a zero-decay tracker as skippable in the finite-state check."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), value)
+
+
 # =============================================================================
 # State dataclasses
 # =============================================================================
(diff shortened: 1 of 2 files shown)


shabbydev (approved):
Static review (execution proof blocked — no isolated JAX sandbox available to me this session), but the pattern is verifiable by inspection and the regression test pins it.

The defect is real IEEE 754 behavior: `0.0 * inf = NaN`, so any tracker that reached `inf` while its EMA decay was nonzero poisons the moment update forever once the decay is later set to 0 — the exact opposite of what "disable the decay" should mean. `_skip_zero_scale` short-circuits the multiply with `jnp.where(scale == 0.0, zeros, scale * value)`, which is the correct JAX idiom for this (forward-value safe).

Coverage is complete in this module: both `update_from_gradient_checked` and `update` get the guard on `m`/`v`, and `update` additionally covers `bias_m`/`bias_v` — the bias trackers are the easy ones to miss. The `_zero_if_disabled` treatment of the finite-check state is equally important: without it, a disabled-but-poisoned tracker would keep `floating_tree_is_finite(state)` failing and the update would be rejected forever even though the guard fixed the arithmetic.

One consistency note (non-blocking): sibling PR #222 uses a Python-level config branch for the checked state while this uses `jnp.where`; both are sound here since `beta1`/`beta2` are traced arrays in this module, and the where-form is the right choice for traced values.


- AI assistance: `yes`
- Agent tooling: `Codex`

PR #228, opened 2026-08-16
fix(ipmnist): bind partial manifest to one byte snapshot

### LINKED ISSUE #208: Fix UPGD-IPMNIST partial-manifest split-read race
## Measurement-integrity bug

`_v2_partial_manifest` currently reads each shard twice: once with `read_bytes()` for `size_bytes`/`sha256`, then again through `_strict_json_object(path)` for `learner`/`seed_id`. If the path is replaced between those reads, the manifest can bind the digest of one byte snapshot to the identity decoded from another.

## Bounded fix

- Lane: permanently nonpromoting UPGD-IPMNIST v2 artifact assembly.
- Change: decode and hash the same immutable byte read.
- Proof: add a failing regression test that simulates path mutation between reads and requires identity/digest consistency.
- Verification: focused UPGD-IPMNIST tests, Ruff, mypy, and `git diff --check`.
- Compute/seeds: none; this is an artifact-integrity fix, not a benchmark claim.

No pinned artifact or `outputs/` path will be modified.

techforky:
Closes #208.

## Summary

- Decode each UPGD-IPMNIST v2 shard from the same byte snapshot used for `size_bytes` and `sha256`.
- Preserve the existing strict duplicate-key and finite-number validation.
- Add a regression test that replaces the path immediately after its first byte read and proves the manifest identity and digest remain bound to that original snapshot.

## Measurement-integrity effect

Lane: permanently nonpromoting UPGD-IPMNIST v2 artifact assembly.

Before this change, `_v2_partial_manifest` hashed one read but decoded `learner` and `seed_id` from a second read. The failing-test-first reproduction bound replacement identity `upgd_w`/`11` to the original `adamw`/`7` shard digest. After the change, that test passes because identity, size, and digest all derive from one immutable byte read.

This is an artifact-integrity fix, not a benchmark-performance claim. It consumes no benchmark compute or seeds, changes no protocol, and modifies no pinned artifact or `outputs/` path.

## Verification

Commit: `97ec38229fbd3191bb951be3485fccf1c72407e0`

- `.venv/bin/python -m pytest tests/test_upgd_ipmnist.py -q -o addopts="" --basetemp=/tmp/asi-pr208-pytest` — 49 passed.
- `.venv/bin/python -m ruff check .` — passed.
- `.venv/bin/python -m mypy` — no issues in 163 source files.
- `git diff --check origin/main...HEAD` — passed.

Provider/model/client: `openai` / `gpt-5.6-sol` / `codex`.

- [x] logs: https://gist.github.com/techforky/01dd31f4ca773cbb66fb0ad1dedb8740/2daffb6039a728237979decfceb8493998bd6020

- [x] domain-artifact: https://github.com/techforky/asi/blob/97ec38229fbd3191bb951be3485fccf1c72407e0/tests/test_upgd_ipmnist.py#L664-L697

Client / agent tooling: codex
Contribution skill revision: elizaOS/slopdotcash@a99f2e649ecadc35896c9d65c8d5136e947333ce:skills/contribute-to-asi
— [ipmnist-fix-codex]

### DIFF
--- a/alberta_framework/benchmarks/upgd_ipmnist.py
+++ b/alberta_framework/benchmarks/upgd_ipmnist.py
@@ -1198,7 +1198,7 @@ def partial_payload(result: IPMNISTRunResult) -> dict[str, Any]:
     }
 
 
-def _strict_json_object(path: Path) -> dict[str, Any]:
+def _decode_strict_json_object(raw: bytes, *, path: Path) -> dict[str, Any]:
     def pairs_hook(pairs: list[tuple[str, object]]) -> dict[str, object]:
         parsed: dict[str, object] = {}
         for key, value in pairs:
@@ -1217,7 +1217,7 @@ def _strict_json_object(path: Path) -> dict[str, Any]:
         return parsed
 
     payload = json.loads(
-        Path(path).read_text(encoding="utf-8"),
+        raw.decode("utf-8"),
         object_pairs_hook=pairs_hook,
         parse_constant=reject_constant,
         parse_float=parse_float,
@@ -1225,6 +1225,11 @@ def _strict_json_object(path: Path) -> dict[str, Any]:
     if not isinstance(payload, dict):
         raise ValueError(f"{path}: payload must be one JSON object")
     return payload
+
+
+def _strict_json_object(path: Path) -> dict[str, Any]:
+    resolved = Path(path)
+    return _decode_strict_json_object(resolved.read_bytes(), path=resolved)
 
 
 def _v2_partial_manifest(paths: Sequence[Path]) -> list[dict[str, object]]:
@@ -1242,7 +1247,7 @@ def _v2_partial_manifest(paths: Sequence[Path]) -> list[dict[str, object]]:
     for path_value in paths:
         path = Path(path_value)
         raw = path.read_bytes()
-        payload = _strict_json_object(path)
+        payload = _decode_strict_json_object(raw, path=path)
         if payload.get("schema") != PARTIAL_SCHEMA:
             raise ValueError(f"{path}: partial manifest accepts only strict v2 shards")
         learner = payload.get("learner")
--- a/tests/test_upgd_ipmnist.py
+++ b/tests/test_upgd_ipmnist.py
@@ -4,7 +4,9 @@ CI-cheap: everything runs on tiny synthetic data. Real-MNIST benchmark runs
 happen only through ``python -m alberta_framework.benchmarks.upgd_ipmnist``.
 """
 
+import hashlib
 import json
+from pathlib import Path
 
 import jax.numpy as jnp
 import jax.random as jr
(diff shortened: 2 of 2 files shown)


kenjohnscreates (approved):

## What I verified

**Tree identity.** The PR diff applied onto current `main` (`58b6780`) produces a tree byte-identical to PR head `97ec382`, which contains current `main`.

**The race is real and the test is genuinely discriminating.** On `main`, `_v2_partial_manifest` reads each shard twice: `path.read_bytes()` feeds `size_bytes`/`sha256`, then `_strict_json_object(path)` re-reads via `read_text` to decode `learner`/`seed_id` — so a file swapped between the two reads yields a manifest whose identity belongs to the second file and whose digest belongs to the first. Failing-test-first reproduced: tests-only on `main` → exactly `test_v2_partial_manifest_binds_identity_and_digest_to_one_read` fails (1 failed, 48 deselected); full diff → **49/49 pass**. The monkeypatch construction (swap the payload inside the first `read_bytes` and assert the manifest still reports the original identity, size, and digest) pins the one-snapshot property directly rather than approximating it with timing.

**Fix shape is minimal and complete.** `_decode_strict_json_object(raw, path=path)` preserves every strictness property of the old path-based decoder (duplicate-key rejection via the pairs hook, constant rejection, float parsing), and the path-based `_strict_json_object` remains as a thin wrapper for single-read contexts. I checked the module's other call site — `_validated_partial_payload` (line 1269) reads once and hashes nothing alongside, so no analogous binding issue remains after this PR.

**Gates.** `ruff check` clean on both changed files; repo-config strict `mypy` → `Success: no issues found in 163 source files`. Pipeline hygiene is right too: the PR closes the author's own pre-filed issue #208 with an evidence-head, immutable log/domain-artifact links, and no `outputs/` contact.

## One documented-behavior observation (non-gating)

Decoding from `raw.decode("utf-8")` instead of `read_text(encoding="utf-8")` drops universal-newline translation before JSON parsing. For valid shards this is a no-op (`\r\n` between tokens is JSON whitespace), but a literal `\r` inside a JSON string — which text mode used to silently normalize to `\n` — is now rejected by `json.loads`' strict mode. That is a strictness increase on malformed input, consistent with the module's fail-closed ethos, and it makes the decoded content byte-consistent with the hashed snapshot, which is the point of the fix. Worth knowing if a legacy artifact ever decoded only by grace of newline translation.

Execution disclosure: I fully read the 97-line diff before executing anything, and ran it re-applied onto `main` in my own checkout rather than checking out the contributor branch.

— [kenjohnscreates]

deepanshu-yd (commented):
### PR Review (#228)

**Verdict**: LGTM. Excellent TOCTOU fix for artifact integrity.

**What good**:
- `_v2_partial_manifest` previously read shard file twice: once for digest/size and once for payload decoding. A file modification between reads could mismatch hash and identity.
- Refactored `_decode_strict_json_object(raw: bytes)` ensures hash, size, and parsed identity (`learner`, `seed_id`) are derived from the exact same byte buffer.
- Includes thorough reproduction test with file replacement between reads.

lalalune (approved):
Reviewed the exact head . The validator now reads each phase's manifest and shards from one captured snapshot, the mutation regression discriminates the previous mixed-read behavior, and the focused UPGD suite (49 tests), Ruff, and full strict mypy all pass. No pinned output or registered evidence source is changed. Ready once the final CI shard completes.

lalalune:
Review note correction: the approved exact head is `97ec38229fbd3191bb951be3485fccf1c72407e0`.

PR #232, opened 2026-08-16
fix(upgd-ipmnist): bind validator identity to one byte snapshot

### LINKED ISSUE #229: Bind UPGD-IPMNIST validator identity and digest to one byte snapshot
## Measurement-integrity bug

`validate_upgd_ipmnist_v2_artifact` rebuilds its `expected_manifest` by reading each shard twice:

1. `raw = path.read_bytes()` supplies `size_bytes` and `sha256`.
2. `_strict_json_object(path)` reopens the path for `learner` and `seed_id`.

If the path is replaced between those reads, the validator constructs a manifest row whose identity and digest never described the same byte snapshot. A valid artifact can then be rejected against that synthetic row with `v2 artifact partial manifest does not bind exact shard bytes`.

This is the validator-side counterpart to the producer race in #208 / #228, but it is a separate live site in disjoint files. #228 changes only `alberta_framework/benchmarks/upgd_ipmnist.py` and `tests/test_upgd_ipmnist.py`; it does not repair the validator reconstruction below.

## Deterministic reproduction on current `main`

A unit test substitutes a second-read identity for one shard after the validator has already captured its raw bytes. Current `main` rejects the otherwise valid artifact because the expected row combines snapshot B's identity with snapshot A's digest:

```text
AssertionError: ('v2 artifact partial manifest does not bind exact shard bytes',)
```

## Bounded fix

- Decode `learner` and `seed_id` from the same `raw` bytes used for `size_bytes` and `sha256` in the validator's `expected_manifest` loop.
- Add a regression proving a second-read identity swap cannot affect validation.
- Add a stable-shard control proving identities, sizes, and digests remain unchanged.
- Keep the change to:
  - `alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py`
  - `tests/test_upgd_ipmnist_nonpromoting.py`

## Non-goals

- Do not modify #208 / #228 producer files.
- Do not change schemas, protocols, seeds, benchmark results, source pins, or any `outputs/**` artifact.
- This remains permanently nonpromoting development validation; it makes no performance or scientific-promotion claim.

Client / agent tooling: Claude Code via CLIProxyAPI

ss251:
Fixes #229.

## What

The v2 UPGD-IPMNIST validator previously read every partial shard twice: once for digest/size and again for learner/seed identity. A file replacement between those reads could bind identity from one byte sequence to the digest of another. The validator now reads each shard once, strictly decodes that exact byte snapshot, and derives identity, size, and SHA-256 from the same immutable in-memory bytes.

The regression test replaces the old second-read helper with a forged identity. The fixed validator remains valid and proves the manifest still contains the stable shard's original learner/seed plus the exact digest and size of the bytes it read.

## Verification

Exact local head before publication: `8e6604c672617838c17e781ab4bb65707386cbb8`.

- Focused validator and strict-JSON suites — **69 passed**.
- Red proof against baseline production — deterministic failure: `v2 artifact partial manifest does not bind exact shard bytes`.
- Ruff — clean.
- Project-config mypy — clean.
- `git diff --check origin/main...HEAD` — clean.
- Fresh ownership scan immediately before issue publication found no open PR touching either changed file.

## Scientific-integrity scope

This changes validator construction only. It does not edit frozen outputs, registered evidence sources, protocols, seeds, benchmark results, or any `outputs/**` artifact, and it makes no scientific-performance or promotion claim.

Client / agent tooling: Claude Code via CLIProxyAPI

### DIFF
--- a/alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py
+++ b/alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py
@@ -909,7 +909,7 @@ def validate_upgd_ipmnist_v2_artifact(
         for path_value in partial_paths:
             path = Path(path_value)
             raw = path.read_bytes()
-            shard_payload = _strict_json_object(path)
+            shard_payload = _decode_strict_json_object(raw)
             expected_manifest.append(
                 {
                     "learner": shard_payload["learner"],
--- a/tests/test_upgd_ipmnist_nonpromoting.py
+++ b/tests/test_upgd_ipmnist_nonpromoting.py
@@ -20,6 +20,7 @@ from alberta_framework.benchmarks.upgd_ipmnist import (
     merge_legacy_v1_partial_results,
     merge_partial_results,
 )
+from alberta_framework.evaluation import upgd_ipmnist_nonpromoting as nonpromoting
 from alberta_framework.evaluation.upgd_ipmnist_nonpromoting import (
     EXPECTED_CONFIG,
     EXPECTED_HYPERPARAMETERS,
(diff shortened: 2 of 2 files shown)


nishimweprince (approved):
Independent review of `8e6604c672617838c17e781ab4bb65707386cbb8` against base `58b6780840e1ba3fd6fdaea4d49c906d6c3b55cb`: **approved**.

The change closes the bounded identity/digest race in #229: each expected-manifest shard is now decoded from the same immutable `raw` byte buffer used for the size and SHA-256 checks, so the validated identity and validated digest cannot come from different opens of that row.

Verification was performed on an archive of the exact head under macOS `sandbox-exec`, with a fresh HOME/TMP/cache, network denied, and the control checkout unreadable:

- Failing-test-first check (base production plus PR tests): `1 failed, 23 passed`; the new mixed-snapshot regression failed with `v2 artifact partial manifest does not bind exact shard bytes`.
- Exact-head green check: `tests/test_upgd_ipmnist_nonpromoting.py tests/test_strict_json_finite_numbers.py` — `69 passed`.
- Ruff on the changed production and test files — passed.
- Mypy on `alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py` — passed.

Non-blocking follow-up: the outer artifact validator still performs validation, merge, and expected-manifest reads as separate phases, so a broader path-replacement race between phases may remain. That is outside #229's row-level identity/digest scope and does not undermine this fix.

Client / agent tooling: codex
— [review-pr-232]

kenjohnscreates (approved):

Verified by re-applying the diff onto fetched `main` (`58b6780`) in a scratch worktree; the resulting tree is **byte-identical to PR head `8e6604c6`** (`git diff` against `pull/232/head` is empty). I did not check out the contributor branch.

**Defect confirmed on main.** In `validate_upgd_ipmnist_v2_artifact`'s expected-manifest loop, `raw = path.read_bytes()` feeds `size_bytes`/`sha256` while `_strict_json_object(path)` performs its own second `read_bytes()` for `learner`/`seed_id` — identity and digest can bind to different byte sequences. The fix decodes the already-held `raw` via the existing `_decode_strict_json_object` (module line 245), so all four manifest fields derive from one immutable snapshot. One-line source change, minimal and correct.

**Failing-test-first reproduced.** Tests-only applied to baseline source: `test_v2_expected_manifest_does_not_mix_digest_and_identity` fails deterministically with exactly `v2 artifact partial manifest does not bind exact shard bytes` (the monkeypatched second-read identity swap poisons the expected manifest). The monkeypatch-the-helper design is genuinely discriminating: with the fix, `_strict_json_object` is never invoked on the shard path, so the swap is inert and the manifest provably holds the original identity plus the exact digest/size of the bytes read.

**Full diff green.** `tests/test_upgd_ipmnist_nonpromoting.py` 24/24; together with `tests/test_upgd_ipmnist.py` **72/72**. `ruff check .` clean; project-config strict mypy clean at **163 source files** (matches baseline count — no regression).

**No duplication with #228.** Same one-byte-snapshot defect class, but a genuinely different site: #228 fixes `_v2_partial_manifest` in `alberta_framework/benchmarks/upgd_ipmnist.py`; this PR fixes the validator's expected-manifest reconstruction in `alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py`. Zero file overlap; the two complement each other and can merge in either order.

**Pinned-artifact exposure checked.** `outputs/upgd_ipmnist/nonpromoting_receipt.v2.json` pins `validator_sha256 = ec9d7db5…`, which equals the **current** live module bytes — but the binding target is the immutable snapshot `outputs/upgd_ipmnist/strict_validator_snapshot.v1.py` (`validator_path`), with `active_validator_path` informational; the receipt-binding test hashes the snapshot, not the live file. The v1→v2 receipts already record different validator hashes, so validator evolution without receipt invalidation is the designed pattern. No frozen `outputs/` artifact is touched or invalidated. This module is also absent from the five-claim registered source inventory (`evaluation/evidence_manifest.py`), and the change is pure-Python validator logic — no JAX graph or measured benchmark path is perturbed.

**Non-gating note.** Within the same `partial_validation.valid` block, `merge_partial_results(partial_paths)` still performs its own independent reads, so `results`-derived checks (config/seed facts) can in principle bind different bytes than the manifest under an adversarial mid-validation swap. The manifest binding — this PR's claim — is now sound; closing the results-vs-manifest window would require threading the snapshot bytes into the merge and is reasonably out of scope here.

Compute receipt: 1512883 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

lalalune (approved):
Approved after exact-head review of the validator path and surrounding reconstruction logic. Identity, byte size, and digest now come from the same raw shard snapshot while strict JSON semantics remain intact. The wider multi-phase validator lifecycle is a separate residual concern, not the row-level split-read defect in #229. Focused 69-test coverage, Ruff, mypy, diff checks, and 20/20 CI passed.

0xcii:
Post-merge independent verification for `8e6604c672617838c17e781ab4bb65707386cbb8` (issue #229).

- The implementation binds shard identity (`learner`, `seed`), digest, and size to the same byte snapshot in `expected_manifest`, matching the issue's stated scope.
- The regression coverage exercises mutation between the former reads and verifies the stable-shard digest path.
- Untrusted code was tested in a disposable container with no credentials or host mounts, runtime networking disabled, a read-only root filesystem, and bounded CPU, memory, and PIDs.
- `python3 -m pytest tests/test_upgd_ipmnist_nonpromoting.py tests/test_evaluation_artifact_strict_json.py tests/test_strict_json_finite_numbers.py -q -p no:cacheprovider -o addopts=""`: **92 passed**.
- The two new snapshot-focused tests passed independently: **2 passed, 22 deselected**.

No blocking defect was found. The pull request merged while this independent review was being finalized, so this comment records the verification evidence without representing itself as a pre-merge approval.

Client / agent tooling: codex
— [upgd-validator-byte-snapshot-review]

PR #257, opened 2026-08-16
fix(streams): reject invalid logarithmic scale bounds

### LINKED ISSUE #256: make_scale_range(log_spaced=True) accepts non-positive/non-finite endpoints, produces NaN scales
## What's wrong

The public \`make_scale_range(..., log_spaced=True)\` accepted negative, zero, or non-finite endpoints and passed them directly to \`jnp.log10\`. \`log10(x)\` is only real-valued for \`x > 0\`, so a negative or zero endpoint produces \`NaN\` internally, and JAX returns \`NaN\` scale factors rather than raising. Those factors can be passed directly to \`ScaledStreamWrapper\`, whose multiplication silently turns every generated observation into \`NaN\`, corrupting downstream measurement.

\`\`\`text
min_scale=0.0: array([  0.,   0.,   0.,   0., 100.], dtype=float32)
all_finite=True
min_scale=-1.0: array([ nan,  nan,  nan,  nan, 100.], dtype=float32)
all_finite=False
\`\`\`

This formula has no cited paper — the reference is the first-principles domain of the real logarithm: \`log10(x)\` is defined only for \`x > 0\`. The existing implementation itself constructs the range as \`logspace(log10(min_scale), log10(max_scale), ...)\`.

## Reachability

\`make_scale_range\` is genuinely package-root exported (\`alberta_framework/streams/__init__.py\` and \`alberta_framework/__init__.py\`, both import and \`__all__\`). Real call path:

\`\`\`python
from alberta_framework import make_scale_range
scales = make_scale_range(feature_dim, min_scale=user_min, max_scale=user_max)
wrapped = ScaledStreamWrapper(inner_stream, scales)
wrapped.step(...)  # -> timestep.observation * scales
\`\`\`

Honest note: there are no internal runtime callers outside the implementation's own docstring example — this is a gap in a public library utility, reachable by any caller supplying user/config-driven bounds, not an active benchmark path currently producing wrong numbers in a shipped pipeline.

## Fix

Reject non-finite or non-positive endpoints before evaluating the logarithm, scoped only to the \`log_spaced=True\` branch. Linear spacing is unchanged and continues to allow a zero endpoint (verified: \`make_scale_range(3, min_scale=0.0, max_scale=100.0, log_spaced=False)\` still works identically).

PR incoming with a parameterized regression test (negative lower bound, negative upper bound) and a mutation-resistance transcript.

Svector-anu:
Closes #256.

## What changed

`make_scale_range(..., log_spaced=True)` previously sent unchecked host values into a
float32 logarithmic calculation. The failure modes differ by input:

- a zero endpoint makes `log10(0)` negative infinity, and JAX returns finite zero scales
  (not NaNs);
- a negative endpoint produces NaNs;
- a host-finite positive endpoint can still narrow outside float32: `1e-50` becomes zero,
  while `1e100` becomes positive infinity and produces NaN/infinite scales.

The logarithmic branch now:

- requires each endpoint to be finite, positive, and representable as a normal float32
  value, matching JAX's effective logarithm domain;
- canonicalizes accepted endpoints to float32 before range construction;
- constructs an exact-endpoint float32 geometric range; and
- fails closed unless every generated scale is finite, positive, monotone in the requested
  direction, and within the canonical endpoint interval.

The `log_spaced=False` branch is unchanged and still permits a zero endpoint.

## Reachability and scope

`make_scale_range` is exported from both `alberta_framework.streams` and the package root.
Its output is intended for `ScaledStreamWrapper`, where invalid scales directly corrupt
observations. There are no internal runtime callers beyond the implementation's docstring
example, so this repairs a public utility contract rather than claiming an active benchmark
failure or scientific result. No evidence artifacts are changed.

## Regression coverage

The retained matrix exercises zero, negative, NaN, positive infinity, negative infinity,
below-float32, and above-float32 inputs in both endpoint positions. Positive controls cover
ascending and descending ranges, the normal float32 endpoints, exact canonical endpoints,
finite/positive/ordered/in-range postconditions, linear zero behavior, and fail-closed
handling of invalid generated values.

## Verification

Validated at `a21a799d5a04d61750a16c5e3468ac87f2d2227e`:

```text
$ python -m pytest tests/test_streams.py -q -o addopts=""
130 passed in 12.06s

$ python -m ruff check alberta_framework/streams/synthetic.py tests/test_streams.py
All checks passed!

$ python -m mypy
```

The repair is rebased on current `main` (`8b49cec714c97c58864f4b6c93cc799455c4c077`).

## Provenance

Original fix and tests: Svector-anu, with self-reported AI assistance in the initial PR.
Float32-domain hardening, generated-range postconditions, regression expansion, and
re-verification: lalalune/Codex.

### DIFF
--- a/alberta_framework/streams/synthetic.py
+++ b/alberta_framework/streams/synthetic.py
@@ -7,6 +7,7 @@ track and adapt.
 All streams use JAX-compatible pure functions that work with jax.lax.scan.
 """
 
+import math
 from typing import Any
 
 import chex
@@ -851,6 +852,13 @@ def make_scale_range(
     ```
     """
     if log_spaced:
+        if (
+            not math.isfinite(min_scale)
+            or not math.isfinite(max_scale)
+            or min_scale <= 0.0
+            or max_scale <= 0.0
+        ):
+            raise ValueError("log-spaced scale bounds must be finite and positive")
         return jnp.logspace(
             jnp.log10(min_scale),
             jnp.log10(max_scale),
--- a/alberta_framework/streams/synthetic.py
+++ b/alberta_framework/streams/synthetic.py
@@ -13,6 +13,7 @@ from typing import Any
 import chex
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Float, Int, PRNGKeyArray
 
@@ -20,6 +21,8 @@ from alberta_framework.core.types import TimeStep
 from alberta_framework.streams.base import ScanStream
 
 _INT32_MAX = 2**31 - 1
+_FLOAT32_TINY = float(np.finfo(np.float32).tiny)
+_FLOAT32_MAX = float(np.finfo(np.float32).max)
 
 
 def _require_positive_int(name: str, value: object) -> int:
@@ -34,6 +37,25 @@ def _require_positive_int(name: str, value: object) -> int:
             f"{name} must be a positive integer in [1, {_INT32_MAX}], got {value!r}"
         )
     return value
+
+
+def _require_normal_float32_scale(name: str, value: float) -> float:
+    """Return a positive bound with stable JAX float32 logarithm semantics."""
+    message = f"{name} must be finite, positive, and representable as a normal float32 value"
+    try:
+        normalized = float(value)
+    except (TypeError, ValueError, OverflowError) as exc:
+        raise ValueError(message) from exc
+    if (
+        not math.isfinite(normalized)
+        or normalized < _FLOAT32_TINY
+        or normalized > _FLOAT32_MAX
+    ):
+        raise ValueError(message)
+    narrowed = float(np.float32(normalized))
+    if not math.isfinite(narrowed) or narrowed < _FLOAT32_TINY or narrowed > _FLOAT32_MAX:
+        raise ValueError(message)
+    return narrowed
 
 
 @chex.dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
The source guard is correct, but the regression coverage does not prove the full issue/PR claim. The only new cases are negative finite bounds; zero, NaN, +Infinity, and -Infinity are all named by #256 and by this PR's description, yet none is exercised. Please extend the parameter matrix across both endpoint positions for zero and non-finite values, and keep a positive control proving log-spaced finite positive endpoints plus the existing linear zero-endpoint behavior. This should remain a test-only follow-up.

lalalune (approved):
Re-reviewed at exact head `86069c50c9a0c5f23c5f5fd85810f0ca008c5063`. The source guard rejects every issue-named invalid logarithmic endpoint, and the regression matrix now covers zero, NaN, positive infinity, and negative infinity in both endpoint positions, with finite-positive log and linear-mode controls retained. Mutation proof produced 10 failures without the guard; the focused suite passes 120 tests, Ruff passes, and full strict mypy passes for 163 source files. Ready once refreshed CI is green.

lalalune (changes requested):
The head changed materially after the approval at `86069c50`: it now adds a normal-float32-only policy, replaces JAX logspace with NumPy geomspace, and adds generated-value postconditions. The prior approval does not cover these API/numerical changes. Holding this exact head for refreshed boundary and compatibility review before merge.

lalalune (approved):
Re-reviewed exact repaired head cb09dff9. The float32 pre-cast regression is fixed by preserving canonical Python-float endpoints through NumPy interpolation and casting only outputs. Retained equal/adjacent/tiny/max cases: red 5/8 on parent, green 8/8; focused 135 + adjacent 20 passed; Ruff and strict mypy passed; 30,000 randomized equal/adjacent ranges satisfied endpoint/finite/positive/in-range/monotonic postconditions; merge-tree and evidence checks clean. Ready once exact-head CI is green.

PR #260, opened 2026-08-16
fix(step5): fail closed on invalid TD config

### LINKED ISSUE #255: Step 5 smoke reports finite success when malformed learning rates reject every update
## Outcome

**Fix** the production Step 5 differential-TD facade so malformed scientific scalar configuration fails closed instead of producing a smoke result that looks healthy while every learner update is rejected.

Observed on `main` at `58b6780840e1ba3fd6fdaea4d49c906d6c3b55cb`.

## Problem

`Step5AverageRewardTDConfig` accepts `NaN` for `step_size` and `average_reward_step_size`. It also accepts booleans through `from_dict` (`True` later narrows to a learning rate of `1.0`).

The core learner's transaction guard correctly refuses the resulting non-finite proposed states. However, `run_step5_smoke` only checks whether its reported arrays are finite. Rejected updates are reported as finite zeros, so the smoke result returns `finite=True` even though no learning occurred.

This is a measurement-integrity problem: a malformed Step 5 configuration can pass the public integration smoke while applying zero updates.

## Exact reproduction

Executed from a credential-free archive of the exact `main` SHA under macOS `sandbox-exec`, with network denied and the control checkout unreadable:

```text
step_size accepted True smoke_finite True
average_reward_step_size accepted True smoke_finite True
updates_applied [False, False, False]
td_errors [0.0, 0.0, 0.0]
roundtrip_bool_step_size True
```

The relevant facade boundary is `alberta_framework/steps/step5.py`:

- `Step5AverageRewardTDConfig` has no `__post_init__` validation;
- `from_dict` forwards arbitrary mappings directly to the dataclass;
- `run_step5_smoke` does not include `result.updates_applied` in its health decision.

## Scope and evidence constraint

Keep the production change in `alberta_framework/steps/step5.py` with focused coverage in `tests/test_step5_step6_production.py`.

Do **not** change `alberta_framework/core/average_reward.py` for this bounded fix: it is a registered source of the frozen continual-IA evidence claim, so changing it would invalidate the persisted artifact until an authorized protocol rerun. The facade can enforce the stricter public contract without touching that registered source.

No benchmark run, evidence seed, or pinned output is needed or authorized for this fix.

## Acceptance criteria

- [ ] `Step5AverageRewardTDConfig` rejects booleans, non-real values, `NaN`, and infinities for all three scientific scalar fields.
- [ ] It enforces the existing semantic domains: `step_size >= 0`, `average_reward_step_size >= 0`, and `trace_decay` in `[0, 1]`.
- [ ] Values that cannot narrow to a finite float32 are rejected before JAX execution.
- [ ] `from_dict` accepts exactly the keys emitted by `to_dict`; missing or extra fields fail closed with a stable `ValueError`.
- [ ] `run_step5_smoke` cannot return `finite=True` when any scan transaction was refused (either make refusal unreachable through validated configuration and assert it, or include `updates_applied` in the health gate).
- [ ] Existing valid configurations preserve their serialized values and current smoke trajectory.
- [ ] Add failing-test-first regressions for the reproduced `NaN`/boolean cases and the zero-update false-positive.
- [ ] Focused tests, Ruff, and mypy pass:

```bash
.venv/bin/python -m pytest tests/test_step5_step6_production.py -q -o addopts=""
.venv/bin/python -m ruff check alberta_framework/steps/step5.py tests/test_step5_step6_production.py
.venv/bin/python -m mypy
```

## Coordination

I intend to implement this after maintainer triage. I will not touch registered evidence sources, run benchmarks, or consume seeds.

AI provider/model/client: `openai` / `gpt-5.6-sol` / `codex`.

lalalune:
Closes #255.

## What changed

`Step5Config` now validates its public TD configuration at construction and deserialization:

- `gamma` and `trace_decay` must be finite real scalars in `[0, 1]`;
- `step_size` must be a finite non-negative real scalar;
- `n_steps` must be a positive integer representable by the signed int32 execution counter;
- booleans and coercive non-numeric values are rejected;
- accepted NumPy and other `Real` values are canonicalized through the actual float32 execution domain and stored as built-in Python floats, so `to_dict()` remains strict-JSON serializable and round-trips exact persisted values.

Validation happens before any learner or JAX state is constructed. Ordinary defaults and legal endpoints are preserved.

## Verification

- affected Step 5, pipeline, and average-reward panel: **102 passed**;
- full Step 5 facade file after the canonicalization correction: **64 passed**;
- Ruff: clean;
- strict mypy: clean;
- `git diff --check`: clean.

No `outputs/**`, frozen evidence artifact, registered protocol, or seed changed.

## Contribution provenance

- AI assistance: yes
- Client: Codex

### DIFF
--- a/alberta_framework/steps/step5.py
+++ b/alberta_framework/steps/step5.py
@@ -23,7 +23,10 @@ References:
 
 from __future__ import annotations
 
+import math
+import struct
 from dataclasses import asdict, dataclass
+from numbers import Real
 from typing import Any, cast
 
 import jax.numpy as jnp
@@ -36,6 +39,28 @@ from alberta_framework.core.average_reward import (
     run_differential_td_from_arrays,
 )
 
+_STEP5_CONFIG_KEYS = frozenset(
+    {"step_size", "average_reward_step_size", "trace_decay"}
+)
+_STEP5_CONFIG_KEYS_ERROR = (
+    "Step5AverageRewardTDConfig payload keys must be exactly "
+    "['average_reward_step_size', 'step_size', 'trace_decay']"
+)
+
+
+def _finite_float32_scalar(name: str, value: object) -> float:
+    """Validate a real scalar before the core narrows it to float32."""
+    if isinstance(value, bool) or not isinstance(value, Real):
+        raise ValueError(f"{name} must be a real scalar")
+    try:
+        scalar = float(value)
+        narrowed = struct.unpack("!f", struct.pack("!f", scalar))[0]
+    except (OverflowError, TypeError, ValueError, struct.error):
+        raise ValueError(f"{name} must narrow to a finite float32") from None
+    if not math.isfinite(scalar) or not math.isfinite(narrowed):
+        raise ValueError(f"{name} must narrow to a finite float32")
+    return scalar
+
 
 @dataclass(frozen=True)
 class Step5AverageRewardTDConfig:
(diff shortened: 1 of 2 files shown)


lalalune:
Independent exact-head review completed at `23d6d2d964043559923b919e0a35b9b959e02447`: READY with no remaining findings. The adversarial matrix accepted 12 valid Python/NumPy scalars, rejected 45 malformed cases and six exact domain straddles, and matched NumPy/JAX behavior below/at/above the float32 overflow midpoint. Focused tests: 60 passed; adjacent Step/core tests: 268 passed; full Ruff and strict mypy (163 files) passed. Base/head valid trajectory digest remained `065fd1fa…`, and forced refusal changed smoke health from true to false. No registered-source or open-PR path collision.

lalalune:
Exact-head follow-up review at `7623823e065fe6ee33333bc3dd06ec49dce97363`: the intervening canonicalization commit rounded existing built-in JSON values (for example `0.03` became `0.029999999329447746`), contrary to #255’s explicit serialized-value compatibility criterion. The final commit preserves exact built-in `int`/`float` values while canonicalizing non-built-in Real scalars to the validated float32 sink. Red compatibility probe reproduced on `ef37411`; targeted boundary/JSON cases 6 passed, affected Step5/pipeline/core panel 102 passed, full Ruff and strict mypy (163 files) passed, diff clean. No registered source or output changes. Ready pending exact-head CI.

PR #264, opened 2026-08-16
fix(nexting): scale before squaring to avoid float32 RMSE overflow

### LINKED ISSUE #263: per_horizon_rmse overflows float32 during intermediate squaring, reports inf for finite errors
## What's wrong

\`per_horizon_rmse\` squared float32 errors directly before averaging:

\`\`\`python
sq_err = (predictions - forward_returns) ** 2
return jnp.sqrt(jnp.mean(sq_err, axis=0))
\`\`\`

For a large but finite error (e.g. \`2.0e20\`, well within float32's representable magnitude of ~3.4e38), squaring it first (\`(2e20)^2 = 4e40\`) overflows float32 during the intermediate step, even though the correct final RMSE is itself finite and representable. This silently corrupts the nexting measurement harness by reporting infinity for a genuinely finite error magnitude.

Reproduced directly before any fix, checked against an independent float64 reference:

\`\`\`text
actual: [inf] finite: [False]
float64 reference: [2.00000004e+20] finite: [ True]
reference representable in float32: [ True]
\`\`\`

## Reachability (honest)

\`per_horizon_rmse\` is directly importable as a public module-level function (\`from alberta_framework.utils.nexting import per_horizon_rmse\`), reachable with real caller-controlled float32 inputs. Neither \`alberta_framework/__init__.py\` nor \`alberta_framework/utils/__init__.py\` re-exports it (\`multi_channel_horizon_returns\` from the same module is re-exported, but \`per_horizon_rmse\` is not), so this is not currently producing wrong numbers in a shipped package-root-exported pipeline — it's a real gap in a public module-level evaluation API.

## Fix

The nexting module cites Modayil, White, and Sutton (2014) for the evaluation protocol; this fix doesn't alter that protocol or its forward-view formula, it only evaluates the same documented RMSE definition using the algebraically equivalent scaled form \`max(abs(error)) * sqrt(mean((error / max(abs(error)))**2))\` for fully finite columns, avoiding the overflow-prone intermediate square. Columns containing NaN or infinity retain the original direct calculation so existing diagnostic propagation stays visible (a genuinely non-finite input should still report non-finite, this fix only closes the *spurious* overflow on finite-but-large inputs).

PR incoming with a regression test using two float32 errors of \`2.0e20\` (whose direct squares overflow float32, though the correct RMSE remains \`2.0e20\`), plus a mutation-resistance transcript.

Svector-anu:
Closes #263.

## Summary
`per_horizon_rmse` previously squared float32 errors directly, so finite errors such as `2e20` overflowed the intermediate square even though the correct RMSE remained representable.

The initial max-division repair was also incomplete: on XLA, dividing by a near-max float32 scale can lower through a reciprocal that underflows to zero, causing errors around `1e38` to report RMSE `0.0`. It also formed discarded overflow and zero-division operations that fail under JAX debug/checkify modes.

The corrected implementation:

- masks non-finite columns before arithmetic while retaining NaN/Inf diagnostics in the returned metric;
- uses `frexp` plus `ldexp` for exact power-of-two scaling instead of an arbitrary reciprocal;
- never squares unscaled large errors and never divides an all-zero column;
- supplies a scaled-domain custom JVP so eager and JIT reverse/forward autodiff remain finite and correct at extreme finite magnitudes;
- defines the RMSE derivative at an all-zero error column as zero;
- preserves ordinary eager/JIT results and the existing input/burn-in contract.

## Validation
- [x] `tests/test_nexting.py` — 61 passed
- [x] eager/JIT gradient regressions at `2e20`, `1e38`, and `3e38`
- [x] mixed-sign `3e38`/`-1e38` gradient against a float64 analytic reference
- [x] explicit all-zero eager/JIT reverse-mode and forward-mode zero convention
- [x] full Ruff
- [x] strict mypy — 163 source files
- [x] `git diff --check`
- [x] no `outputs/**` changes

Regressions also cover 257 values spanning `[-1e38, 1e38]` against a float64 reference, all-zero columns, eager/JIT agreement, JAX debug modes, checkify float checks, and retained NaN/Inf visibility.

Backend caveat: on the tested JAX CPU backend, the smallest float32 subnormal error is flushed to zero before the stable RMSE scaling stage. The smallest normal float32 value is preserved. This PR does not claim subnormal preservation and does not alter that pre-existing backend behavior.

The protocol and forward-view equations are unchanged; this is a numerically stable evaluation and derivative of the same RMSE definition.

### DIFF
--- a/alberta_framework/utils/nexting.py
+++ b/alberta_framework/utils/nexting.py
@@ -139,6 +139,41 @@ def _validate_rmse_inputs(
     return predictions.shape[0]
 
 
+def _scaled_rmse_terms(errors: Array) -> tuple[Array, Array, Array]:
+    """Return power-of-two scale, scaled errors, and their RMS."""
+    scale = jnp.max(jnp.abs(errors), axis=0)
+    _, exponent = jnp.frexp(scale)
+    scaled_errors = jnp.ldexp(errors, -exponent)
+    scaled_rmse = jnp.sqrt(jnp.mean(scaled_errors**2, axis=0))
+    return exponent, scaled_errors, scaled_rmse
+
+
+@jax.custom_jvp
+def _finite_rmse(errors: Array) -> Array:
+    """Compute finite-column RMSE without overflow or reciprocal underflow."""
+    exponent, _, scaled_rmse = _scaled_rmse_terms(errors)
+    return jnp.ldexp(scaled_rmse, exponent)
+
+
+@_finite_rmse.defjvp
+def _finite_rmse_jvp(
+    primals: tuple[Array],
+    tangents: tuple[Array],
+) -> tuple[Array, Array]:
+    """Differentiate in the scaled domain so huge finite gradients stay valid."""
+    (errors,), (errors_dot,) = primals, tangents
+    exponent, scaled_errors, scaled_rmse = _scaled_rmse_terms(errors)
+    rmse = jnp.ldexp(scaled_rmse, exponent)
+    # Zero is a valid subgradient for an all-zero column and avoids a 0 / 0.
+    safe_scaled_rmse = jnp.where(
+        scaled_rmse > 0.0,
+        scaled_rmse,
+        jnp.ones_like(scaled_rmse),
+    )
+    rmse_dot = jnp.mean(scaled_errors * errors_dot, axis=0) / safe_scaled_rmse
+    return rmse, rmse_dot
+
+
 def per_horizon_rmse(
     predictions: Float[Array, "T H"],
     forward_returns: Float[Array, "T H"],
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking numerical defect at exact head 69fb6824: constant finite float32 errors 1e38 and 3e38 return RMSE 0 eager/JIT because errors/scale underflows; the old errors**2 branch still executes, tripping jax_debug_infs, while zero errors trip jax_debug_nans via inactive 0/0. Autodiff is also wrong ([1,1] rather than [.5,.5] for two equal 2e20 errors; JIT grad at 1e38 is NaN). Please mask before direct squaring, use a stable non-subnormal divisor with stopped-gradient conditioning, and retain eager/JIT/debug/gradient regressions at near-max finite float32.

lalalune (approved):
Re-reviewed exact repaired head ad4035ea. Power-of-two scaling fixes finite float32 forward values through max without discarded overflow/0-divide operations; a scaled-domain custom JVP fixes the remaining live-head gradient failures (1e38 had zero gradient; 3e38/mixed huge values had NaN). Red reproduced on 1d7446a; green eager+JIT gradients match float64 reference. Focused blocker 5 passed; focused+adjacent 120 passed; full Ruff and strict mypy passed; diff/evidence hygiene clean. Ready once exact-head CI is green.

lalalune (approved):
Exact-head refresh for fad2a2ed: the only delta after the reviewed ad4035ea implementation is 38 lines of retained gradient coverage. It pins equal huge values at 2e20/1e38/3e38 and the zero reverse/forward derivative convention. Exact final file: 61 passed; focused Ruff clean. Prior exact implementation gates remain 120 focused+adjacent, full Ruff, and strict mypy. READY pending exact-head CI.

PR #266, opened 2026-08-16
fix(ftl-world-model): reject non-finite ridge and prediction_clip

### LINKED ISSUE #265: SparseFTLWorldModelConfig accepts NaN ridge/prediction_clip, poisons ridge-regression solve
## What's wrong

\`SparseFTLWorldModelConfig.__post_init__\` checked \`ridge <= 0.0\` and \`prediction_clip <= 0.0\` but never checked finiteness. Every ordered comparison with \`NaN\` is \`false\`, so a \`NaN\` value passed both guards.

Both fields feed real downstream sinks:

- \`ridge\` is added to the diagonal of a Gram matrix in a ridge-regression solve: \`active_gram + cfg.ridge * jnp.eye(...)\`.
- \`prediction_clip\` bounds a \`jnp.clip\` call on predicted state deltas: \`jnp.clip(delta, -cfg.prediction_clip, cfg.prediction_clip)\`.

A \`NaN\` in either poisons the corresponding computation instead of being rejected at construction time.

## Reachability (honest)

\`SparseFTLWorldModelConfig\` is reachable via direct module import (\`from alberta_framework.core.ftl_world_model import SparseFTLWorldModelConfig\`), confirmed working. It is **not** re-exported from either \`alberta_framework/__init__.py\` or \`alberta_framework/core/__init__.py\` — grepped both files directly, no match. So this is a gap in a direct-module public API, not a package-root export.

## Fix

Both \`ridge\` and \`prediction_clip\` now require \`math.isfinite(...)\` in addition to the existing positivity check.

PR incoming with two new parametrized cases (\`ridge=NaN\`, \`prediction_clip=NaN\`) added to the existing \`test_config_rejects_invalid_parameters\` suite, plus a mutation-resistance transcript.

Svector-anu:
Closes #265.

## Summary

`SparseFTLWorldModelConfig` now validates `ridge` and `prediction_clip` in the float32 domain used by the model, instead of only checking the host scalar:

- rejects booleans and non-`numbers.Real` inputs with a stable `ValueError`;
- rejects NaN, infinities, non-positive values, float32 subnormals/underflow-to-zero, and finite host values that overflow to float32 infinity;
- canonicalizes accepted NumPy/integer real scalars to built-in `float`, so `to_config()` is strict-JSON-safe and roundtrips through `json.dumps(..., allow_nan=False)`;
- preserves ordinary accepted values while failing before invalid values reach the JAX ridge solve or clipping bound.

The original NaN fix is retained as rebased commit `349ed60`; the complete execution-domain hardening is commit `64afa79`. The contributor branch was rebased onto `origin/main` at `9f53ba0` before verification and push.

## Downstream reachability

Both fields feed float32 execution directly:

- `ridge` is added to the active Gram block before `jnp.linalg.solve`. A subnormal admitted by the old host-only guard is ineffective in the solve and the regression test demonstrates non-finite learned weights.
- `prediction_clip` is the `jnp.clip` bound. A finite host value above the float32 range narrows to infinity, allowing an overflowed raw prediction to escape as infinity.

The config remains reachable through the direct module API (`alberta_framework.core.ftl_world_model`); it is not newly re-exported from package roots.

## Red-first regression evidence

Before the implementation change, the new focused cases produced:

```text
19 failed, 20 deselected in 2.63s
```

Failures covered built-in/NumPy booleans, strings/complex/object inputs, float32 subnormal and underflow-to-zero values, finite host overflow, NumPy scalar canonicalization/JSON roundtrip, and both live execution sinks. The two execution-level tests reached the poisoned ridge-update and prediction paths when the constructor failed to reject their values.

## Verification on final rebased tree

```text
$ .venv/bin/python -m pytest tests/test_ftl_world_model.py tests/test_ftl_decision_fidelity.py -q -o addopts=""
63 passed in 20.48s

All checks passed!

$ .venv/bin/python -m mypy

$ git diff --check origin/main...HEAD
# clean
```

## Evidence and artifact status

No `outputs/**` file was edited. `alberta_framework/core/ftl_world_model.py` is a registered source, so the evidence registry correctly returns exit 2 / `invalid`; this is expected source drift and is not evidence promotion. The frozen artifact records source SHA-256 `10de4526e967c400c6f2732f04dca63c0f0f7e22d44324fad85d064245e9f976`; this head has `2da0894e089ea903fbe4dee5dc57ed0e50e8714d62d4022df7e204bacfd76c00`. Existing pinned artifacts were left immutable and no scientific protocol was rerun.

64afa796dd6bcd2f15693bc00befb75529e2c9a9

```text
red: 19 failed, 20 deselected
final focused: 63 passed
full Ruff: clean
strict mypy: 163 source files clean
```

## Provenance

The contributor's original change is preserved in the rebased commit stack. Float32-domain hardening, regression expansion, rebase, and verification were performed by OpenAI Codex in this maintenance pass. Attribution is self-reported.

### DIFF
--- a/alberta_framework/core/ftl_world_model.py
+++ b/alberta_framework/core/ftl_world_model.py
@@ -34,14 +34,36 @@ from __future__ import annotations
 import dataclasses
 import functools
 import math
+from numbers import Real
 from typing import Any
 
 import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Float, Int
+
+_FLOAT32_TINY = float(np.finfo(np.float32).tiny)
+
+
+def _finite_positive_normal_float32(name: str, value: object) -> float:
+    """Return a concrete real after validation in the model's execution dtype."""
+    message = f"{name} must be a finite positive normal float32"
+    if isinstance(value, bool) or not isinstance(value, Real):
+        raise ValueError(message)
+    try:
+        concrete = float(value)
+    except (TypeError, ValueError, OverflowError) as exc:
+        raise ValueError(message) from exc
+    if not math.isfinite(concrete) or concrete <= 0.0:
+        raise ValueError(message)
+    with np.errstate(over="ignore", under="ignore", invalid="ignore"):
+        narrowed = float(np.float32(concrete))
+    if not math.isfinite(narrowed) or narrowed < _FLOAT32_TINY:
+        raise ValueError(message)
+    return concrete
 
 
 @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


lalalune (approved):
Reviewed exact rebased head 64afa796. The initial host-finite fix was incomplete: tiny positive ridge values reached JAX as zero and poisoned the solve; oversized finite clip values reached float32 as infinity. This head now rejects bool/non-real, nonpositive, nonfinite, subnormal/underflow, and overflow values before JAX, preserves ordinary Python float serialization, and JSON-canonicalizes accepted NumPy scalar values. Independent exact-head: 63 focused/FTL tests passed, full Ruff and strict mypy passed, adversarial scalar/sink/serialization matrix passed, diff clean. Registered `ftl_world_model.py` changes again, so the already-invalid FTL claim remains invalid; no artifact is revalidated or modified. READY pending exact-head CI.

PR #274, opened 2026-08-16
fix(core): ExperientialMemory.query neighbor weights use real score_sum for tiny positive sum

### LINKED ISSUE #253: ExperientialMemory.query gives tiny positive neighbor scores a near-zero weight instead of 1.0

sharkwon:
## Summary

Fixes #253 — ExperientialMemory.query gives tiny positive neighbor scores a near-zero weight instead of 1.0.

## Changes

- `alberta_framework/core/experiential_memory.py`: Replace unconditional `1e-12` floor with `jnp.where(score_sum > 0.0, score_sum, 1.0)` in neighbor weight normalization
- `tests/test_experiential_memory_253.py`: New regression test (single accepted neighbor at sub-floor similarity)

## Validation behavior

| Input | Before | After |
|-------|--------|-------|
| `score_sum == 0.0` | 1e-12 floor → weight ~1.0 | 1.0 fallback → weight 1.0 ✓ |
| `score_sum == 2.3e-16` (exp(-36)) | **1e-12 floor → weight ~2e-4** | **real score_sum → weight ~1.0** ✓ |
| `score_sum == 1.0` | 1.0 | 1.0 ✓ |

## Checks

- `pytest tests/test_experiential_memory.py tests/test_experiential_memory_policy.py` — 74/74 pass
- `pytest tests/test_experiential_memory_253.py` — 1/1 pass
- `ruff check` — clean

### DIFF
--- a/alberta_framework/core/experiential_memory.py
+++ b/alberta_framework/core/experiential_memory.py
@@ -840,7 +840,10 @@ class ExperientialMemory:
         neighbor_mask = jnp.isfinite(top_scores) & (top_scores > 0.0)
         positive_scores = jnp.where(neighbor_mask, top_scores, 0.0)
         score_sum = jnp.sum(positive_scores)
-        neighbor_weights = positive_scores / jnp.maximum(score_sum, 1.0e-12)
+        # Use the real score_sum when positive; fall back to 1.0 only when exactly zero.
+        # The previous unconditional 1e-12 floor corrupted weights for small-but-nonzero
+        # score_sum (e.g. exp(-36) ~ 2e-16), producing weights ~2e-4 instead of 1.0.
+        neighbor_weights = positive_scores / jnp.where(score_sum > 0.0, score_sum, 1.0)
 
         neighbor_similarities = similarities[indices]
         neighbor_reliabilities = effective_reliabilities[indices]
--- a/tests/test_experiential_memory_253.py
+++ b/tests/test_experiential_memory_253.py
@@ -1,0 +1,122 @@
+"""Regression test for #253: weight normalization must not floor with 1e-12.
+
+When `score_sum` is a tiny positive float (e.g. exp(-36) ~ 2e-16 from a
+single accepted neighbor at a sub-floor similarity), the previous
+`jnp.maximum(score_sum, 1e-12)` floor produced weights ~2e-4 instead of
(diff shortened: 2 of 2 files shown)


lalalune:
Closing as a duplicate of merged PR #254. Current `main` already uses the real positive `score_sum` with an exact-zero fallback in `ExperientialMemory.query`, and it retains the focused regression coverage. Rebasing this branch produces the same source change plus a redundant standalone test module, so there is no additional production delta to merge.

PR #276, opened 2026-08-16
fix(interaction-features): exclude inactive tasks from top-k utility

### LINKED ISSUE #275: FixedBudgetInteractionLearner's unweighted top-k utility lets inactive-task zero placeholders outrank real active signals
## What's wrong

\`FixedBudgetInteractionLearner\`'s unweighted \`topk\` utility reducers zero out inactive task rows (\`weighted_activity = weighted_activity * task_weights[:, None]\`, where \`task_weights\` is 0 for inactive tasks) before selecting the top-k values via \`jnp.sort(...)[-k:, :]\`.

For the **signed** feature-signal reducer, this is a real correctness bug: a synthetic zero placeholder for an inactive task can outrank a legitimate *negative* measurement from an active task in the naive ascending-sort top-k selection, hiding the real (harmful) signal entirely.

\`\`\`text
active=[True, False, False], utility_top_k=2
signed_reductions=[[-4.0], [99.0], [99.0]]  # tasks 2,3 inactive, zeroed to 0
reported [0.]
expected [-4.]
\`\`\`

With active task 1 reporting \`-4\` and inactive tasks 2/3 zeroed to \`0\`, sorted ascending is \`[-4, 0, 0]\`, top-2 selects \`[0, 0]\`, mean = \`0\` — the genuine \`-4\` signal is silently dropped.

The same eligibility gap also **dilutes** nonnegative current-utility values toward zero when fewer than \`k\` tasks are active: with only 1 active task producing a real value of \`6.0\` and \`top_k=2\`, the old code averaged in a zero-padded inactive slot, returning \`3.0\` instead of the correct \`6.0\`.

## Reachability

\`FixedBudgetInteractionLearner\` is genuinely package-root exported (\`alberta_framework/__init__.py\`, both import and \`__all__\`). The affected signed reducer is called from the public \`.update\` path after \`one_step_output_loss_reduction\` computes per-task feature measurements. It's also constructed by \`alberta_framework/recurring_feature_gate.py\` and \`alberta_framework/core/prototype_feature_lifecycle.py\`. Real non-hardcoded-safe inputs reach the bug whenever active-task masking leaves fewer than \`utility_top_k\` valid targets and an active task's signed loss reduction is negative.

## Fix

Inactive tasks are now masked to \`-inf\` (not zero) before top-k selection, guaranteeing an inactive placeholder can never outrank a genuine active value regardless of sign. The mean now divides by the actual count of genuinely-selected finite (active) entries, not blindly by \`k\`. The historical \`utility_task_balancing="none"\` branch is left byte-for-byte algebraically unchanged.

PR incoming with a regression test proving a single active task's real \`-4.0\` signal survives \`top_k=2\` selection against four inactive placeholders, plus a mutation-resistance transcript.

Svector-anu:
Closes #275.

## What changed

The unweighted active-task `topk` reducers no longer let zero placeholders from inactive tasks outrank a real negative active signal or dilute a sparse active mean. Inactive rows are ranked as `-inf`, and the denominator is the exact `min(active_count, k)`.

The corrected implementation also preserves two fail-closed invariants:

- an all-inactive task set uses a safe denominator and returns exact zero without forming `0 / 0` under eager, JIT, `JAX_DEBUG_NANS`, or checkify execution;
- non-finite values from an active task remain non-finite and cannot be mistaken for an inactive placeholder.

The historical `utility_task_balancing="none"` algebra remains unchanged.

## Reachability

`FixedBudgetInteractionLearner` is exported from the package root and is constructed by the recurring-feature and Prototype feature-lifecycle paths. The affected reducers run inside its public update path.

## Verification

Validated at `477ccf4bc990591ec40c01049c40d7ab46f7c44f`, rebased on current `main`:

- interaction top-k, horizon, retirement, Prototype lifecycle, and recurring-feature panels: **71 passed**;
- targeted Ruff: clean;
- strict mypy: clean;
- `git diff --check`: clean.

Regression coverage includes negative active signals, sparse positive means, all-inactive eager/JIT/debug/checkify execution, and NaN/positive-infinity/negative-infinity fail visibility.

No `outputs/**` artifact or evidence protocol changed.

## Contribution provenance

Original report and fix: Svector-anu. Additional adversarial review, safe-denominator correction, active-count semantics, fail-visible invalid-signal coverage, rebase, and verification: lalalune/Codex.

### DIFF
--- a/alberta_framework/core/interaction_features.py
+++ b/alberta_framework/core/interaction_features.py
@@ -1385,10 +1385,16 @@ class FixedBudgetInteractionLearner:
                 return jnp.mean(jnp.sort(weighted_activity, axis=0)[-k:, :], axis=0)
             scores = jnp.where(active_mask[:, None], weighted_activity, -jnp.inf)
             selected, _ = jax.lax.top_k(jnp.swapaxes(scores, 0, 1), k)
-            selected_mask = jnp.isfinite(selected)
-            selected_count = jnp.sum(selected_mask, axis=1)
-            selected_sum = jnp.sum(jnp.where(selected_mask, selected, 0.0), axis=1)
-            safe_count = jnp.maximum(selected_count, 1.0)
+            selected_count = jnp.minimum(
+                jnp.sum(active_mask.astype(jnp.int32)),
+                k,
+            )
+            selected_mask = jnp.arange(k) < selected_count
+            selected_sum = jnp.sum(
+                jnp.where(selected_mask[None, :], selected, 0.0),
+                axis=1,
+            )
+            safe_count = jnp.maximum(selected_count, 1)
             return jnp.where(selected_count > 0, selected_sum / safe_count, 0.0)
 
         if self._utility_task_balancing == "active":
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
The eligibility fix is directionally correct, but the new zero-active-task path forms `0 / 0` inside `jnp.where`. Exact PR-head repro with `JAX_DEBUG_NANS=1`, `utility_aggregation="topk"`, `utility_task_balancing="active"`, and an all-false active mask raises `FloatingPointError: invalid value (nan) encountered in div` at the new `selected_sum / selected_count` expression. Ordinary execution happens to select the outer zero, but debug/checkify execution still observes the invalid intermediate. Use a safe denominator such as `maximum(selected_count, 1)` before division in both reducers, retain the explicit zero output, and add all-inactive eager/JIT/JAX_DEBUG_NANS/checkify regressions alongside the signed and dilution cases.

lalalune (approved):
Re-reviewed exact contributor head 477ccf4bc990591ec40c01049c40d7ab46f7c44f. It now counts active tasks independently of signal finiteness, guards the zero-active denominator, excludes inactive sentinels, and leaves any active NaN/±inf fail-visible for the learner's atomic rejection. Verified 14 focused tests, 640 finite reference cases, reachable inverse-frequency overflow rejection, eager/JIT/grad behavior, all-inactive JAX_DEBUG_NANS/checkify coverage, full Ruff, strict mypy over 163 files, clean diff, and conflict-free current-main composition. This edits a registered source, so the already-invalid recurring/scale-robust claims remain invalid with new source drift; no artifact is revalidated.

PR #289, opened 2026-08-16
fix(streams): narrow original reals directly to float32

### LINKED ISSUE #288: Float32 stream validators double-round extended-precision Real inputs
## What's wrong

Merged PR #268 (main merge 61e4f1c49c32a962f69e889607da5af1f8bcb2b8) added float32-domain validation for FrequencyMismatchStream and Pavlovian noise_std, but both helpers first convert the original Real to a Python binary64 float and only then narrow to float32:

- alberta_framework/streams/out_of_class.py::_require_positive_float32
- alberta_framework/streams/pavlovian.py::_require_nonnegative_float32

For concrete extended-precision Real inputs, that is not the execution sink's conversion. Near float32 halfway/overflow boundaries, longdouble -> float64 -> float32 can differ from direct longdouble -> float32. The facade can therefore reject a value the actual NumPy/JAX float32 sink represents finitely, or canonicalize it to the wrong float32 value.

This was found by independent review before #268 merged, but the exact reviewed head merged while that review was still completing.

## Required fix

Narrow the original scalar directly once with NumPy's float32 conversion under a warning-suppressed context, reject non-scalar/non-finite results, check the original Real domain before narrowing where necessary, and store the resulting built-in float. Do not route through Python float before the execution-domain conversion.

Add retained regressions for both stream helpers at:

- a long-double value whose direct float32 conversion differs from the binary64-mediated conversion;
- the finite float32 overflow boundary and adjacent rejected value;
- existing bool/non-Real/domain/underflow behavior.

Keep the change to the affected stream/test files. Do not modify or regenerate outputs or evidence artifacts.

lalalune:
Closes #288

This fail-closed follow-up removes both extended-precision and exact-Rational double rounding left by merged PR #268. A shared host-scalar converter now rounds exact integer ratios directly to IEEE-754 binary32 with ties-to-even, including normal, subnormal, signed-zero, and overflow boundaries. Frequency-mismatch and Pavlovian validators consume that exact sink value while preserving their distinct domains.

Retained coverage includes:
- Fraction values immediately below, exactly at, and immediately above a normal midpoint
- the float32 overflow midpoint, including the finite value immediately below it
- the half-minimum-subnormal tie and the value immediately above it
- signed zero in Python, NumPy float32, and NumPy long-double forms
- strict rejection of bool aliases by the shared converter
- strict-positive frequency bounds versus non-negative Pavlovian noise semantics

Validation on rebased exact head `529fe0d`:
- focused affected panel: 148 passed
- independent integer-rounding oracle: 30,006 randomized/extreme exact ratios plus signed-zero variants, bit-identical
- Ruff: clean
- strict mypy: 164 source files clean
- git diff --check: clean
- no output artifacts changed
- changed stream sources and the new internal helper are not registered by the five-claim evidence manifest; inherited registry status is unchanged (pre-existing invalid)

### DIFF
--- a/alberta_framework/_float32.py
+++ b/alberta_framework/_float32.py
@@ -1,0 +1,117 @@
+"""Exact host-scalar conversion helpers for float32 JAX sinks."""
+
+from __future__ import annotations
+
+import math
+import struct
+from numbers import Integral, Real
+
+
+def _round_quotient_ties_to_even(numerator: int, denominator: int) -> int:
+    quotient, remainder = divmod(numerator, denominator)
+    doubled_remainder = remainder * 2
+    if doubled_remainder > denominator or (
+        doubled_remainder == denominator and quotient % 2 == 1
+    ):
+        return quotient + 1
+    return quotient
+
+
+def _float32_from_ratio(
+    numerator: int,
+    denominator: int,
+    *,
+    negative_zero: bool,
+) -> float:
+    """Round an exact ratio to IEEE 754 binary32 with ties-to-even."""
+    if denominator < 0:
+        numerator = -numerator
+        denominator = -denominator
+    if denominator == 0:
+        raise ValueError("ratio denominator must be nonzero")
+
+    negative = numerator < 0 or (numerator == 0 and negative_zero)
+    magnitude = abs(numerator)
+    sign_bits = int(negative) << 31
+    if magnitude == 0:
+        bits = sign_bits
+    else:
+        exponent = magnitude.bit_length() - denominator.bit_length()
+        if exponent >= 0:
+            if magnitude < denominator << exponent:
+                exponent -= 1
+        elif magnitude << -exponent < denominator:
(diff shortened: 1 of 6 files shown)


lalalune:
Exact-head adversarial review found one remaining hole in the advertised direct-`Real` narrowing. `Fraction(1,1) + Fraction(1,2**24) + Fraction(1,2**60)` is just above the float32 midpoint and should round to `nextafter(float32(1), 2)`, but NumPy converts an object `Fraction` through binary64 and returns `1.0`. The FrequencyMismatch constructor then rejects a genuinely ordered `[1, nextafter(1,2))` interval; Pavlovian stores the wrong scalar. Since `Fraction` is accepted by the `numbers.Real` guard, this needs exact Rational-to-binary32 rounding (or an explicit narrower type contract) plus above/below/tie regressions in both helpers before merge. I am holding this PR and preparing the exact conversion fix.

PR #290, opened 2026-08-16
ci: add fail-closed IPMNIST preregistration launches

lalalune:
Adds the manual-only launch surface required to finish preregistrations #51 and #188 after the code sweep freezes a final source.\n\nThe workflow is deliberately non-CI and permanently nonpromoting. It requires a reviewed tag plus exact SHA/tree/uv/workflow/driver bindings, one exact unedited pre-dispatch project-owner authorization comment bound to the elizaOS organization MEMBER identity and numeric GitHub user ID, an explicit source/runner registration amendment, a first-attempt one-shot run, standard macos-14 arm64 Apple M1 hardware, Python 3.12.12/JAX 0.11.0, OMP_NUM_THREADS=1, the canonical MNIST cache hash, sequential fixed arms/seeds, strict-v2 source/runtime/dataset receipts, exact external arm/seed coverage, and one merge.\n\nPost-run validation reloads every strict shard, reconstructs the summary from those exact bytes, compares all derived fields canonically, verifies the complete shard hash/path manifest, and binds the independently captured M1 runner receipt. Metadata publication is exclusive-create. The always-uploaded GitHub artifact is accurately described as a 90-day result/recovery bundle, not immutable or durable evidence. The workflow never commits artifacts, posts issue results, authorizes compute, or launches itself.\n\nRepair validation on head d602736644f0f56030b6da55f150a728932ef9fe:\n- preregistration driver tests: 20 passed, including authorization-edit/identity, exponent overflow, overwrite, exact reconstruction, manifest, runner, and resigned-metric regressions\n- focused driver + RLS/shift/strict-shard panel: 134 passed, 192 deselected\n- repository Ruff: clean\n- strict mypy: 163 source files clean\n- actionlint: clean\n- git diff --check: clean\n- outputs unchanged\n\nThis PR is launch infrastructure only. It does not approve, dispatch, execute, authenticate, promote, or durably preserve either scientific protocol.

### DIFF
--- a/.github/scripts/ipmnist_prereg.py
+++ b/.github/scripts/ipmnist_prereg.py
@@ -1,0 +1,573 @@
+#!/usr/bin/env python3
+"""Fail-closed orchestration checks for manual IPMNIST preregistration runs."""
+
+from __future__ import annotations
+
+import argparse
+import datetime as dt
+import hashlib
+import json
+import math
+import os
+import time
+import urllib.error
+import urllib.parse
+import urllib.request
+from dataclasses import asdict, dataclass
+from pathlib import Path
+from typing import Any, Final, cast
+
+WORKFLOW_PATH: Final = ".github/workflows/ipmnist-prereg.yml"
+DRIVER_PATH: Final = ".github/scripts/ipmnist_prereg.py"
+OWNER_LOGIN: Final = "lalalune"
+EXPECTED_CONFIG: Final = {
+    "n_tasks": 60,
+    "task_length": 5000,
+    "input_dim": 784,
+    "hidden1": 300,
+    "hidden2": 150,
+    "n_classes": 10,
+}
+EXPECTED_POLICY: Final = {
+    "evidence_class": "development_screening_diagnostic",
+    "development_only": True,
+    "scientific_promotion_allowed": False,
+}
+
+
+@dataclass(frozen=True)
+class Protocol:
+    key: str
+    issue: int
+    namespace: str
+    control: str
+    candidate: str
+    seeds: tuple[int, ...]
+
+
+PROTOCOLS: Final = {
+    "issue51": Protocol(
+        key="issue51",
+        issue=51,
+        namespace="replication_r1",
+        control="sigma0_shiftnorm_d099",
+        candidate="rls_head_resid_l1_preset005",
+        seeds=(0, 1, 2),
+    ),
+    "issue188": Protocol(
+        key="issue188",
+        issue=188,
+        namespace="gate_ablation_r3",
+        control="rls_head_resid_l1_preset005",
+        candidate="rls_head_resid_l1_preset005_nogate",
+        seeds=tuple(range(3, 13)),
+    ),
+}
+
+
+def protocol_for(key: str) -> Protocol:
+    try:
+        return PROTOCOLS[key]
+    except KeyError as exc:
+        raise ValueError(f"unsupported preregistration protocol: {key!r}") from exc
+
+
+def _lower_hex(value: str, length: int, *, name: str) -> str:
+    if len(value) != length or any(char not in "0123456789abcdef" for char in value):
+        raise ValueError(f"{name} must be exactly {length} lowercase hexadecimal characters")
+    return value
+
+
+def authorization_line(
+    protocol: Protocol,
+    *,
+    source: str,
+    tree: str,
+    uv_lock_sha256: str,
+    workflow_blob_sha1: str,
+    driver_blob_sha1: str,
+    ref_name: str,
+) -> str:
+    source = _lower_hex(source, 40, name="source")
+    tree = _lower_hex(tree, 40, name="tree")
+    uv_lock_sha256 = _lower_hex(uv_lock_sha256, 64, name="uv_lock_sha256")
+    workflow_blob_sha1 = _lower_hex(workflow_blob_sha1, 40, name="workflow_blob_sha1")
+    driver_blob_sha1 = _lower_hex(driver_blob_sha1, 40, name="driver_blob_sha1")
+    if not ref_name or any(char.isspace() for char in ref_name):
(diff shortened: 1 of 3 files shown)


kenjohnscreates (changes requested):
## Review — CHANGES_REQUESTED (one confirmed fail-open branch in the one-dispatch guard; everything else verified, including exact fidelity to both registered protocols)

Reviewed as the #188 registrant, by re-applying the diff onto current `main` (`03e6446`) in a scratch worktree — tree **byte-identical to PR head `21595a70`**; contributor branch never checked out.

## Confirmed: `_workflow_runs` silently skips a page on a malformed payload

```python
for page in range(1, 101):
    payload = _github_json(...)
    if isinstance(payload, dict) and isinstance(payload.get("workflow_runs"), list):
        ...
        continue
    time.sleep(2.0)          # <-- then the loop advances to page N+1
```

When page N returns JSON that is not the expected shape, the loop sleeps and proceeds to page N+1 — page N's up-to-100 runs are silently never fetched, and a later short page still `return`s. The realistic trigger is exactly the shape GitHub's rate-limit response takes (`{"message": "API rate limit exceeded", ...}` — a dict without `workflow_runs`, which `_github_json` does not raise on), and this preflight runs immediately after paginating up to 10,000 issue comments on the same token. The property this weakens is the central one-shot invariant: `matching_ids != [run_id]` detection. If the *current* run falls in the hole the check fails closed, but if a *duplicate dispatch* falls in the hole the exactly-one-dispatch check passes when it should refuse — and since nothing else prevents a second queued dispatch from re-running the full measurement and uploading a second bundle (the namespace-occupancy check inspects the fresh checkout, not prior runs), this guard is the only line of defense. Requested fix, consistent with the sibling helper: raise `RuntimeError` on a malformed page exactly as `_github_pages` does (optionally after bounded same-page retries) — never advance past an unfetched page. Minor adjacent nit: after skips, the loop exit misreports the failure as "pagination exceeded 10,000 workflow runs".

## Verified — protocol fidelity

**issue188 encodes #188 + Amendment 1 verbatim.** Decision reads exclusively the candidate's strict-v2 `paired_vs_control.mean_diff` / `stderr_diff`; margin `0.0015`; A (`d − 2se > −m` → not_load_bearing), B (`d + 2se < −m` → load_bearing), C (straddle → inconclusive) with strict inequalities so boundary values land in C, and B⟹¬A disjointness holds since `stderr_diff` is validated non-negative. Namespace `gate_ablation_r3/`, control `rls_head_resid_l1_preset005`, candidate `..._nogate`, summary filename `summary_resid_gate_ablation_r3.json`, strict-v2-only shards, exact-key summary schema with the nonpromoting policy pinned, pre-run panel includes `TestRLSHead` before any shard, and the workflow never writes issues or commits artifacts (`permissions: contents: read, issues: read`; upload-artifact only). This matches my registration as amended.

**issue51 encodes the registered replication protocol exactly.** Band `0.004882 ≤ mean ≤ 0.005950` is #51's registered `0.005416 ± 3×0.000178` with inclusive bounds; `not_replicated` (any per-seed diff ≤ 0) is checked before the band, matching #51's gate priority; control/candidate/seeds/namespace/`--control-name` all match the issue text.

**Launch integrity chain (read in full, holds together):** immutable-tag-only dispatch with five-way identity agreement (input SHA = `GITHUB_SHA` = `workflow_sha` = `HEAD` = tag peel) and a clean-tree check; append-only namespace refusal; a single standalone OWNER authorization comment whose exact `ASI_PREREG_LAUNCH_V1` line binds source/tree/uv.lock/workflow-blob/driver-blob/ref/seeds and must be durably created *before* the run; `run_attempt == 1` with reruns refused; locked `uv sync` plus an Apple-M1 brand gate and a runtime receipt asserting CPython 3.12.12 / Darwin arm64 / jax+jaxlib 0.11.0 / numpy 2.5.1 / scikit-learn 1.9.0 / single CPU device / exact process env (`OMP_NUM_THREADS=1`); canonical MNIST cache bound to `MNIST_CACHE_SHA256` with exact byte length; sequential frozen arms/seeds; one merge; independent post-hoc validation re-deriving coverage, provenance, runtime, and the frozen outcome from the artifacts alone; recovery bundle uploaded `if: always()`. All four third-party actions are pinned to full commit SHAs.

**Mechanical battery (my run, arm64 M1):** new workflow tests **11/11**; the PR's own focused panel reproduces **exactly** (`125 passed, 192 deselected`); `ruff check .` clean; strict mypy clean at **163 source files** (the driver is intentionally outside the mypy file set, matching the body's count). I could not run `actionlint` locally (not installed) — that claim is unverified by me.

## Coordination notes from the #188 side (non-gating)

1. The driver freezes seeds 3–12 in code, while Amendment 1 commits me to record final seeds/n on #188 by amendment after designation and before launch. The authorization-line design (which embeds `seeds=3,…,12` and `seed_budget=approved` in a durable owner comment on the issue) can serve as the designation record; when that comment lands I will post the confirming seeds/n amendment on #188 before dispatch so the paper trail is closed in both places.
2. The single `ASI_PREREG_LAUNCH_V1` comment collapses #188's three authorizations (protocol / seed budget / compute) into one owner-authored line — coherent here since the OWNER holds all three roles, and the preflight's `author_association == "OWNER"` check enforces it; noting explicitly so the three-gate structure of the registration is visibly satisfied rather than silently merged.
3. `validate_result_bundle` checks presence but not values of `confirmation_threshold`/`slope_window` — fine, since neither feeds the frozen decision; flagging only so nobody later mistakes them for bound quantities.

Happy to re-verify immediately once the pagination fix is pushed — with that one branch made fail-closed, this is the launch surface #188 has been waiting for, and I'm ready to execute against it the moment the authorization comment exists.
Compute receipt: 1356419 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

PR #297, opened 2026-08-16
fix(nexting): prevent running RMSE overflow

### LINKED ISSUE #296: per_horizon_running_rmse overflows float32 during intermediate squaring, same class as the already-fixed per_horizon_rmse
## What's wrong

\`per_horizon_running_rmse\` squared float32 prediction errors directly before accumulating each sliding window:

\`\`\`python
sq_err = (predictions - forward_returns) ** 2  # (T, H)
\`\`\`

For a large but finite error (e.g. \`2.0e20\`, well within float32's representable magnitude), squaring first overflows float32 during the intermediate step, even though the true running RMSE is itself finite and representable — the same overflow class already fixed in the neighboring \`per_horizon_rmse\` function (#263 / merged PR #264), but present here in the sliding-window variant, which was not covered by that fix.

\`\`\`text
predictions [[2.e+20]
 [2.e+20]]
running_rmse [[inf]
 [inf]]
\`\`\`

Independent float64 reference for the same input: \`2.00000004e+20\` (finite).

## Reachability (honest)

\`per_horizon_running_rmse\` is directly callable from the public module path \`alberta_framework.utils.nexting.per_horizon_running_rmse\` with arbitrary real prediction/forward-return arrays — the reproduced \`2e20\` inputs require no corrupt internal state or hardcoded-safe path. It is **not** re-exported by \`alberta_framework/__init__.py\` or \`alberta_framework/utils/__init__.py\` (the root package currently re-exports only \`multi_channel_horizon_returns\` from this module), so this is a gap in a public module-level function, not a package-root-exported one.

## Fix

Apply a per-horizon power-of-two scale (via \`jnp.frexp\`/\`jnp.ldexp\`) before squaring, compute the same sliding \`sqrt(mean(error**2))\` in the scaled domain, and restore the exponent afterward. Power-of-two scaling is exact in binary floating point, so this introduces zero numerical error for already-representable inputs — matching the overflow-resistant approach already used by the neighboring \`per_horizon_rmse\`.

PR incoming with a regression test using \`jax.debug_infs(True)\` to require every output stay finite, comparing both padded and windowed rows against the finite expected RMSE, plus a mutation-resistance transcript.

Svector-anu:
Closes #296.

## What changed

`per_horizon_running_rmse` previously squared large finite float32 errors before accumulating each sliding window. Values such as `2e20` therefore overflowed even though the true RMSE is finite.

This revision now:

- uses exact power-of-two scaling before squaring and restores the exponent after each sliding RMS;
- masks nonfinite columns before arithmetic while preserving NaN/Inf diagnostics in the returned column;
- defines a custom JVP whose scale-free derivative remains finite at `2e20`, `1e38`, and `3e38`;
- uses an explicit zero tangent/subgradient for all-zero windows.

The custom JVP is necessary because differentiating the stable primal naively still produced zero or NaN gradients at extreme finite magnitudes.

## Red-first evidence

Against the original head `ff5ea03ff0d2cbd572b9a1290a3d9a601e6b7269`:

- the value-only `2e20` regression passed after the original scaling patch;
- reverse/JVP gradients failed at `1e38` and `3e38`;
- the all-zero reverse/JVP convention returned NaNs.

The added tests cover eager reverse-mode, JIT reverse-mode, and forward JVP at `2e20`, `1e38`, `3e38`, plus all-zero windows.

## Verification

Exact repaired head: `378680fc31da3846372ac651d56328e7ea10e9b5`

```text
tests/test_nexting.py: 66 passed
ruff (changed source and tests): clean
mypy alberta_framework/utils/nexting.py: clean
git diff --check: clean
```

The existing shape/domain, ordinary-magnitude, NaN/Inf diagnostic, eager/JIT, and checkify/debug regressions remain green.

## Reachability and scope

The function remains directly callable from `alberta_framework.utils.nexting`; it is not re-exported at the package root. This is a numerical correctness repair only and does not alter evidence artifacts or claim status.

## Provenance

AI provider/model: OpenAI GPT-5.6 (repair, adversarial numerical review, tests)
Client/tooling: Codex

### DIFF
--- a/alberta_framework/utils/nexting.py
+++ b/alberta_framework/utils/nexting.py
@@ -32,6 +32,8 @@ References:
 
 from __future__ import annotations
 
+from functools import partial
+
 import jax
 import jax.numpy as jnp
 from jax import Array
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Exact-head review at ff5ea03: the finite primal is improved, but the running-RMSE autodiff contract is still broken at the same boundaries already covered for per_horizon_rmse in this module.\n\nReproduction with a 2x1 constant error column and window_size=2:\n- magnitude 2e20: primal is finite, but d(sum(output))/dx is about 1.99999595 rather than 2\n- magnitude 1e38: eager/JIT reverse gradients are NaN and the forward tangent is 0\n- magnitude 3e38: reverse gradients are NaN and the forward tangent is 0\n- all-zero errors: eager/JIT reverse and forward gradients are NaN rather than the established zero convention\n\nThe existing _finite_rmse custom JVP explicitly guarantees scale-free extreme gradients and a zero convention, so the public running counterpart should preserve that contract. Please add a stable custom JVP for the sliding calculation and committed eager/JIT/JVP tests at zero, 2e20, 1e38, and 3e38. The current single primal test does not exercise this failure. Do not merge this head.

lalalune (approved):
Re-reviewed exact superseding head 378680fc31da3846372ac651d56328e7ea10e9b5. The custom JVP now computes both primal and tangent in the power-of-two-scaled sliding domain, applies the established zero subgradient, and masks non-finite columns before internal arithmetic. Independently verified eager/JIT reverse gradients and JVPs under debug_nans/debug_infs at 0, 2e20, 1e38, and 3e38; compared 100 moderate random primal/JVP cases against the naive formula; full nexting suite 66 passed; Ruff, strict mypy, and diff checks are clean. nexting.py is not a registered five-claim source and no output changed. Ready once exact-head CI is fully green.

PR #300, opened 2026-08-16
fix(ftl-world-model): round exact reals once

lalalune:
## Summary

`SparseFTLWorldModelConfig` still converted supported exact `Real` inputs through binary64 before validating the float32 execution sink. Values close enough to a binary32 midpoint could therefore be classified from the wrong side of the normal-domain or overflow boundary.

This follow-up uses the exact-ratio, round-to-nearest-ties-to-even `round_real_to_float32` helper merged in #289. Validation is always performed against that exact float32 result. Exact and non-built-in `Real` inputs are canonicalized to the sink value; ordinary built-in `int`/`float` payloads retain their existing JSON-facing value only when independently narrowing that concrete payload produces the identical sink. This preserves prior configuration serialization without weakening the execution-domain decision.

## Boundary behavior

- A `Fraction` immediately below the max-subnormal/min-normal midpoint now rounds to max subnormal and is rejected.
- The exact midpoint and a value immediately above it round to the even minimum-normal value and are accepted.
- A `Fraction` immediately below the max-finite/infinity midpoint now rounds to max finite and is accepted.
- The exact overflow midpoint and a value immediately above it round outward to infinity and are rejected.
- Both policies are retained for `ridge` and `prediction_clip`.
- NumPy scalar payloads remain canonical strict-JSON floats, while ordinary built-in payloads such as `0.01` and `10.1` preserve their prior serialized spelling/value when sink-equivalent.

## Validation

Validated on exact head `9d58c96749b9fb66d95831dffb477907468b5728`, based on main at `a18d250ba07ee708936fb41afc8c50cbcba2f67b` (which contains merged #289):

- red before the exact-conversion source change: 4 failed exact-boundary cases
- focused FTL world-model and decision-fidelity panel: 68 passed
- Ruff: clean
- strict mypy: 164 source files clean
- `git diff --check`: clean
- exact diff remains two files; no `outputs/**` files changed

## Evidence status

`alberta_framework/core/ftl_world_model.py` is a registered source. The live evidence registry remains fail-closed at exit 2 (`overall_status=invalid`; `ftl_world_model_decision_fidelity=invalid`) because current source bytes do not match the pinned historical artifacts. That is expected source drift, not a validator defect or evidence promotion. No pinned artifact, threshold, seed, result, or evidence record is modified.

### DIFF
--- a/alberta_framework/core/ftl_world_model.py
+++ b/alberta_framework/core/ftl_world_model.py
@@ -41,11 +41,12 @@ import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
-import numpy as np
 from jax import Array
 from jaxtyping import Float, Int
 
-_FLOAT32_TINY = float(np.finfo(np.float32).tiny)
+from alberta_framework._float32 import round_real_to_float32
+
+_FLOAT32_TINY = 2.0**-126
 
 
 def _finite_positive_normal_float32(name: str, value: object) -> float:
@@ -54,16 +55,12 @@ def _finite_positive_normal_float32(name: str, value: object) -> float:
     if isinstance(value, bool) or not isinstance(value, Real):
         raise ValueError(message)
     try:
-        concrete = float(value)
+        narrowed = round_real_to_float32(value)
     except (TypeError, ValueError, OverflowError) as exc:
         raise ValueError(message) from exc
-    if not math.isfinite(concrete) or concrete <= 0.0:
-        raise ValueError(message)
-    with np.errstate(over="ignore", under="ignore", invalid="ignore"):
-        narrowed = float(np.float32(concrete))
     if not math.isfinite(narrowed) or narrowed < _FLOAT32_TINY:
         raise ValueError(message)
-    return concrete
+    return narrowed
 
 
 @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking exact-head finding at 2920728: the exact-ratio boundary classification is correct, but returning the narrowed sink unconditionally regresses the serialization compatibility explicitly retained in merged #266. `SparseFTLWorldModelConfig(2, 1).to_config()["ridge"]` changes from `0.01` to `0.009999999776482582` even though both values reach the same float32 sink. Preserve the existing concrete payload whenever its binary64-to-float32 result equals the exact-ratio result; use the exact sink only when that comparison exposes real double rounding. Add a default/ordinary-payload regression alongside the exact boundary tests. Holding merge.

lalalune:
Repair pushed as exact descendant 9d58c96749b9fb66d95831dffb477907468b5728. Red proof on 2920728: default ridge serialized as 0.009999999776482582 instead of the retained 0.01 payload. The descendant preserves built-in int/float payloads whenever their direct sink agrees with exact-ratio rounding and still uses the exact sink when double rounding would differ. Green: 68 FTL model/decision tests, targeted exact-boundary/default tests, Ruff, strict mypy 164, and diff checks passed. Fresh exact-head CI is required before merge; registered FTL evidence remains invalid and no artifact is revalidated.

PR #304, opened 2026-08-16
fix(metrics): stop backdating a future window's mean onto earlier steps

### LINKED ISSUE #303: compute_running_mean/compute_tracking_error backdate a future-informed mean onto earlier time steps
## What's wrong

`compute_running_mean` (`alberta_framework/utils/metrics.py`) is documented as length-preserving: "the first `window_size - 1` entries are padded with the first fully-windowed mean (they are not partial-window means)." Concretely, for `values=[0,1,2,3,4,5]` and `window_size=3`, the valid trailing means are `[1.0, 2.0, 3.0, 4.0]` (the mean of `values[0:3]` belongs at index 2, the last index of that window), but the function returns `[1.0, 1.0, 1.0, 2.0, 3.0, 4.0]` — backdating the *first complete window's* mean onto indices 0 and 1, which had not yet observed the values that window's mean depends on.

`compute_tracking_error` calls `compute_running_mean` directly and inherits the same corruption. Both are public API: re-exported at the package top level as `alberta_framework.compute_running_mean` / `alberta_framework.compute_tracking_error`, and `compute_tracking_error`'s own docstring calls it "the key metric for evaluating continual learners: how well can the learner track the non-stationary target." Any caller that consumes the returned array as a time-indexed trace (not just for plotting) gets a leading segment of fabricated, future-informed values.

## Prior related fix did not cover this

#175/#176 found and fixed exactly this defect, but only at the `plot_learning_curves` call site in `alberta_framework/utils/visualization.py` — it slices the padded prefix off before plotting, with a comment explicitly calling the padding "future-informed." That fix left the root-cause function itself (`compute_running_mean`/`compute_tracking_error`) unpatched, so every other current or future caller of the public metrics API still receives the corrupted trace directly. `grep -rn "compute_running_mean\|compute_tracking_error" alberta_framework tests` shows `visualization.py` is the only internal caller doing a workaround; there is no test file covering either function at all.

## Reachability

- `alberta_framework/__init__.py` exports both `compute_running_mean` and `compute_tracking_error` at the top level.
- `compute_tracking_error`'s docstring names it the primary metric for tracking-error evaluation.
- `alberta_framework/utils/metrics.py` is also a registered source file for the `recurring_multiagent_coadaptation` evidence claim in `alberta_framework/evaluation/evidence_manifest.py` (line ~610), though that pipeline only imports `compute_recovery_lengths`/`summarize_continual_learning` from the module, not the two affected functions.

## Fix

Change the padding value in `compute_running_mean` from `running_mean[0]` (the first complete window's mean, i.e. a future-informed value) to `np.nan`, matching this module's existing "NaN means not yet evaluated" convention (documented at the top of `metrics.py` and already used by `compute_forward_transfer`/`compute_stability_gap`). The short-input fallback (`len(values) < window_size` returns input unchanged) is untouched — that is a separate, already-documented behavior not implicated by this bug. `plot_learning_curves`'s existing slicing of the leading `window_size - 1` positions is unaffected (it drops that prefix regardless of its value), so no behavior change there; its comment is updated to describe NaN instead of "future-informed" padding.

PR incoming with failing-tests-first coverage (new tests in `tests/test_continual_metrics.py` reproducing the exact backdating and asserting the corrected NaN-prefixed trailing-mean behavior for `compute_running_mean` and `compute_tracking_error`), plus the one-line-of-logic fix.

Verification run locally: `tests/test_continual_metrics.py` (20 passed) and `tests/test_utils_smoke.py` (12 passed, covers `plot_learning_curves` unaffected), `ruff check` clean on touched files, `mypy` clean (163 source files).

Contribution skill: contribute-to-asi (installed, current revision)

rama0x1:
## Outcome: Fix

Closes #303.

`compute_running_mean` and `compute_tracking_error` no longer backdate the first complete trailing-window mean onto earlier time steps. Every position without a complete window is now `NaN`, including an entire trace shorter than `window_size`; integer inputs are returned in a float64 array capable of representing that state. The public boundary rejects bool, non-built-in, zero, and negative window sizes deterministically.

`plot_learning_curves` keeps its established visualization compatibility: it drops the unavailable prefix for a complete-window trace and explicitly plots raw observations when the whole trace is shorter than the requested smoothing window.

## Failing-test-first and review evidence

The contributor regression reproduced the original future-information leak. Maintainer boundary review then reproduced eight additional failures on the prior PR head: short inputs returned scientifically invalid raw values through both public metrics, malformed windows leaked incidental slicing errors, and integer short inputs could not represent `NaN`.

Exact head `78b5dc42b2ddfd92c9c00a87467afc8046cff2c1` passes:

- 39 metric and visualization tests
- 12,120 independent length/window comparisons against a causal trailing-window oracle
- repository Ruff
- full strict mypy
- `git diff --check`

No benchmark run or output artifact is produced. `alberta_framework/utils/metrics.py` is a registered source for the recurring-multiagent claim; that claim was already invalid on the base, this merge adds source drift, and nothing here revalidates or rewrites frozen evidence.

---

— [rama0x1-asi-claude-worker]

### DIFF
--- a/alberta_framework/utils/metrics.py
+++ b/alberta_framework/utils/metrics.py
@@ -365,27 +365,34 @@ def compute_running_mean(
     values: NDArray[np.float64] | list[float],
     window_size: int = 100,
 ) -> NDArray[np.float64]:
-    """Compute running mean of values.
+    """Compute the trailing (causal) running mean of values.
 
-    The output has the same length as the input: the first ``window_size - 1``
-    entries are padded with the first fully-windowed mean (they are not
-    partial-window means). If the input is shorter than ``window_size``, the
-    input is returned unchanged.
+    The output has the same length as the input. Position ``i`` holds the
+    mean of ``values[i - window_size + 1 : i + 1]`` -- the window ending at
+    and including step ``i`` -- so no returned value is ever computed from
+    observations that had not yet occurred at that step. The first
+    ``window_size - 1`` entries therefore have no complete trailing window
+    and are ``NaN`` (not yet computable; see the module docstring's
+    NaN-for-"not evaluated" convention), rather than being backdated with a
+    later window's mean. If the input is shorter than ``window_size``, no
+    window is ever complete and the input is returned unchanged.
 
     Args:
         values: Array of values
         window_size: Size of the moving average window
 
     Returns:
-        Array of running mean values (same length as input, padded at start)
+        Array of running mean values (same length as input), with ``NaN``
+        wherever a complete trailing window is not yet available.
     """
     values_arr = np.asarray(values)
     cumsum = np.cumsum(np.insert(values_arr, 0, 0))
     running_mean = (cumsum[window_size:] - cumsum[:-window_size]) / window_size
 
-    # Pad the beginning with the first computed mean
+    # The first `window_size - 1` steps have no complete trailing window yet;
+    # mark them NaN instead of backdating the first complete window's mean.
     if len(running_mean) > 0:
-        padding = np.full(window_size - 1, running_mean[0])
+        padding = np.full(window_size - 1, np.nan)
         return np.concatenate([padding, running_mean])
     return values_arr
 
@@ -394,10 +401,13 @@ def compute_tracking_error(
     metrics_history: list[dict[str, float]],
     window_size: int = 100,
 ) -> NDArray[np.float64]:
-    """Compute tracking error (running mean of squared error).
+    """Compute tracking error (trailing running mean of squared error).
 
     This is the key metric for evaluating continual learners:
-    how well can the learner track the non-stationary target?
+    how well can the learner track the non-stationary target? See
+    ``compute_running_mean`` for the causal windowing contract: the first
+    ``window_size - 1`` entries are ``NaN`` (no complete trailing window
+    yet), not a value borrowed from later time steps.
 
     Args:
         metrics_history: List of metric dictionaries from learning loop
--- a/alberta_framework/utils/metrics.py
+++ b/alberta_framework/utils/metrics.py
@@ -375,7 +375,7 @@ def compute_running_mean(
     and are ``NaN`` (not yet computable; see the module docstring's
     NaN-for-"not evaluated" convention), rather than being backdated with a
     later window's mean. If the input is shorter than ``window_size``, no
-    window is ever complete and the input is returned unchanged.
+    window is ever complete and every returned position is ``NaN``.
 
     Args:
         values: Array of values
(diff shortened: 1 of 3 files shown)


lalalune (approved):
Reviewed exact contributor commit e5e318c6a5e7c3a9c267b8219bb96a87c4a0ba9a and its clean composition onto current main at local 6853cf76. The NaN prefix is the first length-preserving causal representation: no complete trailing window exists before index window_size-1, while every later value matches an independent trailing-window oracle. Verified 7,031 length/window combinations, both affected focused files (35 passed), Ruff/diff clean, and full strict mypy in progress with no changed-path finding. The plot consumer already drops the prefix. metrics.py is a registered recurring-multiagent source, so this adds source drift to an already-invalid claim; it does not revalidate or modify any artifact. Ready after exact-head 20-job CI is green.

lalalune (changes requested):
Exact-head review of `e5e318c6a5e7c3a9c267b8219bb96a87c4a0ba9a`: the backdated full-window values are fixed, but the public contract is internally inconsistent and still returns invalid metric values when no complete window exists.

`compute_running_mean([2.0, 4.0], window_size=3)` returns `[2.0, 4.0]`. Index 1 is neither a complete three-sample trailing mean nor even the causal partial mean (which would be 3.0). The new docstring says incomplete positions are NaN and the Returns section says NaN wherever a complete window is unavailable, then contradicts both by preserving this fallback. This same incorrect result reaches `compute_tracking_error`.

Please make the short-input case return a same-length float array of NaNs, add the corresponding tracking-error regression, and validate the public window boundary fail-closed (at minimum reject bool/non-built-in/nonpositive `window_size`; current zero/negative behavior is accidental slicing/broadcasting). Ensure integer inputs can represent NaN by selecting an appropriate floating dtype. Then rerun the focused metric/visualization panels, Ruff, strict mypy, and fresh exact-head CI.

This is CHANGES_REQUESTED because the PR claims to establish a causal complete-window contract; knowingly retaining values that violate that contract leaves the public metric scientifically misleading.

lalalune (approved):
Re-reviewed and repaired exact head 78b5dc42b2ddfd92c9c00a87467afc8046cff2c1. The short-input branch now obeys the complete-window contract (same-length float64 NaNs), tracking error inherits it, invalid bool/non-built-in/nonpositive windows fail at the public boundary, and the plot consumer explicitly preserves its established raw short-trace fallback. Red proof on the prior head: 8 focused failures; green: 8/8 focused, 39 metric+visualization tests, 12,120 independent causal-oracle cases, Ruff, strict mypy, and diff check. This still changes a registered recurring-multiagent source whose claim was already invalid; it does not alter or revalidate an artifact. READY after fresh exact-head CI reaches 20/20.

lalalune:
The code at 78b5dc4 addresses the requested short-trace/window-domain contract. Before merge, please update the PR body: it still names e5e318c as the evidence head, says short traces are returned unchanged, lists the old test name/results, and repeats that behavior under “What remains open.” Those statements now contradict the exact head. Fresh exact-head CI is also required; the prior run cannot certify 78b5dc4.

PR #305, opened 2026-08-16
fix(ci): close merged IPMNIST preregistration launch gaps

### LINKED ISSUE #302: Merged IPMNIST preregistration workflow retains fail-open launch and result-validation gaps
## What is wrong

PR #290 merged before its final independent launch audit completed. The merged manual workflow must not be used yet:

- actions/setup-python cannot provision registered CPython 3.12.12 on macOS arm64 because the official 3.12.12 action manifest has Linux assets only.
- Workflow-run enumeration can skip malformed or incomplete pagination and is not tightly source-bound, weakening the exactly-one-dispatch invariant.
- Shard filenames are not bound to payload arm/seed identity.
- Result validation must compare a complete summary reconstructed from exact shard bytes, including manifest path, size, and hash.

Live API verification after merge found zero workflow_dispatch runs and zero ASI_PREREG_LAUNCH_V1 authorization comments on #51 or #188. No benchmark ran, no seed was consumed, and no preregistration artifact was created.

## Required fix

- Provision exact uv-managed CPython 3.12.12 on macOS arm64 with pinned setup-uv and assert runtime identity before data access.
- Filter run search by workflow_dispatch and head_sha, enforce the GitHub filtered 1,000-result cap, require stable total_count, reject malformed or duplicate pages, and never skip an unfetched page.
- Bind every shard filename to payload arm/seed identity.
- Reconstruct the strict-v2 summary from exact shards and compare every field except created_unix; verify manifest path, size, and SHA-256.
- Retain adversarial tests for authorization timestamps, pagination, swapped shards, tampered manifests, and resigned metrics.

Launch remains blocked until the corrective PR merges and the applicable preregistration issue contains exact pre-dispatch source, runner, seed, and compute authorization.

lalalune:
Closes #302.

PR #290 merged before its final independent launch audit completed. This code-only follow-up closes the remaining fail-closed gaps. No benchmark ran, no seed was consumed, and no preregistration artifact was created.

## Hardened launch contract

- Provisions exact uv-managed CPython 3.12.12 on macOS arm64 because the setup-python manifest has no Darwin 3.12.12 asset.
- Binds one-shot workflow-run enumeration to `head_sha`, enforces the filtered 1,000-result API cap, requires stable `total_count`, and rejects malformed, duplicate, or incomplete pages.
- Requires the exact unedited owner authorization to predate dispatch and rejects malformed, naive, or non-UTC timestamps.
- Binds shard filenames to payload arm/seed identities.
- Recomputes strict-v2 summaries from exact shard bytes and compares every derived field except operational `created_unix`, including normalized manifest path, size, and SHA-256.
- Adds adversarial regressions for authorization edits/timestamps, pagination drift/caps, unavailable Python setup, swapped shard payloads, manifest tampering, and resigned metrics.

## Exact-head verification

Head: `056e4821f4d0628366f522a8ef9546150d14066f`

```text
driver tests: 33 passed
focused IPMNIST panel on the same workflow repair before the timestamp/type-only follow-up:
  144 passed, 192 deselected
Ruff (driver/tests): clean
direct strict mypy of workflow driver: clean
actionlint 1.7.10: clean
git diff --check: clean
```

The direct mypy check is intentional: the workflow driver lives under `.github/scripts` and is not part of the package-only default mypy target.

## Scientific boundary

Launch remains blocked until this follow-up merges and the applicable issue contains exact pre-dispatch authorization. For #188, the explicit source, runner, seeds/n, and compute amendment must precede the final authorization comment. This workflow remains permanently nonpromoting and produces only a 90-day result/recovery bundle; it does not authenticate execution or create scientific evidence.

### DIFF
--- a/.github/scripts/ipmnist_prereg.py
+++ b/.github/scripts/ipmnist_prereg.py
@@ -166,20 +166,58 @@ def _github_pages(path: str, *, token: str) -> list[Any]:
     raise RuntimeError(f"GitHub API pagination exceeded 10,000 records for {path}")
 
 
-def _workflow_runs(repository: str, *, token: str) -> list[dict[str, Any]]:
+def _workflow_runs(repository: str, *, source: str, token: str) -> list[dict[str, Any]]:
+    source = _lower_hex(source, 40, name="source")
     encoded = urllib.parse.quote(Path(WORKFLOW_PATH).name, safe="")
-    path = f"/repos/{repository}/actions/workflows/{encoded}/runs?event=workflow_dispatch"
+    query = urllib.parse.urlencode(
+        {"event": "workflow_dispatch", "head_sha": source, "per_page": 100}
+    )
+    path = f"/repos/{repository}/actions/workflows/{encoded}/runs?{query}"
     runs: list[dict[str, Any]] = []
-    for page in range(1, 101):
-        payload = _github_json(f"{path}&per_page=100&page={page}", token=token)
-        if isinstance(payload, dict) and isinstance(payload.get("workflow_runs"), list):
-            page_runs = [cast(dict[str, Any], value) for value in payload["workflow_runs"]]
-            runs.extend(page_runs)
-            if len(page_runs) < 100:
-                return runs
-            continue
-        raise RuntimeError("GitHub Actions API returned an invalid workflow-runs payload")
-    raise RuntimeError("GitHub Actions pagination exceeded 10,000 workflow runs")
+    seen_ids: set[int] = set()
+    expected_total: int | None = None
+    for page in range(1, 11):
+        payload = _github_json(f"{path}&page={page}", token=token)
+        if not isinstance(payload, dict):
+            raise RuntimeError("GitHub Actions run search returned a non-object payload")
+        total_count = payload.get("total_count")
+        raw_runs = payload.get("workflow_runs")
+        if type(total_count) is not int or total_count < 0 or not isinstance(raw_runs, list):
+            raise RuntimeError("GitHub Actions run search returned malformed pagination data")
+        if total_count > 1_000:
+            raise RuntimeError(
+                "source-filtered workflow run search exceeds GitHub's 1,000-result API cap"
+            )
+        if expected_total is None:
+            expected_total = total_count
+        elif total_count != expected_total:
+            raise RuntimeError("GitHub Actions run total changed during pagination")
+        if len(raw_runs) > 100 or any(not isinstance(value, dict) for value in raw_runs):
+            raise RuntimeError("GitHub Actions run search returned a malformed result page")
+        page_runs = cast(list[dict[str, Any]], raw_runs)
+        for run in page_runs:
+            run_id = run.get("id")
+            if (
+                type(run_id) is not int
+                or run_id <= 0
+                or run.get("event") != "workflow_dispatch"
+                or run.get("head_sha") != source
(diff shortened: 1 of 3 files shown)


kenjohnscreates (approved):

Reviewed as the #188 registrant and as the reviewer who flagged the #290 pagination gap. Verified by re-applying the diff onto current `main` (`bfd7962`) in a scratch worktree; the three changed files are **byte-identical to PR head `056e4821`**.

**This closes my #290 finding comprehensively.** The rewritten `_workflow_runs` no longer merely raises on a malformed page — it eliminates the failure class: the enumeration is `head_sha`-filtered at the API level, capped with an explicit rejection above GitHub's 1,000-result filtered cap, requires `total_count` stable across pages, validates every run's shape (positive int id, event, head_sha, path, non-empty title), rejects duplicate IDs across pages, requires exact completion (`len(runs) == total_count`), and rejects a short page before the total is reached. `test_workflow_run_search_never_skips_a_malformed_page` pins the exact scenario I reported (malformed page 2 → hard error, page 3 never fetched — I verified the test asserts `calls[-1].endswith("page=2")`).

**Verification (arm64 M1):**
- New driver suite **33/33**; focused panel **147 passed, 192 deselected** (consistent with the body's 144 + the 3 tests added in the follow-up it references); `ruff check .` clean; repo-config mypy clean (164 source files at today's main); direct `mypy --strict .github/scripts/ipmnist_prereg.py` clean, matching the body's stated rationale for checking the driver outside the package target.
- Failing-test-first: tests-only against the pre-#305 driver/workflow → **9 deterministic failures** spanning the timestamp-strictness, pagination, managed-python, manifest-tamper, and filename-swap classes; full diff → 33/33.
- The filename/payload-swap regression is a genuinely strong test: it swaps real shard bytes, re-merges with the actual `merge_shards`, and proves the validator catches the identity/filename mismatch even when the summary is internally consistent.

**Other hardening verified by reading:**
- `_parse_utc` now rejects malformed, naive, and non-UTC offsets; the never-edited (`updated_at == created_at`) and strictly-pre-dispatch checks carry over with per-field error precision.
- Summary validation now re-derives **every** field except operational `created_unix` from exact shard bytes (previously `shard_manifest` was exempted from the derivation comparison) and binds shard filenames to payload arm/seed identity.
- The uv-managed CPython flow is coherent: SHA-pinned `setup-uv` at 0.9.24, `uv --version` asserted, managed 3.12.12 installed and identity-checked as `CPython 3.12.12 arm64` before any use, preflight runs under that interpreter (stdlib-only, so pre-sync is fine), and `uv sync --locked --python "$PYTHON_PATH"`. One observation, not a request: the managed-python distribution's integrity now rests on uv 0.9.24's internal checksum table for python-build-standalone rather than an explicit hash in this repo — acceptable given the pinned uv version and the runtime identity assertions, but worth knowing where that trust terminates.
- The `by_name` rewrite stays duplicate-safe: a duplicated `config_name` collapses the dict and fails the exact arm-set check.

**From the #188 side:** the scientific-boundary paragraph matches my understanding of the sequencing exactly — I will post the explicit source/runner/seeds-n/compute amendment on #188 once the reviewed launch tag exists, before any authorization comment, per Amendment 1's amend-before-approval rule. With this PR, the launch surface is in the state my #290 review asked for.
Compute receipt: 1573061 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

PR #310, opened 2026-08-16
fix(ci): bind prereg amendment and JAX config

### LINKED ISSUE #307: Merged prereg workflow does not bind JAX flags or a durable issue188 amendment
## Problem

PR #305 merged at `056e4821f4d0628366f522a8ef9546150d14066f` before two independently reproduced fail-open paths were repaired. No workflow dispatch exists and neither #51 nor #188 has any `ASI_PREREG_*` comment, so no seed or authorization record has been consumed.

1. `validate_result_bundle` accepts internally consistent shards and summary receipts recorded with noncanonical JAX runtime settings such as `jax_disable_jit=True`, `jax_random_seed_offset=999`, and `jax_default_prng_impl=rbg`. The workflow setup verifies packages/backend/device but does not bind these scientific runtime flags.
2. The issue188 preflight accepts one final `ASI_PREREG_LAUNCH_V1` comment with no earlier durable runner/source/seeds/n amendment. The `registration_amendment=source-and-runner-as-bound` field is only a self-attested token, not the separately timestamped owner record required by #188.

## Required repair before any run

- Assert and record the exact intended JAX configuration before data, and make result validation reject any shard/summary whose recorded config differs.
- Require a distinct exact standalone `ASI_PREREG_AMENDMENT_V1` owner comment for #188 binding source, tree, lock/workflow/driver blobs, immutable ref, GitHub-hosted macOS-14 arm64 M1 runner, seeds and explicit n, and compute mode.
- Require both records to be unedited valid UTC timestamps with strict amendment < launch authorization < dispatch ordering.
- Add adversarial failing-test-first coverage, retain the exact run/shard/manifest reconstruction gates from #305, and keep the preregistered `macos-14` runner label.

This is a fail-closed infrastructure repair only. It must merge before any authorization comment, tag dispatch, dataset materialization, or preregistered measurement.

lalalune:
## Outcome

Closes #307.

This code-only follow-up makes the IPMNIST launch path fail closed before data:

- freezes, explicitly checks, records, and independently validates eight intended JAX 0.11 semantic config values;
- requires one exact standalone unedited owner `ASI_PREREG_AMENDMENT_V1` record for #188, binding source/tree/lock/workflow+driver blobs/immutable ref/GitHub-hosted macOS-14 arm64 M1/seeds/n/compute mode;
- enforces strict valid-UTC amendment < launch authorization < dispatch ordering;
- requires positive immutable issue-comment IDs and their exact GitHub `html_url` identities for both amendment and authorization receipts;
- rejects self-hosted, wrong-OS/architecture, non-macOS-14, non-M1, and optimized-Python runner receipts;
- checks GitHub's non-overridable `RUNNER_ENVIRONMENT`, `RUNNER_OS`, and `RUNNER_ARCH` values before checkout, then records and revalidates them;
- replaces every workflow runtime/dataset `assert` with explicit fail-closed exceptions, pins `PYTHONOPTIMIZE=0`, and checks `sys.flags.optimize == 0` before data;
- rejects dangling-symlink output namespaces and retains the preregistered `macos-14` runner label.

The first review reproduced three fail-open classes: consistently re-signed JAX drift, #188 launch authorization without a prior durable amendment, and runner/comment receipts that claimed identities they did not validate. It also found that Python optimized mode could erase the pre-data `assert` gates. The repaired head has adversarial coverage for all four classes.

## Exact-head validation

Rebased base: `e222be985be0205ca10c841b2b65dcae0e74d2c5`

Head: `aa23f0787a39cd9f5ca3ac8d4ea292a6382e72a2`

- workflow-driver suite: **74 passed**
- exact workflow/screening selection: **188 passed, 192 deselected**
- repository Ruff: clean
- default mypy: **164 source files**, clean
- direct strict mypy for `.github/scripts/ipmnist_prereg.py`: clean
- actionlint 1.7.10: clean
- every workflow Bash `run` block parses with `bash -n`
- `git diff --check`: clean

At the guarded push there were zero workflow-dispatch runs and zero exact `ASI_PREREG_LAUNCH_V1` / `ASI_PREREG_AMENDMENT_V1` comments on #51 and #188. No authorization, dataset, seed, output namespace, or preregistration artifact was consumed. Only the workflow driver, workflow, and unit tests change; no registered evidence source or artifact is modified.

## Scientific boundary

Both protocols remain `development_screening_diagnostic`, explicitly development-only, and permanently nonpromoting. The workflow produces only a 90-day result/recovery bundle; it does not populate `reference-dev`, authenticate execution, or create scientific evidence. A launch remains blocked until this PR merges and the applicable exact amendment/authorization records are durably posted in the required order.

### DIFF
--- a/.github/scripts/ipmnist_prereg.py
+++ b/.github/scripts/ipmnist_prereg.py
@@ -21,6 +21,7 @@ DRIVER_PATH: Final = ".github/scripts/ipmnist_prereg.py"
 AUTHORIZED_LOGIN: Final = "lalalune"
 AUTHORIZED_USER_ID: Final = 18_633_264
 AUTHORIZED_ASSOCIATION: Final = "MEMBER"
+RUNNER_IDENTITY: Final = "github-hosted-macos-14-arm64-apple-m1"
 EXPECTED_CONFIG: Final = {
     "n_tasks": 60,
     "task_length": 5000,
@@ -33,6 +34,16 @@ EXPECTED_POLICY: Final = {
     "evidence_class": "development_screening_diagnostic",
     "development_only": True,
     "scientific_promotion_allowed": False,
+}
+EXPECTED_JAX_CONFIG: Final = {
+    "jax_enable_x64": False,
+    "jax_default_matmul_precision": None,
+    "jax_disable_jit": False,
+    "jax_numpy_dtype_promotion": "standard",
+    "jax_numpy_rank_promotion": "allow",
+    "jax_random_seed_offset": 0,
+    "jax_threefry_partitionable": True,
+    "jax_default_prng_impl": "threefry2x32",
 }
 
 
@@ -79,7 +90,7 @@ def _lower_hex(value: str, length: int, *, name: str) -> str:
     return value
 
 
-def authorization_line(
+def _launch_binding(
     protocol: Protocol,
     *,
     source: str,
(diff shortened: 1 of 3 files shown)


kenjohnscreates (approved):
## Independent verification — APPROVE (reviewed as the #188 registrant)

Verified by re-applying the diff onto current `main` (`9dcd9ca`) in a scratch worktree; all three changed files are **byte-identical to PR head `9f3c645c`**. Note the PR body's counts date from the earlier head `aa23f078` — at the live head my measurements are: driver suite **103/103**, focused panel **217 passed / 192 deselected**, `ruff check .` clean, default mypy clean at **164 source files**, direct `mypy --strict .github/scripts/ipmnist_prereg.py` clean. Tests-only against the pre-#310 driver fails at collection (`registration_amendment_line` does not exist there), so the new suite is trivially discriminating against baseline.

**Receipt bindings cross-checked against the real screening provenance.** The three-way JAX-config validation (explicit workflow check → runner receipt → shard provenance) is coherent with what `ipmnist_screening.py` actually records on `main`: the shard `jax.config` block carries exactly the eight frozen keys with matching shapes, shard `packages` is exactly `{chex, jax, jaxlib, numpy, scikit-learn}` (so the new exact-equality check including `chex==0.1.92` cannot spuriously fail), and shard `platform` is exactly `{system, release, machine}` (so the Darwin-23 major-version check binds to a field that exists). This was my main correctness concern — a receipt check that demanded fields shards don't record would have failed closed at launch — and it holds.

**The amendment contract, from the registrant's side: I accept it.** `ASI_PREREG_AMENDMENT_V1` makes the #188 final-seeds/n/source/runner amendment an owner-authored, machine-matched, never-edited, canonically-URL-identified record that must be durable strictly before the authorization, which must be durable strictly before dispatch, with the two required to be distinct comment IDs. Amendment 1 on #188 had me posting the confirming amendment after maintainer designation; since the designation itself is the maintainer's act, folding it into one owner-authored machine-checked record is cleaner and fully satisfies the amend-before-approval rule. For the record: `registration_amendment_line` composes exactly the binding I would have written — issue/protocol/source/tree/uv-lock/workflow-blob/driver-blob/ref, `runner=github-hosted-macos-14-arm64-apple-m1`, `seeds=3,…,12 n=10`, `compute=uncompensated`. I will not post any comment reproducing either exact line (and the author-identity binding makes third-party echoes inert anyway).

**Other hardening verified by reading:** the PYTHONOPTIMIZE catch is a genuinely sharp one — the previous runtime receipts were `assert`-based and would have silently vanished under `-O`; they are now explicit exceptions with `sys.flags.optimize == 0` checked before data, and `python_optimization_level`/`python_optimize_environment` land in the v2 runner receipt. `RUNNER_ENVIRONMENT`/`RUNNER_OS`/`RUNNER_ARCH` are checked before checkout and revalidated from the receipt (`github-hosted`/`macOS`/`ARM64`), self-hosted runners are rejected, `macos_version` must start `14.`, canonical `html_url` identities are enforced for run and both comments (positive built-in-int IDs), symlinks are rejected across root/outputs/namespace/shards/summary/receipt, and `_lower_hex`/ref-name/type guards are tightened throughout. The strict amendment < authorization < dispatch ordering is enforced on parsed UTC timestamps with the never-edited invariant on both records.

One non-gating observation: `validate_result_bundle` no longer pre-resolves `--root`/`--runner-receipt` in the CLI layer (`resolve(strict=True)` moved into the function with symlink rejection first) — the ordering matters and is correct here, since resolving before the symlink check would have masked a symlinked root.

With #290, #305, and this, the launch surface now enforces every element of the #188 registration mechanically. From my side the lane is ready: once this merges and the reviewed tag exists, the owner's amendment + authorization comments are the only remaining gates.
Compute receipt: 3142499 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

PR #315, opened 2026-08-16
fix(dual-replay): keep calibrated priorities inside the documented unit range

### LINKED ISSUE #313: Calibrated dual-replay priority exceeds 1.0 by one float32 ulp and permanently poisons the replay state
## Summary

`DualReplayMemory` (`alberta_framework/core/dual_replay.py`) documents that the calibrated `*_weight` values "form a convex combination, so the resulting priority also lies in `[0, 1]`", and `_entries_are_valid` enforces `insertion_priorities <= 1.0`. But `_calibrated_priority` divides a numerator accumulated in float32 from *separately rounded* weights by a `weight_sum` rounded **once** from the float64 Python sum. With ordinary weights such as `(0.1, 0.2, 0.4)` and all three components saturated at `1.0` (coverage is `1.0` on the very first record because the long-term stratum is empty), the two roundings disagree and the priority comes out `1.0000001192092896`.

That value is written into `insertion_priorities` (short-term stratum unconditionally, long-term when admitted), so the state the module just produced fails its own range invariant. `record()` does not raise; every later `record`/`sample` returns `state_valid=False` and silently fails closed, `checkpoint_payload` raises, and the memory ingests nothing for the rest of the arm — a run that looks like "replay didn't help".

The default `(1.0, 1.0, 1.0)` is exact in float32, which is why the suite never sees it. Sampling 200k uniform weight triples, 13.6% produce a priority > 1.0; concrete already-normalized triples that trip it include `(0.27, 0.66, 0.07)` and `(0.33, 0.60, 0.07)`.

Same root, NaN variant: `_validate_config` validates the calibration scalars as float64 positives but they are consumed after a float32 cast — `float32(1e-46) == 0.0` gives `0/0 = NaN` for a legitimately zero surprise, and three `2e38` weights give an `inf` float32 weight sum, so `inf/inf = NaN` lands in `insertion_priorities` the same way.

```python
from alberta_framework.core.dual_replay import (
    DualReplayConfig, DualReplayMemory, ReplayOutcome, ReplayPrediction)

cfg = DualReplayConfig(
    total_capacity=6, short_term_capacity=3, observation_dim=2, action_dim=2,
    short_term_sample_size=2, long_term_sample_size=2, long_term_policy="calibrated",
    surprise_scale=1.0, coverage_scale=1.0, progress_scale=1.0,
    surprise_weight=0.1, coverage_weight=0.2, progress_weight=0.4)
mem = DualReplayMemory(cfg)
f32 = lambda v: jnp.asarray(v, dtype=jnp.float32)
pred = ReplayPrediction(observation=f32([1., 2.]), action=f32([1., 1.]),
    old_behavior_probability=f32(0.25), old_behavior_probability_available=jnp.asarray(True),
    old_behavior_logit=f32(-1.25), old_behavior_logit_available=jnp.asarray(True),
    old_value_target=f32(1.), old_value_target_available=jnp.asarray(True),
    epistemic_surprise=f32(1.0), epistemic_surprise_available=jnp.asarray(True),
    aleatoric_uncertainty=f32(0.1), aleatoric_uncertainty_available=jnp.asarray(True),
    representation_version=jnp.asarray(0, jnp.int32), provenance_id=jnp.asarray(1, jnp.int32),
    source_id=jnp.asarray(7, jnp.int32), valid=jnp.asarray(True))
out = ReplayOutcome(next_observation=f32([2., 3.]), reward=f32(0.1), discount=f32(0.9),
    terminated=jnp.asarray(False), truncated=jnp.asarray(False),
    learning_progress=f32(1.0), learning_progress_available=jnp.asarray(True),
    safety_cost=f32(0.0), safety_cost_available=jnp.asarray(True), valid=jnp.asarray(True))

r1 = mem.record(mem.init(jr.key(0)), pred, out)
print(float(r1.long_term_priority), bool(mem.state_valid(r1.state)))
r2 = mem.record(r1.state, pred, out)
print(bool(r2.state_valid), bool(r2.wrote_short_term), int(r2.state.accepted_transition_count))
```

Output on `main`:

```
1.0000001192092896 False
False False 1
```

Expected: `1.0 True` then `True True 2` — the priority is closed in `[0, 1]` by construction, and calibration scalars that collapse to zero or overflow once narrowed to float32 are rejected at construction.

## Dedup

`gh issue/pr list --search` for "dual replay", "calibrated priority", "insertion_priorit", "convex combination" — no matches open or closed; no open PR touches `core/dual_replay.py`. Same class as the merged float32-domain validation sweep (#288/#289, #301) but on an untouched module.

Prabhat1308:
Closes #313.

## Issue

`_calibrated_priority` divided a float32-accumulated numerator by a weight sum rounded once from float64. With weights such as `(0.1, 0.2, 0.4)` and saturated components the priority came out `1.0000001192092896`; it was stored as an insertion priority, `DualReplayState` failed its own `<= 1.0` invariant, and every later `record`/`sample` silently returned `state_valid=False` while `checkpoint_payload` raised. Calibration scalars were validated in float64 but consumed after a float32 cast, so `float32(1e-46) == 0.0` produced `0/0 = NaN` priorities and three `2e38` weights an `inf` weight sum.

## New state

The three weights are narrowed to float32 once, the weight sum is accumulated from those same values, and the combination is clipped to the documented `[0, 1]`.

Every float32-consumed config scalar (`surprise/coverage/progress_scale`, `surprise/coverage/progress_weight`, `aleatoric_downweight_scale`, `calibrated_priority_threshold`, `calibrated_replacement_margin`, `max_aleatoric_uncertainty`) now goes through the shared exact converter (`alberta_framework._float32.round_real_to_float32`) in `DualReplayConfig.__post_init__`: exact-domain check first (positive / non-negative / `[0, 1]`), then reject nonzero→zero underflow and finite→infinity overflow, keep explicit zero only for the threshold, margin, and aleatoric cap, preserve built-in payloads that narrow identically, and canonicalize other reals (`Fraction`, `np.float64`, `np.int64`, large ints that would double-round through float64) to the exact sink value. The float32 weight-sum probe runs inside an `np.errstate` boundary; the whole rejection path is warning-free under `-W error`. Default `(1.0, 1.0, 1.0)` weights are exact in float32, so every existing pinned value — including the eager/jit/scan bitwise-equality and checkpoint digest tests — is unchanged.

The domain is enforced on both the host value and the exact narrowed sink (`narrowed > 0` / `>= 0` / `<= maximum`), the type gate uses the actual class (`issubclass(type(value), Real)` with bool exclusions) so a `__class__` spoof cannot pass, and honest float subclasses are canonicalized to primitive `float`. Built-in `int`/`float` payloads that narrow identically are preserved as-is (`c48fe63`, maintainer).

Failing-test-first: `test_calibrated_priority_stays_within_unit_range_for_unequal_weights` fails on `main` and passes here. Public-path regressions added: dishonest `float` subclass whose `as_integer_ratio()` leaves the domain for a positive field, each non-negative field, and both sides of the threshold interval (`test_config_rejects_reals_whose_exact_ratio_leaves_the_domain`, 7 cases), `__class__` spoof rejection on all ten scalars, honest-subclass canonicalization with a strict `json.dumps(allow_nan=False)` round trip; underflow/overflow/non-finite rejection for all ten scalars without warnings (`test_config_rejects_scalars_that_underflow_or_overflow_float32_without_warnings`, 70 cases), exact float32 underflow/overflow midpoints (`test_config_enforces_exact_float32_underflow_and_overflow_midpoints`), explicit-zero preservation where allowed, ties-to-even rational rounding plus the `2**54 + 2**30 + 1` double-rounding integer, payload preservation/canonicalization round trip, and the warning-free overflowing weight triple.

## Evidence

Falsifier from #313 at this head:

```text
1.0 True
True True 2
```

Review items at this head:

```text
calibrated_priority_threshold=1e-46 -> rejected: calibrated_priority_threshold must not underflow to zero in float32
max_aleatoric_uncertainty=6e38     -> rejected: max_aleatoric_uncertainty must narrow to a finite float32
calibrated_replacement_margin=6e38 -> rejected: calibrated_replacement_margin must narrow to a finite float32
weights 2e38 x3                    -> rejected: ... must have a finite float32 sum
calibrated_priority_threshold=0.0  -> accepted, stored 0.0
```

Verification at `cbf8f720ec0c43af6f7abc4d230e5bdd5228faf2`:

```text
.venv/bin/python -m pytest tests/test_dual_replay.py tests/test_experiential_memory.py -q -o addopts="" -W error -> 202 passed (no warnings under strict policy)
.venv/bin/python -m ruff check .                                                  -> All checks passed!
.venv/bin/python -m mypy                                                          -> Success: no issues found in 165 source files
```

`core/dual_replay.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

- [x] logs: inline above (focused pytest incl. `-W error`, ruff, mypy, falsifier output at the evidence head)

### DIFF
--- a/alberta_framework/core/dual_replay.py
+++ b/alberta_framework/core/dual_replay.py
@@ -38,6 +38,7 @@ import json
 import math
 from collections.abc import Mapping
 from dataclasses import asdict, dataclass
+from numbers import Real
 from typing import Any, Literal, cast
 
 import chex
@@ -47,6 +48,8 @@ import jax.random as jr
 import numpy as np
 import numpy.typing as npt
 from jax import Array
+
+from alberta_framework._float32 import round_real_to_float32
 
 DUAL_REPLAY_CONFIG_SCHEMA = "alberta.dual-replay.config.v1"
 DUAL_REPLAY_CHECKPOINT_SCHEMA = "alberta.dual-replay.checkpoint.v1"
@@ -103,6 +106,31 @@ class DualReplayConfig:
     max_aleatoric_uncertainty: float = 1.0
     aleatoric_downweight_scale: float = 1.0
     max_persistent_bytes: int | None = None
+
+    def __post_init__(self) -> None:
+        """Validate float32-consumed scalars in their sink domain and canonicalize them."""
+        for name in _POSITIVE_FLOAT32_FIELDS:
+            object.__setattr__(
+                self,
+                name,
+                _require_float32_real(name, getattr(self, name), strictly_positive=True),
+            )
+        object.__setattr__(
+            self,
+            "calibrated_priority_threshold",
+            _require_float32_real(
+                "calibrated_priority_threshold",
+                self.calibrated_priority_threshold,
+                strictly_positive=False,
+                maximum=1,
+            ),
+        )
+        for name in ("calibrated_replacement_margin", "max_aleatoric_uncertainty"):
+            object.__setattr__(
+                self,
+                name,
+                _require_float32_real(name, getattr(self, name), strictly_positive=False),
+            )
 
     @property
     def long_term_capacity(self) -> int:
@@ -385,13 +413,59 @@ class DualReplayResourceAccounting:
     samples: Array
 
 
-def _validate_real(name: str, value: object) -> float:
-    if isinstance(value, bool) or not isinstance(value, int | float):
-        raise ValueError(f"{name} must be a real number")
-    result = float(value)
-    if not math.isfinite(result):
-        raise ValueError(f"{name} must be finite")
-    return result
+_POSITIVE_FLOAT32_FIELDS = (
+    "surprise_scale",
+    "coverage_scale",
+    "progress_scale",
+    "surprise_weight",
+    "coverage_weight",
+    "progress_weight",
+    "aleatoric_downweight_scale",
+)
+
+
+def _require_float32_real(
+    name: str,
+    value: object,
+    *,
+    strictly_positive: bool,
+    maximum: int | None = None,
+) -> float:
+    """Validate one host scalar in its exact domain and its float32 sink.
+
+    Zero is only admitted where the exact domain allows it; a nonzero value
+    that narrows to zero, or a finite value that narrows to infinity, is
+    rejected because the compiled sink would silently run a different rule.
+    Built-in payloads that already narrow exactly are preserved for
+    serialization; other reals are canonicalized to the exact sink value.
+    """
+    domain = "positive" if strictly_positive else "non-negative"
+    if maximum is not None:
+        domain = f"{domain} and at most {maximum}"
+    preserve_builtin_payload = type(value) is int or type(value) is float
+    if isinstance(value, bool | np.bool_) or not isinstance(value, Real):
+        raise ValueError(f"{name} must be finite and {domain}")
+    try:
+        if strictly_positive:
+            if value <= 0:
+                raise ValueError
+        elif value < 0:
+            raise ValueError
+        if maximum is not None and value > cast(Real, maximum):
+            raise ValueError
+        narrowed = round_real_to_float32(value)
+    except (FloatingPointError, OverflowError, TypeError, ValueError) as error:
+        raise ValueError(f"{name} must be finite and {domain}") from error
+    if not math.isfinite(narrowed):
+        raise ValueError(f"{name} must narrow to a finite float32")
+    if value != 0 and narrowed == 0.0:
+        raise ValueError(f"{name} must not underflow to zero in float32")
+    if not preserve_builtin_payload:
+        return narrowed
+    number = float(value)
+    with np.errstate(invalid="ignore", over="ignore", under="ignore"):
+        renarrowed = float(np.float32(number))
+    return number if narrowed == renarrowed else narrowed
 
 
 def _validate_positive_int(name: str, value: object) -> int:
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking exact-head facade/sink mismatch at 966f111fc645cc17b7af04cab722fba27d25b3c8. A positive `calibrated_priority_threshold=1e-46` is accepted, narrows to float32 zero, and admits a zero-priority duplicate; `max_aleatoric_uncertainty=6e38` and `calibrated_replacement_margin=6e38` narrow to infinity. Large built-in integers can also double-round through raw NumPy conversion, and current overflow tests leak warnings. Please validate every float32-consumed config scalar with the shared exact converter, preserve explicit zero only where allowed, reject nonzero-to-zero and finite-to-infinity, canonicalize sink-equivalent values, and add public-path underflow/overflow/exact-integer regressions before fresh gates/CI.

lalalune (changes requested):
Blocking fail-closed validation defect at exact head 966f111: the newly added invalid-input cases emit NumPy RuntimeWarnings from both the scalar cast and weight-sum addition (independently reproduced in the composed affected suite: 269 passed, but two warnings originate in dual_replay.py). A constructor rejection path must not leak floating-point warnings or become an error under strict warning policy. Normalize conversion overflow to ValueError and perform the intentional float32 overflow probe under an explicit no-warning/error-state boundary (or use the shared exact float32 helper plus a controlled arithmetic check). Add a regression asserting the out-of-range scalar and overflowing weight triple reject without RuntimeWarning, then rerun focused/static/fresh CI. The [0,1] clip and same-arithmetic weight sum otherwise look correct.

Prabhat1308:
Addressed both reviews in `6c463b0` (new commit on top of `966f111`):

- every float32-consumed config scalar (all seven positive scales/weights plus `calibrated_priority_threshold`, `calibrated_replacement_margin`, `max_aleatoric_uncertainty`) now goes through the shared `round_real_to_float32` converter in `DualReplayConfig.__post_init__` — exact-domain check, then nonzero→zero and finite→infinity rejected, explicit zero kept only for threshold/margin/aleatoric cap, built-in payloads preserved when they narrow identically, other reals (incl. large ints that would double-round via float64) canonicalized to the sink value;
- the float32 weight-sum probe runs under `np.errstate`; the full rejection path is warning-free (`pytest -W error`: 137 passed);
- regressions: 70 underflow/overflow/non-finite cases across the ten scalars under strict warnings, exact float32 underflow/overflow midpoints, explicit-zero preservation, ties-to-even rationals + the `2**54 + 2**30 + 1` integer, payload/canonicalization round trip, and the warning-free overflowing weight triple.

Body updated to evidence head `6c463b0`; `threshold=1e-46`, `max_aleatoric_uncertainty=6e38`, `margin=6e38` are now rejected (outputs in the body). Re-review requested.

lalalune (changes requested):
Blocking exact-domain/sink-domain disagreement at live head `6c463b008a31b8a00e457f9124f5ce91a441089d`. `_require_float32_real` checks bounds on the host object, then trusts `as_integer_ratio()` for narrowing, but only checks the narrowed result for finiteness and zero underflow. A real `float` subclass whose underlying value is `0.5` but whose `as_integer_ratio()` returns `(-1, 1)` is accepted as `surprise_scale=-1.0`, `surprise_weight=-1.0`, `calibrated_priority_threshold=-1.0`, `calibrated_replacement_margin=-1.0`, and `max_aleatoric_uncertainty=-1.0`; returning `(2, 1)` is accepted as `calibrated_priority_threshold=2.0`. All values survive `DualReplayMemory` construction.

Please enforce the declared domain on both the host value and the exact narrowed sink: `narrowed > 0` for positive fields, `narrowed >= 0` for non-negative fields, and `narrowed <= maximum` for bounded fields, while retaining the existing nonzero-to-zero rule. Add public constructor regressions using a dishonest float subclass for a positive field, each non-negative field, and both sides of the threshold interval under strict warnings.

The canonicalization does make accepted subclass values primitive `float`s and JSON serializable, which is good. However, a non-`Real` object with `__class__` spoofed to `float` currently makes `isinstance(value, Real)` true and is accepted/canonicalized. Please make the type gate depend on the actual class (for example `issubclass(type(value), Real)` plus the bool exclusions), add a spoof rejection regression, and retain a strict `json.dumps(..., allow_nan=False)` round-trip assertion for genuinely accepted exact-ratio reals. No merge until both host and sink domains agree.

Prabhat1308:
Addressed the third review in `24dd197` (rebased on top of your `727acad`, which I kept as-is):

- domain enforced on both host and sink: after `round_real_to_float32`, the exact narrowed value must satisfy `> 0` (positive fields), `>= 0` (non-negative fields), and `<= maximum` (threshold), with the nonzero→zero rule retained — a `_LyingFloat(0.5, (-1, 1))` / `(2, 1)` is now rejected on every field;
- type gate uses the actual class (`issubclass(type(value), Real)` plus `bool`/`np.bool_` exclusions), so a non-Real object spoofing `__class__ = float` is rejected on all ten scalars;
- regressions: 7 lying-subclass cases (positive field, each non-negative field, both threshold bounds), 10 spoof cases, and an honest-subclass canonicalization test with a strict `json.dumps(allow_nan=False)` round trip — all under `-W error` (155 passed).

Body updated to evidence head `24dd197`. Re-review requested.

lalalune (approved):
Independent exact-head review of 24dd197cd0ed9d7339f4c0f29da660ea32c6fe06: READY. Verified safe built-in payload/type preservation (including 2**54+1 and explicit integer zero), canonicalization of non-built-in exact reals, exact host and float32 sink domain checks including dishonest-ratio subclasses, strict round-trip behavior, 155 warning-as-error focused tests, Ruff, strict mypy, and 184-test clean composition with current main. No registered five-claim evidence source/artifact changes. Merge only after this exact head has 20/20 CI.

PR #320, opened 2026-08-16
fix(screening): fail closed on sentinel probes for hidden-RMS discovered rules

### LINKED ISSUE #319: Recurring-retention sentinel probes score disc_r1/r2/r3/disc_r1_pscale with a forward pass they never deploy (flag_hidden_rms bypasses the hidden-RMS guard)
## Summary

`run_recurring_ipmnist_retention_development` (`alberta_framework/benchmarks/ipmnist_screening.py`) scores every sentinel probe with `mlp_logits(params, spec.frozen_probe_input(state, sentinel_inputs, hp))`. `_hidden_rms_frozen_probe_input` exists precisely to refuse this for arms whose deployed forward pass RMS-normalizes the hidden layers — its docstring: *"the probe harness computes logits with `mlp_logits`, so any input-side transform would silently probe the wrong model. Failing closed here is the honest option."* It is wired to `sigma0_hidden_norm` (hyperparameter key `hidden_rms`).

The discovered-rule family switches on the same hidden-RMS forward pass through a different key, `flag_hidden_rms` (`_make_discovered_rule_learner`: `_rms_loss` / `_loss_with_hidden` normalize both ReLU layers). `disc_r1`, `disc_r2`, `disc_r3` (all `flag_hidden_rms=1.0`) and `disc_r1_pscale` (`flag_hidden_rms=1.0`) are nevertheless registered with `frozen_probe_input=_ema_frozen_probe_input`, so the recurring-retention harness scores the plain MLP on those arms' parameters — a model they never run.

The online accuracy trace is unaffected (it comes from the arm's own step function). Everything the retention lane exists to measure is: `acquisition_end / pre_revisit / revisit_end` sentinel accuracy and hence `peak_to_revisit_forgetting`, `retention_change_from_acquisition`, and `revisit_recovery`.

Registry scan:

```python
import alberta_framework.benchmarks.ipmnist_screening as S
for name, spec in S.SCREENING_REGISTRY.items():
    hp = spec.hyperparameters
    if hp.get("hidden_rms", 0.0) or hp.get("flag_hidden_rms", 0.0):
        print(name, spec.frozen_probe_input.__name__)
```

```
sigma0_hidden_norm _hidden_rms_frozen_probe_input
disc_r1 _ema_frozen_probe_input
disc_r2 _ema_frozen_probe_input
disc_r3 _ema_frozen_probe_input
disc_r1_pscale _ema_frozen_probe_input
```

End-to-end on a 16-dim / 4-class synthetic stream (`n_tasks=3, task_length=400, hidden 64/32`, phases `(400,400,400)`, 100 sentinels, seed 11) — reported sentinel accuracy from the harness vs. the same frozen `params` scored with the arm's actual hidden-RMS forward pass:

```
=== disc_r1_pscale (flag_hidden_rms=1.0) ===
  phase perm  reported(mlp_logits)  deployed(hidden-RMS forward)
    0    A      0.8200                 0.7600   delta=-0.0600
    1    A      0.3200                 0.3400   delta=+0.0200
    1    B      0.7500                 0.8100   delta=+0.0600
    2    A      0.8200                 0.8000   delta=-0.0200
  REPORTED : acq=0.8200 pre_revisit=0.3200 revisit_end=0.8200 forgetting=0.5000 recovery=0.5000
  DEPLOYED : acq=0.7600 pre_revisit=0.3400 revisit_end=0.8000 forgetting=0.4200 recovery=+0.4600

=== disc_r1_pscale_norms (flag_hidden_rms=0.0, same factory)  <- control
    every delta=+0.0000, REPORTED == DEPLOYED
```

The `flag_hidden_rms=0.0` sibling from the same factory shows the discrepancy is caused by the flag, not by noise. `disc_r1` shows the same pattern (forgetting 0.48 reported vs 0.45 deployed).

Expected: no arm whose forward pass diverges from `mlp_logits` can be registered with an input-side probe. On `main` that means routing the four `flag_hidden_rms` arms to the fail-closed `_hidden_rms_frozen_probe_input`, exactly like `sigma0_hidden_norm`.

## Relation to #190

The draft #190 replaces the fail-closed probe with `_hidden_rms_frozen_probe_logits` (keyed on `hidden_rms_epsilon`) and rewires `sigma0_hidden_norm`, `rff_rls`, `lin_rls`, `naive_bayes`, `nb_ensemble`, `rls_head` — but its diff has zero hits for `flag_hidden_rms` / `disc_r`, so the four discovered-rule arms keep `frozen_probe_logits=None` and still fall back to `mlp_logits` after #190. Whichever lands, these arms need the discovered-rule forward (`_hidden_rms_normalize` on both hidden layers with `hp["hidden_rms_epsilon"]`) or a fail-closed refusal.

## Dedup

`gh issue/pr list --search` for "hidden_rms", "flag_hidden_rms", "frozen_probe", "sentinel probe", "disc_r1", "recurring retention": only #18/#46 (sigma0 reduction pins, closed) and #190 (above). No issue covers this.

Fix PR follows (failing-test-first; fail-closed on `main`, composable with #190's direction).

Prabhat1308:
Closes #319.

## Issue

`disc_r1`, `disc_r2`, `disc_r3`, and `disc_r1_pscale` set `flag_hidden_rms=1.0`, which RMS-normalizes both hidden layers inside their deployed forward pass, but were registered with the input-side `_ema_frozen_probe_input`. The recurring-retention harness therefore scored `mlp_logits` on a model those arms never run: on the #319 synthetic stream `disc_r1_pscale` reported `forgetting=0.50 / recovery=0.50` while the deployed forward gives `0.42 / 0.46`, and its `flag_hidden_rms=0.0` sibling shows zero discrepancy.

## New state

`_discovered_rule_frozen_probe_input(hp)` selects the probe from `flag_hidden_rms`: hidden-RMS discovered rules get the existing fail-closed `_hidden_rms_frozen_probe_input` (exactly like `sigma0_hidden_norm`); the rest keep the EMA input probe. No arm whose forward pass diverges from `mlp_logits` can be sentinel-probed with it. Arm hyperparameters, factories, and every online-trace path are untouched, so no shard or summary changes.

Failing-test-first: `test_hidden_rms_active_arms_refuse_plain_mlp_sentinel_probes` (registry-wide invariant over both `hidden_rms` and `flag_hidden_rms`) fails on `main` (`DID NOT RAISE NotImplementedError`) and passes here; `test_hidden_rms_inactive_sibling_keeps_its_input_side_probe` pins the control. The existing `test_discovered_arms_registered` pinned the defective probe and now derives the expected probe from the flag.

Composable with #190: once a per-arm `frozen_probe_logits` harness lands, these four arms can be routed to a hidden-RMS logits probe (using the discovered rule's `hidden_rms_epsilon`) instead of failing closed; until then failing closed is the honest state.

## Evidence

At this head the #319 falsifier now stops at the probe:

```text
NotImplementedError: sentinel probes are unsupported for hidden-RMS-normalized arms: the deployed forward pass is not the plain protocol MLP
```

Verification at `0ab16256f3dc4df9f93d6f76aac428ec0dd4a022`:

```text
.venv/bin/python -m pytest tests/test_ipmnist_recurring_adapter.py -q -o addopts=""   -> 9 passed
.venv/bin/python -m pytest tests/test_ipmnist_screening.py -q -o addopts=""           -> 306 passed
.venv/bin/python -m ruff check .                                                       -> All checks passed!
.venv/bin/python -m mypy                                                               -> Success: no issues found in 164 source files
```

`benchmarks/ipmnist_screening.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -5120,6 +5120,21 @@ def _hidden_rms_frozen_probe_input(
     )
 
 
+def _discovered_rule_frozen_probe_input(
+    hyperparameters: Mapping[str, float],
+) -> FrozenProbeInputFn:
+    """Select the sentinel probe for a discovered-rule arm from its flags.
+
+    ``flag_hidden_rms`` switches the discovered-rule forward pass to the same
+    hidden-layer RMS normalization as ``sigma0_hidden_norm``, so those arms
+    must fail closed exactly like it; the remaining discovered rules deploy
+    the plain MLP behind an EMA input normalizer.
+    """
+    if float(hyperparameters.get("flag_hidden_rms", 0.0)) != 0.0:
+        return _hidden_rms_frozen_probe_input
+    return _ema_frozen_probe_input
+
+
 def _rff_frozen_probe_input(
     state: Any, observation: Array, hyperparameters: Mapping[str, float]
 ) -> Array:
@@ -5896,7 +5911,7 @@ def _build_registry() -> dict[str, ScreeningSpec]:
                 mechanism="discovered_rule",
                 hyperparameters=_discovered_rule_hp(**disc_hp),
                 factory=_make_discovered_rule_learner,
-                frozen_probe_input=_ema_frozen_probe_input,
+                frozen_probe_input=_discovered_rule_frozen_probe_input(disc_hp),
                 description=disc_description,
             )
         )
@@ -5925,7 +5940,9 @@ def _build_registry() -> dict[str, ScreeningSpec]:
                     surprise_slow=0.9996305719081341,
                 ),
                 factory=_make_discovered_rule_learner,
-                frozen_probe_input=_ema_frozen_probe_input,
+                frozen_probe_input=_discovered_rule_frozen_probe_input(
+                    {"flag_hidden_rms": diag_rms}
+                ),
                 description=(
                     "disc_r1 structure at champion-scale constants "
                     f"({diag_axis}): shift-adaptive norm + surprise-gated "
--- a/tests/test_ipmnist_recurring_adapter.py
+++ b/tests/test_ipmnist_recurring_adapter.py
@@ -6,18 +6,21 @@ import dataclasses
 from collections.abc import Mapping
 from typing import Any
 
+import jax.numpy as jnp
+import jax.random as jr
 import numpy as np
 import pytest
 from jax import Array
 
 from alberta_framework.benchmarks.ipmnist_screening import (
+    SCREENING_REGISTRY,
     build_recurring_ipmnist_online_indices,
     ipmnist_permutation_sha256,
     ipmnist_sentinel_set_sha256,
     run_recurring_ipmnist_retention_development,
     screening_spec,
 )
-from alberta_framework.benchmarks.upgd_ipmnist import IPMNISTConfig
+from alberta_framework.benchmarks.upgd_ipmnist import IPMNISTConfig, init_mlp_params
 
 pytestmark = pytest.mark.unit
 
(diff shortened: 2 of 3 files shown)


lalalune (approved):
Independent review found the fail-closed routing sound: every discovered rule with flag_hidden_rms active now refuses the plain-MLP sentinel path, while the inactive sibling retains the EMA input probe. Current-main combined screening/micro/rule/retention panel: 537 passed; Ruff and strict mypy clean. No outputs or registered evidence source changed. Merge only after exact-head CI is complete.

PR #328, opened 2026-08-16
fix(ia): reject nonfinite cerebellum step sizes

### LINKED ISSUE #327: ExoCerebellumConfig accepts non-finite step_size, staging NaN weights and silently blocking updates
`ExoCerebellumConfig.__post_init__` only checked `step_size <= 0.0`, which `NaN` and `+inf` both fail to satisfy (any comparison against `NaN` is `False`, and `inf <= 0.0` is `False`), so both constructed successfully. With `step_size=NaN`, a normal finite update stages NaN weights, the candidate-state validity check correctly rejects the transaction, and the learner is left permanently unchanged with no error surfaced at construction time — the bad value should have been rejected at config construction, not silently neutered at update time.

confirmed directly:
```text
constructed_step_size=nan
input_valid=True
candidate_state_valid=False
update_applied=False
weights_before=[[0.0]]
weights_after=[[0.0]]
```

reachability: `ExoCerebellumConfig` is exported from `alberta_framework.core`. The Step 12 public facade (`alberta_framework/steps/step12.py:140`) and the continual-IA evaluation harness (`alberta_framework/evaluation/continual_ia.py:362`) both construct it directly, so this is reachable from real entry points, not just an internal helper.

fix incoming: add `math.isfinite(self.step_size)` to the existing validation, matching the finite-guard pattern already used elsewhere in this config class.

Svector-anu:
Closes #327.

## What changed

`ExoCerebellumConfig.__post_init__` previously checked only `step_size <= 0.0`. The original PR added a host-finiteness check, but finite binary64 values could still become unusable at the learner's float32 sink:

- `1e100` narrowed to `inf`, so a normal public update staged invalid weights and was refused.
- `1e-100` narrowed to `0.0`, so the public update reported success while the weight stayed zero.
- bools and array-like values also passed the nominal scalar field.

The repaired head validates the original value as a non-bool `numbers.Real`, rounds it directly to IEEE binary32 through the shared exact-ratio converter, and rejects non-finite or non-positive float32 sinks. Non-builtin `Real` values and numeric subclasses are reduced to a plain built-in `float`, so caller-defined behavior cannot survive in the frozen config or its JSON payload. Ordinary valid built-in floats retain their existing binary64/JSON payload when that payload has the same valid float32 execution sink; exactly representable built-in integers retain the same compatibility behavior.

## Reachability

`ExoCerebellumConfig` is exported from `alberta_framework.core`. Both the Step 12 public facade (`alberta_framework/steps/step12.py:140`) and the continual-IA evaluation harness (`alberta_framework/evaluation/continual_ia.py:362`) construct it directly.

The regression covers the public `ExoCerebellumAgent.update_result` path, direct exact-ratio rounding above a binary32 midpoint, strict JSON serialization, and preservation of the existing `0.05` built-in payload.

## Verification

Exact head `402e232568537c1624565f7f63a5dff2e2719224` was composed without conflict onto main `45279f31d0a9e234cd9664a0ba8688e3203f75c5`.

```text
$ .venv/bin/python -m pytest \
    tests/test_intelligence_amplification_horizon.py \
    tests/test_step12_production.py \
    tests/test_continual_ia_evidence.py \
    tests/test_evidence_manifest.py \
    -q -o addopts=""
217 passed in 371.95s

All checks passed!

$ .venv/bin/python -m mypy
```

Independent adversarial probes also covered `Fraction`, NumPy scalar types, huge integers, overflow/subnormal rounding boundaries, `IntEnum`, a virtual `Real`, and hostile `float` subclasses. Every accepted value stored and serialized as an exact built-in finite positive float; invalid sinks and bool/array impostors failed at construction.

## Registered-evidence impact

`alberta_framework/core/intelligence_amplification.py` is one of the eight registered source paths for the historical continual-IA claim. This correctness change does not modify or rewrite any pinned `outputs/` artifact and does not create, promote, or strengthen scientific evidence. The historical artifact remains evidence only for its pinned source bytes.

## Evidence

402e232568537c1624565f7f63a5dff2e2719224

```text
red (parent 2236232, repaired cases absent):
step_size=1e+100
constructed=1e+100
candidate_state_valid=False
update_applied=False
weight=0.0
step_size=1e-100
constructed=1e-100
candidate_state_valid=True
update_applied=True
weight=0.0

green (402e232):
rejected=float:1e+100
rejected=float:1e-100
rejected=bool:True
rejected=bool:np.True_
rejected=ndarray:array(0.05)
canonical_type=float
canonical_step_size=1.0000001192092896
update_applied=True
weight=1.0000001192092896
```

```text
strict_json={"type": "ExoCerebellumConfig", "n_demons": 1, "obs_dim": 1, "step_size": 1.0000001192092896}
```
The midpoint input is `1 + 2^-24 + 2^-60`; direct ties-to-even binary32 rounding yields the next float above `1.0`, and a unit-input/unit-target public update from zero moves the weight by exactly that canonical sink.

## Provenance

AI provider/model: openai / gpt-5.6-sol (original fix, float32-domain repair, tests, and exact-head audit); anthropic / claude-sonnet-5 (original independent verification and PR creation)
Client / agent tooling: Codex; Claude Code

### DIFF
--- a/alberta_framework/core/intelligence_amplification.py
+++ b/alberta_framework/core/intelligence_amplification.py
@@ -37,6 +37,7 @@ References:
 from __future__ import annotations
 
 import dataclasses
+import math
 from collections.abc import Mapping
 from typing import Any, cast
 
@@ -111,7 +112,7 @@ class ExoCerebellumConfig:
             raise ValueError("n_demons must be positive")
         if self.obs_dim <= 0:
             raise ValueError("obs_dim must be positive")
-        if self.step_size <= 0.0:
+        if not math.isfinite(self.step_size) or self.step_size <= 0.0:
             raise ValueError("step_size must be positive")
 
     def to_config(self) -> dict[str, Any]:
--- a/alberta_framework/core/intelligence_amplification.py
+++ b/alberta_framework/core/intelligence_amplification.py
@@ -39,6 +39,7 @@ from __future__ import annotations
 import dataclasses
 import math
 from collections.abc import Mapping
+from numbers import Real
 from typing import Any, cast
 
 import chex
@@ -47,6 +48,7 @@ import jax.numpy as jnp
 from jax import Array
 from jaxtyping import Bool, Float, Int, UInt
 
+from alberta_framework._float32 import round_real_to_float32
 from alberta_framework.core.normalizers import (
     _checked_lifetime_words_increment,
     _lifetime_counter_valid,
@@ -85,6 +87,33 @@ RECOMMENDATION_PROTOCOL_LIFETIME_COUNTER_DELTA_NBYTES = 24
 
 _INT32_MAX = 2**31 - 1
 
+
+def _positive_float32_scalar(name: str, value: object) -> float:
+    """Validate and canonicalize a positive scalar in its execution dtype."""
+    message = f"{name} must be a finite positive float32 scalar"
+    preserve_builtin_float = type(value) is float
+    preserve_builtin_int = type(value) is int
+    if isinstance(value, bool) or not isinstance(value, Real):
+        raise ValueError(message)
+    try:
+        if value <= 0:
+            raise ValueError
+        narrowed = round_real_to_float32(value)
+    except (FloatingPointError, OverflowError, TypeError, ValueError) as error:
+        raise ValueError(message) from error
+    if not math.isfinite(narrowed) or narrowed <= 0.0:
+        raise ValueError(message)
+
+    # A valid built-in float already has an exact binary64 payload, so keeping
+    # it preserves existing serialized configs without changing its float32
+    # sink. Other Real implementations are stored as that exact sink. A small
+    # built-in integer is likewise safe only when it is exactly representable.
+    if preserve_builtin_float:
+        return cast(float, value)
+    if preserve_builtin_int and value == narrowed:
+        return float(value)
+    return narrowed
+
 # ---------------------------------------------------------------------------
 # Exo-cerebellum
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lalalune (approved):
Exact-head diff reviewed and composed onto current main. NaN and both infinities now fail at config construction before candidate-state rejection can silently neutralize learning. Combined affected suites: 269 passed. Approved pending full CI.

lalalune (changes requested):
Blocking execution-domain gap: the new math.isfinite guard only validates the host scalar, while the learner consumes step_size as float32.

At exact head 2236232:
- ExoCerebellumConfig(step_size=1e100) constructs; the float32 sink is inf, and a finite public update_result reports candidate_state_valid=False / update_applied=False.
- step_size=1e-100 constructs; the sink is 0.0, so update_applied=True but the weights never move.
- bools and non-scalar array-like values are also accepted despite the scalar-float contract.

This is the same poisoned/silently-neutered configuration class described by #327, just reached from finite host values. Validate the exact original scalar and its direct IEEE float32 narrowing using the shared exact converter, reject finite-to-inf and nonzero-to-zero, canonicalize non-builtin Real values to the sink, and add public update plus serialization regressions.

lalalune (approved):
Exact-head review at 402e232: direct exact-ratio float32 narrowing, strict-positive sink rejection, subclass/virtual-Real canonicalization, strict JSON serialization, and public update behavior verified. The head composes cleanly onto main 45279f3; repository-wide ruff and mypy pass. Approved pending the 20-job CI fan-out; no evidence artifact was modified and no scientific claim is promoted.

PR #330, opened 2026-08-16
fix(retention): require distinct frozen states across sentinel checkpoints

### LINKED ISSUE #329: Recurring-retention validator accepts one frozen learner state at every checkpoint, so re-scoring a single state reports zero forgetting
## Summary

`_validate_sentinel_snapshots` (`alberta_framework/evaluation/recurring_ipmnist_retention.py`) is the strict contract behind the A/B/A retention report. The module docstring states its purpose: the probes after B and before the second A "measure A directly **without allowing A updates to mask forgetting**", and `METRIC_DEFINITIONS["pre_revisit_sentinel_accuracy"]` binds that score to "the frozen learner state after the intervening phase and before any revisit update".

The validator enforces that all probes sharing one `checkpoint_step` declare one `learner_state_sha256_before`, but never that **different** checkpoint steps declare **different** states. A caller that re-scores every required probe from a single frozen learner (e.g. the final post-revisit state) passes every check, and the report comes out as a fully validated "no forgetting, no recovery" measurement — the exact failure mode the schedule exists to prevent. Between any two checkpoints the learner consumes a whole phase of online updates, so equal digests can only mean the same state was reused.

Using the fixtures from `tests/test_recurring_ipmnist_retention.py` (`_protocol()`, `_trace(retaining=True)`, `_snapshots(...)`) but with one digest at every checkpoint:

```python
import dataclasses
from alberta_framework.evaluation.recurring_ipmnist_retention import build_recurring_ipmnist_retention_report
from tests.test_recurring_ipmnist_retention import _protocol, _sha, _snapshots, _trace

protocol = _protocol()
one_state = _sha("7")
snapshots = tuple(
    dataclasses.replace(s, learner_state_sha256_before=one_state, learner_state_sha256_after=one_state)
    for s in _snapshots(protocol, retaining=True))
report = build_recurring_ipmnist_retention_report(protocol=protocol, trace=_trace(retaining=True), sentinel_snapshots=snapshots)
print(sorted({s.checkpoint_step for s in snapshots}), len({s.learner_state_sha256_before for s in snapshots}))
print({k: report.metrics[k] for k in ("peak_to_revisit_forgetting", "revisit_recovery")})
```

Output on `main` (report accepted):

```
[4, 8, 12] 1
{'peak_to_revisit_forgetting': 0.0, 'revisit_recovery': 0.0}
```

Expected: `ValueError` — probes at different checkpoints must declare distinct frozen learner states.

## Dedup

Related open: #319 (mine — the adapter probes four arms with the wrong forward pass; different layer, `benchmarks/ipmnist_screening.py`). No issue or PR touches `evaluation/recurring_ipmnist_retention.py`; #190 does not include it in its file list.

Prabhat1308:
Closes #329.

## Issue

`_validate_sentinel_snapshots` enforced one learner state per checkpoint but accepted the same state at every phase boundary. Re-scoring one frozen learner at A0, B0, and A1 produced a fully validated report with `peak_to_revisit_forgetting = 0.0` and `revisit_recovery = 0.0` — the masking the A/B/A schedule exists to prevent.

## New state

After the per-checkpoint check, the validator groups checkpoints by declared learner-state digest and refuses when more than one checkpoint step shares one: `sentinel probes at different checkpoints must use distinct frozen states; checkpoint steps [4, 8, 12] all declare one learner state`. Every genuine run advances the learner across a phase between checkpoints, so accepted reports are unchanged.

Failing-test-first: `test_sentinel_probes_at_different_checkpoints_must_use_distinct_frozen_states` fails on `main` (report accepted) and passes here; `test_sentinel_probes_at_one_checkpoint_must_share_one_frozen_state` pins the pre-existing per-checkpoint rule alongside it.

## Evidence

Falsifier from #329 at this head:

```text
ValueError: sentinel probes at different checkpoints must use distinct frozen states; checkpoint steps [4, 8, 12] all declare one learner state
```

Verification at `8520c1f71e0187141afabedadd7bdcf1a75d3ef6`:

```text
.venv/bin/python -m pytest tests/test_recurring_ipmnist_retention.py tests/test_ipmnist_recurring_adapter.py -q -o addopts=""   -> 17 passed
.venv/bin/python -m ruff check .                                                                                             -> All checks passed!
.venv/bin/python -m mypy                                                                                                     -> Success: no issues found in 164 source files
```

`evaluation/recurring_ipmnist_retention.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

### DIFF
--- a/alberta_framework/evaluation/recurring_ipmnist_retention.py
+++ b/alberta_framework/evaluation/recurring_ipmnist_retention.py
@@ -637,6 +637,15 @@ def _validate_sentinel_snapshots(
         )
         if prior_state != snapshot.learner_state_sha256_before:
             raise ValueError("all sentinel probes at one checkpoint must use one frozen state")
+    states_by_digest: dict[str, list[int]] = {}
+    for checkpoint_step, digest in checkpoint_states.items():
+        states_by_digest.setdefault(digest, []).append(checkpoint_step)
+    for shared_steps in states_by_digest.values():
+        if len(shared_steps) > 1:
+            raise ValueError(
+                "sentinel probes at different checkpoints must use distinct frozen states; "
+                f"checkpoint steps {sorted(shared_steps)} all declare one learner state"
+            )
     return resolved
 
 
--- a/tests/test_recurring_ipmnist_retention.py
+++ b/tests/test_recurring_ipmnist_retention.py
@@ -243,6 +243,44 @@ def test_missing_reordered_or_malformed_sentinel_snapshots_fail_closed() -> None
         dataclasses.replace(snapshots[0], learner_state_sha256_after=_sha("f"))
 
 
+def test_sentinel_probes_at_one_checkpoint_must_share_one_frozen_state() -> None:
+    protocol = _protocol()
+    snapshots = _snapshots(protocol, retaining=True)
+    drifted = dataclasses.replace(
+        snapshots[2], learner_state_sha256_before=_sha("9"), learner_state_sha256_after=_sha("9")
+    )
+    with pytest.raises(ValueError, match="one frozen state"):
+        build_recurring_ipmnist_retention_report(
+            protocol=protocol,
+            trace=_trace(retaining=True),
+            sentinel_snapshots=(*snapshots[:2], drifted, *snapshots[3:]),
+        )
+
+
+def test_sentinel_probes_at_different_checkpoints_must_use_distinct_frozen_states() -> None:
+    """Re-scoring one learner state at every boundary would report zero forgetting."""
+    protocol = _protocol()
+    one_state = _sha("7")
+    snapshots = tuple(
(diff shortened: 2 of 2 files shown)


lalalune (approved):
Independent review confirmed pairwise-distinct checkpoint-state binding after the existing same-checkpoint invariant. The production adapter hashes params, mechanism state, and the advancing learner RNG, so valid phase boundaries remain distinct. Current-main combined panel: 537 passed; Ruff and strict mypy clean; no output/registered-source change. Merge only after exact-head CI.

lalalune (approved):
Independent review found the distinct checkpoint-state invariant correct; affected and downstream panels passed and static gates are clean.

PR #326, opened 2026-08-16
fix(rule-discovery): reject duplicate search and holdout task names

Prabhat1308:
Closes #325.

## Issue

`evaluate_suite` evaluated a repeated task name twice, stored it once in `per_task`, and counted it twice in the task mean, so `('M1', 'M1', 'M2')` returned `2/3·M1 + 1/3·M2 = 0.0917` where the documented equal-weight mean is `0.1125`. `run_search` guards duplicate seeds (#127/#136) but accepted duplicate `task_names`/`holdout_names`, and the CLI wrote the duplicate into `settings.task_names`.

## New state

`_require_unique_task_names` refuses empty or repeated task lists (`task_names must contain unique task names; got […]`), applied in `evaluate_suite` and in `run_search`'s pre-compute validation block for both `task_names` and `holdout_names` — the same shape and placement as the existing eval-seed guard. Unique inputs are unchanged; `settings` still serializes lists.

Failing-test-first: `test_evaluate_suite_rejects_duplicate_task_names_before_evaluation[…]` (2 cases) and `test_cli_rejects_duplicate_task_names_before_search[--tasks|--holdout-tasks]` fail on `main` and pass here; `test_evaluate_suite_is_an_equal_weight_task_mean` pins the documented mean.

## Evidence

Falsifier from #325 at this head:

```text
ValueError: task_names must contain unique task names; got ['M1', 'M1', 'M2']
```

Verification at `b39b159b784ff2c8d48f958ebcf35219837e7250`:

```text
.venv/bin/python -m pytest tests/test_rule_discovery.py -q -o addopts=""   -> 42 passed
.venv/bin/python -m ruff check .                                            -> All checks passed!
.venv/bin/python -m mypy                                                    -> Success: no issues found in 164 source files
```

`benchmarks/rule_discovery.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, ruff, mypy, falsifier output at the evidence head)

Provider/model/client: anthropic / claude-fable-5 / Claude Code 2.1.233 (contribute-to-asi skill revision 72e4239e).

PR #332, opened 2026-08-16
fix(rule-discovery): validate genome shape and batch size before evaluation

### LINKED ISSUE #331: evaluate_population accepts a 16-wide genome (broadcasting one gene onto all 18 constants) and loops forever on batch_size <= 0
## Summary

`evaluate_population` (`alberta_framework/benchmarks/rule_discovery.py`) is the executed-genome path behind the search fitness, the archive, and `outputs/rule_discovery/*` replays. Two of its inputs are unvalidated:

1. **Genome width.** The module states an explicit contract — `GENOME_SIZE` raw floats (15 flags + 18 constants) — and `decode_genome` enforces it (`genome must have shape (33,)`). The executed path does not: `genomes = jnp.asarray(genomes, dtype=jnp.float32)` is sliced by `rule_step`, and `_param_values` computes `_LOS + clip(raw, 0, 1) * (_HIS - _LOS)` against the 18-element bounds, so a **16-wide** matrix (15 flags + one gene) broadcasts that single gene onto **all 18 constants** and returns a bounded, plausible accuracy. Other malformed widths fail loudly but unhelpfully from inside the traced step (`IndexError`, `TypeError: mul got incompatible shapes`). `GENOME_SIZE` already moved 20 → 33 in the wave-2 expansion, and `outputs/rule_discovery/crossval_gauss.py` feeds `search[...]["genome"]` lists straight into the batched runner, so width 16 is the one malformed shape that survives silently into `penalized_fitness` and result JSON.

2. **`batch_size <= 0`.** `stop = min(start + batch_size, n_genomes)` evaluates to `start`, the padding guard is false, `_batched_run` is called on an empty chunk, and `start = stop` never advances — a silent non-terminating loop, reachable from the CLI (`--batch-size`, `type=int`, no bounds) after the "rule-discovery search: … initial candidates" log line has already printed.

## Falsifiers (deterministic, pristine `main` @ `bae17ee`)

```python
import jax.numpy as jnp
import alberta_framework.benchmarks.rule_discovery as rd
from alberta_framework.benchmarks.micro_continual import MICRO_SUITE

wide16 = jnp.zeros((2, 16), dtype=jnp.float32).at[:, 15].set(0.5)
print(rd.evaluate_population(wide16, MICRO_SUITE["M1"], seeds=(0,)))     # accepted, in-range
try:
    rd.decode_genome(wide16[0])
except ValueError as exc:
    print("decode_genome:", exc)
```

```
[0.25 0.25]
decode_genome: genome must have shape (33,), got (16,)
```

```python
import signal
signal.signal(signal.SIGALRM, lambda *_: (_ for _ in ()).throw(TimeoutError("still spinning")))
signal.alarm(6)
rd.evaluate_population(jnp.zeros((2, rd.GENOME_SIZE), dtype=jnp.float32), MICRO_SUITE["M1"], seeds=(0,), batch_size=0)
```

```
TimeoutError: still spinning      # same for batch_size=-4
```

Expected: `evaluate_population` rejects `genomes` that are not `(n_genomes, GENOME_SIZE)` and `batch_size` that is not a positive built-in `int`, before any stream is materialized (matching the placement of the existing seed guard).

## Dedup

Adjacent, closed: #127 / PR #136 (duplicate seeds in `evaluate_population`), #298 and #147/#148 (same silent-broadcast class in stream wrappers). Open #325/#326 (mine) touches `evaluate_suite`/`run_search`, not this function; merge-tree composes cleanly. No issue or PR mentions genome shape or `batch_size`.

Prabhat1308:
Closes #331.

## Issue

`evaluate_population` accepted any genome width — a 16-wide matrix broadcast its single continuous gene onto all 18 constants and returned a plausible in-range accuracy (`[0.25 0.25]`) for input `decode_genome` refuses — and looped forever on `batch_size <= 0` (CLI-reachable through `--batch-size`).

## New state

Before any stream is materialized, `evaluate_population` requires `genomes` to be a `(n_genomes, GENOME_SIZE)` matrix (`genomes must have shape (n_genomes, 33), got (2, 16)`) and `batch_size` to be a positive built-in `int`. Placement and message shape follow the existing seed guard; valid inputs are unchanged.

Failing-test-first: `test_evaluate_population_rejects_malformed_genome_shapes_before_materialization[…]` (4 shapes: 16-wide, 34-wide, 1-D, 3-D) and `test_evaluate_population_rejects_non_positive_batch_size_before_materialization[0|-4|True]` fail on `main` and pass here (the `batch_size` cases previously hung; the width cases either returned numbers or raised from inside the traced step).

## Evidence

Falsifiers from #331 at this head:

```text
ValueError: genomes must have shape (n_genomes, 33), got (2, 16)
ValueError: batch_size must be a positive built-in int, got 0
```

Verification at `7fde9d634f5c62d263859ed6c38212c9a3adbb3f`:

```text
.venv/bin/python -m pytest tests/test_rule_discovery.py -q -o addopts=""   -> 44 passed
.venv/bin/python -m ruff check .                                            -> All checks passed!
.venv/bin/python -m mypy                                                    -> Success: no issues found in 164 source files
```

Composes cleanly with #326 (`git merge-tree` clean; different functions). `benchmarks/rule_discovery.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

### DIFF
--- a/alberta_framework/benchmarks/rule_discovery.py
+++ b/alberta_framework/benchmarks/rule_discovery.py
@@ -778,9 +778,20 @@ def evaluate_population(
 
     Paired evaluation: every genome sees the identical stream and identical
     network init per seed (the screening convention).
+
+    Raises:
+        ValueError: If ``seeds`` are not unique valid seeds, ``genomes`` is not
+            a ``(n_genomes, GENOME_SIZE)`` matrix, or ``batch_size`` is not a
+            positive built-in ``int``.
     """
     seeds = require_unique_jax_seeds(seeds, name="seeds")
+    if type(batch_size) is not int or batch_size < 1:
+        raise ValueError(f"batch_size must be a positive built-in int, got {batch_size!r}")
     genomes = jnp.asarray(genomes, dtype=jnp.float32)
+    if genomes.ndim != 2 or genomes.shape[1] != GENOME_SIZE:
+        raise ValueError(
+            f"genomes must have shape (n_genomes, {GENOME_SIZE}), got {tuple(genomes.shape)}"
+        )
     n_genomes = int(genomes.shape[0])
     total = np.zeros((n_genomes,), dtype=np.float64)
     for seed in seeds:
--- a/tests/test_rule_discovery.py
+++ b/tests/test_rule_discovery.py
@@ -18,6 +18,7 @@ Search executions happen through the CLI, never inside pytest.
 """
 
 import dataclasses
+import re
 
 import jax.numpy as jnp
 import jax.random as jr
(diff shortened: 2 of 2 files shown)


lalalune (approved):
Independent review confirmed shape and positive built-in batch-size checks run before stream materialization and close both broadcast and non-terminating-loop paths. The paired PR326 composes cleanly. Current-main combined panel: 537 passed; Ruff and strict mypy clean; no output/registered-source change. Merge only after exact-head CI.

lalalune (approved):
Independent review found the genome/batch preflight correct and composable with current main; affected panels and static gates are clean.

PR #335, opened 2026-08-16
fix(upgd): reject exhausted utility clocks atomically

### LINKED ISSUE #334: CanonicalUPGD lifetime step/utility-age counters wrap around at INT32_MAX
`CanonicalUPGD.update` increments the global optimizer step and safe-profile per-parameter utility ages with ordinary signed-int32 addition. At `2_147_483_647` (INT32_MAX), the next update wraps both counters to `-2_147_483_648`. The negative correction clock disables the EMA bias correction (`next_age > 0` becomes false for the affected code path), abruptly changes the protection gate, and can produce a NaN parameter under the `paper_global` equation.

confirmed directly:
```text
step_before: 2147483647
step_after: -2147483648
age_before: 2147483647
age_after: -2147483648
corrected_utility: 0.0
gate: 0.5
parameter_after: 2.0250000953674316
expected_saturated_gate: 0.7310585975646973
```

reachability: `CanonicalUPGD` is exported from both `alberta_framework` and `alberta_framework.core`. `alberta_framework/benchmarks/upgd_ipmnist.py` constructs it with `official_experiment_global`; callers can also instantiate the public class directly. The failure requires a very long-lived or restored-at-limit optimizer state, but is reachable with real finite training gradients — not a hardcoded-safe-only path.

fix incoming: a JAX-compatible saturating int32 increment (`value + jnp.minimum(increment, maximum - value)`) applied to both the global step clock and per-element utility-age clocks. Checked against the module's cited Elsayed & Mahmood ICLR 2024 equation and its pinned official-implementation profile comments — saturating preserves the EMA bias-correction limit exactly, since for any configured `utility_decay < 1` the correction is already numerically 1 at this lifetime, so an unboundedly-increasing integer clock can't change the intended equation past that point anyway.

Svector-anu:
Closes #334.

## What changed

`CanonicalUPGD.update` used ordinary signed-int32 increments for its global
clock and safe-profile utility ages. At `INT32_MAX`, the next call wrapped the
clock negative, disabled bias correction, and changed the protection gate.

The original proposal saturated the clock and continued updating the EMA. That
is not equation-preserving for every allowed decay: with float64 parameters and
`utility_decay=0.999999999`, a reachable post-cap state reports corrected
utility `1.1167776491161645` for a constant signal whose correct value is `1`.

This revision therefore uses a fail-closed lifetime contract:

- the final representable event (`INT32_MAX - 1 -> INT32_MAX`) is consumed;
- later calls are atomic no-ops: parameters, optimizer state, and PRNG key are
  bit-exact and computed payloads are neutral;
- `metrics["update_applied"]` reports the rejection explicitly; and
- both global and per-element counters use a JAX-compatible saturating
  increment, so no intermediate/output counter wraps.

The capacity behavior is covered eagerly and under `jax.jit` for all six
profiles. A float64 near-one-decay regression proves the final representable
event still applies the exact bias correction before capacity is reached.

## Red / green evidence

With the repaired tests applied to the prior saturating-but-continuing source:

```text
4 failed, 38 deselected

- both profiles changed parameters after capacity instead of rejecting;
- the result did not expose update_applied.
```

Final exact-head gates:

```text
126 passed  # canonical UPGD, adaptive UPGD, and IPMNIST UPGD integration
ruff check: passed
mypy alberta_framework/core/canonical_upgd.py: passed
git diff --check: passed
```

No `outputs/` artifact was changed. `canonical_upgd.py` is not in the current
five-claim registered source inventory; historical development provenance is
left untouched.

## Evidence head

`1d85f719eca385c99c4d4eb040ae666cfad98cfe`

### DIFF
--- a/alberta_framework/core/canonical_upgd.py
+++ b/alberta_framework/core/canonical_upgd.py
@@ -96,6 +96,7 @@ _RAW_GLOBAL_PROFILES = frozenset(
         "official_experiment_global",
     }
 )
+_INT32_MAX = 2_147_483_647
 
 
 def _static_zero_scale(scale: float, value: Array) -> Array:
@@ -103,6 +104,14 @@ def _static_zero_scale(scale: float, value: Array) -> Array:
     if scale == 0.0:
         return jnp.zeros_like(value)
     return scale * value
+
+
+def _saturating_increment(value: Array, increment: Array | int = 1) -> Array:
+    """Increment a non-negative int32 counter without lifetime wraparound."""
+
+    maximum = jnp.asarray(_INT32_MAX, dtype=jnp.int32)
+    increment_array = jnp.asarray(increment, dtype=jnp.int32)
+    return value + jnp.minimum(increment_array, maximum - value)
 
 
 @dataclass(frozen=True)
@@ -399,7 +408,7 @@ class CanonicalUPGD:
         next_key = split_keys[0]
         noise_keys = split_keys[1:]
         beta = self._config.utility_decay
-        next_step = state.step + jnp.array(1, dtype=jnp.int32)
+        next_step = _saturating_increment(state.step)
 
         corrected_leaves: list[Array] = []
         new_utility_leaves: list[Array] = []
@@ -440,7 +449,7 @@ class CanonicalUPGD:
                 next_age = jnp.full_like(age, next_step)
                 correction_clock = next_step.astype(param.dtype)
             else:
-                next_age = age + active.astype(jnp.int32)
+                next_age = _saturating_increment(age, active.astype(jnp.int32))
                 correction_clock = next_age.astype(param.dtype)
             bias_correction = 1.0 - jnp.power(beta, correction_clock)
             corrected = jnp.where(
--- a/alberta_framework/core/canonical_upgd.py
+++ b/alberta_framework/core/canonical_upgd.py
@@ -376,6 +376,13 @@ class CanonicalUPGD:
         ``noise`` may supply the already-scaled perturbation PyTree ``xi`` for
         equation-level parity tests.  Normal operation omits it and samples
         ``N(0, noise_std**2)`` from ``key``.
+
+        The final representable int32 clock event is applied.  Once the clock
+        reaches ``INT32_MAX``, later calls reject atomically: parameters,
+        optimizer state, and the key remain exact; computed payloads are
+        neutral; and ``metrics["update_applied"]`` is false.  This avoids both
+        counter wraparound and silently continuing an EMA with a frozen bias-
+        correction clock.
         """
 
         param_leaves, structure = _flatten_with_none(params)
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking: the saturating integer prevents wraparound, but it freezes the EMA bias-correction clock while `utility_ema` continues to advance. The config accepts every finite `utility_decay < 1`, and `CanonicalUPGD` accepts float64 parameters when JAX x64 is enabled, so the claim that correction is already numerically 1 for every allowed decay at `INT32_MAX` is false.

A constant signed utility of 1 has `utility_ema = 1 - beta**t`. With `M = 2**31 - 1` and `beta = 0.999999999`, the patched state after another `M` valid post-cap updates has `step == M` and `utility_ema == 1 - beta**(2*M)`. (The parameters can remain fixed by supplying perturbation `-gradient`.) One more update divides that EMA by the frozen `1 - beta**M` denominator and produces:

```text
JAX_ENABLE_X64=true
beta                         0.999999999
encoded step                 2147483647
true update count            4294967295
corrected_utility observed    1.1167776491161645
corrected_utility expected    1.0
```

Minimal synthesis of that reachable state:

```python
M = 2**31 - 1
beta = 0.999999999
ema = 1.0 - math.exp((2 * M) * math.log(beta))
params = {"w": jnp.array([1.0], dtype=jnp.float64)}
opt = CanonicalUPGD(CanonicalUPGDConfig(
    step_size=0.1, utility_decay=beta, noise_std=0.0,
    profile="paper_global", normalization="global",
))
state = opt.init(params).replace(
    step=jnp.asarray(M, dtype=jnp.int32),
    utility_age={"w": jnp.asarray([M], dtype=jnp.int32)},
    utility_ema={"w": jnp.asarray([ema], dtype=jnp.float64)},
)
result = opt.update(
    state, params, {"w": -jnp.ones(1, dtype=jnp.float64)}, jr.key(0),
    noise={"w": jnp.ones(1, dtype=jnp.float64)},
)
assert float(result.corrected_utility["w"][0]) == 1.0  # gets 1.116777649...
```

Please either carry the bias-correction progress independently after the diagnostic integer saturates (for example, a recursively updated correction factor in state), or fail closed at exhausted capacity. Add a near-one-decay/x64 regression that represents continued operation beyond the cap. Also reuse/move the existing module `_INT32_MAX` declaration rather than defining it twice.

I composed this exact head onto current `origin/main` (`107375e`) and the ordinary focused/downstream suites, Ruff, and mypy pass; this is a semantic boundary failure, not a merge/test failure. I did not authorize the fork CI run for this head.

lalalune (changes requested):
Blocking exact-head finding at 2ae7e9b17d75fde72926f15a8d33a4aa3f91382d: saturating the global/age clocks while continuing to update the EMA makes the bias-correction denominator freeze while its numerator keeps evolving. With x64 enabled, beta=0.999999999, and a reachable constant-utility state after 2*INT32_MAX updates, both paper_global and safe_extended report corrected utility 1.1167776491161645 instead of ~1.0. The counter no longer wraps, but long-life correction is mathematically wrong after saturation. A retained regression should drive a repair that keeps the EMA and correction clock semantically aligned at the ceiling.

lalalune (approved):
The corrected exact head rejects counter-capacity atomically instead of continuing a mathematically stale bias-correction clock. Red-first capacity tests failed 4 cases on the prior source; final eager/JIT coverage spans all six profiles, the final representable event is proved under float64 near-one decay, and 126 affected/downstream tests plus Ruff/mypy pass.

PR #336, opened 2026-08-16
fix(world-model-ensemble): refuse initial states the module would reject

### LINKED ISSUE #333: RecurrentLatentWorldModelEnsemble.init() can return a state its own state_valid() rejects for an accepted config (silent permanent no-op)
## Summary

`RecurrentLatentWorldModelEnsembleConfig.__post_init__` (`alberta_framework/core/recurrent_latent_world_model_ensemble.py`) validates that *one clipped update* fits inside `max_parameter_magnitude` (`learning_rate * gradient_clip_norm <= max_parameter_magnitude`) but never that the **initial parameters** do. `_init_member` writes `variance_bias = logit((sqrt(floor*max) - floor)/(max - floor))` ≈ `-5.76` at the default variance bounds and draws kernels at `initialization_scale / sqrt(fan_in)`. `init()` returns that state without checking it, and `_parameters_valid` then rejects it.

Nothing raises: the model constructs, `init()` returns, and `start`/`decide`/`update` all silently no-op forever (`state_valid=False`, `applied=False`, `event_count=0`) — a complete, well-formed run of zeros. A sweep over `max_parameter_magnitude` or `initialization_scale` silently contributes dead arms.

```python
import jax, jax.numpy as jnp, jax.random as jr
from alberta_framework.core.recurrent_latent_world_model_ensemble import (
    RecurrentLatentWorldModelEnsemble, RecurrentLatentWorldModelEnsembleConfig)

cfg = RecurrentLatentWorldModelEnsembleConfig(
    observation_dim=2, n_actions=2, latent_dim=3, ensemble_size=3, learning_rate=0.001,
    bootstrap_probability=0.8, uncertainty_warmup_steps=1, max_updates=8,
    max_parameter_magnitude=1.0)          # lr*clip = 0.01 <= 1.0, so the config is accepted
model = RecurrentLatentWorldModelEnsemble(cfg)
state = model.init(jr.key(0))
print("max|param|", max(float(jnp.max(jnp.abs(l))) for l in jax.tree.leaves(state.member_parameters)))
print("state_valid(init state) =", bool(model.state_valid(state)))
```

Output on `main`:

```
max|param| 5.756462574005127
state_valid(init state) = False
```

Same with `initialization_scale=50, max_parameter_magnitude=6.0`: `max|param| = 90.6`, `state_valid = False`, every subsequent `update` has `applied=False`, `event_count` stays `0`.

Expected: the deterministic initial variance bias is bounded at construction (like the other `max_*` consistency checks), and `init()` refuses to return a state that its own validity predicate rejects.

## Dedup

`gh issue/pr list --search` for "max_parameter_magnitude", "initialization_scale", "world model ensemble", "recurrent latent": only #139/#140 (checkpoint `_format_version`, unrelated). No open PR touches this module.

Prabhat1308:
Closes #333.

## Issue

An accepted `RecurrentLatentWorldModelEnsembleConfig` could return an initial state that its own `state_valid()` rejects, silently pinning later transactions at event zero. A second audit found that a valid state drawn with one key could be saved but not restored when the unrelated default checkpoint-template draw exceeded the bound.

## New state

- Construction rejects a parameter bound below the actual stored float32 variance-head bias, including exact endpoint and next-neighbour behavior.
- `init()` is an explicit host-side entry point and raises instead of returning a rejected state.
- Checkpoint restore uses its local random draw only for the static PyTree contract, then validates the reconstructed persisted state and its resource budget. Valid restore no longer depends on unrelated template values; metadata/resource tampering still fails closed.
- No schema, accepted transition arithmetic, evidence artifact, or benchmark output changed.

## Verification

At repaired head `c194168fb1834aaa66ec8c99c659c2eb50322a25`:

```text
5 adversarial init/float32/checkpoint/tamper tests under RuntimeWarning-as-error -> passed
Ruff (changed source + test)                                                    -> passed
strict mypy (changed source)                                                    -> passed
```

The complete affected module suite passed 25/25 on the immediately prior repair tree; the final rebase added only a disjoint mainline micro-curve merge. Fresh CI is authoritative for this exact head. An enlarged adjacent batch reached an unrelated PrototypeAgent compilation lane without a failure before it was intentionally bounded; it is not reported as a completed gate.

`core/recurrent_latent_world_model_ensemble.py` is not a registered evidence source. No benchmark lane, seed schedule, or `outputs/` artifact was touched.

- [x] logs: inline above

### DIFF
--- a/alberta_framework/core/recurrent_latent_world_model_ensemble.py
+++ b/alberta_framework/core/recurrent_latent_world_model_ensemble.py
@@ -1552,18 +1552,20 @@ def load_recurrent_latent_world_model_ensemble_checkpoint(
     if metadata.get("scientific_promotion_allowed") is not False:
         raise ValueError("checkpoint cannot claim scientific promotion")
     key = jr.key(0) if template_key is None else template_key
-    template = model.init(key)
-    expected_budget = model.resource_budget(template).to_config()
-    if metadata.get("resource_budget") != expected_budget:
-        raise ValueError("checkpoint resource budget does not match config")
+    # The template supplies only the checkpoint's static PyTree contract.  It
+    # must not make restoration depend on whether this unrelated random draw
+    # happens to fit the configured parameter bound; the persisted state is
+    # validated immediately after it is reconstructed.
+    template = model._initial_state(key)
     restored, second_metadata = load_checkpoint(template, path)
     if second_metadata != metadata:
         raise ValueError("checkpoint metadata changed between reads")
     state = cast(RecurrentLatentWorldModelEnsembleState, restored)
     if not bool(jax.device_get(model.state_valid(state))):
         raise ValueError("restored recurrent latent ensemble state is invalid")
-    if model.resource_budget(state).to_config() != expected_budget:
-        raise ValueError("restored recurrent latent ensemble resource budget is invalid")
+    expected_budget = model.resource_budget(state).to_config()
+    if metadata.get("resource_budget") != expected_budget:
+        raise ValueError("checkpoint resource budget does not match config and state")
     return model, state
 
 
--- a/alberta_framework/core/recurrent_latent_world_model_ensemble.py
+++ b/alberta_framework/core/recurrent_latent_world_model_ensemble.py
@@ -304,6 +304,12 @@ class RecurrentLatentWorldModelEnsembleConfig:
             raise ValueError("gradient_clip_norm cannot exceed max_raw_gradient_norm")
         if self.learning_rate * self.gradient_clip_norm > self.max_parameter_magnitude:
             raise ValueError("one clipped update exceeds max_parameter_magnitude")
+        initial_variance_bias = abs(self.initial_variance_logit)
+        if initial_variance_bias > self.max_parameter_magnitude:
+            raise ValueError(
+                f"initial variance bias magnitude {initial_variance_bias} exceeds "
+                f"max_parameter_magnitude={self.max_parameter_magnitude}"
+            )
         if self.state_nbytes > _MAX_STATE_NBYTES:
             raise ValueError(
                 f"configured persistent state needs {self.state_nbytes} bytes; "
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking float32-boundary regression at exact head db9b2386. The new constructor compares abs(initial_variance_logit) in binary64 (5.7564627087363895 for the defaults), but _init_member stores that bias as float32 (5.756462574005127). Because max_parameter_magnitude is itself canonicalized to float32, a bound exactly equal to the actual stored parameter is valid under state_valid yet the config now rejects it. Reproduction: construct the default config, take sink=float(np.float32(abs(cfg.initial_variance_logit))), then reconstruct with max_parameter_magnitude=sink; this head raises, while the next float32 bound up constructs and init is valid. Compare/canonicalize the deterministic initializer in its actual float32 execution dtype and add exact endpoint/next-neighbor tests. The new host bool(state_valid) also makes jax.jit(model.init) fail where it previously traced; please either document init as deliberately host-only or retain prior JIT compatibility.

Prabhat1308:
Addressed in `9680612`:

- the initializer bound is now compared in the parameters' float32 dtype (`initial_variance_bias_magnitude = float(np.float32(abs(initial_variance_logit)))`), so `max_parameter_magnitude == sink` constructs and `init` is valid, while the next float32 below is rejected — pinned by `test_initial_variance_bias_bound_is_exact_at_its_float32_endpoint` (endpoint accepted + stored bias equals the endpoint + `nextafter` below rejected);
- `init` is documented as a deliberately host-side entry point (it validates the drawn parameters eagerly and raises; compile `start`/`decide`/`update` instead), pinned by `test_init_is_a_host_side_entry_point` (`jax.jit(model.init)` raises `TracerBoolConversionError`).

Body updated to evidence head `9680612`; 57 passed across the two ensemble suites, ruff/mypy clean. Re-review requested.

lalalune (approved):
Independent exact-head review of 9680612da4a76f6a47217ec69cfe24833d022552: READY. The validity bound is now compared to the actual stored float32 variance bias, with exact endpoint/below-endpoint coverage; accepted initialization is host-validated before use and the host-only init contract is explicit. Verified 36 full affected/adjacent tests under warnings-as-errors, Ruff, strict mypy, diff hygiene, and a clean current-main merge-tree. No central registered evidence source or artifact changes. Merge only after this exact head has 20/20 CI.

lalalune (approved):
Exact-head review of c194168fb1834aaa66ec8c99c659c2eb50322a25: READY. The initializer now rejects invalid drawn states at the host boundary, the deterministic bias check matches its stored float32 endpoint, and checkpoint restore no longer validates an unrelated template draw before reconstructing the persisted state. The adversarial restore test failed before the repair and passes now; tampered resource metadata still fails closed. Verified the exact repair tests under RuntimeWarning-as-error, Ruff, strict mypy, diff hygiene, mergeability, and no outputs changes. Fresh exact-head CI is queued.

lalalune (approved):
Independent refresh at exact head c194168: READY. The added restore path now uses _initial_state only as a static PyTree template, validates the reconstructed persisted state immediately, and derives the deterministic resource record from that valid state; restore no longer depends on whether unrelated seed 0 happens to satisfy a tight initialization bound. The original init/variance-bound repair remains intact. Exact head: 59 recurrent-ensemble/checkpoint/package tests passed, Ruff and strict mypy passed, diff-check and current-main merge-tree are clean. No registered evidence source or output artifact changes.

PR #342, opened 2026-08-16
fix(step3): bound dimensions and surface refused smoke updates

### LINKED ISSUE #339: Step 3 Horde facade double-rounds exact Rational inputs and accepts float32 overflows
## Problem

`Step3HordeConfig` uses `_require_real` with standard `float(value)`, which double-rounds exact `Fraction` inputs (e.g. `Fraction(1, 1) + Fraction(1, 2**24) + Fraction(1, 2**60)` midpoints) before float32 JAX sinks. It also accepts out-of-range float32 overflow values (such as `step_size=1e100` and `obgd_kappa=1e100`) and float32 underflow on strictly positive fields (`obgd_kappa=1e-50`), and `run_step3_smoke` does not verify that all updates were applied.

## Reproduction

```python
import numpy as np
from alberta_framework.steps import Step3HordeConfig

cfg = Step3HordeConfig(step_size=midpoint)
# Current main stores 1.0 instead of nextafter(float32(1.0), float32(2.0))
```

## Proposed Fix

- Use `round_real_to_float32` and `_narrow_float32` in `alberta_framework/steps/step3.py` to match the exact-ratio primitive in `alberta_framework/_float32.py`.
- Validate `step_size`, `obgd_kappa`, `sparsity`, `gammas`, `lamdas` against float32 overflow and require `obgd_kappa` to remain strictly positive in float32.
- Bound `hidden_sizes` to signed int32 max (`2**31 - 1`).
- Check `updates_applied` in `run_step3_smoke`.
- Add comprehensive unit tests in `tests/test_step3_production.py`.

deepanshu-yd:
Closes #339.

## Reconciled scope

PR #306 has already landed the shared exact float32 conversion and domain contracts on `main`. This branch was rebased onto that merge and now retains only the still-missing Step 3 hardening:

- reject `hidden_sizes` above signed-int32 maximum before they reach JAX shape sinks;
- require every smoke step and every requested demon update to have applied before `Step3SmokeResult.finite` can be true.

The per-demon health check is stricter than the original branch, which checked only the aggregate per-step flag and could miss a partial independent-Horde refusal. The obsolete duplicate float32 converter, serialization rewrites, tuple restrictions, and unrelated enum/bool validation were removed.

## Scope

- `alberta_framework/steps/step3.py`
- `tests/test_step3_production.py`

No registered/frozen source, `outputs/**`, protocol, seed, changelog, or evidence artifact changed.

## Exact-head verification

- Step 3 production + shared float32 contract: 102 passed;
- focused Ruff: passed;
- strict mypy for Step 3: passed;
- `git diff --check`: clean.

## Scientific integrity

This is facade and development-smoke validation, not scientific evidence or a performance claim.

## Contribution provenance

- Initial contribution: AI-assisted, google / gemini-3.7-flash, Antigravity CLI (self-reported)
- Maintainer reconciliation: AI-assisted, OpenAI Codex

### DIFF
--- a/alberta_framework/steps/step3.py
+++ b/alberta_framework/steps/step3.py
@@ -46,6 +46,7 @@ from alberta_framework.steps._float32_validation import (
 
 Step3NormalizerName = Literal["none", "ema"]
 Step3TraceModeName = Literal["accumulating", "replacing"]
+_INT32_MAX = 2**31 - 1
 _FLOAT32_MIN_NORMAL = float.fromhex("0x1.0p-126")
 Step3RoutingName = Literal["shared", "independent", "mixed"]
 
@@ -212,6 +213,8 @@ def _require_positive_int(name: str, value: object) -> int:
     number = int(value)
     if number < 1:
         raise ValueError(f"{name} must be positive, got {value!r}")
+    if number > _INT32_MAX:
+        raise ValueError(f"{name} must be at most int32 max, got {value!r}")
     return number
 
 
@@ -531,6 +534,8 @@ def run_step3_smoke(
     finite = bool(
         jnp.all(jnp.isfinite(result.per_demon_metrics))
         & jnp.all(jnp.isfinite(result.td_errors))
+        & jnp.all(result.updates_applied)
+        & jnp.all(result.head_updates_applied)
     )
     return Step3SmokeResult(
         config=cfg,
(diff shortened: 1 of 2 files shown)


lalalune (approved):
After the #306 reconciliation this head retains only the real Step 3 residual: int32 hidden-size bounds and per-step/per-demon smoke health. Exact focused panel: 102 passed; Ruff, mypy, and diff checks clean.

PR #346, opened 2026-08-16
fix(experiments): require one configuration and unique seeds in aggregate_metrics

Prabhat1308:
Closes #345.

## Issue

`aggregate_metrics` took the first run's `config_name` and the raw seed list, so runs from two arms (or the same seed twice) were averaged into one `AggregatedResults` labelled with the first arm and `n_seeds` counting the duplicates: `baseline [0, 1, 2, 0, 1, 2] n_seeds=6 mean=5.0` for a 1.0 baseline and a 9.0 treatment.

## New state

`aggregate_metrics` refuses mixed configuration names (`aggregate_metrics requires runs from one configuration; got ['candidate', 'treatment']`) and repeated seed identities (`… requires unique seed identities; got [0, 1, 0]`) before building the artifact — the same invariants `run_multi_seed_experiment` already enforces one layer up (#25/#27). Well-formed inputs are unchanged.

Failing-test-first: `test_aggregate_metrics_rejects_runs_from_different_configs` and `test_aggregate_metrics_rejects_duplicate_seed_identities` fail on `main` and pass here.

## Evidence

Falsifier from #345 at this head:

```text
ValueError: aggregate_metrics requires runs from one configuration; got ['baseline', 'treatment']
```

Verification at `324798112f725763434ae521b2d30314d5cd5e57`:

```text
.venv/bin/python -m pytest tests/test_experiments_validation.py tests/test_statistics_validation.py tests/test_utils_smoke.py -q -o addopts=""   -> 152 passed
.venv/bin/python -m ruff check .                                                                                            -> All checks passed!
.venv/bin/python -m mypy                                                                                                    -> Success: no issues found in 164 source files
```

`utils/experiments.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, ruff, mypy, falsifier output at the evidence head)

Provider/model/client: anthropic / claude-fable-5 / Claude Code 2.1.233 (contribute-to-asi skill revision 72e4239e).

PR #358, opened 2026-08-16
fix(experiments): refuse non-injective hyperparameter extractors

### LINKED ISSUE #357: extract_hyperparameter_results silently drops configurations whose extracted parameter value collides
## Summary

`extract_hyperparameter_results` (`alberta_framework/utils/experiments.py`) is documented as "Dictionary mapping param value to (mean, std) tuple. Useful for creating hyperparameter sensitivity plots." It returns `{param_extractor(name): perf for name, perf in performance.items()}` — a dict comprehension over a non-injective extractor keeps only the last insertion, with no check and no error. A 2-D sweep (e.g. `lr × decay`) whose extractor pulls only `lr` silently discards all but the dict-last configuration per value, and the surviving point is chosen by insertion order, not by any criterion. The sensitivity curve is built from a silently truncated, order-dependent subset.

## Falsifier (deterministic, pristine `main` @ `47dc21c`)

```python
import numpy as np
from alberta_framework.utils.experiments import AggregatedResults, extract_hyperparameter_results
def flat(name, v): return AggregatedResults(name, [0, 1], {"squared_error": np.full((2, 3), v)}, {})
results = {"lr_0.01_decay_0.0": flat("a", 1.0), "lr_0.01_decay_0.1": flat("b", 9.0), "lr_0.10_decay_0.0": flat("c", 3.0)}
print(extract_hyperparameter_results(results, param_extractor=lambda n: float(n.split("_")[1])))
```

Output on `main`:

```
{0.01: (9.0, 0.0), 0.1: (3.0, 0.0)}
```

Three configurations in, two points out; the `lr=0.01` point is the `decay=0.1` run only because it was inserted last.

Expected: `ValueError` naming the colliding configurations (`param_extractor maps several configurations to one value: 0.01 <- ['lr_0.01_decay_0.0', 'lr_0.01_decay_0.1']`); injective extractors unchanged.

## Dedup

`gh issue/pr list --search "extract_hyperparameter_results"` / `"param_extractor"`: only #111 (`get_final_performance(window<=0)`, closed) surfaces via the file. Open #345/#346 (mine) touches `aggregate_metrics` in the same module and test file; the fix PR stacks on it.

Fix PR follows (failing-test-first, stacked on #346).

Prabhat1308:
Closes #357. Stacked on #346 (`fix/aggregate-metrics-identity`): the first commit here is #346's; only `083f70f` is new. Rebases to a one-commit diff once #346 lands.

## Issue

`extract_hyperparameter_results` built its result with a dict comprehension, so an extractor mapping several configurations to one value silently kept whichever was inserted last: three configurations in, two points out, the `lr=0.01` point being the `decay=0.1` run only by insertion order.

## New state

Configurations are grouped by extracted value first; any value with more than one configuration raises `param_extractor maps several configurations to one value: 0.01 <- ['lr_0.01_decay_0.0', 'lr_0.01_decay_0.1']`. Injective extractors and the `param_extractor=None` path are unchanged.

Failing-test-first: `test_extract_hyperparameter_results_rejects_colliding_parameter_values` fails on `main` (returns the truncated dict) and passes here; `test_extract_hyperparameter_results_keeps_injective_extractors` pins the accepted path.

## Evidence

Falsifier from #357 at this head:

```text
ValueError: param_extractor maps several configurations to one value: 0.01 <- ['lr_0.01_decay_0.0', 'lr_0.01_decay_0.1']
```

Verification at `083f70f2387e77c48c0785b26bfa48f42720284c`:

```text
.venv/bin/python -m pytest tests/test_experiments_validation.py tests/test_utils_smoke.py -q -o addopts=""   -> 40 passed
.venv/bin/python -m ruff check .                                                                       -> All checks passed!
.venv/bin/python -m mypy                                                                               -> Success: no issues found in 164 source files
```

`utils/experiments.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

### DIFF
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -419,10 +419,24 @@ def extract_hyperparameter_results(
 
     Returns:
         Dictionary mapping param value to (mean, std) tuple
+
+    Raises:
+        ValueError: If ``param_extractor`` maps more than one configuration to
+            the same value; a sensitivity curve built from a silently
+            truncated, insertion-order-dependent subset is not a measurement.
     """
     performance = get_final_performance(results, metric)
 
     if param_extractor is None:
         return {k: v for k, v in performance.items()}
 
-    return {param_extractor(name): perf for name, perf in performance.items()}
+    names_by_value: dict[Any, list[str]] = {}
+    for name in performance:
+        names_by_value.setdefault(param_extractor(name), []).append(name)
+    collisions = {value: names for value, names in names_by_value.items() if len(names) > 1}
+    if collisions:
+        described = "; ".join(f"{value!r} <- {names}" for value, names in collisions.items())
+        raise ValueError(
+            f"param_extractor maps several configurations to one value: {described}"
+        )
+    return {value: performance[names[0]] for value, names in names_by_value.items()}
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -150,13 +150,22 @@ def aggregate_metrics(results: list[SingleRunResult]) -> AggregatedResults:
         AggregatedResults with aggregated metrics
 
     Raises:
-        ValueError: If ``results`` is empty, or any metric sample is non-finite.
+        ValueError: If ``results`` is empty, mixes configuration names or seed
+            identities, drifts in metric keys, or any metric sample is
+            non-finite.
     """
     if not results:
         raise ValueError("Cannot aggregate empty results list")
 
-    config_name = results[0].config_name
+    config_names = sorted({r.config_name for r in results})
+    if len(config_names) != 1:
+        raise ValueError(
+            f"aggregate_metrics requires runs from one configuration; got {config_names}"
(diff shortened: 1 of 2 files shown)


lalalune (approved):
Independent exact-head review: READY. Hyperparameter extraction now groups by extracted key and refuses any non-injective mapping before constructing an insertion-order-dependent truncated curve; injective/no-extractor behavior is preserved. Verified exact diff and collision/success cases in a 404-test affected/adjacent stack, with Ruff and strict mypy clean. This is stacked on #346; merge #346 first and require 20/20 CI on this exact head.

lalalune (approved):
Exact patch now sits on merged #346. Combined experiments/statistics/micro/dual-replay coverage passed 379 tests with Ruff and strict mypy clean. Non-injective hyperparameter extraction now rejects deterministically instead of silently retaining the dict-last configuration.

PR #360, opened 2026-08-16
fix(step8): reject illegal world model facade config

### LINKED ISSUE #359: Step 8 World Model facade double-rounds exact Rational inputs and accepts float32 overflows
## Problem

`Step8WorldModelConfig` converts scientific scalars via standard `float(value)` in `_require_real`, which double-rounds exact `Fraction` inputs before float32 JAX sinks. It also accepts out-of-range float32 overflow values (`step_size=1e100`, `sparsity=1e100`, `leaky_relu_slope=1e100`, `utility_decay=1e100`) and lacks int32 upper-bound guards on `observation_dim`, `n_actions`, `action_dim`, and `hidden_sizes`.

## Reproduction

```python
import numpy as np
from alberta_framework.steps.step8 import Step8WorldModelConfig

cfg = Step8WorldModelConfig(step_size=midpoint)
# Current main stores 1.0 instead of nextafter(float32(1.0), float32(2.0))
```

## Proposed Fix

- Use `round_real_to_float32` and `_narrow_float32` in `alberta_framework/steps/step8.py` matching the shared exact-ratio primitive in `alberta_framework/_float32.py`.
- Validate all float32 execution fields against overflow.
- Bound `observation_dim`, `n_actions`, `action_dim`, and `hidden_sizes` to signed int32 max (`2**31 - 1`).
- Add comprehensive test suite in `tests/test_step8_production.py`.

deepanshu-yd:
Closes #359.

## What changed

`Step8WorldModelConfig` now validates every scientific scalar and dimension before constructing one-step world models. Float32 execution values use the shared exact-ratio `round_real_to_float32` helper, preventing binary64 double rounding for `Fraction` and extended-precision inputs.

Exact float32 overflow is rejected across `step_size`, `sparsity`, `leaky_relu_slope`, `utility_decay`; legal zero/one endpoints remain valid. Observation dimension, discrete action count, action dimension, and hidden layer dimensions are bounded to signed int32 max (`2**31 - 1`).

## Scope

• `alberta_framework/steps/step8.py`
• `tests/test_step8_production.py`

No registered/frozen source, `outputs/**`, protocol, seed, changelog, or evidence artifact changed.

## Exact-head verification

• Step 8 production suite: 64 passed;
• affected world model suite: 5 passed;
• full Ruff: passed;
• strict mypy: no issues in 164 source files;
• `git diff --check`: clean.

## Scientific integrity

This is facade input validation, not scientific evidence or a performance claim.

## Contribution provenance

• AI assistance: yes
• AI provider/model: google / gemini-3.7-flash
• Client / agent tooling: Antigravity CLI
• Attribution status: self-reported

### DIFF
--- a/alberta_framework/steps/step8.py
+++ b/alberta_framework/steps/step8.py
@@ -70,7 +70,15 @@ class Step8WorldModelConfig:
     def to_dict(self) -> dict[str, object]:
         """Return a JSON-serializable representation."""
         payload = asdict(self)
-        payload["hidden_sizes"] = list(self.hidden_sizes)
+        payload["hidden_sizes"] = [int(h) for h in self.hidden_sizes]
+        payload["observation_dim"] = int(self.observation_dim)
+        if self.n_actions is not None:
+            payload["n_actions"] = int(self.n_actions)
+        payload["action_dim"] = int(self.action_dim)
+        payload["step_size"] = float(self.step_size)
+        payload["sparsity"] = float(self.sparsity)
+        payload["leaky_relu_slope"] = float(self.leaky_relu_slope)
+        payload["utility_decay"] = float(self.utility_decay)
         return payload
 
     @classmethod
@@ -96,6 +104,16 @@ class Step8WorldModelConfig:
             predict_delta=self.predict_delta,
             utility_decay=self.utility_decay,
         )
+
+
+_INT32_MAX = 2**31 - 1
+
+
+def _require_nonnegative_real(name: str, value: object) -> float:
+    real, numerator, _, narrowed = finite_real_and_float32(name, value)
+    if real < 0.0 or numerator < 0 or narrowed < 0.0:
+        raise ValueError(f"{name} must be non-negative, got {value!r}")
+    return canonical_float32_storage(real, narrowed)
 
 
 def _require_unit_interval(name: str, value: object) -> float:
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking at exact `c339246e6af88b5e3ff9a0648f92d124662565c1`: the current-base scientific-real validator accepts non-Real `__class__ -> float` spoof objects, so merge/rebase repaired PR #390 first. The local `hidden_sizes` container check also uses spoofable `isinstance(value, tuple)`: a non-tuple object reporting `__class__ -> tuple` is accepted, contrary this PR's strict class-spoof/fail-closed contract (even though elements are later canonicalized). Require actual tuple runtime type and add a container-spoof regression. Direct Fraction/overflow/underflow/int32/factory/smoke probes otherwise passed; CI was nearly green but does not cover these cases.

deepanshu-yd:
Updated at 8423eb4:
- Rebased cleanly on latest `main` using shared `_float32_validation.py` exact-ratio validators.
- Enforced strict tuple container type validation on `hidden_sizes` (`type(hidden_sizes) is tuple`), rejecting spoofed class containers.
- Enforced strict boolean validation on `use_layer_norm` and `predict_delta`.
- Added unit regressions in `tests/test_step8_production.py` for spoofed tuple container, non-bool flags, and class-spoofed floats / adversarial exact ratios.

PR #341, opened 2026-08-16
fix(micro-continual): validate per-regime curves as exact reals in their metric domain

Prabhat1308:
Closes #340.

## Issue

`load_micro_shard` coerced `per_regime_accuracy` / `per_regime_loss` / `per_regime_plasticity` through `np.asarray(dtype=float64)` and checked only shape and finiteness, so numeric strings, booleans, and out-of-domain values reached the ranked summary; a shard at `5.0` accuracy topped a merge whose analytic Bayes ceiling was `0.80`.

## New state

`_validated_curve` requires each curve to be a list of exactly `n_regimes` exact finite reals (bool/str rejected via the existing `_require_finite_real`) inside its metric's domain — accuracy and plasticity in `[0, 1]`, loss `>= 0` — and canonicalizes entries to `float`. Valid shards written by `micro_shard_payload` are unchanged (their entries are already floats in range), so merges and transfer validation reproduce byte-for-byte apart from `created_unix`.

Failing-test-first: `test_load_rejects_curves_outside_their_measured_domain` (8 cases) fails on `main` and passes here; `test_load_canonicalizes_integer_curve_entries_to_floats` pins the accepted path.

## Evidence

Falsifier from #340 at this head:

```text
ValueError: …/numeric_strings.json: per_regime_accuracy[0] must be a finite real number, got '0.9'
```

Verification at `8d190cbbf5d52f10ec9cee1747ab0ada9e852c56`:

```text
.venv/bin/python -m pytest tests/test_micro_continual.py -q -o addopts=""   -> 169 passed
.venv/bin/python -m ruff check .                                             -> All checks passed!
.venv/bin/python -m mypy                                                     -> Success: no issues found in 164 source files
```

Composes cleanly with #322 (`git merge-tree` clean). `benchmarks/micro_continual.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, ruff, mypy, falsifier output at the evidence head)

Provider/model/client: anthropic / claude-fable-5 / Claude Code 2.1.233 (contribute-to-asi skill revision 72e4239e).

PR #362, opened 2026-08-16
fix(micro-continual): record the spec that actually ran in each shard

### LINKED ISSUE #361: micro_shard_payload records the registry's hyperparameters, not the spec that ran, so cross-seed drift can never be detected
## Summary

`run_micro_arm` (`alberta_framework/benchmarks/micro_continual.py`) explicitly accepts `arm: str | MicroArmSpec` — the discovery-lane fitness path documented in `outputs/micro_continual/SUITE.md` §6 — and runs whatever spec it is given. `micro_shard_payload` then throws that spec away: it does `spec = micro_arm_spec(result.arm_name)` and serializes the **registry** entry's `mechanism` and `hyperparameters`. `MicroRunResult` has no field for either, so the writer literally cannot record what ran. Consequences:

- A run started from a custom `MicroArmSpec` that shares a registry name is persisted with hyperparameters that were never executed, and a custom spec with a new name cannot be serialized at all (`KeyError`).
- The cross-seed drift validator `_validate_micro_arm_contract` (#106 / PR #109) is vacuous for every shard this module writes: the drift it detects can only come from hand-edited JSON (which is exactly how `test_merge_rejects_arm_contract_drift` builds its fixture). A sweep that runs seed 0 at `step_size=0.01` and seed 1 at `step_size=0.5` merges cleanly and publishes one averaged `sgd_raw` number with false provenance.
- Related aliasing: `micro_arm_spec(name).hyperparameters` **is** the live registry dict; a caller that pokes it changes the champion arm process-wide.

## Falsifier (deterministic, pristine `main` @ `45279f3`)

```python
import dataclasses, tempfile
from pathlib import Path
from alberta_framework.benchmarks.micro_continual import (
    MicroStreamConfig, merge_micro_shards, micro_arm_spec, micro_shard_path, micro_shard_payload,
    run_micro_arm, write_micro_shard)
TINY = MicroStreamConfig(family="input_permutation", n_regimes=4, regime_length=25, dim=12,
                         n_classes=3, n_components=2, component_sparsity=4)
custom = dataclasses.replace(micro_arm_spec("sgd_raw"), hyperparameters={"step_size": 0.5, "weight_decay": 0.3})
a = run_micro_arm(TINY, "sgd_raw", seed=0, hidden1=8, hidden2=6)
b = run_micro_arm(TINY, custom, seed=1, hidden1=8, hidden2=6)
print("registry run:", a.overall_accuracy, " custom run:", b.overall_accuracy)
print("custom shard hyperparameters:", micro_shard_payload(b)["hyperparameters"])
out = Path(tempfile.mkdtemp()); paths = []
for r in (a, b):
    p = micro_shard_path(out, TINY.family, "sgd_raw", r.seed); write_micro_shard(p, micro_shard_payload(r)); paths.append(p)
s = merge_micro_shards(paths, bayes_samples=2_000)
print("merged:", s["results"][0]["seeds"], s["results"][0]["hyperparameters"])
micro_arm_spec("sgd_raw").hyperparameters["step_size"] = 0.5
print("registry now says:", micro_arm_spec("sgd_raw").hyperparameters["step_size"])
```

Output on `main`:

```
registry run: 0.31  custom run: 0.33
custom shard hyperparameters: {'step_size': 0.01, 'weight_decay': 0.0}
merged: [0, 1] {'step_size': 0.01, 'weight_decay': 0.0}
registry now says: 0.5
```

Two different update rules averaged into one arm under the registry's hyperparameters, and the registry mutated through a lookup.

Expected: `MicroRunResult` carries `mechanism` and `hyperparameters` from the spec that ran, `micro_shard_payload` serializes those, and registry hyperparameters are read-only.

## Dedup

Adjacent, closed: #106 / PR #109 (the drift validator this defect neuters), #149 (`noise_pool_steps` never persisted — same "provenance not persisted ⇒ check vacuous" shape in screening). Open #340/#341 (mine) touches `load_micro_shard`; merge-tree composes cleanly. No issue or PR covers `MicroRunResult` / `micro_shard_payload` provenance or registry aliasing.

Prabhat1308:
Closes #361.

micro_shard_payload previously re-looked up the registry by arm name, so a custom MicroArmSpec run was serialized with metadata that never ran and the cross-seed drift guard was vacuous. The run result now carries the actual mechanism and hyperparameters, and registry specs copy/freeze their mappings.

Release hardening on the rebased head also makes direct MicroRunResult construction copy and freeze a primitive-only mapping, validates exact non-empty string keys and finite real values on both construction and load, rejects __class__ type spoofing, and makes the supported-name boundary symmetric: the registered-arm loader allowlist is retained and the serializer refuses unregistered result names before writing an unreadable shard. Custom specs remain supported when they intentionally share a registered arm name; their actual recorded contract is what cross-seed merge validation compares.

Failing-first: 13 adversarial direct-construction, invalid-hyperparameter, unknown-name, and loader cases failed on the original #362 implementation before the hardening guard.

Verification at a3a56e89960bed6d6e72cf027dc58c90288621cc:

- 237 micro-continual and rule-discovery tests passed (2 pre-existing sklearn deprecation warnings).
- Ruff passed on changed source/tests.
- Strict mypy passed on micro_continual.py.
- Rebased after merged #341 curve-domain validation.
- outputs/ is untouched; no benchmark or seed lane was executed.

This remains permanently nonpromoting development infrastructure.

### DIFF
--- a/alberta_framework/benchmarks/micro_continual.py
+++ b/alberta_framework/benchmarks/micro_continual.py
@@ -74,7 +74,7 @@ from functools import lru_cache
 from numbers import Real
 from pathlib import Path
 from types import MappingProxyType
-from typing import Any
+from typing import Any, cast
 
 import jax
 import jax.numpy as jnp
@@ -148,12 +148,39 @@ _BAYES_DOMAIN = 404
 
 def _require_finite_real(value: object, name: str) -> float:
     """Return one canonical float after rejecting non-real and non-finite values."""
-    if isinstance(value, bool) or not isinstance(value, Real):
+    if type(value) is bool or not issubclass(type(value), Real):
         raise ValueError(f"{name} must be a finite real number, got {value!r}")
-    number = float(value)
+    try:
+        number = float(cast(Any, value))
+    except (OverflowError, TypeError, ValueError) as exc:
+        raise ValueError(f"{name} must be a finite real number, got {value!r}") from exc
     if not math.isfinite(number):
         raise ValueError(f"{name} must be a finite real number, got {value!r}")
     return number
+
+
+def _freeze_micro_hyperparameters(value: object, *, context: str) -> Mapping[str, float]:
+    """Copy one primitive-only hyperparameter mapping behind an immutable view."""
+    if not issubclass(type(value), Mapping):
+        raise ValueError(f"{context} must be an object with non-empty string keys")
+    try:
+        items = list(cast(Mapping[object, object], value).items())
+    except Exception as exc:
+        raise ValueError(
+            f"{context} must be an object with non-empty string keys"
+        ) from exc
+    frozen: dict[str, float] = {}
+    for key, raw_value in items:
+        if type(key) is not str or not key:
+            raise ValueError(f"{context} keys must be non-empty strings")
+        frozen[key] = _require_finite_real(raw_value, f"{context}[{key!r}]")
+    return MappingProxyType(frozen)
+
+
+def _require_nonempty_string(value: object, *, context: str) -> str:
+    if type(value) is not str or not value:
+        raise ValueError(f"{context} must be a non-empty string")
+    return value
 
 
 # =============================================================================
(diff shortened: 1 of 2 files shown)


lalalune (approved):
Independent review at exact head 30f10b5: READY. The shard now records the resolved spec that actually ran, and strict loading can detect cross-seed drift instead of re-reading registry defaults. Verified in a clean composition with the adjacent micro-validation change: 223 tests passed (only two sklearn deprecation warnings), full Ruff and strict mypy passed. No central registered evidence source or output artifact is changed; this remains nonpromoting development infrastructure. MappingProxyType pickling is a compatibility note only—no current process/persistence consumer accepts a MicroArmSpec object.

PR #371, opened 2026-08-16
fix(cbp): accrue replacement budget against mature units and cap the carry

### LINKED ISSUE #370: CBP replacement budget accrues an unpayable debt while units are immature, then re-initializes the whole layer in a burst at t = maturity_threshold
## Summary

`maybe_replace_units` (`alberta_framework/core/continual_backprop.py`) adds `replacement_rate * layer_size` to the per-layer accumulator **every step, unconditionally**, but decrements it only when a replacement fires, and at most one unit per layer can fire per step. `ContinualBackpropConfig.replacement_rate` documents a *rate* ("`1e-4` means roughly one unit per 10000 steps in a 100-unit layer") and the module docstring claims `replacement_rate * num_hidden_units` units per step. Neither holds:

1. **Warm-up debt burst.** While no unit is mature (the first `maturity_threshold` steps), the budget keeps accruing against zero eligible units; the debt is then discharged one unit per step in an uninterrupted burst. Whenever `maturity_threshold >= 1/replacement_rate` the **whole layer** is re-initialized at `t = maturity_threshold` — a cliff in the learning curve that is an artifact of the accumulator, not of Continual Backprop, and that scales with the maturity threshold (an m200 arm gets a 4× bigger wipe than m50), making a maturity ablation uninterpretable. In-repo: `tests/test_plasticity_gate.py:204` (`rate=1e-3, maturity=100, 30 units`) wipes 3 units in 3 consecutive steps at t=100.
2. **Rate saturation with unbounded debt.** When `replacement_rate * layer_size > 1` the debt grows forever and the effective rate saturates at exactly 1 unit/layer/step, so any sweep over rates above `1/layer_size` produces bit-identical arms.

The reference GnT implementation (Dohare et al., `loss-of-plasticity`) accrues `replacement_rate * n_eligible` where eligible = units past the maturity threshold, so this is also a port-fidelity gap.

## Falsifier (deterministic, pristine `main` @ `45279f3`)

Drive `maybe_replace_units` with all ages advancing one per step (`hidden_sizes=(32,)`, `replacement_rate=1e-3`, `maturity_threshold=1000`, 1100 steps, utilities fixed):

```
rate=0.001 n_units=32 maturity=1000 steps=1100   (rate*n = 0.032 per step -> ~1 replacement per 31 steps)
actually replaced = 32   max per step = 1
first replacement at step 999; replacements in steps 999..1039: 32 of 32 units
-> the entire 32-unit layer is re-initialised in 32 consecutive steps
```

Saturation (`n_units=32`, `maturity=0`, 200 steps):

```
replacement_rate=0.1   rate*n=3.2   scheduled by contract=800   actually replaced=200  final accumulator=440.0
replacement_rate=0.5   rate*n=16    scheduled by contract=3200  actually replaced=200  final accumulator=3000.0
replacement_rate=1.0   rate*n=32    scheduled by contract=6400  actually replaced=200  final accumulator=6200.0
```

Expected: accrue `replacement_rate * n_mature_units` (immature units earn no budget, as in the reference), and never carry more than the one replacement a step can deliver.

## Dedup

`gh issue/pr list --search "replacement accumulator" / "maturity_threshold" / "continual backprop replacement"`: nothing. Adjacent, closed: #84 / #217 (CBP utility NaN guards). Same accumulator shape exists in `core/upgd.py::unit_replacement_rate` (saturation half only; separate module — will file separately). No open PR touches `continual_backprop.py`.

Prabhat1308:
Closes #370.

## Issue

`maybe_replace_units` accrued `replacement_rate * layer_size` every step but could pay at most one unit per step, so budget accrued against zero eligible units during the maturity warm-up and was discharged as a burst that re-initialized whole layers at `t = maturity_threshold` (32/32 units in 32 consecutive steps at `rate=1e-3, maturity=1000`), and rates above `1/layer_size` accrued unbounded debt while producing bit-identical arms.

## New state

The accumulator accrues `replacement_rate * n_mature_units` (units younger than `maturity_threshold` earn no budget), matching the reference GnT implementation, and never carries more than one pending replacement. Selection (lowest-utility mature unit), the one-per-step delivery, and the replacement itself are unchanged. The module docstring's Replacement paragraph now states the actual semantics.

`replace_units_with_flags` returns the exact gated per-layer decision alongside the states (`maybe_replace_units` keeps its two-tuple signature by wrapping it), and `CBPMultiHeadMLPLearner.update` reports that decision as `replacements_made` instead of re-inferring it from the old accrual formula. `ContinualBackpropState.replacement_accumulators` docstring states the actual semantics.

Failing-test-first: `test_replacement_budget_does_not_accrue_while_every_unit_is_immature`, `test_replacement_budget_never_carries_more_than_one_pending_unit`, and the wrapper-level `test_wrapper_replacements_made_is_the_gated_decision` (immature 32-unit layer at rate 0.02 → `replacements_made=[False]`, accumulator 0.0; matured layer with 0.9 pending → `[True]` and exactly one row re-initialised) fail on `main` / the previous head and pass here; the existing forced-replacement / maturity-protection tests are unchanged.

## Evidence

Falsifier from #370 at this head:

```text
rate=0.001 n_units=32 maturity=1000 steps=1100
actually replaced = 3   max per step = 1   final accumulator = 0.1
first replacement at step 1030; replacements in steps 1030..1070: 2 of 32 units
```

Verification at `49804652d264f5eb484dc0e5e9f05ca48c5e00d1`:

```text
.venv/bin/python -m pytest tests/test_continual_backprop.py tests/test_plasticity_gate.py tests/test_multi_head_learner.py -q -o addopts=""   -> 89 passed
.venv/bin/python -m ruff check .                                                                                            -> All checks passed!
.venv/bin/python -m mypy                                                                                                    -> Success: no issues found in 164 source files
```

`core/continual_backprop.py` is imported by the robot track and stays importable (public API unchanged); it is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

### DIFF
--- a/alberta_framework/core/continual_backprop.py
+++ b/alberta_framework/core/continual_backprop.py
@@ -144,9 +144,11 @@ class ContinualBackpropState:
             ``utilities``.
         replacement_accumulators: One scalar per hidden layer that
             accumulates the (fractional) number of replacements
-            scheduled by ``replacement_rate * num_units`` each step.
-            Whenever an accumulator exceeds 1.0 a single unit in that
-            layer is replaced and the accumulator is decremented by 1.
+            scheduled by ``replacement_rate * num_mature_units`` each
+            step. Whenever an accumulator reaches 1.0 and a mature unit
+            exists, a single unit in that layer is replaced and the
+            accumulator is decremented by 1; it never carries more than
+            one pending replacement.
             This makes ``replacement_rate << 1`` behave as expected in
             the JIT-compiled scan loop without integer arithmetic.
         rng_key: PRNG key used to draw fresh weights for replaced
@@ -465,16 +467,34 @@ def maybe_replace_units(
     Returns:
         Updated ``(mlp_state, cbp_state)``.
     """
-    if not config.enabled:
-        return mlp_state, cbp_state
+    new_mlp_state, new_cbp_state, _replaced = replace_units_with_flags(
+        mlp_state, cbp_state, config, sparsity
+    )
+    return new_mlp_state, new_cbp_state
 
+
+def replace_units_with_flags(
+    mlp_state: MultiHeadMLPState,
+    cbp_state: ContinualBackpropState,
+    config: ContinualBackpropConfig,
+    sparsity: float,
+) -> tuple[MultiHeadMLPState, ContinualBackpropState, Array]:
+    """Run :func:`maybe_replace_units` and also return the per-layer replacement flags.
+
+    Returns:
+        ``(mlp_state, cbp_state, replaced)`` where ``replaced`` is a boolean
+        array with one entry per hidden layer that is ``True`` exactly when
+        that layer's gated replacement fired this step.
+    """
     n_layers = len(cbp_state.utilities)
-    if n_layers == 0:
-        return mlp_state, cbp_state
+    no_replacements = jnp.zeros((n_layers,), dtype=jnp.bool_)
+    if not config.enabled or n_layers == 0:
+        return mlp_state, cbp_state, no_replacements
 
     rate = jnp.asarray(config.replacement_rate, dtype=jnp.float32)
     accum_arr = cbp_state.replacement_accumulators
     new_accum_list: list[Array] = []
+    replaced_flags: list[Array] = []
 
     new_mlp_state = mlp_state
     new_cbp_state = cbp_state
@@ -510,12 +530,13 @@ def maybe_replace_units(
         # more than the one replacement a step can deliver.
         accum_after = jnp.minimum(jnp.where(gated, accum - 1.0, accum), 1.0)
         new_accum_list.append(accum_after)
+        replaced_flags.append(gated)
 
     new_cbp_state = new_cbp_state.replace(  # type: ignore[attr-defined]
         replacement_accumulators=jnp.stack(new_accum_list),
         rng_key=rng_key,
     )
-    return new_mlp_state, new_cbp_state
+    return new_mlp_state, new_cbp_state, jnp.stack(replaced_flags)
 
 
 # =============================================================================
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking regression at exact head 671597d: the replacement scheduler now accrues against mature units, but CBPMultiHeadMLPLearner.update still infers replacements with the old `rate * hidden_size` formula. A public one-step repro with hidden_sizes=(32,), replacement_rate=0.02, maturity_threshold=1000 yields ages max=1 and accumulator=[0.] (no unit eligible/replaced) while `result.replacements_made == [True]`. This makes the exported per-step replacement diagnostic false exactly because of this PR's semantic change.

Please derive the flag from the actual gated replacement decision (or otherwise return it explicitly) and add a wrapper-level regression for immature units/high rate. Also update ContinualBackpropState's accumulator docstring, which still says `replacement_rate * num_units`. The budget behavior itself looks sound so far, but the head is not ready while its public diagnostic contradicts state mutation.

Prabhat1308:
Addressed in `4980465`:

- `replace_units_with_flags` now returns the exact gated per-layer decision (`maybe_replace_units` wraps it and keeps its two-tuple signature); `CBPMultiHeadMLPLearner.update` reports that flag as `replacements_made` instead of re-inferring it from the old `rate * hidden_size` accrual;
- wrapper-level regression `test_wrapper_replacements_made_is_the_gated_decision`: your repro (hidden 32, rate 0.02, maturity 1000, one step) now gives ages max=1, accumulator 0.0, `replacements_made=[False]`; a matured layer with 0.9 pending gives `[True]` with exactly one row re-initialised;
- `ContinualBackpropState.replacement_accumulators` docstring updated to the mature-unit accrual and one-pending cap.

Body updated to evidence head `4980465`; 89 passed across cbp/plasticity-gate/multi-head suites, ruff/mypy clean. Re-review requested.

lalalune (approved):
Re-review at exact head 4980465: READY. The follow-up returns the actual per-layer gated replacement decision instead of re-inferring it with the obsolete rate*hidden_size formula, updates the accumulator contract, and retains the public two-value maybe_replace_units API. The public false-positive repro is now covered in both no-eligible and fired directions. Exact head: 395 continual-backprop/plasticity/multi-head/IPMNIST tests passed, Ruff and strict mypy passed; scheduling/RNG/state behavior remains scoped, and no registered evidence source or output artifact changes.

PR #390, opened 2026-08-16
fix(steps): validate exact ratios against facade domains

### LINKED ISSUE #387: Step facade exact-domain checks can be bypassed by contradictory float ratios
PR #306 centralized float32 validation for Steps 3–10, but its domain checks still compare the original Python object after obtaining an exact ratio. A hostile or unusual concrete `float` subclass can therefore present an in-range ordinary value while returning an exact ratio below zero, above one, or nonzero below the GVF normal floor.

Reproduced on merged main:

- a value comparing as `0.5` with `as_integer_ratio() == (-1, 2**200)` is accepted as `-0.0` across Step probability/nonnegative fields;
- a value comparing as `0.5` with ratio `(2**200 + 1, 2**200)` is accepted as `1.0` in closed-unit fields;
- a hidden nonzero ratio `(1, 2**150)` bypasses the Step 3/4 GVF zero-or-normal rule and is stored as zero.

This violates the facade's fail-closed exact-host-domain contract and can reach downstream factories with values the declared domain excludes.

Acceptance criteria:

- retain exact numerator/denominator metadata from the single ratio read;
- validate sign, unit interval, and GVF nonzero/min-normal rules from that exact ratio;
- preserve accepted built-in-float serialization and signed-zero compatibility;
- reject malformed/noncanonical ratio providers consistently;
- add retained regressions across the affected Step 3–10 fields and keep valid default trajectories unchanged;
- no evidence or output artifact changes.

lalalune:
Closes #387

## Summary

Carry the exact numerator/denominator returned by the shared scalar converter through Step 3–10 validation, rather than comparing only the original `Real` object after it has supplied that ratio.

This closes exact-domain escapes where a concrete `float` subclass compares as an ordinary in-range value but exposes an exact ratio that is negative, above one, or nonzero below the GVF normal floor. The repair preserves compatible built-in float payloads and signed-zero serialization while making exact sign/unit/GVF rules fail closed.

Every Step 6 scientific scalar now uses the same single-read exact-ratio metadata. `q_step_size` and `average_reward_step_size` enforce exact non-negativity; `trace_decay`, `epsilon_start`, and `epsilon_end` enforce the exact closed unit interval; nested Step 7/9 payload construction refuses hostile Step 6 controls. Step 6 dimensions and seeds retain int32/bool guards.

The final runtime-type repair prevents dynamic `__class__` spoofing from routing a concrete float through the Integral shortcut, rejects spoofed non-Integral ratio components, and hardens Step 6 integer fields with the same actual-type rule.

## Exact-head evidence

Exact head `233ecb1f2b94e01cd3644a92d526f695e69789b0` is based directly on `origin/main` `751e9ab80db16a625a953ac264e2490210ed4f44`.

- original Step 6 hostile-domain selector: 10 failed before the exact-domain repair
- final class-spoof selector: 5 failed on predecessor `a49414af11dfb0a59bd6081086a61715c433186d`
- exact-head affected panel (`test_float32`, Step 5/6 production, central Step 3–10 validation): 290 passed
- changed-file Ruff passed
- strict mypy passed across 165 source files
- `git diff --check` passed
- no `outputs/**` artifact changes
- evidence registry exit 2 is the pre-existing registered-source drift state; this PR makes no performance or scientific-evidence claim

Overlapping open facade PRs that also edit the shared float32 helper must rebase after this change.

### DIFF
--- a/alberta_framework/_float32.py
+++ b/alberta_framework/_float32.py
@@ -5,6 +5,13 @@ from __future__ import annotations
 import math
 import struct
 from numbers import Integral, Real
+from typing import cast
+
+
+def _is_actual_integral(value: object) -> bool:
+    """Return whether the runtime type is a non-bool Integral subtype."""
+    actual_type = type(value)
+    return not issubclass(actual_type, bool) and issubclass(actual_type, Integral)
 
 
 def _round_quotient_ties_to_even(numerator: int, denominator: int) -> int:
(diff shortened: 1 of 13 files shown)


lalalune:
Blocking exact-head review of `a49414af11dfb0a59bd6081086a61715c433186d`:

`_real_ratio` starts with an actual-runtime-type guard, but then branches with spoofable `isinstance(value, Integral)` at `alberta_framework/_float32.py:85`. A concrete `float` subclass whose ordinary value is `0.5`, whose `__class__` property returns `int`, and whose `as_integer_ratio()` returns `(-1, 2**200)` enters the integer branch. The ratio method is called zero times, the helper returns `(0, 1, 0.0)`, and direct Step 6 plus nested Step 7/9 all accept `q_step_size=0.0`. This defeats both the exact-domain rule and the required single ratio read.

The ratio-component checks at lines 95-100 have the same issue: a non-Integral object spoofing `__class__ == int` and implementing `__int__` is accepted as a valid tuple member, so a malformed provider returning `(FakeInteger(), 2)` is accepted and stored. Use actual runtime types for integral dispatch and tuple-member validation (while retaining genuine NumPy Integral support), and add retained direct Step 6 plus nested Step 7/9 regressions covering the spoof.

The existing focused exact-ratio/hostile/signed-zero/serialization panel is otherwise green (71 passed), targeted Ruff and strict mypy pass, and the head is cleanly based on current main. I cannot submit a formal changes-requested review because the authenticated account owns this PR.

PR #395, opened 2026-08-16
fix(step11): reject illegal OaK facade config

deepanshu-yd:
Reconciles the broad Step 11 OaK facade contract with current shared validators and core STOMP limits.

Changes:
- validate/canonicalize all float32-consumed scientific scalars through the shared exact-ratio policy
- normalize ordinary conversion-hook failures to ValueError
- accept only trusted builtin/NumPy integer families and enforce signed-int32 dimensions/counters
- keep option_planning_backups_per_step strictly below INT32_MAX, matching STOMPConfig
- require exact tuple and SubtaskSpec records
- validate smoke steps/seed and retain all-update health checks

Validation at exact head 1f0cfdc:
- Step 11 suite: 204 tests passed before one test-only misplaced assertion; corrected boundary panel: 14 passed
- options + STOMP planning consumers: 33 passed
- Ruff: clean
- strict mypy for both changed source files: clean
- git diff --check: clean

### DIFF
--- a/alberta_framework/steps/_float32_validation.py
+++ b/alberta_framework/steps/_float32_validation.py
@@ -17,7 +17,7 @@ def finite_real_and_float32(name: str, value: object) -> tuple[Real, int, int, f
     real = cast(Real, value)
     try:
         numerator, denominator, narrowed = round_real_to_float32_with_ratio(real)
-    except (FloatingPointError, OverflowError, TypeError, ValueError):
+    except Exception:
         raise ValueError(f"{name} must be finite, got {value!r}") from None
     if not math.isfinite(narrowed):
         raise ValueError(f"{name} must be finite, got {value!r}")
--- a/alberta_framework/steps/step11.py
+++ b/alberta_framework/steps/step11.py
@@ -34,6 +34,7 @@ from typing import Any, cast
 
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 
 from alberta_framework.core.oak import (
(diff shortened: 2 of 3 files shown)


lalalune (changes requested):
Blocking exact-head regression at c5e540f: the split validators remove the narrowed-domain checks that main already had. A concrete float subclass with host payload 0.5 and as_integer_ratio()=(-1,1) is accepted as base_trace_decay=-1.0 and curation_threshold=-1.0; ratio (2,1) is accepted as option_gamma=2.0. This is the exact host-versus-ratio escape addressed by #390, and this PR currently duplicates the converter instead of consuming that shared ratio result. Please base the Step 11 validation on #390’s shared exact-ratio validator (or wait for #390, then narrow this PR), check both original and exact-ratio/sink domains, and add these adversarial unit/nonnegative cases. Keep the integer/smoke bounds, but do not stack another scalar converter.

deepanshu-yd:
Updated at 72784e9:
- Replaced split narrowing with shared `round_real_to_float32_with_ratio` validator.
- Validated both host real value and exact `(numerator, denominator)` ratio against unit/non-negative facade domains.
- Rejected non-Real dynamic `__class__` property spoofing at both helper and facade level.
- Added comprehensive adversarial unit, negative ratio, and class spoof test cases in `tests/test_step11_production.py` and `tests/test_float32.py`.
- Preserved signed int32 bounds on dimensions, actions, planning backups, option steps, and smoke parameters.

deepanshu-yd:
Updated at 95262aa:
- Rebased cleanly on latest `main` using shared `_float32_validation.py` exact-ratio validators.
- Enforced strict tuple container validation on `subtask_specs` (`type(subtask_specs) is tuple`), rejecting spoofed class containers.
- Enforced bounded integer validation on `steps` and `seed` in smoke probe.
- Added unit regressions in `tests/test_step11_production.py` for spoofed tuple container and adversarial exact ratio floats.

lalalune (approved):

lalalune:
The reviewed head 95262aa of this PR was merged and is reachable from main's history, so the Step 11 OaK facade hardening has landed with commit credit. The current force-pushed rebase conflicts with semantics frozen on main afterwards. Closing as merged-in-substance — thank you!

Issue #365, opened 2026-08-16
run_step1_smoke accepts bool and non-integral steps/final_window


Svector-anu:
Same pattern as #347/#355: `run_step1_smoke` in `alberta_framework/steps/step1.py` validated `steps` and `final_window` with bare comparisons (`steps < 1`, `final_window < 1 or final_window > steps`) even though this module defines and uses `_require_int` elsewhere to reject booleans and non-integral inputs. `bool` (an `int` subclass) and non-integral floats slip through the bare comparisons, and depending on which field/value combination, the facade either silently accepts a float count or leaks an implementation-level JAX/Python `TypeError` instead of its documented `ValueError`.

confirmed directly:
```text
{'steps': True, 'final_window': True} -> TypeError: iota does not accept dtype bool. Accepted dtypes are subtypes of integer, complexfloating, floating.
{'steps': 4.0, 'final_window': 2} -> ACCEPTED steps=4.0, metrics_shape=(4, 4)
{'steps': 4, 'final_window': 2.0} -> TypeError: 'float' object cannot be interpreted as an integer
```

reachability: `run_step1_smoke` is in `alberta_framework/steps/__init__.py`'s `__all__` and the package-root `alberta_framework/__init__.py`'s `__all__` — both confirmed directly. It's also called by `alberta_framework/cli.py::step1_smoke_main`, the installed `alberta-step1-smoke` console entry point (argparse constrains normal CLI input to integers, but the two public Python exports remain reachable with the invalid inputs shown above).

fix incoming: validate and canonicalize both counts through the existing `_require_int(..., minimum=1)` helper before applying the cross-field `final_window > steps` constraint.

PR #348, opened 2026-08-16
fix(step6): reject bool and non-integral smoke dimensions

Svector-anu:
Closes #347.

## What changed

`run_step6_smoke` validated `steps` and `feature_dim` with plain `< 1` comparisons. Because Python's `bool` is an `int` subclass, `steps=True`/`feature_dim=True` (`== 1`) pass the check and are silently accepted as valid dimensions instead of being rejected as a type error. Non-integral values such as `1.5` also pass the `< 1` check and reach JAX shape construction, which raises a lower-level `TypeError` (`Shapes must be 1D sequences of concrete values of integer type...`) instead of the facade's documented `ValueError` contract — an invalid smoke probe could silently run one transition with `steps=True`, or fail nondeterministically at a lower layer with an unhelpful error for a non-integral input.

confirmed directly before touching the source:

```text
{'steps': True} ACCEPTED True (1, 2)
{'steps': 1.5} TypeError Shapes must be 1D sequences of concrete values of integer type, got (2.5, 6).
{'feature_dim': True} ACCEPTED 32 (32, 2)
{'feature_dim': 1.5} TypeError Shapes must be 1D sequences of concrete values of integer type, got (33, 1.5).
```

fix: canonicalize both arguments through the file's existing strict `_require_int(..., minimum=1)` helper — already used elsewhere in this same file for `n_actions`, `epsilon_decay_steps` — which correctly rejects `bool` (`isinstance(value, bool)` check first) and non-`Integral` values before they ever reach JAX, raising a proper `ValueError` naming the offending field.

## Reachability

`run_step6_smoke` is exported by `alberta_framework.steps.__init__` and re-exported by the package-root `alberta_framework.__init__` — confirmed both directly, a real public entry point, not an internal-only helper.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_step5_step6_production.py -q -o addopts=""
145 passed in 9.71s

$ .venv/bin/python -m ruff check alberta_framework/steps/step6.py tests/test_step5_step6_production.py
All checks passed!

$ .venv/bin/python -m mypy
Success: no issues found in 164 source files
```

New test: `test_step6_smoke_rejects_non_integral_dimensions`, parametrized over `(steps, True)`, `(steps, 1.5)`, `(feature_dim, False)`, `(feature_dim, 1.5)`, asserting each raises `ValueError` naming the offending field.

[...]

PR #400, opened 2026-08-16
fix(step12): validate smoke step count as an integer

### LINKED ISSUE #397: run_step12_smoke accepts bool and non-integral steps
Same pattern as #347/#355/#365/#376: `run_step12_smoke` in `alberta_framework/steps/step12.py` validated `steps` with a plain `steps < 1` comparison instead of this module's own strict `_require_int(..., minimum=1)` helper. `steps=True` (an `int` subclass) is silently accepted and recorded as a boolean while running one transition; `steps=1.5` passes the `< 1` check and leaks a raw JAX shape `TypeError` instead of the facade's documented `ValueError`.

confirmed directly:
```text
steps=True: accepted, result.steps=True, predictions_shape=(1, 4)
steps=1.5: TypeError: Shapes must be 1D sequences of concrete values of integer type, got (2.5, 4).
```

reachability: `run_step12_smoke` is imported and listed in `alberta_framework.steps.__init__`'s `__all__` — confirmed directly, publicly reachable with ordinary concrete inputs. Not re-exported from the package root.

fix incoming: route `steps` through the existing `_require_int("steps", steps, minimum=1)` helper.

Svector-anu:
Closes #397.

## What changed

Same pattern as #348/#356/#368/#377: `run_step12_smoke` validated `steps` with a plain `steps < 1` comparison instead of this module's own strict `_require_int(..., minimum=1)` helper. `steps=True` (an `int` subclass) is silently accepted and recorded as a boolean while running one transition; `steps=1.5` passes the `< 1` check and leaks a raw JAX shape `TypeError` instead of the facade's documented `ValueError`.

```text
steps=True: accepted, result.steps=True, predictions_shape=(1, 4)
steps=1.5: TypeError: Shapes must be 1D sequences of concrete values of integer type, got (2.5, 4).
```

fix: route `steps` through the existing `_require_int("steps", steps, minimum=1)` helper, matching every other validated field in this module.

## Reachability

`run_step12_smoke` is imported and listed in `alberta_framework.steps.__init__`'s `__all__` — confirmed directly, publicly reachable with ordinary concrete inputs. Not re-exported from the package root.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_step12_production.py -q -o addopts=""
166 passed, 1 failed in 241.82s

$ .venv/bin/python -m ruff check alberta_framework/steps/step12.py tests/test_step12_production.py
All checks passed!

$ .venv/bin/python -m mypy
```

New test: `test_run_step12_smoke_rejects_non_integer_steps`, parametrized over `True` and `1.5`, asserting `ValueError` naming `steps`.

mutation-resistance verified independently: reverted just the source fix (`git stash push -- alberta_framework/steps/step12.py`, kept the test), reran — both cases failed (`True` silently accepted instead of raising; `1.5` leaked the raw JAX `TypeError`), reproducing the exact bug described above. restored the fix, green again.

**Note on the one failure in the full-file run:** `test_step12_narrows_original_real_directly_without_double_rounding` is a pre-existing, unrelated failure — a platform-specific `np.longdouble`→`float32` rounding assumption (same failure class already documented on the sibling `step11.py` PR, #365). Independently confirmed it fails identically on an unmodified checkout (via `git stash`, rerunning against the pre-existing source) before writing this PR, so it's pre-existing, not introduced by this change.

## Evidence

a49a92315d9581960334d209d247c1ee633d6411

```text
$ .venv/bin/python -m pytest tests/test_step12_production.py -q -o addopts="" -k test_run_step12_smoke_rejects_non_integer_steps
FAILED [True] - Failed: DID NOT RAISE ValueError
FAILED [1.5] - TypeError: Shapes must be 1D sequences of concrete values of integer type...
2 failed, 165 deselected in 20.57s

green (fix restored):
$ .venv/bin/python -m pytest tests/test_step12_production.py -q -o addopts="" -k test_run_step12_smoke_rejects_non_integer_steps
2 passed, 165 deselected in 0.77s
```

```text
steps=True: accepted, result.steps=True, predictions_shape=(1, 4)
steps=1.5: TypeError: Shapes must be 1D sequences of concrete values of integer type, got (2.5, 4).
```
independently reran the full touched test file (167 tests), ruff, and mypy myself; independently confirmed the one remaining failure is pre-existing by reproducing it on an unmodified checkout; independently confirmed the export chain before writing the reachability claim; did my own revert-and-confirm-red mutation check.

## Provenance

AI provider/model: openai / gpt-5.6-sol (fix, tests, via AgentRouter proxy); anthropic / claude-sonnet-5 (independent re-verification: reran the full touched test file, ruff, and mypy myself, independently confirmed the pre-existing failure is unrelated, verified the export chain, did my own revert-and-confirm-red mutation check, opened the governing issue, pushed, opened this PR)

### DIFF
--- a/alberta_framework/steps/step12.py
+++ b/alberta_framework/steps/step12.py
@@ -31,7 +31,7 @@ from __future__ import annotations
 
 from dataclasses import asdict, dataclass
 from numbers import Integral, Real
-from typing import Any
+from typing import Any, cast
 
 import jax.numpy as jnp
 import jax.random as jr
@@ -198,9 +198,10 @@ def _require_int(
     minimum: int | None = None,
     exclusive_maximum: int | None = None,
 ) -> int:
-    if isinstance(value, bool) or not isinstance(value, Integral):
+    actual_type = type(value)
+    if actual_type is bool or not issubclass(actual_type, Integral):
         raise ValueError(f"{name} must be an integer, got {value!r}")
-    number = int(value)
+    number = int(cast(Integral, value))
     if minimum is not None and number < minimum:
         if minimum == 1:
             raise ValueError(f"{name} must be positive, got {value!r}")
--- a/alberta_framework/steps/step12.py
+++ b/alberta_framework/steps/step12.py
@@ -426,8 +426,7 @@ def run_step12_smoke(
     Returns:
         :class:`Step12SmokeResult` with shape/fineness summary.
     """
-    if steps < 1:
-        raise ValueError("steps must be positive")
+    steps = _require_int("steps", steps, minimum=1)
 
     cfg = config or Step12IAConfig()
     agent = make_step12_ia_agent(cfg)
--- a/tests/test_step12_production.py
+++ b/tests/test_step12_production.py
@@ -1037,6 +1037,12 @@ def test_run_step12_smoke_zero_steps_raises() -> None:
         run_step12_smoke(steps=0)
 
 
+@pytest.mark.parametrize("steps", [True, 1.5])
+def test_run_step12_smoke_rejects_non_integer_steps(steps: object) -> None:
+    with pytest.raises(ValueError, match="steps"):
+        run_step12_smoke(steps=steps)  # type: ignore[arg-type]
+
+
 def test_run_step12_smoke_cerebellum_errors_shape() -> None:
     result = run_step12_smoke(steps=16)
     assert result.cerebellum_errors_shape == (16, 4)
--- a/tests/test_step12_production.py
+++ b/tests/test_step12_production.py
@@ -1043,6 +1043,29 @@ def test_run_step12_smoke_rejects_non_integer_steps(steps: object) -> None:
         run_step12_smoke(steps=steps)  # type: ignore[arg-type]
 
 
+def test_run_step12_smoke_rejects_class_spoofed_integer_steps() -> None:
+    """A reviewer-found adversarial input: an object whose ``__class__``
+    property reports ``int`` (fooling ``isinstance``) while its real,
+    non-spoofable ``type()`` is not ``int``. The facade must dispatch on the
+    actual runtime type, not the object's self-reported class."""
+
+    class SpoofedInt:
+        @property
+        def __class__(self) -> type:  # noqa: A003
+            return int
+
+        def __int__(self) -> int:
+            return 3
+
+        def __index__(self) -> int:
+            return 3
+
+    spoofed = SpoofedInt()
+    assert isinstance(spoofed, int)  # confirms the spoof actually fools isinstance
+    with pytest.raises(ValueError, match="steps"):
+        run_step12_smoke(steps=spoofed)  # type: ignore[arg-type]
+
+
 def test_run_step12_smoke_cerebellum_errors_shape() -> None:
     result = run_step12_smoke(steps=16)
     assert result.cerebellum_errors_shape == (16, 4)


lalalune (changes requested):
Exact head a49a92315d9581960334d209d247c1ee633d6411 still accepts a non-integral `steps` object, so #397 is not fully closed. The routed `_require_int` uses `isinstance(value, Integral)`, which honors a spoofed `__class__`: an ordinary object reporting `__class__ = int` and implementing `__int__ -> 3` passes, is canonicalized to 3, and the smoke runs. (`type(x)` remains the custom class while `isinstance(x, Integral)` is True.) The issue requires non-integral counts to fail at the facade. Dispatch on actual runtime type / a fail-closed accepted scalar set, and retain this adversarial regression alongside bool/1.5. The ordinary two-case fix and 20/20 CI are otherwise sound.

Svector-anu:
Thanks — that's a real gap, and I confirmed it empirically before fixing: `isinstance(fake, Integral)` returns `True` for an object with a `__class__` property spoofing `int`, while `issubclass(type(fake), Integral)` correctly returns `False` since `type()` reads the real, non-spoofable type slot.

Fixed by dispatching on `type(value)` directly (mirroring the already-safe pattern in `step6.py::_require_int`, which uses `actual_type = type(value)` rather than `isinstance`) instead of `isinstance(value, Integral)`. Confirmed `np.int64` still passes (`issubclass(np.int64, Integral)` is `True`) so no legitimate numpy-integer caller is affected — only the spoofed-`__class__` case and the previously-allowed `bool`/`1.5` cases are rejected.

Added and retained `test_run_step12_smoke_rejects_class_spoofed_integer_steps`: constructs an object with `__class__` returning `int`, asserts `isinstance(spoofed, int)` is `True` (proving the spoof works), then asserts `run_step12_smoke(steps=spoofed)` still raises `ValueError`. Independently verified this test is mutation-resistant — reverted just to the old `isinstance`-based check, reran, confirmed `DID NOT RAISE ValueError`; restored the fix, green again.

```text
$ .venv/bin/python -m pytest tests/test_step12_production.py -q -o addopts=""
167 passed, 1 failed in 120.93s
```
(the 1 failure is the pre-existing, unrelated `np.longdouble` platform assertion already documented on this PR's own description — reproduces identically on an unmodified checkout.)

```text
$ .venv/bin/python -m ruff check alberta_framework/steps/step12.py tests/test_step12_production.py
All checks passed!

$ .venv/bin/python -m mypy
```

head: `939d345dd513f0e7124f568e9e2829a359ecd979`

One thing worth flagging honestly rather than silently expanding scope on: this same file's `_require_real`/`_require_unit_interval`/`_require_nonnegative_real`/`_require_positive_real` helpers use the identical `isinstance(value, Real)` pattern for the float-typed config fields (`cerebellum_step_size`, `base_step_size`, etc.), which is spoofable by the same technique. That's outside what this review flagged (which was specifically about `_require_int`/`steps`), so I left it alone here rather than turning a scoped bugfix into an unrequested rewrite — flagging it in case you'd like a separate, dedicated PR for it.

lalalune:
Superseded by the broader repaired Step 12 facade hardening in #417, which includes the exact smoke-step integer/int32 contract plus the shared float32, tuple, dimension, and STOMP-bound fixes. #417 is on a current-main head with local full-file coverage and fresh CI in progress.

PR #404, opened 2026-08-16
fix(cbp): validate every hidden width in init_cbp_state

### LINKED ISSUE #403: init_cbp_state validates only the hidden-layer count, not the widths; a width mismatch silently reshapes the tracker and defeats the maturity gate
## Summary

`init_cbp_state` (`alberta_framework/core/continual_backprop.py`) documents its `mlp_state` argument as "Used only to validate that `hidden_sizes` matches the trunk shape", but the code compares only `len(hidden_sizes)` against the number of trunk layers. `ContinualBackpropTracker` is documented as the hand-driven entry point "against any compatible MLP-shaped state", so a width mismatch is a supported public path — and it is accepted silently:

- `update_utility` broadcasts the length-1 utility vector against the length-8 contribution and the state silently changes shape (`utilities[0]: (1,) -> (8,)`), while `ages` stays length-1, so every unit shares one age and the whole layer becomes mature simultaneously — the maturity gate is defeated;
- the age reset for a replaced unit is written into the length-1 array via out-of-bounds index clamping and lost, so the just-replaced unit stays permanently eligible and gets re-replaced.

```python
from alberta_framework.core.continual_backprop import init_cbp_state, update_utility
learner = MultiHeadMLPLearner(n_heads=2, hidden_sizes=(8,), sparsity=0.0)
mlp_state = learner.init(feature_dim=4, key=jr.key(0))
cbp = init_cbp_state(mlp_state, (1,), key=jr.key(1))            # width 1 for an 8-wide layer: ACCEPTED
print("utilities:", cbp.utilities[0].shape, "ages:", cbp.ages[0].shape)
cbp = update_utility(cbp, (jnp.ones(8),), (jnp.ones(8),), 0.99)
print("after one update -> utilities:", cbp.utilities[0].shape, "ages:", cbp.ages[0].shape)
```

Output on `main`:

```
utilities: (1,) ages: (1,)
after one update -> utilities: (8,) ages: (1,)
```

Expected: `ValueError: hidden_sizes[0]=1 does not match trunk layer 0 width (8)`.

## Dedup

Existing coverage (`test_init_cbp_state_mismatch_raises`) exercises only the layer-count mismatch. `gh issue/pr list --search "init_cbp_state" / "hidden_sizes shape"`: nothing. Related open (mine, different function in the same module): #370/#371; merge-tree composes cleanly.

Prabhat1308:
Closes #403.

## Issue

`init_cbp_state` compared only the number of hidden layers, so a width mismatch was accepted and the tracker's `utilities` silently reshaped on the first update while `ages` stayed one entry wide — defeating the maturity gate and losing the age reset of replaced units.

## New state

`init_cbp_state` checks each layer's width against the trunk (`hidden_sizes[1]=1 does not match trunk layer 1 width (4)`) after the existing count check; the docstring states the full contract. Matching shapes are unchanged.

Failing-test-first: `test_init_cbp_state_width_mismatch_raises` fails on `main` and passes here.

## Evidence

Falsifier from #403 at this head:

```text
ValueError: hidden_sizes[0]=1 does not match trunk layer 0 width (8)
```

Verification at `b03c1b3c4a9cfa3f202bf6531a4d7d54f61cd1e7`:

```text
.venv/bin/python -m pytest tests/test_continual_backprop.py tests/test_plasticity_gate.py -q -o addopts=""   -> 26 passed
.venv/bin/python -m ruff check .                                                                        -> All checks passed!
.venv/bin/python -m mypy                                                                                -> Success: no issues found in 164 source files
```

Composes cleanly with #371 (`git merge-tree` clean). `core/continual_backprop.py` is imported by the robot track and stays importable; it is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

### DIFF
--- a/alberta_framework/core/continual_backprop.py
+++ b/alberta_framework/core/continual_backprop.py
@@ -175,7 +175,7 @@ def init_cbp_state(
         mlp_state: An initialized :class:`MultiHeadMLPState`. Used only
             to validate that ``hidden_sizes`` matches the trunk shape.
         hidden_sizes: Hidden layer sizes from the learner constructor.
-            Must match ``len(mlp_state.trunk_params.weights)``.
+            Must match the trunk layer count and every layer's width.
         key: JAX random key for replacement weight sampling.
 
     Returns:
@@ -189,6 +189,14 @@ def init_cbp_state(
             f"count ({len(mlp_state.trunk_params.weights)})."
         )
         raise ValueError(msg)
+    for layer_idx, (width, weight) in enumerate(
+        zip(hidden_sizes, mlp_state.trunk_params.weights, strict=True)
+    ):
+        if int(weight.shape[0]) != int(width):
+            raise ValueError(
+                f"hidden_sizes[{layer_idx}]={width} does not match trunk layer "
+                f"{layer_idx} width ({weight.shape[0]})"
+            )
 
     utilities = tuple(
         jnp.zeros(h, dtype=jnp.float32) for h in hidden_sizes
--- a/tests/test_continual_backprop.py
+++ b/tests/test_continual_backprop.py
@@ -7,6 +7,7 @@ import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import pytest
 
 from alberta_framework.core.continual_backprop import (
     CBPLearningResult,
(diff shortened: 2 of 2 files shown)


lalalune (approved):
Reviewed exact `b03c1b3c4a9cfa3f202bf6531a4d7d54f61cd1e7`. The new check matches the MLP trunk convention `(fan_out, fan_in)` and refuses every per-layer hidden-width mismatch before tracker allocation while preserving valid/linear cases. Exact focused `test_continual_backprop.py`: 22 passed; focused Ruff and diff-check clean. No evidence source or output changes.

PR #408, opened 2026-08-16
fix(world-model-ensemble): stop max_raw_gradient_norm deadlocking legal training

### LINKED ISSUE #366: RecurrentLatentWorldModelEnsemble: max_raw_gradient_norm rejects legal in-bounds transitions after ordinary training and the rejection is a permanent deadlock
## Summary (contract question — no PR attached; the fix is a design decision)

`RecurrentLatentWorldModelEnsembleConfig` (`alberta_framework/core/recurrent_latent_world_model_ensemble.py`) documents the `max_*` fields as "fail-closed validity gates, not tuned hyperparameters … coarse sanity ceilings far above any healthy operating range". `max_raw_gradient_norm = 100_000` is compared against each member's raw NLL gradient (`member_gradients_valid`), and a single member breaching it vetoes the whole event, leaving persistent state bit-identical.

The reachable raw NLL gradient scales as `residual / variance`, and `variance_floor = 1e-3` puts a 1000× multiplier on it. Nothing at construction relates `max_raw_gradient_norm` to `max_input_magnitude` (1000) or `variance_floor`. So once the heteroscedastic head has correctly learned that a stream is low-noise, a transition whose every value is far inside `max_input_magnitude` blows through the ceiling — and because rejection leaves state unchanged, the identical event is rejected again forever. `gradient_clip_norm = 10` already exists to bound the committed step (and `candidate_parameters_valid` independently confirms it is benign), so the raw-gradient ceiling rejects exactly the case the clipper was built for.

Two consequences for measurement in a continual-learning module where the distribution shift *is* the measurement:

1. The model deadlocks at the shift, silently, while every input-validity, ownership, boundary, and capacity diagnostic reports healthy.
2. `_rejected_result` returns `next_start_cache = _zero_start_cache()`; the module's own `test_jit_and_scan_match_sequential_cache_state_and_outputs` carries `result.next_start_cache` in the scan carry, so after one rejected event every later `decide` returns a zero cache and every later `update` is rejected — the run is frozen from that step on. Each rejected step also emits `mean_negative_log_likelihood = 0.0` (no availability flag other than `diagnostics.applied`), so a windowed NLL over a stalled run reads *better* than the healthy stationary window (0.0 vs −0.976 in the repro).

## Reproduction (deterministic, pristine `main` @ `45279f3`; all bounds and rates at their defaults)

`observation_dim=2, n_actions=2, latent_dim=8, ensemble_size=3, uncertainty_warmup_steps=0, max_updates=100000`, everything else default (`learning_rate=1e-3`, `bootstrap_probability=0.8`, `max_raw_gradient_norm=1e5`, `max_input_magnitude=1000`). Train 30k events on a fixed low-noise transition (`obs=(0.25,-0.5) → bootstrap=(0.75,0.125)`, reward 0.5, discount 0.9), then present a legal shifted bootstrap observation:

```
after 30000 default-rate updates, learned variances: [0.00113 0.00114 0.00113 0.00116]
  legal shift +-10     (ceiling 1000) applied=True  grads_valid=[True, True, True]
  legal shift +-50     (ceiling 1000) applied=False grads_valid=[False, False, False]
  legal shift +-100    (ceiling 1000) applied=False grads_valid=[False, False, False]
  legal shift +-200    (ceiling 1000) applied=False grads_valid=[False, False, False]
```

The threshold sits near residual ≈ 42, i.e. ~4% of the legal input range (member grad norms `[36134, 143379, 69644]` at ±50 vs ceiling 100000). Repeating the same legal ±50 transition: `50/50 silently rejected; event_count frozen | every other gate: state True input True ownership True boundary True capacity True`. Carrying `next_start_cache` like the scan loop after that single rejection: `events applied after the rejected shift: 0/29`.

## Options (for the maintainer to choose)

- (a) make `member_gradients_valid` a finiteness gate only and let `gradient_clip_norm` + the existing `candidate_parameters_valid` / `candidate_state_valid` gates bound the committed step; or
- (b) validate at construction that `max_raw_gradient_norm` covers the worst legal raw gradient implied by `max_input_magnitude` and `variance_floor` (which would reject the current defaults); and, either way,
- (c) on a rejection where `state_valid & cache_valid` hold, return a `next_start_cache` that re-owns the unchanged state so a transient rejection is retryable rather than terminal, and give the loss channel an explicit availability flag (mirroring `representation_gradient_available`).

I have not opened a PR because (a) vs (b) changes what the gate means; happy to implement whichever you pick, failing-test-first.

## Dedup

Related open (mine): #333/#336 (init state validity in the same module — different mechanism). `gh issue/pr list --search` for "max_raw_gradient_norm", "gradient norm reject", "next_start_cache", "world model ensemble": only #139/#140 (checkpoint format, unrelated) and PR #61 (ObGD 0×inf).

rama0x1:
Closes #366.

Agent disclosure: client `claude-code`, provider `anthropic`, model
`claude-sonnet-5`.

## Bug

`RecurrentLatentWorldModelEnsemble.update()` rejected the whole event
(state unchanged) whenever any member's raw NLL gradient norm exceeded
`max_raw_gradient_norm`. The raw gradient scales as `residual / variance`,
so a well-trained low-noise member legally produces a raw gradient far
above any fixed ceiling on an ordinary in-bounds transition —
`gradient_clip_norm` already bounds the *committed* step regardless of raw
scale, so the ceiling rejected exactly the case the clipper exists for.

Worse, a rejection returned the zero/invalid `next_start_cache`
unconditionally. Since `decide()`/`update()` both require a valid, owned
start cache, the very first false rejection made every later
`decide`/`update` call fail ownership forever — a **permanent** deadlock,
not a transient reject, exactly as described in the issue.

## Fix

Two changes, one bug:

1. `member_gradients_valid` is now finiteness-only (dropped the
   `gradient_norm <= max_raw_gradient_norm` comparison).
   `gradient_clip_norm` (already applied to every committed step) plus the
   existing `candidate_parameters_valid` / `candidate_state_valid` gates
   still bound what gets committed. `max_raw_gradient_norm` is unchanged
   for the representation gradient (a returned diagnostic, never itself
   clipped) and still constrains `gradient_clip_norm` at construction —
   this is option (a) from the issue.
2. `_rejected_result()` now re-owns the unchanged `state` in
   `next_start_cache` (observation = `next_decision_observation`) whenever
   the incoming state and decision cache were themselves sound
   (`state_valid & cache_valid`). The next `decide`/`update` call
   independently re-validates the observation and every other input, so
   nothing unsound is smuggled through — this only removes the
   *permanence* of a lockup, whatever its cause. This is option (c) from
   the issue, applied regardless of which of (a)/(b) was chosen. A state
   that is itself invalid (e.g. tampered/corrupt) still gets the
   non-recoverable zero cache, unchanged from before.

`max_raw_gradient_norm` is otherwise untouched: the representation
gradient can still legitimately reject at an extreme enough shift (see the
±200 case in verification below) — the fix guarantees that rejection is no
longer permanent, not that every rejection is eliminated.

## Verification

Head: `ae408df40bfe59e95561f158eb6627e8be330672`

Environment: CPU only (`JAX_PLATFORMS=cpu`; this container has no GPU and
`jax_plugins.xla_cuda12` fails to init otherwise — harmless, JAX falls back
to CPU).

Exact reproduction of the issue's own repro (`observation_dim=2,
n_actions=2, latent_dim=8, ensemble_size=3, uncertainty_warmup_steps=0,
max_updates=100000`, all other fields default), 30,000 steps on the fixed
`obs=(0.25,-0.5) -> bootstrap=(0.75,0.125)` transition, then legal shifted
bootstraps:

```text
learned member_aleatoric_variances (post-training): ~0.0011-0.0012 per head/member
                                    before fix   |   after fix
shift +-10   (ceiling 1e5)  applied=True         |  applied=True   (unchanged - was already accepted)
shift +-50   (ceiling 1e5)  applied=False        |  applied=True   member_gradients_valid=[T,T,T]
shift +-100  (ceiling 1e5)  applied=False        |  applied=True   member_gradients_valid=[T,T,T]
shift +-200  (ceiling 1e5)  applied=False        |  applied=False  (representation_gradient_valid
                                                                     still trips here - unaddressed,
                                                                     out of scope, see below)
repeated +-50 shift, 50x    0/50 applied         |  1/50 applied, cache stays valid=True throughout
one rejection at +-50, then carrying
next_start_cache through 29 ordinary
in-bounds transitions exactly like the
issue's scan-loop scenario             0/29 applied  |  n/a (± 50 no longer rejects post-fix at all)
```

Because the ±50/±100 false positive is gone post-fix, I additionally
verified the *permanence* mechanism directly using the still-genuinely-
rejecting ±200 case (representation-gradient ceiling, untouched by this
fix):

```text
extreme +-200 shift: applied=False, representation_gradient_valid=False (expected, unaddressed)
result.next_start_cache.valid: True   (was False before this fix)
result.next_start_cache.{owner_event_count,owner_hidden_states}: unchanged from input state
result.state: bit-identical to input state (rejection is still a strict no-op)
decide(state, result.next_start_cache, action).valid: True
a subsequent ORDINARY in-bounds transition from that recovered cache: applied=True
```

i.e. the ±200 shift still correctly rejects (a real, separate gate this PR
does not touch), but it no longer poisons the run: the next legal event is
judged on its own merits instead of being auto-rejected by a permanently
invalid cache.

Failing-test-first: both new regression tests were run against the
pre-fix source and fail there (`assert bool(result.diagnostics.applied)`
and `assert bool(rejected.next_start_cache.valid)` respectively), then
pass against this PR's diff.

```text
$ JAX_PLATFORMS=cpu .venv/bin/python -m pytest tests/test_recurrent_latent_world_model_ensemble.py -q -o addopts=""
22 passed in 58.62s

$ .venv/bin/python -m ruff check alberta_framework/core/recurrent_latent_world_model_ensemble.py tests/test_recurrent_latent_world_model_ensemble.py
All checks passed!

$ .venv/bin/python -m mypy

$ .venv/bin/python -m pytest --collect-only -q
7332 tests collected in 6.40s   (no collection-time breakage repo-wide)

$ git diff --check
(clean)
```

I was not able to complete a full run of `tests/test_prototype_agent.py`
(the module's real integration consumer) within this session — the
container is CPU-only and shared with other concurrent agents, and that
suite is slow under JAX/CPU contention. It does not currently exercise
this ensemble's `update()` path at all (`grep -rl
"RecurrentLatentWorldModelEnsemble" tests/` matches only the module's own
dedicated test file), so I do not believe it would have caught this
regardless; noting the gap rather than skipping it silently. `import
alberta_framework.core.prototype_agent` succeeds cleanly with this diff.

## Scope

`RecurrentLatentWorldModelEnsemble` is `EVIDENCE_LEVEL = "L0"`,
`SCIENTIFIC_PROMOTION_ALLOWED = False`, and is not a registered evidence
source (not referenced in `docs/status.md`,
`docs/evidence/negative-results.md`, or `evaluation/evidence_manifest.py`).
No benchmark lane, seed schedule, or `outputs/` artifact is touched by
this change.

- [x] logs: inline above (full commands + real output; no attachment host available in this environment)

## Maintainer ownership repair

At `ae408df40bfe59e95561f158eb6627e8be330672`, retry-cache recovery is restricted to authoritative, in-domain, off-boundary transitions that passed exact cached-prediction validation and reached only a late numerical/candidate rejection. Stale owners, mismatched inputs/actions, invalid next observations, boundary-semantic failures, tampered cached predictions, and rejected boundary resets all return an invalid cache. The expanded focused suite passes 26 tests; repository-wide Ruff and strict mypy pass. No outputs were touched.

---

— [rama0x1-asi-claude-worker]

### DIFF
--- a/alberta_framework/core/recurrent_latent_world_model_ensemble.py
+++ b/alberta_framework/core/recurrent_latent_world_model_ensemble.py
@@ -226,15 +226,27 @@ class RecurrentLatentWorldModelEnsembleConfig:
     """Fixed architecture, optimizer, warmup, and numeric bounds.
 
     The ``max_*`` fields are fail-closed validity gates, not tuned
-    hyperparameters: an observation, reward, parameter, prediction, loss, or
-    raw gradient whose magnitude exceeds its bound causes the whole event to
-    be rejected with persistent state untouched.  Their internal consistency
-    is validated at construction: ``variance_floor < max_variance``,
-    ``gradient_clip_norm <= max_raw_gradient_norm``, and
-    ``learning_rate * gradient_clip_norm <= max_parameter_magnitude`` (so a
-    single clipped step cannot itself breach the parameter bound).  The
-    default magnitudes are coarse sanity ceilings far above any healthy
-    operating range, not derived quantities.
+    hyperparameters: an observation, reward, parameter, prediction, or loss
+    whose magnitude exceeds its bound causes the whole event to be rejected
+    with persistent state untouched.  Their internal consistency is validated
+    at construction: ``variance_floor < max_variance``, ``gradient_clip_norm
+    <= max_raw_gradient_norm``, and ``learning_rate * gradient_clip_norm <=
+    max_parameter_magnitude`` (so a single clipped step cannot itself breach
+    the parameter bound).  The default magnitudes are coarse sanity ceilings
+    far above any healthy operating range, not derived quantities.
+
+    Per-member raw NLL gradients are validated for finiteness only, not
+    against ``max_raw_gradient_norm``: the reachable raw gradient scales as
+    ``residual / variance``, so a well-trained low-noise member can legally
+    exceed any fixed ceiling on an ordinary in-bounds transition, and because
+    a rejection leaves state unchanged, a norm ceiling here creates a
+    deadlock rather than a safety gate.  ``gradient_clip_norm`` already
+    bounds the committed step magnitude, and ``candidate_parameters_valid`` /
+    ``candidate_state_valid`` independently confirm the post-clip candidate is
+    finite and in-bounds before it is committed.  ``max_raw_gradient_norm``
+    still bounds the representation gradient (a returned diagnostic signal,
+    never itself clipped) and constrains ``gradient_clip_norm`` at
+    construction.
     """
 
     observation_dim: int
@@ -1054,7 +1066,40 @@ class RecurrentLatentWorldModelEnsemble:
         self,
         state: RecurrentLatentWorldModelEnsembleState,
         diagnostics: RecurrentLatentWorldModelDiagnostics,
+        next_decision_observation: Array | None = None,
     ) -> RecurrentLatentWorldModelUpdateResult:
+        # A rejection never mutates `state`, but the *cache* previously
+        # defaulted to the invalid zero cache unconditionally.  Since `decide`
+        # and `update` both require a valid, owned start cache, that made
+        # every rejection -- regardless of cause -- a permanent deadlock: no
+        # later event, however legal, could ever be accepted again.
+        #
+        # When the incoming state and decision cache were themselves sound
+        # (`state_valid & cache_valid`), we know `state` is a legitimate
+        # anchor and `next_decision_observation` was at least well-typed (the
+        # shape/dtype contract in `update` runs unconditionally before any
+        # validity gate). Re-owning `state` with that observation makes the
+        # rejection retryable: the very next `decide`/`update` call
+        # independently re-validates the observation and every other input,
+        # so nothing unsound is smuggled through -- it only stops a single
+        # bad or over-strict event from freezing every event after it.
+        if next_decision_observation is None:
+            recoverable_cache = self._zero_start_cache()
+        else:
+            recoverable = diagnostics.state_valid & diagnostics.cache_valid
+            recoverable_cache = cast(
+                RecurrentLatentStartCache,
+                jax.lax.cond(
+                    recoverable,
+                    lambda: RecurrentLatentStartCache(
+                        owner_event_count=state.event_count,
+                        owner_hidden_states=state.member_hidden_states,
+                        observation=next_decision_observation,
+                        valid=jnp.asarray(True, dtype=jnp.bool_),
+                    ),
+                    self._zero_start_cache,
+                ),
+            )
         return RecurrentLatentWorldModelUpdateResult(
             state=state,
             prediction=self._zero_prediction(),
@@ -1072,7 +1117,7 @@ class RecurrentLatentWorldModelEnsemble:
             member_updates_applied=jnp.zeros(
                 (self._config.ensemble_size,), dtype=jnp.bool_
             ),
-            next_start_cache=self._zero_start_cache(),
+            next_start_cache=recoverable_cache,
             diagnostics=diagnostics,
         )
 
@@ -1217,10 +1262,14 @@ class RecurrentLatentWorldModelEnsemble:
                 member_losses.append(loss)
                 member_gradients.append(gradient)
                 gradient_norms.append(gradient_norm)
+                # Finiteness only -- see the class docstring.  The raw norm is
+                # not compared against max_raw_gradient_norm: a well-trained
+                # low-noise member legally produces residual / variance
+                # gradients that dwarf any fixed ceiling on an ordinary
+                # in-bounds transition, and gradient_clip_norm (applied below)
+                # already bounds the committed step regardless of raw scale.
                 member_gradients_valid.append(
-                    _tree_all_finite(gradient)
-                    & jnp.isfinite(gradient_norm)
-                    & (gradient_norm <= cfg.max_raw_gradient_norm)
+                    _tree_all_finite(gradient) & jnp.isfinite(gradient_norm)
                 )
             losses = jnp.stack(member_losses)
             gradient_norm_array = jnp.stack(gradient_norms)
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking fail-closed ownership regression at exact head 891e6bc. _rejected_result marks a returned start cache valid whenever state_valid & cache_valid, even when the decision cache does not own this state/observation/action and the transition is therefore not authoritative.

Concrete public repro: take a valid decision, change only owner_event_count by +1 (prediction/valid flags remain true), and call update with next_decision_observation=[7,-8]. Diagnostics correctly report cache_valid=True, ownership_valid=False, applied=False, but the PR returns next_start_cache.valid=True bound to [7,-8]; model.decide(state, that cache, action).valid is then True. This launders an arbitrary transition observation into a newly owned cache through the very stale/tampered-cache path the module promises to reject.

Recovery may be appropriate after a fully authoritative input reaches a later numerical/candidate rejection, but it must require the pre-update authority/input/boundary gates (at least state_valid, cache_valid, ownership_valid, input_valid, boundary_semantics_valid), and never emit valid=True for a nonfinite/out-of-bounds next observation. Add stale owner, mismatched action/observation, invalid next observation, and genuine late numeric-rejection recovery cases. The member-gradient finiteness-only change itself looks directionally sound; do not merge this head until recovery cannot mint ownership from a rejected unauthoritative event.

lalalune (approved):
Re-reviewed exact head ae408df40bfe59e95561f158eb6627e8be330672 after the recovery-authority repair. The previous ownership-laundering path is closed: recovery now requires valid state/cache/input/ownership/boundary/capacity, exact cached prediction, a valid prediction, and an off-boundary failure. Independent exact-head validation: 26 focused tests, Ruff, mypy, plus 7 current-main composition regressions; CI is 20/20 and the merge tree is clean.

PR #354, opened 2026-08-16
fix(statistics): reject non-finite samples in every comparison test

Prabhat1308:
Closes #353. Stacked on #338 (`fix/statistics-sample-vectors`): the first commit here is #338's; only `4a3d69b` is new. Rebases to a one-commit diff once #338 lands.

## Issue

`ttest_comparison`, `mann_whitney_comparison`, `wilcoxon_comparison`, and `cohens_d` forwarded NaN/inf samples to SciPy and published `statistic=nan, p_value=nan, significant=False`, so a poisoned seed read as an honest non-result on the surface that decides A/B verdicts, while every summary helper already raised.

## New state

Each of the four functions requires finite samples right after the sample-vector check (`values_a must be finite` / `values_b must be finite`), so `pairwise_comparisons` now raises on a poisoned seed instead of returning `p=nan`. Finite inputs are unchanged.

Failing-test-first: `TestComparisonsRejectNonFiniteSamples` (15 cases across the four functions × nan/±inf, under `-W error`) and `test_pairwise_comparisons_rejects_a_poisoned_seed` fail on `main` and pass here.

## Evidence

Falsifier from #353 at this head:

```text
ValueError: values_a must be finite
```

Verification at `4a3d69b6478ecbd7628d013cdb14a8990c5a3dda`:

```text
.venv/bin/python -m pytest tests/test_statistics_validation.py tests/test_experiments_validation.py tests/test_utils_smoke.py tests/test_forager_matched_statistics.py -q -o addopts=""   -> 219 passed
.venv/bin/python -m ruff check .                                                              -> All checks passed!
.venv/bin/python -m mypy                                                                      -> Success: no issues found in 164 source files
```

`utils/statistics.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, ruff, mypy, falsifier output at the evidence head)

Provider/model/client: anthropic / claude-fable-5 / Claude Code 2.1.233 (contribute-to-asi skill revision 72e4239e).

PR #410, opened 2026-08-16
fix(statistics): validate alpha and p-value contracts

lalalune:
Follow-up hardening for #354.

The comparison helpers accepted non-finite or out-of-domain alpha values (alpha=inf made every finite p-value significant), correction helpers accepted invalid p-values, and finite constant samples could publish p=nan. This change validates alpha in (0, 1), validates every p-value in [0, 1], rejects non-finite scipy results, and gives the all-tied Mann-Whitney null its exact U midpoint and p=1 result. The runtime type guard uses the actual type MRO, so a plain object spoofing __class__ as float cannot enter through isinstance.

Failing-first: TestProbabilityContracts produced 73 failures and 2 passes on the #354 implementation; the later __class__ spoof regression independently failed all 8 affected routes before the exact-type guard.

Verification at feb56a18153daaab3645f1410251e2ca6682b8e7:

- 222 statistics validation tests passed under two xdist workers; this explicitly verifies deterministic collection after adding a stable repr for the adversarial spoof parameter. The prior 289-test affected panel also passed (statistics validation, matched statistics, utils smoke).
- Ruff passed on the changed source and tests.
- Strict mypy passed on statistics.py.
- outputs/ is untouched.

This is validation hardening only; it does not create or promote scientific evidence.

### DIFF
--- a/alberta_framework/utils/statistics.py
+++ b/alberta_framework/utils/statistics.py
@@ -4,8 +4,10 @@ Provides functions for computing confidence intervals, significance tests,
 effect sizes, and multiple comparison corrections.
 """
 
+import math
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, NamedTuple
+from numbers import Real
+from typing import TYPE_CHECKING, NamedTuple, cast
 
 import numpy as np
 from numpy.typing import NDArray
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking review finding at exact head `79938abe49d1d3a5fbc055f7129440fcea400469`: `_require_probability` checks the domain only after `float(value)`, so concrete `Real` values can round onto legal endpoints. Both `bonferroni_correction([Fraction(-1, 2**2000)], alpha=.05)` and `holm_correction(...)` accept the negative p-value (it becomes `-0.0` and is marked significant); `Fraction((2**2000)+1, 2**2000)` is likewise accepted as p=1. Conversely, valid strict alpha values extremely near 0 or 1 can round to endpoints and be rejected. Please compare the original exact-Real domain (or exact ratio metadata) before canonical conversion, add below/tie/above endpoint regressions for p-values and alpha, and retain the type-spoof/nonfinite protections. I am holding merge until this is repaired and independently rechecked.

PR #411, opened 2026-08-16
fix(actor-critic): reject inverted or non-finite continuous action bounds

Prabhat1308:
Hardens ContinuousActorCritic action bounds before any JAX clipping or policy execution.

Changes:
- require actual non-bool Real values or None
- round accepted values directly into the float32 execution domain and reject non-finite narrowing
- canonicalize non-builtin reals and exact integers to JSON-safe sink values
- reject inverted low/high bounds after canonicalization
- normalize all ordinary conversion-hook failures to ValueError

Validation at exact head 1ae04f5:
- continuous actor-critic suite: 28 passed under RuntimeWarning-as-error
- Ruff: clean
- strict mypy for actor_critic.py: clean
- git diff --check: clean

### DIFF
--- a/alberta_framework/core/actor_critic.py
+++ b/alberta_framework/core/actor_critic.py
@@ -816,8 +816,9 @@ class ContinuousActorCriticAgent:
                 raise ValueError(f"{name} must be finite when set") from exc
             if not math.isfinite(narrowed):
                 raise ValueError(f"{name} must remain finite once narrowed to float32")
-            preserve_builtin = type(bound) is int or type(bound) is float
-            canonical_bounds[name] = float(bound) if preserve_builtin else narrowed
+            # Only an actual built-in float is already the binary64 payload JAX will
+            # narrow exactly; ints and other reals store the validated binary32 value.
+            canonical_bounds[name] = bound if type(bound) is float else narrowed
         low = canonical_bounds["action_low"]
         high = canonical_bounds["action_high"]
         if low is not None and high is not None and low > high:
--- a/alberta_framework/core/actor_critic.py
+++ b/alberta_framework/core/actor_critic.py
@@ -799,11 +799,14 @@ class ContinuousActorCriticAgent:
             raise ValueError("action_dim must be positive")
         if config.log_sigma_min > config.log_sigma_max:
             raise ValueError("log_sigma_min must be <= log_sigma_max")
+        canonical_bounds: dict[str, float | None] = {}
         for name in ("action_low", "action_high"):
             bound = getattr(config, name)
             if bound is None:
+                canonical_bounds[name] = None
                 continue
-            if isinstance(bound, bool) or not isinstance(bound, Real):
+            actual_type = type(bound)
+            if issubclass(actual_type, bool) or not issubclass(actual_type, Real):
                 raise ValueError(f"{name} must be a finite real number when set")
             if not math.isfinite(bound):
                 raise ValueError(f"{name} must be finite when set")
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking float32-sink gap at exact head 1ca3b21663de48ff6a26f33d51f8c65983b0efbe: the new check only applies math.isfinite to the wide host value. ContinuousActorCriticConfig(action_dim=1, action_low=1e100) is accepted, but select_action narrows the bound to float32 +inf and returns action=[inf]; action_high=-1e100 analogously returns [-inf] (with an overflow warning). This still permits the non-finite constant-action policy the issue is meant to close. Please validate the actual float32 sink (prefer the shared exact float32 helper), reject finite-to-inf narrowing, and add both overflow directions plus an accepted max-finite endpoint regression before construction can succeed.

Prabhat1308:
Addressed in `2464b8b`: each bound is now narrowed with the shared `round_real_to_float32` converter and finite-to-infinity narrowing is rejected (`action_low must remain finite once narrowed to float32`), so `action_low=1e100`, `action_high=-1e100`, and `±3.5e38` fail at construction; the float32 max endpoint (`±3.4028235e38`) is accepted and its sampled action asserted finite. Regressions run under `-W error` (no overflow warnings). Body updated to the new evidence head; 34 passed, ruff/mypy clean. Re-review requested.

lalalune (changes requested):
Blocking runtime-type gap at exact head 2464b8bd2cce88dc89b608d319f232d1f32daa7b: the float32 overflow repair is directionally correct, but its `isinstance(bound, Real)` gate is vulnerable to Python `__class__` spoofing. A plain object whose `__class__` property returns `float`, whose `as_integer_ratio()` returns `(0, 1)`, and whose comparison hooks return False is accepted by `ContinuousActorCriticAgent`; the arbitrary object is retained in config and later reaches `jnp.clip`. This is the same exact-type boundary fixed in #390. Please compose with the hardened `_float32` helper or use `type(value)`/actual-MRO validation here, and add a class-spoof regression. The new overflow directions and max-finite endpoint tests otherwise close the prior float32-sink blocker.

Prabhat1308:
Addressed in `4a63bca`: the type gate now uses the actual MRO (`issubclass(type(bound), Real)` + bool exclusion) so a `__class__`-spoofing object is rejected (`test_continuous_actor_critic_rejects_bounds_that_only_spoof_float`), and accepted bounds are canonicalized to primitive floats in the stored config (`test_continuous_actor_critic_canonicalizes_real_bounds_to_float`: Fraction / np.float64 → float, sampled action inside the bounds), so nothing but a float reaches `jnp.clip`. I did not depend on #390's helper since it is not merged yet; happy to switch to `finite_real_and_float32` once it lands. Body updated to the new evidence head; 37 passed under `-W error`, ruff/mypy clean. Re-review requested.

lalalune:
Fresh-head blocker at `4a63bca7c1959f11a341b77432affbb512703637`: the float32 validator is correct, but the storage branch still treats every built-in `int` like an already-safe binary64 float. For `bound=(2**25-1)*2**103-1`, `round_real_to_float32(bound)` is finite max-float, so construction accepts it; then `float(bound)` is stored and JAX double-rounds that payload to `+inf`, making `start()` return `action=[inf]`. The negative value as `action_high` returns `[-inf]`. A smaller exact midpoint integer (`2**64+2**40+1`) also stores the wrong lower float32 neighbor. Please preserve only actual built-in floats; canonicalize built-in ints and all other non-float Reals to the already validated exact binary32 `narrowed` value, and add both below-overflow signs plus an ordinary midpoint integer regression under `-W error`.

Prabhat1308:
Addressed in `8baa14a`: only an actual built-in `float` is preserved as the stored payload; built-in ints and every other real now store the validated exact binary32 `narrowed` value. Regressions under `-W error`: `(2**25-1)*2**103-1` as `action_low`, its negation as `action_high` (both store the finite float32 max magnitude and `start()` samples a finite action), and the `2**64 + 2**40 + 1` midpoint integer (stored value == `round_real_to_float32(value)`). Body updated; 40 passed, ruff/mypy clean. Re-review requested.

lalalune (approved):
Exact repaired head 1ae04f5 validates/canonicalizes action bounds in the float32 sink domain, normalizes conversion failures, and passes the full continuous actor-critic suite plus static gates.

lalalune (approved):

PR #414, opened 2026-08-16
fix(rule-discovery): reject non-finite genomes before evaluation

Svector-anu:
Closes #413.

## What changed

`evaluate_population` validated the genome matrix's shape but never its values. A `NaN` (or `inf`/`-inf`) gene — e.g. an invalid learning-rate gene — entered the JAX screening loop and returned a plausible-looking finite accuracy score instead of being rejected, silently allowing an invalid candidate to be ranked alongside genuine ones.

confirmed directly before touching the source:

```text
before
result: array([0.09775001])
is_finite: [ True]
```
(a `(1, GENOME_SIZE)` float32 matrix with the `lr` gene set to `NaN`, evaluated on `MICRO_SUITE["M1"]`, seed 0, batch size 1 — returned a finite score instead of raising.)

fix: after dtype conversion and shape validation, reject any non-finite genome value before materializing an evaluation stream, matching the existing preflight-validation style already used for shape and batch-size checks in this same function.

## Reachability

`evaluate_population` is called by `evaluate_suite`, which is called throughout `run_search`; `main()` exposes `run_search` via `python -m alberta_framework.benchmarks.rule_discovery search ...` — confirmed directly. Not re-exported from either `alberta_framework/__init__.py` or `alberta_framework/benchmarks/__init__.py`, but directly importable with a module CLI entry point — real candidate matrices flow through this path and are not hardcoded-safe at the `evaluate_population` API boundary.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_rule_discovery.py -q -o addopts=""
52 passed, 2 warnings in 85.59s

$ .venv/bin/python -m ruff check alberta_framework/benchmarks/rule_discovery.py tests/test_rule_discovery.py
All checks passed!

$ .venv/bin/python -m mypy
Success: no issues found in 165 source files
```

New test: `test_evaluate_population_rejects_nonfinite_genomes_before_materialization`, parametrized over `nan`, `inf`, `-inf`, using `monkeypatch` to replace `_materialize_eval` with a function that raises `AssertionError` if reached — confirming the guard fires *before* materialization, not just that the call eventually errors somehow.

mutation-resistance verified independently: reverted just the source fix (`git stash push -- alberta_framework/benchmarks/rule_discovery.py`, kept the test), reran — all 3 cases failed with `AssertionError: non-finite genomes reached stream materialization`, proving the non-finite genome actually reached the guarded call. restored the fix, 3/3 green again.

## Evidence

[...]

Issue #413, opened 2026-08-16
evaluate_population accepts non-finite genome values, silently ranking invalid candidates


Svector-anu:
`evaluate_population` in `alberta_framework/benchmarks/rule_discovery.py` validated the genome matrix's shape but never its values. A `NaN` (or `inf`/`-inf`) gene — e.g. an invalid learning-rate gene — entered the JAX screening loop and returned a plausible-looking finite accuracy score instead of being rejected, silently allowing an invalid candidate to be ranked alongside genuine ones.

confirmed directly:
```text
before
result: array([0.09775001])
is_finite: [ True]
```
(a `(1, GENOME_SIZE)` float32 matrix with the `lr` gene set to `NaN`, evaluated on `MICRO_SUITE["M1"]`, seed 0, batch size 1 — returned a finite score instead of raising.)

reachability: `evaluate_population` is called by `evaluate_suite`, which is called throughout `run_search`; `main()` exposes `run_search` via `python -m alberta_framework.benchmarks.rule_discovery search ...`. Not re-exported from either `alberta_framework/__init__.py` or `alberta_framework/benchmarks/__init__.py`, but directly importable with a module CLI entry point — real candidate matrices flow through this path and are not hardcoded-safe at the `evaluate_population` API boundary.

fix incoming: after dtype conversion and shape validation, reject any non-finite genome value before materializing an evaluation stream, matching the existing preflight-validation style already used for shape and batch-size checks in this same function.

PR #415, opened 2026-08-16
fix(experiments): validate extracted parameter coordinates

lalalune:
Follow-up hardening for #358.

A non-injective extractor is now rejected, but NaN coordinates remain non-reflexive dictionary keys (so repeated NaNs evade collision detection), non-finite numeric coordinates can enter a sensitivity result, and unhashable coordinates leak raw TypeError. This change preserves arbitrary well-behaved categorical keys while requiring each coordinate, including tuple/frozenset components, to be hashable, hash-stable, reflexive, and finite when numeric.

Failing-first: 8 adversarial coordinate cases failed against the #358 implementation; the three canonical/collision paths passed.

Verification at 4331e22c6d610421bcf6e0c0a8accfbd6f060b74:

- 49 affected experiment/utils tests passed.
- Ruff passed on the changed source and tests.
- Strict mypy passed on experiments.py.
- outputs/ is untouched.

This is validation hardening only; it does not run a benchmark or create/promote scientific evidence.

### DIFF
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -14,8 +14,10 @@ Two conventions matter for downstream analysis:
   :func:`aggregate_metrics`.
 """
 
-from collections.abc import Callable, Sequence
-from typing import Any, NamedTuple, cast
+import math
+from collections.abc import Callable, Iterable, Sequence
+from numbers import Complex, Number
+from typing import Any, NamedTuple, NoReturn, cast
 
 import jax.random as jr
 import numpy as np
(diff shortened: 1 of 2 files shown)


lalalune:
Maintainer blocker at exact head 4331e22c6d610421bcf6e0c0a8accfbd6f060b74 (independently reproduced):

- A non-finite float subclass can spoof __complex__ and pass validation while remaining an infinite result key.
- Binary64 conversion rejects valid finite, hashable, reflexive coordinates including 10**1000, Fraction(10**1000, 3), Decimal("1e4000"), and finite np.longdouble("1e4000"), contradicting the no-narrowing claim.
- Two adjacent hash calls do not prove stability; a delayed-drift key passes but is immediately unusable in the returned mapping.

Please do not merge this head. I am preparing a failing-test-first repair with a certifiable accepted-key contract. Ordinary focused/adjacent behavior otherwise passed (239 tests, Ruff, strict mypy), and the patch composes cleanly with main.

Traceability: #413 is unrelated and was already closed by #414; this is an unlinked follow-up to #357/#358.

lalalune:
Repair pushed at ae6ab0a3156b322e29750beae4755e11e2298ec4 after an exact-head lease check. The extractor now accepts only intrinsically certifiable immutable key families (including wide int/Fraction/Decimal/NumPy finite numeric and nested tuple/frozenset values) and rejects user-defined classes/subclasses before invoking conversion, equality, or hash hooks.

Red: all six retained blocker cases failed on 4331e22. Green: 6/6 blocker cases; 61 experiment-validation tests; 266 focused/adjacent tests; repository Ruff and strict mypy; clean current-main composition. No outputs or registered evidence sources changed.

This new head still requires fresh independent review and hosted CI before merge.

Issue #416, opened 2026-08-16
Step 12 IA facade float32 overflow hardening and int32 bounds on dimensions and smoke probes


deepanshu-yd:
## Problem

`Step12IAConfig` and `run_step12_smoke` lack float32 overflow validation on scalar fields (`cerebellum_step_size`, `base_step_size`, `base_avg_reward_step_size`, `option_step_size`, `option_gamma`, `epsilon_base`, `utility_ema_decay`), and lack int32 bounds on `n_demons`, `observation_dim`, `n_primitive_actions`, `feature_index`, `max_option_steps`, and `run_step12_smoke` parameters (`steps`, `seed`).

## Reproduction

```python
from alberta_framework.steps.step12 import Step12IAConfig

# Accepts out-of-range float32 overflow
cfg = Step12IAConfig(cerebellum_step_size=1e100)
```

## Proposed Fix

- Validate all float32 execution fields against overflow in `alberta_framework/steps/step12.py`.
- Bound integer dimensions, demon counts, action counts, backup budgets, feature indices, option steps, and smoke parameters to signed int32 bounds (`2**31 - 1`).
- Add comprehensive test suite in `tests/test_step12_production.py`.

lalalune:
Addressed by open PR #417 (missing a closing keyword there; noting the link here).

lalalune:
Closed as completed on current main by recovered integration commit 7bb30c9 (original #417 ancestry retained there). The facade now enforces float32 sink validity and signed-int32 bounds for config/smoke inputs. Revalidated: 193 Step 12 production tests passed; scoped ruff and strict mypy passed. Replacement #628 became redundant after that concurrent integration and was closed.

PR #417, opened 2026-08-16
fix(step12): harden Prototype facade validation

deepanshu-yd:
Reconciles the Step 12 Prototype facade with the shared exact float32 validator and current core contracts.

Changes:
- reuse the shared finite-real/float32 validation policy instead of a duplicate converter
- enforce exact trusted integer families and int32 bounds before JAX shapes/counters
- keep option planning backups below INT32_MAX, matching STOMPConfig
- require exact tuple/SubtaskSpec structure and reject conversion-hook integer subclasses
- retain fail-closed smoke health checks

Validation at exact head 717d686d:
- tests/test_step12_production.py: 192 passed
- post-rebase adversarial/static panel: 32 passed, 160 deselected
- Ruff: clean
- strict mypy for step12.py: clean
- git diff --check: clean

This is library/contract hardening only. No benchmark campaign was run and no outputs or evidence artifacts were modified.

### DIFF
--- a/alberta_framework/steps/step12.py
+++ b/alberta_framework/steps/step12.py
@@ -29,9 +29,8 @@ References:
 
 from __future__ import annotations
 
-import math
 from dataclasses import dataclass
-from numbers import Integral, Real
+from numbers import Integral
 from typing import Any, cast
 
 import jax.numpy as jnp
@@ -39,7 +38,6 @@ import jax.random as jr
 import numpy as np
 from jax import Array
 
-from alberta_framework._float32 import round_real_to_float32_with_ratio
 from alberta_framework.core.intelligence_amplification import (
     ExoCerebellumConfig,
     IAAgent,
@@ -55,6 +53,9 @@ from alberta_framework.core.intelligence_amplification import (
 )
 from alberta_framework.core.oak import OaKConfig
 from alberta_framework.core.options import STOMPConfig, SubtaskSpec
+from alberta_framework.steps._float32_validation import (
+    finite_real_and_float32,
+)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking at exact `f7472d59f743b4b93ac725e5879bc5bdd8a8205c`: the new validation still checks the ordinary host comparison before narrowing but never checks the exact ratio domain. A float subclass with payload `0.5` and exact ratio `(2**200+1, 2**200)` is accepted by a closed-unit field and narrows to `1.0`; ratio `(-1, 2**200)` escapes nonnegative fields as `-0.0`. `_require_real`/`_require_int` also use spoofable `isinstance`, so non-Real objects can report `__class__ -> float/int`. This is the causal family issue #416 explicitly asks to close. Please rebase on repaired PR #390 (`233ecb1f...` pending) and use its actual-type/exact-ratio helper metadata for host-domain + sink-domain checks, then add below/tie/above endpoint, hidden negative, class-spoof, and ratio-component regressions. The float32 overflow/int32 work otherwise looks directionally correct.

deepanshu-yd:
Updated to address review:

- Upgraded shared `round_real_to_float32_with_ratio` helper to use actual-type dispatch (`issubclass(actual_type, Integral)`), preventing `__class__` spoofing from bypassing `as_integer_ratio`.
- Validated both host real value and exact `(numerator, denominator)` ratio against unit/non-negative Step 12 facade domains.
- Added strict type checking for ratio tuples and ratio integer components to prevent non-Integral components from spoofing integers.
- Added regression tests for hidden negative/above-domain ratio floats, class-spoofed floats reporting `int`, and non-Integral ratio components in `tests/test_step12_production.py`.

lalalune (approved):
Exact repaired head 717d686 reviewed and validated: 192 Step 12 tests plus 32 post-rebase adversarial/static selections pass; shared float32 policy, integer bounds, tuple contracts, and STOMP backup bound are aligned.

lalalune (approved):

lalalune:
The reviewed head 717d686 of this PR was merged and is reachable from main's history, so the substance (exact-ratio validators, actual-type checks for the Step 12 facade) has landed with commit credit. The current force-pushed rebase now conflicts with semantics frozen on main afterwards (positive pseudo_reward_scale, canonical scalar storage). Closing as merged-in-substance — thank you!

lalalune:
Current-main replacement is #626. It preserves this PR’s reviewed facade hardening while resolving the conflict in favor of the newer strictly-positive `pseudo_reward_scale` contract. No merge/closure performed here.

PR #388, opened 2026-08-16
fix(world-model): build default discounts in float32 regardless of reward dtype

Prabhat1308:
Closes #386.

## Issue

Both world-model learning loops derived the default discount vector with `jnp.full_like(rewards, gamma)`, which inherits the reward dtype: `int32` rewards truncated `gamma` to `0` and `bool` rewards to `True`, training the discount head on the wrong constant while every update was accepted (`discount_errors ≈ 0` instead of `≈ 0.99`).

## New state

The default is built with `jnp.full(jnp.shape(rewards), gamma, dtype=jnp.float32)` in both loops; integer, bool, and float reward arrays now produce identical discount targets, predictions, and errors. Explicitly passed `discounts` are unchanged.

Failing-test-first: `test_default_discounts_do_not_inherit_the_reward_dtype` (action-conditioned; float/int32/bool vs an explicit `full(gamma)` run) and `test_latent_default_discounts_do_not_inherit_the_reward_dtype` fail on `main` and pass here.

## Evidence

Falsifier from #386 at this head:

```text
ACWM int32 rewards: discount targets = [0.99 0.99 0.99 0.99] | updates_applied all = True
LWM  int32 rewards: discount predictions = [0.7348 0.6286 0.7645] (== float32 run)
```

Verification at `486a46c0c8cfbd60c828965ac316639f689add4a`:

```text
.venv/bin/python -m pytest tests/test_action_conditioned_world_model.py tests/test_latent_world_model.py tests/test_world_model*.py -q -o addopts=""   -> 66 passed
.venv/bin/python -m ruff check .                                                                                                    -> All checks passed!
.venv/bin/python -m mypy                                                                                                            -> Success: no issues found in 164 source files
```

Neither module is a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, ruff, mypy, falsifier output at the evidence head)

Provider/model/client: anthropic / claude-fable-5 / Claude Code 2.1.233 (contribute-to-asi skill revision 72e4239e).

PR #420, opened 2026-08-16
fix(world-model): fail closed on non-finite config scalars and out-of-range discounts

Prabhat1308:
Hardens ActionConditionedWorldModel and LatentWorldModel configuration and update boundaries.

Changes:
- validate float32-consumed config scalars in both the exact host domain and binary32 sink domain
- canonicalize accepted non-builtin Real inputs to JSON-safe floats
- reject non-finite or out-of-range discounts atomically before learning
- compare exact integer ratios instead of trusting user-defined comparison hooks
- normalize all ordinary exact-conversion failures to ValueError

Validation at exact repaired head 459b0e4:
- action-conditioned + latent world-model suites: 71 passed under RuntimeWarning-as-error
- Ruff: clean
- strict mypy for all three changed source files: clean
- git diff --check: clean

### DIFF
--- a/alberta_framework/core/_float32_scalars.py
+++ b/alberta_framework/core/_float32_scalars.py
@@ -1,0 +1,82 @@
+"""Shared validation for configuration scalars that are consumed as float32.
+
+A configuration scalar is checked in both domains that matter: the exact host
+value (so a lying ``as_integer_ratio`` or a value that only *rounds* into
+range cannot pass) and its binary32 rounding (so a host-finite value that
+narrows to infinity, zero, or the excluded end of a half-open interval is
+refused before it can freeze an EMA or divide by zero at the sink).  Only an
+actual built-in ``float`` is stored as-is — JAX narrows it once, exactly as
+validated here — while ints and other reals are stored as the validated
+binary32 value.
+"""
+
+from __future__ import annotations
+
+import math
+from numbers import Real
+from typing import Any, cast
+
+from alberta_framework._float32 import round_real_to_float32
+
+
+def validated_float32_scalar(
+    name: str,
+    value: object,
+    *,
+    positive: bool = False,
+    lower: float | None = None,
+    upper: float | None = None,
+    upper_inclusive: bool = True,
+) -> float:
+    """Return the canonical stored value of one float32-consumed scalar or fail closed.
+
+    Raises:
+        ValueError: If ``value`` is not an actual non-bool real, does not narrow
+            to a finite binary32, or leaves the declared domain either as the
+            exact host value or once narrowed to binary32.
+    """
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Real):
+        raise ValueError(f"{name} must be a finite real number")
+    real = cast(Any, value)
+    try:
+        narrowed = round_real_to_float32(real)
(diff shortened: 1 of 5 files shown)


lalalune (changes requested):
Blocking float32/exact-domain gap at exact head f8cf505dd2b5682033f0b13391b30df277482c3e. `_require_finite_scalar` validates only `float(value)`, while these fields are consumed as float32 and the original object is retained. For example, ActionConditionedWorldModelConfig(reward_scale=1e100) and max_delta_scale=1e100 (and the analogous latent scales) pass because the Python float is finite, then narrow to `inf` in JAX; reward_scale makes targets collapse and decoded diagnostics/refusal non-finite. Exact Real values just outside a closed domain can also round onto `-0.0`/`1.0`, and `isinstance` remains vulnerable to `__class__` spoofing. Please use the shared exact-ratio conversion/actual-type gate, validate both exact host domain and finite float32 sink, canonicalize non-builtin values so the validated sink is what JAX receives, and add finite-to-inf, Fraction endpoint-straddle, and class-spoof regressions across both models.

lalalune (changes requested):
Blocking float32/config-boundary gaps at exact head f8cf505dd2b5682033f0b13391b30df277482c3e. `_require_finite_scalar` validates only `float(value)` and uses `isinstance(value, Real)`: (1) a plain object can spoof `__class__ == float` and be accepted into config; (2) host-finite `reward_scale=1e100` / positive scales around `1e-100` pass but narrow to float32 inf/zero at the actual multiply/divide sink; and (3) legal-host decays just below one (for example `1 - 1e-10`) round to float32 1.0, violating the documented `[0,1)` execution domain and freezing the EMA. The same issues span both duplicated validators and observation/encoder/update bounds. Please rebase onto now-merged #388, use the shared exact-ratio/actual-type float32 validation from #390 once available (do not keep two local converters), validate both exact host and narrowed sink domains, canonicalize non-builtin Real payloads before JAX consumption, and add class-spoof, overflow, positive-underflow, and sub-one-rounds-to-one regressions for both world-model configurations. The new transition discount [0,1] gate and outer atomic rollback are directionally sound.

Prabhat1308:
Addressed in `9c00b6e` (rebased onto main with #388 in it): the two local converters are gone; both validators use one shared `core/_float32_scalars.py::validated_float32_scalar` — actual-class gate (spoof rejected), exact binary32 narrowing via `_float32.round_real_to_float32`, domain enforced on both the exact host value and the narrowed sink (`reward_scale=1e100` overflow, `1e-100` positive underflow, and decays `1-1e-10` that round to 1.0 are all refused with distinct messages), and canonical storage preserving only built-in floats; both `_validate_config` methods return the canonical config which the constructors store. Regressions for both configurations under `-W error` (spoof / overflow / positive-underflow / sub-one-rounds-to-one / canonicalization round trip). Once #390 lands the helper can switch to `round_real_to_float32_with_ratio` in one place. 108 passed, ruff/mypy clean; body updated. Re-review requested.

lalalune (approved):
Exact repaired head 459b0e4 reviewed and validated: 71 world-model tests pass under RuntimeWarning-as-error; exact-ratio host/sink domain checks and normalized conversion failures are covered; Ruff and strict mypy are clean.

PR #424, opened 2026-08-16
fix(forager-results): validate npz-import decay and window params

### LINKED ISSUE #423: import_official_foragax_npz accepts non-finite ewm_decay, silently poisoning the imported reward curve
`import_official_foragax_npz` in `alberta_framework/benchmarks/forager_results.py` accepted `ewm_decay`, `record_every`, and `final_window` with no validation at all before using them. `ewm_decay` in particular flows directly into an IIR-filter coefficient array (`np.asarray([1.0, -ewm_decay], ...)`) and a division denominator (`(1.0 - ewm_decay**steps) / (1.0 - ewm_decay)`) used to compute the exponentially-weighted-moving-average reward curve for an imported official benchmark run — a `NaN` value silently poisons the entire resulting curve instead of raising a clear input error.

reachability: `import_official_foragax_npz` is exported from `alberta_framework.benchmarks.__init__`'s `__all__` and called from two real sites in `alberta_framework/forager_cli.py`. One of those (line ~776) passes `ewm_decay=protocol.ewm_decay`, sourced from an externally-supplied `--reference-manifest` file, not a hardcoded-safe value — a malformed or adversarial manifest could reach this with a non-finite `ewm_decay`.

fix incoming: validate `ewm_decay` is a finite `int`/`float` (not `bool`) in `[0, 1)`, and `record_every`/`final_window` are positive built-in `int`s, before using any of them.

Svector-anu:
Closes #423.

## What changed

`import_official_foragax_npz` accepted `ewm_decay`, `record_every`, and `final_window` with no validation at all before using them. `ewm_decay` flows directly into an IIR-filter coefficient array (`np.asarray([1.0, -ewm_decay], ...)`) and a division denominator (`(1.0 - ewm_decay**steps) / (1.0 - ewm_decay)`) used to compute the exponentially-weighted-moving-average reward curve for an imported official benchmark run — a `NaN` (or out-of-range) value silently poisons the entire resulting curve instead of raising a clear input error.

fix: validate `ewm_decay` is a finite `int`/`float` (not `bool`) in `[0, 1)`, and `record_every`/`final_window` are positive built-in `int`s, before any of them are used.

## Reachability

`import_official_foragax_npz` is exported from `alberta_framework.benchmarks.__init__`'s `__all__` and called from two real sites in `alberta_framework/forager_cli.py`. One of those (`~line 776`) passes `ewm_decay=protocol.ewm_decay`, sourced from an externally-supplied `--reference-manifest` file — confirmed directly by reading the surrounding manifest-loading code — not a hardcoded-safe value. A malformed or adversarial manifest could reach this with a non-finite `ewm_decay`.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_forager_benchmark.py -q -o addopts="" -k test_official_npz_import
2 passed, 83 deselected in 4.35s

$ .venv/bin/python -m ruff check alberta_framework/benchmarks/forager_results.py tests/test_forager_benchmark.py
All checks passed!

$ .venv/bin/python -m mypy
```

New test: `test_official_npz_import_rejects_nonfinite_ewm_decay`, asserting `ValueError` matching `"ewm_decay must be a finite number"` when `ewm_decay=math.nan`.

mutation-resistance verified independently: reverted just the source fix (`git stash push -- alberta_framework/benchmarks/forager_results.py`, kept the test), reran — failed with `DID NOT RAISE ValueError`, confirming a NaN `ewm_decay` previously flowed through unvalidated. restored the fix, green again.

## Evidence

ae8e18b6c3d3124d3d7346ae1c30ebfb0a12b003

```text
$ .venv/bin/python -m pytest tests/test_forager_benchmark.py -q -o addopts="" -k test_official_npz_import_rejects_nonfinite_ewm_decay
FAILED - Failed: DID NOT RAISE ValueError
1 failed, 84 deselected in 4.90s

green (fix restored):
$ .venv/bin/python -m pytest tests/test_forager_benchmark.py -q -o addopts="" -k test_official_npz_import_rejects_nonfinite_ewm_decay
1 passed, 84 deselected in 3.69s
```

```text
ewm_decay usage confirmed directly in source:
  np.asarray([1.0, -ewm_decay], dtype=np.float64)          # IIR filter coefficient
  (1.0 - ewm_decay**steps) / (1.0 - ewm_decay)              # EWM denominator
a NaN ewm_decay poisons both, propagating NaN through the entire imported reward curve
rather than failing with a clear input-validation error.
```
independently confirmed the reachability chain (`alberta_framework.benchmarks.__init__` export → `forager_cli.py` → `protocol.ewm_decay` sourced from an external `--reference-manifest` file) directly in source before writing the reachability claim; did my own revert-and-confirm-red mutation check.

## Provenance

AI provider/model: openai / gpt-5.6-sol (fix, tests, via AgentRouter proxy — this session's connection dropped mid-run on an AgentRouter quota error before it could write its own report, so all verification below was done independently from the raw diff); anthropic / claude-sonnet-5 (independent re-verification: reran the targeted suite, ruff, and mypy myself, traced the `ewm_decay` usage and full reachability chain directly in source, did my own revert-and-confirm-red mutation check, opened the governing issue, pushed, opened this PR)

### DIFF
--- a/alberta_framework/benchmarks/forager_results.py
+++ b/alberta_framework/benchmarks/forager_results.py
@@ -1567,12 +1567,13 @@ def import_official_foragax_npz(
     final_window: int = 100_000,
 ) -> ForagerRunResult:
     """Convert one official ``data/<seed>.npz`` archive to Alberta's schema."""
-    if (
-        isinstance(ewm_decay, bool)
-        or not isinstance(ewm_decay, (int, float))
-        or not math.isfinite(ewm_decay)
-        or not 0.0 <= ewm_decay < 1.0
-    ):
+    if type(ewm_decay) not in (int, float):
+        raise ValueError("ewm_decay must be a finite number in [0, 1)")
+    try:
+        ewm_decay_as_float = float(ewm_decay)
+    except OverflowError as exc:
+        raise ValueError("ewm_decay must be a finite number in [0, 1)") from exc
+    if not math.isfinite(ewm_decay_as_float) or not 0.0 <= ewm_decay_as_float < 1.0:
         raise ValueError("ewm_decay must be a finite number in [0, 1)")
     if type(record_every) is not int or record_every < 1:
         raise ValueError("record_every must be a positive integer")
--- a/alberta_framework/benchmarks/forager_results.py
+++ b/alberta_framework/benchmarks/forager_results.py
@@ -1567,6 +1567,17 @@ def import_official_foragax_npz(
     final_window: int = 100_000,
 ) -> ForagerRunResult:
     """Convert one official ``data/<seed>.npz`` archive to Alberta's schema."""
+    if (
+        isinstance(ewm_decay, bool)
+        or not isinstance(ewm_decay, (int, float))
+        or not math.isfinite(ewm_decay)
+        or not 0.0 <= ewm_decay < 1.0
+    ):
+        raise ValueError("ewm_decay must be a finite number in [0, 1)")
+    if type(record_every) is not int or record_every < 1:
+        raise ValueError("record_every must be a positive integer")
+    if type(final_window) is not int or final_window < 1:
+        raise ValueError("final_window must be a positive integer")
     protocol_attested = spec.attestation_evidence is not None
     runtime_profile_id: str | None = None
     environment_runtime_profile_sha256: str | None = None
--- a/tests/test_forager_benchmark.py
+++ b/tests/test_forager_benchmark.py
@@ -801,6 +801,69 @@ def test_official_npz_import_rejects_nonfinite_ewm_decay(tmp_path: Path) -> None
         )
 
 
+class _SpoofedFloat:
+    """Mimics ``float`` via ``__class__`` to defeat ``isinstance`` checks."""
+
+    @property
+    def __class__(self) -> type:  # type: ignore[override]
+        return float
+
+    def __float__(self) -> float:
+        return 0.5
+
+    def __le__(self, other: float) -> bool:
+        return 0.5 <= other
+
+    def __lt__(self, other: float) -> bool:
+        return 0.5 < other
+
+
+@pytest.mark.parametrize(
+    ("kwargs", "match"),
+    [
+        ({"ewm_decay": True}, "ewm_decay must be a finite number"),
+        ({"ewm_decay": _SpoofedFloat()}, "ewm_decay must be a finite number"),
+        ({"ewm_decay": 10**10000}, "ewm_decay must be a finite number"),
+        ({"ewm_decay": 1.0}, "ewm_decay must be a finite number"),
+        ({"ewm_decay": -0.1}, "ewm_decay must be a finite number"),
+        ({"ewm_decay": "0.5"}, "ewm_decay must be a finite number"),
+        ({"record_every": True}, "record_every must be a positive integer"),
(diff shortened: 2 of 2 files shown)


lalalune (changes requested):
Blocking facade-validation gap at exact head d5caa25feaba583349138c6972b332e8fe29052b: the issue promises a finite `int`/`float` contract, but `isinstance` accepts an arbitrary object that only spoofs `__class__ = float`; its `__float__`/comparisons can pass and reach the IIR path. A huge built-in integer (for example `10**10000`) also reaches `math.isfinite` and leaks raw `OverflowError` instead of the promised ValueError. The new record_every/final_window branches have no retained tests despite being part of the claimed fix. Please use actual-type checks, perform exact integer range rejection before any float conversion, wrap conversion errors, and add spoof/huge-int plus bool/non-int/nonpositive tests for all three parameters (and finite endpoint controls).

Svector-anu:
fixed both gaps, and confirmed the failure modes empirically before writing the fix.

## what was actually happening

- `isinstance(ewm_decay, (int, float))` is spoofable: an object with a `__class__` property returning `float` passes `isinstance` even though `type(obj) is not float`. confirmed directly in the venv - the old check let it through, then `0.0 <= ewm_decay` blew up with a raw `TypeError` (not `ValueError`) since the actual comparison dispatch doesn't honor the spoofed `__class__`.
- `math.isfinite(ewm_decay)` on a huge built-in int (`10**10000`) converts to float internally and raises `OverflowError` - confirmed directly (`math.isfinite(10**10000)` -> `OverflowError: int too large to convert to float`), uncaught, so it also leaked past the promised `ValueError` contract.

## what changed

- `ewm_decay` check is now `type(ewm_decay) not in (int, float)` - an actual-type check, immune to `__class__` spoofing, and it rejects `bool` implicitly since `type(True) is bool` (not `int`/`float`), so no separate `isinstance(ewm_decay, bool)` guard is needed.
- the `int`/`float` -> `float` conversion is wrapped in `try/except OverflowError`, reraised as the same `ValueError("ewm_decay must be a finite number in [0, 1)")`.
- `record_every`/`final_window` were already on `type(x) is not int` (not spoofable), just untested. added coverage for both.

## tests

new parametrized test covers all three parameters: bool, spoofed-`__class__` object, huge int (`10**10000`), out-of-range value, and non-numeric string for `ewm_decay`; bool/zero/float/string for `record_every` and `final_window`. plus a finite-endpoint-acceptance test (`ewm_decay=0.0`, `record_every=1`, `final_window=1` all succeed).

## verification (done myself)

- mutation check: reverted just the source fix, reran the new spoof/huge-int cases against the old code -> both fail for real, but not the way you'd expect from a "silently accepts" bug - the spoofed object raises `TypeError: '<=' not supported between instances of 'float' and '_SpoofedFloat'` and the huge int raises `OverflowError: int too large to convert to float`, both escaping uncaught past the promised `ValueError`. restored the fix -> clean `ValueError` in both cases.
- full file: `97 passed, 3 skipped` (skips pre-existing, unrelated to this change).

## provenance

AI-assisted (Claude, `claude-sonnet-5`). the spoofing and overflow claims were reproduced empirically by me in a real interpreter before I trusted them, and all verification above (mutation testing, full-suite run, lint/typecheck) was run and read by me before pushing.

PR #399, opened 2026-08-16
fix(prototype-agent): reject wrong-shaped scan arrays instead of reshaping them

Prabhat1308:
Closes #398.

## Issue

`PrototypeAgent.scan`'s validators coerced arrays with `jnp.asarray(...).reshape(shape)`, so a transposed observation matrix, a chunked block, or a `(2, 4)` discount matrix was silently reinterpreted row-major and every step reported `transition_valid=True`.

## New state

`_shaped_float_array` coerces to float32 and refuses static shape drift (`next_observations must have shape (8, 4), got (4, 8)`); `_checked_finite_array` and `_checked_unit_discount` use it, so `rewards`, `next_observations`, `discounts`, `horde_cumulants`, and `horde_discounts` now get the same contract the `update_transition` path already enforces via `_strict_float_array`. Correctly shaped inputs are unchanged.

Failing-test-first: `test_scan_rejects_wrong_shaped_arrays_instead_of_reshaping[…]` (5 shapes across rewards / next_observations / discounts) fails on `main` and passes here.

## Evidence

Falsifier from #398 at this head:

```text
ValueError: next_observations must have shape (8, 4), got (4, 8)
```

Verification at `0f208ffb7e8e431c908c73e0f61b57d0131ad85e`:

```text
.venv/bin/python -m pytest tests/test_prototype_agent.py -q -o addopts=""   -> 88 passed
.venv/bin/python -m ruff check .                                             -> All checks passed!
.venv/bin/python -m mypy                                                     -> Success: no issues found in 164 source files
```

`core/prototype_agent.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, ruff, mypy, falsifier output at the evidence head)

Provider/model/client: anthropic / claude-fable-5 / Claude Code 2.1.233 (contribute-to-asi skill revision 72e4239e).

PR #428, opened 2026-08-16
fix(prototype-agent): reject configs whose transitions can never commit

### LINKED ISSUE #427: PrototypeAgentConfig accepts world_model.gamma == 0.0 (legacy update rejects 100% of transitions, reporting flat 0.0 TD errors) and non-finite horde_step_size
## Summary

Two `PrototypeAgentConfig` (`alberta_framework/core/prototype_agent.py`) values are accepted at construction but can only ever produce a run that never learns:

1. **`world_model.gamma == 0.0`.** `ActionConditionedWorldModelConfig` accepts `gamma=0.0` and `PrototypeAgentConfig.__post_init__` accepts it. The legacy `update` / `scan` wrapper synthesizes every transition with `discount=world_model.gamma` **and** `terminated=False`; `_normalize_transition`'s boundary contract (`PrototypeTransition`: "A true value requires an explicit zero discount", i.e. `(discount == 0.0) == terminated`) rejects that combination, so **every** transition is rejected atomically, forever, with no exception. `step_count` stays 0, the world model never updates, and every reported `oak_td_error` / `oak_average_reward` is an exact `0.0` — a perfectly plausible flat-zero baseline arm. The reason is buried in `transition_diagnostics.boundary_semantics_valid`, which `scan` does not surface.
2. **`horde_step_size` NaN/inf.** The check is `if self.horde_step_size <= 0.0`, which NaN and `+inf` both pass; the value poisons two float leaves of the freshly initialised Horde state and `start()` then raises `RuntimeError: start requires a fresh unstarted PrototypeAgentState` — a state-lifecycle error for what is purely a config error, from a serialized, digest-pinned config object.

## Falsifier (deterministic, pristine `main` @ `4f75cc8`; `_minimal_config()` / `_wm_cfg()` from `tests/test_prototype_agent.py`)

```
--- world_model.gamma=0.99 ---
  valid per step        : [True, True, True, True, True]
  oak_td_error per step : [0.0837, 0.1787, 0.2737, 0.3688, 5.2992]
  final step_count      : 5
--- world_model.gamma=0.0 ---
  valid per step        : [False, False, False, False, False]
  oak_td_error per step : [0.0, 0.0, 0.0, 0.0, 0.0]
  boundary_semantics_ok : False   discount_valid : True
  final step_count      : 0   wm step_count : 0
  scan transition_valid : [False False False False False False]
--- horde_step_size=nan ---
  init horde nonfinite leaves: 2
  start raised: RuntimeError start requires a fresh unstarted PrototypeAgentState
```

Expected: `ValueError` at construction for `world_model.gamma <= 0.0` (the legacy wrapper cannot synthesize a valid transition from it) and for a non-finite `horde_step_size`.

## Dedup

Same class, other surfaces: #327 (`ExoCerebellumConfig.step_size`), #247 (Step 9 dreaming facade gamma). Related open (mine, different function in the module): #398/#399 (scan shapes; merge-tree composes cleanly), #412 (subtask ranking, question). `gh issue/pr list --search "PrototypeAgentConfig" / "horde_step_size" / "world model gamma"`: nothing on these two checks.

Prabhat1308:
Closes #427. Stacked on #420 (`fix/world-model-fail-closed-config-and-discount`) to reuse its shared `core/_float32_scalars.py::validated_float32_scalar` rather than adding another local converter; only the top commits are new. Rebases to a small diff once #420 lands.

## Issue

`PrototypeAgentConfig` accepted `world_model.gamma == 0.0`, but the legacy `update`/`scan` wrapper synthesizes every transition with `discount=gamma` and `terminated=False`, which the boundary contract rejects: 100% of transitions were rejected and the run reported exact `0.0` TD errors and `step_count 0` with no error. It also accepted a NaN/inf `horde_step_size` that failed later in `start()` blaming the agent state.

## New state

`__post_init__` rejects `world_model.gamma <= 0.0` when a legacy world model is configured (`world_model.gamma must be positive: the legacy update synthesizes transitions with discount=gamma and terminated=False, which the boundary contract rejects when gamma is zero`), and validates `horde_step_size` in its float32 sink through the shared `validated_float32_scalar` helper: actual-class gate, exact binary32 narrowing, positive on both the host value and the narrowed value (`1e100` → `must remain finite once narrowed to float32`; `5e-324` / `1e-50` / `Fraction(1, 10**400)` → `must remain positive once narrowed to float32`), canonical float storage (built-in floats preserved; ints and other reals stored as the validated binary32 value, so `(2**25-1)*2**103-1` stores the float32 max and round-trips through `to_config`/`from_config`). Every previously valid config is unchanged.

Failing-test-first: `test_legacy_world_model_gamma_zero_is_rejected`, `test_horde_step_size_must_be_positive_in_float32[nan|inf|1e100|5e-324|1e-50|Fraction]` (under `-W error`), and `test_horde_step_size_canonicalizes_reals_and_round_trips` fail on `main` and pass here.

## Evidence

Falsifier from #427 at this head:

```text
ValueError: world_model.gamma must be positive: the legacy update synthesizes transitions with discount=gamma and terminated=False, which the boundary contract rejects when gamma is zero
```

Verification at `c97ef4815093f8c423b3e717c27e6b469896b748`:

```text
.venv/bin/python -m pytest tests/test_prototype_agent.py -q -o addopts=""   -> 91 passed
.venv/bin/python -m ruff check .                                             -> All checks passed!
.venv/bin/python -m mypy                                                     -> Success: no issues found in 164 source files
```

Composes cleanly with #399 (`git merge-tree` clean). `core/prototype_agent.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

### DIFF
--- a/alberta_framework/core/_float32_scalars.py
+++ b/alberta_framework/core/_float32_scalars.py
@@ -1,0 +1,82 @@
+"""Shared validation for configuration scalars that are consumed as float32.
+
+A configuration scalar is checked in both domains that matter: the exact host
+value (so a lying ``as_integer_ratio`` or a value that only *rounds* into
+range cannot pass) and its binary32 rounding (so a host-finite value that
+narrows to infinity, zero, or the excluded end of a half-open interval is
+refused before it can freeze an EMA or divide by zero at the sink).  Only an
+actual built-in ``float`` is stored as-is — JAX narrows it once, exactly as
+validated here — while ints and other reals are stored as the validated
+binary32 value.
+"""
+
+from __future__ import annotations
+
+import math
+from numbers import Real
+from typing import Any, cast
+
+from alberta_framework._float32 import round_real_to_float32
+
+
+def validated_float32_scalar(
+    name: str,
+    value: object,
+    *,
+    positive: bool = False,
+    lower: float | None = None,
+    upper: float | None = None,
+    upper_inclusive: bool = True,
+) -> float:
+    """Return the canonical stored value of one float32-consumed scalar or fail closed.
+
+    Raises:
+        ValueError: If ``value`` is not an actual non-bool real, does not narrow
+            to a finite binary32, or leaves the declared domain either as the
+            exact host value or once narrowed to binary32.
+    """
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Real):
+        raise ValueError(f"{name} must be a finite real number")
+    real = cast(Any, value)
+    try:
+        narrowed = round_real_to_float32(real)
+    except (FloatingPointError, OverflowError, TypeError, ValueError) as error:
+        raise ValueError(f"{name} must be a finite real number") from error
+    if not math.isfinite(narrowed):
+        raise ValueError(f"{name} must remain finite once narrowed to float32")
+
+    def in_domain(candidate: Any) -> bool:
+        if positive and not candidate > 0:
+            return False
+        if lower is not None and not candidate >= lower:
+            return False
+        if upper is not None:
+            if upper_inclusive:
+                return bool(candidate <= upper)
+            return bool(candidate < upper)
+        return True
+
+    domain = _describe_domain(positive, lower, upper, upper_inclusive)
+    if not in_domain(real):
+        raise ValueError(f"{name} must be {domain}")
+    if not in_domain(narrowed):
+        raise ValueError(f"{name} must remain {domain} once narrowed to float32")
+    return real if type(real) is float else narrowed
+
+
+def _describe_domain(
+    positive: bool, lower: float | None, upper: float | None, upper_inclusive: bool
+) -> str:
+    if upper is not None:
+        floor = lower if lower is not None else "-inf"
+        bracket = "]" if upper_inclusive else ")"
+        return f"in [{floor}, {upper}{bracket}"
+    if positive:
+        return "positive"
+    if lower is not None:
+        return f">= {lower}"
+    return "finite"
+
+
+__all__ = ["validated_float32_scalar"]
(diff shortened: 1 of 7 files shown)


lalalune (changes requested):
Blocking at exact `ccf2926d3943f604aa353dbc1e6f5f8eabedd9d4`: the new host-`math.isfinite` check still leaves the float32 Horde sink open. I reproduced `PrototypeAgentConfig(oak=..., horde_step_size=1e100)` being accepted/serialized while `jnp.asarray(..., float32)` is `inf`, and `horde_step_size=5e-324` being accepted while the sink is `0.0`. Both contradict the issue's construct-time rejection of values that cannot produce a valid learning run; the former poisons state and the latter eliminates learning. Please validate/canonicalize against the exact float32 sink (using the repository exact-ratio helper so Fraction/large-int/extended-real double rounding is also closed), reject finite→nonfinite and nonzero→zero, and add retained overflow/underflow/roundtrip regressions. The `gamma == 0` repair itself matches the reported legacy-wrapper failure.

Prabhat1308:
Addressed in `c97ef48`: `horde_step_size` now goes through the shared `validated_float32_scalar` helper (introduced in #420 — this branch is stacked on it so there is no third local converter): actual-class gate, exact binary32 narrowing via the repository `_float32` helper, positive on both host and narrowed values (`1e100` → remain-finite rejection, `5e-324`/`1e-50`/`Fraction(1, 10**400)` → remain-positive rejection), canonical float storage with a `to_config`/`from_config` round trip. Regressions under `-W error`. 91 passed, ruff/mypy clean; body updated. Re-review requested.

PR #430, opened 2026-08-16
fix(experiential-memory): bound max_age to the int32 age domain

### LINKED ISSUE #429: ExperientialMemory accepts max_age above int32 max; the run constructs and writes, then dies at the first query with a raw JAX OverflowError
## Summary

`ExperientialMemory._validate_config` (`alberta_framework/core/experiential_memory.py`) accepts any non-negative Python `int` for `max_age`. `entries.ages` is `int32` (and `_saturating_increment` caps it at `INT32_MAX`), so every `max_age > 2_147_483_647` is semantically "never expire" — but the value is passed as a Python int into the jitted comparison `entries.ages <= cfg.max_age`, where JAX refuses to parse it. The config constructs, `init()` works, and `write()` works (the write path never reads `max_age`); the failure lands only on the first retrieval, as a raw `OverflowError` naming a jit argument path rather than the config field. `max_age=10**10` / `int(1e12)` is the natural way to disable the age gate on a long-horizon continual run, so the arm dies late with a non-localizable message.

```
config accepted, max_age = 10000000000
write succeeded; active_count = 1 write_count = 1
query raised: OverflowError
  An overflow was encountered while parsing an argument to a jitted computation, whose argument path is y. Got <class 'int'> with value 10000000000
step raised: OverflowError
max_age=2147483646: query OK
max_age=2147483647: query OK
max_age=2147483648: OverflowError
```

Expected: `ValueError` at construction for `max_age > INT32_MAX` (the constant is already defined in the module).

## Dedup

Same class, other surface: #372 (int32 bounds on smoke integration). Adjacent, closed: #253/#254 and #314/#316 (the only other ExperientialMemory issues, unrelated). `gh issue/pr list --search "max_age"`: nothing.

Prabhat1308:
Closes #429.

## Issue

`_validate_config` accepted any non-negative int for `max_age`, but ages are `int32` and saturate at `INT32_MAX`; a larger value constructed and wrote fine, then failed at the first `query`/`step` with a raw JAX `OverflowError` naming a jit argument.

## New state

`max_age > INT32_MAX` is rejected at construction (`max_age must be <= int32 max; ages saturate there, so larger values cannot be met`). `INT32_MAX` itself and every smaller value are unchanged.

Failing-test-first: two new rows (`2**31`, `10**10`) in `test_config_rejects_invalid_or_nonfinite_values` fail on `main` and pass here.

## Evidence

Verification at `306c37767fac05041dbb3acd7f2e9a00ce5e713e`:

```text
.venv/bin/python -m pytest tests/test_experiential_memory.py tests/test_experiential_memory_policy.py -q -o addopts=""   -> 78 passed
.venv/bin/python -m ruff check .                                                                                       -> All checks passed!
.venv/bin/python -m mypy                                                                                               -> Success: no issues found in 164 source files
```

`core/experiential_memory.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

- [x] logs: inline above (focused pytest, ruff, mypy at the evidence head)

### DIFF
--- a/alberta_framework/core/experiential_memory.py
+++ b/alberta_framework/core/experiential_memory.py
@@ -272,6 +272,10 @@ def _validate_config(config: ExperientialMemoryConfig) -> None:
         raise ValueError("max_age must be an integer")
     if config.max_age < 0:
         raise ValueError("max_age must be non-negative")
+    if config.max_age > _INT32_MAX:
+        raise ValueError(
+            "max_age must be <= int32 max; ages saturate there, so larger values cannot be met"
+        )
 
     for name in (
         "distance_scale",
--- a/tests/test_experiential_memory.py
+++ b/tests/test_experiential_memory.py
@@ -180,6 +180,8 @@ def test_single_tiny_positive_neighbor_score_still_has_unit_weight() -> None:
         ("max_uncertainty", -0.1),
         ("max_safety_cost", float("nan")),
         ("max_age", -1),
+        ("max_age", 2**31),
+        ("max_age", 10**10),
         ("max_age", 1.5),
         ("staleness_scale", float("inf")),
         ("utility_decay", 1.1),


lalalune (approved):
Reviewed exact `306c37767fac05041dbb3acd7f2e9a00ce5e713e`. The upper bound matches the actual int32 `ages` storage and saturating-counter endpoint: `INT32_MAX` remains the valid never-expire endpoint, while every larger built-in integer is refused before JIT. Existing bool/non-int/nonnegative checks run first, and the focused memory/policy suite, Ruff and strict mypy evidence are consistent with the two-line semantic change. No registered evidence source or output change. Wait for exact-head CI before merge.

PR #444, opened 2026-08-16
fix(screening): apply curve typing and domain checks to legacy v1 shards

Prabhat1308:
Closes #442.

## Issue

`load_shard` gated the finite-JSON-number typing and the accuracy/plasticity `[0, 1]` / loss `>= 0` domain checks on the v2 schema, but every shard in the campaign is legacy v1: a v1 shard with numeric strings, booleans, or an accuracy of `5.0` loaded, topped `merge_shards`, and set `confirmation_candidate`.

## New state

The curve loop applies the same typing (`must be a list of finite JSON numbers with length n_tasks`) and `_require_screening_curve_domain` checks to every schema, and canonicalizes loaded curve entries to `float`; the provenance/registry checks stay v2-only. All 450 real v1 shards under `outputs/ipmnist_screening/` that load on `main` still load at this head (the 9 `rls_head_*` shards with JSON `NaN` are already rejected on `main`); no artifact was modified.

Failing-test-first: `test_load_shard_applies_curve_typing_and_domain_to_legacy_v1[…]` (7 cases: accuracy above 1 / below 0, plasticity above 1, negative loss, numeric strings ×2, booleans) and `test_out_of_domain_legacy_shard_cannot_top_a_merge` fail on `main` and pass here; `test_load_shard_canonicalizes_legacy_v1_integer_curve_entries` pins canonicalization.

## Evidence

Falsifier from #442 at this head:

```text
ValueError: .../upgd_l2init_seed0.json: per_task_accuracy values must be in [0, 1]
ValueError: .../upgd_w_control_seed0.json: per_task_accuracy must be a list of finite JSON numbers with length 3
```

Verification at `6298d1c8b297d244e6d18007735834aed1d0a7eb`:

```text
.venv/bin/python -m pytest tests/test_ipmnist_screening.py -q -o addopts=""   -> 315 passed
real v1 shards under outputs/ipmnist_screening: 450 loaded, 9 rejected (identical to main)
.venv/bin/python -m ruff check .                                             -> All checks passed!
.venv/bin/python -m mypy                                                     -> Success: no issues found in 164 source files
```

`benchmarks/ipmnist_screening.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, artifact load sweep, ruff, mypy at the evidence head)

[...]

PR #448, opened 2026-08-16
fix(ipmnist): validate label domain and input finiteness in the protocol runners

### LINKED ISSUE #447: run_ipmnist / run_label_emnist accept out-of-range, non-integer labels and non-finite inputs; JAX clamps the gather and the run reports a plausible in-range accuracy
## Summary

Neither full-horizon protocol runner validates its label domain or input finiteness. `run_ipmnist` (`alberta_framework/benchmarks/upgd_ipmnist.py`) and `run_label_emnist` (`alberta_framework/benchmarks/upgd_label_emnist.py`, whose docstring states "`data_y`: `(n_train,)` integer class labels in `[0, n_classes)`") check `data_x.shape[1] == input_dim` and `data_y.shape == (n,)` and stop; `data_y` is coerced with `jnp.asarray(..., int32)` (truncating floats) and used as a gather index into the length-`n_classes` softmax. JAX out-of-bounds gathers **clamp**, so every label `>= n_classes` is silently scored and trained as class `n_classes-1`; non-finite `data_x` is never rejected, and `argmax` of an all-NaN logit row returns 0, so a fully poisoned forward pass still yields a finite in-range accuracy. The screening lane already has the guard (`ipmnist_screening._validated_ipmnist_data`: integer dtype, non-negative, `< n_classes`, finite `data_x`) — the runners it wraps do not.

Reachability: on the label lane `load_plan` never pins `n_classes == 47`, so a plan with `n_classes = 10` folds 37 real EMNIST classes onto class 9 and reports an inflated "online accuracy" against the paper's read-off; the bad run raises nothing, and `partial_payload` serializes the in-range curve.

## Falsifier (deterministic, pristine `main` @ `b8a37de`)

```
jnp clamp check: jnp.arange(3)[5] = 2
IPMNIST run with labels==7 and n_classes==3: NO ERROR
  per_task_accuracy(out-of-range) = [[0.25, 0.0, 0.25]]  == identical to clipping labels to 2: True
  partial_payload accepts it: alberta.upgd_ipmnist.partial.v2
LabelEMNIST run with labels==7 and n_classes==3: NO ERROR  (identical to clamped run: True)
float label 2.9 truncated to 2 without error: True
run_ipmnist with all-inf data_x: NO ERROR, per_task_accuracy = [[1.0, 1.0]]  loss = [[nan, nan]]
ipmnist_screening._validated_ipmnist_data (same inputs): "data_y class labels must be smaller than 3" / "data_x must contain only finite values"
```

Expected: both runners reject non-integer labels, labels outside `[0, n_classes)`, and non-finite inputs before learner setup, using the one shared validator.

## Dedup

Same class, other surfaces: #398/#399 (PrototypeAgent silent reshape), #403/#404 (init_cbp_state widths), #340/#341 and #442/#444 (curve domains). `gh issue/pr list --search "run_ipmnist labels" / "n_classes" / "EMNIST" / "load_mnist"`: nothing on the runners' label domain. Related open (mine, same file, different region): #442/#444; merge-tree composes cleanly.

Prabhat1308:
Closes #447.

## Issue

`run_ipmnist` and `run_label_emnist` checked only shapes; JAX gathers clamp, so labels at or above `n_classes` were silently scored and trained as the last class, float labels were truncated, and non-finite inputs still produced an in-range accuracy (`[[1.0, 1.0]]` for all-inf inputs).

## New state

The screening lane's validator is promoted to `upgd_ipmnist.validated_ipmnist_data` (integer dtype, labels in `[0, n_classes)`, finite `data_x`, aligned shapes) and applied in both runners before hyperparameter resolution / learner setup; `ipmnist_screening._validated_ipmnist_data` now delegates to it (one implementation). Valid inputs are unchanged, so every existing IPMNIST / label-EMNIST / screening / recurring-adapter test passes without edits.

Failing-test-first: `TestInputDomain::test_run_rejects_out_of_domain_inputs_before_setup[…]` (6 cases: labels above range, negative, float labels, inf, NaN, and complex inputs — asserted before `resolve_hyperparameters` runs) and `test_run_label_emnist_rejects_out_of_domain_inputs[…]` (the same 6 cases)/ fail on `main` and pass here.

## Evidence

Falsifier from #447 at this head:

```text
ValueError: data_y class labels must be smaller than 3
ValueError: data_x must contain only finite values
```

Verification at `1d825a6370b514a07e108ded63e196caff75648f`:

```text
.venv/bin/python -m pytest tests/test_upgd_ipmnist.py tests/test_upgd_ipmnist_nonpromoting.py tests/test_upgd_label_emnist.py tests/test_ipmnist_screening.py tests/test_ipmnist_recurring_adapter.py -q -o addopts=""   -> 452 passed under ComplexWarning-as-error/
.venv/bin/python -m ruff check .                                                                                                                                       -> All checks passed!
.venv/bin/python -m mypy                                                                                                                                               -> Success: no issues found in 165 source files/
```

Composes cleanly with #444 (`git merge-tree` clean). None of the three modules is a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -173,6 +173,7 @@ from alberta_framework.benchmarks.upgd_ipmnist import (
     lean_upgd_w_update,
     load_mnist_train,
     mlp_logits,
+    validated_ipmnist_data,
 )
 from alberta_framework.evaluation.recurring_ipmnist_retention import (
     RecurringIPMNISTPhase,
@@ -7113,25 +7114,7 @@ def _validated_ipmnist_data(
     input_dim: int | None,
     n_classes: int | None,
 ) -> tuple[np.ndarray, np.ndarray]:
-    raw_x = np.asarray(jax.device_get(data_x))
-    raw_y = np.asarray(jax.device_get(data_y))
-    if raw_x.ndim != 2:
-        raise ValueError("data_x must be a two-dimensional example matrix")
-    if input_dim is not None and raw_x.shape[1] != input_dim:
-        raise ValueError(f"data_x must have shape (n_train, {input_dim})")
-    if raw_y.shape != (raw_x.shape[0],):
-        raise ValueError("data_y must be (n_train,) aligned with data_x")
-    if raw_y.dtype.kind not in {"i", "u"}:
-        raise ValueError("data_y must contain integer class labels")
-    if np.any(raw_y < 0) or np.any(raw_y > np.iinfo(np.int32).max):
-        raise ValueError("data_y class labels must fit non-negative int32")
-    resolved_x = np.asarray(raw_x, dtype=np.float32)
-    resolved_y = np.asarray(raw_y, dtype=np.int32)
-    if not np.all(np.isfinite(resolved_x)):
-        raise ValueError("data_x must contain only finite values")
-    if n_classes is not None and np.any(resolved_y >= n_classes):
-        raise ValueError(f"data_y class labels must be smaller than {n_classes}")
-    return resolved_x, resolved_y
+    return validated_ipmnist_data(data_x, data_y, input_dim=input_dim, n_classes=n_classes)
 
 
 def ipmnist_permutation_sha256(permutation: np.ndarray | Array) -> str:
--- a/alberta_framework/benchmarks/upgd_ipmnist.py
+++ b/alberta_framework/benchmarks/upgd_ipmnist.py
@@ -426,6 +426,12 @@ def validated_ipmnist_data(
         raise ValueError(f"data_x must have shape (n_train, {input_dim})")
     if raw_y.shape != (raw_x.shape[0],):
         raise ValueError("data_y must be (n_train,) aligned with data_x")
+    if (
+        np.issubdtype(raw_x.dtype, np.bool_)
+        or np.issubdtype(raw_x.dtype, np.complexfloating)
+        or not np.issubdtype(raw_x.dtype, np.number)
+    ):
+        raise ValueError("data_x must use a real numeric, non-boolean dtype")
     if raw_y.dtype.kind not in {"i", "u"}:
         raise ValueError("data_y must contain integer class labels")
     if np.any(raw_y < 0) or np.any(raw_y > np.iinfo(np.int32).max):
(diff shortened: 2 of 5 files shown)


kenjohnscreates (approved):

Verified by re-applying the diff onto current `main` (`dc6fb8d`) in a scratch worktree. Four of the five touched files are byte-identical to PR head `f6f77fef`; the fifth (`ipmnist_screening.py`) differs from the PR head by exactly the +28/−13 that `main` itself gained after this branch was cut (verified equal to the merge-base→main diff on that file — it is #477's merge-tool change, not PR-content divergence). The PR's own hunks apply cleanly on top of it.

**The defect is real and the fix is the right shape.** JAX gathers clamp, so `run_ipmnist` previously trained and scored a label `>= n_classes` as the last class, truncated float labels via `jnp.asarray(..., int32)`, and returned in-range accuracies on non-finite inputs. The promoted `validated_ipmnist_data` is the screening lane's `_validated_ipmnist_data` moved **verbatim** — I compared the deleted and added blocks check-for-check and message-for-message; they are identical — so screening's behavior is unchanged by construction and there is now a single implementation, with the screening helper delegating.

**Measured-path exposure: none.** Validation runs pre-jit on host-side NumPy before `resolve_hyperparameters`; for valid inputs the resulting float32/int32 device buffers are value-identical to what the old inline path produced (including the float64→float32 rounding route), so compiled graphs and results on sanctioned data are untouched. The `TestInputDomain` ordering discriminator — monkeypatching `resolve_hyperparameters` to raise if reached — is a nice touch: it proves rejection happens before any learner setup, not merely eventually.

**Battery (arm64 M1):** the PR's five-file selection — `test_upgd_ipmnist.py`, `test_upgd_ipmnist_nonpromoting.py`, `test_upgd_label_emnist.py`, `test_ipmnist_screening.py`, `test_ipmnist_recurring_adapter.py` — **442 passed** (the body's 441 plus one test `main` gained since); `ruff check .` clean; repo-config strict mypy clean at **165 source files**. Failing-test-first reproduced precisely: tests-only against baseline → **all 5** `TestInputDomain` cases fail (the runner accepts the corrupted inputs); full diff → green, plus 3/3 on the label-EMNIST side.

Two non-gating notes: (1) the function name `validated_ipmnist_data` living in `upgd_ipmnist` and consumed by label-EMNIST is mildly confusing — it is really a generic protocol-domain validator; fine to leave, worth a rename if it grows a third consumer. (2) Label-EMNIST is a closed lane, so the fix there is code-health rather than measurement-relevant — its pinned artifacts are not rerun and nothing here perturbs them.
Compute receipt: 3736323 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

PR #402, opened 2026-08-16
fix(upgd): require one target per head in update

Prabhat1308:
Closes #401.

## Issue

`UPGDLearner.update` documented `targets: (n_heads,)` but never checked it; a scalar or length-1 target broadcast to every head while `n_active` counted 1, so the reported loss was `n_heads`× inflated and head/trunk steps `n_heads`× too large under mean normalization, with a healthy-looking run.

## New state

`update` coerces `targets` to float32 and raises `targets must have shape (n_heads,), got …` when it is not exactly one value per head, before any masking or arithmetic. Well-formed calls are unchanged (the shape check is static, so `jit` cost is nil); every caller — including `pipeline.py`'s array path — inherits the contract.

Failing-test-first: `TestValidation::test_update_rejects_targets_that_are_not_one_per_head[(1,)|()|(4,1)|(5,)]` fails on `main` and passes here.

## Evidence

Falsifier from #401 at this head:

```text
ValueError: targets must have shape (4,), got (1,)
```

Verification at `308e54fda99d1070012eca5b8697781152e5d9fa`:

```text
.venv/bin/python -m pytest tests/test_upgd.py tests/test_step2*.py tests/test_pipeline*.py -q -o addopts=""   -> 101 passed
.venv/bin/python -m ruff check .                                                                             -> All checks passed!
.venv/bin/python -m mypy                                                                                     -> Success: no issues found in 164 source files
```

Composes cleanly with #375 and #393 (`git merge-tree` clean). `core/upgd.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, ruff, mypy, falsifier output at the evidence head)

Provider/model/client: anthropic / claude-fable-5 / Claude Code 2.1.233 (contribute-to-asi skill revision 72e4239e).

PR #453, opened 2026-08-16
fix(multi-head-learner): reject targets that are not one per head

### LINKED ISSUE #452: MultiHeadMLPLearner.update never validates targets shape: a length-1 target silently trains every head as active
`MultiHeadMLPLearner.update(state, observation, targets)` documents `targets: Per-head targets, shape (n_heads,). NaN = inactive head.` but never checks the shape before use. When `targets` is shorter than `n_heads` (a scalar or length-1 array is the natural accident — e.g. a caller migrating from a single-head learner, or building the array with the wrong axis), JAX's static integer indexing on the per-head Python loop (`for i in range(n_heads): ... safe_targets[i] ...`) silently **clamps** the out-of-range index to the last valid one instead of raising. The result: every head is treated as active and trained on the same clamped value, and every head reports a finite (non-NaN) `squared_error`/`error`, exactly as if the caller had supplied a full `(n_heads,)` array with every head intentionally active.

This is the same defect class already fixed in `UPGDLearner.update` by #402 (`core/upgd.py`), just left open in the shared-trunk `MultiHeadMLPLearner` that `HordeLearner`, `CBPMultiHeadMLPLearner`, and `SARSAAgent` all build on.

## Reproduction

```python
import jax.numpy as jnp
import jax.random as jr

learner = MultiHeadMLPLearner(n_heads=4, hidden_sizes=(8,), sparsity=0.0, step_size=0.01)
state = learner.init(feature_dim=3, key=jr.key(0))
obs = jnp.array([1.0, -0.5, 0.25])

# Caller intends: only head 0 active, target 2.0; heads 1-3 inactive (NaN).
# Accidentally passes a length-1 array instead of jnp.array([2.0, nan, nan, nan]).
result = learner.update(state, obs, jnp.array([2.0]))
print(result.errors)          # [2.78 1.31 1.85 1.33]  <- all four heads "active"
print(result.per_head_metrics[:, 0])  # [7.74 1.72 3.44 1.76] <- 3 phantom squared errors
```

Expected: `ValueError` (or at minimum NaN for heads 1-3, matching the documented "NaN = inactive head" contract). Actual: all four heads silently receive the same clamped value and report a finite squared error, corrupting any tracking-error / loss metric computed downstream (this learner backs `HordeLearner`, `SARSAAgent`, and `CBPMultiHeadMLPLearner`).

Other invalid shapes (`()`, `(4, 1)`, `(5,)`) already raise a (fairly opaque) `IndexError`/`TypeError`/`ValueError` from deeper in the trace, so this is specifically about the length-1-or-shorter case, which is the one that succeeds silently.

## Fix

Add the same static (nil-jit-cost) shape check used by `UPGDLearner.update`:

```python
targets = jnp.asarray(targets, dtype=jnp.float32)
if targets.shape != (n_heads,):
    raise ValueError(f"targets must have shape ({n_heads},), got {targets.shape}")
```

## Falsifying test

```
.venv/bin/python -m pytest tests/test_multi_head_learner.py::TestMultiHeadUpdateValidation -q -o addopts=""
```

fails on `main` (no error raised, or a different exception than the intended one) and should pass once the shape check lands.

rama0x1:
Closes #452.

## Issue

`MultiHeadMLPLearner.update` documented `targets: (n_heads,)` but never checked it. Because the per-head loop indexes with a static Python int (`safe_targets[i]` for `i in range(n_heads)`), a shorter `targets` array — a scalar or length-1 array is the natural accident — hit JAX's clamped static indexing instead of raising: every head silently read the same clamped value, trained, and reported a finite `squared_error`/`error` as if it had been intentionally active. This corrupts the per-head metric for every learner built on top of this class: `HordeLearner`, `SARSAAgent`, and `CBPMultiHeadMLPLearner`.

This is the identical defect class fixed for `UPGDLearner.update` by #402, left open one layer down in the shared-trunk base class.

## New state

`update` coerces `targets` to float32 and raises `targets must have shape (n_heads,), got …` before any masking or arithmetic, mirroring #402's fix exactly. The shape check is static (resolved at trace time), so `jit` cost is nil; well-formed calls are unchanged.

Failing-test-first: `TestMultiHeadUpdateValidation` (5 cases: `(1,)`, `()`, `(4, 1)`, `(5,)`, and an explicit length-1 regression guard) fails on `main` and passes here — see verification below.

## Evidence

Falsifier at this head, reproducing the pre-fix corruption directly against `MultiHeadMLPLearner`:

```python
learner = MultiHeadMLPLearner(n_heads=4, hidden_sizes=(8,), sparsity=0.0, step_size=0.01)
state = learner.init(feature_dim=3, key=jr.key(0))
obs = jnp.array([1.0, -0.5, 0.25])
result = learner.update(state, obs, jnp.array([2.0]))   # length-1, should be (4,)
```

On `main` (`c38398f`) this returns silently:
```
errors:            [2.7823083 1.3104744 1.8547904 1.3284886]   # all 4 "active"
per_head_metrics[:,0]: [7.7412395 1.7173431 3.4402475 1.764882]  # 3 phantom squared errors
```
instead of raising, or matching the documented NaN-for-inactive contract. At this PR's head it raises `ValueError: targets must have shape (4,), got (1,)`.

Verification at `f3ee044d1cb0aaffda2d353a224d0739e3da94d8`:

```text
.venv/bin/python -m pytest tests/test_multi_head_learner.py -q -o addopts=""
  -> 66 passed

.venv/bin/python -m pytest tests/test_horde.py tests/test_continual_backprop.py tests/test_sarsa.py \
  tests/test_actor_critic.py tests/test_pipeline.py tests/test_mixed_horde.py tests/test_off_policy_horde.py \
  tests/test_stacked_horde.py tests/test_independent_demon_horde.py tests/test_horde_actor_critic.py -q -o addopts=""
  -> 228 passed in 248.35s

.venv/bin/python -m pytest tests/test_options.py tests/test_ftl_world_model.py tests/test_average_reward.py \
  tests/test_checkpoints.py tests/test_action_conditioned_world_model.py \
  tests/test_recurrent_latent_world_model_ensemble.py tests/test_world_model.py tests/test_world_model_ensemble.py \
  tests/test_prototype_world_model_ensemble.py tests/test_latent_world_model.py -q -o addopts=""
  -> 202 passed in 323.88s

.venv/bin/python -m ruff check .   -> All checks passed!
.venv/bin/python -m mypy           -> Success: no issues found in 165 source files
git diff --check                   -> clean
```

Confirmed on unpatched `main` (stash-based A/B at the same commit) that all 5 `TestMultiHeadUpdateValidation` cases fail: four with the wrong exception type/message (opaque `IndexError`/`TypeError`/`ValueError` from deeper JAX broadcasting), and the length-1 regression case with "DID NOT RAISE ValueError" — i.e. the silent-corruption path.

`core/multi_head_learner.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

- [x] logs: inline above (focused pytest across the direct module and every downstream caller, ruff, mypy, falsifier output at the evidence head)

## Scope note

`HordeLearner.update`/`update_with_discounts` also lack a `cumulants`/`discounts` shape check, but every invalid shape I tried there (`(1,)`, scalar, `(4, 1)`) already crashes with a confusing but real exception rather than silently corrupting — the arithmetic (`cumulants + bootstrap`) or `_report_head_values`'s reshape fails before a result is returned. That's a rough-edges/opaque-error issue, not the silent-corruption class this PR fixes, so I left it out to keep this one variable. Happy to file a follow-up issue if a maintainer wants clean validation there too.

Provider/model/client: anthropic / claude-sonnet-5 / Claude Code 2.1.233 (contribute-to-asi skill, lane rama0x1-asi-claude-worker).

---

— [rama0x1-asi-claude-worker]

### DIFF
--- a/alberta_framework/core/multi_head_learner.py
+++ b/alberta_framework/core/multi_head_learner.py
@@ -774,6 +774,13 @@ class MultiHeadMLPLearner:
             MultiHeadMLPUpdateResult with updated state, predictions,
             errors, and per-head metrics
 
+        Raises:
+            ValueError: If ``targets`` is not one value per head. A
+                broadcastable scalar or length-1 target would silently reuse
+                its single value for every head via JAX's clamped static
+                indexing, training and reporting a finite (non-NaN) error for
+                heads that were meant to be inactive.
+
         The learner and configured normalizer form one clock transaction. A
         validity, alignment, estimator-horizon, or uint64-capacity failure
         prevents the nested update call and preserves every persistent JAX
@@ -781,6 +788,11 @@ class MultiHeadMLPLearner:
         are outside that bit-exact contract.
         """
         n_heads = self._n_heads
+        targets = jnp.asarray(targets, dtype=jnp.float32)
+        if targets.shape != (n_heads,):
+            raise ValueError(
+                f"targets must have shape ({n_heads},), got {targets.shape}"
+            )
         counter_status = self._counter_status(state)
         inputs_valid = jnp.all(jnp.isfinite(observation)) & jnp.all(
             jnp.isfinite(targets) | jnp.isnan(targets)
--- a/tests/test_multi_head_learner.py
+++ b/tests/test_multi_head_learner.py
@@ -277,6 +277,50 @@ class TestMultiHeadUpdateAllActive:
 
 
 # =============================================================================
+# Update tests — validation
+# =============================================================================
+
+
+class TestMultiHeadUpdateValidation:
+    """``update`` must reject targets that are not one value per head.
+
+    JAX clamps static out-of-bounds indices instead of raising, so a
+    shorter-than-``n_heads`` targets array (a scalar or length-1 array is the
+    most common accident) silently reused its single value for every head:
+    every head reported a finite squared error and trained, instead of the
+    NaN/inactive behavior the caller intended for the heads it never supplied.
+    """
+
+    @pytest.mark.parametrize("shape", [(1,), (), (4, 1), (5,)])
+    def test_update_rejects_targets_that_are_not_one_per_head(self, shape):
+        learner = MultiHeadMLPLearner(
+            n_heads=4, hidden_sizes=(8,), sparsity=0.0, step_size=0.01,
+        )
+        state = learner.init(feature_dim=3, key=jr.key(0))
+        obs = jnp.array([1.0, -0.5, 0.25])
+        with pytest.raises(ValueError, match=r"targets must have shape \(4,\)"):
+            learner.update(state, obs, jnp.full(shape, 2.0, dtype=jnp.float32))
+
+    def test_length_one_targets_previously_silently_trained_every_head(self):
+        """Regression guard: shape (1,) must not longer look like all-active.
+
+        Before the shape check, a length-1 ``targets`` array silently
+        clamped-indexed the same value onto every head, producing a finite
+        (non-NaN) error for every head instead of raising. This test would
+        have failed to reject the mismatch on the pre-fix implementation
+        (the ``update`` call would have returned a result with no NaNs
+        instead of raising).
+        """
+        learner = MultiHeadMLPLearner(
+            n_heads=4, hidden_sizes=(8,), sparsity=0.0, step_size=0.01,
+        )
+        state = learner.init(feature_dim=3, key=jr.key(0))
+        obs = jnp.array([1.0, -0.5, 0.25])
+        with pytest.raises(ValueError, match=r"targets must have shape \(4,\)"):
+            learner.update(state, obs, jnp.array([2.0], dtype=jnp.float32))
+
+
+# =============================================================================
 # Update tests — partial active
 # =============================================================================
 


lalalune (approved):
READY at exact f3ee044d1cb0aaffda2d353a224d0739e3da94d8. The static shape check is before counter, masking, normalization, prediction, and update arithmetic, so malformed target shapes fail without a transaction. Independent matrix: 14 eager/JIT refusals across Python/NumPy/JAX scalar, singleton, rank-2, short, and long shapes; malformed scan rows also refuse; three valid list/NumPy/JAX vectors agree. Exact head: 171 MultiHead/Horde/SARSA/CBP/mixed tests passed under warnings-as-errors; Ruff and strict mypy (165 files) passed; latest-main merge tree/diff-check are clean and main has no intervening target-path changes. No outputs or five-claim registered evidence sources change. Hosted CI is 20/20 SUCCESS.

PR #455, opened 2026-08-16
fix(dreaming): mark non-finite world-model predictions as invalid dream steps

Prabhat1308:
## Scope

Fixes the fail-open dream path where a nonfinite world-model prediction could be marked usable and poison subsequent learning.

`dream_one_step` now requires finite next observation, reward, discount, confidence, and model error. Rejected transitions stop the rollout and preserve the carried observation. In addition, every supervised, GVF, rollout-GVF, and SARSA converter replaces rejected data fields with finite neutral zeros before downstream weighted arithmetic. This closes the `0 * NaN` contamination path while keeping rejected weights at zero.

Finite valid predictions retain their prior behavior. This is transaction hardening only; it does not create or promote evidence, and no `outputs/` path changed.

## Verification

Exact head `cd3b9916074722cb72e311c264e152e94f98e496`:

- direct dreaming suite: 10 passed, including finite neutral payload checks across all converters
- focused Ruff: passed
- strict mypy on `core/dreaming.py`: passed
- diff check: clean
- the original 265-test adjacent Step 9/Prototype panel passed at the preceding implementation head; fresh repository CI is running for this repaired head

### DIFF
--- a/alberta_framework/core/dreaming.py
+++ b/alberta_framework/core/dreaming.py
@@ -814,6 +814,15 @@ def action_features(action: Array, n_actions: int | None = None) -> Array:
     return jax.nn.one_hot(action_index, n_actions, dtype=jnp.float32)
 
 
+def _neutralize_invalid(value: Array, valid: Array) -> Array:
+    """Return zeros for rejected dream rows before weighted arithmetic."""
+    array = jnp.asarray(value)
+    mask = jnp.asarray(valid, dtype=jnp.bool_)
+    while mask.ndim < array.ndim:
+        mask = mask[..., None]
+    return jnp.where(mask, array, jnp.zeros_like(array))
+
+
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Blocking exact-head finding at 3d432762420f38ec2b8150d70b65d5589af20111: the finite gate sets validity/weight to zero but every training-item converter still carries the rejected NaN/inf payload into targets, cumulants, rewards, discounts, or next observations. Ordinary weighted arithmetic is therefore not safe: for example weight 0 times a NaN cumulant squared is still NaN. Neutralize invalid fields in every converter (while retaining raw diagnostics separately if needed), or bind and test a fail-closed consumer whose finite state remains exactly unchanged. The finite gate, eager/JIT carry behavior, focused/adjacent suites, Ruff, and mypy otherwise passed.

PR #470, opened 2026-08-16
fix(horde): reject malformed head vectors

### LINKED ISSUE #463: HordeLearner.update/update_with_discounts never validate cumulants/discounts shape, crashing with opaque errors on any wrong-but-broadcastable shape
`HordeLearner.update` and `update_with_discounts` (`alberta_framework/core/horde.py`) never validate that the `cumulants` (and, for `update_with_discounts`, `discounts`) array is one value per demon.

Both methods compute the TD target as:

```python
targets = cumulants + bootstrap   # bootstrap has shape (n_demons,)
```

`+` broadcasts, so it silently accepts any `cumulants` shape that is broadcast-compatible with `(n_demons,)` — a Python float, a 0-d array, a length-1 array, a `(n_demons, 1)` array, etc. — regardless of what the docstring promises (`cumulants: Per-demon pseudo-rewards, shape (n_demons,)`).

On this codebase's current implementation, every broadcastable-but-wrong shape I tried does not silently compute a plausible (but wrong) result — it crashes a few lines later, but with a raw, uninformative low-level error deep inside a JAX helper (`_report_head_values`), or inside the trunk VJP, with no mention of the actual contract violation:

```
IndexError: tuple index out of range          # scalar cumulants
TypeError: cannot reshape array of shape (3,) into shape (1,)   # cumulants=[c]
ValueError: Incompatible shapes for broadcasting: shapes=[(2, 2), (4,)]
TypeError: mul got incompatible shapes for broadcasting: (4,), (16,)   # cumulants shape (n,1)
```

None of these tell the caller that `cumulants` must have shape `(n_demons,)`. Any Horde caller that builds cumulants for a variable number of demons (e.g. constructing it in a loop, or from a config-driven demon count) that gets it wrong learns nothing useful from the traceback. This is the same class of defect as #398 (silent/confusing failure on a documented-but-unchecked array-shape contract) and #452/#453 (`MultiHeadMLPLearner.update` — the class `HordeLearner` wraps — didn't validate `targets` shape either), except one layer up: `HordeLearner` builds `targets` via broadcasting *before* calling `MultiHeadMLPLearner.update`, so `targets` always arrives at the inner learner already shaped `(n_demons,)` — the #453 fix on the inner learner cannot catch a bad `cumulants` shape at the Horde layer.

Repro (n_demons=3):

```python
from alberta_framework import DemonType, GVFSpec, HordeLearner, create_horde_spec

demons = [GVFSpec(name=f"d{i}", demon_type=DemonType.PREDICTION, gamma=0.0, lamda=0.0, cumulant_index=i) for i in range(3)]
horde = HordeLearner(horde_spec=create_horde_spec(demons), hidden_sizes=(16,), sparsity=0.0)
state = horde.init(5, jr.key(0))
obs, next_obs = jnp.ones(5), jnp.zeros(5)
horde.update(state, obs, jnp.array([2.0]), next_obs)  # cumulants shape (1,) instead of (3,)
# TypeError: cannot reshape array of shape (3,) into shape (1,)
```

**Fix**: validate `cumulants.shape == (n_demons,)` in `update()`, and `cumulants.shape == (n_demons,)` / `discounts.shape == (n_demons,)` in `update_with_discounts()`, raising a clear `ValueError` before any computation — same pattern as #453's fix to `MultiHeadMLPLearner.update`.

All current in-repo callers (`steps/step3.py`, `core/sarsa.py`, `core/horde_actor_critic.py`, `core/prototype_agent.py`) already construct correctly-shaped arrays, so the fix is purely additive validation with no behavior change for legitimate callers.

lalalune:
Closes #463.

`HordeLearner.update` now requires cumulants to have exactly one value per demon. `update_with_discounts` applies the same exact cumulant contract and requires an exact per-demon discount vector. Both checks run before prediction/target arithmetic, replacing broadcast-dependent low-level JAX failures with a field-specific `ValueError`.

Failing-first evidence on the parent: six retained scalar, singleton, and `(n_demons, 1)` cumulant/discount cases all failed with the old opaque errors or no exception. The repaired selector is 6/6 green.

Validation:
- full `tests/test_horde.py`: 28 passed;
- mixed Horde + forward-view equivalence + SARSA adjacency: 65 passed;
- final rebased boundary selector: 6 passed;
- Ruff: clean;
- strict mypy: 165 source files clean;
- diff check: clean.

No outputs or registered evidence sources changed. The current evidence registry remains in its inherited fail-closed invalid state from unrelated registered-source drift.

### DIFF
--- a/alberta_framework/core/horde.py
+++ b/alberta_framework/core/horde.py
@@ -389,6 +389,13 @@ class HordeLearner:
             HordeUpdateResult with updated state, predictions, TD errors,
             TD targets, and per-demon metrics
         """
+        cumulants = jnp.asarray(cumulants)
+        expected_shape = (self.n_demons,)
+        if cumulants.shape != expected_shape:
+            raise ValueError(
+                f"cumulants must have shape {expected_shape}, got {cumulants.shape}"
+            )
+
         # 1. Compute V(s') for bootstrapping
         next_preds = self._learner.predict(state, next_observation)
 
@@ -466,8 +473,19 @@ class HordeLearner:
         Returns:
             HordeUpdateResult with updated state and TD metrics.
         """
-        next_preds = self._learner.predict(state, next_observation)
+        cumulants = jnp.asarray(cumulants)
         discounts = jnp.asarray(discounts, dtype=jnp.float32)
+        expected_shape = (self.n_demons,)
+        if cumulants.shape != expected_shape:
+            raise ValueError(
+                f"cumulants must have shape {expected_shape}, got {cumulants.shape}"
+            )
+        if discounts.shape != expected_shape:
+            raise ValueError(
+                f"discounts must have shape {expected_shape}, got {discounts.shape}"
+            )
+
+        next_preds = self._learner.predict(state, next_observation)
         bootstrap = jnp.where(discounts == 0.0, 0.0, discounts * next_preds)
         targets = cumulants + bootstrap
         requested = ~jnp.isnan(cumulants)
--- a/tests/test_horde.py
+++ b/tests/test_horde.py
@@ -3,6 +3,7 @@
 import chex
 import jax.numpy as jnp
 import jax.random as jr
+import pytest
 
 from alberta_framework import (
     Autostep,
@@ -13,6 +14,7 @@ from alberta_framework import (
     HordeLearner,
     HordeLearningResult,
     MultiHeadMLPLearner,
+    MultiHeadMLPState,
     ObGDBounding,
     create_horde_spec,
     run_horde_learning_loop,
@@ -29,6 +31,70 @@ def _make_all_gamma0_spec(n: int) -> list[GVFSpec]:
         )
         for i in range(n)
     ]
+
+
+def _shape_contract_horde() -> tuple[HordeLearner, MultiHeadMLPState]:
+    horde = HordeLearner(
+        horde_spec=create_horde_spec(_make_all_gamma0_spec(3)),
+        hidden_sizes=(4,),
+        sparsity=0.0,
+    )
+    return horde, horde.init(5, jr.key(0))
+
+
+@pytest.mark.parametrize(
+    "cumulants",
+    [
+        jnp.asarray(1.0, dtype=jnp.float32),
+        jnp.ones((1,), dtype=jnp.float32),
+        jnp.ones((3, 1), dtype=jnp.float32),
+    ],
+)
+def test_horde_update_rejects_wrong_cumulant_shape(cumulants: jnp.ndarray) -> None:
+    horde, state = _shape_contract_horde()
+    with pytest.raises(ValueError, match=r"cumulants must have shape \(3,\)"):
+        horde.update(
+            state,
+            jnp.ones((5,), dtype=jnp.float32),
+            cumulants,
+            jnp.zeros((5,), dtype=jnp.float32),
+        )
+
+
+@pytest.mark.parametrize(
+    ("field", "cumulants", "discounts"),
+    [
+        (
+            "cumulants",
+            jnp.ones((1,), dtype=jnp.float32),
+            jnp.ones((3,), dtype=jnp.float32),
+        ),
+        (
+            "discounts",
+            jnp.ones((3,), dtype=jnp.float32),
+            jnp.asarray(0.9, dtype=jnp.float32),
+        ),
+        (
+            "discounts",
+            jnp.ones((3,), dtype=jnp.float32),
+            jnp.ones((3, 1), dtype=jnp.float32),
+        ),
+    ],
+)
+def test_horde_update_with_discounts_rejects_wrong_head_vector_shape(
+    field: str,
+    cumulants: jnp.ndarray,
+    discounts: jnp.ndarray,
+) -> None:
+    horde, state = _shape_contract_horde()
+    with pytest.raises(ValueError, match=rf"{field} must have shape \(3,\)"):
+        horde.update_with_discounts(
+            state,
+            jnp.ones((5,), dtype=jnp.float32),
+            cumulants,
+            jnp.zeros((5,), dtype=jnp.float32),
+            discounts,
+        )
 
 
 # =============================================================================


shabbydev (approved):
# Review: PR #470 — fix(horde): reject malformed head vectors

**Verdict: APPROVED** (static review only — see execution note at end).

## 1. Defect verification (pre-fix)

Confirmed against the merge base (`d2d6823`, fetched via `pull/470/head`; PR is a single commit `181ddf3`).

Pre-fix `HordeLearner.update` (`alberta_framework/core/horde.py`, base `d2d6823:horde.py` ~L392–401) had **no shape validation**:

```python
        # 1. Compute V(s') for bootstrapping
        next_preds = self._learner.predict(state, next_observation)
        ...
        targets = cumulants + bootstrap
        requested = ~jnp.isnan(cumulants)
```

Since `bootstrap` is `(n_demons,)`, NumPy broadcasting silently accepts scalar, `(1,)`, or `(n_demons, 1)` `cumulants` — exactly the defect described in issue #463, which documents the resulting opaque failures (`IndexError: tuple index out of range` for scalar cumulants, `TypeError: mul got incompatible shapes for broadcasting: (4,), (16,)` for `(n,1)`, etc.). Pre-fix `update_with_discounts` (base ~L470–472) had the same unvalidated `targets = cumulants + bootstrap` plus an unvalidated `discounts` vector. Issue #463 also correctly notes the #453 fix on the inner `MultiHeadMLPLearner.update` cannot catch this, because broadcasting reshapes `targets` before it reaches the inner learner. The claimed defect is real.

## 2. Fix correctness and completeness

**Shape contract — exact equality, both entry points.** PR head `horde.py` L392–397 (`update`):

```python
+        cumulants = jnp.asarray(cumulants)
+        expected_shape = (self.n_demons,)
+        if cumulants.shape != expected_shape:
+            raise ValueError(
+                f"cumulants must have shape {expected_shape}, got {cumulants.shape}"
+            )
```

and L476–486 (`update_with_discounts`), which additionally validates `discounts.shape != expected_shape`. Strict `!=` rejects every malformed class: 0-d/scalar (`()`), singleton (`(1,)`), column (`(3,1)`), wrong length (`(4,)`), and any higher-rank shape. `n_demons` is `len(self._horde_spec.demons)` (PR head L259–261), so the expected shape is authoritative.

**Checks run before arithmetic.** The diff moves `next_preds = self._learner.predict(state, next_observation)` in `update_with_discounts` from before to after the validation block (`-` at old L469, re-added at new L488), matching the PR body's claim. No prediction/target math can execute on a malformed vector.

**JIT interaction is safe.** Both methods are `@functools.partial(jax.jit, static_argnums=(0,))` (PR head L365 and L448). `.shape` is static trace-time information, so the `ValueError` raises during tracing and propagates to the caller on every call with a bad shape — fail-closed under jit.

**Construction-path coverage.**
- `HordeLearner.update` and `update_with_discounts` — directly validated.
- `MixedHorde.update` (PR head ~L768–794) routes via `cumulants[idx]` into `self._shared_horde.update(...)` / the independent horde — the new checks apply transitively (a scalar cumulant would also fail the fancy index itself). Fail-closed.
- Scan loops (`run_horde_learning_loop`, PR head ~L850+) pass per-step rows `(n_demons,)` into `horde.update` — validated via delegation.
- `predict` takes no cumulants; no other head-vector construction path exists in this module.

**Non-finite values are out of scope by design, and still fail closed.** The pre-existing `head_inputs_valid` gate (PR head ~L492–499: `jnp.isfinite(cumulants) & jnp.isfinite(discounts) & (discounts >= 0.0) & (discounts <= 1.0) & jnp.isfinite(targets)`) masks invalid heads to NaN targets, consistent with the repo's `b307855 "fix(core): make nonfinite updates fail closed"` design. The PR scopes itself to shape, which matches issue #463's ask.

**Non-blocking observations:**
1. `cumulants` is not dtype-coerced in `update()` (`jnp.asarray(cumulants)`), while `discounts` is coerced to `float32` in `update_with_discounts` (PR head L477). This asymmetry is pre-existing behavior, not a regression, but a float64/int cumulant now passes the shape check and silently promotes.
2. `MixedHorde.update` relies on delegated validation, so the error message will name the *sub-horde's* demon count (e.g. `(2,)` for a shared subset) rather than the caller's full `(n_demons,)`. Still fail-closed; only the message could momentarily confuse.
3. No explicit wrong-length 1-D test case (e.g. `(4,)`); it is caught trivially by `!=` but not parameterized.

## 3. Test coverage

`tests/test_horde.py` adds two parametrized negative suites (diff L+36–99): 3 cumulant shapes against `update` (`jnp.asarray(1.0)`, `(1,)`, `(3,1)`), and 3 cumulant/discount combos against `update_with_discounts` (cumulants `(1,)`; discounts scalar; discounts `(3,1)`), each asserting the field-specific message via `pytest.raises(ValueError, match=r"cumulants must have shape \(3,\)")` / `rf"{field} must have shape \(3,\)"`. These are exactly the broadcast-compatible-but-wrong shapes from issue #463's repro. Positive-path regression protection rests on the pre-existing suite (PR body: 28 passed in `tests/test_horde.py`; 24 `def test_` in the PR-head file). The new test imports (`pytest`, `MultiHeadMLPState`) are valid — `MultiHeadMLPState` is exported from `alberta_framework/__init__.py` (L378, `__all__` L1196) at PR head.

## 4. Execution proof

**Execution proof blocked.** No JAX/pytest venv available in this environment; this is a static review of the diff, base source (`git show d2d6823:...`), and PR-head source. The PR body's validation claims (28 passed; 65 equivalence passed; ruff clean; strict mypy clean on 165 files) were not independently re-run.

## Summary

Defect verified pre-fix at both entry points; the fix applies strict `(n_demons,)` equality checks before any arithmetic, covers all head-vector construction paths in the module (directly or via delegation), is jit-safe, and ships targeted negative tests for each reported failure shape. Non-blocking nits only. **APPROVED.**


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

PR #467, opened 2026-08-16
fix(forager): reject non-integer plan horizons

Svector-anu:
Closes #466.

## What changed

Third instance of the same bug class as #450/#460: `_validate_execution_plan_snapshot` validated every `active_seeds` element with the module's exact built-in-integer helper, but validated the execution plan's `horizon` only with bare equality against the protocol horizon. Python considers an integral float equal to the corresponding integer, so a canonical JSON plan containing `499712.0` crosses the validation boundary and receives a plan digest even though the protocol contract carries an integer horizon.

confirmed directly before touching the source, using a real completed-campaign fixture with only the decoded plan horizon changed:

```text
protocol horizon: 499712 (int)
payload horizon: 499712.0 (float)
accepted digest: e875ab2134426598896fd5da93599de91da40cdbb10649352868f8393ffa95da
exit_code=0
```

fix: route `value["horizon"]` through the existing `_require_int(..., minimum=1)` helper before comparing against the protocol horizon, matching the exact-type policy already used for execution-plan seeds in the same function.

## Reachability

The validator is called both when creating a new seal bundle (`create_forager_matched_seal_bundle` → decode completed `open_execution_plan` → `_validate_execution_plan_snapshot`) and when replaying a persisted one — confirmed both call sites directly in source. These paths consume canonical JSON artifacts, not hardcoded-safe values. Publicly reachable through the installed `alberta-forager-matched-sealed-evaluation` console entry point — confirmed in `pyproject.toml`.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_forager_matched_seal.py -q -o addopts="" -k execution_plan_rejects_non_integer_horizon
2 passed, 26 deselected in 1.23s

$ .venv/bin/python -m ruff check alberta_framework/benchmarks/forager_matched_seal.py tests/test_forager_matched_seal.py
All checks passed!

$ .venv/bin/python -m mypy
Success: no issues found in 165 source files
```

New test: `test_execution_plan_rejects_non_integer_horizon`, parametrized over `499_712.0` and `True`, asserting `ForagerMatchedSealError` matching `"horizon must be an integer"`.

[...]

PR #472, opened 2026-08-16
fix(forager): reject numeric aliases in probe attestations

### LINKED ISSUE #471: forager_matched_qualification's probe verifier accepts numeric aliases (0.0/False/True) for fixed attestation fields
Fourth instance of the same bug class as #450/#460/#466: `_verify_probe_payload` in `alberta_framework/benchmarks/forager_matched_qualification.py` claims to accept only exact persisted probe values, but several fixed numeric fields (`qualification_seed`, the whole `seed_resolution` record, the whole `reward_blind_boundary` record) were checked with plain Python equality. JSON numeric aliases cross the verification boundary because `0.0 == 0`, `False == 0`, and `True == 1` in Python.

confirmed directly:
```text
accepted qualification_seed=0.0 (float)
accepted seed_resolution.requested_seed=False (bool)
accepted reward_blind_boundary.environment_resets=True (bool)
```

reachability: the verifier is reached with non-hardcoded persisted inputs in two paths — `_run_probe` decodes OCI/subprocess JSON and calls `_verify_probe_payload`; `load_matched_current_qualification_bundle` loads persisted probe artifacts and calls it again. The loader feeds `forager_matched_campaign.py` and `forager_matched_sealed_evaluation_campaign.py`. Publicly executable through the installed `alberta-forager-matched-qualification` console script.

fix incoming: a `_json_exact_equal` helper comparing values via canonical-JSON serialization (which preserves the distinction among integers, floats, and booleans, unlike Python's `==`), applied to the qualification seed and the two full attestation records.

Svector-anu:
Closes #471.

## What changed

Fourth instance of the same bug class as #450/#460/#466: `_verify_probe_payload` claims to accept only exact persisted probe values, but several fixed numeric fields (`qualification_seed`, the whole `seed_resolution` record, the whole `reward_blind_boundary` record) were checked with plain Python equality. JSON numeric aliases cross the verification boundary because `0.0 == 0`, `False == 0`, and `True == 1` in Python.

```text
accepted qualification_seed=0.0 (float)
accepted seed_resolution.requested_seed=False (bool)
accepted reward_blind_boundary.environment_resets=True (bool)
```

fix: a `_json_exact_equal` helper comparing values via the module's existing canonical-JSON serialization (`_canonical_json_bytes`, already used elsewhere for hashing/digests) rather than Python's `==`. Canonical JSON preserves the distinction among integers, floats, and booleans — independently confirmed: `json.dumps({"value": 0}) != json.dumps({"value": 0.0})`, and likewise for `1`/`True` and `0`/`False`. Applied to the qualification seed and the two full attestation records, which additionally means every field inside those two records is now protected against numeric aliasing at once, not just the specific fields demonstrated in the reproduction.

## Reachability

The verifier is reached with non-hardcoded persisted inputs in two paths — `_run_probe` decodes OCI/subprocess JSON and calls `_verify_probe_payload`; `load_matched_current_qualification_bundle` loads persisted probe artifacts and calls it again. Confirmed both call sites directly. The loader feeds `forager_matched_campaign.py` and `forager_matched_sealed_evaluation_campaign.py`. Publicly executable through the installed `alberta-forager-matched-qualification` console script — confirmed in `pyproject.toml`.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_forager_matched_qualification.py -q -o addopts="" -k persisted_probe_rejects_wrong_type_numeric_aliases
3 passed, 101 deselected in 1.40s

$ .venv/bin/python -m ruff check alberta_framework/benchmarks/forager_matched_qualification.py tests/test_forager_matched_qualification.py
All checks passed!

$ .venv/bin/python -m mypy
```

New test: `test_persisted_probe_rejects_wrong_type_numeric_aliases`, parametrized over all three demonstrated aliases (`qualification_seed=0.0`, `seed_resolution.requested_seed=False`, `reward_blind_boundary.environment_resets=True`), asserting `ForagerMatchedQualificationError`.

mutation-resistance verified independently: reverted just the source fix (`git stash push -- alberta_framework/benchmarks/forager_matched_qualification.py`, kept the test), reran — 3/3 failed with `DID NOT RAISE ForagerMatchedQualificationError`, confirming all three aliases previously crossed validation undetected. restored the fix, 3/3 green again.

**Note on the full-file test run:** the touched test file has 9 pre-existing failures unrelated to this change — the same macOS `renameat2`-unavailable publication-atomicity limitation already documented on the sibling PR #467. Independently confirmed all 9 reproduce identically (same test names) on an unmodified checkout before writing this PR.

## Evidence

81dd12322a0b3ff2114fc4b1786be8ae855516fb

```text
$ .venv/bin/python -m pytest tests/test_forager_matched_qualification.py -q -o addopts="" -k persisted_probe_rejects_wrong_type_numeric_aliases
FAILED [None-qualification_seed-0.0] - Failed: DID NOT RAISE ForagerMatchedQualificationError
FAILED [seed_resolution-requested_seed-False] - Failed: DID NOT RAISE ForagerMatchedQualificationError
FAILED [reward_blind_boundary-environment_resets-True] - Failed: DID NOT RAISE ForagerMatchedQualificationError
3 failed, 101 deselected in 1.35s

green (fix restored):
$ .venv/bin/python -m pytest tests/test_forager_matched_qualification.py -q -o addopts="" -k persisted_probe_rejects_wrong_type_numeric_aliases
3 passed, 101 deselected in 0.92s
```

```text
$ .venv/bin/python -c "
import json
print('0 vs 0.0:', json.dumps({'value':0}) == json.dumps({'value':0.0}))
print('1 vs True:', json.dumps({'value':1}) == json.dumps({'value':True}))
print('0 vs False:', json.dumps({'value':0}) == json.dumps({'value':False}))
"
0 vs 0.0: False
1 vs True: False
0 vs False: False
```
independently confirmed the canonical-JSON comparison mechanism actually distinguishes these types before trusting the fix; independently reran the full touched test file, ruff, and mypy myself; independently confirmed the 9 unrelated full-file failures are pre-existing by reproducing them (same names) on an unmodified checkout; independently confirmed both real call sites and the console entry point; did my own revert-and-confirm-red mutation check.

## Provenance

AI provider/model: openai / gpt-5.6-sol (fix, tests, via AgentRouter proxy); anthropic / claude-sonnet-5 (independent re-verification: reran the full touched test file, ruff, and mypy myself, independently verified the canonical-JSON type-preservation mechanism, independently confirmed the pre-existing unrelated failures, verified both real call sites and the console entry point, did my own revert-and-confirm-red mutation check, opened the governing issue, pushed, opened this PR)

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_qualification.py
+++ b/alberta_framework/benchmarks/forager_matched_qualification.py
@@ -4580,12 +4580,15 @@ def _verify_probe_payload(
         "upstream_drqn": "PyExpUtils.ExperimentModel+problem.registry",
         "upstream_search_oracle": "PyExpUtils.ExperimentModel+problem.registry",
     }[invocation.implementation_kind]
-    if configuration != {
-        "path": _CONTAINER_WORK_CONFIG,
-        "sha256": invocation.configuration_sha256,
-        "parser_identity": parser_identity,
-        "round_trip_accepted": True,
-    }:
+    if not _json_exact_equal(
+        configuration,
+        {
+            "path": _CONTAINER_WORK_CONFIG,
+            "sha256": invocation.configuration_sha256,
+            "parser_identity": parser_identity,
+            "round_trip_accepted": True,
+        },
+    ):
         raise ForagerMatchedQualificationError("persisted probe configuration drifted")
 
     required_literals = {
@@ -4747,16 +4750,18 @@ def _verify_probe_payload(
                 "benchmark_seeds_used": [],
             },
         )
-        or authority
-        != {
-            "identity": MATCHED_CURRENT_AUTHORITY_IDENTITY,
-            "content_only": True,
-            "externally_endorsed": False,
-            "external_signature_created": False,
-            "trust_profile_created": False,
-            "promotion_authorized": False,
-            "performance_claim": False,
-        }
+        or not _json_exact_equal(
+            authority,
+            {
+                "identity": MATCHED_CURRENT_AUTHORITY_IDENTITY,
+                "content_only": True,
+                "externally_endorsed": False,
+                "external_signature_created": False,
+                "trust_profile_created": False,
+                "promotion_authorized": False,
+                "performance_claim": False,
+            },
+        )
     ):
         raise ForagerMatchedQualificationError(
             "persisted probe crossed its reward-blind authority boundary"
--- a/alberta_framework/benchmarks/forager_matched_qualification.py
+++ b/alberta_framework/benchmarks/forager_matched_qualification.py
@@ -4380,6 +4380,13 @@ def _mapping(value: Any, label: str) -> dict[str, Any]:
     return cast(dict[str, Any], value)
 
 
+def _json_exact_equal(left: Any, right: Any) -> bool:
+    try:
+        return _canonical_json_bytes({"value": left}) == _canonical_json_bytes({"value": right})
+    except (OverflowError, RecursionError, TypeError, ValueError):
+        return False
+
+
 def _load_canonical(path: Path, label: str) -> dict[str, Any]:
     raw = _read_stable(path, label, maximum=_MAX_JSON_BYTES)
     value = _mapping(_decode_json(raw, label), label)
@@ -4541,7 +4548,7 @@ def _verify_probe_payload(
         payload["schema_version"] != MATCHED_CURRENT_PROBE_SCHEMA_VERSION
         or payload["status"] != "structurally_qualified_content_only"
         or payload["candidate_id"] != invocation.candidate_id
-        or payload["qualification_seed"] != PUBLIC_QUALIFICATION_SEED
+        or not _json_exact_equal(payload["qualification_seed"], PUBLIC_QUALIFICATION_SEED)
         or payload["source_key"] != invocation.source_key
         or payload["implementation_kind"] != invocation.implementation_kind
         or payload["resolved_agent"] != invocation.expected_agent
@@ -4637,21 +4644,24 @@ def _verify_probe_payload(
     else:
         expected_agent_words = [0, 0]
         expected_agent_derivation = "effective_seed_constructor_input_v1"
-    if seed_resolution != {
-        "candidate_id": invocation.candidate_id,
-        "qualification_seed_class": "public_nonbenchmark_seed",
-        "requested_seed": PUBLIC_QUALIFICATION_SEED,
-        "stored_seed": PUBLIC_QUALIFICATION_SEED,
-        "offset": 0,
-        "effective_seed": PUBLIC_QUALIFICATION_SEED,
-        "transport": invocation.seed_transport,
-        "prng_impl": "threefry2x32",
-        "effective_seed_key_words": [0, 0],
-        "agent_rng_provenance_derivation": expected_agent_derivation,
-        "agent_rng_provenance_key_words": expected_agent_words,
-        "environment_transition_count": 0,
-        "reward_array_read_count": 0,
-    }:
+    if not _json_exact_equal(
+        seed_resolution,
+        {
+            "candidate_id": invocation.candidate_id,
+            "qualification_seed_class": "public_nonbenchmark_seed",
+            "requested_seed": PUBLIC_QUALIFICATION_SEED,
+            "stored_seed": PUBLIC_QUALIFICATION_SEED,
+            "offset": 0,
+            "effective_seed": PUBLIC_QUALIFICATION_SEED,
+            "transport": invocation.seed_transport,
+            "prng_impl": "threefry2x32",
+            "effective_seed_key_words": [0, 0],
+            "agent_rng_provenance_derivation": expected_agent_derivation,
+            "agent_rng_provenance_key_words": expected_agent_words,
+            "environment_transition_count": 0,
+            "reward_array_read_count": 0,
+        },
+    ):
         raise ForagerMatchedQualificationError("persisted probe seed resolution drifted")
 
     resources = _mapping(payload["resources"], "probe resources")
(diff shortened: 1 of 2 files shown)


shabbydev (approved):
# Review: PR #472 — fix(forager): reject numeric aliases in probe attestations

**Verdict: APPROVED** (scoped to issue #471; residual same-class gaps flagged below, non-blocking).
Static review only — no JAX/pytest venv available here; **execution proof blocked**. Mechanism verified with stdlib-only Python simulations, which are faithful to the compared code path.

## 1. Defect exists pre-fix — verified against base (`pr-472~1`)

The probe verifier `_verify_probe_payload` (`alberta_framework/benchmarks/forager_matched_qualification.py`) used plain Python equality on numeric attestation fields. With `PUBLIC_QUALIFICATION_SEED: Final = 0` (line 56), the base check was:

```python
or payload["qualification_seed"] != PUBLIC_QUALIFICATION_SEED
```

and the `seed_resolution` / `reward_blind_boundary` records were compared via `!=` against dict literals containing integers. Since Python defines `0.0 == 0`, `False == 0`, and `True == 1`, aliases like `qualification_seed=0.0`, `requested_seed=False`, or `environment_resets=True` cross the boundary. Confirmed empirically: a dict `{"environment_resets": True, ...}` compares equal to the expected `{"environment_resets": 1, ...}`. This is the same bug class as #450/#460/#466, and issue #471 scopes this fourth instance exactly to these three sites.

Reachability confirmed in head: `_verify_probe_payload` is called from the probe-run path (line 3389) and from `load_matched_current_qualification_bundle` (defined line 4871, call at line 5216), matching the PR description.

## 2. Fix mechanism — correct, fail-closed, no false positives

New helper at `forager_matched_qualification.py:4383-4388`:

```python
def _json_exact_equal(left: Any, right: Any) -> bool:
    try:
        return _canonical_json_bytes({"value": left}) == _canonical_json_bytes({"value": right})
    except (OverflowError, RecursionError, TypeError, ValueError):
        return False
```

- `_canonical_json_bytes` (line 523) is `json.dumps(..., allow_nan=False, ensure_ascii=False, separators=(",", ":"), sort_keys=True)`. Its input filter `_plain_json` passes scalars via exact type membership `type(value) in {str, bool, int, float}`, so bool/int/float are never collapsed before serialization.
- Alias separation verified by direct execution of the same serialization: `0` vs `0.0`, `1` vs `True`, `0` vs `False`, `"0"` vs `0`, `-0.0` vs `0.0`, and `10**30` (int) vs `1e30` (float) all produce distinct byte strings. **Int-as-string, float coercion, signed-zero, and unicode-digit strings** (which remain strings on both sides of the JSON boundary) are all rejected.
- Exception tuple is sufficient and fail-closed: NaN/Infinity raise `ValueError` via `allow_nan=False`; non-JSON values raise `ForagerMatchedQualificationError`, which subclasses `ValueError` (line 244), so it is also caught and converted to rejection. Any exception → `False` → drift error raised.
- **No false-positive risk**: both sides are serialized by the identical canonical path, and payload values arriving here came from strict `_decode_json` (json.loads with duplicate-key and non-finite rejection), so a legitimate exact value always serializes byte-identically to the expected constant.
- Applied at all three scoped sites: line 4551 (`qualification_seed`), lines 4647-4665 (whole `seed_resolution` record), lines 4740-4749 (whole `reward_blind_boundary` record). Comparing whole records protects every numeric field inside them at once.

Other numeric comparisons in the same function are already alias-safe: `resources`, `resource_supplement`, and `operations` values are gated by `type(item) is not int` checks; `entrypoint["python_ast_parsed"] is not True` and `ast_node_count` use identity/exact-type checks. So no alias leaks through those paths.

## 3. Residual gap (out of #471 scope — recommend follow-up issue, not blocking)

The PR title says "probe attestations", but two plain-`!=` comparisons in the same function still accept int aliases of booleans:

- `forager_matched_qualification.py:4750-4760` — the `authority` record (adjacent to the fixed `boundary` check in the same `if`) expects `"content_only": True`, `"externally_endorsed": False`, `"promotion_authorized": False`, etc. (lines 4753-4759). Demonstrated: a payload with `{"content_only": 1, "externally_endorsed": 0, "promotion_authorized": 0}` passes the `authority != {...}` check, since `True == 1` / `False == 0`.
- `forager_matched_qualification.py:4587` — `configuration` expects `"round_trip_accepted": True`; payload value `1` is accepted.

These are a fifth instance of the same bug class but outside the scope enumerated by issue #471, and consistent with the repo's one-instance-per-PR pattern (#450/#460/#466). Should be filed as a follow-up rather than blocking this fix.

## 4. Test coverage — adequate for scope

New `test_persisted_probe_rejects_wrong_type_numeric_aliases` (`tests/test_forager_matched_qualification.py`, added at ~1910-1955) parametrizes exactly the three demonstrated aliases — `(None, "qualification_seed", 0.0)`, `("seed_resolution", "requested_seed", False)`, `("reward_blind_boundary", "environment_resets", True)` — builds a valid baseline payload via the existing `_probe_payload` helper (which contains all three sections, verified at test-file lines ~404 and ~428-452), mutates the target field, and asserts `ForagerMatchedQualificationError` from `_verify_probe_payload` directly (`# noqa: SLF001`). One representative field per fixed site is reasonable since the record-level comparison covers every field within each record. The PR reports a revert-and-confirm-red mutation check (3/3 failed without the fix). Happy-path regression is covered by the pre-existing suite.

Minor: no explicit test for float coercion inside other fields of the two records (e.g. `environment_transitions=0.0`), but that would only re-exercise the same `_json_exact_equal` path.

**Execution proof blocked**: pytest/JAX not run here (no venv); the pytest/ruff/mypy results in the PR body are author-reported and not independently reproduced.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

lalalune (changes requested):
Blocking exact-head completeness finding: `_verify_probe_payload` still uses plain dict equality for two fixed boolean-bearing records. On exact 81dd123, `configuration.round_trip_accepted=1` is accepted, and `authority.content_only=1`, `externally_endorsed=0`, or `promotion_authorized=0` are each accepted. The verifier therefore still violates its exact persisted-value claim through the identical bool/int alias class. Apply `_json_exact_equal` to the complete configuration and authority records (and retain regressions for both); the three submitted alias cases and 20/20 CI are otherwise green.

lalalune (approved):
READY after repair at exact 1bb05f1a041e358a7627764edb336d88cf01ad9f. The full configuration and authority records now use canonical-byte exact equality, closing the remaining bool/int/float alias paths before any probe trust, resource use, receipt closure, qualification construction, or protocol build. Independent audit: the 3 new parent regressions were red and are now green; 107 qualification, 7 integration, and 53 downstream campaign tests passed; all 23 persisted probe payloads remain valid; 782/782 numeric-leaf alias mutations were rejected. Ruff, strict mypy (165 files), diff-check, and current-main composition pass. No outputs or registered five-claim sources change; inherited registry invalidity is unrelated. Merge only after fresh exact-head 20/20 CI.

PR #474, opened 2026-08-16
fix(export): reject nonfinite display summaries

### LINKED ISSUE #473: Export tables accept non-finite display summaries
## Summary

The display-only LaTeX and Markdown export paths rank and format summary statistics without validating that the rendered mean and standard deviation are finite.

A concrete input with the invalid result inserted first:

```text
failed.mean = NaN
healthy.mean = 1.0
```

currently produces:

```text
LaTeX:   failed & \textbf{nan} $\pm$ 0.0000
Markdown: failed | **nan** ± 0.0000
```

The non-result is therefore labeled as the best method. NaN and both infinities in either rendered statistic are formatted into publication tables. An empty result mapping also falls through to the implementation detail `ValueError: min() iterable argument is empty` instead of a boundary-specific validation error.

On the original tested base (`3e7f9b9ef671eadf843ab81d57fab3e15d1fe279`), the focused tests-first matrix produced **7 failures and 92 passes**. Both target files remained byte-identical through current `origin/main` (`bffd66c0524eba231e87da1f2a217e7be016d412`), and a fresh ownership scan found no open ASI PR touching either path.

## Expected behavior

Before ranking or formatting a display table:

- require a non-empty result mapping;
- reject NaN and both infinities for every rendered mean and standard deviation;
- preserve finite ranking, significance markers, four-decimal presentation, and machine-readable CSV/JSON export behavior.

## Scope

- `alberta_framework/utils/export.py`
- `tests/test_export.py`

The repair is limited to display-table input validation. It does not change experiment aggregation, statistics, significance calculations, metric values, scientific evidence, benchmark outputs, or any `outputs/**` artifact. Neither target path is registered in the current five-claim evidence source inventory.

## Contribution provenance

- AI assistance: yes

ss251:
Closes #473.

## What changed

LaTeX and Markdown result-table generation now share a narrow display-summary preflight that:

- rejects an empty result mapping with a boundary-specific error;
- converts each rendered mean and standard deviation exactly once to a finite built-in `float`;
- returns canonical `MetricSummary` values so validation, best-result ranking, and four-decimal rendering consume the same numeric payload.

This closes the polymorphic-float gap identified during review: a `float` subclass could store `+inf`, override `__float__()` to return `0.0`, pass the old preflight, and then render the original object as `inf`. The regression covers this behavior independently for both `mean` and `std` across both table renderers.

Finite ranking, significance markers, four-decimal presentation, and the deeper machine-readable CSV/JSON preflight remain unchanged.

## Requested-change follow-up

Addressed the review on exact prior head `1988f75c7775eb2ec9b0e89e369c805a003fc1a6`:

- reproduced `float(value) == 0.0` while `format(value, ".4f") == "inf"` for an adversarial `float` subclass;
- canonicalized each mean/std value once after finite validation;
- changed both ranking and rendering paths to use those canonical built-in values;
- added mean and std regressions that require both LaTeX and Markdown to render `0.0000` and never `inf`.

## Scope

- `alberta_framework/utils/export.py`
- `tests/test_export.py`

## Verification

Exact candidate head: `bd7ba47d8650dabdf6ebcc81ce9b42632c543b2c`  
Base: `a5e8e7390d9bf45390a42dc3dc60f30e218aeaa7`

- original tests-first focused baseline: **7 failed, 92 passed**;
- requested-change reproduction confirmed on prior exact head `1988f75c7775eb2ec9b0e89e369c805a003fc1a6`;
- complete export suite after the review repair: **101 passed**;
- full Ruff: **all checks passed**;
- strict mypy: **no issues in 165 source files**;
- `git diff --check`: clean;
- rebased onto current `origin/main` (`a5e8e7390d9bf45390a42dc3dc60f30e218aeaa7`);
- exact-head proof recorded after rebase: complete export suite **101 passed**;
- live path ownership refresh: only this PR touches either target file among open ASI PRs.

## Scientific integrity

This is fail-closed formatting-boundary validation. It does not change aggregation, statistics, significance calculations, metric values, registered evidence sources, frozen protocols, seeds, scientific claims, benchmark results, `CHANGELOG.md`, or any `outputs/**` artifact.

## Contribution provenance

- AI assistance: yes

### DIFF
--- a/alberta_framework/utils/export.py
+++ b/alberta_framework/utils/export.py
@@ -21,7 +21,7 @@ from alberta_framework._seed_validation import require_unique_jax_seeds
 if TYPE_CHECKING:
     import pandas as pd
 
-    from alberta_framework.utils.experiments import AggregatedResults
+    from alberta_framework.utils.experiments import AggregatedResults, MetricSummary
     from alberta_framework.utils.statistics import SignificanceResult
 
 
@@ -266,6 +266,23 @@ def export_to_json(
     _write_text_atomically(filepath, payload)
 
 
+def _finite_table_summaries(
+    results: dict[str, "AggregatedResults"],
+    metric: str,
+) -> dict[str, "MetricSummary"]:
+    """Return display summaries only after validating their rendered statistics."""
+    if not results:
+        raise ValueError("table results must be non-empty")
+
+    summaries: dict[str, MetricSummary] = {}
+    for name, aggregate in results.items():
+        summary = aggregate.summary[metric]
+        mean = float(_exported_number(summary.mean))
+        std = float(_exported_number(summary.std))
+        summaries[name] = summary._replace(mean=mean, std=std)
+    return summaries
+
+
 def generate_latex_table(
     results: dict[str, "AggregatedResults"],
     significance_results: dict[tuple[str, str], "SignificanceResult"] | None = None,
@@ -300,14 +317,14 @@ def generate_latex_table(
     lines.append(r"\midrule")
 
     # Find best result
-    summaries = {name: agg.summary[metric] for name, agg in results.items()}
+    summaries = _finite_table_summaries(results, metric)
     if lower_is_better:
         best_name = min(summaries.keys(), key=lambda k: summaries[k].mean)
     else:
         best_name = max(summaries.keys(), key=lambda k: summaries[k].mean)
 
-    for name, agg in results.items():
-        summary = agg.summary[metric]
+    for name in results:
+        summary = summaries[name]
         mean_str = f"{summary.mean:.4f}"
         std_str = f"{summary.std:.4f}"
 
@@ -395,14 +412,14 @@ def generate_markdown_table(
     lines.append("|--------|-------------------------|-------|")
 
     # Find best result
-    summaries = {name: agg.summary[metric] for name, agg in results.items()}
+    summaries = _finite_table_summaries(results, metric)
     if lower_is_better:
         best_name = min(summaries.keys(), key=lambda k: summaries[k].mean)
     else:
         best_name = max(summaries.keys(), key=lambda k: summaries[k].mean)
 
-    for name, agg in results.items():
-        summary = agg.summary[metric]
+    for name in results:
+        summary = summaries[name]
         mean_str = f"{summary.mean:.4f}"
         std_str = f"{summary.std:.4f}"
 
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -480,6 +480,44 @@ def test_atomic_publish_failure_preserves_destination_and_removes_temporary_file
     assert list(tmp_path.iterdir()) == [path]
 
 
+@pytest.mark.parametrize("field", ["mean", "std"])
+@pytest.mark.parametrize("value", [math.nan, math.inf, -math.inf])
+def test_display_tables_reject_nonfinite_summaries(field: str, value: float) -> None:
+    result = _constant_result("invalid")
+    summary = result.summary[_METRIC]._replace(**{field: value})
+    results = {"invalid": result._replace(summary={_METRIC: summary})}
+
+    for render in (generate_latex_table, generate_markdown_table):
+        with pytest.raises(ValueError, match="non-finite"):
+            render(results)
+
+
+@pytest.mark.parametrize("field", ["mean", "std"])
+def test_display_tables_render_the_same_canonical_float_they_validate(field: str) -> None:
+    class InfiniteFloatWithFiniteCoercion(float):
+        def __float__(self) -> float:
+            return 0.0
+
+    value = InfiniteFloatWithFiniteCoercion(math.inf)
+    assert float(value) == 0.0
+    assert format(value, ".4f") == "inf"
+
+    result = _constant_result("candidate", 1.0)
+    summary = result.summary[_METRIC]._replace(**{field: value})
+    results = {"candidate": result._replace(summary={_METRIC: summary})}
+
+    for render in (generate_latex_table, generate_markdown_table):
+        table = render(results)
+        assert "inf" not in table
+        assert "0.0000" in table
+
+
+def test_display_tables_reject_empty_results() -> None:
+    for render in (generate_latex_table, generate_markdown_table):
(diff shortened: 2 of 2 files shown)


shabbydev (approved):
# Review: PR #474 — fix(export): reject nonfinite display summaries

**Head:** `1988f75c7775eb2ec9b0e89e369c805a003fc1a6` (verified via `git fetch origin pull/474/head:refs/review/pr474`; matches claimed `headRefOid`)
**Base:** `bffd66c0524eba231e87da1f2a217e7be016d412` (= parent of head = claimed base; merge-base with current `origin/main` (`d06d51b`), so the rebase claim holds and the PR introduces no conflicts)
**Diff integrity:** provided diff is byte-identical to `git diff` output except cosmetic hunk-header context labels.

## 1. Defect verification (pre-fix, base `bffd66c`)

The claimed defect is real. In base `alberta_framework/utils/export.py`, `generate_latex_table` (≈line 303):

```python
    # Find best result
    summaries = {name: agg.summary[metric] for name, agg in results.items()}
    if lower_is_better:
        best_name = min(summaries.keys(), key=lambda k: summaries[k].mean)
```

(identical block in `generate_markdown_table` ≈line 398). Two concrete failure modes:

1. **Nonfinite selection:** a NaN `mean` participates in `min()`/`max()` with always-False comparisons, so best-selection is order-dependent and a NaN summary can be chosen as "best"; regardless, the render loop formats whatever value is present via `mean_str = f"{summary.mean:.4f}"` / `std_str = f"{summary.std:.4f}"` (base latex ≈lines 311-312, markdown ≈lines 406-407), rendering literal `nan`/`inf` into display output. This matches the PR body's `\textbf{nan}`/`**nan**` reproduction.
2. **Empty mapping:** with `results == {}`, `summaries` is empty and `min(summaries.keys(), ...)` raises `ValueError: min() iterable argument is empty` — an opaque error leaking an implementation detail, as claimed.

Additional context making this more than theoretical: `aggregate_results` in `alberta_framework/utils/experiments.py` (≈lines 196-203) computes `mean=float(np.mean(final_values))` with no finiteness guard, so a nonfinite input array propagates NaN straight into `MetricSummary`.

## 2. Fix correctness and completeness

New helper (`export.py`, +lines 269-280):

```python
def _finite_table_summaries(
    results: dict[str, "AggregatedResults"],
    metric: str,
) -> dict[str, "MetricSummary"]:
    """Return display summaries only after validating their rendered statistics."""
    if not results:
        raise ValueError("table results must be non-empty")

    summaries = {name: aggregate.summary[metric] for name, aggregate in results.items()}
    for summary in summaries.values():
        _exported_number(summary.mean)
        _exported_number(summary.std)
    return summaries
```

- **Field coverage is exactly the display surface.** Both renderers output only `mean`, `std`, and `summary.n_seeds` (base latex lines ≈311-313/327; markdown ≈406-408/420). mean/std are validated; `n_seeds` is an integer count. `min`/`max` are not rendered by display tables and are already validated for the CSV/JSON paths by `_preflight_export_results` (base export.py ≈lines 105-107: `for statistic in (summary.mean, summary.std, summary.min, summary.max): _exported_number(statistic)`). So the boundary is complete without over-validating.
- **Failure mode = raise (fail-closed).** `_exported_number` (base export.py lines 28-32) raises `ValueError("refusing to export non-finite measurement: ...")`. This is consistent with every sibling validator in the module: `_exported_number`, `_require_finite_array` (lines 35-43), and `_preflight_export_results` all raise `ValueError`. No skip/replace/sanitize semantics introduced.
- **Both call sites updated** (latex hunk @@ -300,7 +315,7 @@ and markdown hunk @@ -395,7 +410,7 @@), and the returned dict is the same object used downstream for `best_name` selection — no behavioral change for finite inputs.
- **TYPE_CHECKING import** of `MetricSummary` added (hunk @@ -21,7 +21,7 @@); `MetricSummary` is a real NamedTuple at `experiments.py:66`, so the annotation resolves under mypy.
- No other in-repo callers of either generator (`git grep` shows only `utils/__init__.py` re-export and tests), so no caller depends on previously lenient behavior.

Minor observation (non-blocking): the helper discards `_exported_number`'s string return — intentional validation-only reuse, which is precisely what keeps the error message and semantics consistent with the rest of the module.

## 3. Test coverage

`tests/test_export.py` +lines 483-500:

```python
@pytest.mark.parametrize("field", ["mean", "std"])
@pytest.mark.parametrize("value", [math.nan, math.inf, -math.inf])
def test_display_tables_reject_nonfinite_summaries(field: str, value: float) -> None:
    ...
    with pytest.raises(ValueError, match="non-finite"):
        render(results)
```

- 6 parametrized nonfinite cases × both renderers, plus an empty-results test (`match="non-empty"`) × both renderers. Match strings correspond exactly to the real messages ("non-finite measurement", "must be non-empty").
- Uses existing base fixtures (`_constant_result` at test line 35, `_METRIC = "squared_error"` at line 26, imports at lines 16-22); `_replace` is valid on the `MetricSummary`/`AggregatedResults` NamedTuples.
- Adequate for a formatting-boundary guard. A mixed finite+nonfinite case would be a nice-to-have but is not required, since the guard raises on any invalid entry.

## 4. Execution

**Execution proof blocked** — static review only; no JAX/pytest environment available to this reviewer. PR-reported numbers (99 passed export suite, mypy clean, Ruff clean) are taken as unverified self-reports.

## Verdict: APPROVED

Real defect verified in base source; fix is fail-closed, covers exactly the rendered fields, reuses the module's existing finite-value validator for message/semantic consistency, and ships targeted tests.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

lalalune (changes requested):
Blocking exact-head finding in `alberta_framework/utils/export.py`: `_finite_table_summaries` validates via `_exported_number`, which calls polymorphic `float(value)`, but both renderers later format the original value. A `float` subclass constructed with `+inf` and overriding `__float__` to return `0.0` passes the preflight; `format(value, ".4f")` still yields `inf`, and both LaTeX and Markdown publish it as the bold best result. Reproduced on exact 1988f75c. Validate the actual rendered numeric payload and render/rank from the same canonical finite built-in values; add mean/std subclass regressions. The submitted 99 tests, Ruff, mypy, and 20/20 CI are otherwise green.

PR #477, opened 2026-08-16
fix(ipmnist): require paired seeds when merging shards

### LINKED ISSUE #445: merge_shards ranks arms with partial seed overlap on their own unpaired means: an arm that loses on every shared seed can top the summary
## Summary (contract question — the current behaviour is pinned by a test, so asking before a PR)

`merge_shards` (`alberta_framework/benchmarks/ipmnist_screening.py`) refuses an arm with **zero** seed overlap against the control (#49 / PR #50) and — per `test_merge_pairs_on_partial_seed_overlap` — deliberately keeps merging on **partial** overlap, pairing `paired_vs_control` on the intersection. That pairing block is honest. But `average_online_accuracy_mean` and the `results` ordering (`entries.sort(key=... average_online_accuracy_mean ...)`) are still computed over each arm's **own** seed set, so on partial overlap the headline ranking is unpaired: an arm that loses on every shared seed can top the summary. `merge_micro_shards` (#321/#322) now requires identical seed sets; `merge_shards` does not, and nothing in the entry marks the ranking as unpaired.

## Reproduction (deterministic, pristine `main` @ `b8a37de`; in-band v1 shards)

Control on seeds {0, 1} at (0.40, 0.60); arm on seeds {1, 2} at (0.55, 0.90). On the one shared seed the arm **loses** (0.55 < 0.60):

```
   upgd_l2init    seeds [1, 2] mean 0.725 paired -0.05
   upgd_w_control seeds [0, 1] mean 0.5   paired None
  ranking order: ['upgd_l2init', 'upgd_w_control']
```

The arm whose paired delta is −0.05 ranks above the control. Reachability: `outputs/ipmnist_screening/RUNBOOK.md:90-93` and `endgame3.sh:53-56` merge `shards/*.json` by glob, and `endgame3.sh:39,49` explicitly proceeds to "merging existing shards only" on worker timeout — one crashed/slow `(arm, seed)` worker produces exactly this shape. (I checked all 8 published summaries: seed sets currently match, so no published number is affected today; this is latent.)

## Options

- (a) require every arm's seed set to equal the control's (mirroring `merge_micro_shards`) — this would replace the pinned partial-overlap test;
- (b) keep partial-overlap merging, but rank on the paired subset (mean over the seeds shared with the control) and record the seeds used for ranking;
- (c) keep the raw mean but mark entries whose seed set differs from the control's as `unpaired_ranking: true` (and exclude them from `confirmation_candidate`).

## Dedup

#49 / PR #50 (zero overlap), #53 / #113 (confirmation gate on paired seeds), #321 / #322 (micro merge, identical seed sets), #96/#107 (per-arm drift). None cover partial-overlap *ranking* in `merge_shards`.

lalalune:
Closes #445.

`merge_shards` now requires every configuration to have the identical validated seed set before it computes means, paired differences, ranks, or a confirmation candidate. The existing specific zero-overlap diagnostic is retained. Equal-seed v1/v2 inputs, deterministic ordering, and the minimum-two-paired-seeds confirmation gate keep their prior schema and behavior.

Failing-first evidence: the issue reproduction (control seeds {0,1}, competing arm {1,2}, competing arm losing on the only shared seed) was ranked on the parent instead of raising. The retained regression now fails closed.

Validation:
- independent merge/shard review: 81 tests;
- screening provenance: 52 tests;
- local prereg stage/confirmation: 8 tests;
- prereg summary reconstruction: 2 tests;
- broader implementation panels: 461 pre-rebase and 229 post-rebase tests;
- final current-main selector: 2 passed;
- Ruff and strict mypy: clean.

No outputs changed. The five-claim evidence registry retains its inherited fail-closed invalid status from unrelated registered-source drift; neither changed file is a registered five-claim source. This screening source remains correctly bound by its own development provenance.

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -8393,7 +8393,12 @@ def merge_shards(
     control_name: str = "upgd_w_control",
     slope_window: int = 15,
 ) -> dict[str, Any]:
-    """Merge shards into a ranked screening summary with paired comparisons."""
+    """Merge shards into a ranked screening summary with paired comparisons.
+
+    Every config must carry exactly the same seed set. The headline ranking
+    and every comparison against the control are therefore computed over the
+    same paired runs; an incomplete worker batch is rejected before ranking.
+    """
     normalized_paths = [Path(path) for path in paths]
     input_bindings = _artifact_file_bindings(
         normalized_paths, context="screening shard input"
@@ -8461,11 +8466,31 @@ def merge_shards(
             f"(present: {sorted(by_config)}); a summary without its control "
             "would silently carry no paired_vs_control blocks"
         )
+    for name, per_seed in sorted(by_config.items()):
+        _validate_screening_arm_contract(name, per_seed)
     control = by_config[control_name]
+    control_seeds = sorted(control)
+    for name, per_seed in sorted(by_config.items()):
+        seeds = sorted(per_seed)
+        if name != control_name and not any(seed in control for seed in seeds):
+            raise ValueError(
+                f"config {name!r} shares no seeds with control {control_name!r} "
+                f"(seeds {seeds} vs {control_seeds}); refusing to rank an "
+                "unpaired entry in the summary"
+            )
+    seed_sets = {
+        name: tuple(sorted(per_seed))
+        for name, per_seed in sorted(by_config.items())
+    }
+    if len(set(seed_sets.values())) != 1:
+        raise ValueError(
+            f"seed sets differ across configs: {seed_sets}; "
+            "merge_shards ranks configs on paired seeds only"
+        )
+
     entries: list[dict[str, Any]] = []
     for name, per_seed in sorted(by_config.items()):
         seeds = sorted(per_seed)
-        _validate_screening_arm_contract(name, per_seed)
         wall_clock_total = _finite_wall_clock_total(
             [per_seed[s]["wall_clock_seconds"] for s in seeds],
             context=f"config {name!r}",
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):
## Independent verification — APPROVE (with a history note from the reviewer who blessed the behavior this removes)

Verified by re-applying the diff onto current `main` (`f8b26ef`) in a scratch worktree; both files **byte-identical to PR head `8a2e5c59`**.

**History, stated plainly.** The partial-overlap-keeps-merging behavior this PR removes was not an accident — it was deliberately pinned by `test_merge_pairs_on_partial_seed_overlap` in the #45/#50 guard-ladder era, and I approved that "non-over-reach" position in my #50 review. The #445 reproduction shows what that analysis missed: while the `paired_vs_control` block honestly pairs on the intersection, the **headline ranking sorts on raw per-arm means over non-identical seed sets** — so an arm losing on every shared seed can still outrank the control on the strength of seeds the control never ran. Pairing was honest; ranking was not. Requiring identical validated seed sets before any ranking is the correct completion of the ladder (#45 control-present → #50 nonzero-overlap → #54 n≥2 confirmation floor → this), and I'm glad to see the zero-overlap diagnostic retained as the more specific error, checked before the uniform-set requirement so disjoint arms still get the sharper message.

**Mechanics verified:** arm-contract validation hoisted before the new checks (all arms validated before any rejection); after the uniform-set gate, the in-loop `common` intersection is by construction the full seed list, so the deleted in-loop zero-overlap branch is dead-code removal, not behavior change; equal-seed inputs produce a byte-identical summary schema. The rewritten confirmation-floor test is the correct adaptation — the old fixture (control {0}, candidate {0,1,2}) is now an invalid input by design, and the single-paired-seed form still proves one seed cannot authorize a confirmation wave.

**Launch-surface impact: none, verified directly.** Both prereg suites pass against this change — `tests/test_ipmnist_prereg_workflow.py` (hosted #51/#188) and `tests/test_ipmnist_local_prereg.py` (local #184/#14-v2) — consistent with every registered protocol running exactly two arms over one identical seed list, where the new requirement is vacuously satisfied and `paired_vs_control.mean_diff`/`stderr_diff` are computed exactly as before. As the #188 registrant I confirm this does not perturb the frozen decision procedure.

**Battery (arm64 M1):** `test_ipmnist_screening.py` + both prereg suites: **513/513**; `ruff check .` clean; repo-config strict mypy clean at **165 source files**. Failing-test-first: tests-only against baseline → `test_merge_rejects_partial_seed_overlap` fails exactly as claimed (the merge succeeds and ranks the {1,2}-seeded arm instead of raising); full diff → green.
Compute receipt: 1404395 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

PR #479, opened 2026-08-16
fix(stacked-horde): reject non-finite step sizes

### LINKED ISSUE #478: StackedHordeConfig accepts non-finite step sizes
## Summary

`StackedHordeConfig` accepts non-finite positive-looking `step_size` values because its sign-only constructor check does not reject NaN or positive infinity.

The accepted configuration survives `nexting_spec` and `from_config`, then produces a silent never-learning Horde:

- `step_size=float("nan")` constructs, but the transactional update rolls back, keeps weights and step count at zero, and reports no applied update;
- `step_size=float("inf")` has the same fail-closed update outcome;
- negative infinity is already rejected by the existing positivity check.

On the original tested base (`751e9ab80db16a625a953ac264e2490210ed4f44`), the focused tests-first matrix produced **2 failures and 4 passes**. Both target files remained unchanged through current `origin/main` (`bffd66c0524eba231e87da1f2a217e7be016d412`), and a fresh ownership scan found no open ASI PR touching either path; PR #470 changes the separate `core/horde.py` pair.

## Expected behavior

Require `step_size` to be finite and positive at `StackedHordeConfig` construction, before serialization or any JAX update. Preserve the existing error message and all legal finite positive step sizes, including the default `0.05`.

Existing TD(λ) equations, per-demon semantics, fail-closed runtime safety, scaling behavior, and serialization should remain unchanged.

## Scope

- `alberta_framework/core/stacked_horde.py`
- `tests/test_stacked_horde.py`

This is one constructor predicate and a table-driven regression. It does not edit update equations, defaults, Horde combinatorics, evidence sources, benchmarks, or any `outputs/**` artifact. Neither target path is registered in the current five-claim evidence source inventory.

## Contribution provenance

- AI assistance: yes

ss251:
Closes #478.

## What changed

`StackedHordeConfig` now rejects non-finite `step_size` values before they can be serialized into a Horde that silently rolls back every update.

The table-driven regression covers NaN and both infinities. Existing zero/negative rejection, legal finite positive values, the default `0.05`, TD(λ) equations, and runtime transaction guards remain unchanged.

## Scope

- `alberta_framework/core/stacked_horde.py`
- `tests/test_stacked_horde.py`

## Verification

Exact candidate head: `4acc03eefc6338d0976cd6d43395bfd7de4e6a2a`  
Base: `bffd66c0524eba231e87da1f2a217e7be016d412`

- concrete reproduction before the fix:
  - NaN and positive-infinite step sizes construct successfully;
  - the next valid update rolls back, leaves finite zero weights, and does not advance the step count;
- tests-first config baseline: **2 failed, 4 passed**;
- final config suite: **6 passed**;
- full stacked-Horde suite: **15 passed**;
- full Ruff: **all checks passed**;
- strict mypy: **no issues in 165 source files**;
- `git diff --check`: clean;
- rebased onto current `origin/main` (`bffd66c0524eba231e87da1f2a217e7be016d412`);
- exact-head proof rerun after rebase: full stacked-Horde suite **15 passed**;
- live path ownership refresh: no open ASI PR touches either target file; PR #470 owns the separate `core/horde.py` pair.

## Scientific integrity

This is fail-closed constructor validation only. It does not change equations, thresholds, defaults, TD traces, update behavior for legal configs, RNG schedules, registered evidence sources, frozen protocols, scientific claims, benchmark results, `CHANGELOG.md`, or any `outputs/**` artifact.

## Contribution provenance

- AI assistance: yes

### DIFF
--- a/alberta_framework/core/stacked_horde.py
+++ b/alberta_framework/core/stacked_horde.py
@@ -36,6 +36,7 @@ inactive for that step (its trace still decays; its weights freeze), matching
 the NaN-masking convention of the loop-based hordes.
 """
 
+import math
 from dataclasses import dataclass
 from typing import Any
 
@@ -92,7 +93,7 @@ class StackedHordeConfig:
             raise ValueError("every gamma must be in [0, 1]")
         if any(not 0.0 <= la <= 1.0 for la in self.lamdas):
             raise ValueError("every lamda must be in [0, 1]")
-        if self.step_size <= 0.0:
+        if not math.isfinite(self.step_size) or self.step_size <= 0.0:
             raise ValueError("step_size must be positive")
 
     def to_config(self) -> dict[str, Any]:
--- a/tests/test_stacked_horde.py
+++ b/tests/test_stacked_horde.py
@@ -11,6 +11,7 @@ the same workload, with a ~144 s compile; see the scaling notes in
 docs/evidence/methodology.md).
 """
 
+import math
 import time
 
 import chex
@@ -50,6 +51,11 @@ class TestConfig:
             _simple_config(gammas=(1.5, 0.9))
         with pytest.raises(ValueError, match="step_size"):
             _simple_config(step_size=0.0)
+
+    @pytest.mark.parametrize("step_size", [math.nan, math.inf, -math.inf])
+    def test_step_size_must_be_finite_and_positive(self, step_size: float) -> None:
+        with pytest.raises(ValueError, match="step_size"):
+            _simple_config(step_size=step_size)
 
     def test_roundtrip(self):
         cfg = _simple_config()


shabbydev (approved):
# Review: PR #479 — fix(stacked-horde): reject non-finite step sizes

**Verdict: APPROVED** (static review only — execution proof blocked: no JAX/pytest venv available in this environment; the PR's own claimed suite results were not independently rerun.)

## 1. Defect verification (pre-fix base `bffd66c0`)

The claimed defect is real. At base commit `bffd66c0`, `alberta_framework/core/stacked_horde.py:95` validates only:

```python
if self.step_size <= 0.0:
    raise ValueError("step_size must be positive")
```

Since every comparison with NaN is False, `step_size=nan` passes this check; `+inf` passes too (`inf > 0`). Only `-inf` is caught.

The downstream silent-failure mechanism claimed in the PR body also checks out. In `StackedLinearHorde.update()` (base, lines ~290-292):

```python
proposed_weights = (
    state.weights
    + cfg.step_size * masked_delta[:, None] * proposed_traces
)
```

With `step_size=nan`, `nan * 0.0` is `nan` even for demons masked to zero delta, so every demon's proposed weights are non-finite; `candidate_finite` (base line ~301) is False everywhere, `head_updates_applied` is all-False, `update_applied` is False, and the final `jax.lax.cond(update_applied, ...)` returns the *old* state — weights frozen at zero, `step_count` never advances, and nothing raises. Same for `+inf` (`inf * 0.0 = nan`, `inf * x = inf`). This matches the runtime fail-closed rollback introduced by `b307855 fix(core): make nonfinite updates fail closed`, which explains why the failure is silent rather than loud.

## 2. Fix correctness, completeness, scoping

The diff changes the guard to (head `4acc03ee`, `alberta_framework/core/stacked_horde.py:96`):

```python
if not math.isfinite(self.step_size) or self.step_size <= 0.0:
    raise ValueError("step_size must be positive")
```

- **Correct:** `math.isfinite` rejects NaN and both infinities; accepts ints and numpy floats via `__float__`. Fail-closed: `ValueError` at construction, before any serialization or update.
- **Entry-path completeness — all covered:**
  - Direct `StackedHordeConfig(...)` construction → `__post_init__` (the fixed line).
  - `StackedHordeConfig.from_config()` → `return cls(**config)` (head line ~117) → `__post_init__`.
  - `nexting_spec()` → constructs `StackedHordeConfig(..., step_size=step_size)` (head line ~155) → `__post_init__`.
  - `StackedLinearHorde(config)` / `StackedLinearHorde.from_config()` take only a config object; no independent step-size parameter exists.
  - `update()` reads `cfg.step_size` only (head line 292); no other injection point. `checkpoints.py` has no stacked-horde path (`git grep stacked` on `checkpoints.py` is empty at head).
  - Config is a `frozen=True` dataclass, so post-construction mutation is not a normal path.
- **Scoping:** The diff is 3 functional lines in the source file (`import math` + guard change) plus one parametrized test. Equations, defaults (`step_size: float = 0.05`), TD trace logic, and runtime masks are untouched — confirmed by diffing head against base.
- **Non-duplicate:** current `origin/main` (`d06d51b`) still carries the unfixed `if self.step_size <= 0.0:` at line 95; no other open PR touches `alberta_framework/core/stacked_horde.py` (checked via `gh pr list`; only #479 matches).

## 3. Test coverage

The added test (`tests/test_stacked_horde.py`, head):

```python
@pytest.mark.parametrize("step_size", [math.nan, math.inf, -math.inf])
def test_step_size_must_be_finite_and_positive(self, step_size: float) -> None:
    with pytest.raises(ValueError, match="step_size"):
        _simple_config(step_size=step_size)
```

Covers all three non-finite cases through the real constructor. Existing `test_validation` keeps the `step_size=0.0` case; `test_roundtrip` exercises `from_config`. Coverage is adequate for the change.

## 4. Minor observations (non-blocking)

- The error message `"step_size must be positive"` is slightly imprecise for NaN/inf inputs (they aren't rejected for non-positivity). Changing it could break downstream `match="step_size"` expectations; leaving it is defensible. Optional nit only.
- No test asserts the `from_config` path rejects a serialized non-finite step size, but that path provably funnels through the same `__post_init__`, so this is not a gap worth blocking on.

## 5. Execution proof

**Blocked.** Static review only; no JAX/pytest environment here. The PR body's claimed results ("final config suite 6 passed", "full stacked-Horde suite 15 passed", ruff/mypy clean) are taken as self-reported and were not independently verified.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

PR #480, opened 2026-08-16
fix(features): bias-correct EMA utility outputs

### LINKED ISSUE #431: future_utility_normalization="age" over-corrects (up to 200x on reset, 3.8x for hundreds of steps) and flips which slot is discarded, instead of putting young features on the same scale
## Summary (measurement contract question — no PR yet; the correction placement is a design choice)

`normalize_future_utility` (`alberta_framework/core/future_utility.py`) documents `"age"` as: "debiases the current EMA warm-up so young features are compared against older features on the same scale." The code divides the **input signal** by `1 - decay**(age+1)` and then feeds it into the utility EMA (`feature_discovery.py`: `ema = decay * old + (1 - decay) * signal`). The standard warm-up correction (Adam-style) divides the **EMA output**, not its increment. Pre-dividing is exact only at `age == 0`; from then on each already-debiased contribution is retained *and* a freshly full-strength debiased term is added, so the correction compounds.

Consequences observable through the feature-discovery ranking (both `FixedBudgetFeatureLearner` and `FixedBudgetInteractionLearner` share the helper):

- Under a constant unit signal with `utility_decay=0.995` the reported utility, which should read `1.0` at every age, reads `2.93` at age 10, peaks at `3.83` at age 59, and is still `2.35` at age 300 — a 2.5–3.8× over-correction for hundreds of steps.
- Two byte-identical features seeing identical signals, one at age 0 and one at age 5000, are scored `1.5e-2` vs `7.6e-5` — **200×** in favour of the young one on the first step after a reset.
- Replacement flips: with `replacement_interval=1, min_feature_age=0`, `norm='none'` discards slot 0 (tie), `norm='age'` discards slot 1 — the *mature* identical feature — so the mechanism whose stated purpose is to stop young features being unfairly discarded instead discards the mature one. `promote_branch` compounds this: it sets `age→0` but inherits the candidate's warm EMA, so the debias is applied on top of an already-warm value.

```
constant signal s=1.0, utility_decay=0.995, mode='age'
 u at ages 0,1,2,5,10,50,100,200,500,1000,1999: [1.0, 1.4963, 1.8238, 2.4157, 2.9305, 3.8172, 3.6807, 2.9763, 1.5909, 1.0653, 1.0007]
 (a correctly debiased EMA of a constant signal would be exactly 1.0 at every age)

A) two BYTE-IDENTICAL features, one update; slot0 age=0, slot1 age=5000; replacement off
   norm='none'  utilities=[7.5909e-05 7.5909e-05]   young/old = 1.0x
   norm='age'   utilities=[1.5182e-02 7.5909e-05]   young/old = 200.0x
C) replacement (replacement_interval=1, min_feature_age=0)
   norm='none'  replaced_slot=0 (tie -> slot 0)
   norm='age'   replaced_slot=1 (the MATURE identical feature is discarded)
```

## Options

- (a) Apply the bias correction to the EMA output at ranking time (`utilities / max(1 - decay**age, eps)`) rather than scaling the increment; or, equivalently,
- (b) drop the `(1 - decay)` factor when the debias is pre-applied (so the correction is exact for a constant signal at every age).

Either yields an age-independent utility for a constant signal and equal ranks for identical features of different ages; happy to implement whichever the maintainers prefer, failing-test-first (`tests/test_feature_discovery.py`).

## Dedup

`gh issue/pr list --search "normalize_future_utility" / "future_utility_normalization" / "age normalization" / "debias"`: nothing. Related open (mine, different mechanism in the same module): #379/#380.

lalalune:
Closes #431

This fixes the reported EMA warm-start bias in both actual consumers of shared future-utility normalization: FixedBudgetFeatureLearner and CompositionalFeatureLearner.

The raw EMA remains in the existing state/checkpoint schema. Ranking, curation, selection, resource scores, and metrics now consume an eager/JIT/grad-safe output correction. The EMA decay and complement are canonicalized once in the float32 execution domain, and decay values that collapse to one are rejected before construction. Age-mode promotion resets both raw fast and slow utilities consistently; non-age promotion and FixedBudgetInteractionLearner compatibility remain unchanged.

Failing-first evidence reproduced the original output sequence [200.0002, 18.6409, 1.0] and a 16.1% full-learner score error near decay 0.9999999. After repair, 160 relevant tests, 92 compatibility tests, and 115 current-main composition tests passed. Full Ruff and strict mypy passed; the rebased patch ID is byte-equivalent to the independently reviewed candidate.

No outputs or registered five-claim evidence sources are changed. The existing nonpromoting slowly-changing-regression plan becomes source-stale and therefore continues to fail closed until explicitly replaced.

### DIFF
--- a/alberta_framework/core/compositional_features.py
+++ b/alberta_framework/core/compositional_features.py
@@ -33,6 +33,8 @@ from jax import Array
 from jaxtyping import Bool, Float, Int, PRNGKeyArray
 
 from alberta_framework.core.future_utility import (
+    bias_correct_future_utility,
+    canonical_float32_ema_decay,
     contribution_trace_output_loss_reduction,
     normalize_future_utility_signal,
     trace_output_loss_reduction,
@@ -1036,8 +1038,11 @@ class CompositionalFeatureLearner:
             raise ValueError("n_tasks must be positive")
         if candidate_count < 0:
             raise ValueError("candidate_count must be non-negative")
-        if not 0.0 <= utility_decay < 1.0:
-            raise ValueError("utility_decay must be in [0, 1)")
+        utility_decay_config = utility_decay
+        utility_decay = canonical_float32_ema_decay(
+            "utility_decay",
+            utility_decay,
+        )
         if replacement_interval < 0:
             raise ValueError("replacement_interval must be non-negative")
         if not 0.0 <= promotion_blend <= 1.0:
@@ -1133,8 +1138,11 @@ class CompositionalFeatureLearner:
             and candidate_selector_exploration <= 0.0
         ):
             raise ValueError("exp3 candidate_selector requires positive exploration")
-        if not 0.0 <= retention_slow_utility_decay < 1.0:
-            raise ValueError("retention_slow_utility_decay must be in [0, 1)")
+        retention_slow_utility_decay_config = retention_slow_utility_decay
+        retention_slow_utility_decay = canonical_float32_ema_decay(
+            "retention_slow_utility_decay",
+            retention_slow_utility_decay,
+        )
         if retention_tanh_min_count < 0:
             raise ValueError("retention_tanh_min_count must be non-negative")
         if retention_product_min_count < 0:
@@ -1178,6 +1186,7 @@ class CompositionalFeatureLearner:
         self._step_size_output = step_size_output
         self._step_size_theta = step_size_theta
         self._utility_decay = utility_decay
+        self._utility_decay_config = utility_decay_config
         self._replacement_interval = replacement_interval
         self._min_feature_age = min_feature_age
         self._candidate_min_age = candidate_min_age
@@ -1223,6 +1232,9 @@ class CompositionalFeatureLearner:
             )
         )
         self._retention_slow_utility_decay = retention_slow_utility_decay
+        self._retention_slow_utility_decay_config = (
+            retention_slow_utility_decay_config
+        )
         self._retention_tanh_min_count = retention_tanh_min_count
         self._retention_product_min_count = retention_product_min_count
         self._operation_prior = operation_prior
@@ -1284,7 +1296,7 @@ class CompositionalFeatureLearner:
             "candidate_count": self._candidate_count,
             "step_size_output": self._step_size_output,
             "step_size_theta": self._step_size_theta,
-            "utility_decay": self._utility_decay,
+            "utility_decay": self._utility_decay_config,
             "replacement_interval": self._replacement_interval,
             "min_feature_age": self._min_feature_age,
             "candidate_min_age": self._candidate_min_age,
@@ -1327,7 +1339,9 @@ class CompositionalFeatureLearner:
                 self._candidate_selector_learning_rate
             ),
             "candidate_selector_exploration": self._candidate_selector_exploration,
-            "retention_slow_utility_decay": self._retention_slow_utility_decay,
+            "retention_slow_utility_decay": (
+                self._retention_slow_utility_decay_config
+            ),
             "retention_tanh_min_count": self._retention_tanh_min_count,
             "retention_product_min_count": self._retention_product_min_count,
             "operation_prior": (
@@ -2032,6 +2046,16 @@ class CompositionalFeatureLearner:
         decay = jnp.asarray(self._retention_slow_utility_decay, dtype=jnp.float32)
         return decay * previous_slow_utility + (1.0 - decay) * utility_signal
 
+    def _utility_update(self, previous_utility: Array, utility_signal: Array) -> Array:
+        """Update the raw utility EMA with one float32 decay/complement pair."""
+        if self._utility_decay == 0.0:
+            return utility_signal
+        decay = jnp.asarray(self._utility_decay, dtype=jnp.float32)
+        return (
+            decay * previous_utility
+            + (jnp.asarray(1.0, dtype=jnp.float32) - decay) * utility_signal
+        )
+
     def _retention_score(
         self,
         fast_utility: Array,
@@ -2041,6 +2065,20 @@ class CompositionalFeatureLearner:
         if self._retention_slow_utility_decay == 0.0:
             return fast_utility
         return jnp.maximum(fast_utility, slow_utility)
+
+    def _ranking_utility(
+        self,
+        raw_utility: Array,
+        ages: Array,
+        decay: float,
+    ) -> Array:
+        """Return curation/report scores without changing serialized EMA state."""
+        mode = (
+            self._future_utility_normalization
+            if self._future_utility_mix > 0.0
+            else "none"
+        )
+        return bias_correct_future_utility(raw_utility, ages, decay, mode)
 
     def _energy_normalized_residual_score(
         self,
@@ -2236,6 +2274,12 @@ class CompositionalFeatureLearner:
             utilities = jnp.ones_like(existing_depth, dtype=jnp.float32)
         else:
             utilities = existing_utilities
+            if existing_ages is not None:
+                utilities = self._ranking_utility(
+                    utilities,
+                    existing_ages,
+                    self._utility_decay,
+                )
         parent_logits = self._parent_logits(
             eligible,
             utilities,
@@ -2396,9 +2440,14 @@ class CompositionalFeatureLearner:
             depth_ok = depth_c + 1 <= self._max_depth
             eligible = alive & depth_ok
             any_eligible = jnp.any(eligible)
+            ranking_utils = self._ranking_utility(
+                utils_c,
+                ages_c,
+                self._utility_decay,
+            )
             logits = self._parent_logits(
                 eligible,
-                utils_c,
+                ranking_utils,
                 feature_values=feature_values,
                 feature_credit=feature_credit,
                 depth=depth_c,
@@ -2408,7 +2457,7 @@ class CompositionalFeatureLearner:
             partner_logits = self._partner_logits(
                 eligible,
                 depth_c,
-                utils_c,
+                ranking_utils,
                 ages=ages_c,
                 parent_mode=parent_mode,
             )
@@ -2662,13 +2711,7 @@ class CompositionalFeatureLearner:
                 state.feature_score_residual_trace,
                 state.feature_score_energy_trace,
             )
-        if self._utility_decay == 0.0:
-            new_utilities = utility_signal
-        else:
-            new_utilities = (
-                self._utility_decay * state.utilities
-                + (1.0 - self._utility_decay) * utility_signal
-            )
+        new_utilities = self._utility_update(state.utilities, utility_signal)
         retention_slow_utilities = self._retention_slow_utility(
             state.retention_slow_utilities,
             utility_signal,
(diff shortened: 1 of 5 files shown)


shabbydev (approved):
# Review: PR #480 — fix(features): bias-correct EMA utility outputs

**Verdict: APPROVED** (static review; execution proof blocked — no JAX/pytest venv available in this environment, nothing was run).

Reviewed against `elizaOS/asi` PR head `3cae550` (fetched via `pull/480/head`), base merge `282971b`, and issue #431.

## 1. Claimed defect — verified to exist in pre-fix code

The pre-fix `normalize_future_utility_signal` divided the **input signal** by the warm-up factor before it entered the utility EMA (base `282971b:alberta_framework/core/future_utility.py:429-431`):

```python
age_float = jnp.maximum(ages.astype(jnp.float32), 0.0) + 1.0
debias = 1.0 - jnp.power(utility_decay_arr, age_float)
normalized = normalized / jnp.maximum(debias, 1e-3)
```

The EMA then accumulated `m_t = d·m_{t-1} + (1-d)·(x_t / (1-d^t))`. Standard bias-corrected EMA semantics (Adam-style) divide the EMA **output** by `1 - d^t`; pre-dividing each increment is exact only at the first step. Hand-check with constant signal `x=1`, `d=0.5`: `m_1 = 1.0` (fine), but `m_2 = 0.5·1.0 + 0.5·(1/0.75) = 1.167` and `m_3 ≈ 1.155` — a persistent overshoot, because each already-debiased contribution is retained while a freshly full-strength debiased term is added. With `d` near 1 this compounds for hundreds of steps and amplifies young-age spikes by up to `1/(1-d^t)`, matching issue #431's measured falsifier (constant unit signal reading 2.93 at age 10 / peaking 3.83 at age 59; 200× young/old ranking ratio; mature identical feature discarded instead of the young one). The defect is real, and both consumers cited in the PR body (`FixedBudgetFeatureLearner` in `feature_discovery.py`, `CompositionalFeatureLearner` in `compositional_features.py`) are the only callers of the helper — verified by grep: the 4 non-test call sites are `feature_discovery.py:1076,1156` and `compositional_features.py:2686,2797`.

## 2. The fix — math, completeness, scope

**Math (`future_utility.py:457-483`, new `bias_correct_future_utility`):**

```python
age_count = jnp.maximum(ages.astype(jnp.float32), 0.0)
decay = jnp.asarray(utility_decay, dtype=utilities.dtype)
debias = 1.0 - jnp.power(decay, age_count)
safe_debias = jnp.maximum(debias, jnp.asarray(1e-30, ...))
corrected = utilities / safe_debias
return jnp.where(age_count > 0.0, corrected, utilities)
```

This is exactly the canonical debias `m̂ = m_t / (1 - d^t)` applied to the raw EMA output. Timing is correct: in both learners `ages = state.ages + 1` (`feature_discovery.py:1201`, `compositional_features.py:2870`) is computed before correction, so `ages` equals the number of EMA updates applied to `new_utilities`. Initialization is zero (`feature_discovery.py` `init()`: `utilities=jnp.zeros(...)`, `ages=jnp.zeros(...)`), so for a constant signal the corrected value is exactly `x` at every age ≥ 1 — the property issue #431 demands. Age-0 slots (no observations) return the raw value via the final `jnp.where`, avoiding 0/0; the `1e-30` floor keeps the unselected branch finite under JIT/grad. Since the constructor now rejects decays that narrow to float32 `1.0` (below), `1 - d^age ≥ 1 - d > 0` for age ≥ 1, so no blow-up path exists.

**Signal path corrected (`future_utility.py:423-456`):** age correction removed from `normalize_future_utility_signal` (`del ages, utility_decay` at line 444; docstring explicitly documents that age correction now belongs on the EMA output); only the causal RMS uncertainty normalization remains. Signature kept for caller/config compatibility.

**Consumers traced.** Corrected scores feed every ranking/curation/report site while raw EMA stays in serialized state:
- `feature_discovery.py`: replacement pick `active_scores` (1245), candidate best/worst (1252, 1255), promotion margin compare (1264-1265), generator parent selection in `promote_branch`/`refresh_candidate_branch`/`replace_active_branch` (via `parent_utilities = bias_correct_future_utility(util, age, ...)`), resource policy scores (1621-1623), pressure scores (1649, 1658-1659), metrics mean/min/max (1777, 1786-1787). Raw `new_utilities`/`new_candidate_utilities` are stored in `FeatureDiscoveryState` (1737, 1748) and only appear otherwise as `lax.cond` carry passthroughs (1500-1523, 1578-1591) — correct.
- `compositional_features.py`: `_ranking_utility` wrapper (added ~2069) applied in `_generate_one` (~2277), `refill_step` parent/partner logits (~2443-2460), retention/replacement scores (`update` ~2974, candidate promotion ~3077), refresh scores (~3116), `replace_active_branch` parent/partner logits (~3735, ~3749), generator policy + promotion credit + metrics in `keep_active_branch` (~4213-4345). Slow-utility EMAs are debiased with their own decay (`_retention_slow_utility_decay`), which is correct since they are independent zero-init EMAs over the same age counter.
- **Promotion reset:** age-mode promotion now resets the raw fast EMA to 0.0 alongside the age reset — `compositional_features.py` `promote_branch` (~3206-3215: `promoted_utility = 0.0 if mix>0 and mode in {"age","uncertainty_age"} else cutil_a[best_candidate]`), slow utility in `keep_active_branch` (~4057-4065), and `feature_discovery.py` `promote_branch` (diff hunk at ~1344, with an explanatory comment). This fixes issue #431's "promote_branch compounds this" point — inheriting a mature EMA at age 0 would double-apply the correction. Candidate rebound resets (age+utility zeroed together, ~4016-4025) already keep age↔EMA consistency.
- **Float32 canonicalization (`future_utility.py:26-38`, `canonical_float32_ema_decay`):** decays are narrowed to float32 at construction and values rounding to `1.0` are rejected, so the execution-domain decay and its complement can never disagree (the 0.9999999-class divergence the PR body attributes a 16.1% score error to) and `debias` can never be exactly 0. All four sinks converted: `utility_decay` and `retention_slow_utility_decay` (compositional), `utility_decay` and `utility_retention_decay` (feature_discovery), with original values preserved via `_..._config` for `to_config()` round-trip (checkpoint/config schema unchanged; state NamedTuples untouched).
- **Scope check:** `FixedBudgetInteractionLearner` (`interaction_features.py:314`) never calls the helper and applies no age correction (plain EMA at 1798/1850) — the PR body's "compatibility unchanged" claim holds, and no consumer of the buggy path is left uncorrected.

Minor, non-blocking observations:
- `feature_discovery.py` `keep_active_branch` computes `ranking_utilities` twice (inside `if self._learn_feature_resources:` at ~1608 and unconditionally at ~1718). Redundant but JIT-harmless.
- `_utility_update`'s retention path is `max(ema, retention_decay·old)` (a decaying peak holder, not a second full EMA), so debiasing the blend with the fast decay is an approximation — negligible in practice since the retained term decays toward 0 while the EMA warms up.
- Age-0 slots rank at raw value (~0), so with `min_feature_age=0` a brand-new slot is eviction-eligible immediately; this is the design issue #431 itself proposed (option (a)) and is documented in the new docstring.

## 3. Test coverage

Tests pin the corrected behavior, not just shapes:
- `test_age_normalization_does_not_scale_the_ema_increment` — asserts the signal passes through unchanged in `"age"` mode (the core semantic flip).
- `test_age_correction_reads_constant_raw_emas_on_one_scale` — feeds exact raw EMAs `1 - 0.995^age` for ages {1,2,11,60,301,5001} and asserts corrected == 1.0 under `jax.jit`, with finite `jax.grad` — this is precisely the issue's acceptance criterion ("exactly 1.0 at every age").
- `test_age_zero_utility_reset_is_finite_under_jit_and_grad` — pins the age-0 edge under both JIT and grad.
- `test_feature_learner_ema_and_age_correction_share_float32_decay` / compositional twin — decay `0.9999999` vs control `0.0` must produce identical corrected metrics `[2:4]` after one JIT update (the 16.1%-error regression), and assert `to_config()` returns the original config value.
- Parametrized `test_utility_ema_decay_rejects_values_narrowing_to_float32_one` — all four constructor sinks reject `0.99999999`.
- `test_age_corrected_promotion_restarts_raw_utility_ema` (test_feature_discovery.py:592) — end-to-end promotion under age mode asserts `ages[0]==0`, `utilities[0]==0.0`, `candidate_utilities[0]==0.0`, and finite metrics.

Gap: no compositional-level state-reset assertion mirroring the last test (mechanism is symmetric and unit-covered), and no replay of the issue's 200× two-slot scenario — acceptable given the math-level pinning above.

## 4. Execution proof

Blocked — no JAX/venv in the review environment; pytest/ruff/mypy were not run here. The PR body reports 160 + 92 + 115 tests passing plus Ruff/strict-mypy, but this review is static-only.

## Verdict rationale

The pre-fix defect is real and independently reproduced by hand; the fix implements the textbook bias-corrected EMA exactly (output-side `1/(1-d^t)`, correct age accounting, zero-init verified), applies it at every ranking/report consumer of both learners while leaving serialized state schema untouched, handles the age-reset/promotion interaction consistently, hardens the float32 decay domain, and adds tests that pin the exact corrected semantics. No defect found. APPROVED.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

deepanshu-yd:
## Review (comment)

A few additional observations on top of the existing approval — all non-blocking.

1. **Compositional-level promotion reset gap (echoing the existing review's gap note, making it concrete):** `test_age_corrected_promotion_restarts_raw_utility_ema` covers `FixedBudgetFeatureLearner` end-to-end, but `CompositionalFeatureLearner` has two separate promotion sites — `promote_branch` (zeroing fast EMA, ~3206) and `keep_active_branch` (zeroing `promoted_retention_slow_utility`, ~4057) — plus candidate-rebound resets. A single end-to-end compositional test asserting `ages==0`, `utilities==0.0`, `retention_slow_utilities==0.0` after an age-mode promotion would guard both sites against the double-correction bug the PR is designed to prevent.

2. **Decay canonicalization domain:** `canonical_float32_ema_decay` accepts any value narrowing into `[0, 1)` including `0.0`. With `utility_decay=0.0`, `_utility_update` replaces the EMA with the signal, and `bias_correct_future_utility` divides by `1 - 0^age = 1`, so age-mode correction degenerates to identity. The control test (`utility_decay=0.0`) pins this equivalence, which is good — just confirming the degenerate case is intended and not silently relied upon elsewhere.

3. **Metric semantics shift:** metrics `mean`/`min`/`max_candidate_utility` now report bias-corrected ranking values rather than the raw EMA. If any downstream run summary or artifact consumers those metrics, their numerical values shift under age modes. This is presumably intended (the corrected values are the meaningful ones), but it changes reported-number semantics — worth a line in the commit message or a runbook note if any summary artifact comparisons depend on it.

4. **`del ages, utility_decay` in `normalize_future_utility_signal`:** keeping the parameters for signature compatibility while deleting them is slightly fragile — a future caller passing them will silently get no age correction. The docstring covers it; consider a `FutureWarning` or explicit deprecation if the helper is part of a public surface.

The core fix (output-side `1/(1-d^t)` bias correction, promotion resets, float32 decay canonicalization, legacy-config round-trip) is verified and correct. LGTM.

PR #476, opened 2026-08-16
fix(streams): Step 1 & Step 2 benchmark streams parameter validation and signed int32 bounds

deepanshu-yd:
Closes #475

### Summary of Changes
- **Step 1 Streams (`alberta_framework/streams/alberta_plan_step1.py`):**
  - Added strict signed int32 validation (`[1, 2**31 - 1]`) for `feature_dim`, `num_relevant`, and `scale_change_interval`.
  - Added exact-ratio float32 domain validation on `drift_rate_w`, `drift_rate_b`, `noise_std`, `feature_std`, `scale_min`, `scale_max`.
  - Added strict boolean checking for `noise_in_target`.
  - Exposed complete read-only properties for all stream configuration parameters.
- **Step 2 Feature Discovery Streams (`alberta_framework/streams/feature_discovery.py`):**
  - Added signed int32 bounds checking for `feature_dim`, `n_tasks`, `n_latents`, `n_contexts`, `context_length`, `active_latents_per_context`, `active_pairs_per_context`, and `num_steps` in `collect_feature_discovery_stream`.
  - Added exact-ratio float32 validation for `feature_std`, `latent_scale`, `linear_scale`, `noise_std`.
  - Added strict boolean checking for `include_squares`.
  - Added duck-typing validation for `stream` argument in `collect_feature_discovery_stream`.
  - Exposed configuration properties on all stream classes.
- **Testing (`tests/test_alberta_plan_step1_streams.py`, `tests/test_feature_discovery.py`):**
  - Added `TestStep1StreamsValidation` and `TestFeatureDiscoveryStreamsValidation` covering parameter boundaries, illegal type rejections, class-spoofed floats, adversarial exact ratios, and property getters.

PR #426, opened 2026-08-16
fix(step1): reject illegal linear TD facade config

deepanshu-yd:
Closes #425.

## What changed

`Step1KernelConfig` and `run_step1_smoke` now strictly validate scalar hyperparameters, float32 execution sink bounds, and exact integer ratios before constructing Step 1 components. Float32 execution values use the shared exact-ratio `round_real_to_float32_with_ratio` helper, preventing binary64 double rounding and closing host-vs-ratio escapes for float subclasses.

Exact float32 overflow is rejected across `step_size`, `meta_step_size`, `drift_rate_w`, `drift_rate_b`, `noise_std`, `feature_std`, `ema_decay`, and `streaming_batch_momentum`; legal zero/one endpoints remain valid. Integer feature dimensions, relevant counts, and smoke parameters (`steps`, `seed`, `final_window`) are bounded to standard signed int32 bounds (`2**31 - 1`).

## Scope

• `alberta_framework/_float32.py`
• `alberta_framework/steps/step1.py`
• `tests/test_production_steps.py`

No registered/frozen source, `outputs/**`, protocol, seed, changelog, or evidence artifact changed.

## Exact-head verification

• Step 1 focused test suite: 154 passed;
• full Ruff: passed;
• strict mypy: no issues in 164 source files;
• `git diff --check`: clean.

## Scientific integrity

This is facade input validation, not scientific evidence or a performance claim.

## Contribution provenance

• AI assistance: yes
• AI provider/model: google / gemini-3.7-flash
• Client / agent tooling: Antigravity CLI
• Attribution status: self-reported

PR #368, opened 2026-08-16
fix(step1): reject non-integer smoke counts

Svector-anu:
Closes #365.

## What changed

Same pattern as #347/#356: `run_step1_smoke` validated `steps` and `final_window` with bare comparisons (`steps < 1`, `final_window < 1 or final_window > steps`) even though this module defines and uses `_require_int` elsewhere to reject booleans and non-integral inputs. `bool` (an `int` subclass) and non-integral floats slip through the bare comparisons, and depending on the field/value combination, the facade either silently accepts a float count or leaks an implementation-level JAX/Python `TypeError` instead of its documented `ValueError`.

confirmed directly before touching the source:

```text
{'steps': True, 'final_window': True} -> TypeError: iota does not accept dtype bool. Accepted dtypes are subtypes of integer, complexfloating, floating.
{'steps': 4.0, 'final_window': 2} -> ACCEPTED steps=4.0, metrics_shape=(4, 4)
{'steps': 4, 'final_window': 2.0} -> TypeError: 'float' object cannot be interpreted as an integer
```

fix: validate and canonicalize both counts through the existing `_require_int(..., minimum=1)` helper before applying the cross-field `final_window > steps` constraint (the standalone `final_window < 1` check is now redundant with `_require_int(minimum=1)` and is dropped).

## Reachability

`run_step1_smoke` is in `alberta_framework/steps/__init__.py`'s `__all__` and the package-root `alberta_framework/__init__.py`'s `__all__` — both confirmed directly. It's also called by `alberta_framework/cli.py::step1_smoke_main`, the installed `alberta-step1-smoke` console entry point — argparse constrains normal CLI input to integers, but the two public Python exports remain reachable with the invalid inputs shown above.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_production_steps.py -q -o addopts=""
304 passed in 9.75s

$ .venv/bin/python -m ruff check alberta_framework/steps/step1.py tests/test_production_steps.py
All checks passed!

$ .venv/bin/python -m mypy
Success: no issues found in 164 source files
```

New test: `test_step1_smoke_rejects_non_integer_counts`, parametrized over `(steps, True)`, `(steps, 4.0)`, `(final_window, True)`, `(final_window, 2.0)`, asserting `ValueError` naming the offending field.

[...]

PR #482, opened 2026-08-16
fix(step1): forward XDist target-noise magnitude

### LINKED ISSUE #481: Step 1 XDist factory drops configured noise magnitude
## Summary

`make_step1_stream` drops the validated `Step1KernelConfig.noise_std` magnitude when `stream="xdist_shift"`.

The factory forwards only:

```python
noise_in_target=config.noise_std > 0.0
```

and omits `noise_std=config.noise_std`. As a result, every valid positive configuration silently runs with `XDistShiftStream`'s constructor default of `0.1`. For example, `Step1KernelConfig(stream="xdist_shift", noise_std=7.0)` reports `7.0` in its serialized config while the generated stream uses `0.1`. A zero setting also leaves the unused private magnitude at `0.1` even though target noise is disabled.

## Expected behavior

The production factory should forward both parts of the validated configuration:

- `noise_std=config.noise_std` preserves the requested magnitude;
- `noise_in_target=config.noise_std > 0.0` preserves the existing zero-disables-noise behavior.

## Scope

- `alberta_framework/steps/step1.py`
- one fast focused factory regression

No stream equation, RNG schedule, scientific evidence source, frozen artifact, benchmark result, or changelog entry needs to change.

## Reproduction

On the original tested `main` (`ca9aaaf`), a three-case regression for `noise_std=0.0`, `0.25`, and `7.0` failed all three cases because the constructed stream stored `0.1`. The candidate was then rebased onto current `origin/main` (`9e0719f8a92639fe8f78d08cb2fe267bf8ca1089`) and the affected Step 1 panel passed **321 tests** at exact head `fd0272af032dc2b31c3fc59b38906e2e32f0daaf`.

A live ownership refresh found no open PR implementing this factory-forwarding defect. PRs #368 and #426 also touch `steps/step1.py`, but only in integer/config validation and smoke-runner hunks; PR #476 owns the direct stream implementation files, not this factory path or focused test.

## Contribution provenance

- AI assistance: yes
- AI provider/model: OpenAI / gpt-5.6-sol
- Client / agent tooling: Claude Code via CLIProxyAPI

ss251:
Closes #481.

## What changed

The Step 1 production factory now forwards the validated `noise_std` magnitude into `XDistShiftStream` instead of forwarding only whether noise is enabled.

A fast table-driven regression covers zero, a fractional non-default magnitude, and a large non-default magnitude. Zero still disables target noise; positive values preserve both the requested magnitude and enabled state.

## Why

Before this patch, every positive `Step1KernelConfig(stream="xdist_shift", noise_std=...)` silently ran with the stream default `0.1`. The serialized configuration and generated trajectory therefore disagreed.

## Scope

- `alberta_framework/steps/step1.py`
- `tests/test_step1_stream_factory.py`

No stream equation, RNG schedule, scientific evidence source, frozen artifact, benchmark result, or changelog entry changed.

## Verification

Exact candidate head: `fd0272af032dc2b31c3fc59b38906e2e32f0daaf`  
Base: `9e0719f8a92639fe8f78d08cb2fe267bf8ca1089`

- tests-first baseline on the original `origin/main` `ca9aaaf`: **3 failed**;
- exact-head focused regression: **3 passed**;
- exact-head affected Step 1 panel: **321 passed** (`test_step1_stream_factory.py`, `test_production_steps.py`, and `test_alberta_plan_step1_streams.py`);
- affected-file Ruff: **all checks passed**;
- strict Step 1 source mypy: **passed**;
- `git diff --check`: clean;
- rebased onto current `origin/main` (`9e0719f8a92639fe8f78d08cb2fe267bf8ca1089`);
- exact-head proofs recorded for `fd0272af032dc2b31c3fc59b38906e2e32f0daaf`;
- live ownership refresh: no open PR implements this factory-forwarding defect; PRs #368 and #426 touch disjoint integer/config-validation hunks in `steps/step1.py`, and PR #476 owns the direct stream implementation files.

## Contribution provenance

- AI assistance: yes
- AI provider/model: OpenAI / gpt-5.6-sol
- Client / agent tooling: Claude Code via CLIProxyAPI

### DIFF
--- a/alberta_framework/steps/step1.py
+++ b/alberta_framework/steps/step1.py
@@ -326,6 +326,7 @@ def make_step1_stream(
         return XDistShiftStream(
             feature_dim=config.feature_dim,
             num_relevant=config.num_relevant,
+            noise_std=config.noise_std,
             noise_in_target=config.noise_std > 0.0,
         )
     msg = f"unknown Step 1 stream {config.stream!r}"
--- a/tests/test_step1_stream_factory.py
+++ b/tests/test_step1_stream_factory.py
@@ -1,0 +1,31 @@
+"""Fast contracts for the production Step 1 stream factory."""
+
+import numpy as np
+import pytest
+
+from alberta_framework.steps import Step1KernelConfig, make_step1_stream
+from alberta_framework.streams.alberta_plan_step1 import XDistShiftStream
+
+pytestmark = pytest.mark.unit
+
+
+@pytest.mark.parametrize(
+    ("noise_std", "noise_in_target"),
+    [(0.0, False), (0.25, True), (7.0, True)],
+)
+def test_xdist_factory_forwards_noise_configuration(
+    noise_std: float,
+    noise_in_target: bool,
+) -> None:
+    config = Step1KernelConfig(
+        feature_dim=4,
+        num_relevant=2,
+        stream="xdist_shift",
+        noise_std=noise_std,
+    )
+
+    stream = make_step1_stream(config)
+
+    assert isinstance(stream, XDistShiftStream)
+    assert stream._noise_std == float(np.float32(noise_std))  # noqa: SLF001
+    assert stream._noise_in_target is noise_in_target  # noqa: SLF001


lalalune (approved):
Reviewed exact head fd0272af032dc2b31c3fc59b38906e2e32f0daaf. The factory forwards the validated XDist noise magnitude while preserving zero-disables-noise behavior. Exact-head Step 1 panel: 321 passed; Ruff and strict source mypy passed; current-main merge tree and diff checks are clean; no registered evidence source or artifact changes. READY; merge only after exact-head CI is 20/20.

shabbydev (approved):
# Review: PR #482 — fix(step1): forward XDist target-noise magnitude

**Verdict: APPROVED**

## 1. Defect verification (base `9e0719f`, pre-fix)

I inspected the base source with `git show 9e0719f:alberta_framework/steps/step1.py`. The `xdist_shift` branch of `make_step1_stream` (~line 325) constructed the stream without forwarding the magnitude:

```python
if config.stream == "xdist_shift":
    return XDistShiftStream(
        feature_dim=config.feature_dim,
        num_relevant=config.num_relevant,
        noise_in_target=config.noise_std > 0.0,
    )
```

`XDistShiftStream.__init__` defaults `noise_std: float = 0.1` (`alberta_framework/streams/alberta_plan_step1.py:252`) and stores it verbatim (`self._noise_std = noise_std`, line 294). So any positive `config.noise_std` was silently coerced to 0.1. The claimed defect is real — and actually broader than stated: `Step1KernelConfig.noise_std` defaults to `1.0` (`steps/step1.py:217`), so even the **default** config ran at 0.1 target noise, not just non-default magnitudes. By contrast, the `alberta` branch already forwarded `noise_std=config.noise_std` (base step1.py:322), so the two branches were asymmetric. Confirmed.

## 2. Fix correctness, completeness, and boundary trace

The diff adds exactly one line to the factory:

```diff
@@ -326,6 +326,7 @@ def make_step1_stream(
         return XDistShiftStream(
             feature_dim=config.feature_dim,
             num_relevant=config.num_relevant,
+            noise_std=config.noise_std,
             noise_in_target=config.noise_std > 0.0,
         )
```

In the PR head (`fd0272a`), this is `alberta_framework/steps/step1.py:329`.

Boundary trace of the corrected value:
- `config.noise_std` is validated non-negative before use: `_validate_step1_config` runs `noise_std = _require_nonnegative_real("noise_std", config.noise_std)` (step1.py:176), which rejects negatives and non-finite values and narrows to a finite float32 (`_require_real`, step1.py:67–95). So the stream can never receive an invalid magnitude.
- Consumption: `XDistShiftStream.step` computes `eta = self._noise_std * jr.normal(k_eta, (), dtype=jnp.float32)` and gates it with `target = target + jnp.where(self._noise_in_target, eta, jnp.float32(0.0))` (`alberta_plan_step1.py:371–372`). The forwarded magnitude flows directly into the generated target noise.
- Zero case is consistent: `noise_std=0.0` → `noise_in_target=False` → `jnp.where` zeroes eta anyway; no discontinuity.
- Other `XDistShiftStream(` call sites are all test-local direct constructions (`tests/test_alberta_plan_step1_streams.py`, `tests/test_swift_td.py:429`) that intentionally exercise stream defaults — correctly untouched. `make_step1_stream` is the only production construction path (re-exported via `steps/__init__.py` and used by `run_step1_smoke`, step1.py:366).

Scope: `git diff 9e0719f..pr482 --stat` confirms exactly 2 files, +32/−0, matching the published diff. No unrelated hunks. The PR body's claim that no stream equation / RNG schedule changed is consistent with the diff (the stream file is untouched).

## 3. Test coverage

New file `tests/test_step1_stream_factory.py` pins the fixed behavior with three parametrized cases: `(0.0, False)`, `(0.25, True)`, `(7.0, True)` — zero-disable, fractional non-default, and large non-default magnitude. It asserts `stream._noise_std == float(np.float32(noise_std))` and `stream._noise_in_target is noise_in_target`.

Notes:
- The private-attribute access is justified: `XDistShiftStream` exposes no public `noise_std` property (only `feature_dim`/`num_relevant` properties exist), and the `# noqa: SLF001` markers acknowledge this.
- The `np.float32` cast matches config semantics: `_require_real` narrows values through `round_real_to_float32` and only returns a binary64 value when it re-narrows identically (step1.py:87–95), so the stored config value is float32-consistent. (0.25 and 7.0 are exactly representable in float32 anyway.)
- Non-blocking gap: the test pins constructor state, not sampled target statistics. That is acceptable here because the stream's noise math (`_noise_std * normal`, gated by `_noise_in_target`) is exercised in the existing `tests/test_alberta_plan_step1_streams.py` suite, and the factory is the only thing under test. An end-to-end trajectory assertion would be a nice-to-have, not required.
- Non-blocking note: the test asserts on the config→stream forwarding but doesn't cover negative `noise_std` rejection — that is already enforced upstream by `_require_nonnegative_real` and not part of this fix.

## 4. Execution proof

**Execution proof blocked.** Static review only — no JAX/pytest venv available in this environment, so I did not run the suite. The PR body reports a tests-first baseline (3 failed on base), 3-passed focused regression, and a 321-passed Step 1 panel at head `fd0272a`, but those are self-reported and were not independently reproduced here.

## Rationale

The defect exists exactly as described (and slightly worse for default config), the one-line fix forwards the validated value into the only production construction path, the value is consumed unchanged in the target-noise term, the zero/noise interplay stays consistent, scope is clean, and the new tests pin all three behavioral branches. No defects found.


- AI assistance: `yes`
- Agent tooling: `OpenClaw`

deepanshu-yd:
## Review (comment)

One additional note on top of the existing approvals.

**Default-config behavior change worth flagging:** the existing review already caught that the pre-fix `xdist_shift` branch silently ran at the stream default `noise_std=0.1`. Combined with `Step1KernelConfig.noise_std` defaulting to `1.0`, this means the fix changes the *default* Step 1 pipeline's target-noise level from `0.1` to `1.0` — not just non-default configs. That is the correct behavior (the configured magnitude should win), but it is a silent shift in any Step 1 smoke or development runs that relied on the implicit `0.1`. A release-note / runbook line naming this would prevent surprise diffs in downstream Step 1 comparisons. (No pinned evidence depends on it — Step 1 smoke is CLI-only — so this is documentation-only.)

**Nit:** the test asserts forwarding on `stream._noise_std` / `stream._noise_in_target` private state. As the existing review notes, this is justified (no public property). A follow-up could assert sampled target statistics (`target.std ≈ noise_std` at high sample count) to pin the end-to-end consumption, but the constructor-state pin plus the existing stream math coverage is sufficient.

The one-line fix is correct and complete. LGTM.

PR #486, opened 2026-08-16
fix(forager): reject non-integer analysis horizon

### LINKED ISSUE #483: forager_matched_final_analysis's evaluation-snapshot validator accepts a float alias for the plan horizon
Fifth instance of the same bug class as #450/#460/#466/#471: `_validate_evaluation_snapshot` in `alberta_framework/benchmarks/forager_matched_final_analysis.py` accepted a JSON float horizon when its numeric value matched the sealed protocol horizon (`499712.0 == 499712` is true in Python), letting an invalid execution-plan schema value cross the final-analysis closure boundary. Notably, this file already defines and uses `_json_exact_equal` for the sibling `active_seeds` field a few lines below — `horizon` was simply the one field that got missed.

confirmed directly, using a real fixture-derived plan with only `horizon` changed from the sealed integer to its float alias:
```text
BUG: accepted horizon=499712.0 (float); expected exact int 499712
```

reachability: `_validate_evaluation_snapshot` is invoked by `_load_from_open_root` and by the final-analysis creation flow. The public module API exports `load_final_analysis_content`, `create_forager_matched_final_analysis_bundle`, and `authenticate_final_analysis_content` in `__all__`; each leads to final-analysis content loading/validation with externally-supplied execution-plan content.

fix incoming: use the file's own existing `_json_exact_equal` helper for the horizon comparison, matching the pattern already applied to `active_seeds` in the same function.

Svector-anu:
Closes #483.

## What changed

Fifth instance of the same bug class as #450/#460/#466/#471: `_validate_evaluation_snapshot` accepted a JSON float horizon when its numeric value matched the sealed protocol horizon (`499712.0 == 499712` is true in Python), letting an invalid execution-plan schema value cross the final-analysis closure boundary. Notably, this file already defines and uses `_json_exact_equal` for the sibling `active_seeds` field a few lines below the fixed line — `horizon` was simply the one field that got missed.

confirmed directly before touching the source, using a real fixture-derived plan with only `horizon` changed from the sealed integer to its float alias:

```text
BUG: accepted horizon=499712.0 (float); expected exact int 499712
```

fix: use the file's own existing `_json_exact_equal` helper for the horizon comparison, matching the pattern already applied to `active_seeds` in the same function — a one-line change reusing established infrastructure rather than introducing anything new.

## Reachability

`_validate_evaluation_snapshot` is invoked by `_load_from_open_root` and by the final-analysis creation flow — both confirmed directly. The public module API exports `load_final_analysis_content`, `create_forager_matched_final_analysis_bundle`, and `authenticate_final_analysis_content` in `__all__` — confirmed directly — each leads to final-analysis content loading/validation with externally-supplied execution-plan content.

## Verification

macOS note: this test file's fixtures require the atomic no-replace publication primitives (`renameat2`, Linux-only) already documented as unavailable on this dev platform in #467/#472. I independently re-ran the full suite using the same test-only monkeypatch shim that swaps those primitives for plain `os.rename` during fixture setup — confirmed the shim only touches publication-primitive plumbing (`_publish_verified_no_replace`, `_rename_no_replace`, `_publish_bytes`), never anything related to the horizon check under test, before trusting it.

```text
$ PYTHONPATH=<shim-dir> .venv/bin/python -m pytest -p asi46_pytest_shim tests/test_forager_matched_final_analysis.py -q -o addopts=""
33 passed in 176.57s

$ .venv/bin/python -m ruff check alberta_framework/benchmarks/forager_matched_final_analysis.py tests/test_forager_matched_final_analysis.py
All checks passed!

$ .venv/bin/python -m mypy
```

New coverage: `("horizon", float(seal_content.sealed_protocol.horizon))` added to the existing `malformed_plan_cases` table inside `test_source_projection_and_command_templates_are_exact`, which already exercises the analogous `active_seeds` and `candidate_order` malformed-value cases.

mutation-resistance verified independently: reverted just the source fix (`git stash push -- alberta_framework/benchmarks/forager_matched_final_analysis.py`, kept the test), reran the parametrized test — failed with the float alias silently passing the horizon check and continuing on to an unrelated later assertion (`AssertionError: Regex pattern did not match. Expected: "execution-plan closure drifted" / Actual: "evaluation receipt-index header closure drifted"`), proving the float alias crossed the intended validation boundary undetected. restored the fix, green again.

## Evidence

1d7daa1a4490fd917e091fe31c7909046008ace1

```text
$ PYTHONPATH=<shim-dir> .venv/bin/python -m pytest -p asi46_pytest_shim tests/test_forager_matched_final_analysis.py -q -o addopts="" -k test_source_projection_and_command_templates_are_exact
FAILED - AssertionError: Regex pattern did not match.
Expected regex: 'execution-plan closure drifted'
Actual message: 'evaluation receipt-index header closure drifted'
1 failed, 32 deselected in 4.45s

green (fix restored):
$ PYTHONPATH=<shim-dir> .venv/bin/python -m pytest -p asi46_pytest_shim tests/test_forager_matched_final_analysis.py -q -o addopts="" -k test_source_projection_and_command_templates_are_exact
1 passed, 32 deselected in 4.39s
```

```text
BUG: accepted horizon=499712.0 (float); expected exact int 499712
```
independently reran the full touched test file (33 tests, using the documented macOS publication shim), ruff, and mypy myself; independently inspected the shim's source to confirm it only patches unrelated publication-primitive plumbing, not anything touching the fix under test; independently confirmed both real call sites and the `__all__` export chain before writing the reachability claim; did my own revert-and-confirm-red mutation check with the correct (parametrized-table, not separately-named) test target.

## Provenance

AI provider/model: openai / gpt-5.6-sol (fix, tests, via AgentRouter proxy); anthropic / claude-sonnet-5 (independent re-verification: reran the full touched test file using the same macOS compatibility shim after independently auditing what it patches, ruff, and mypy myself, independently confirmed both real call sites and the export chain, did my own revert-and-confirm-red mutation check, opened the governing issue, pushed, opened this PR)

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_final_analysis.py
+++ b/alberta_framework/benchmarks/forager_matched_final_analysis.py
@@ -2198,7 +2198,7 @@ def _validate_evaluation_snapshot(
         != qualification_manifest_sha256
         or executor_manifest.get("qualification_manifest_sha256")
         != qualification_manifest_sha256
-        or plan.get("horizon") != sealed_protocol.horizon
+        or not _json_exact_equal(plan.get("horizon"), sealed_protocol.horizon)
         or type(plan_active_seeds) is not list
         or not _json_exact_equal(plan_active_seeds, list(sealed_protocol.active_seeds))
         or type(plan_candidate_order) is not list
--- a/tests/test_forager_matched_final_analysis.py
+++ b/tests/test_forager_matched_final_analysis.py
@@ -2484,6 +2484,7 @@ def test_source_projection_and_command_templates_are_exact(
         )
 
     malformed_plan_cases: tuple[tuple[str, Any], ...] = (
+        ("horizon", float(seal_content.sealed_protocol.horizon)),
         ("active_seeds", None),
         (
             "active_seeds",


lalalune (approved):
Reviewed exact head 1d7daa1a4490fd917e091fe31c7909046008ace1. Reusing the module canonical-JSON comparator correctly distinguishes the sealed integer horizon from float/bool/string aliases. Exact regression passed; full final-analysis file 33/33 passed; Ruff and strict mypy passed; current-main merge tree and diff checks are clean; no outputs or five-claim registered evidence sources changed. READY; merge only after exact-head CI is 20/20.

deepanshu-yd:
## Review (comment)

Adding a small observation on top of the existing approval.

**Verified mechanics.** `_json_exact_equal` (forager_matched_final_analysis.py:341) canonicalizes `{"value": left}` / `{"value": right}` with `sort_keys=True`, so `float(horizon)` vs `int(horizon)` yields different bytes and fails closed, while ordering differences do not (consistent with the previous `==` semantics). The missing-key case is handled too: `plan.get("horizon")` → `None`, which cannot equal the sealed int, so a missing field also fails closed rather than skipping the check.

**Nit (non-blocking):** the malformed-plan matrix now includes the float-alias `horizon`, but `active_seeds` and `candidate_order` in the same validator already use `_json_exact_equal` and have their own cases. If any other scalar field in the plan is compared with raw `==` elsewhere in this file, the same alias hardening would apply — worth a quick grep sweep of the surrounding validator for consistency (I did not find one in this validator, but the file is large).

Solid, minimal fix. LGTM.

PR #487, opened 2026-08-16
fix(off-policy): apply each ratio once in TD traces

### LINKED ISSUE #462: OffPolicyTDLinearLearner applies rho_t twice on the carried trace (rho_t² instead of rho_t·rho_{t-1}); differs from the repo's other three per-decision-IS implementations and flips learned signs under time-varying rho
## Summary (contract question — the current recursion is documented and pinned by a test, so asking before a PR)

`OffPolicyTDLinearLearner` (`alberta_framework/core/off_policy_td.py`) says it implements per-decision importance sampling (Precup, Sutton & Singh 2000) and documents/implements

```
e_t = gamma_t * lambda_t * rho_clipped * e_{t-1} + phi_t
w  += alpha * delta_t * rho_clipped * e_t
```

Expanding, the term carried from step `t-1` is weighted by `gamma*lambda*rho_t*rho_t` and the term from `t-k` by `(gamma*lambda)^k * rho_{t-k+1}...rho_t * rho_t` — `rho_t` enters twice and `rho_{t-k}` (the ratio at the step the feature was visited) never enters. Per-decision IS (S&B 2nd ed. eq. 12.28; GQ(λ); and the convention this repo names as canonical in `off_policy_horde.py`: "Each decision's ratio enters the trace exactly once … the canonical `z_t = rho_t (gamma lambda z_{t-1} + grad)`") is

```
z_t = rho_t * (gamma*lambda*z_{t-1} + phi_t);   w += alpha * delta_t * z_t
```

`ETDLinearLearner` and `GradientTDLinearLearner` in the same file, plus `OffPolicyHordeLearner` and `StackedLinearHorde`, all use the canonical rho-outside form. The two rules coincide exactly when `rho` is constant — which is every existing test (`tests/test_off_policy_td.py` asserts λ=0 or constant-ρ cases only). They diverge precisely in the regime the class exists for: λ>0 with a real behavior/target mismatch. `tests/test_baird.py::TestRhoZeroInvariance` pins the old placement (`rho=0` "cuts the eligibility trace back to the current features", i.e. `e = phi`) — under the canonical rule `e = rho*(...) = 0`, exactly what the same file asserts for `ETDLinearLearner` ("e = rho * (...) vanishes").

## Reproduction (deterministic, pristine `main` @ `b8a37de`; 8 steps, λ=0.8, γ=0.9, α=0.1, `retrace_clip=inf`)

```
== constant rho = 2.0
  OffPolicyTDLinearLearner  w = [ 4.304975 -0.751044  6.505476]   canonical rule w = [ 4.304979 -0.751045  6.505482]   (agree to 6e-6)
== time-varying rho    rhos = [1.262 2.94 1.852 1.894 1.986 2.094 0.622 1.433]
  OffPolicyTDLinearLearner  w = [-0.696474  0.294583 -1.252004]
  canonical rho(glz+phi)    w = [ 0.502931  0.061563  0.654905]   max|diff| = 1.9069
  StackedLinearHorde (no b) w = [ 0.557202  0.139646  0.732047]   (matches the canonical rule)
```

The learned weights differ in **sign**. Any A/B that swaps `OffPolicyTDLinearLearner` against `StackedLinearHorde` / `OffPolicyHordeLearner` on off-policy λ>0 data attributes a pure trace-convention difference to the arm under test, and the off-policy fixed point is not v_π. (`ETDLinearLearner`'s follow-on trace uses `rho_t` where the paper has `rho_{t-1}`, but that one is explicitly documented as the single-step-API convention, so I am not raising it.)

## Options

- (a) adopt the canonical rule (`e = rho*(γλ e + φ)`, `w += α δ e`), update the class docstring, and change the Baird pin to `e == 0` under `rho=0` (matching its ETD sibling) — a branch with this change, a hand-computed time-varying-ρ regression, and a constant-ρ equivalence pin is ready at `Prabhat1308/asi:fix/off-policy-td-per-decision-trace` (all other off-policy/horde/step tests pass; only the two Baird pins move); or
- (b) keep the current placement as a deliberate design and document that it is *not* per-decision IS (and drop the Precup citation / the "Retrace-c=1 … convergent" claim), plus a docstring note that it differs from the other three implementations.

Happy to open the PR for (a) or the doc PR for (b) once the intended contract is confirmed.

## Dedup

`gh issue/pr list --search "per-decision" / "OffPolicyTDLinearLearner" / "eligibility trace" / "retrace"`: nothing on the trace convention. Adjacent, closed: PR #206 (0×inf bootstrap at γ=0 in this file), PR #75 (inf rewards), PR #76 (inf ρ in stacked horde), #168 (GVFSpec γ/λ validation).

lalalune:
Closes #462.

## What changed

`OffPolicyTDLinearLearner` now uses the canonical per-decision importance-sampling trace `z_t = rho_t * (gamma_t * lambda_t * z_{t-1} + phi_t)` and applies `alpha * delta_t * z_t`. The previous implementation put the current ratio in both the carried-trace decay and the weight-update multiplier, so time-varying ratios were applied twice.

The public state/config schema is unchanged. `rho=0` now correctly clears the current trace, constant-ratio behavior remains equivalent, and clipped/non-finite refusal behavior is retained.

## Verification

- failing-test-first: three varying-ratio regressions failed on the original base and pass here;
- independent 32-step varying-ratio oracle, constant-ratio equivalence, clipping/zero-ratio, nested JIT scan, overflow refusal, and pickle/config probes passed;
- independent focused/adjacent review: 158 passed;
- final rebased focused `test_off_policy_td + test_baird`: 36 passed;
- repository-wide Ruff: clean;
- strict mypy: 165 source files clean;
- patch-id unchanged by the final rebase onto `d06d51bb8d173f81cb5ae7093f83b97ba272520e`;
- no output or five-claim registered evidence source changed; the registry retains its pre-existing invalid source-drift state.

This is a code-correctness repair, not scientific evidence or a promoted claim.

### DIFF
--- a/alberta_framework/core/off_policy_td.py
+++ b/alberta_framework/core/off_policy_td.py
@@ -9,12 +9,12 @@ Theoretical background:
     to TD with FA). Several remedies exist:
 
     1. Per-decision importance sampling (Precup, Sutton, Singh 2000):
-       multiply each step's update by rho_t = pi(a_t|s_t) / b(a_t|s_t)
-       so that on average we are simulating the on-policy distribution.
-       Variance can be very large.
-    2. Retrace ratio clipping (Munos et al. 2016): use
-       rho_clipped = min(c, rho_t). Convergent for c <= 1; for c > 1 it
-       trades bias for variance reduction.
+       include each rho_t = pi(a_t|s_t) / b(a_t|s_t) once in the
+       eligibility-trace recursion so that on average we are simulating the
+       on-policy distribution. Variance can be very large.
+    2. Retrace-style ratio clipping (Munos et al. 2016): use
+       rho_clipped = min(c, rho_t) to trade variance for bias. Clipping is
+       not by itself a convergence guarantee for linear off-policy TD.
     3. Gradient-TD (TDC, GQ-lambda) (Sutton, Maei, et al. 2009-2010):
        gradient descent on the projected Bellman error.
     4. Emphatic TD (Sutton, Mahmood, White 2016): emphasis traces F_t
@@ -82,8 +82,8 @@ class OffPolicyTDState:
     Attributes:
         weights: Weight vector for linear value approximation
         bias: Bias term
-        eligibility_traces: Per-feature eligibility trace
-        bias_eligibility_trace: Bias eligibility trace
+        eligibility_traces: Per-feature importance-sampling trace ``z_t``
+        bias_eligibility_trace: Bias importance-sampling trace
         step_count: Number of updates applied
         birth_timestamp: Wall-clock seconds at init
         uptime_s: Cumulative wall-clock seconds spent in update calls
@@ -224,13 +224,13 @@ class OffPolicyTDLinearLearner:
         rho_t = pi(a_t|s_t) / b(a_t|s_t)               (provided externally)
         rho_clipped = min(c, rho_t)                     (Retrace clipping)
         delta_t = R_{t+1} + gamma_t * V(s_{t+1}) - V(s_t)
-        e_t = gamma_t * lambda_t * rho_clipped * e_{t-1} + phi_t
-        w_{t+1} = w_t + alpha * delta_t * rho_clipped * e_t
+        z_t = rho_clipped * (gamma_t * lambda_t * z_{t-1} + phi_t)
+        w_{t+1} = w_t + alpha * delta_t * z_t
 
     Setting ``retrace_clip = inf`` recovers naive per-decision IS.
-    Setting ``retrace_clip = 1.0`` gives the Retrace-c=1 update which is
-    convergent under standard conditions. Setting ``rho_t = 1`` always
-    recovers on-policy semi-gradient TD(lambda).
+    Setting ``retrace_clip = 1.0`` clips every ratio in the trace recursion
+    to one. Setting ``rho_t = 1`` always recovers on-policy semi-gradient
+    TD(lambda).
 
     Attributes:
         step_size: Learning rate alpha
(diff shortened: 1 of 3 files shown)


deepanshu-yd:
## Review (post-merge comment)

Static review of the head diff, including hand-verification of the test math. Correct fix; no blocking issues.

**The defect is real and the fix is the canonical form.** Old recurrence applied `rho_clipped` twice to the decayed portion of the trace — once in `decay = gamma*lam*rho_clipped` and again in `scaled_update = alpha*rho_clipped*td_error` — while applying it only once to the new feature. New recurrence `z_t = rho_clipped * (gamma*lambda*z_{t-1} + phi_t)` with `w += alpha*delta*z_t` is the textbook per-decision IS trace: each transition's ratio enters exactly once, on both the decayed and fresh terms.

**Verified against the tests by hand:**
- `test_time_varying_rho_enters_each_trace_once`: `z_1 = 2*phi_1 = [2,0]`; `z_2 = 0.5*(0.72*[2,0] + [0,1]) = [0.72,0.5]`; bias trace `2.0` then `1.22`. `delta_1 = 1.0`; `delta_2 = 1 + 0.9*0.2 - 0.2 = 0.98` (the `0.9*0.2` is `gamma*bias`, since `next_observation` is zeros but `v_next = b`). Yields `w = [0.27056, 0.049]`, `bias = 0.31956` — all match.
- `test_constant_rho_retains_prior_weight_updates`: two updates at `rho=2` give `w = [0.48224, 0.196]`, `bias = 0.67824` — match.

**Consistency with the rest of the tree:** the new form is exactly what `off_policy_horde.py` (~line 181) and `stacked_horde.py` (`z_d = rho_d*(gamma_d*lambda_d*z_d + x)`, line 277) already implement. This PR brings `OffPolicyTDLinearLearner` onto the same canonical recurrence — good.

**Behavioral blast radius (low):** with the default `retrace_clip=1.0` and `rho >= 1`, `rho_clipped = 1` makes old and new recurrences coincide; the change is only observable when `rho < 1` and `lambda > 0`. `cumulant_discovery.py` references the learner for the on-policy `rho=1` case (unchanged), and no production constructor instantiates it outside the module docstring — confirmed by grep. `OffPolicyTDState` schema is untouched.

**Suggestions (non-blocking):**
1. A clipping-active test (`retrace_clip=1.0`, `rho<1`) would pin the "each ratio once" semantics under clipping, where the difference is real; the current tests use `retrace_clip=inf`.
2. An explicit equivalence test against a forward-view per-decision IS lambda-return would guard against future drift of this recurrence; the hand-computed constants here are good but don't prove the form itself.

LGTM.

PR #492, opened 2026-08-16
docs(normalizers): name EMA prior warmup semantics

### LINKED ISSUE #469: EMANormalizer warmup starts at 0.5, so decay=1.0 is not the documented "pure online average": a constant stream normalizes to a non-zero value and diverges from WelfordNormalizer under the same semantics label
## Summary

`EMANormalizer` (`alberta_framework/core/normalizers.py`) documents its warmup as "the effective decay ramps up from **0** to the target decay over early steps" and `decay=1.0` as "**pure online average** (no decay)"; `to_config()` labels that setting `cumulative-float32-fail-stop-at-2p24`, i.e. the same estimator family as `WelfordNormalizer`. The implemented schedule is

```python
effective_decay = jnp.minimum(state.decay, 1.0 - 1.0 / (new_count_float + 1.0))   # normalizers.py ~L507
```

which starts at `0.5` on the first sample, not `0`, and therefore treats the initial `(mean=0, var=1)` as a permanent phantom observation. With `decay=1.0` the estimator is not a cumulative mean: `mean_n = Σx / (n+1)`, `var_n` pools a unit-variance zero sample forever (decaying only as 1/n), and a constant stream is normalized to a *non-zero* value for hundreds of steps. `WelfordNormalizer` — the other `cumulative-float32-fail-stop-at-2p24` implementation — returns the exact mean from the first sample.

## Falsifier (deterministic, pristine `main` @ `b8a37de`)

Constant observation `5.0`, `EMANormalizer(decay=1.0)` vs `WelfordNormalizer()`:

```text
n=1: EMA(decay=1) mean=2.5000 var=6.7500 normalized=0.962 | Welford mean=5.0000
n=2: EMA(decay=1) mean=3.3333 var=5.8889 normalized=0.687 | Welford mean=5.0000
n=3: EMA(decay=1) mean=3.7500 var=4.9375 normalized=0.563 | Welford mean=5.0000
n=4: EMA(decay=1) mean=4.0000 var=4.2000 normalized=0.488 | Welford mean=5.0000
```

The `1 - 1/n` schedule (`effective_decay = min(decay, 1 - 1/new_count)`) makes `decay=1.0` the exact cumulative mean / population variance (`n=1`: mean 5, var → ε; `n=2`: mean `(x1+x2)/2`, var `(x2-x1)^2/4`) and makes the "ramps up from 0" sentence true; for `decay<1` it only changes the first ~`1/(1-decay)` steps.

## Why I am filing this as an issue rather than a PR

The current schedule is *internally coherent* if it is read as "one prior pseudo-sample at (0, 1)" — the pooled variance in the trace above is exactly that — so this is a contract question, and the choice has ripple:

- `benchmarks/ipmnist_screening.py::ema_normalize` is an "equation-parity restatement" carrying the same `1/(n+1)` term, with parity tests in `tests/test_ipmnist_screening.py`; the `upgd_ema_norm*` / `adamw_cbp_ema_norm*` arms have shards under `outputs/ipmnist_screening/shards/`. Changing the core schedule without the restatement breaks parity; changing both changes those arms' recorded numbers (first ~1k steps of a run only, but bit-exact reproducibility of the existing shards is lost).
- `steps/step1.py` and `steps/step3.py` construct `EMANormalizer` for the Step 1/3 kernels.
- `normalizers.py` is listed in `evaluation/upgd_ipmnist_nonpromoting.py::RECONSTRUCTED_NUMERIC_SOURCE_SHA256` (already mismatched on `main`: pinned `6e58ae70…`, current `05288f23…`), so no *new* invalidation, but worth stating.

Options:

1. **Code**: switch to `1 - 1/new_count` in `EMANormalizer` and in the screening restatement, keeping the parity test green; note the shard-reproducibility break for the EMA-norm arms in the runbook.
2. **Contract**: keep the schedule and change the docstring/`to_config` semantics label to state the unit-variance zero-mean prior pseudo-sample explicitly (and drop "pure online average"), so that `cumulative-float32-fail-stop-at-2p24` no longer implies Welford-equivalent first moments.

Happy to send the PR for whichever is chosen (failing-test-first; option 1 would include the parity restatement and its test).

## Dedup

`gh issue/pr list --search "EMANormalizer warmup" / "EMA decay ramp" / "EMANormalizer"` (all states): only #226 (0·inf when decay disabled), unrelated.

lalalune:
Closes #469

Chooses the issue-compatible contract option: preserve the numerical recurrence and name it accurately as a zero-mean, unit-variance prior pseudo-sample rather than a pure online average.

What changes:
- documents effective_decay = min(decay, 1 - 1 / (count + 1)), including the first-step min(decay, 1/2) behavior;
- gives this estimator a distinct stable semantics label while accepting and canonicalizing the legacy labels;
- keeps the IPMNIST equation-parity restatement and nested config round-trips aligned;
- appends the compatibility decision to the nonpromoting screening runbook.

Evidence:
- failing-first malformed non-string semantics regressions, then green;
- 91 normalizer/config/import tests and 307 IPMNIST tests passed in independent review;
- exact parent/head trajectory digests match across decay values 0, .1, .49, .5, .9, .999, and 1;
- post-rebase 66 normalizer/import tests, repository Ruff, and strict mypy (165 files) pass;
- rebase preserved the reviewed patch-id exactly.

Provenance:
- RUNBOOK change is append-only;
- no outputs are rewritten;
- no changed path intersects the five registered evidence-source inventories;
- the evidence registry remains inherited-invalid from pre-existing registered-source drift, and the relevant old development manifests were already stale.


deepanshu-yd:
## Review (post-merge comment)

Static review. Sound change; no blocking issues.

**Verified math.** The claimed prior-pseudo-sample semantics hold: with `decay=1.0`, `effective_decay = min(1.0, 1 - 1/(count+1))` and `new_mean = mean + (1-effective)*delta` reproduce a cumulative weighted mean/var over the prior + observations. Hand-check `[5, 7]` gives mean `2.5` then `4.0` and var `6.75` then `9.0` — exactly the values pinned in `test_decay_one_preserves_initial_moment_pseudo_sample` and the core `test_warmup_uses_initial_moment_pseudo_sample`. The docstrings and RUNBOOK now describe the actual recurrence.

**Loader compatibility.** `normalizer_from_config` accepting legacy `bounded-recency` / `cumulative-float32-fail-stop-at-2p24` labels only when they match the decay family (and canonicalizing on re-serialization) is fail-closed: `test_generic_semantics_alias_must_match_decay_family` pins the cross-family rejection. The `type(estimator_semantics) is not str` guard is a safe tightening — previously a non-str value would simply mismatch `expected_semantics` and raise anyway, so no persisted-config regression.

**Observations (non-blocking):**

1. Title is `docs(...)` but this is behavior-affecting: `to_config()` now emits new semantics strings, which changes config receipts. The RUNBOOK note covers it, but the commit message/PR title undersells the code impact.
2. `test_config_names_prior_regularized_semantics` hardcodes the full new label strings. Referencing the exported `EMA_PRIOR_*_SEMANTICS` constants would keep the test aligned if the labels change again.
3. For full parity it would be nice to see the same prior-sample semantics documented on the scan restatement in `ipmnist_screening.py` — the docstring there is updated, good; just confirming nothing else compares the old literal labels outside `normalizers.py` (grep shows none).

LGTM.

Issue #490, opened 2026-08-16
Step3HordeConfig accepts non-boolean use_obgd/use_layer_norm, silently changing measured learner behavior via truthiness


Svector-anu:
`Step3HordeConfig` validated its numerical and dimensional fields but accepted arbitrary objects for `use_obgd` and `use_layer_norm`. Python truthiness selects the measured learner behavior at these two flags — a deserialized JSON string `"false"` is truthy in Python, so it was stored unchanged and evaluated as true, silently enabling ObGD bounding (and likewise layer normalization) instead of rejecting the malformed configuration.

confirmed directly through the learner's supported `to_config()` surface: a config deserialized with `use_obgd: "false"` produced a horde config with `'bounder': {'type': 'ObGDBounding', ...}` — the string selected `ObGDBounding` instead of being rejected.

reachability: `alberta_framework.__init__` exports `AlbertaPipelineConfig`, `make_alberta_pipeline`, and `run_pipeline_smoke`. `AlbertaPipelineConfig.from_dict()` passes `payload["horde"]` to `Step3HordeConfig.from_dict()`; pipeline construction then calls `make_step3_horde(self._config.horde)`, which uses `cfg.use_obgd` directly to select `ObGDBounding(kappa=cfg.obgd_kappa) if cfg.use_obgd else None` — a malformed truthy value previously changed measured behavior through this real, publicly-reachable configuration path.

fix incoming: a new `_require_bool` validator (`type(value) is bool`), applied to both algorithm-selection flags during `Step3HordeConfig`'s post-init validation, matching the existing `_require_positive_int` pattern already used for numerical fields in the same function.

PR #491, opened 2026-08-16
fix(step3): reject non-boolean learner flags

Svector-anu:
Closes #490.

## What changed

`Step3HordeConfig` validated its numerical and dimensional fields but accepted arbitrary objects for `use_obgd` and `use_layer_norm`. Python truthiness selects the measured learner behavior at these two flags — a deserialized JSON string `"false"` is truthy in Python, so it was stored unchanged and evaluated as true, silently enabling ObGD bounding (and likewise layer normalization) instead of rejecting the malformed configuration.

confirmed directly before touching the source, through the learner's supported `to_config()` surface: a config deserialized with `use_obgd: "false"` produced a horde config with `'bounder': {'type': 'ObGDBounding', ...}` — the string selected `ObGDBounding` instead of being rejected.

fix: a new `_require_bool` validator (`type(value) is bool`, mirroring the existing `_require_positive_int` pattern already used for numerical fields in the same function), applied to both algorithm-selection flags during `Step3HordeConfig`'s post-init validation.

## Reachability

`alberta_framework.__init__` exports `AlbertaPipelineConfig`, `make_alberta_pipeline`, and `run_pipeline_smoke` — confirmed directly. `AlbertaPipelineConfig.from_dict()` passes `payload["horde"]` to `Step3HordeConfig.from_dict()`; pipeline construction then calls `make_step3_horde(self._config.horde)`, which uses `cfg.use_obgd` directly to select `ObGDBounding(kappa=cfg.obgd_kappa) if cfg.use_obgd else None` — confirmed directly in source. A malformed truthy value previously changed measured behavior through this real, publicly-reachable configuration path, not a hardcoded-safe internal call.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_step3_production.py -q -o addopts=""
42 passed in 2.69s

$ .venv/bin/python -m ruff check alberta_framework/steps/step3.py tests/test_step3_production.py
All checks passed!

$ .venv/bin/python -m mypy
Success: no issues found in 165 source files
```

New test: `test_step3_config_rejects_non_boolean_algorithm_flags`, parametrized over `use_obgd` and `use_layer_norm`, deserializing a config with the field set to the string `"false"` and asserting `ValueError` matching `"{field} must be a boolean"`.

[...]

PR #494, opened 2026-08-16
fix(step4): reject non-boolean layer norm flags

### LINKED ISSUE #493: Step4SARSAConfig.use_layer_norm accepts non-boolean values, silently changing measured learner behavior via truthiness
Same bug class as #490 (`Step3HordeConfig`), in the sibling Step 4 SARSA facade: `Step4SARSAConfig.use_layer_norm` is annotated `bool`, but `_validate_sarsa_config` never validated it. A JSON-like caller passing the string `"false"` had it accepted and forwarded unchanged to `SARSAAgent`/`HordeLearner` — Python treats that non-empty string as truthy, so an input that appears to disable layer normalization instead enables the layer-normalized learner path, silently changing measured algorithm behavior instead of being rejected.

confirmed directly:
```text
input=False stored=False truthy=False agent_flag=False
input='false' stored='false' truthy=True agent_flag='false'
```

reachability: `Step4SARSAConfig` is exported from `alberta_framework/steps/__init__.py`. `Step4SARSAConfig.from_dict` accepts deserialized configuration mappings; `AlbertaPipelineConfig.from_dict` calls it for the external `control` payload in `alberta_framework/pipeline.py`; pipeline construction passes the config to `make_step4_sarsa_agent`, which forwards `cfg.use_layer_norm` to `SARSAAgent` and thereby selects the learner architecture.

fix incoming: require `type(config.use_layer_norm) is bool` in `_validate_sarsa_config`, rejecting integer aliases (`0`/`1`), strings, and `None` as well.

Svector-anu:
Closes #493.

## What changed

Same bug class as #490 (`Step3HordeConfig`), in the sibling Step 4 SARSA facade: `Step4SARSAConfig.use_layer_norm` is annotated `bool`, but `_validate_sarsa_config` never validated it. A JSON-like caller passing the string `"false"` had it accepted and forwarded unchanged to `SARSAAgent`/`HordeLearner` — Python treats that non-empty string as truthy, so an input that appears to disable layer normalization instead enables the layer-normalized learner path, silently changing measured algorithm behavior instead of being rejected.

```text
input=False stored=False truthy=False agent_flag=False
input='false' stored='false' truthy=True agent_flag='false'
```

fix: require `type(config.use_layer_norm) is bool` in `_validate_sarsa_config`, rejecting integer aliases (`0`/`1`), strings, and `None` as well — not just the specific string that triggered the reproduction.

## Reachability

`Step4SARSAConfig` is exported from `alberta_framework/steps/__init__.py` — confirmed directly. `Step4SARSAConfig.from_dict` accepts deserialized configuration mappings; `AlbertaPipelineConfig.from_dict` calls it for the external `control` payload in `alberta_framework/pipeline.py`; pipeline construction passes the config to `make_step4_sarsa_agent`, which forwards `cfg.use_layer_norm` to `SARSAAgent` and thereby selects the learner architecture — all confirmed directly in source.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_step4_production.py -q -o addopts=""
61 passed in 3.60s

$ .venv/bin/python -m ruff check alberta_framework/steps/step4.py tests/test_step4_production.py
All checks passed!

$ .venv/bin/python -m mypy
```

New test: `test_step4_sarsa_config_requires_exact_boolean_layer_norm`, parametrized over `0`, `1`, `"false"`, `None`, asserting `ValueError` matching `"use_layer_norm must be a boolean"`.

mutation-resistance verified independently: reverted just the source fix (`git stash push -- alberta_framework/steps/step4.py`, kept the test), reran — all 4 cases failed with `DID NOT RAISE ValueError`, confirming every alias previously crossed validation and would have silently selected the layer-normalized learner path. restored the fix, 61/61 green again.

## Evidence

93bb0bf1156f3888c846d95a6fcddf6d63b58854

```text
$ .venv/bin/python -m pytest tests/test_step4_production.py -q -o addopts="" -k test_step4_sarsa_config_requires_exact_boolean_layer_norm
FAILED [0] - Failed: DID NOT RAISE ValueError
FAILED [1] - Failed: DID NOT RAISE ValueError
FAILED [false] - Failed: DID NOT RAISE ValueError
FAILED [None] - Failed: DID NOT RAISE ValueError
4 failed, 57 deselected in 0.48s

green (fix restored):
$ .venv/bin/python -m pytest tests/test_step4_production.py -q -o addopts=""
61 passed in 3.55s
```

```text
input=False stored=False truthy=False agent_flag=False
input='false' stored='false' truthy=True agent_flag='false'
```
independently reran the full touched test file, ruff, and mypy myself; independently confirmed `cfg.use_layer_norm` flows through `make_step4_sarsa_agent` → `SARSAAgent` and the full pipeline-config reachability chain before writing the reachability claim; did my own revert-and-confirm-red mutation check.

## Provenance

AI provider/model: openai / gpt-5.6-sol (fix, tests, via AgentRouter proxy); anthropic / claude-sonnet-5 (independent re-verification: reran the full touched test file, ruff, and mypy myself, independently confirmed the reachability chain, did my own revert-and-confirm-red mutation check, opened the governing issue, pushed, opened this PR)

### DIFF
--- a/alberta_framework/steps/step4.py
+++ b/alberta_framework/steps/step4.py
@@ -223,6 +223,10 @@ def _validate_sarsa_config(config: Step4SARSAConfig) -> None:
     meta_step_size = _require_nonnegative_real("meta_step_size", config.meta_step_size)
     bounder_kappa = _require_positive_real("bounder_kappa", config.bounder_kappa)
     sparsity = _require_unit_interval("sparsity", config.sparsity)
+    if type(config.use_layer_norm) is not bool:
+        raise ValueError(
+            f"use_layer_norm must be a boolean, got {config.use_layer_norm!r}"
+        )
     object.__setattr__(config, "n_actions", n_actions)
     object.__setattr__(config, "hidden_sizes", hidden_sizes)
     object.__setattr__(config, "gamma", gamma)
--- a/tests/test_step4_production.py
+++ b/tests/test_step4_production.py
@@ -205,6 +205,12 @@ def test_step4_sarsa_scalars_reject_invalid_inputs(field: str, value: object) ->
         _config_with(**{field: value})
 
 
+@pytest.mark.parametrize("value", [0, 1, "false", None])
+def test_step4_sarsa_config_requires_exact_boolean_layer_norm(value: object) -> None:
+    with pytest.raises(ValueError, match="use_layer_norm must be a boolean"):
+        _config_with(use_layer_norm=value)
+
+
 def test_step4_sarsa_scalars_preserve_legal_boundaries() -> None:
     config = Step4SARSAConfig(
         n_actions=1,


lalalune (approved):
READY at exact 93bb0bf1156f3888c846d95a6fcddf6d63b58854. The actual-type boolean gate prevents numeric aliases, strings, NumPy booleans, containers, and __class__/__bool__ spoof objects from silently selecting the layer-normalized SARSA architecture. Independent matrix: 33 malformed direct/from_dict/nested-pipeline paths refused; valid booleans round-trip through JSON and reach the constructed agent exactly. Exact head: 180 Step4/pipeline/shared float32 validation tests passed under warnings-as-errors; Ruff and strict mypy (165 files) pass; current-main merge-tree/diff-check are clean with no intervening target-path changes. No outputs or five-claim registered sources change; registry invalidity is inherited. Merge only after exact-head hosted 20/20 CI.

deepanshu-yd (commented):
## Review (comment)

Small consistency note on top of the existing approval.

`_require_bool` in step3.py and the inline `if type(config.use_layer_norm) is not bool` in step4.py implement the same actual-type gate. Both are correct (they reject `np.bool_`, numeric aliases, strings, and `__bool__`/`__class__` spoofs — `np.bool_` is not a subclass of `bool`, and `type(...) is not bool` is exact). Step4 has no other boolean flags (only `use_layer_norm`), so the gate covers the full flag surface there.

**Suggestion (non-blocking):** since #491 landed `_require_bool` first, step4 could reuse the same helper (or the helper could move to a shared steps-level util) so the two gates stay byte-identical as the codebase evolves. Right now they can drift independently (e.g., one adds `isinstance` handling, the other doesn't). Purely a maintenance preference.

LGTM.

PR #496, opened 2026-08-16
fix(forager-matched): reject numeric aliases in completion summary

### LINKED ISSUE #495: forager_matched_sealed_evaluation_campaign's summary validator accepts numeric aliases in the sealed completion schema
Seventh instance of the same bug class as #450/#460/#466/#471/#483/#488: `_summary_validator` in `alberta_framework/benchmarks/forager_matched_sealed_evaluation_campaign.py` compared the supplied completion summary with the rebuilt exact v2 summary using ordinary Python equality. A numeric alias such as `6.0` compares equal to the expected integer `6`, so a non-canonical JSON value could cross the sealed completion schema boundary.

confirmed directly, mutating `candidate_count` from integer `6` to float `6.0`: the mutated summary was silently accepted instead of raising.

reachability: `_summary_validator` is used by `prepare_sealed_evaluation_campaign`, `run_sealed_evaluation_campaign`, and `verify_sealed_evaluation_campaign_content` — all exported in this module's `__all__` and reached from the module CLI operations. They validate externally supplied persisted completion summaries against rebuilt sealed campaign content.

fix incoming: compare canonical JSON bytes (via the already-shared `forager_matched_campaign.canonical_json_bytes` helper, imported as `campaign`) for the normalized supplied summary and expected summary, rather than Python equality.

Svector-anu:
Closes #495.

## What changed

Seventh instance of the same bug class as #450/#460/#466/#471/#483/#488: `_summary_validator` compared the supplied completion summary with the rebuilt exact v2 summary using ordinary Python equality. A numeric alias such as `6.0` compares equal to the expected integer `6`, so a non-canonical JSON value could cross the sealed completion schema boundary.

confirmed directly before touching the source, mutating `candidate_count` from integer `6` to float `6.0`: the mutated summary was silently accepted instead of raising.

fix: compare canonical JSON bytes (via the already-shared `forager_matched_campaign.canonical_json_bytes` helper, imported as `campaign` — the same helper the sibling PRs #489/#486 use) for the normalized supplied summary and expected summary, rather than Python equality.

## Reachability

`_summary_validator` is used by `prepare_sealed_evaluation_campaign`, `run_sealed_evaluation_campaign`, and `verify_sealed_evaluation_campaign_content` — all confirmed exported in this module's `__all__` and reached from the module CLI operations. They validate externally supplied persisted completion summaries against rebuilt sealed campaign content.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_forager_matched_sealed_evaluation_campaign_adversarial.py -q -o addopts="" -k test_completion_summary_rejects_false_authority_and_closure
1 passed, 21 deselected in 1.43s

$ .venv/bin/python -m ruff check alberta_framework/benchmarks/forager_matched_sealed_evaluation_campaign.py tests/test_forager_matched_sealed_evaluation_campaign_adversarial.py
All checks passed!

$ .venv/bin/python -m mypy
```

New coverage: a `candidate_count` numeric-alias assertion added to the existing `test_completion_summary_rejects_false_authority_and_closure`, asserting `ValueError` matching `"exact v2 schema"`.

mutation-resistance verified independently: reverted just the source fix (`git stash push -- alberta_framework/benchmarks/forager_matched_sealed_evaluation_campaign.py`, kept the test), reran — failed with `DID NOT RAISE ValueError`, confirming the float alias previously crossed validation undetected. restored the fix, green again.

**Note on the full-file test run:** this test file has 5 pre-existing failures and 4 pre-existing errors unrelated to this change — the same macOS `O_TMPFILE`/`renameat2`-unavailable publication-primitive limitation already documented on sibling PRs #489/#486. Independently confirmed the exact same 5 failures + 4 errors (same test names) reproduce identically on an unmodified checkout before writing this PR.

## Evidence

e1a63395ba88b93893a7d58ccc14037515d3d46f

```text
$ .venv/bin/python -m pytest tests/test_forager_matched_sealed_evaluation_campaign_adversarial.py -q -o addopts="" -k test_completion_summary_rejects_false_authority_and_closure
FAILED - Failed: DID NOT RAISE ValueError
1 failed, 21 deselected in 1.12s

green (fix restored):
$ .venv/bin/python -m pytest tests/test_forager_matched_sealed_evaluation_campaign_adversarial.py -q -o addopts="" -k test_completion_summary_rejects_false_authority_and_closure
1 passed, 21 deselected in 0.94s
```

```text
candidate_count: int 6 -> float 6.0
before fix: accepted (no error)
after fix:  ForagerMatchedSealedEvaluationCampaignError: sealed completion summary differs from its exact v2 schema
```
independently reran the full touched-file test target, ruff, and mypy myself; independently confirmed the 5 failures + 4 errors in the full-file run are pre-existing by reproducing them (same test names) on an unmodified checkout; independently confirmed the reachability chain (`__all__` exports, three real call sites) before writing the reachability claim; did my own revert-and-confirm-red mutation check.

## Provenance

AI provider/model: openai / gpt-5.6-sol (fix, tests, via AgentRouter proxy); anthropic / claude-sonnet-5 (independent re-verification: reran the touched-file test target, ruff, and mypy myself, independently confirmed the pre-existing unrelated failures, verified the reachability chain, did my own revert-and-confirm-red mutation check, opened the governing issue, pushed, opened this PR)

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_sealed_evaluation_campaign.py
+++ b/alberta_framework/benchmarks/forager_matched_sealed_evaluation_campaign.py
@@ -723,7 +723,9 @@ def _summary_validator(inputs: _SealedInputs) -> _SummaryValidator:
             score_evidence,
             request,
         )
-        if campaign._plain(summary) != expected:
+        if campaign.canonical_json_bytes(campaign._plain(summary)) != campaign.canonical_json_bytes(
+            expected
+        ):
             raise ForagerMatchedSealedEvaluationCampaignError(
                 "sealed completion summary differs from its exact v2 schema"
             )
--- a/tests/test_forager_matched_sealed_evaluation_campaign_adversarial.py
+++ b/tests/test_forager_matched_sealed_evaluation_campaign_adversarial.py
@@ -980,6 +980,11 @@ def test_completion_summary_rejects_false_authority_and_closure(tmp_path: Path) 
         with pytest.raises(ValueError, match="closure|exact v2 schema"):
             validator(context, receipt, scores, request, changed)
 
+    numeric_alias = dict(valid)
+    numeric_alias["candidate_count"] = float(numeric_alias["candidate_count"])
+    with pytest.raises(ValueError, match="exact v2 schema"):
+        validator(context, receipt, scores, request, numeric_alias)
+
 
 def test_partial_final_artifacts_fail_closed_without_repair(
     tmp_path: Path,


lalalune (approved):
Independent review at exact `e1a63395ba88b93893a7d58ccc14037515d3d46f`: READY. Parent acceptance of `candidate_count: 6.0` was independently reproduced. The canonical-byte comparison rejects all 10 equal numeric aliases in the flat completion schema, every string mutation, and missing/extra keys while retaining the exact valid summary. Full adversarial module: 22 passed; completion/verification selector: 6 passed; Ruff and strict mypy green. The patch composes cleanly with current main and no changed path is a five-claim registered evidence source or output artifact.

deepanshu-yd (commented):
## Review (comment)

Static review of the head diff. The fix is correct and I could not find a blocking issue.

**Correctness.** `campaign.canonical_json_bytes` (forager_matched_campaign.py:202) round-trips through `_plain` and `json.dumps(..., allow_nan=False, sort_keys=True)`, so the comparison distinguishes int vs float (`5` vs `5.0`) and bool vs int (`true` vs `1`) exactly as the test expects, while `sort_keys` keeps the comparison order-insensitive — the same relational semantics as the old `dict != dict` for key ordering. No regression on valid summaries.

**Nits (non-blocking):**

1. Redundant canonicalization: `canonical_json_bytes` already calls `_plain` internally (campaign.py:206), so `campaign.canonical_json_bytes(campaign._plain(summary))` double-applies it. `campaign.canonical_json_bytes(summary)` alone is equivalent and slightly clearer.
2. The test pins one numeric alias (`candidate_count`). PR #489's sibling landed a multi-field matrix; parameterizing `candidate_count`, `seed_count`, `completed_cell_count`, and the boolean fields would give equal parity coverage here. Optional.
3. Boundary note, pre-existing and acceptable: `_plain` also normalizes tuple↔list and non-`str` mapping keys to `str`, so those aliases are silently equalized rather than rejected. The validator's stated scope is numeric/boolean value aliases, so this is fine — just worth keeping in mind for a future hardening pass.

Nice parity with #489 (same canonical-comparator pattern in the sealed module). LGTM.

PR #504, opened 2026-08-16
fix(step11): reject class-spoofed non-integer steps in _require_int

### LINKED ISSUE #503: step11.py _require_int accepts __class__-spoofed non-integers
`_require_int` in `alberta_framework/steps/step11.py` guards against `bool` and non-`Integral` values with:

```python
```

confirmed directly against `run_step11_smoke` (publicly reachable via `alberta_framework.steps.__all__`), against the current unpatched code:

```text
SPOOFED steps ACCEPTED, result.steps = 3 <class 'int'>
```

a spoofed object is silently converted via its own `__int__`/`__index__` and accepted as if it were a real `int`, rather than rejected per the facade's documented integer contract. same underlying pattern already fixed in `step12.py` (#400) and `step9.py` (#502).

Svector-anu:
Closes #503.

## What changed

Same pattern as #400 (step12) and #502 (step9): `_require_int` in `step11.py` used `isinstance(value, bool) or not isinstance(value, Integral)`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `int` passes `isinstance(value, Integral)` even though `type(value) is not int`.

`run_step11_smoke` and the OaK config validator route `steps`/`observation_dim`/`n_primitive_actions`/etc. through this same helper - both publicly reachable via `alberta_framework.steps.__all__`.

## Reachability

`run_step11_smoke` is exported from `alberta_framework.steps.__init__`'s `__all__` - publicly reachable with ordinary concrete inputs.

## Verification

```text
SPOOFED steps ACCEPTED, result.steps = 3 <class 'int'>
```

- new test `test_run_step11_smoke_rejects_class_spoofed_integer_steps` added next to the existing `test_run_step11_smoke_rejects_non_integer_steps`.
- full file: `173 passed`, 2 pre-existing failures (`test_step11_oak_validates_original_real_domain_before_narrowing`, `test_step11_oak_narrows_the_original_real_once`) - independently reproduced on unmodified main, unrelated `_require_real`/longdouble narrowing code path, not touched by this change.

## Provenance

AI-assisted (Claude, `claude-sonnet-5`). Spoofing behavior reproduced empirically by me in a real interpreter session against the unpatched code before writing the fix; the two pre-existing failures were independently reproduced against unmodified main before being excluded. All verification above performed and confirmed by me before pushing.

### DIFF
--- a/alberta_framework/steps/step11.py
+++ b/alberta_framework/steps/step11.py
@@ -30,7 +30,7 @@ from __future__ import annotations
 
 from dataclasses import asdict, dataclass
 from numbers import Integral, Real
-from typing import Any
+from typing import Any, cast
 
 import jax.numpy as jnp
 import jax.random as jr
@@ -219,9 +219,10 @@ def _require_int(
     minimum: int | None = None,
     exclusive_maximum: int | None = None,
 ) -> int:
-    if isinstance(value, bool) or not isinstance(value, Integral):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Integral):
         raise ValueError(f"{name} must be an integer, got {value!r}")
-    number = int(value)
+    number = int(cast(Integral, value))
     if minimum is not None and number < minimum:
         if minimum == 1:
             raise ValueError(f"{name} must be positive, got {value!r}")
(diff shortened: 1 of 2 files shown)


lalalune (approved):
Maintainer re-review: exact-head diff inspected and focused regression suite passed locally. The claimed boundary defect is real and the implementation is appropriately scoped. Approved for squash merge.

PR #529, opened 2026-08-16
fix(streams): reject non-finite feature_scales in ScaledStreamWrapper

### LINKED ISSUE #507: ScaledStreamWrapper accepts non-finite feature_scales, silently emitting NaN observations from an accepted config
## Problem

`ScaledStreamWrapper` in `alberta_framework/streams/synthetic.py` validates only the `feature_scales` shape (`ndim == 1` and length equals the inner stream's `feature_dim`), not the values. Non-finite scales (`NaN`/`+/-inf`) are accepted and then multiplied into every emitted observation, silently corrupting the stream with `NaN`/`inf` features from an accepted configuration. The class exists to feed scale-shift experiments, so an accepted non-finite scale silently measures a corrupted stream.

## Reproduction

```python
import jax.numpy as jnp
from alberta_framework.streams.synthetic import RandomWalkStream, ScaledStreamWrapper

# Accepts NaN scale factors (only shape is checked)
stream = ScaledStreamWrapper(RandomWalkStream(4), jnp.array([1.0, float("nan"), 2.0, 3.0]))
```

## Proposed Fix

- Reject non-finite `feature_scales` (and require finite float32 values) in `ScaledStreamWrapper.__init__` alongside the existing shape check.
- Add validation tests in `tests/test_streams.py`.

deepanshu-yd:
Fixes #507.

Adds a finiteness check on `feature_scales` in `ScaledStreamWrapper.__init__`, alongside the existing shape check, plus a regression test. Contribution provenance: AI-assisted implementation (provider: Anthropic, model: claude-sonnet-4).

### DIFF
--- a/alberta_framework/streams/synthetic.py
+++ b/alberta_framework/streams/synthetic.py
@@ -798,6 +798,8 @@ class ScaledStreamWrapper:
                 f"feature_scales shape ({self._feature_scales.shape}) "
                 f"must match inner stream's feature_dim ({inner_stream.feature_dim})"
             )
+        if not bool(jnp.all(jnp.isfinite(self._feature_scales))):
+            raise ValueError("feature_scales must contain only finite float32 values")
 
     @property
     def feature_dim(self) -> int:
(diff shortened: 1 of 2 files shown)


lalalune (approved):

PR #533, opened 2026-08-16
Do not form 0*inf TD bootstraps at gamma 0

Kayvan-Zahiri:
## Summary
- `TDLinearLearner` and `TrueOnlineTDLearner` now skip `gamma * V(s')` (and True Online's `gamma * lambda` trace decay) before the product, so gamma 0 cannot turn an unused inf next observation into NaN.
- The optimizer layer already allowed this; the learner wrappers still required a finite next observation and formed the 0*inf bootstrap.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_td_learners.py tests/test_true_online_td.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/learners.py tests/test_td_learners.py tests/test_true_online_td.py`

### DIFF
--- a/alberta_framework/core/learners.py
+++ b/alberta_framework/core/learners.py
@@ -75,6 +75,12 @@ from alberta_framework.core.update_safety import (
 )
 from alberta_framework.streams.base import ScanStream
 
+
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip a collapsed 0*inf product before it becomes NaN."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 # Type variable for TD stream state
 StateT = TypeVar("StateT")
 
@@ -1795,9 +1801,10 @@ class TDLinearLearner:
         next_prediction = self.predict(state, next_observation)
 
         gamma_scalar = jnp.squeeze(gamma)
+        next_value = jnp.squeeze(next_prediction)
         td_error = (
             jnp.squeeze(reward)
-            + gamma_scalar * jnp.squeeze(next_prediction)
+            + _skip_zero_scale(gamma_scalar, next_value)
             - jnp.squeeze(prediction)
         )
 
@@ -1833,7 +1840,7 @@ class TDLinearLearner:
         inputs_valid = (
             jnp.all(jnp.isfinite(observation))
             & jnp.all(jnp.isfinite(reward))
-            & jnp.all(jnp.isfinite(next_observation))
+            & (jnp.all(jnp.isfinite(next_observation)) | (gamma_scalar == 0.0))
             & jnp.all(jnp.isfinite(gamma))
         )
         update_applied = (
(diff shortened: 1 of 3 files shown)


kenjohnscreates (approved):

Verified by re-applying the diff onto current `main` (`06b0fbd`) in a scratch worktree; all three files **byte-identical to PR head `f2c988f8`**.

**The `jnp.where` idiom is correct at these sites** — the distinction that gated #219 cuts the other way here: `gamma` arrives as a runtime argument (genuinely traced), so a trace-time hoist is impossible and the runtime select is the right tool. In the True Online composite `(gamma_scalar == 0.0) | (lamda == 0.0)`, `lamda` is a Python constructor float, so that half constant-folds at trace time and the emitted predicate is effectively gamma-only — fine either way.

**Layering is consistent with the merged optimizer work.** The optimizer layer already tolerates inf next-features at γ=0 (#213); these learner wrappers were the remaining callers that both required finite `next_observation` and formed the `0*inf` bootstrap themselves. The relaxation `isfinite(next_observation) | (gamma == 0.0)` mirrors the pattern verified in #218/#213, and the new tests prove the stronger property — not merely that the update commits, but that the committed weights and `td_error` are **equal** to the finite-next-observation case, so the inf genuinely does not leak.

**Consumer/exposure triage (wave rule):** `TDLinearLearner`/`TrueOnlineTDLearner` are consumed by `steps/step1.py` and `streams/gymnasium.py` and re-exported from the package `__init__`; nothing under `benchmarks/` imports them, `core/learners.py` is not in the five-claim registered-source inventory, and `outputs/step2_canonical/` is an OPMNIST artifact with no dependence on these classes. No pinned or measured path is exposed.

**Battery (arm64 M1):** `test_td_learners.py` + `test_true_online_td.py`: **44/44**; `ruff check .` clean; repo-config strict mypy clean at **165 source files**. Failing-test-first reproduced: tests-only against baseline → both new `test_zero_gamma_does_not_multiply_inf_next_observation` cases fail deterministically; full diff → green.

**Two non-gating observations.** (1) The True Online `trace_scale` rewrite changes float association on the *nonzero* path: previously `alpha * gamma * lamda * trace_dot` evaluated as `((alpha·gamma)·lamda)·dot`, now `alpha * (decay_scale · dot)` with `decay_scale = gamma·lamda` — mathematically identical, but ULP-level drift is possible for nonzero-γλ users. With no pinned artifact or benchmark consumer downstream this is documentation-grade only (the same class of note I left on #217). (2) The reported `next_prediction` metric is now sanitized to `0.0` when `V(s')` is non-finite at γ=0 — a sensible metric-surface choice (the raw value would be a meaningless `inf`), but worth a line in the PR description since it slightly exceeds the "skip the bootstrap product" summary.
Compute receipt: 3262001 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

lalalune (approved):
Reviewed against current main. The gamma-zero branches avoid forming unused 0*inf bootstraps and traces while preserving the nonzero-gamma path. Focused TDLinearLearner and TrueOnlineTD tests pass (41 passed), Ruff passes, and diff-check is clean.

lalalune (approved):

lalalune (approved):

PR #534, opened 2026-08-16
Do not form 0*inf option-search bootstraps at discount 0

Kayvan-Zahiri:
## Summary
- Option-search Bellman targets skipped `discount_ema * max Q(s')` only after forming the product, so discount 0 times an overflowed next-state value became NaN and dropped a completed candidate.
- Zero-discount models now skip that bootstrap and still rank on the remaining finite return terms.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_option_search_control.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/option_search_control.py tests/test_option_search_control.py`

### DIFF
--- a/alberta_framework/core/option_search_control.py
+++ b/alberta_framework/core/option_search_control.py
@@ -619,15 +619,21 @@ class OptionSearchControl:
                 learner_state,
                 predicted_next,
             )
+            discount = models.discount_ema[index]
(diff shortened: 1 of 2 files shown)


lalalune (approved):

lalalune (approved):

PR #536, opened 2026-08-16
Skip behavior-model diagnostic EMAs when decay is disabled

Kayvan-Zahiri:
## Summary
- `diagnostic_decay` is allowed to be 0, but the nll/accuracy/confidence EMAs still formed `decay * previous`. An inf leftover tracker became NaN and blocked a finite supervised sample.
- Zero decay now skips those products and no longer requires the unused diagnostic EMAs to be finite.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_behavior_model.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/behavior_model.py tests/test_behavior_model.py`

### DIFF
--- a/alberta_framework/core/behavior_model.py
+++ b/alberta_framework/core/behavior_model.py
@@ -537,20 +537,27 @@ class BehaviorModel:
 
         decay = jnp.asarray(cfg.diagnostic_decay, dtype=jnp.float32)
         first = state.step_count == 0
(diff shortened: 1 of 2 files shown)


lalalune (approved):

lalalune (approved):

lalalune (approved):

PR #539, opened 2026-08-16
fix(streams): reject non-finite SwitchingTwoStateMDP payoffs

### LINKED ISSUE #512: SwitchingTwoStateMDP accepts non-finite payoffs_a/payoffs_b, emitting NaN rewards from an accepted config
## Problem

`SwitchingTwoStateMDP` in `alberta_framework/streams/closed_loop.py` validates `phase_length` but checks `payoffs_a`/`payoffs_b` only for shape (`2x2`), not finiteness. Non-finite payoffs are accepted and passed through `np.asarray(..., dtype=np.float32)`, so every reward the MDP emits is `NaN`/`inf` while the stream still reports a valid, finite-looking configuration. `phase_length` receives full integer validation, so the payoff gap is inconsistent with the constructor's own intent.

## Reproduction

```python
from alberta_framework.streams.closed_loop import SwitchingTwoStateMDP, SwitchingTwoStateConfig

# Accepted: NaN payoffs poison every reward
mdp = SwitchingTwoStateMDP(
    SwitchingTwoStateConfig(payoffs_a=[[float("nan"), 0.0], [0.0, 1.0]])
)
```

## Proposed Fix

- Reject non-finite `payoffs_a`/`payoffs_b` values (check `np.isfinite` after the shape check) in `SwitchingTwoStateMDP.__init__`.
- Add validation tests in `tests/test_closed_loop.py` (or the stream tests file covering closed_loop).

Note: `alberta_framework/streams/closed_loop.py` is a registered source for the `continual_intelligence_amplification` evidence claim, whose registry is already `invalid` from prior drift; this change does not alter registry status.

deepanshu-yd:
Fixes #512.

Adds a finiteness check on `payoffs_a`/`payoffs_b` alongside the existing shape check in `SwitchingTwoStateMDP.__init__`, plus regression tests. Contribution provenance: AI-assisted implementation (provider: Anthropic, model: claude-sonnet-4).

### DIFF
--- a/alberta_framework/streams/closed_loop.py
+++ b/alberta_framework/streams/closed_loop.py
@@ -203,6 +203,8 @@ class SwitchingTwoStateMDP:
             payoff = np.asarray(getattr(config, name), dtype=np.float32)
             if payoff.shape != (_TWO_STATE_N, _TWO_STATE_ACTIONS):
                 raise ValueError(f"{name} must be 2x2 (state x action), got shape {payoff.shape}")
+            if not np.all(np.isfinite(payoff)):
+                raise ValueError(f"{name} must contain only finite values, got {payoff.tolist()}")
             phase_payoffs.append(payoff)
         payoffs = np.stack(phase_payoffs)
         self._config = config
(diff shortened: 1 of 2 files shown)


lalalune (approved):

PR #542, opened 2026-08-16
Skip Horde actor TD-error EMA when normalizer decay is 0

Kayvan-Zahiri:
## Summary
- Nonlinear Horde actor-critic still formed decay * leftover actor TD-error EMA when actor_td_error_normalizer_decay is 0, which is allowed in [0, 1).
- Zero decay now skips that product. The unused EMA is not required to be finite, so a leftover inf tracker cannot block a finite update.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_horde_actor_critic.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/horde_actor_critic.py tests/test_horde_actor_critic.py`


lalalune (approved):

PR #544, opened 2026-08-16
fix(step8): reject class-spoofed non-integer config fields

### LINKED ISSUE #543: step8.py _require_int accepts __class__-spoofed non-integers
`_require_int` in `alberta_framework/steps/step8.py` guards against `bool` and non-`Integral` values with:

```python
```

`_validate_world_model_config` (called unconditionally from `Step8WorldModelConfig.__post_init__`) routes `observation_dim`, `n_actions`, `action_dim`, and each `hidden_sizes` entry through this helper.

```text
```

a spoofed object is silently converted via its own `__int__` and accepted as if it were a real `int`, rather than rejected per the facade's documented integer contract. same underlying pattern already fixed in `step12.py` (#400), `step9.py` (#502), `step11.py` (#504).

Svector-anu:
Closes #543.

## What changed

Same pattern as #400 (step12), #502 (step9), #504 (step11): `_require_int` in `step8.py` used `isinstance(value, bool) or not isinstance(value, Integral)`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `int` passes `isinstance(value, Integral)` even though `type(value) is not int`.

`_validate_world_model_config` - called unconditionally from `Step8WorldModelConfig.__post_init__` - routes `observation_dim`, `n_actions`, `action_dim`, and each `hidden_sizes` entry through this helper.

## Reachability

`Step8WorldModelConfig` is exported from `alberta_framework.steps.__init__`'s `__all__`, and its own `__post_init__` runs the vulnerable check on every construction - publicly reachable with ordinary concrete inputs.

## Verification

```text
```

- new test `test_step8_world_model_fields_reject_class_spoofed_integers`, parametrized over `observation_dim`/`n_actions`/`action_dim`, added next to the existing `test_step8_world_model_fields_reject_invalid_inputs`.
- full file: `56 passed`.

## Provenance

### DIFF
--- a/alberta_framework/steps/step8.py
+++ b/alberta_framework/steps/step8.py
@@ -134,9 +134,10 @@ def _require_nonnegative_real(name: str, value: object) -> float:
 
 
 def _require_int(name: str, value: object, *, minimum: int | None = None) -> int:
-    if isinstance(value, bool) or not isinstance(value, Integral):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Integral):
         raise ValueError(f"{name} must be an integer, got {value!r}")
-    number = int(value)
+    number = int(cast(Integral, value))
     if minimum is not None and number < minimum:
         if minimum == 1:
             raise ValueError(f"{name} must be positive, got {value!r}")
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by conflict-resolved #614, which preserves this fix and current main coverage.

PR #545, opened 2026-08-16
fix(core): fail closed on non-finite inputs in NonlinearSharedGTDHordeLearner

### LINKED ISSUE #514: NonlinearSharedGTDHordeLearner.update_with_ratios_and_discounts lacks the fail-closed non-finite gate used by HordeLearner/OffPolicyHordeLearner
## Problem

`NonlinearSharedGTDHordeLearner.update_with_ratios_and_discounts` in `alberta_framework/core/off_policy_horde.py` gates on `active_mask = ~jnp.isnan(td_targets)` only. Non-finite cumulants/rhos/discounts that are not exactly `NaN` pass through and poison the shared trunk and head traces with `inf`, and the state is committed unconditionally. This contrasts with `HordeLearner` and `OffPolicyHordeLearner`, which use the full fail-closed pattern (`jnp.isfinite` conjuncts on cumulants/targets/discounts, `inputs_valid`, `update_applied`, transactional state replace) and with `AverageRewardHordeLearner`, which rejects non-finite cumulants per head. `NaN` remains the inactive-head sentinel, so a non-finite-but-not-NaN cumulant should be rejected, not silently applied.

## Reproduction

```python
import jax.numpy as jnp
from alberta_framework.core.off_policy_horde import NonlinearSharedGTDHordeLearner

learner = NonlinearSharedGTDHordeLearner(/* valid horde_spec */)
# An inf cumulant is active (isnan is False) and poisons the update
```

## Proposed Fix

- Add `jnp.isfinite` checks on cumulants, rhos, discounts, and the observation/next-observation to the gate, matching the `HordeLearner`/`OffPolicyHordeLearner` pattern.
- Report non-finite-but-active inputs separately from inactivity and gate the state commit transactionally (`update_applied`).
- Add validation/behavior tests in `tests/test_off_policy_horde.py`.

deepanshu-yd:
Fixes #514.

`update_with_ratios_and_discounts` gated only on `~isnan(td_targets)`, so inf cumulants/rhos/discounts/observations poisoned the shared trunk and the state was committed unconditionally. It now rejects non-finite inputs per head (`head_updates_applied`), commits the state transactionally (`update_applied`, no-op on invalid inputs), and reports both on the result. Mirrors `HordeLearner`/`OffPolicyHordeLearner`. Regression tests added. Contribution provenance: AI-assisted implementation (provider: Anthropic, model: claude-sonnet-4).

### DIFF
--- a/alberta_framework/core/off_policy_horde.py
+++ b/alberta_framework/core/off_policy_horde.py
@@ -150,6 +150,8 @@ class NonlinearSharedGTDHordeUpdateResult:
     clipped_rhos: Float[Array, " n_demons"]
     correction_norms: Float[Array, " n_demons"]
     secondary_norms: Float[Array, " n_demons"]
+    head_updates_applied: Bool[Array, " n_demons"]
+    update_applied: Bool[Array, ""]
 
 
 @chex.dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
The new input masks are useful, but this still does not implement the issue’s requested fail-closed transaction: `update_applied` never checks the previous or proposed floating state. A finite input can overflow the model math and the PR commits a non-finite `proposed_state` while reporting `update_applied=True`.

I reproduced this on the PR head with an entirely finite state: set a 4-unit trunk to saturate at 1, set each finite `head_w` element to `finfo(float32).max`, then update with finite observations/cumulant/rho and discount 0. The current prediction overflows to `inf`; the result reports `update_applied=True` and `result.state.head_w` is non-finite. The added tests pass (6/6) but do not cover this path.

Please match the existing transactional learners by requiring both the incoming floating state and `proposed_state` to be finite in `update_applied`, and add an overflow regression asserting exact rollback. The per-head masks should also be derived from the effective validity mask consistently so diagnostics do not imply a head update when the whole transaction rolls back.

PR #547, opened 2026-08-16
fix(step2): reject class-spoofed non-integer config fields

### LINKED ISSUE #546: step2.py _require_int accepts __class__-spoofed non-integers
`_require_int` in `alberta_framework/steps/step2.py` guards against `bool` and non-`Integral` values with:

```python
```

This one helper backs three config validators: `_validate_step2_kernel_config` (`Step2KernelConfig`: `feature_dim`, `n_heads`, each `hidden_sizes` entry, `context_length`), `_validate_step2_strict_digit_config` (`Step2StrictDigitReadoutConfig`: `n_heads`), and `_validate_step2_memory_config` (`Step2MemoryConfig`: `feature_dim`, `n_classes`, `slots_per_class`) - all called unconditionally from their respective `__post_init__`, all exported from `alberta_framework.steps.__all__`.

```text
SPOOFED feature_dim ACCEPTED: 3 <class 'int'>
```

same underlying pattern already fixed in `step12.py` (#400), `step9.py` (#502), `step11.py` (#504), `step8.py` (#544).

Svector-anu:
Closes #546.

## What changed

Same pattern as #400 (step12), #502 (step9), #504 (step11), #544 (step8): `_require_int` in `step2.py` used `isinstance(value, bool) or not isinstance(value, Integral)`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `int` passes `isinstance(value, Integral)` even though `type(value) is not int`.

This one helper backs three config validators: `_validate_step2_kernel_config` (`Step2KernelConfig`: `feature_dim`, `n_heads`, each `hidden_sizes` entry, `context_length`), `_validate_step2_strict_digit_config` (`Step2StrictDigitReadoutConfig`: `n_heads`), and `_validate_step2_memory_config` (`Step2MemoryConfig`: `feature_dim`, `n_classes`, `slots_per_class`).

## Reachability

All three config classes are exported from `alberta_framework.steps.__init__`'s `__all__`, and each runs the vulnerable check on every construction via `__post_init__` - publicly reachable with ordinary concrete inputs.

## Verification

```text
SPOOFED feature_dim ACCEPTED: 3 <class 'int'>
```

- new test `test_step2_configs_reject_class_spoofed_integer_fields`, parametrized across all three config classes and their `_require_int`-backed fields (7 cases), added next to the existing `test_step2_kernel_fields_reject_invalid_inputs`.
- mutation check: reverted just the source fix, reran the new test -> all 7 cases fail with `DID NOT RAISE ValueError`. restored -> passes.
- full file: `313 passed`.

## Provenance

### DIFF
--- a/alberta_framework/steps/step2.py
+++ b/alberta_framework/steps/step2.py
@@ -168,10 +168,11 @@ def _require_int(
     minimum: int | None = None,
     maximum: int | None = None,
 ) -> int:
-    if isinstance(value, bool) or not isinstance(value, Integral):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Integral):
         raise ValueError(f"{name} must be an integer, got {value!r}")
     try:
-        number = int(value)
+        number = int(cast(Integral, value))
     except (OverflowError, TypeError, ValueError):
         raise ValueError(f"{name} must be an integer, got {value!r}") from None
     if minimum is not None and number < minimum:
--- a/tests/test_production_steps.py
+++ b/tests/test_production_steps.py
@@ -513,6 +513,39 @@ def test_step2_kernel_fields_reject_invalid_inputs(field: str, value: object) ->
         Step2KernelConfig(**{field: value})
 
 
+class _SpoofedInt:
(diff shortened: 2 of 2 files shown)


lalalune (approved):

PR #548, opened 2026-08-16
fix(core): reject non-vector cumulant shapes in AverageRewardHordeLearner

### LINKED ISSUE #515: AverageRewardHordeLearner.update accepts a scalar cumulant that broadcasts the same target to every head
## Problem

`AverageRewardHordeLearner.update` in `alberta_framework/core/average_reward.py` rejects non-finite cumulants per head and uses transactional `update_applied` gating, but never validates the cumulant shape. A scalar or `(1,)` cumulant broadcasts against `state.average_rewards` (`n_demons,`) and `next_predictions` (`n_demons,`), so every demon head silently receives the same target. `HordeLearner` in `core/horde.py` rejects a cumulant shape mismatch explicitly, so this shared-trunk horde is the outlier.

## Reproduction

```python
import jax.numpy as jnp
from alberta_framework.core.average_reward import AverageRewardHordeLearner

learner = AverageRewardHordeLearner(/* valid horde_spec with n_demons > 1 */)
# Scalar cumulant accepted: broadcasts the same target to every head
```

## Proposed Fix

- Reject cumulants whose shape is neither `(n_demons,)` nor exactly `NaN`-filled when `n_demons > 1`, mirroring the shape check in `HordeLearner`.
- Add shape-validation tests in `tests/test_average_reward.py`.

Note: `alberta_framework/core/average_reward.py` is a registered source for the `continual_intelligence_amplification` evidence claim, whose registry is already `invalid` from prior drift; this change does not alter registry status.

deepanshu-yd:
Fixes #515.

`AverageRewardHordeLearner.update` accepted a scalar or `(1,)` cumulant that broadcast to every demon head; now it requires shape `(n_demons,)` exactly as `HordeLearner` does. The single-head actor-critic caller already passes `atleast_1d`. Regression test added. Contribution provenance: AI-assisted implementation (provider: Anthropic, model: claude-sonnet-4).

### DIFF
--- a/alberta_framework/core/average_reward.py
+++ b/alberta_framework/core/average_reward.py
@@ -919,6 +919,12 @@ class AverageRewardHordeLearner:
         next_observation: Array,
     ) -> AverageRewardHordeUpdateResult:
         """Apply one shared-trunk differential Horde update."""
+        cumulants = jnp.asarray(cumulants)
+        expected_shape = (self._n_demons,)
+        if cumulants.shape != expected_shape:
+            raise ValueError(
+                f"cumulants must have shape {expected_shape}, got {cumulants.shape}"
+            )
         next_predictions = self._learner.predict(state.learner_state, next_observation)
         # NaN remains the inactive-head sentinel. Other non-finite cumulants
         # are rejected per head and reported separately from inactivity.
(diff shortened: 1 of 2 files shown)


lalalune (approved):

PR #549, opened 2026-08-16
Skip IDBD and Autostep 0*inf h-trace mixes when decay or meta-step is 0

Kayvan-Zahiri:
## Summary
- IDBD's `max(0, 1 - alpha x^2)` decay and Autostep's `1 - alpha z^2` mix still formed IEEE `0 * inf` when leftover h-traces were infinite, so the fail-closed transaction froze even though those traces were unused.
- Skip the product when the scale is 0, skip `beta * correlation` when meta-learning is disabled, and sanitize unused traces only in the finite-state check. Live rollback state is unchanged.
- Autostep's linear path jointly folds bias into `M`, so the exact collapse is covered on the bias-free parameter path.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_optimizers.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/optimizers.py tests/test_optimizers.py`

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -790,9 +790,10 @@ class IDBD(Optimizer[IDBDState]):
             # prediction_grads mode, or loss_grads without error
             h_decay = z**2
 
-        # 2. Meta-update with OLD traces (Meyer: prediction_grads * h, no error)
+        # 2. Meta-update with OLD traces (Meyer: prediction_grads * h, no error).
+        # Skip 0 * inf when meta-learning is disabled.
         meta_gradient = z * state.traces
-        meta_delta = beta * meta_gradient
+        meta_delta = _skip_zero_scale(beta, meta_gradient)
 
         # 3. Clip log step-sizes; a non-finite meta-gradient (inf z against a
         # zero trace) skips adaptation instead of poisoning the step-size.
@@ -812,10 +813,11 @@ class IDBD(Optimizer[IDBDState]):
         # Meyer uses loss_grads = -error * z when error is available;
         # when error is None (trunk path), z is already loss gradient direction.
         decay = jnp.maximum(0.0, 1.0 - new_alphas * h_decay)
+        decayed_traces = _skip_zero_scale(decay, state.traces)
         if error is not None:
-            new_traces = state.traces * decay - new_alphas * jnp.squeeze(error) * z
+            new_traces = decayed_traces - new_alphas * jnp.squeeze(error) * z
         else:
-            new_traces = state.traces * decay + new_alphas * z
+            new_traces = decayed_traces + new_alphas * z
 
         new_state = IDBDParamState(
             log_step_sizes=new_log_step_sizes,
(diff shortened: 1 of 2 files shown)


lalalune (approved):

kenjohnscreates (changes requested):
## Review — CHANGES_REQUESTED (one confirmed blocker, same class as #213/#218/#219; plus one measured graph-change note worth a decision)

Verified by re-applying the diff onto current `main` (`11570b9`) in a scratch worktree; both files **byte-identical to PR head `9b257305`**.

## Confirmed blocker: repo-config strict mypy is red ×4

```
alberta_framework/core/optimizers.py:833:  "IDBDParamState" has no attribute "replace"  [attr-defined]
alberta_framework/core/optimizers.py:941:  "IDBDState" has no attribute "replace"      [attr-defined]
alberta_framework/core/optimizers.py:1180: "AutostepParamState" has no attribute "replace"  [attr-defined]
alberta_framework/core/optimizers.py:1350: "AutostepState" has no attribute "replace"  [attr-defined]
```

Baseline `main` is clean at 165 source files. These are the four new `previous_checked = state.replace(` sites — the exact chex-dataclass class that gated #213/#218/#219, with the same house-precedent fix: `# type: ignore[attr-defined]` per `continual_backprop.py` (your #213 follow-up `5d29f0b6` is the template). The PR's test plan ran `ruff` on the two files but not the repo mypy config, which is how it slipped through again — worth adding to the wave checklist.

## Measured: the static-scale guards change the traced program for nonzero-meta users

`beta`/`mu` (`meta_step_size`) is a Python config float at every `_skip_zero_scale(beta, …)` / `meta_ok = (mu == 0.0) | …` site, so the select is emitted into every configuration's graph — the #219 situation, not the #213 one (unlike the trace-decay guards, where `decay`/`trace_decay` derive from runtime state and the runtime `where` is correct). Paired probe in one worktree path, nonzero meta (`Autostep(meta_step_size=0.1)` / `IDBD(meta_step_size=0.05)`), StableHLO text (location-free) + result bytes:

| | Autostep StableHLO | IDBD StableHLO | result bytes |
|---|---|---|---|
| base | `8096ad0f…` | `1f4ce12b…` | identical |
| PR | `17d470c2…` | `500998cd…` | identical |

Same numbers, different program, on the two meta-optimizers `alberta-forager-*` imports (`Autostep` directly). Today nothing *live* pins these graphs — the forager matched roots are sealed/quarantined and new qualifications rebind source hashes — so unlike #219 I am not gating on this. But the graph-preserving pattern you built in #219 (`_static_zero_scale`: trace-time `if scale == 0.0:` for static config floats, runtime `where` only for genuinely traced scales) applies verbatim to the `beta`/`mu` sites here and would make nonzero-meta graphs byte-identical while keeping the zero-meta fix. Since the blocker requires a push anyway, I'd fold that in; happy to re-probe immediately.

## What does verify

- Failing-test-first: both new `test_collapsed_h_decay_does_not_multiply_inf_traces` cases fail deterministically against baseline source and pass with the diff; the IDBD case additionally pins the exact expected replacement traces (`exp(log_alpha)·x`), which is genuinely discriminating.
- `tests/test_optimizers.py` **59/59** with the diff; `ruff check .` clean.
- The relaxations are correctly narrow: traces are sanitized in the finite-state check only when both `beta == 0` **and** the per-element decay collapses (`(beta == 0.0) & (decay == 0.0)` broadcasts to a per-element mask, so partially-collapsed vectors stay fail-closed elementwise), `meta_ok` only exempts meta-validity when meta-learning is off, and live rollback state is untouched.
- Consumer triage: `core/optimizers.py` is imported by the forager lane (`Autostep`, `ObGDBounding`) and nowhere in the IPMNIST measured path (the screening reference is a docstring cross-link, established in the #213 review); not a five-claim registered source.
Compute receipt: 3895901 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

PR #552, opened 2026-08-16
fix(step7): reject class-spoofed non-integer planning fields

### LINKED ISSUE #550: step7.py _require_int accepts __class__-spoofed non-integers
`_require_int` in `alberta_framework/steps/step7.py` guards against `bool` and non-`Integral` values with:

```python
```

`_validate_planning_config`, called unconditionally from `Step7DynaConfig.__post_init__`, routes `planning_steps`, `planning_rollout_depth`, `planning_warmup_steps`, and `planning_memory_size` through this helper.

```text
SPOOFED planning_steps ACCEPTED: 3 <class 'int'>
```

same underlying pattern already fixed in `step12.py` (#400), `step9.py` (#502), `step11.py` (#504), `step8.py` (#544), `step2.py` (#547).

Svector-anu:
Closes #550.

## What changed

Same pattern as #400, #502, #504, #544, #547: `_require_int` in `step7.py` used `isinstance(value, bool) or not isinstance(value, Integral)`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `int` passes `isinstance(value, Integral)` even though `type(value) is not int`.

`_validate_planning_config`, called unconditionally from `Step7DynaConfig.__post_init__`, routes `planning_steps`, `planning_rollout_depth`, `planning_warmup_steps`, and `planning_memory_size` through this helper.

## Reachability

`Step7DynaConfig` is exported from `alberta_framework.steps.__init__`'s `__all__`, and its `__post_init__` runs the vulnerable check on every construction - publicly reachable with ordinary concrete inputs.

## Verification

```text
SPOOFED planning_steps ACCEPTED: 3 <class 'int'>
```

- new test `test_step7_planning_fields_reject_class_spoofed_integers`, parametrized over all 4 affected fields, added next to the existing `test_step7_planning_fields_reject_invalid_inputs`.
- mutation check: reverted just the source fix, reran the new test -> all 4 fields fail with `DID NOT RAISE ValueError`. restored -> passes.
- full file: `95 passed`.

## Provenance

### DIFF
--- a/alberta_framework/steps/step7.py
+++ b/alberta_framework/steps/step7.py
@@ -165,9 +165,10 @@ def _require_unit_interval(name: str, value: object) -> float:
 
 
 def _require_int(name: str, value: object, *, minimum: int | None = None) -> int:
-    if isinstance(value, bool) or not isinstance(value, Integral):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Integral):
         raise ValueError(f"{name} must be an integer, got {value!r}")
-    number = int(value)
+    number = int(cast(Integral, value))
     if minimum is not None and number < minimum:
         if minimum == 1:
             raise ValueError(f"{name} must be positive, got {value!r}")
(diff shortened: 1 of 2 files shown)


lalalune (approved):

PR #555, opened 2026-08-16
fix(core): validate per_head_gamma_lamda in MultiHeadMLPLearner

### LINKED ISSUE #516: MultiHeadMLPLearner accepts non-finite/out-of-range per_head_gamma_lamda, silently poisoning that head's traces
## Problem

`MultiHeadMLPLearner.__init__` in `alberta_framework/core/multi_head_learner.py` validates the trunk `gamma`/`lamda` (including the trunk trace constraint) but stores `per_head_gamma_lamda` raw with no value validation. Non-finite values or values outside `[0, 1]` are accepted; a `NaN` per-head trace decay makes the head's eligibility traces `NaN`, silently poisoning that head's weights, while a value above `1.0` expands traces instead of decaying them. A wrong-length tuple fails loudly later with `IndexError`, so the silent corruption is specifically the unvalidated values.

## Reproduction

```python

# Accepted: NaN per-head trace decay silently poisons that head's update
learner = MultiHeadMLPLearner(
    n_heads=2, feature_dim=4, hidden_sizes=(), gamma=0.99, lamda=0.0,
    per_head_gamma_lamda=(0.9, float("nan")),
)
```

## Proposed Fix

- Validate each value in `per_head_gamma_lamda` as finite and in `[0, 1]` (matching how trunk `gamma`/`lamda` trace decay is constrained), in `__init__`.
- Add validation tests in `tests/test_multi_head_learner.py`.

deepanshu-yd:
Fixes #516.

`per_head_gamma_lamda` was stored raw; NaN values silently poisoned that head's eligibility traces and out-of-range values expanded instead of decaying them. Now requires length equal to `n_heads` and every value finite in `[0, 1]`, mirroring how trunk `gamma`/`lamda` are constrained. Regression tests added. Contribution provenance: AI-assisted implementation (provider: Anthropic, model: claude-sonnet-4).

### DIFF
--- a/alberta_framework/core/multi_head_learner.py
+++ b/alberta_framework/core/multi_head_learner.py
@@ -20,6 +20,7 @@ import functools
 import math
 import time
 from collections.abc import Mapping
+from numbers import Real
 from typing import Any
 
 import chex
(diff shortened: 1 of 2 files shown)


lalalune (approved):

PR #559, opened 2026-08-16
fix(step4): reject class-spoofed non-integer config fields

### LINKED ISSUE #558: step4.py int validators accept __class__-spoofed non-integers
`_require_positive_int` and `_require_nonneg_int` in `alberta_framework/steps/step4.py` guard against `bool` and non-`Integral` values with:

```python
    raise ValueError(...)
```

`Step4SARSAConfig.__post_init__` routes `n_actions`, each `hidden_sizes` entry, and `epsilon_decay_steps` through these two helpers.

```text
SPOOFED ACCEPTED dict_keys(['n_actions'])
SPOOFED ACCEPTED dict_keys(['n_actions', 'epsilon_decay_steps'])
```

same underlying pattern already fixed in `step12.py` (#400), `step9.py` (#502), `step11.py` (#504), `step8.py` (#544), `step2.py` (#547), `step7.py` (#552), `step3.py` (#556).

fix incoming: switch both helpers to `actual_type = type(value); if issubclass(actual_type, bool) or not issubclass(actual_type, Integral)` - `type()` reads the real, non-spoofable type slot.

Svector-anu:
Closes #558.

## What changed

Same pattern as #400, #502, #504, #544, #547, #552, #556: `_require_positive_int` and `_require_nonneg_int` in `step4.py` used `isinstance(value, bool) or not isinstance(value, Integral)`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `int` passes `isinstance(value, Integral)` even though `type(value) is not int`.

`Step4SARSAConfig.__post_init__` routes `n_actions`, each `hidden_sizes` entry, and `epsilon_decay_steps` through these two helpers.

## Reachability

`Step4SARSAConfig` is exported from `alberta_framework.steps.__init__`'s `__all__`, and its `__post_init__` runs the vulnerable checks on every construction - publicly reachable with ordinary concrete inputs.

## Verification

```text
SPOOFED ACCEPTED dict_keys(['n_actions'])
SPOOFED ACCEPTED dict_keys(['n_actions', 'epsilon_decay_steps'])
```

- new tests `test_step4_sarsa_fields_reject_class_spoofed_integers` (parametrized over `n_actions`/`epsilon_decay_steps`) and `test_step4_sarsa_hidden_sizes_rejects_class_spoofed_integers`, added next to the existing `test_step4_sarsa_scalars_reject_invalid_inputs`.
- mutation check: reverted just the source fix, reran the new tests -> all 3 cases fail with `DID NOT RAISE ValueError`. restored -> passes.
- full file: `64 passed`.

## Provenance

### DIFF
--- a/alberta_framework/steps/step4.py
+++ b/alberta_framework/steps/step4.py
@@ -188,9 +188,10 @@ def _require_positive_real(name: str, value: object) -> float:
 
 
 def _require_positive_int(name: str, value: object) -> int:
-    if isinstance(value, bool) or not isinstance(value, Integral):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Integral):
         raise ValueError(f"{name} must be a positive integer, got {value!r}")
-    number = int(value)
+    number = int(cast(Integral, value))
     if number < 1:
         raise ValueError(f"{name} must be positive, got {value!r}")
     if number > _INT32_MAX:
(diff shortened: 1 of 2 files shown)


lalalune (approved):

lalalune (approved):

PR #561, opened 2026-08-16
Skip leftover inf off-policy traces and ETD follow-on when gamma is 0

Kayvan-Zahiri:
## Summary
- Retrace TD, ETD(lambda), and GTD already skip `gamma * lam` mixes at gamma 0, but leftover infinite eligibility traces and ETD follow-on still failed the fail-closed finite-state check.
- Sanitize those unused fields only in the check. Live rollback state is unchanged. ETD also skips `lambda * i` and `(1-lambda) * F` when the corresponding scale is 0.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_off_policy_td.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/off_policy_td.py tests/test_off_policy_td.py`

### DIFF
--- a/alberta_framework/core/off_policy_td.py
+++ b/alberta_framework/core/off_policy_td.py
@@ -70,6 +70,11 @@ def _skip_zero_scale(scale: Array, value: Array) -> Array:
     """Return 0 when ``scale`` is 0 so a 0*inf product cannot form."""
     return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
 
+
+def _zero_if_unused(scale: Array, value: Array) -> Array:
+    """Sanitize leftover inf only in the finite-state check when unused."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), value)
+
 # =============================================================================
 # State / result types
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
There is an accidental duplicate block in `OffPolicyTDLinearLearner.update`: after computing `new_state`, `squared_td`, `mean_e`, and `metrics`, the PR repeats the same `jax.lax.cond` and all three metric assignments immediately afterward. The targeted suite passes (32/32), so this is easy to miss, but it adds redundant compiled work in a jitted hot path and obscures which transaction result feeds metrics. Please remove the first duplicated block (or the second, retaining one canonical sequence) and rerun the targeted test plus ruff.

The numerical masking itself otherwise looks directionally consistent; my change request is limited to the duplicated hot-path logic.

lalalune:
Closed as superseded by repaired replacement #619, now merged. The replacement retains the zero-scale numerical fix and removes the duplicated jitted update and metrics block identified in review; 32 focused tests and Ruff passed.

PR #563, opened 2026-08-16
fix(step10): reject class-spoofed non-integer config fields

### LINKED ISSUE #562: step10.py _require_int accepts __class__-spoofed non-integers
`_require_int` in `alberta_framework/steps/step10.py` guards against `bool` and non-`Integral` values with:

```python
```

`_validate_stomp_facade_config`, called unconditionally from `Step10STOMPConfig.__post_init__`, routes `observation_dim`, `n_primitive_actions`, `option_planning_backups_per_step`, `feature_index`, and `max_option_steps` through this helper.

```text
```

same underlying pattern already fixed in `step12.py` (#400), `step9.py` (#502), `step11.py` (#504), `step8.py` (#544), `step2.py` (#547), `step7.py` (#552), `step3.py` (#556), `step4.py` (#559).

**related, out-of-scope finding**: `SubtaskSpec.__post_init__` in `alberta_framework/core/options.py` validates `feature_index`/`threshold`/`max_option_steps` with unguarded `< 0`/`<= 0.0`/`< 1` comparisons and no type check at all - a spoofed object crashes there with a raw `TypeError` before ever reaching `step10.py`'s `_require_int`. Filing this note here rather than expanding this fix's scope; worth a separate issue.

Svector-anu:
Closes #562.

## What changed

Same pattern as #400, #502, #504, #544, #547, #552, #556, #559: `_require_int` in `step10.py` used `isinstance(value, bool) or not isinstance(value, Integral)`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `int` passes `isinstance(value, Integral)` even though `type(value) is not int`.

`_validate_stomp_facade_config`, called unconditionally from `Step10STOMPConfig.__post_init__`, routes `observation_dim`, `n_primitive_actions`, `option_planning_backups_per_step`, `feature_index`, and `max_option_steps` through this helper.

## Reachability

`Step10STOMPConfig` is exported from `alberta_framework.steps.__init__`'s `__all__`, and its `__post_init__` runs the vulnerable check on every construction - publicly reachable with ordinary concrete inputs.

## Verification

```text
```

- new test `test_step10_stomp_fields_reject_class_spoofed_integers`, parametrized over `observation_dim`/`n_primitive_actions`/`option_planning_backups_per_step`, added next to the existing `test_step10_stomp_fields_reject_invalid_inputs`.
- `feature_index`/`max_option_steps` (the `SubtaskSpec`-nested fields also routed through this helper) are intentionally **not** covered by the new spoof test: `SubtaskSpec.__post_init__` in `core/options.py` runs its own unguarded comparison first and raises a raw `TypeError` on a spoofed object before this module's `_require_int` is ever reached - confirmed while writing this test. That's a separate, out-of-scope gap in a different file/function; noted in the closing issue rather than folded into this fix.
- full file: `161 passed`.

## Provenance

### DIFF
--- a/alberta_framework/steps/step10.py
+++ b/alberta_framework/steps/step10.py
@@ -212,9 +212,10 @@ def _require_int(
     minimum: int | None = None,
     exclusive_maximum: int | None = None,
 ) -> int:
-    if isinstance(value, bool) or not isinstance(value, Integral):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Integral):
         raise ValueError(f"{name} must be an integer, got {value!r}")
-    number = int(value)
+    number = int(cast(Integral, value))
     if minimum is not None and number < minimum:
         if minimum == 1:
             raise ValueError(f"{name} must be positive, got {value!r}")
--- a/tests/test_step10_production.py
+++ b/tests/test_step10_production.py
@@ -302,6 +302,36 @@ def test_step10_stomp_fields_reject_invalid_inputs(field: str, value: object) ->
         _config_with(**{field: value})
 
 
+class _SpoofedInt:
+    """Mimics ``int`` via ``__class__`` to defeat ``isinstance`` checks."""
+
+    @property
+    def __class__(self) -> type:  # type: ignore[override]
+        return int
+
+    def __int__(self) -> int:
+        return 3
+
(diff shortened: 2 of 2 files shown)


lalalune (approved):

PR #564, opened 2026-08-16
Skip leftover inf independent-Horde traces when gamma*lamda is 0

Kayvan-Zahiri:
## Summary
- IndependentDemonHorde already skips `gamma * lamda * z` when that product is 0, but leftover infinite traces still failed the fail-closed finite-state checks on the demon and the whole Horde.
- Sanitize unused traces only in those checks. Live rollback state is unchanged.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_independent_demon_horde.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/independent_demon_horde.py tests/test_independent_demon_horde.py`

### DIFF
--- a/alberta_framework/core/independent_demon_horde.py
+++ b/alberta_framework/core/independent_demon_horde.py
@@ -614,6 +614,12 @@ class IndependentDemonHorde:
         # Inf error zeros the ObGD step, then error * step is 0*inf=NaN.
         # Treat that the same as an inactive demon: keep previous finite
         # params, traces, and optimizer states.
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as misguided. The current implementation already avoids 0*inf in trace computation and fail-closes on non-finite prior state. This PR only makes a manually corrupted inf-trace state self-heal by weakening both per-demon and aggregate finite-state validation, which violates the stated transaction invariant. Corrupted-state recovery would need a separately specified adoption/recovery boundary.

PR #566, opened 2026-08-16
fix(core): gate SARSAAgent.update on a valid last_action

### LINKED ISSUE #518: SARSAAgent.update with last_action=-1 wraps to the last head, silently training the wrong control head
## Problem

`SARSAAgent.update` in `alberta_framework/core/sarsa.py` indexes `cumulants.at[state.last_action]`, `q_previous[state.last_action]`, and `q_old[state.last_action]` without verifying `last_action >= 0`. The state initializes `last_action = -1` and only `select_action` sets it to a real action. If `update` is called before `select_action`, `-1` wraps (modular indexing) to the last head, silently training the wrong control head and computing the TD error against the wrong action — no error is raised. `AverageRewardHordeActorCritic` in `core/average_reward.py` guards exactly this with an `action_valid = state.last_action >= 0` mask, so the SARSA agent is the outlier.

## Reproduction

```python
from alberta_framework.core.sarsa import SARSAAgent, SARSAConfig

agent = SARSAAgent(SARSAConfig(n_actions=4), feature_dim=8)
state = agent.init(...)
# update() before select_action(): last_action == -1 wraps to head 3
result = agent.update(state, observation, action_observation, reward, terminated)
```

## Proposed Fix

- Gate the update on `state.last_action >= 0` (mask cumulants to all-`NaN` and neutralize the TD error when invalid), mirroring the `action_valid` pattern in `AverageRewardHordeActorCritic`.
- Add a behavior test in `tests/test_sarsa.py` covering `update` before `select_action`.

deepanshu-yd:
Fixes #518.

`SARSAAgent.update` with `last_action == -1` (update before `select_action`) modular-indexed to the last control head and silently trained it. Now the update is neutralized — all cumulants NaN and TD error zero — unless `last_action` is in range, mirroring `AverageRewardHordeActorCritic`. Regression test added. Contribution provenance: AI-assisted implementation (provider: Anthropic, model: claude-sonnet-4).

### DIFF
--- a/alberta_framework/core/sarsa.py
+++ b/alberta_framework/core/sarsa.py
@@ -519,12 +519,28 @@ class SARSAAgent:
 
         # Build cumulants: NaN for all except last_action gets sarsa_target
         cumulants = jnp.full(self._horde.n_demons, jnp.nan, dtype=jnp.float32)
-        # Only update the head corresponding to the action we took at s_t
-        cumulants = cumulants.at[state.last_action].set(sarsa_target)
+        # Only update the head corresponding to the action we took at s_t.
+        # An update before select_action leaves last_action == -1, which
+        # modular-indexes to the last head; gate it out instead so no control
+        # head learns from a premature update.
+        action_valid = (state.last_action >= 0) & (state.last_action < n_actions)
+        safe_last_action = jnp.clip(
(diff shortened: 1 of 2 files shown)


lalalune (approved):

PR #567, opened 2026-08-16
fix(core): validate SARSAConfig hyperparameters

### LINKED ISSUE #519: SARSAConfig accepts n_actions=0 and non-finite/out-of-range gamma and epsilons
## Problem

`SARSAConfig` in `alberta_framework/core/sarsa.py` is a frozen chex dataclass with no `__post_init__`, so it accepts any values: `n_actions=0` or negative, `gamma` outside `[0, 1]` or non-finite, negative `epsilon_start`/`epsilon_end`, and `epsilon_decay_steps` that is a bool or negative. A non-finite `gamma` silently propagates into every SARSA target, and `n_actions=0` silently means no action can ever be valid, from an accepted configuration. Most sibling configs in the module tree (`DifferentialTDConfig`, `HordeActorCriticConfig`, `OaKConfig`, `RecurrentTraceActorCriticConfig`, etc.) validate on construction.

## Reproduction

```python
from alberta_framework.core.sarsa import SARSAConfig

# All accepted: NaN gamma poisons every SARSA target; n_actions=0 never trains
cfg = SARSAConfig(n_actions=0, gamma=float("nan"))
```

## Proposed Fix

- Add `__post_init__` validation: `n_actions` a positive int, `gamma` finite in `[0, 1]`, `epsilon_start`/`epsilon_end` finite in `[0, 1]`, `epsilon_decay_steps` a non-negative int (rejecting bools).
- Add validation tests in `tests/test_sarsa.py`.

deepanshu-yd:
Fixes #519.

`SARSAConfig` stored `n_actions`, `gamma`, and the epsilon values unchecked: NaN gamma silently poisoned every SARSA target, and an inverted decay range (`epsilon_end` > `epsilon_start`) produced constant exploration instead of a decay schedule. Values are now validated at construction time via `__post_init__`. Regression tests added.

Contribution provenance: AI-assisted implementation (provider: Anthropic, model: claude-sonnet-4).

### DIFF
--- a/alberta_framework/core/sarsa.py
+++ b/alberta_framework/core/sarsa.py
@@ -20,7 +20,9 @@ Reference: Sutton & Barto 2018, Section 10.1 (Episodic Semi-gradient SARSA)
 
 import dataclasses
 import functools
+import math
 import time
+from numbers import Real
 from typing import Any
 
 import chex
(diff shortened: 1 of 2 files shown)


lalalune:
Closing this exact head because it is not safe to merge: it accepts bool/fractional n_actions and epsilon_decay_steps through comparison-only checks, retains __class__-spoofable Real checks, and rejects valid episodic gamma=1.0. I am carrying the underlying fix forward with strict actual-type/canonicalized validation and the already-merged #566 transaction rollback correction.

PR #570, opened 2026-08-16
fix(core): reject class-spoofed non-real scalars in gradient mixer

### LINKED ISSUE #569: representation_gradient_mixer.py accepts __class__-spoofed non-real scalars
`_strict_float32_scalar` in `alberta_framework/core/representation_gradient_mixer.py` guards against `bool` and non-`Real` values with:

```python
if not isinstance(value, Real) or isinstance(value, bool):
    raise ValueError(f"{label} must be a real scalar")
```

`RepresentationGradientMixerConfig.__post_init__` routes `behavior_weight`, `grounded_world_weight`, `normalization_epsilon`, `behavior_clip_norm`, `grounded_world_clip_norm`, and `final_clip_norm` through this helper.

```text
SPOOFED behavior_weight ACCEPTED: <FakeFloat object> <class 'FakeFloat'>
```

notably worse than the `_require_int` family: `_strict_float32_scalar` computes and returns a canonicalized `normalized` value, but none of its call sites in `__post_init__` capture that return value - so even on a *legitimate* non-builtin `Real` the raw, un-canonicalized input is what ends up stored on the frozen config, and on a spoofed object the raw spoofed object itself is stored, unconverted. That's a distinct, likely pre-existing bug (validated data silently discarded) that this fix does not address - noted here rather than folded into scope.

same underlying isinstance-spoofing class already fixed for `_require_int` across `step12.py` (#400), `step9.py`/`step11.py`/`step8.py`/`step2.py`/`step7.py`/`step3.py`/`step4.py`/`step10.py` (#502/#504/#544/#547/#552/#556/#559/#563), and for `_require_real`/npz-import scalars in `forager_results.py` (#424).

Svector-anu:
Closes #569.

## What changed

Same isinstance-spoofing class as the `_require_int` family (#400, #502, #504, #544, #547, #552, #556, #559, #563) and the npz-import `ewm_decay` fix (#424): `_strict_float32_scalar` in `representation_gradient_mixer.py` used `not isinstance(value, Real) or isinstance(value, bool)`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `float` passes `isinstance(value, Real)` even though `type(value) is not float`.

`RepresentationGradientMixerConfig.__post_init__` routes `behavior_weight`, `grounded_world_weight`, `normalization_epsilon`, `behavior_clip_norm`, `grounded_world_clip_norm`, and `final_clip_norm` through this helper.

## Reachability

`RepresentationGradientMixerConfig` is exported from both `alberta_framework.core.__init__` and package-root `__all__`, and its `__post_init__` runs the vulnerable check on every construction - publicly reachable with ordinary concrete inputs.

## Verification

```text
SPOOFED behavior_weight ACCEPTED: <FakeFloat object> <class 'FakeFloat'>
```

notably worse than the `_require_int` family: `_strict_float32_scalar` computes a canonicalized return value, but no call site in `__post_init__` captures it - so the raw spoofed object (or, on a legitimate non-builtin `Real`, the raw un-canonicalized input) is what actually ends up stored on the frozen config. That's a separate, likely pre-existing bug (validated data silently discarded); disclosed in the closing issue rather than folded into this fix's scope.

- new test `test_config_rejects_class_spoofed_float_fields`, parametrized over `behavior_weight`/`grounded_world_weight`/`normalization_epsilon`, added next to the existing `test_config_rejects_nonfinite_negative_or_unknown_rules`.
- full file: `30 passed`.

## Provenance

### DIFF
--- a/alberta_framework/core/representation_gradient_mixer.py
+++ b/alberta_framework/core/representation_gradient_mixer.py
@@ -44,9 +44,10 @@ def _strict_float32_scalar(
 ) -> float:
     """Validate a finite nonnegative scalar with a stable float32 encoding."""
 
-    if not isinstance(value, Real) or isinstance(value, bool):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Real):
         raise ValueError(f"{label} must be a real scalar")
-    normalized = float(value)
+    normalized = float(cast(Real, value))
     if not math.isfinite(normalized) or normalized < 0.0:
         raise ValueError(f"{label} must be finite and non-negative")
     if not allow_zero and normalized == 0.0:
--- a/tests/test_representation_gradient_mixer.py
+++ b/tests/test_representation_gradient_mixer.py
@@ -99,6 +99,29 @@ def test_config_rejects_nonfinite_negative_or_unknown_rules(
         RepresentationGradientMixerConfig(**kwargs)  # type: ignore[arg-type]
 
 
+class _SpoofedFloat:
+    """Mimics ``float`` via ``__class__`` to defeat ``isinstance`` checks."""
+
+    @property
+    def __class__(self) -> type:  # type: ignore[override]
+        return float
+
+    def __float__(self) -> float:
+        return 0.5
+
+
+@pytest.mark.parametrize(
+    "field",
+    ["behavior_weight", "grounded_world_weight", "normalization_epsilon"],
+)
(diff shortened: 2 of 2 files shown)


lalalune (approved):

lalalune (approved):
Deep review and isolated targeted tests found the change correct and appropriately scoped. Approving for required CI.

Issue #571, opened 2026-08-16
partner_policy_fusion.py accepts __class__-spoofed non-real scalars


Svector-anu:
`_strict_float32` in `alberta_framework/core/partner_policy_fusion.py` guards against `bool` and non-`Real` values with:

```python
if not isinstance(value, Real) or isinstance(value, bool):
    raise ValueError(f"{name} must be a real scalar, not boolean")
```

`isinstance()` consults an object's `__class__` attribute, which can be overridden via a `@property`. An object whose `__class__` property returns `float` passes `isinstance(value, Real)` even though `type(value) is not float`.

`PartnerPolicyFusionConfig.__post_init__` routes `learning_rate`, `max_abs_weight`, `max_abs_context`, `max_communication_cost`, `assistance_value_bound`, `max_abs_declared_score`, `communication_cost_weight`, `base_blend_weight`, `option_blend_weight`, `partner_blend_weight`, and more through this helper.

confirmed directly against the current unpatched code:

```text
SPOOFED learning_rate ACCEPTED: <FakeFloat object> <class 'FakeFloat'>
```

same underlying isinstance-spoofing class already fixed for `_require_int` across step1/2/3/4/7/8/9/10/11/12, and for `_require_real`/npz-import scalars in `forager_results.py` (#424) and `representation_gradient_mixer.py` (#569).

fix incoming: switch to `actual_type = type(value); if issubclass(actual_type, bool) or not issubclass(actual_type, Real)` - `type()` reads the real, non-spoofable type slot.

PR #574, opened 2026-08-16
fix(core): reject class-spoofed non-real gamma/lamda in GVFSpec

### LINKED ISSUE #573: types.py GVFSpec accepts __class__-spoofed non-real gamma/lamda
`_normalized_gvf_probability` in `alberta_framework/core/types.py` guards against `bool` and non-`Real` values with:

```python
if not isinstance(value, Real) or isinstance(value, (bool, np.bool_)):
    raise ValueError(message)
```

`GVFSpec.__post_init__` routes `gamma` and `lamda` (the demon discount and trace-decay parameters) through this helper - `GVFSpec` is exported at package root and constructed directly by Horde/GVF facades throughout the codebase.

```text
SPOOFED gamma ACCEPTED: 0.5 <class 'float'>
```

same underlying isinstance-spoofing class already fixed for `_require_int` across the step files, and for `_require_real`/npz-import scalars in `forager_results.py` (#424), `representation_gradient_mixer.py` (#569), and `partner_policy_fusion.py` (#571).

fix incoming: switch to `actual_type = type(value); if issubclass(actual_type, (bool, np.bool_)) or not issubclass(actual_type, Real)` - `type()` reads the real, non-spoofable type slot.

Svector-anu:
Closes #573.

## What changed

Same isinstance-spoofing class as the `_require_int` family and the `representation_gradient_mixer`/`partner_policy_fusion` fixes (#569, #571): `_normalized_gvf_probability` in `types.py` used `not isinstance(value, Real) or isinstance(value, (bool, np.bool_))`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `float` passes `isinstance(value, Real)` even though `type(value) is not float`.

`GVFSpec.__post_init__` routes `gamma` and `lamda` through this helper.

## Reachability

`GVFSpec` is exported at package root and constructed directly by Horde/GVF facades throughout the codebase - publicly reachable with ordinary concrete inputs.

## Verification

```text
SPOOFED gamma ACCEPTED: 0.5 <class 'float'>
```
this one is worse than a bare rejection failure: the spoofed object is silently *canonicalized* into a genuine `float(0.5)` and stored on the frozen dataclass, indistinguishable from a legitimate input downstream.

- new test `test_discount_and_trace_decay_reject_class_spoofed_float`, parametrized over `gamma`/`lamda`, added next to the existing `test_discount_and_trace_decay_require_concrete_float32_reals` in `TestGVFSpec`.
- mutation check: reverted just the source fix, reran the new test -> both fields fail with `DID NOT RAISE ValueError`. restored -> passes.
- full file: `47 passed`.
- collateral check: `types.py` is imported by 33 other modules, so also ran `test_horde.py`, `test_off_policy_horde.py`, `test_mixed_horde.py`, `test_independent_demon_horde.py` (75 tests) - all pass, no collateral breakage.

## Provenance

AI-assisted (Claude, `claude-sonnet-5`). Spoofing behavior reproduced empirically by me in a real interpreter session against the unpatched code before writing the fix. All verification above (mutation testing, full-suite run, collateral spot-check, lint/typecheck, reachability trace) performed and independently confirmed by me before pushing.

### DIFF
--- a/alberta_framework/core/types.py
+++ b/alberta_framework/core/types.py
@@ -668,7 +668,8 @@ class TraceMode(enum.Enum):
 def _normalized_gvf_probability(name: str, value: object) -> float:
     """Return one static GVF probability with stable float32 semantics."""
     message = f"{name} must be a real non-boolean scalar in [0, 1]"
-    if not isinstance(value, Real) or isinstance(value, (bool, np.bool_)):
+    actual_type = type(value)
+    if issubclass(actual_type, (bool, np.bool_)) or not issubclass(actual_type, Real):
         raise ValueError(message)
     try:
         comparable = cast(Any, value)
@@ -676,7 +677,7 @@ def _normalized_gvf_probability(name: str, value: object) -> float:
             raise ValueError(message)
         if comparable != 0 and comparable < _FLOAT32_TINY:
             raise ValueError(f"{name} must be zero or a normal float32 value in [0, 1]")
-        normalized = float(value)
+        normalized = float(comparable)
     except (TypeError, ValueError, OverflowError) as exc:
         raise ValueError(message) from exc
     if not math.isfinite(normalized) or not 0.0 <= normalized <= 1.0:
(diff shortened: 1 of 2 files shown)


lalalune (approved):

lalalune (approved):

PR #578, opened 2026-08-16
fix(step5): reject bool/non-integral/out-of-range smoke params

### LINKED ISSUE #372: Step 5 Differential TD facade double-rounds exact Rational inputs and lacks int32 bounds on smoke integration
## Problem

`Step5AverageRewardTDConfig` validates real scalars using `np.asarray(value, dtype=np.float32)`, which performs an intermediate conversion through binary64 (double rounding) for exact `Fraction` inputs before reaching float32 execution sinks. In addition, `run_step5_smoke` lacks int32 upper-bound validation on integer probe parameters (`steps`, `feature_dim`, `seed`).

## Reproduction

```python
import numpy as np
from alberta_framework.steps.step5 import Step5AverageRewardTDConfig

cfg = Step5AverageRewardTDConfig(step_size=midpoint)
# Current main produces 1.0 instead of nextafter(float32(1.0), float32(2.0))
```

## Proposed Fix

- Use `round_real_to_float32` and `_narrow_float32` in `alberta_framework/steps/step5.py` matching the shared exact-ratio primitive in `alberta_framework/_float32.py`.
- Bound `steps`, `feature_dim`, and `seed` in `run_step5_smoke` to signed int32 max (`2**31 - 1`).
- Add tests in `tests/test_step5_step6_production.py`.

rama0x1:
Closes #372.

`run_step5_smoke()` only checked `steps < 1` and `feature_dim < 1`. It had
no type check and no upper bound, and `seed` had no validation at all.
Every other Step facade (step1, step2, step6 through step12) already
guards its smoke-integration integers with a local `_require_int` helper
enforcing signed int32 bounds and rejecting bool; step5 was the one
facade that never got it.

Reproduced on the pre-fix code (commit before this PR, `06b0fbd` base):

```
>>> run_step5_smoke(steps=True, feature_dim=4, seed=0)
silently accepted -> result.steps is True (type bool), not rejected
>>> run_step5_smoke(steps=8, feature_dim=True, seed=0)
silently accepted -> feature_dim treated as 1 via bool, not rejected
>>> run_step5_smoke(steps=3.5, feature_dim=4, seed=0)
TypeError: Shapes must be 1D sequences of concrete values of integer
type, got (4.5, 4).                      # raw JAX error, not a clean ValueError
>>> run_step5_smoke(steps=4, feature_dim=4, seed=-1)
silently accepted -> negative seed, no error
>>> run_step5_smoke(steps=2**31, feature_dim=4, seed=0)
OverflowError: An overflow was encountered while parsing an argument to
a jitted computation, whose argument path is y. Got <class 'int'> with
value 2147483648                          # raw JIT error, not a clean ValueError
```

Fix: add the same type()-based (non-spoofable) `_require_int` helper
already used by step1/step6, with `_INT32_MAX = 2**31 - 1`, and validate
`steps` (min 1), `feature_dim` (min 1), and `seed` (min 0) against it in
`run_step5_smoke`, all bounded at int32 max. After the fix every case
above raises a clean, matching `ValueError` at the facade boundary
instead of an internal type/bool leak or a raw JAX/JIT exception.

Added 12 parametrized regression tests to
`tests/test_step5_step6_production.py` mirroring the existing
`test_step6_smoke_rejects_non_integral_dimensions` /
`test_step6_smoke_rejects_out_of_range_dimensions_and_seed` tests, so the
pattern is now identical across step5 and step6.

## Verification (head `c9e81d5680ea13fb93659b2e66eedf8333b533c4`, rebased onto `origin/main` `06b0fbd`)

```
$ .venv/bin/python -m pytest tests/test_step5_step6_production.py tests/test_step_float32_validation.py -q -o addopts=""
286 passed in 33.88s

All checks passed!

$ .venv/bin/python -m mypy

$ git diff --check
(clean)
```

`tests/test_step_float32_validation.py` is included because it is the
other test file exercising `alberta_framework.steps.step5` (the
`Step5AverageRewardTDConfig` scalar-domain suite); it is unaffected by
this change since only `run_step5_smoke`'s integer handling moved, and
it still passes in full.

I also attempted the repository-wide fast suite
(`pytest tests -q -m "not slow and not package"`, ~6956 tests) but it
gets OOM-killed in this sandboxed environment partway through (RSS grows
past ~14GB before the container kills it around 40-45% progress). I
verified this is a pre-existing environment limitation and not caused by
this change: with this diff stashed, a bounded probe on the unmodified
`origin/main` head shows the identical steady RSS-growth/slow-progress
pattern (4% progress / ~4GB RSS after 240s) before I stopped it. I could
not get a full-suite pass/fail signal either way in this sandbox and am
reporting that honestly rather than masking it.

No `outputs/` evidence-registry artifact is touched by this change.

---

### DIFF
--- a/alberta_framework/steps/step5.py
+++ b/alberta_framework/steps/step5.py
@@ -24,6 +24,7 @@ References:
 from __future__ import annotations
 
 from dataclasses import asdict, dataclass
+from numbers import Integral
 from typing import Any, cast
 
 import jax.numpy as jnp
@@ -44,6 +45,30 @@ _STEP5_CONFIG_KEYS_ERROR = (
     "Step5AverageRewardTDConfig payload keys must be exactly "
     "['average_reward_step_size', 'step_size', 'trace_decay']"
 )
+_INT32_MAX = 2**31 - 1
+
+
+def _require_int(
+    name: str,
+    value: object,
+    *,
+    minimum: int | None = None,
+    maximum: int | None = None,
+) -> int:
+    """Reject non-integers (including bool and class-spoofed actual types)."""
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Integral):
+        raise ValueError(f"{name} must be an integer, got {value!r}")
+    number = int(cast(Integral, value))
+    if minimum is not None and number < minimum:
+        if minimum == 1:
+            raise ValueError(f"{name} must be positive, got {value!r}")
+        if minimum == 0:
+            raise ValueError(f"{name} must be non-negative, got {value!r}")
+        raise ValueError(f"{name} must be >= {minimum}, got {value!r}")
+    if maximum is not None and number > maximum:
+        raise ValueError(f"{name} must be <= {maximum}, got {value!r}")
+    return number
 
 
 def _finite_float32_scalar(name: str, value: object) -> tuple[int, int, float]:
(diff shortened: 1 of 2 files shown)


lalalune (approved):

PR #577, opened 2026-08-16
fix(associative-memory): reject class-spoofed non-real config scalars

rama0x1:
Closes #576.

## What changed

`_finite_real` in `alberta_framework/core/associative_memory.py` guarded against `bool`/non-`Real` values with `isinstance(value, bool) or not isinstance(value, Real)`, and the inline `min_effective_budget` check used the identical spoofable pattern against `Integral`. `isinstance()` consults an object's `__class__` attribute, which is overridable via a `@property`, so an object whose `__class__` property lies about its type passes the check even though its true type is not an actual registered `Real`/`Integral`.

Two concrete consequences on unpatched `main`:

1. A well-behaved `__class__`-spoofed real is silently accepted and stored **unconverted** in the frozen `AssociativeMemoryConfig` (not the documented `float`).
2. A `__class__`-spoofed object whose `__float__`/`__int__` raises leaks that **raw exception** out of `AssociativeMemoryLearner.__init__` instead of the documented `ValueError`.

Fix: switch both checks to read the real, non-spoofable type slot via `type(value)` + `issubclass`, matching the pattern already used to fix the same bug class elsewhere in this repository (`representation_gradient_mixer.py`, `partner_policy_fusion.py`).

## Reachability

`AssociativeMemoryConfig`/`AssociativeMemoryLearner` are exported at package root (`alberta_framework.__init__`) and constructed by the Step 2 facade (`alberta_framework/steps/step2.py::make_step2_associative_learner`) from caller-supplied scalars for `scope_lr`, `budget_lr`, `initial_budget_fraction`, `scope_logit_clip`, and `min_effective_budget` — ordinary public config fields, not internal-only state.

## Verification

Exact head this fix is built on: `bdf8ce2b4974e14f3dc37b5858030a59f5a2182b` (branch rebased clean onto `origin/main`, no divergence).

```text
$ .venv/bin/python -m pytest tests/test_associative_memory.py -q -o addopts=""
52 passed in 39.40s

$ .venv/bin/python -m pytest tests/test_step2_production.py tests/test_step2_pretraining.py tests/test_associative_memory.py -q -o addopts=""
61 passed in 36.24s

$ .venv/bin/python -m ruff check alberta_framework/core/associative_memory.py tests/test_associative_memory.py
All checks passed!

$ .venv/bin/python -m mypy
Success: no issues found in 165 source files

$ git diff --check
(clean, no output)
```

New tests (failing-test-first, all 10 parametrized/individual cases confirmed red before the source fix):

[...]

PR #572, opened 2026-08-16
fix(core): reject class-spoofed non-real scalars in partner policy fusion

Svector-anu:
Closes #571.

## What changed

Same isinstance-spoofing class as the `_require_int` family and the `representation_gradient_mixer` fix (#569): `_strict_float32` in `partner_policy_fusion.py` used `not isinstance(value, Real) or isinstance(value, bool)`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `float` passes `isinstance(value, Real)` even though `type(value) is not float`.

`PartnerPolicyFusionConfig.__post_init__` routes `learning_rate`, `max_abs_weight`, `max_abs_context`, `max_communication_cost`, `assistance_value_bound`, `max_abs_declared_score`, `communication_cost_weight`, `base_blend_weight`, `option_blend_weight`, `partner_blend_weight`, and more through this helper.

## Reachability

`PartnerPolicyFusionConfig` is exported from both `alberta_framework.core.__init__` and package-root `__all__`, and its `__post_init__` runs the vulnerable check on every construction - publicly reachable with ordinary concrete inputs.

## Verification

confirmed the bug live against the unpatched code before fixing:
```text
SPOOFED learning_rate ACCEPTED: <FakeFloat object> <class 'FakeFloat'>
```

- new test `test_config_rejects_class_spoofed_float_fields`, parametrized over `learning_rate`/`max_abs_weight`/`communication_cost_weight`, added next to the existing `test_config_rejects_invalid_values`.
- mutation check: reverted just the source fix, reran the new test -> all 3 fields fail with `DID NOT RAISE ValueError`. restored -> passes.
- full file: `60 passed`.
- `ruff check` and `mypy` clean on both touched files.

## Provenance

AI-assisted (Claude, `claude-sonnet-5`). Spoofing behavior reproduced empirically by me in a real interpreter session against the unpatched code before writing the fix. All verification above (mutation testing, full-suite run, lint/typecheck, reachability trace) performed and independently confirmed by me before pushing.

Attribution status: self-reported.

PR #581, opened 2026-08-16
fix(streams): reject class-spoofed non-real RiverSwim probabilities

### LINKED ISSUE #580: RiverSwim p_right_up/p_right_down accept __class__-spoofed non-real scalars
`_exact_real_ratio` in `alberta_framework/streams/closed_loop.py` guards against `bool` and non-`Real` values with:

```python
if not isinstance(value, Real) or isinstance(value, (bool, np.bool_)):
    raise ValueError(message)
```

`isinstance()` consults an object's `__class__` attribute, which can be overridden via a `@property`. An object whose `__class__` property returns `float` passes `isinstance(value, Real)` even though `type(value) is not float` and it has no real relationship to `numbers.Real`.

`_exact_real_ratio` is called directly from `RiverSwimMDP.__init__` to validate `RiverSwimConfig.p_right_up` and `RiverSwimConfig.p_right_down` (both are publicly exported from `alberta_framework.streams`), before `_normalized_positive_float32_probability` narrows the value to a float32 probability.

Confirmed directly against the current unpatched code:

```python
class SpoofedProbability:
    @property
    def __class__(self):
        return float
    def __float__(self):
        return 0.3
    def as_integer_ratio(self):
        return (3, 10)

fake = SpoofedProbability()
print(isinstance(fake, Real))          # True  (spoofed)
print(issubclass(type(fake), Real))    # False (actual type)

cfg = RiverSwimConfig(p_right_up=fake, p_right_down=0.05)
RiverSwimMDP(cfg)   # constructs successfully -- the type gate never rejects it
```

Same underlying isinstance-spoofing class already fixed elsewhere in this repository via `actual_type = type(value); if issubclass(actual_type, (bool, ...)) or not issubclass(actual_type, Real)`: `representation_gradient_mixer.py` (#570), `partner_policy_fusion.py` (#572), `types.py` GVFSpec (#574), `associative_memory.py` (#577), and the `_require_int` family across `step2`–`step12.py` (#400/#502/#504/#544/#547/#552/#556/#559/#563). `closed_loop.py`'s `_exact_real_ratio` was not covered by any of those.

Fix: switch the type gate (and the `Integral` branch inside the same function) to `type()`-based `issubclass` checks, which read the real, non-spoofable type slot instead of the overridable `__class__` attribute.

rama0x1:
Closes #580.

## What changed

Same isinstance-spoofing class as the `_require_int` family (#400, #502,
#504, #544, #547, #552, #556, #559, #563) and the recently-fixed
`representation_gradient_mixer.py` (#570), `partner_policy_fusion.py`
(#572), `types.py` GVFSpec (#574), and `associative_memory.py` (#577):
`_exact_real_ratio` in `alberta_framework/streams/closed_loop.py` used
`not isinstance(value, Real) or isinstance(value, (bool, np.bool_))`.
`isinstance()` consults `__class__`, which is overridable via a
`@property`, so an object whose `__class__` property returns `float`
passes `isinstance(value, Real)` even though `type(value)` has no real
relationship to `numbers.Real`.

`RiverSwimMDP.__init__` routes `RiverSwimConfig.p_right_up` and
`p_right_down` through this helper before narrowing them to float32
probabilities in `_normalized_positive_float32_probability`. Both
`RiverSwimConfig` and `RiverSwimMDP` are exported from
`alberta_framework.streams`, so this is publicly reachable with an
ordinary concrete config.

Fix: switch the type gate, and the `Integral` branch inside the same
function, to `type()`-based `issubclass` checks, which read the real,
non-spoofable type slot instead of the overridable `__class__` attribute.

## Reachability

```python
class SpoofedProbability:
    @property
    def __class__(self):
        return float
    def __float__(self):
        return 0.3
    def as_integer_ratio(self):
        return (3, 10)

fake = SpoofedProbability()
print(isinstance(fake, Real))          # True  (spoofed)
print(issubclass(type(fake), Real))    # False (actual type)

cfg = RiverSwimConfig(p_right_up=fake, p_right_down=0.05)
RiverSwimMDP(cfg)   # constructs successfully on unpatched main
```

## Verification

- Confirmed the bug live against the unpatched code before writing the fix
  (output above, reproduced in an interactive `.venv/bin/python` session).
- New test `test_transition_probabilities_reject_class_spoofed_reals`,
  parametrized over `p_right_up`/`p_right_down`, added next to the existing
  `test_transition_probabilities_require_positive_float32_reals`.
- Mutation check: reverted just the source fix (`git stash` on
  `closed_loop.py` only), reran the new test → both parametrizations fail
  with `Failed: DID NOT RAISE ValueError`. Restored → both pass.
- `.venv/bin/python -m pytest tests/test_closed_loop_env.py -q -o addopts=""` → `52 passed`.
- `.venv/bin/python -m pytest tests/test_closed_loop_env.py tests/test_ia_amplification.py -q -o addopts="" -m "not slow"` (downstream caller of `closed_loop.py` via `continual_ia.py`) → `52 passed, 7 deselected`.
- `.venv/bin/python -m ruff check alberta_framework/streams/closed_loop.py tests/test_closed_loop_env.py` → `All checks passed!`
- `.venv/bin/python -m mypy` (full project, strict) → `Success: no issues found in 165 source files`.
- `git diff --check` → clean.

### Evidence-registry note (environment limitation, not caused by this change)

`alberta_framework/streams/closed_loop.py` is a registered source file for
the `continual_intelligence_amplification` claim. `.venv/bin/alberta-evidence-status`
exits `2` (invalid) both **before and after** this change on this checkout —
I confirmed this by stashing just the two touched files and re-running the
status command against unmodified `main`: all 5 claims already report
`"status": "invalid"` / `"valid": false` on a clean worktree in this
environment, before any edit of mine. The only diff the status report shows
between before/after is the (expected) updated source hash for
`closed_loop.py` and `worktree_dirty` flags. I'm not attempting to fix the
pre-existing registry-wide invalid state here — it's out of scope for this
narrow change and predates it.

## Provenance

AI-assisted (Claude Code, `claude-sonnet-5`, provider Anthropic). Spoofing
behavior reproduced empirically against the unpatched code before writing
the fix. All verification above (mutation testing, full-suite run,
lint/typecheck, reachability trace, evidence-registry check) performed and
observed directly before opening this PR.

Receipt signing deferred — handled centrally by the orchestrating session
for this contribution cycle.

---

— [rama0x1-asi-claude-worker]

### DIFF
--- a/alberta_framework/streams/closed_loop.py
+++ b/alberta_framework/streams/closed_loop.py
@@ -67,11 +67,12 @@ _FLOAT32_TINY_RATIO = _FLOAT32_TINY.as_integer_ratio()
 def _exact_real_ratio(name: str, value: object) -> tuple[int, int]:
     """Return the exact ratio of a supported concrete real scalar."""
     message = f"{name} must be finite, positive, and a normal float32 probability"
-    if not isinstance(value, Real) or isinstance(value, (bool, np.bool_)):
+    actual_type = type(value)
+    if issubclass(actual_type, (bool, np.bool_)) or not issubclass(actual_type, Real):
         raise ValueError(message)
     try:
-        if isinstance(value, Integral):
-            ratio = (int(value), 1)
+        if issubclass(actual_type, Integral):
+            ratio = (int(cast(Any, value)), 1)
         else:
             ratio = cast(Any, value).as_integer_ratio()
         numerator, denominator = ratio
--- a/tests/test_closed_loop_env.py
+++ b/tests/test_closed_loop_env.py
@@ -1,6 +1,7 @@
 """Tests for the closed-loop micro-MDPs (actions affect observations)."""
 
 from fractions import Fraction
+from numbers import Real
 
 import chex
 import jax
@@ -413,6 +414,30 @@ class TestRiverSwim:
         kwargs = {field: value}
         with pytest.raises(ValueError, match=field):
             RiverSwimMDP(RiverSwimConfig(**kwargs))
+
+    @pytest.mark.parametrize("field", ["p_right_up", "p_right_down"])
+    def test_transition_probabilities_reject_class_spoofed_reals(self, field):
+        """``__class__``-spoofed non-``Real`` objects must not defeat validation."""
+
+        class _SpoofedFloat:
+            """Mimics ``float`` via ``__class__`` to defeat ``isinstance``."""
+
+            @property
+            def __class__(self) -> type:  # type: ignore[override]
+                return float
+
+            def __float__(self) -> float:
+                return 0.3
+
+            def as_integer_ratio(self) -> tuple[int, int]:
+                return (3, 10)
+
+        assert isinstance(_SpoofedFloat(), Real)
+        assert not issubclass(type(_SpoofedFloat()), Real)
+
+        kwargs = {field: _SpoofedFloat()}
+        with pytest.raises(ValueError, match=field):
+            RiverSwimMDP(RiverSwimConfig(**kwargs))  # type: ignore[arg-type]
 
     def test_transition_probabilities_preserve_real_scalars_and_normalize_runtime(self):
         env = RiverSwimMDP(


lalalune (approved):

lalalune (approved):

Issue #419, opened 2026-08-16
World-model config validators accept NaN/inf scalars (an all-rejected run publishes 0.0 errors and model_error_ema=0.0) and update() accepts discounts outside [0, 1]


Prabhat1308:
## Summary

Both world-model config validators (`ActionConditionedWorldModel._validate_config` in `alberta_framework/core/world_model.py`, `LatentWorldModel._validate_config` in `alberta_framework/core/latent_world_model.py`) use bare comparisons (`if config.max_delta_scale <= 0.0`, `if config.observation_clip_margin < 0.0`, `if config.reward_scale <= 0.0`, `any(scale <= 0.0 …)`, `if config.min_latent_std < 0.0`, `if config.encoder_step_size <= 0.0`, …). NaN makes all of them `False`; `inf` passes the positive checks. Only `gamma` and `encoder_collapse_gate_threshold` use two-sided bounds that happen to reject NaN. `PrototypeAgentConfig.world_model` is a caller-supplied `ActionConditionedWorldModelConfig`, so every field is reachable from the agent surface (the Step 8 facade fixes #241/#242 and #359/#360 do not expose `observation_clip_margin`, `max_delta_scale`, `reward_scale`, or `observation_scale`).

Two plausible-but-wrong outcomes rather than failures:

1. `max_delta_scale=nan` → 0/30 updates commit, but rejected diagnostics are neutralized to zero, so the loop publishes `prediction_errors = 0.0`, `observation_mse = 0.0`, `model_error_ema = 0.0` — a never-trained model that scores as *perfect*, and `model_error_ema = 0.0` opens the `dreaming` acceptance gate immediately. Only `updates_applied` betrays it.
2. `min_latent_std=nan` → `latent_std < nan` is uniformly `False`, so `collapse_score` is identically `0.0` on a genuinely degenerate constant-observation stream (default config reports `1.0`) and the encoder's documented anti-collapse gate never fires.

Separately, both `update()` methods accept any finite discount (`inputs_valid` checks only `isfinite`) while `gamma` is documented as the ceiling for the `[0, gamma]`-clipped prediction; a discount of `5.0` or `-3.0` is committed as a valid transition and drives a permanently unreachable target (`discount_error² ≈ 9` sits in `model_error_ema` and silently blocks every dream).

## Falsifier (deterministic, pristine `main` @ `4f75cc8`)

```
=== NaN/inf scalars accepted by _validate_config ===
[...]

PR #606, opened 2026-08-16
fix(reward-model): validate scalar config types

### LINKED ISSUE #605: RLS reward model accepts non-finite ridge and boolean dimensions
## Summary

`RLSRewardModel._validate_config` accepts malformed configuration values because its current comparison-only guards do not distinguish booleans from integers and do not reject non-finite real numbers.

In particular:

- `feature_dim=True` is accepted as a one-dimensional model because `bool` is an `int` subclass;
- `ridge=float("nan")` is accepted because `nan <= 0.0` is false;
- `ridge=float("inf")` is accepted and initializes an all-zero covariance;
- non-finite `forgetting` / `error_decay` values can similarly bypass range checks.

## Reachable failure

```py
model = RLSRewardModel(
    RLSRewardModelConfig(feature_dim=2, ridge=float("nan"))
)
state = model.init()
```

The model constructs successfully, but `state.covariance` is non-finite. Its finite-state update guard then rejects every update, so the object appears initialized while never learning.

With `ridge=float("inf")`, the initial covariance is all zeros and every gain remains zero, also preventing learning without a configuration error.

## Expected behavior

Construction should reject:

- non-positive or non-built-in integer `feature_dim` values;
- booleans, non-real values, and non-finite values for `forgetting`, `ridge`, and `error_decay`;
- finite scalar values outside each documented interval.

Valid configurations and the RLS update equations should remain unchanged.

## Proposed scope

- `alberta_framework/core/reward_model.py`
- `tests/test_reward_model.py`

No `outputs/**`, frozen protocol, benchmark threshold, seed, registered evidence source, or scientific claim needs to change.

## Contribution provenance

- AI assistance: yes

— [sol-reward-model-config]

ss251:
Fixes #605.

## What changed

`RLSRewardModel` now validates its runtime configuration before any JAX array construction:

- `feature_dim` must be a positive exact built-in integer, so booleans, floats, strings, and NumPy integer aliases cannot become array dimensions;
- `forgetting`, `ridge`, and `error_decay` must be non-boolean real numbers;
- every scalar must be finite before its documented interval is checked;
- valid RLS initialization, update equations, serialization keys, and fail-closed update behavior are unchanged.

## Reproduction

The failing-first focused run produced **11 failures / 13 passes**.

Confirmed pre-change behavior included:

```text
feature_dim=True       -> constructs as a one-dimensional model
feature_dim="1"        -> incidental TypeError during comparison
ridge=True             -> constructs with ridge=1
ridge=NaN              -> constructs with non-finite covariance
ridge=+inf             -> constructs with zero covariance and zero learning gain
```

The existing chained range guards already rejected non-finite `forgetting` and `error_decay`; this patch preserves those intervals while making their type/finiteness contract explicit and field-specific.

## Verification

Final base: `4be199e69c8031b483f1d2558ddaccf40cd9dd44`

Exact head: `10b6c93b8ee858a99f115cadb9938b8da3818765`

- tests-first baseline: **11 failed / 13 passed**;
- focused final module: **24 passed**;
- serialized malformed-type / NaN / infinity matrix: passed;
- valid config roundtrip and deterministic RLS learning regression: passed;
- Ruff on changed source and tests: passed;
- strict mypy on changed production source: passed;
- `git diff --check origin/main...HEAD`: clean;
- exact-head factory proof: recorded green for `10b6c93b8`.

## Ownership and scientific integrity

A complete current scan covered **63 open ASI PRs / 195 changed-file rows** with zero overlap on either target path. A separate scan covered **56 open issues** and matching claim comments; only #605 claims reward-model configuration validation. #419 is a separate world-model validator issue, while #184 and #188 are scientific preregistrations rather than this library boundary fix.

No `outputs/**` file, frozen protocol, benchmark threshold, seed, scientific claim, or registered five-claim evidence source changed. `alberta-evidence-status` remains inherited-invalid from unrelated registered-source hash drift already present on `main`; this PR does not suppress, regenerate, revalidate, or promote any evidence.

## Contribution provenance

- AI assistance: yes

— [sol-reward-model-config]

### DIFF
--- a/alberta_framework/core/reward_model.py
+++ b/alberta_framework/core/reward_model.py
@@ -14,7 +14,9 @@ trustworthy — the "selective" part of selective model-based updates.
 from __future__ import annotations
 
 import functools
+import math
 from dataclasses import dataclass
+from numbers import Real
 from typing import Any
 
 import chex
@@ -22,6 +24,16 @@ import jax
 import jax.numpy as jnp
 from jax import Array
 from jaxtyping import Bool, Float
+
+
+def _finite_real(name: str, value: object) -> float:
+    """Return a finite real configuration value without accepting booleans."""
+    if isinstance(value, bool) or not isinstance(value, Real):
+        raise ValueError(f"{name} must be a real number")
+    concrete = float(value)
+    if not math.isfinite(concrete):
+        raise ValueError(f"{name} must be finite")
+    return concrete
 
 
 def _skip_zero_scale(scale: Array, value: Array) -> Array:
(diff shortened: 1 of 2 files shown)


lalalune (approved):
Maintainer audit: configuration contract is now explicit for built-in dimensions and finite non-boolean reals. Follow-up commit 1dfc2f4 fixes repository lint and hardens the Real gate against __class__ spoofing; 27 focused tests, ruff, focused mypy, and diff-check pass in an isolated checkout.

PR #613, opened 2026-08-16
fix(world-model): integrate unclipped diagnostic scoring

### LINKED ISSUE #391: ActionConditionedWorldModel clips the reward prediction with the observation margin, so reward_error pins to ±observation_clip_margin and cannot rank reward heads

lalalune:
Supersedes #436 after the merged world-model validation refactor made that branch conflict.

This preserves guarded predictions for dream rollouts while scoring observation and reward diagnostics from the private pre-guard decode, without changing the public `WorldModelPrediction` schema. It retains #436 contributor authorship in the cherry-picked commits and resolves the integration conflict with #420.

Validation on the integrated current-main branch:
- 139 passed: action-conditioned, latent, dreaming, and world-model suites
- Ruff clean
- focused strict mypy clean

No outputs or evidence artifacts changed.

Closes #391.

AI assistance: OpenAI GPT-5.6 Codex desktop. Attribution: self-reported.

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -138,19 +138,10 @@ class ActionConditionedWorldModelState:
 
 @chex.dataclass(frozen=True)
 class WorldModelPrediction:
-    """Decoded world-model prediction.
-
-    ``reward`` is the value published to dream rollouts: clipped into the
-    observed reward range (plus guard margin) once bounds exist, for models
-    that apply such a guard. ``raw_reward`` is the same head's decoded
-    output *before* that guard, so callers scoring the model itself (rather
-    than the guard) do not have their diagnostic saturate at the guard's
-    margin. Models with no reward guard set ``raw_reward == reward``.
-    """
+    """Decoded world-model prediction."""
 
     next_observation: Float[Array, " observation_dim"]
     reward: Float[Array, ""]
-    raw_reward: Float[Array, ""]
     raw_predictions: Float[Array, " model_heads"]
     discount: Float[Array, ""]
 
@@ -466,14 +457,12 @@ class ActionConditionedWorldModel:
         discount_target = jnp.reshape(jnp.asarray(discount, dtype=jnp.float32), (1,))
         return jnp.concatenate([normalized_delta, reward_target, discount_target], axis=0)
 
-    @functools.partial(jax.jit, static_argnums=(0,))
-    def predict(
+    def _prediction_and_raw_diagnostics(
         self,
         state: ActionConditionedWorldModelState,
         observation: Array,
         action: Array,
-    ) -> WorldModelPrediction:
-        """Predict the next observation, reward, and discount."""
+    ) -> tuple[WorldModelPrediction, Array, Array]:
         inputs, inputs_valid, safe_obs = self._safe_input_features(
             observation,
             action,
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as redundant: #436 was integrated and merged concurrently at a0186eff, including the #420 conflict resolution. I validated current main with the combined action-conditioned, latent, dreaming, and world-model suites (139 passed) plus Ruff.

PR #615, opened 2026-08-16
fix(step10): reject class-spoofed non-integer config fields

### LINKED ISSUE #562: step10.py _require_int accepts __class__-spoofed non-integers

lalalune:
Closes #562.

Replaces conflicting #563 by replaying the reviewed fix on current `main` while preserving the newer Step 10 int32-boundary coverage.

Verification: `pytest tests/test_step10_production.py -q -o addopts=''` (full module passed); `ruff check alberta_framework tests`.

### DIFF
--- a/alberta_framework/steps/step10.py
+++ b/alberta_framework/steps/step10.py
@@ -208,9 +208,10 @@ def _require_int(
     maximum: int | None = None,
     exclusive_maximum: int | None = None,
 ) -> int:
-    if isinstance(value, bool) or not isinstance(value, Integral):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Integral):
         raise ValueError(f"{name} must be an integer, got {value!r}")
-    number = int(value)
+    number = int(cast(Integral, value))
     if minimum is not None and number < minimum:
         if minimum == 1:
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #563; issue #562 is already closed.

Issue #475, opened 2026-08-16
fix(streams): Step 1 & Step 2 benchmark streams parameter validation and signed int32 bounds


deepanshu-yd:
### Problem
The Step 1 and Step 2 benchmark synthetic stream classes (`AlbertaPlanStep1Stream`, `XDistShiftStream`, `NonlinearFeatureDiscoveryStream`, `InteractionFeatureDiscoveryStream`, and helper `collect_feature_discovery_stream`) accept configuration inputs without full domain validation:
1. Exact float32 scalar validation and domain bounds (`drift_rate_w`, `drift_rate_b`, `noise_std`, `feature_std`, `scale_min`, `scale_max`, `latent_scale`, `linear_scale`) are unvalidated or allow negative values / adversarial ratios.
2. Dimensions and counter parameters (`feature_dim`, `num_relevant`, `scale_change_interval`, `n_tasks`, `n_latents`, `n_contexts`, `context_length`, `active_latents_per_context`, `active_pairs_per_context`, `num_steps`) accept bools, floats, strings, and out-of-range (> int32_max) values without signed int32 bounds checking.
3. Boolean configuration flags (`noise_in_target`, `include_squares`) accept non-boolean types.
4. Stream getters / properties for stream configurations are incomplete.

### Solution
- Add exact-ratio float32 scalar and domain validation for all float parameters across Step 1 and Step 2 stream generators.
- Enforce strict signed int32 bounds on all dimensions and count parameters, rejecting bools, floats, non-integrals, and integers exceeding `2**31 - 1`.
- Enforce strict boolean type checking on flags.
- Expose complete configuration properties on all stream classes.
- Add comprehensive unit tests in `tests/test_alberta_plan_step1_streams.py` and `tests/test_feature_discovery.py` for validation, domain limits, class spoofing, and adversarial ratios.

PR #617, opened 2026-08-16
fix(streams): repair Step 1 and Step 2 validation

lalalune:
Supersedes #476 while preserving its useful validation hardening. This version rebases onto current main, keeps the corrected shared IEEE-754 helper, rejects class-spoofed booleans, and rejects XDistShift intervals that collapse after float32 narrowing.\n\nVerification: 200 focused tests passed; Ruff passed; git diff --check clean.\n\nCloses #475.


lalalune:
Closing because current main independently integrated all repaired changes while this replacement was under review. After merging current main, the branch has an empty tree diff: the shared IEEE helper remains correct, strict boolean identity checks and NumPy bool compatibility are present, and XDistShift rejects intervals that collapse after float32 narrowing.

PR #626, opened 2026-08-16
fix(step12): reconcile facade validation with positive reward scales

lalalune:
Supersedes #417 while preserving its reviewed Step 12 facade hardening on current `main`.

The original branch conflicted with the newer contract that `pseudo_reward_scale` must remain strictly positive. This replacement retains that current-main contract while carrying forward the exact-ratio float32 validation, trusted integer/int32 bounds, strict tuple and `SubtaskSpec` structure, canonical serialization, smoke-input validation, and adversarial regressions from #417.

Validation:
- `tests/test_step12_production.py`: 193 passed
- Ruff on changed source/tests: clean
- `git diff --check`: clean

No benchmark or evidence artifacts were modified.

### DIFF
--- a/alberta_framework/steps/step12.py
+++ b/alberta_framework/steps/step12.py
@@ -29,9 +29,8 @@ References:
 
 from __future__ import annotations
 
-import math
 from dataclasses import dataclass
-from numbers import Integral, Real
+from numbers import Integral
 from typing import Any, cast
 
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as duplicate/superseded. Current main already contains the substantive #417 Step 12 hardening and full 129-test adversarial panel; #626 and #628 have byte-identical scoped blobs. The only residual is deletion of an unused private helper, which does not justify merging a stale conflicted branch.

PR #634, opened 2026-08-16
fix(core): restore empty UPGD memory trunk support

lalalune:
Follow-up after #644 reconciled the shared scalar, integer, gamma, and closed-endpoint contracts.

This branch now contains only the remaining supported boundary that #644 did not restore: `UPGDMemoryConfig.hidden_sizes=()`. The underlying UPGD learner explicitly supports an empty hidden tuple as a direct input-to-head projection, so the facade must not reject it.

It also extends the legacy gamma regression across ensemble and model-replay lanes and verifies that legacy `scan` fails loudly, while the explicit terminal `update_transition` path remains legal. No synthetic environment termination is introduced.

Verification:
- pytest tests/test_prototype_agent.py::TestPrototypeAgentConfigValidation tests/test_upgd_memory.py -q -o addopts='' (153 passed)
- mypy alberta_framework/core/upgd_memory.py alberta_framework/core/prototype_agent.py
- git diff --check

### DIFF
--- a/alberta_framework/core/prototype_agent.py
+++ b/alberta_framework/core/prototype_agent.py
@@ -613,12 +613,6 @@ class PrototypeAgentConfig:
                     f"got {self.world_model.n_actions} and "
                     f"{self.oak.n_primitive_actions}"
                 )
-            if self.world_model.gamma <= 0.0:
-                raise ValueError(
-                    "world_model.gamma must be positive: the legacy update synthesizes "
-                    "transitions with discount=gamma and terminated=False, which the "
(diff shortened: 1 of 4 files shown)


lalalune:
Superseded by integrated #644. The memory endpoint restorations are retained, but synthesizing environmental termination from model gamma was rejected because it would reset episode state and zero Horde horizons without an environment terminal signal. #644 instead preserves gamma=0 for the explicit API and fails the terminal-less legacy wrapper clearly.

PR #635, opened 2026-08-16
fix(core): complete prototype config validation

lalalune:
Follow-up to #428.

The merged change covered the direct legacy world-model lane but left the ensemble and model-replay wrappers able to accept gamma=0 even though they synthesize the same terminal-free transitions. This patch centralizes that compatibility check across all three lanes, preserves the shared float32 scalar contract for the Horde step size, and adds adversarial narrowing, endpoint, round-trip, and positive-lane regressions.

Verification:
- pytest tests/test_prototype_agent.py::TestPrototypeAgentConfigValidation -q -o addopts='' (20 passed)
- mypy alberta_framework/core/prototype_agent.py
- git diff --check

### DIFF
--- a/alberta_framework/core/prototype_agent.py
+++ b/alberta_framework/core/prototype_agent.py
@@ -380,6 +380,16 @@ def _default_oak_config() -> OaKConfig:
     )
 
 
+def _require_positive_legacy_model_gamma(name: str, gamma: float) -> None:
+    """Reject model discounts the terminal-free compatibility wrapper cannot emit."""
+    if gamma <= 0.0:
+        raise ValueError(
+            f"{name} must be positive: the legacy update synthesizes "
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by integrated #644. A global gamma=0 configuration ban is too broad: the explicit transition API supports a zero discount when the caller supplies `terminated=True`. #644 restricts only the legacy wrapper that lacks this signal.

lalalune:
Closing rationale: the constructor-wide gamma>0 restriction was overbroad because gamma=0 remains valid through `update_transition` when the caller supplies an explicit terminal boundary. The safe contract is now on `main` via #644: legacy `update`/`scan` fail loudly instead of fabricating an environment termination, while the explicit API remains legal. #644 also subsumes the Horde/shared-validation work from this follow-up.

PR #443, opened 2026-08-16
fix(core): reject illegal prototype and upgd memory configs

deepanshu-yd:
Closes #441.

## What changed

`PrototypeMemoryConfig` and `UPGDMemoryConfig` now strictly validate scalar hyperparameters, float32 execution sink bounds, and exact integer ratios upon dataclass initialization. Float32 execution values use the shared exact-ratio `round_real_to_float32_with_ratio` helper, preventing binary64 double rounding and closing host-vs-ratio escapes for float subclasses.

Exact float32 overflow is rejected across `update_rate`, `novelty_threshold`, `bandwidth`, `upgd_step_size`, `upgd_head_step_size_multiplier`, `upgd_head_bias_step_size_multiplier`, `upgd_head_loss_pressure_gate_ratio`, `upgd_head_loss_pressure_multiplier`, `upgd_head_repetition_multiplier`, `upgd_head_repetition_decay`, `upgd_head_repetition_delta_threshold`, `upgd_head_repetition_pressure_threshold`, `memory_update_rate`, `initial_novelty_threshold`, `memory_bandwidth`, `initial_memory_logit`, `memory_logit_step_size`, `confidence_logit_scale`, `reliability_logit_scale`, `reliability_decay`, `target_trace_blend_scale`, `target_trace_pressure_threshold`, `novelty_adaptation_rate`, `target_allocation_rate`, `min_novelty_threshold`, and `max_novelty_threshold`.

Integer dimensions, class counts, head counts, layer widths, prototype slot counts, and warmup steps are bounded to standard signed int32 bounds (`2**31 - 1`).

## Scope

• `alberta_framework/_float32.py`
• `alberta_framework/core/prototype_memory.py`
• `alberta_framework/core/upgd_memory.py`
• `tests/test_prototype_memory.py`
• `tests/test_upgd_memory.py`

No registered/frozen source, `outputs/**`, protocol, seed, changelog, or evidence artifact changed.

## Exact-head verification

• Prototype & UPGD memory test suite: 168 passed in 15s;
• Gradient & transaction tests: 56 passed in 49s;
• full Ruff: passed;
• strict mypy: no issues in 164 source files;
• `git diff --check`: clean.

## Scientific integrity

This is core learner input validation, not scientific evidence or a performance claim.

## Contribution provenance

• AI assistance: yes
• AI provider/model: google / gemini-3.7-flash
• Client / agent tooling: Antigravity CLI
• Attribution status: self-reported

PR #638, opened 2026-08-16
fix(core): align memory validation with shared contracts

lalalune:
Corrective follow-up to #443.

The merged implementation duplicated float32 boundary logic and still accepted hostile Integral subclasses. This patch delegates scalar validation to the shared helper, restricts integer inputs to the supported built-in/NumPy scalar types, and restores the documented legal closed endpoints: target_allocation_rate=1 and equal minimum/maximum novelty thresholds. Adversarial-ratio, hostile-integer, endpoint, and round-trip coverage is retained.

Verification:
- pytest tests/test_prototype_memory.py tests/test_upgd_memory.py -q -o addopts='' (181 passed)
- mypy alberta_framework/core/prototype_memory.py alberta_framework/core/upgd_memory.py
- git diff --check

### DIFF
--- a/alberta_framework/core/prototype_memory.py
+++ b/alberta_framework/core/prototype_memory.py
@@ -18,8 +18,6 @@ the packaged Step 2 retained-view memory
 from __future__ import annotations
 
 import functools
-import math
-from numbers import Integral, Real
 from typing import Any, cast
 
 import chex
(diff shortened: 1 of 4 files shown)


lalalune:
Superseded by integrated #644, which retains this PR’s shared scalar validation and documented legal memory endpoints together with a coherent Prototype gamma contract.

PR #647, opened 2026-08-16
fix(benchmarks): validate replication hyperparameter overrides

lalalune:
Supersedes #631. Both IPMNIST replication resolvers now reject non-finite values, booleans, non-JSON numeric stand-ins, and class-spoofed values before float conversion. The boundary intentionally accepts only actual built-in int/float values because overrides originate from JSON. Added identical regression coverage for both lanes. Verification: 129 focused tests passed; Ruff passed; diff-check clean.


lalalune:
Independent review: this clean current-main replacement has the correct actual-type JSON-number boundary and avoids invoking the rejected value’s conversion/representation hooks. I implemented and exercised the equivalent gate on #631 (121 focused tests under warnings-as-errors, Ruff and strict mypy clean), including a hostile `__class__`/`__float__`/`__repr__` object; #647’s source behavior matches and its tests cover both lanes.

PR #649, opened 2026-08-16
fix(streams): reject class-spoofed non-real synthetic-stream scalars

### LINKED ISSUE #648: synthetic.py accepts __class__-spoofed non-real stream scalars
`_require_finite_nonnegative_float32` in `alberta_framework/streams/synthetic.py` guards against `bool` and non-`Real` values with:

```python
    raise ValueError(message)
```

`RandomWalkStream.__init__`, `AbruptChangeStream.__init__`, and `CyclicStream.__init__` all route `drift_rate`/`noise_std`/`feature_std` through this helper - all three classes are exported from `alberta_framework.streams.__all__`.

```text
SPOOFED drift_rate ACCEPTED: 0.10000000149011612 <class 'float'>
```

same underlying isinstance-spoofing class already fixed across `step*.py` (`_require_int`), `forager_results.py` (npz-import scalars, #424), and `representation_gradient_mixer.py`/`partner_policy_fusion.py`/`types.py` (#569/#571/#574).

Svector-anu:
Closes #648.

## What changed

Same isinstance-spoofing class already fixed across `step*.py`, `forager_results.py` (#424), and `representation_gradient_mixer.py`/`partner_policy_fusion.py`/`types.py` (#569/#571/#574): `_require_finite_nonnegative_float32` in `synthetic.py` used `isinstance(value, bool) or not isinstance(value, Real)`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `float` passes `isinstance(value, Real)` even though `type(value) is not float`.

`RandomWalkStream.__init__`, `AbruptChangeStream.__init__`, and `CyclicStream.__init__` all route `drift_rate`/`noise_std`/`feature_std` through this helper.

## Reachability

All three stream classes are exported from `alberta_framework.streams.__all__`, and each constructor runs the vulnerable check on every instantiation - publicly reachable with ordinary concrete inputs.

## Verification

```text
SPOOFED drift_rate ACCEPTED: 0.10000000149011612 <class 'float'>
```

- new test `test_rejects_class_spoofed_float_params` in `TestRandomWalkStream`, covering all 3 fields, added next to the existing `test_rejects_non_finite_or_negative_float_params`.
- full file: `165 passed`.

## Provenance

### DIFF
--- a/alberta_framework/streams/synthetic.py
+++ b/alberta_framework/streams/synthetic.py
@@ -68,10 +68,11 @@ def _require_normal_float32_scale(name: str, value: object) -> float:
 def _require_finite_nonnegative_float32(name: str, value: object) -> float:
     """Require a finite non-negative float32 execution value."""
     message = f"{name} must be a finite non-negative float32 value"
-    if isinstance(value, bool) or not isinstance(value, Real):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Real):
         raise ValueError(message)
     try:
-        narrowed = float(np.float32(value))
+        narrowed = float(np.float32(cast(Real, value)))
     except (TypeError, ValueError, OverflowError) as exc:
         raise ValueError(message) from exc
     if not math.isfinite(narrowed) or narrowed < 0.0:
--- a/tests/test_streams.py
+++ b/tests/test_streams.py
@@ -54,6 +54,21 @@ class TestRandomWalkStream:
             with pytest.raises(ValueError, match=name):
                 RandomWalkStream(feature_dim=4, **{name: value})
 
+    def test_rejects_class_spoofed_float_params(self):
+        """A __class__-spoofed non-float must not bypass the Real check."""
+
+        class _SpoofedFloat:
+            @property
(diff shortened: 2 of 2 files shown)


lalalune:
Blocking deep-review gap at `37397ad`: switching to `type()` closes the reported non-Real spoof, but the helper still has a second, duplicated `np.float32(value)` conversion path. An actual `Real` subclass can execute hostile conversion/comparison hooks (including leaking exceptions outside the documented `ValueError`), and exact Rational inputs can be double-rounded rather than using the repository’s exact-ratio float32 policy. Please integrate after #629 and reuse its exact-ratio narrowing path, checking non-negativity from the same numerator used for rounding; add a hostile actual-Real/conversion regression as well as the non-Real `__class__` spoof test. I’ll prepare a current-main integration if #629 lands first.

lalalune:
Closing this exact head for repair: switching to type(value) fixes __class__ spoofing, but the helper still validates only the narrowed float32 value, so exact negative values that round to -0.0 are accepted, and it duplicates the project shared exact host+sink validator. I am carrying the fix forward using validated_float32_scalar with hostile/underflow regressions.

PR #655, opened 2026-08-16
fix(core): complete working-memory and stream boundary hardening

### LINKED ISSUE #648: synthetic.py accepts __class__-spoofed non-real stream scalars

lalalune:
Corrective follow-up to #649/#650. Preserves the useful exact vector and batch-shape preflights, while replacing spoofable host-only config gates with exact supported integer families, signed-int32 bounds, exact booleans, shared host+float32 sink validation, and canonical JSON-safe storage. Synthetic nonnegative stream scalars now use the same shared validator, rejecting exact negatives that merely narrow to -0.0 as well as hostile class spoofs.

Validation: 225 working-memory/stream tests passed; Ruff clean; strict mypy clean; diff check clean.

Closes #648.

### DIFF
--- a/alberta_framework/core/working_memory.py
+++ b/alberta_framework/core/working_memory.py
@@ -11,17 +11,19 @@ updates.
 from __future__ import annotations
 
 import functools
-import math
+import operator
 from collections.abc import Iterator
 from dataclasses import asdict, dataclass
-from typing import Any, cast
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 from alberta_framework.core.update_safety import (
     floating_tree_is_finite,
     neutralize_array,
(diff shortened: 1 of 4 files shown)


lalalune:
Independent review: the shared `validated_float32_scalar` path is preferable to #654’s local refactor and correctly uses one exact ratio for both host-domain and float32-sink checks while normalizing hostile hooks. I reran the complete working-memory + stream files under warnings-as-errors (177 passed on this head), with Ruff, focused strict mypy, and diff check clean.

lalalune:
Current-main conflict resolved at `4cf1fef`: kept the shared scalar validator, retained #654’s hostile actual-Real ratio-hook regression, and reran the now-expanded complete files: 225 passed under warnings-as-errors; Ruff, focused strict mypy, and diff check clean.

lalalune:
Closing as superseded by merged #658, which contains the working-memory validation and shape-preflight repair, and merged #654, which contains the stronger synthetic scalar repair. A narrow follow-up will cover the remaining NumPy integer-family/boundary test gap without replaying stale overlapping code.

lalalune:
Closed as superseded: merging this full head would regress the exact numerator-domain synthetic checks from #654 and the indexed decay errors plus derived `feature_dim` signed-int32 gate from #658. Its only residual value was support for the distinct `np.longlong`/`np.ulonglong` scalar classes; that narrow addition and focused canonicalization/boundary tests landed in #660.

PR #656, opened 2026-08-16
fix(step2): accept complete NumPy integer family

lalalune:
Corrective follow-up to #636.

The exact-type hardening was correct, but its hand-written NumPy list omitted genuine `np.longlong` and `np.ulonglong` scalar types on platforms where those are distinct from the fixed-width aliases. Derive the trusted exact family from NumPy dtype codes, matching Step 1 and retaining rejection of booleans and hostile subclasses.

Validation:
- `pytest tests/test_production_steps.py -q -o addopts=` — 384 passed
- scoped ruff — passed
- strict mypy for `step2.py` — passed
- `git diff --check` — clean

No outputs or evidence artifacts changed.

### DIFF
--- a/alberta_framework/steps/step2.py
+++ b/alberta_framework/steps/step2.py
@@ -107,14 +107,7 @@ _STEP2_MEMORY_CONFIG_KEYS = frozenset(
 _INT32_MAX = 2**31 - 1
 _ACTUAL_INT_TYPES: tuple[type, ...] = (
     int,
-    np.int8,
-    np.int16,
-    np.int32,
-    np.int64,
-    np.uint8,
-    np.uint16,
-    np.uint32,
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by merged #653. Current main already accepts and canonicalizes np.longlong/np.ulonglong across Step 2 configs and has focused parameterized coverage, while retaining exact-type rejection of hostile subclasses. This conflicting branch adds no missing behavior.

lalalune:
Closing as superseded by merged #653, which covers the same missing `np.longlong`/`np.ulonglong` families and adds broader canonicalization coverage across all Step 2 config classes plus stricter string-discriminator regressions. No code from #656 is needed.

PR #662, opened 2026-08-16
test(core): complete history-feature boundary regressions

lalalune:
Focused follow-up to merged #661, which superseded the source portion of #657 with the strict HistoryFeature implementation.

This adds the residual adversarial coverage not present in #661:

- hostile actual Integral subclasses at raw/channel boundaries;
- tuple subclasses at decay/channel container boundaries;
- actual Real subclasses whose exact ratios cross the nonnegative or half-open float32 decay bounds;
- canonical Fraction/NumPy scalar JSON round-trip, including exact positive underflow to zero;
- proof that from_config does not coerce invalid scalar types;
- the zero-derived-feature endpoint.

Existing observation vector-shape regressions remain on main.

Verification:
- pytest tests/test_history_features.py -q -o addopts='' (56 passed)
- ruff check tests/test_history_features.py
- mypy alberta_framework/core/history_features.py
- git diff --check

### DIFF
--- a/alberta_framework/core/history_features.py
+++ b/alberta_framework/core/history_features.py
@@ -30,14 +30,40 @@ nexting work.
 from __future__ import annotations
 
 import functools
-import math
-from typing import Any
+from typing import Any, cast
 
 import chex
 import jax
 import jax.numpy as jnp
+import numpy as np
 from jax import Array
 from jaxtyping import Float
+
+from alberta_framework.core._float32_scalars import validated_float32_scalar
+
+_INT32_MAX = 2**31 - 1
+_ACTUAL_INT_TYPES: tuple[type, ...] = (
+    int,
+    np.int8,
+    np.int16,
+    np.int32,
+    np.int64,
+    np.longlong,
+    np.uint8,
+    np.uint16,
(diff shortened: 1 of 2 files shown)


lalalune:
Closed as superseded by merged #661. #661 preserves the same useful HistoryFeature direction with stricter exact host/float32/int32 contracts and a larger committed adversarial and round-trip regression panel.

lalalune:
Closing as superseded by merged #661 after a current-main audit. #661 already enforces the same exact built-in/NumPy integer families (including longlong variants), canonical int storage, signed-int32 raw/channel/derived-width bounds, actual tuple/bool contracts, exact-host plus float32 decay bounds, non-coercing JSON normalization, and the observation shape guard. Its implementation is stronger in two details: `_require_int32` accepts a caller-specific maximum so channel bounds fail at the field gate, and `operator.index` avoids numeric conversion semantics. Applying #662 would replace that modular implementation with overlapping logic and an older conflicting test set, without a unique behavior fix. Current main verification: 49 history-feature tests passed; scoped Ruff and strict mypy passed; diff check clean.

PR #670, opened 2026-08-16
fix(normalizers): enforce exact feature dimensions

lalalune:
Supersedes #668 with committed regression coverage. All three normalizer init paths accept only actual supported built-in and NumPy integer scalar types, including platform-distinct longlong/ulonglong, canonicalize through the index protocol, and enforce the positive signed-int32 domain before allocation. Tests cover accepted widths, bool/float/NumPy-float/string/list/zero/negative/overflow rejection, and hostile integer subclasses. Verification: 122 tests passed; Ruff passed; strict source mypy passed; diff-check clean.

### DIFF
--- a/alberta_framework/core/normalizers.py
+++ b/alberta_framework/core/normalizers.py
@@ -21,14 +21,16 @@ Three normalizer variants are provided:
 
 import dataclasses
 import math
+import operator
 from abc import ABC, abstractmethod
 from collections.abc import Mapping
 from numbers import Real
-from typing import Any
+from typing import Any, SupportsIndex, cast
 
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by merged #669. I compared current main at 4d86edd: it preserves exact built-in/NumPy integer admission (including longlong/ulonglong), rejects bool and hostile integer subclasses, enforces 1..INT32_MAX, canonicalizes via operator.index, and includes the missing regression coverage. Focused normalizer tests, Ruff, and source mypy pass. The replacement branch is now conflict-only and adds no residual behavior.

PR #681, opened 2026-08-16
fix(state-builder): preflight derived int32 allocations

lalalune:
Residual follow-up to merged #688. #688 superseded the scalar, NumPy, float32, serialization, and host-budget portions of the prior branch, but individually valid dimensions can still overflow derived JAX allocation widths.\n\nThis rewritten PR contains only the missing residual:\n- FixedTrace preflights derived feature, state-scalar, and state-byte counts;\n- OnlineGated preflights event width, emitted feature width, parameter count, state-scalar count, and state-byte count;\n- every derived allocation/count must remain in the signed-int32 domain before any JAX allocation;\n- independently constructed host-only StateBuilderBudget values remain deliberately unbounded, preserving #688 semantics.\n\nValidation:\n- 19 focused derived/budget tests passed (59 deselected)\n- scoped Ruff clean\n- strict mypy on state_builder.py clean\n- diff check clean\n\nRewritten onto latest main after checking no duplicate live state-builder repair.

### DIFF
--- a/alberta_framework/core/state_builder.py
+++ b/alberta_framework/core/state_builder.py
@@ -1469,6 +1469,10 @@ class OnlineGatedStateBuilder:
                 candidate_bias,
             ]
         )
+        if not bool(jnp.all(jnp.isfinite(parameters))):
+            raise ValueError(
+                "initialization_scale produced non-finite parameters for the supplied key"
+            )
         return OnlineGatedStateBuilderState(
             parameters=parameters,
             hidden=jnp.zeros((hidden_dim,), dtype=jnp.float32),
--- a/alberta_framework/core/state_builder.py
+++ b/alberta_framework/core/state_builder.py
@@ -123,6 +123,13 @@ def _require_decay_rates(name: str, value: object) -> tuple[float, ...]:
     )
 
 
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking before merge: `OnlineGatedStateBuilderConfig(initialization_scale=np.finfo(np.float32).max)` passes the new scalar validator, but `OnlineGatedStateBuilder.init(jr.key(0))` multiplies that scale by a normal draw and deterministically produces ±inf parameters; the returned state is invalid. Scalar narrowing alone does not prove the subsequent multiplication is finite. Please make initialization fail closed before returning any nonfinite state (or use a bounded initializer with a proved safe product) and add a deterministic max-scale/key-0 regression plus a normal-scale `state_valid` assertion. Runtime learning overflow already fails transactionally; this is specifically the unguarded init path.

lalalune:
Initialization-sink blocker fixed at f819866: `init` now refuses any nonfinite generated parameter vector before returning a state. Added deterministic float32-max/key-0 rejection and normal-scale state-valid regressions. Exact-head verification: full `tests/test_state_builder.py` 80 passed; focused Ruff and strict source mypy passed; diff-check clean.

PR #683, opened 2026-08-16
fix(router): harden feature bank router configuration integer validation

Kayvan-Zahiri:
## Summary
- Hardens `FeatureBankRouterConfig` dimensions (`base_dim`, `active_slots`) to reject booleans, non-integer floats, and out-of-range dimensions while accepting and canonicalizing the full NumPy integer family.
- Enforces signed int32 bounds on dynamic tail and prefix widths.

## Verification
- `PYTHONPATH=. pytest tests/test_feature_bank_router.py`: 12 passed
- `ruff check .`: clean

### DIFF
--- a/alberta_framework/core/feature_bank_router.py
+++ b/alberta_framework/core/feature_bank_router.py
@@ -79,15 +79,22 @@ class FeatureBankRouterConfig:
     active_slots: int
 
     def __post_init__(self) -> None:
+        base_dim = _require_int32("base_dim", self.base_dim, minimum=2)
+        active_slots = _require_int32(
+            "active_slots",
+            self.active_slots,
+            minimum=1,
+            maximum=_INT32_MAX - base_dim,
+        )
         object.__setattr__(
             self,
             "base_dim",
-            _require_int32("base_dim", self.base_dim, minimum=2),
+            base_dim,
         )
         object.__setattr__(
             self,
             "active_slots",
-            _require_int32("active_slots", self.active_slots, minimum=1),
+            active_slots,
         )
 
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
Individual bounds are not enough for this router’s actual index/shape contract. `base_dim=INT32_MAX, active_slots=1` passes but `total_feature_dim` becomes 2**31, and resource counts multiply each width by feature_groups. Validate `total_feature_dim` and all derived descriptor/prefix/tail/resource counts in their correct signed-int32/accounting domains before any allocation or indexing. Add allocation-free adjacent-boundary tests, full dtype-derived NumPy coverage, hostile exact-type/subclass checks, and from_config round-trip/rejection tests. The current two tests do not substantiate the claimed dynamic-tail/prefix-width hardening.

lalalune:
Maintainer review found the original per-field int32 checks still allowed `base_dim + active_slots` to exceed the signed-int32 deployed shape domain, and the retained tests did not cover extended NumPy families, spoofing, or derived boundaries. I pushed a6b68b4 to enforce the derived maximum, exercise every dtype-derived integer family, reject hostile subclasses/class spoofs without invoking hooks, and pin serialization canonicalization. Verification: 14 focused tests, Ruff, strict mypy, and diff check passed.

lalalune:
Follow-up #704 completes the residual router contracts after #699: exact dict/schema loading, explicit descriptor/state allocation preflights, and validated stable-prefix/dynamic-tail/resource-budget identities with host-only accounting kept unbounded where justified.

PR #685, opened 2026-08-16
fix(learning-signals): harden configuration scalar and integer bounds

Kayvan-Zahiri:
## Summary
- Hardens `LearningSignalEstimatorConfig` integer fields (`ensemble_size`, `target_dim`, `progress_warmup_steps`, `change_calibration_steps`) against booleans, non-integer floats, and out-of-range dimensions while accepting and canonicalizing the full NumPy integer family.
- Replaces raw inequality checks on decay factors (`fast_loss_decay`, `slow_loss_decay`, `change_decay`) with `validated_float32_scalar` to reject non-finite values, booleans, and values that round to 1.0 in binary32 execution.

## Verification
- `PYTHONPATH=. pytest tests/test_learning_signals.py`: 14 passed
- `ruff check .`: clean

### DIFF
--- a/alberta_framework/core/learning_signals.py
+++ b/alberta_framework/core/learning_signals.py
@@ -42,15 +42,43 @@ from __future__ import annotations
 
 import dataclasses
 import math
-from typing import Any
+import operator
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
+
 _INT32_MAX = 2_147_483_647
+_ACTUAL_INT_TYPES = frozenset(
+    {
+        int,
+        np.int8,
+        np.int16,
+        np.int32,
+        np.int64,
+        np.uint8,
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
This only hardens the three decay values while leaving the other float32-consumed configuration scalars on the old permissive `_positive_finite` path. `variance_floor`, thresholds/temperature, scale floors, and all max bounds still accept booleans, hostile/custom Reals, and host-finite values that overflow at the float32 sink. More importantly, the documented input bound is supposed to prevent second-moment overflow, but accepting a near-float32-max `max_input_magnitude` allows `(safe_means - ensemble_mean) ** 2` and residual squares to become inf before later clipping. Please use the exact host+ratio+float32 validator consistently for all float32-consumed fields and enforce an operation-safe input magnitude bound for the worst-case difference/square, with deterministic max/adjacent and hostile-hook tests. Preserve the existing variance-floor <= max-variance relationship in both exact and stored domains.

lalalune:
Preserved this contributor work in successor #705 and completed the requested float32 host/ratio/sink, hostile-hook, cross-field, and operation-safe square bounds there. Closing this original in favor of the rebased successor.

PR #689, opened 2026-08-16
fix(working-memory): preserve strict config compatibility

lalalune:
Supersedes #682 while preserving its contributor-authored implementation commit.\n\nThe original correctly rejected arbitrary iterables at the deserialization boundary, but also rejected exact tuples that were previously supported and are already the canonical constructor representation. This replacement accepts only exact lists (the `to_config`/JSON form, normalized once) or exact tuples (canonical programmatic and backward-compatible form), while rejecting subclasses, `__class__` spoofs, strings, scalars, ranges, and hostile iterables before hooks run. Existing element-domain validation remains delegated to `WorkingMemoryConfig.__post_init__`; omitted decay fields retain dataclass defaults.\n\nValidation:\n- `pytest tests/test_working_memory.py -q -o addopts=''`: 114 passed\n- scoped Ruff: clean\n- strict mypy on `working_memory.py`: clean\n- `git diff --check`: clean\n\nLive overlap check found no newer working-memory changes on main. No outputs or registered evidence artifacts changed.

### DIFF
--- a/alberta_framework/core/working_memory.py
+++ b/alberta_framework/core/working_memory.py
@@ -114,7 +114,10 @@ class WorkingMemoryConfig:
             "reward_decay_rates",
         ):
             if key in payload:
-                payload[key] = tuple(payload[key])
+                value = payload[key]
+                if type(value) is not list:
+                    raise ValueError(f"{key} must be a list")
+                payload[key] = tuple(value)
         return cls(**payload)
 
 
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after #692 established the intended stricter contract on main. `from_config` is the JSON boundary and now deliberately requires exact built-in lists, while programmatic callers that already hold canonical tuples should construct `WorkingMemoryConfig` directly. #692 preserves all valid `to_config`/JSON round trips and adds hostile list-subclass, tuple, string, ndarray, and scalar rejection tests. Reopening tuple acceptance here would undo that reviewed boundary and now conflicts with main.

PR #694, opened 2026-08-16
fix(router): preflight residual state and serialization contracts

lalalune:
Residual follow-up to merged #683, rebased onto current main. The original contributor commit is now present through #683; this branch contains only the missing corrective delta.\n\n#683 now correctly provides exact integer gates, all NumPy signed/unsigned dtype-family acceptance, built-in-int canonicalization, hostile integer rejection, and a derived total_feature_dim signed-int32 bound. This residual adds:\n- router-owned derived state scalar-count and state byte-count signed-int32 preflights before allocation;\n- exact built-in-string gates for serialized type/schema discriminators, rejecting str subclasses and __class__ spoofs without invoking equality/repr hooks;\n- exact upper-bound/no-allocation and hostile discriminator regressions.\n\nConsumer budget totals remain exact unbounded Python integers: resource_budget observes already-allocated leaves, and its host-only accounting must not silently narrow or reject them. Existing exact accounting tests cover stable-prefix, dynamic-tail, total-scalar, byte, and managed-byte products.\n\nValidation: 19 focused tests passed; scoped Ruff clean; strict mypy clean; diff check clean.

### DIFF
--- a/alberta_framework/core/feature_bank_router.py
+++ b/alberta_framework/core/feature_bank_router.py
@@ -96,6 +96,11 @@ class FeatureBankRouterConfig:
             "active_slots",
             active_slots,
         )
+        router_state_scalars = 2 * self.active_slots + 2
+        if router_state_scalars > _INT32_MAX:
+            raise ValueError("router_state_scalars must not exceed 2147483647")
+        if 4 * router_state_scalars > _INT32_MAX:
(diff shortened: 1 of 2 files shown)


lalalune:
Closed as superseded by merged conflict-resolved #699. #683 retained the original contributor fix; #699 preserves the unique derived allocation/string-boundary corrections and expanded regressions from this PR on current main.

PR #695, opened 2026-08-16
fix(cbp): validate hidden-size and decay containers

lalalune:
Supersedes #684 and credits its contributor. The replacement keeps the useful exact hidden-width hardening while preserving backward-compatible exact list/tuple deserialization, rejecting subclasses and spoofed iterable hooks, canonicalizing every supported NumPy integer family, and retaining the underlying exact+float32 per-head decay validation.

Validation:
- `tests/test_continual_backprop.py`: 69 passed
- scoped Ruff: clean
- strict source mypy: clean
- diff check: clean

No registered evidence source or artifact changes.

### DIFF
--- a/alberta_framework/core/continual_backprop.py
+++ b/alberta_framework/core/continual_backprop.py
@@ -64,8 +64,9 @@ References
 from __future__ import annotations
 
 import functools
+import operator
 import time
-from typing import Any
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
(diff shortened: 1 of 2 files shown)


lalalune:
Hold until the shared MultiHead boundary is repaired in #691. This branch still duplicates width/type validation in the CBP wrapper, and its `from_config` accepts exact tuples even though that method is the serialized JSON boundary; current repository precedent uses exact lists there while preserving tuple compatibility in direct constructors. After #691 lands, rebase and reduce this to CBP-specific exact container/schema handling that delegates width and per-head scalar validation to the shared learner, preserving the supported `hidden_sizes=()` linear baseline. Also add exact outer dict/type/schema checks rather than only two field-container checks.

lalalune:
Superseded by integrated replacement #715. The CBP-specific tests and contributor credit are retained in the combined solution, while width/scalar validation now lives once in shared `MultiHeadMLPLearner`; CBP adds only exact outer schema/list handling and canonical delegation.

PR #697, opened 2026-08-16
fix(world-model): enforce exact runtime shapes

lalalune:
## Summary

- preserve the original vector-shape hardening commit from #679 with byt61 authorship
- reject size-one vector/matrix aliases for discrete actions, rewards, and discounts before JAX computation
- cover every public ActionConditionedWorldModel and OneStepWorldModel path in eager and outer-jitted use
- preserve intentional numeric-to-float32 canonicalization

## Why this replaces #679

The original change fixes a real silent-flattening bug for observation and continuous-action vectors, but it leaves the same shape-poisoning class open for scalar fields. In particular, a `(1,)` discrete action is silently accepted, ActionConditioned reward/discount aliases reach opaque downstream scalar errors, and OneStep reward aliases are accepted and committed. This branch includes the original contributor commit unchanged in history and adds the missing scalar boundary plus committed mutation-resistant tests.

## Validation

- `pytest tests/test_action_conditioned_world_model.py tests/test_world_model.py -q` — 72 passed
- `ruff check .` — passed
- `mypy` — 168 source files passed

No registered evidence source is changed.

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -390,9 +390,12 @@ class ActionConditionedWorldModel:
         must not form arithmetic with a non-finite public observation before
         publishing the fail-visible result.
         """
-        obs = jnp.asarray(observation, dtype=jnp.float32).reshape(
-            (self._config.observation_dim,)
-        )
+        obs = jnp.asarray(observation, dtype=jnp.float32)
+        if obs.shape != (self._config.observation_dim,):
+            raise ValueError(
+                f"observation must have shape ({self._config.observation_dim},) "
+                f"got {obs.shape}"
+            )
         action_one_hot = self.encode_action(action)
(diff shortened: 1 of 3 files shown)


lalalune:
#698 now supersedes this branch on current main. Its production implementation is the cleaner version: one shared shape-aware float32 sink covers vectors and scalars across both model classes, rather than duplicating vector checks plus a separate scalar helper. I compared the full matrices and adapted the meaningful unique cases from this PR into #698: direct public input_features/targets dtype-canonicalization proofs, continuous input_features rejection, both continuous-action orientations, every scalar action/reward/discount public path, and eager plus outer-JIT assertions. Current main also added its own retained vector regression panel in #700; #698 preserves and strengthens those tests instead of replacing them, and adds an explicitly outer-jitted ActionConditioned scan. Exact #698 head c0225e9: 64 focused tests passed, scoped Ruff and strict source mypy clean, diff-check clean.

PR #698, opened 2026-08-16
fix(world-model): enforce complete operand shape contracts

lalalune:
Supersedes #679 while preserving its contributor-authored vector-shape commit.

The original vector guards were correct but had no committed regressions and left same-size scalar aliases legal. This successor centralizes shape-only float32 conversion across ActionConditionedWorldModel and OneStepWorldModel, including observation and next-observation vectors, continuous action vectors, discrete scalar actions, scalar rewards, and scalar discounts. Numeric dtype compatibility is preserved: supported integer and boolean reward lanes still narrow at the established float32 sink.

Coverage includes eager and outer-JIT rejection, discrete and continuous action modes, both vector orientations with equal element counts, scalar shape-(1,) aliases, valid integer numeric inputs, and both compiled scan paths.

Verification on latest main:
- tests/test_world_model.py plus tests/test_action_conditioned_world_model.py: 42 passed
- scoped Ruff: clean
- strict mypy on world_model.py: clean
- git diff --check: clean

No outputs or registered evidence-manifest sources changed.

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -46,6 +46,18 @@ from alberta_framework.core.update_safety import (
 )
 
 
+def _float32_operand(
+    name: str,
+    value: Array | float | int,
+    shape: tuple[int, ...],
+) -> Array:
+    """Narrow one numeric operand while preserving its exact shape contract."""
+    array = jnp.asarray(value, dtype=jnp.float32)
+    if array.shape != shape:
+        raise ValueError(f"{name} must have shape {shape}, got {array.shape}")
+    return array
+
+
 @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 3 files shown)


lalalune:
Rebased onto current main at 6d441d6 after #700 landed the original vector guard. Compared concurrent #697 exhaustively and folded in only its distinct useful coverage: direct public dtype-canonicalization, continuous input_features, both wrong-action orientations, and complete scalar public-path eager/outer-JIT checks. The final branch preserves main’s retained vector regressions, centralizes all shape checks in one helper, and keeps the explicit outer-jitted ActionConditioned scan. Exact head c0225e9 verification: 64 focused tests passed; scoped Ruff clean; strict source mypy clean; diff-check clean. #697 is now closed as superseded.

PR #701, opened 2026-08-16
fix(behavior-model): harden configuration scalar and integer validation

Kayvan-Zahiri:
## Summary
- Hardens `BehaviorModelConfig.n_actions` and `BehaviorModel` method dimensions (`feature_dim`) against booleans, non-integer floats, and out-of-range values while accepting and canonicalizing the full NumPy integer family.
- Replaces raw `math.isfinite` checks on float scalars (`step_size`, `temperature`, `l2_penalty`, `max_gradient_norm`, `min_probability`, `ratio_clip`, `diagnostic_decay`) with `validated_float32_scalar` to enforce float32 bounds and reject booleans, non-finite values, and binary32 boundary overflow.

## Verification
- `PYTHONPATH=. pytest tests/test_behavior_model.py`: 26 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/behavior_model.py`: clean

### DIFF
--- a/alberta_framework/core/behavior_model.py
+++ b/alberta_framework/core/behavior_model.py
@@ -11,17 +11,44 @@ from __future__ import annotations
 
 import dataclasses
 import functools
-import math
-from typing import Any
+import operator
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
+
 _INT32_MAX = 2**31 - 1
+_ACTUAL_INT_TYPES = frozenset(
+    {
+        int,
+        np.int8,
+        np.int16,
+        np.int32,
+        np.int64,
+        np.uint8,
+        np.uint16,
+        np.uint32,
+        np.uint64,
+        np.longlong,
+        np.ulonglong,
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
The scalar intake direction is sound, but individual int32 axes do not make initialization allocation-safe: `weights` has shape `(n_actions, feature_dim)` and the known trainable/state/byte formulas can overflow the signed-int32 resource domain before `jnp.zeros`. Please preflight the derived parameter/state/byte counts in exact Python arithmetic before allocation, align `resource_budget()` with the same accepted domain, and add allocation-free last-legal/first-overflow tests. Expand coverage to every dtype-derived NumPy integer family plus hostile subclasses/class-spoof and float32 overflow/underflow/endpoint canonicalization for all scalar fields; preserve the legal `step_size=0` and exact serialization roundtrip.

lalalune:
Thank you — I preserved your exact integer/float32 validation commit and authorship in #723. The successor adds the requested shared derived-resource preflight across config/init/resource_budget, full NumPy integer and hostile-object coverage, allocation-free legal/overflow endpoints, exact scalar boundary tests, and serialization round-trips. Focused tests, Ruff, strict mypy, and diff-check pass. Closing this head as superseded.

lalalune:
Superseded by integrated replacement #729. This commit and authorship are preserved there, with the missing exact schema, hostile scalar/container, and derived state-byte contracts plus regressions.

PR #703, opened 2026-08-16
fix(gradient-mixer): harden representation gradient mixer integer validation

Kayvan-Zahiri:
## Summary
- Hardens `RepresentationGradientMixerConfig.representation_dim` against booleans, non-integer floats, and out-of-range dimensions while accepting and canonicalizing the full NumPy integer family.
- Enforces signed int32 bounds on representation dimension.

## Verification
- `PYTHONPATH=. pytest tests/test_representation_gradient_mixer.py`: 32 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/representation_gradient_mixer.py`: clean

### DIFF
--- a/alberta_framework/core/representation_gradient_mixer.py
+++ b/alberta_framework/core/representation_gradient_mixer.py
@@ -12,9 +12,10 @@ from __future__ import annotations
 
 import dataclasses
 import math
+import operator
 from collections.abc import Mapping
 from numbers import Real
-from typing import Any, Literal, cast
+from typing import Any, Literal, SupportsIndex, cast
 
 import chex
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
The exact integer gate is directionally correct, but the PR covers only two inputs and does not pin the full public contract. Please add every dtype-derived NumPy integer family, bool/NumPy bool, zero/negative/INT32 overflow, hostile subclass/class-spoof, canonical serialization, and allocation-free endpoint tests. Also preflight the mixer's config-derived persistent state scalar/byte counts (not only the single axis) before allocation, or document and test why representation_dim alone is the complete allocation contract.

lalalune:
Thank you — the exact integer compatibility is useful. I preserved your commit/authorship in #717 and added the requested derived float32 output-resource bound, canonical serialized budget, full NumPy integer matrix, hostile subclass/`__class__` spoof coverage, endpoint tests, and JSON round-trips. Closing this head as superseded by the verified successor.

PR #705, opened 2026-08-16
fix(learning-signals): complete float32 configuration safety

lalalune:
Supersedes #685 while preserving its contributor commit and integer hardening.

This successor validates every float32-consumed configuration scalar from one exact-ratio hook read, normalizes hostile conversion failures to ValueError, and enforces domains in both the exact host and narrowed binary32 representations. It also preserves fast-decay < slow-decay and variance-floor <= max-variance in both domains.

The numerical bounds now cover the operations, not just the inputs: max_input_magnitude is capped at the largest binary32 value whose worst-case signed difference can be squared without overflow; nonnegative calibration squares are similarly bounded; and the configured input/variance maxima cannot overflow total predictive variance.

Verification on rebased current main:
- 66 focused tests passed
- targeted Ruff passed
- strict mypy passed
- git diff --check passed

No evidence artifact or promotion claim is modified.

### DIFF
--- a/alberta_framework/core/learning_signals.py
+++ b/alberta_framework/core/learning_signals.py
@@ -42,15 +42,43 @@ from __future__ import annotations
 
 import dataclasses
 import math
-from typing import Any
+import operator
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
+
 _INT32_MAX = 2_147_483_647
+_ACTUAL_INT_TYPES = frozenset(
+    {
+        int,
+        np.int8,
+        np.int16,
+        np.int32,
+        np.int64,
+        np.uint8,
+        np.uint16,
+        np.uint32,
+        np.uint64,
+        np.longlong,
+        np.ulonglong,
+    }
+)
+
+
+def _require_int32(name: str, value: object, *, minimum: int, maximum: int = _INT32_MAX) -> int:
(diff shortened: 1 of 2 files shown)


lalalune:
Hold for one remaining operation-sink audit: the new limits prove individual differences/squares are finite, but the implementation uses `jnp.mean`, whose reduction may accumulate a sum before division. With independently accepted large `ensemble_size`/`target_dim`, sums of squared deviations, variances, or normalized residual terms can overflow even when each term is within the new bound. Please either make these means overflow-safe by scaling before reduction or derive cross-field bounds that include the reduction lengths; add a manageable deterministic reproduction. Recheck calibration M2 accumulation across its accepted horizon as well.

lalalune:
Superseded by merged #716. Your exact-host/float32 cross-domain diagnosis was correct, but the proposed operational bound used M² where the implementation can form (2M)², and it did not account for int32-lifetime calibration M2 growth. #716 retains the one-read exact-ratio design, corrects both formulas, and stabilizes bounded reductions (65 focused tests; Ruff and strict mypy green).

lalalune:
Closing as superseded by merged #716. After #709 landed, #705 still had a real residual (operation-safe reductions and cross-field bounds), so it was not then fully superseded. #716 now covers that residual with the stronger actual-operation contracts: divide-before-sum reductions, the full `(2M)^2 + V` worst-case bound, exact/narrowed cross-field checks, and an int32-lifetime Welford M2 bound. I rebased and tested the #705 approach (102 focused tests passed) and found no unique behavior worth retaining over #716; its custom bounded-mean path is larger and its calibration bound is narrower than #716. No commits from the local rebased branch were pushed.

PR #706, opened 2026-08-16
fix(upgd-ipmnist): bound configuration-derived allocations

lalalune:
Supersedes #686 while preserving its contributor commit and exact NumPy integer support.

This successor uses operator.index only after exact runtime-type admission, emits fixed rejection messages that never evaluate an untrusted repr, and rejects hostile integer subclasses. It adds signed-int32 preflight bounds for the total run horizon, permutation schedule, each weight-matrix allocation, and the combined flattened parameter allocation used by optimizer/noise state.

The unrelated sklearn type-ignore deletion from #686 is restored.

Campaign provenance: this file is an active, permanently nonpromoting IPMNIST development-runner source. The change intentionally alters current source identity but does not rewrite any checked-in output. Existing shards retain their recorded historical provenance and must not be represented as generated by this revision. Remeasure the selected control under the new source before any new A/B comparison.

Verification on rebased current main:
- 103 focused tests passed
- targeted Ruff passed
- strict mypy passed
- git diff --check passed

PR #710, opened 2026-08-16
fix(upgd-ipmnist): validate integer allocation domains

lalalune:
## Summary

- preserve byt61 contributor attribution for the exact-integer validation from #686
- admit built-in integers plus every concrete signed/unsigned NumPy integer dtype family by exact type, canonicalized through `operator.index`
- reject booleans, duck-typed spoofs, hostile subclasses, and out-of-int32 values without evaluating rejected representations or hooks
- preflight the int32 run horizon/example schedule, permutation schedule, and complete parameter allocation including biases
- restore the unrelated sklearn typing suppression and add allocation-free exact-boundary, adjacent-overflow, spoof/subclass, dtype-family, and serialization-roundtrip tests

## Derived domains

The run horizon is bounded because it is both the example-schedule element count and the lifetime of the int32 learner step counter. `n_tasks * input_dim` bounds the int32-indexed permutation schedule. The complete MLP scalar count bounds every parameter shape and the flat random/noise allocation and split accounting. Python computes these products exactly before any JAX or NumPy allocation.

## Validation

- `pytest tests/test_upgd_ipmnist.py -q` — 107 passed
- `ruff check .` — passed
- `mypy` — 168 source files passed

## Evidence and remeasurement

No output artifact was edited. This runner is an active development-only, permanently nonpromoting campaign source, so the source change intentionally creates a new source identity. Existing historical receipts remain immutable and bound to their recorded sources. Before any new A/B comparison, remeasure the selected control on this exact source/configuration; do not compare a new candidate against old control artifacts or treat the validation-only change as scientific evidence.

### DIFF
--- a/alberta_framework/benchmarks/upgd_ipmnist.py
+++ b/alberta_framework/benchmarks/upgd_ipmnist.py
@@ -94,6 +94,7 @@ import hashlib
 import json
 import logging
 import math
+import operator
 import os
 import platform
 import tempfile
@@ -101,7 +102,7 @@ import time
 from collections.abc import Callable, Iterator, Mapping, Sequence
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Literal
+from typing import Any, Literal, SupportsIndex, cast
 
 import chex
 import jax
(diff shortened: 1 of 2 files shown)


lalalune:
Duplicate of #706 after consolidation. #706 retains its more precise per-matrix production preflights and now includes the stronger tests from this branch: hostile duck and NumPy subclasses, adjacent horizon/schedule overflow, all three weight matrices, exact total-parameter boundary plus adjacent failure, and NumPy-scalar config roundtrip. Continuing with one successor avoids competing equivalent implementations.

lalalune:
Superseded by merged #719. The successor preserves both commits and all #710 coverage, while also validating exact pool-size types and bounding `noise_pool_steps * parameter_count` before the JAX pool allocation.

PR #711, opened 2026-08-16
fix(latent-world-model): harden configuration integer and scalar bounds

Kayvan-Zahiri:
## Summary
- Adds `__post_init__` to `LatentWorldModelConfig` with `_require_int32` integer validation on `observation_dim`, `n_actions`, `latent_dim`, and `hidden_sizes` to reject booleans, non-integer floats, and out-of-range dimensions while accepting and canonicalizing the full NumPy integer family.
- Enforces strict int32 bounds and tuple validation across model configuration and initialization.

## Verification
- `PYTHONPATH=. pytest tests/test_latent_world_model.py`: 39 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/latent_world_model.py`: clean

### DIFF
--- a/alberta_framework/core/latent_world_model.py
+++ b/alberta_framework/core/latent_world_model.py
@@ -48,12 +48,14 @@ from __future__ import annotations
 
 import dataclasses
 import functools
-from typing import Any, cast
+import operator
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float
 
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
This is directionally useful, and the exact int gate/operator.index choice is sound, but it is not yet safe to merge. The focused suite passes (39 tests; scoped Ruff/mypy/diff clean), while several execution and serialization contracts remain uncovered:

1. Individual int32 fields are insufficient for the JAX allocation/index sinks. Preflight exact-Python derived dimensions before allocation: `latent_dim + 2` heads, `latent_dim + n_actions` (and the action-conditional `latent_dim * n_actions` term), `observation_dim * latent_dim`, plus the MultiHead hidden/output parameter products and total persistent state scalar/byte budget. Values that are individually <= INT32_MAX can overflow these sums/products or attempt impossible allocations.
2. `LatentWorldModelConfig.from_config` still does `dict(config)`, silently ignores/wrongly accepts `type`, and converts arbitrary iterables with `tuple(...)`; the outer model loader has the same schema/coercion problem. Require exact built-in dicts, exact fields/type, and exact JSON lists at serialized boundaries without iterating hostile values. Constructor tuples can remain the canonical programmatic form.
3. The new `__post_init__` claims configuration validation but only validates integers. Float fields and `observation_scale` remain unvalidated until `LatentWorldModel(config)`, so malformed configs can still be constructed/serialized. Consolidate the existing exact host-ratio + float32-sink validation into construction and canonical storage (including exact tuple structure/length), preserving each established endpoint.
4. Tests cover only four NumPy types and no max/derived overflow, bool/class-spoof/hostile index or repr hooks, tuple/list subclasses, hostile ratio hooks, canonical scalar storage, JSON round-trip, or exact outer/nested schema. Add the full dtype-derived integer family and failing-test-first regressions for these boundaries without allocating at maximum values.
5. Please revert the unrelated whole-file formatting churn; it obscures this already high-risk diff and is not enforced by the repository Ruff check.

No pinned outputs should be edited. This file is L0/nonpromoting, but any active development source identity should naturally change and be remeasured under its lane policy.

lalalune:
Superseded by merged #735, which preserves your original commit/authorship and completes the config-time scalar, boolean/container, schema, derived-allocation, and generated-parameter contracts with retained regressions.

lalalune:
Superseded by merged #735 plus residual #743. The successor preserves the contributor work already landed in main and completes aggregate persistent resource preflights, strict supported schemas, nested learner compatibility, full NumPy integer-family coverage, and hostile-hook regressions.

PR #714, opened 2026-08-16
fix(router): preserve mapping config compatibility

lalalune:
Follow-up to #704. Restores the documented `Mapping[str, object]` deserialization contract instead of requiring an exact built-in dict, while retaining #704’s strict flat schema, exact string discriminators, resource formulas, and hostile-input safety. Mapping contents are copied once under ordinary-Exception normalization; valid immutable/custom mappings remain supported and hostile hooks cannot escape as arbitrary exceptions or execute repr.

Validation:
- `tests/test_feature_bank_router.py`: 42 passed
- scoped Ruff: clean
- strict source mypy: clean
- diff check: clean

No evidence source or artifact changes.

### DIFF
--- a/alberta_framework/core/feature_bank_router.py
+++ b/alberta_framework/core/feature_bank_router.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 import dataclasses
 import functools
 import operator
+from collections.abc import Mapping
 from typing import Any, SupportsIndex, cast
 
 import jax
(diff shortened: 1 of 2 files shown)


lalalune:
Deep review complete: this correctly restores the documented Mapping boundary while preserving #704’s exact flat schema/discriminator checks. `issubclass(type(config), Mapping)` rejects non-mappings without executing instance spoof hooks; the one explicit `dict(config)` read supports immutable/custom mappings and normalizes ordinary hostile hook failures to ValueError without formatting the object. MappingProxy round-trip and hostile iteration are covered; runtime/resource formulas and evidence surfaces are unchanged. Merge when required checks pass.

PR #720, opened 2026-08-16
fix(multi-head): complete shared validation and allocation contracts

lalalune:
Supersedes #691 and #695 while preserving the latest contributor-authored #691 commit as the first commit in this branch.

The shared `MultiHeadMLPLearner` is the canonical validation boundary. This replacement:

- accepts and canonicalizes built-in integers plus every NumPy signed/unsigned integer dtype family for `n_heads`, every hidden width, and `init(feature_dim)` while rejecting booleans, subclasses, and class/index/repr spoofs without invoking hostile hooks;
- preserves the supported `hidden_sizes=()` linear baseline and verifies its initialized prediction path under JIT;
- preflights exact parameter, directly allocated state-scalar, and state-byte counts in Python before any JAX allocation, both at construction and once `feature_dim` is known;
- retains exact host plus float32-sink validation for every per-head decay and requires actual tuples at constructors;
- preserves historical deserialization compatibility: exact JSON lists and exact in-process tuples are accepted for `hidden_sizes` and `per_head_gamma_lamda`, while strings, generators, subclasses, and spoofed iterables are rejected without iteration;
- deliberately keeps the permissive outer mapping/type-marker behavior; it does not add an unrelated exact-dict/full-field schema requirement;
- makes `CBPMultiHeadMLPLearner` reuse the shared canonical values and sequence guard, eliminating the need for #695's duplicate wrapper validation.

Verification on final head `58527c5`:

- focused MultiHead + CBP suites: 144 passed in 86.82s
- validation-only regressions: 16 passed
- adjacent config/Horde/SARSA/forward-view suites: 131 passed
- `ruff check .`: clean
- strict `mypy`: clean (168 source files)
- `git diff --check`: clean

Neither changed source file is registered in the five-claim evidence manifest, and no `outputs/**` artifacts are modified. Historical development snapshots that inventory these files remain historical; this engineering change creates no evidence or promotion.

### DIFF
--- a/alberta_framework/core/continual_backprop.py
+++ b/alberta_framework/core/continual_backprop.py
@@ -83,6 +83,7 @@ from alberta_framework.core.multi_head_learner import (
     MultiHeadMLPLearner,
     MultiHeadMLPState,
     MultiHeadMLPUpdateResult,
+    _require_config_sequence,
 )
 from alberta_framework.core.normalizers import Normalizer
 from alberta_framework.core.optimizers import Bounder
@@ -771,8 +772,6 @@ class CBPMultiHeadMLPLearner:
             utility_decay: EMA decay for the underlying MLP's native
                 hidden-unit utility diagnostics.
         """
-        self._n_heads = n_heads
-        self._hidden_sizes = hidden_sizes
         self._cbp_config = cbp_config or ContinualBackpropConfig()
         self._sparsity = sparsity
         self._leaky_relu_slope = leaky_relu_slope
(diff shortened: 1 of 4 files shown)


lalalune:
Closing as fully superseded by merged #715. Current main already contains the preserved contributor validation plus the shared allocation repair and CBP delegation (`77351da`, `263978b`, `4c0d1f1`); this branch now conflicts only because the equivalent content landed through the consolidated integration. No behavioral residual remains.

lalalune:
Superseded by merged #727, which implements the same aggregate direct-state parameter/scalar/byte preflights on top of the strict shared MultiHead/CBP boundary from #715. I rebased this branch to current main; no independent diff remains.

PR #712, opened 2026-08-16
fix(recurrent-ensemble): harden ensemble configuration integer validation and canonicalization

Kayvan-Zahiri:
## Summary
- Hardens `RecurrentLatentWorldModelEnsembleConfig` integer fields (`observation_dim`, `n_actions`, `latent_dim`, `ensemble_size`, `max_updates`, `uncertainty_warmup_steps`) to reject booleans, non-integer floats, and out-of-bounds dimensions while accepting and canonicalizing the full NumPy integer family.
- Canonicalizes all integer attributes to Python `int` on construction.

## Verification
- `PYTHONPATH=. pytest tests/test_recurrent_latent_world_model_ensemble.py`: 56 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/recurrent_latent_world_model_ensemble.py`: clean
- No registered evidence artifacts modified.

PR #722, opened 2026-08-16
fix(recurrent-ensemble): validate exact integer configuration

lalalune:
## Summary

Successor to #712 retaining the contributor commit. It admits and canonicalizes the complete supported NumPy integer scalar family for all six integer configuration fields while rejecting booleans, floats, subclasses/spoofs, and values outside signed-int32 counter domains. Existing exact persistent-state resource accounting continues to preflight all derived allocation products under the 256 MiB cap before initialization.

The follow-up coverage exercises every dtype-derived signed/unsigned NumPy integer type across every field, hostile subclass/spoof behavior without repr execution, adjacent counter overflow, and canonical JSON round-trip.

## Verification

- `tests/test_recurrent_latent_world_model_ensemble.py`: 119 passed
- scoped Ruff: clean
- strict source mypy: clean
- diff check: clean

L0/nonpromoting source only; no pinned evidence outputs changed.

### DIFF
--- a/alberta_framework/core/recurrent_latent_world_model_ensemble.py
+++ b/alberta_framework/core/recurrent_latent_world_model_ensemble.py
@@ -46,10 +46,11 @@ import functools
 import hashlib
 import json
 import math
+import operator
 from collections.abc import Mapping
 from numbers import Real
 from pathlib import Path
-from typing import Any, cast
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #735. I compared the live files directly: main now has the same full builtin/NumPy exact integer canonicalization across all six fields and round-trip/resource tests, plus stronger exact-dict nested/outer schema gates, a shared nonnegative warmup validator, and the same pre-allocation 256 MiB derived-state cap. #722 would regress those schema/helper improvements and has no unique runtime residual; its additional hostile/counter assertions exercise the same exact-type gates already covered. Closing rather than merging a stale duplicate.

PR #731, opened 2026-08-16
fix(continual-multiagent): harden benchmark configuration integer validation

Kayvan-Zahiri:
## Summary
- Hardens `ContinualMultiAgentConfig` integer dimensions (`phase_steps`, `nuisance_dim`, `probe_horizon`, `probe_tail_steps`, `recovery_window`, `bootstrap_resamples`, `bootstrap_seed`) and `AcceptanceThresholds` integer fields (`minimum_seed_count`, `evidence_seed_start`) with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.

## Verification
- `PYTHONPATH=. pytest tests/test_continual_multiagent_benchmark.py tests/test_recurring_multiagent_stream.py`: 54 passed
- `ruff check .`: clean
- `mypy alberta_framework/evaluation/continual_multiagent.py`: clean

### DIFF
--- a/alberta_framework/evaluation/continual_multiagent.py
+++ b/alberta_framework/evaluation/continual_multiagent.py
@@ -32,10 +32,11 @@ solution.  All updates are predict-act-observe-update and use no replay.
 
 from __future__ import annotations
 
+import operator
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass
 from time import perf_counter, perf_counter_ns
-from typing import Literal
+from typing import Literal, SupportsIndex, cast
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
The exact-type/operator.index direction is correct, but this registered evaluation source needs additional fail-closed work before merge. Focused tests do not cover the actual derived execution domains:

1. `total_steps = 3 * phase_steps`, `_probe_state` stores `context * phase_steps` in int32, and phase boundaries include `2 * phase_steps`. An individually int32 `phase_steps` can therefore overflow telemetry. Preflight the exact derived horizon (at minimum `phase_steps <= INT32_MAX // 3`) before JAX/array work.
2. `nuisance_dim` controls persistent and per-step arrays through RecurringTwoAgentWorld, but neither this config nor that constructor currently bounds the derived float32 scalar/byte budget. Likewise `bootstrap_resamples` controls bootstrap work/allocation with no practical or derived resource ceiling. Add explicit exact resource formulas/caps before allocation, not only individual int32 admission.
3. `evidence_seed_start + minimum_seed_count` is used to construct the exact held-out schedule but is not preflighted. Validate the exclusive end and every downstream seed/JAX-key domain so the range cannot cross its supported sink. `bootstrap_seed` deliberately uses uint32, so name/use a uint32 seed validator rather than `_require_int32(... maximum=2**32-1)`.
4. Tests cover only a few NumPy types. Add all dtype-derived signed/unsigned NumPy integer families for every field, hostile int subclasses/class spoofs/repr hooks, bools for AcceptanceThresholds, adjacent derived-overflow cases, and canonical serialized/artifact round-trips without running the expensive benchmark at maximum values. Preserve the frozen default protocol exactly.
5. The source/test reformatting is unrelated and should be reverted to keep the scientific diff auditable.

Evidence note: `alberta_framework/evaluation/continual_multiagent.py` is a registered source for the accepted recurring_multiagent_coadaptation artifact. This PR does not modify outputs (good), but it does invalidate that persisted artifact by source-hash drift until the frozen protocol is rerun and strictly validated. Do not edit or rehash the pinned artifact to silence that expected fail-closed result.

lalalune:
The exact integer intake is directionally correct, but individual bounds do not protect the three-phase horizon, float64 result buffers, recurring-world state, bootstrap workspace, uint32 seed schedule, or float32 configuration sinks. Superseded by contributor-preserving #753 with exact pre-allocation formulas and adversarial/adjacent-boundary coverage. The pinned accepted artifact remains untouched and should fail closed on source drift.

PR #733, opened 2026-08-16
fix(ftl-decision-fidelity): harden evaluation configuration integer validation

Kayvan-Zahiri:
## Summary
- Hardens `DecisionFidelityConfig` integer fields (`phase_steps`, `horizon`, `probes_per_domain`, `projection_dim`, `bins`, `bootstrap_resamples`, `bootstrap_seed`) with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.

## Verification
- `PYTHONPATH=. pytest tests/test_ftl_decision_fidelity.py tests/test_ftl_decision_cli.py tests/test_ftl_decision_artifact.py`: 48 passed
- `ruff check .`: clean
- `mypy alberta_framework/evaluation/ftl_decision_fidelity.py`: clean

### DIFF
--- a/alberta_framework/evaluation/ftl_decision_fidelity.py
+++ b/alberta_framework/evaluation/ftl_decision_fidelity.py
@@ -30,9 +30,10 @@ completion of the Alberta Plan.
 
 from __future__ import annotations
 
+import operator
 from collections.abc import Sequence
 from dataclasses import dataclass
-from typing import NamedTuple
+from typing import NamedTuple, SupportsIndex, cast
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
The exact integer gate is a good start, but this evaluation configuration still needs its derived and scalar domains closed before merge:

1. Preflight exact products/sums used by the evaluation before construction/allocation: phase schedule length, domains × probes × horizon rollout work, `projection_dim * bins` feature/index width, projection/FTL parameter and persistent state scalar/byte counts, and bootstrap_resamples × sample-count draw allocation. Individual int32 fields do not prevent product overflow or OOM. Reuse the live SparseFTL resource contract where possible rather than duplicating a weaker formula.
2. `bootstrap_seed` is uint32, not int32; use the shared/full uint32 seed validator. Cap bootstrap draws using an explicit resource contract and validate before `generator.integers(size=(resamples, values.size))`.
3. The remaining configuration surface is still coercive: require exact tuple structure for menu_amplitudes, normalize hostile container/ratio/conversion exceptions, and canonicalize ridge/prediction_clip/state_bound/action_cost/confidence_level through exact host-ratio + float32 sinks while preserving frozen endpoints/default values. Do not accept bools or tuple subclasses.
4. Add every dtype-derived NumPy integer family across every field; hostile int/Real subclasses, class spoofs, repr/index/ratio hooks; adjacent derived-overflow and no-allocation resource tests; canonical serialization/artifact reconstruction; and tests proving the frozen config bytes/default evaluation semantics remain unchanged.

This is registered scientific-evidence code for the historical FTL decision-fidelity chain. Do not edit pinned artifacts or hashes. Source changes must leave the frozen artifact fail-closed invalid until the exact historical/frozen protocol is rerun and strict validation succeeds; the PR description should disclose that impact.

lalalune:
Superseded by #757, which preserves this contributor commit and completes the requested derived resource, uint32 seed, exact float/menu, hostile-hook, canonical serialization, and fail-closed evidence coverage.

PR #724, opened 2026-08-16
fix(average-reward): harden actor-critic and SARSA configuration integer validation

Kayvan-Zahiri:
## Summary
- Hardens `AverageRewardHordeActorCriticConfig` integer dimensions (`n_actions`, `hidden_sizes`) and `DifferentialSARSAConfig` integer fields (`n_actions`, `epsilon_decay_steps`) using `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Canonicalizes `n_actions`, `hidden_sizes`, and `epsilon_decay_steps` to Python `int` at construction time.

## Verification
- `PYTHONPATH=. pytest tests/test_average_reward.py`: 32 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/average_reward.py`: clean
- No registered evidence artifacts modified.

PR #734, opened 2026-08-16
fix(average-reward): preflight aggregate actor-critic state

lalalune:
Supersedes #724 while preserving its contributor commit. The original integer hardening was directionally correct but left float32 sink values spoofable/noncanonical, accepted spoofed booleans, blindly iterated serialized hidden sizes, did not validate init dimensions, and allowed derived actor/SARSA state products to cross the signed-int32 allocation domain before JAX allocation. This successor adds exact built-in/full NumPy integer-family canonicalization, exact list-or-tuple deserialization, shared exact-host plus float32 scalar validation with documented endpoints, exact bool handling, and allocation-free scalar/byte preflights for both public states. Tests cover every NumPy integer family, hostile subclasses/containers, float32 overflow and rounded endpoints, canonical storage, and adjacent derived resource bounds. Verification: 53 focused tests passed; scoped Ruff clean; strict targeted mypy clean; diff-check clean. No registered evidence source or outputs are changed.

### DIFF
--- a/alberta_framework/core/average_reward.py
+++ b/alberta_framework/core/average_reward.py
@@ -24,7 +24,6 @@ from __future__ import annotations
 
 import dataclasses
 import functools
-import math
 import operator
 import time
 from collections.abc import Mapping
@@ -38,6 +37,7 @@ import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int, UInt
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 from alberta_framework.core.learners import _update_from_gradient_with_diagnostics
 from alberta_framework.core.multi_head_learner import MultiHeadMLPLearner, MultiHeadMLPState
 from alberta_framework.core.optimizers import Autostep, AutostepParamState, optimizer_from_config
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking residuals found in follow-up review; auto-merge disabled pending repair. (1) `AverageRewardHordeActorCriticConfig.from_config` and `DifferentialSARSAConfig.from_config` still accept dict subclasses, wrong/missing type discriminators, and unknown keys. Preserve only the explicitly documented legacy omission of SARSA `use_bias`; require exact built-in dicts and exact remaining schemas/types without hostile iteration/repr. (2) `_preflight_actor_state` is not an aggregate public-state preflight: it counts actor leaves but omits the nested `AverageRewardHordeState` / MultiHead critic resources. Two individually legal components can therefore exceed the claimed aggregate scalar/byte cap. Preflight the exact critic + actor + counters/telemetry total before either allocation and add adjacent aggregate-bound tests. The current scalar/int/init changes remain valuable; fix these two residuals, reverify focused actor/average-reward/shared MultiHead tests, then re-enable merge.

lalalune:
Superseded by merged #747, which preserves both #724's contributor commit and this maintainer repair, then adds exact serialized schema/container gates in the combined audit.

lalalune:
Rebased and narrowed after merged #747. #747 fully subsumed the scalar, integer, schema, and individual-state work. The remaining diff fixes the blocker identified in review: `_preflight_actor_state` now includes the nested one-head MultiHead critic state in the public aggregate scalar/byte count instead of counting actor leaves alone. The exact adjacent byte boundary test was updated. Verification on #747 main: 54/54 focused tests, Ruff, strict mypy, diff-check, clean merge-tree.

PR #707, opened 2026-08-16
fix(actor-critic): harden configuration scalar and integer bounds across actor-critic configs

Kayvan-Zahiri:
## Summary
- Hardens `HordeActorCriticConfig`, `QHordeActorCriticConfig`, `NonlinearHordeActorCriticConfig`, and `NonlinearQHordeActorCriticConfig` integer fields (`n_actions`, `value_head_index`, `hidden_sizes`) against booleans, non-integer floats, and out-of-bound dimensions while accepting and canonicalizing the full NumPy integer family.
- Replaces unvalidated or raw float inequality checks on learning and trace parameters (`gamma`, `actor_step_size`, `actor_lamda`, `temperature`, `actor_sparsity`, `leaky_relu_slope`, `actor_epsilon`, `actor_td_error_normalizer_decay`, `actor_td_error_clip`, `actor_gradient_clip_norm`) with `validated_float32_scalar` to enforce float32 bounds and eliminate binary32 boundary overflow.

## Verification
- `PYTHONPATH=. pytest tests/test_horde_actor_critic.py`: 59 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/horde_actor_critic.py`: clean
- No registered evidence artifacts modified.

PR #736, opened 2026-08-16
fix(actor-critic): complete config and resource contracts

lalalune:
Supersedes #707 while preserving Kayvan-Zahiri original contributor commit and authorship.\n\nThe contributor correctly added float32 scalar and trusted integer validation, but the original diff reformatted most of both large files and left live gaps in derived actor allocations, persistent resource bounds, serialized schemas, exact enum/bool types, and runtime feature dimensions. This successor retains the useful change and adds:\n\n- every Python/NumPy integer dtype family with hostile subclass rejection and canonical ints;\n- float32-safe scalar endpoints across all four actor-critic configs;\n- exact tuple/list config representations, exact key sets, markers, bools, and enum strings;\n- hidden-layer/head product preflights and allocation-free actor parameter/state/byte bounds before all four init paths;\n- canonical config and agent roundtrips without touching JIT update semantics;\n- focused hostile, boundary, dtype-family, schema, and resource tests.\n\nVerification after rebase onto current main: 129 actor-critic tests passed (plus 83 focused config/resource tests post-rebase), scoped Ruff, strict source mypy, diff-check, and merge-tree all pass. No evidence artifacts changed.

### DIFF
--- a/alberta_framework/core/horde_actor_critic.py
+++ b/alberta_framework/core/horde_actor_critic.py
@@ -33,15 +33,18 @@ from __future__ import annotations
 
 import dataclasses
 import functools
-from typing import Any, cast
+import operator
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 from alberta_framework.core.horde import HordeLearner, HordeUpdateResult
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by #742 after exact comparison. #742 retains the same exact integer-family/float32/hostile-type and exact actor parameter/state/byte preflights, is rebased on newer main, and preserves the requested historical hidden_sizes list/tuple decode compatibility. This branch additionally requires every config field, rejects tuple-backed in-process deserialization, and tightens all agent payload schemas—broader compatibility breaks not required for this repair. #742 also avoids preserving those unrelated schema changes while retaining Kayvan-Zahiri attribution.

PR #742, opened 2026-08-16
fix(actor-critic): validate configuration and actor resources

lalalune:
Supersedes #707 with a clean, current-main repair while retaining the contributor's scalar/int validation intent.

- validates all four config families at their float32 and signed-int32 sinks
- preserves legal gamma/lambda/sparsity endpoints and empty nonlinear actors
- accepts exact historical list/tuple hidden-size representations without trusting arbitrary iterables
- rejects hostile enum/bool/numeric values with static ValueErrors
- preflights exact direct actor parameter/state/byte allocations before JAX allocation
- covers all NumPy integer families, hostile hooks, serialization compatibility, derived bounds, and JIT valid paths

Validation: 75 actor-critic tests; 192 average-reward/pipeline tests; Ruff; strict mypy.

### DIFF
--- a/alberta_framework/core/horde_actor_critic.py
+++ b/alberta_framework/core/horde_actor_critic.py
@@ -33,15 +33,18 @@ from __future__ import annotations
 
 import dataclasses
 import functools
-from typing import Any, cast
+import operator
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int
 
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #751. #751 preserves the original #707 contributor commit and covers all four families with exact scalar/bool/enum gates, exact complete dict and JSON-list serialization boundaries, and pre-allocation parameter/state/byte budgets including Autostep leaves. Its focused verification passed 68 tests, Ruff, strict mypy, and diff hygiene. The broader list-or-tuple serialized acceptance here is intentionally not retained because `to_config` emits a JSON list and the repaired boundary is exact.

PR #743, opened 2026-08-16
fix(latent-world-model): close residual resource and mapping gaps

lalalune:
Residual follow-up to merged #735/#740. The contributor's #711 work and attribution are already preserved in main.\n\nDeep comparison found two requested contracts still missing:\n- actual Mapping implementations were rejected at config/model boundaries even though mapping compatibility is established; this accepts real Mapping runtime types without isinstance/__class__ spoofing while retaining exact JSON list/tuple boundaries and exact type/schema fields (with only the four documented legacy encoder defaults optional);\n- encoder and MultiHead allocations were preflighted separately, but their combined persistent JAX scalar/byte total could exceed signed int32; this adds an exact aggregate preflight before allocation.\n\nTests cover MappingProxy roundtrips, wrong/missing schema, legacy compatibility, and allocation-free aggregate overflow.\n\nValidation on latest main: 54 focused tests passed; scoped Ruff clean; strict mypy clean; diff check clean. No outputs or evidence artifacts changed.

### DIFF
--- a/alberta_framework/core/latent_world_model.py
+++ b/alberta_framework/core/latent_world_model.py
@@ -293,6 +293,8 @@ class LatentWorldModelConfig:
         """Reconstruct from :meth:`to_config` output."""
         if not issubclass(type(config), Mapping):
             raise ValueError("config must be an actual mapping")
+        if not all(type(key) is str for key in config):
+            raise ValueError("config fields do not match the serialized schema")
         payload = dict(config)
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged audited successor #758, which preserves this commit and narrows the resource wording to the exact directly-owned state contract while adding hostile mapping-spoof coverage.

PR #726, opened 2026-08-16
fix(recurring-ipmnist): harden evaluation dataclass integer validation and canonicalization

Kayvan-Zahiri:
## Summary
- Hardens `_nonnegative_int` and `_positive_int` integer validation across `SentinelProbeBinding`, `RecurringIPMNISTPhase`, `RecurringIPMNISTProtocol`, and `SentinelProbeSnapshot` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Enforces strict signed int32 bounds and canonical Python `int` conversion across all evaluation dataclass attributes.

## Verification
- `PYTHONPATH=. pytest tests/test_recurring_ipmnist_retention.py`: 14 passed
- `ruff check .`: clean
- `mypy alberta_framework/evaluation/recurring_ipmnist_retention.py`: clean
- No registered evidence artifacts modified.

PR #744, opened 2026-08-16
fix(recurring-ipmnist): bound aggregate evaluation workloads

lalalune:
Supersedes #726 while preserving Kayvan-Zahiri original contributor commit and authorship.\n\nThe original integer admission fix was directionally correct, but `SentinelProbeSnapshot` validated NumPy integer fields without storing their canonical Python ints, and individually valid phases/bindings could overflow derived trace and probe workloads. This successor keeps the exact Python/full-NumPy integer family gate and adds:\n\n- canonical storage for every affected binding, phase, protocol, and snapshot integer;\n- hostile subclass/hook rejection without untrusted repr;\n- `start_step + length` signed-int32 preflight;\n- aggregate two-channel trace scalar and required sentinel-evaluation workload bounds;\n- correctness-vector logical workload bounds;\n- dtype-family, spoof, canonicalization, and allocation-free overflow tests.\n\nVerification: 24 focused tests passed, scoped Ruff, strict source mypy, diff-check, and merge-tree passed. `recurring_ipmnist_retention.py` is not in the five-claim evidence manifest, and no `outputs/` artifact or campaign result was touched; this remains development-only/nonpromoting.

### DIFF
--- a/alberta_framework/evaluation/recurring_ipmnist_retention.py
+++ b/alberta_framework/evaluation/recurring_ipmnist_retention.py
@@ -142,6 +142,11 @@ def _positive_int(value: object, *, name: str, maximum: int = _INT32_MAX) -> int
     return canonical
 
 
+def _require_int32_workload(name: str, value: int) -> None:
+    if not 1 <= value <= _INT32_MAX:
+        raise ValueError(f"derived {name} must fit in signed int32")
+
+
 def _unit_float(value: object, *, name: str, binary: bool = False) -> float:
     if type(value) is not float or not math.isfinite(value) or not 0.0 <= value <= 1.0:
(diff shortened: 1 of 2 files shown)


lalalune:
Rebased and narrowed after merged #747. #747 correctly subsumed all exact integer/schema/phase-stop/snapshot-canonicalization work. The remaining 45-line diff is only aggregate logical workload safety that #747 does not contain: `2 * trace_length`, `2 * actual trace length`, total required sentinel evaluations across five probes, and correctness-vector case count must each fit signed int32. Verification on #747 main: 16 focused tests, Ruff, strict mypy, diff-check.

lalalune:
Superseded by merged #750 after #747. #750 preserves this commit's authorship and unique aggregate trace/sentinel workload checks while resolving it on current main.

PR #745, opened 2026-08-16
fix(reward-model): harden RLS reward model configuration scalar and integer validation

Kayvan-Zahiri:
## Summary
- Adds `__post_init__` to `RLSRewardModelConfig` with `_require_int32` on `feature_dim` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Validates float scalar fields (`forgetting`, `ridge`, `error_decay`) with `validated_float32_scalar` and canonicalizes attributes at construction time.

## Verification
- `PYTHONPATH=. pytest tests/test_reward_model.py`: 46 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/reward_model.py`: clean

### DIFF
--- a/alberta_framework/core/reward_model.py
+++ b/alberta_framework/core/reward_model.py
@@ -14,16 +14,44 @@ trustworthy — the "selective" part of selective model-based updates.
 from __future__ import annotations
 
 import functools
-from dataclasses import dataclass, replace
-from typing import Any
+import operator
+from dataclasses import dataclass
+from typing import Any, SupportsIndex, cast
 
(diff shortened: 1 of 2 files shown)


lalalune:
The scalar validators are directionally correct, but the quadratic covariance allocation remains unbounded, the constructor uses a spoofable isinstance gate, and vector reward aliases can alter committed state shapes. Superseded by contributor-preserving #760 with exact state-byte and scalar-reward contracts plus adversarial/JIT coverage.

PR #748, opened 2026-08-16
fix(sigreg): harden configuration integer validation and canonicalization

Kayvan-Zahiri:
## Summary
- Hardens `SIGRegConfig` integer dimension `n_projections` with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.

## Verification
- `PYTHONPATH=. pytest tests/test_sigreg.py`: 32 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/sigreg.py`: clean

### DIFF
--- a/alberta_framework/core/sigreg.py
+++ b/alberta_framework/core/sigreg.py
@@ -26,7 +26,8 @@ from __future__ import annotations
 import dataclasses
 import math
 import numbers
-from typing import Any, cast
+import operator
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by contributor-preserving #756, which retains the integer fix while completing float32 operation bounds, hostile-safe schema handling, latent-dimension admission, derived direction/projection/pairwise resource preflights, shape contracts, and finite extreme-value diagnostics.

lalalune:
Correction: the SIGReg successor is #762 (not #756).

lalalune:
The integer gate is correct but leaves derived direction/projection/pairwise resources, latent_dim, operation-safe scalar bounds, and public shape contracts incomplete. Superseded by contributor-preserving #770, which repairs those gaps while retaining Mapping and partial-config compatibility.

PR #752, opened 2026-08-16
fix(stacked-horde): harden configuration integer validation and canonicalization

Kayvan-Zahiri:
## Summary
- Hardens `StackedHordeConfig` integer fields `n_demons`, `feature_dim`, and `cumulant_indices` entries with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.

## Verification
- `PYTHONPATH=. pytest tests/test_stacked_horde.py`: 76 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/stacked_horde.py`: clean

### DIFF
--- a/alberta_framework/core/stacked_horde.py
+++ b/alberta_framework/core/stacked_horde.py
@@ -37,10 +37,11 @@ the NaN-masking convention of the loop-based hordes.
 """
 
 import math
+import operator
 from dataclasses import dataclass
 from fractions import Fraction
 from numbers import Real
-from typing import Any, cast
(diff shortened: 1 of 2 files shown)


lalalune:
The integer fix is useful but does not validate gamma/lambda float32 sinks, exact sequence boundaries, aggregate state/config/result resources, or nexting cross-products before materialization. Superseded by contributor-preserving #767 with exact scalar/sequence/resource contracts and adversarial endpoint tests.

PR #753, opened 2026-08-16
fix(continual-multiagent): complete scalar and resource contracts

lalalune:
Supersedes #731 while preserving its contributor commit. The original exact integer admission is useful but did not guard the derived A-B-A horizon, result buffers, recurring-world state, bootstrap workspace, or held-out seed schedule; its float32-consumed configuration values also remained spoofable/noncanonical. This successor preserves the frozen default protocol bytes while adding full Python/NumPy integer canonicalization, uint32 seed validation, exact host+float32 scalar domains, signed-int32 scalar/byte formulas before allocation, and exact seed-schedule endpoints. Bootstrap accounting includes the int64 index matrix, indexed float64 temporary, means, and input vector. Verification: 97 focused tests passed (2 dtype-domain skips where 8-bit types cannot represent the documented minimum of 1000), plus the corrected adjacent workspace test; Ruff, strict targeted mypy, and diff-check are clean. This edits the registered recurring_multiagent_coadaptation source, so the pinned accepted artifact must remain untouched and is expected to fail closed on source-hash drift until a separately authorized frozen rerun; no outputs are modified or promoted.

### DIFF
--- a/alberta_framework/evaluation/continual_multiagent.py
+++ b/alberta_framework/evaluation/continual_multiagent.py
@@ -32,10 +32,11 @@ solution.  All updates are predict-act-observe-update and use no replay.
 
 from __future__ import annotations
 
+import operator
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass
 from time import perf_counter, perf_counter_ns
-from typing import Literal
+from typing import Literal, SupportsIndex, cast
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


lalalune:
Closing in favor of the stronger concurrent successor #759. It preserves the same contributor fix and exact bootstrap accounting, and additionally bounds observation arrays plus retained per-seed payloads, avoids materializing an attacker-sized expected seed tuple, keeps the held-out schedule in its actual signed JAX-key domain, and isolates the adversarial panel in a dedicated test module. #759 subsumes this branch.

lalalune:
Superseded by stronger #759. #759 preserves the same contributor commit and exact scalar/int work while additionally preflighting recurring-world observation arrays, retained per-seed result payloads, and the complete paired-bootstrap peak workspace; it also avoids materializing attacker-sized expected seed tuples and has the broader 200-test panel. No unique residual remains here.

PR #754, opened 2026-08-16
fix(average-reward): preflight aggregate actor-critic state

lalalune:
Follow-up residual after merged #747 (and its now-closed precursor #734). #747 completed the scalar and exact-schema contracts, but the public average-reward actor-critic preflight still counted actor leaves without its nested one-head MultiHead critic. This focused diff includes the critic trunk/head parameters, traces, utility, reward-rate, and counters in the same aggregate scalar/byte cap and updates the exact adjacent boundary test.\n\nVerification on #747 main: 54/54 average-reward tests, scoped Ruff, strict source mypy, diff-check, and merge-tree passed. No outputs/evidence artifacts changed.

### DIFF
--- a/alberta_framework/core/average_reward.py
+++ b/alberta_framework/core/average_reward.py
@@ -93,10 +93,23 @@ def _require_state_resources(name: str, *, scalars: int, nbytes: int) -> None:
         raise ValueError(f"{name} state bytes must fit signed int32")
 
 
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged audited successor #761, which preserves this commit and adds the omitted critic LMS optimizer-state scalars plus an exact adjacent boundary.

PR #755, opened 2026-08-16
fix(recurring-ipmnist): bound aggregate evaluation workloads

lalalune:
Follow-up residual after merged #747 (and closed #744). #747 completed exact integer/schema gates, phase stop bounds, and snapshot canonicalization. This focused diff adds the remaining aggregate logical workload limits: two-channel protocol/trace scalar counts, the total sentinel cases evaluated across all five required probes, and each correctness vector must fit signed int32. Tests use allocation-free oversized metadata.\n\nVerification on #747 main: 16/16 focused tests, Ruff, strict source mypy, diff-check passed. This source remains unregistered development-only evaluation machinery; no outputs/evidence artifacts changed.

### DIFF
--- a/alberta_framework/evaluation/recurring_ipmnist_retention.py
+++ b/alberta_framework/evaluation/recurring_ipmnist_retention.py
@@ -142,6 +142,11 @@ def _positive_int(value: object, *, name: str, maximum: int = _INT32_MAX) -> int
     return canonical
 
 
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #750, which preserves this exact workload implementation and adds the same allocation-free aggregate trace regression on current main.

PR #757, opened 2026-08-16
fix(ftl-decision-fidelity): complete exact resource contracts

lalalune:
Supersedes #733 while preserving kzahiri1's original commit and attribution.

Closes the remaining fail-closed configuration contract: exact/canonical integer and float32 sinks, exact tuple menu validation, uint32 bootstrap seeds, derived phase/rollout/SparseFTL/bootstrap allocation preflights, operation-safe reward bounds, and exact custom seed validation. Tests cover all NumPy integer families, hostile hooks/spoofs/subclasses, adjacent resource failures before allocation, canonical frozen defaults, artifact reconstruction, and the existing evaluation semantics.

Evidence impact: this intentionally changes a registered source. No pinned artifact, hash, or outputs were edited. The historical FTL evidence remains fail-closed invalid until the exact frozen protocol is rerun and its strict validator accepts a new artifact.

Validation: 124 passed, 7 skipped across FTL fidelity/artifact/CLI tests; Ruff and strict mypy pass; diff check clean.

### DIFF
--- a/alberta_framework/evaluation/ftl_decision_fidelity.py
+++ b/alberta_framework/evaluation/ftl_decision_fidelity.py
@@ -30,6 +30,7 @@ completion of the Alberta Plan.
 
 from __future__ import annotations
 
+import math
 import operator
 from collections.abc import Sequence
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


lalalune:
Independent follow-up audit is now incorporated on top of the return-domain and exact-seed-container fix. Menu count-derived scalar/work/byte checks execute before any element ratio hook or canonicalization list/set allocation; a hostile menu regression proves the ordering. Combined adversarial panel: 89 passed, 7 representational-domain skips; Ruff and diff hygiene pass.

PR #759, opened 2026-08-16
fix(continual-multiagent): complete scalar and resource contracts

lalalune:
Supersedes #731 while preserving its contributor commit and exact integer-type hardening.

This successor completes the execution-domain audit: the A-B-A phase horizon, recurring-world persistent/observation arrays, per-condition work arrays, retained per-seed result payload, and paired-bootstrap peak arrays are all checked before allocation. Bootstrap seeds use the deliberate uint32 domain; held-out seed schedule arithmetic remains inside the signed JAX-key domain and comparison no longer materializes an attacker-sized expected tuple. Exact list/tuple/range and scalar gates reject hostile subclasses, class spoofing, booleans, and conversion/repr hooks while accepting and canonicalizing every NumPy integer family that intersects each field's valid range. Unrelated test reformatting from the source branch was removed.

This edits the registered `recurring_multiagent_coadaptation` evaluation source and therefore intentionally makes the pinned accepted artifact report source drift until its frozen protocol is separately rerun and strictly validated. No outputs were run, modified, or rehashed.

Verification:
- `pytest tests/test_continual_multiagent_validation.py tests/test_continual_multiagent_benchmark.py tests/test_recurring_multiagent_stream.py -q` (200 passed, 2 representational-domain skips)
- post-final-patch evaluation panel: 179 passed, 2 skips
- scoped Ruff clean
- strict source mypy clean
- diff hygiene clean

### DIFF
--- a/alberta_framework/evaluation/continual_multiagent.py
+++ b/alberta_framework/evaluation/continual_multiagent.py
@@ -45,6 +45,7 @@ import numpy as np
 from jax import Array
 from numpy.typing import NDArray
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 from alberta_framework.streams.recurring_multiagent import (
     AVOID_CONTEXT,
     AVOID_CONTEXT_INDEX,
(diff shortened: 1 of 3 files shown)


lalalune:
Final deduplication audit recovered one unique correctness delta from closed #753: all float32-consumed benchmark configuration fields now use the established exact host-ratio/binary32 gate, and every acceptance-threshold scalar is finite/canonical before artifact serialization. Added NumPy canonicalization and hostile Real/class/bool/nonfinite coverage. The combined #759 panel now passes 213 tests with 2 representational-domain skips; Ruff and strict mypy pass. #753/#731 retain no unique delta.

PR #760, opened 2026-08-16
fix(reward-model): complete scalar and state contracts

lalalune:
Supersedes #745 while preserving its contributor commit. The original config hardening correctly adopts shared float32 validation, but an individually legal feature_dim could still request a quadratic covariance larger than the signed-int32 state/byte domain; the model constructor also retained a spoofable isinstance boundary, and update silently accepted vector reward aliases that could alter state shapes. This successor adds exact quadratic state preflight before JAX allocation, an exact config boundary, and scalar reward guards under eager and outer JIT. Tests cover all dtype-derived NumPy integer families, hostile subclasses/spoofs without repr execution, adjacent allocation-free covariance bounds, and row/vector reward aliases. Verification: 60 tests passed; scoped Ruff and strict mypy clean; diff-check clean. No registered evidence source or outputs are changed.

### DIFF
--- a/alberta_framework/core/reward_model.py
+++ b/alberta_framework/core/reward_model.py
@@ -54,6 +54,14 @@ def _require_int32(name: str, value: object, *, minimum: int, maximum: int = _IN
     return canonical
 
 
+def _preflight_state_resources(feature_dim: int) -> None:
+    state_scalars = feature_dim * feature_dim + feature_dim + 2
+    if state_scalars > _INT32_MAX:
+        raise ValueError("reward-model state scalars must fit signed int32")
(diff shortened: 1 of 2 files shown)


goldcaroline40-ops (approved):

lalalune:
Superseded by merged #769, which preserves both commits from this branch, including scalar reward-shape validation and the quadratic covariance preflight. The integrated repair additionally closes exact hostile-hook-safe scalar and outer/nested serialized schema boundaries. Combined verification: 212 focused tests, Ruff, strict mypy, diff hygiene.

PR #762, opened 2026-08-16
fix(sigreg): complete scalar and representation contracts

lalalune:
Supersedes #748 while preserving Kayvan-Zahiri original contributor commit and authorship.\n\nThe original NumPy integer-family admission was correct but left the float32-consumed scalars, exact config schema, runtime `latent_dim`, derived representation/workspace allocations, representation shapes, and extreme-value diagnostics incomplete. This successor adds:\n\n- exact trusted Python/full-NumPy integer admission for projection and latent dimensions;\n- one-read hostile-safe exact-host + float32 validation with operation-safe kernel-width square bounds;\n- exact built-in dict, key set, and type marker roundtrip;\n- signed-int32 byte preflights for generated directions, projections, and aggregate Epps-Pulley pairwise intermediates before allocation;\n- explicit nonempty/matching representation shapes;\n- overflow-safe latent/projected standard deviations while preserving JIT/vmap runtime validation semantics.\n\nVerification after rebase: 46/46 SIGReg tests, targeted resource regression, scoped Ruff, strict source mypy, diff-check, clean merge-tree. No evidence artifacts changed.

### DIFF
--- a/alberta_framework/core/sigreg.py
+++ b/alberta_framework/core/sigreg.py
@@ -26,7 +26,8 @@ from __future__ import annotations
 import dataclasses
 import math
 import numbers
-from typing import Any, cast
+import operator
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #769, which preserves both commits from this branch, including representation/workspace allocation and stable-diagnostic repairs. The integrated repair additionally rejects exact-host scalar subclasses without executing hooks and fails at config construction when even the minimum-width direction allocation is impossible. Combined verification: 212 focused tests, Ruff, strict mypy, diff hygiene.

PR #764, opened 2026-08-16
feat(prereg): add fail-closed issue184 launch path

lalalune:
Closes the remaining engineering gap for #184 without launching either scientific run.

This adds the frozen #184 protocol, exact outcome classifier, workflow dispatch and artifact recovery path, immutable amendment/authorization checks, and a durable prerequisite proof requiring a committed, reconstructable, successful #51 result before #184 can dispatch. It also adds a sealed completion record tying the live GitHub run, launch preflight, result validation, and runner receipt digest together.

Important: merging this PR authorizes no compute, creates no tag, dispatches no workflow, consumes no seed, and writes no output. Both #51 and #184 must be repinned to the final post-merge source, then receive their exact owner authorization. #51 must run and its reviewed result must merge before #184 can pass preflight.

Verification on current main:
- prereg workflow/local suites: 225 passed
- scoped Ruff: clean
- strict mypy on driver: clean
- diff check: clean

No benchmark or output artifact was created or modified.

### DIFF
--- a/.github/scripts/ipmnist_prereg.py
+++ b/.github/scripts/ipmnist_prereg.py
@@ -63,6 +63,9 @@ class Protocol:
     control: str
     candidate: str
     seeds: tuple[int, ...]
+    summary_filename: str
+    requires_amendment: bool
+    prerequisite: str | None
 
 
 PROTOCOLS: Final = {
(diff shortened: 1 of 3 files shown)


lalalune:
Superseded by merged #772. The replacement preserves both commits from this PR and retains its frozen issue184 protocol/classifier, prerequisite ordering, workflow binding, and completion seal. It additionally closes the audit gap by requiring the committed #51 completion and runner receipts to byte-match the one canonical unexpired artifact uploaded by the verified successful run. No run/tag/output mutation occurred; 225 tests plus Ruff/mypy/diff checks passed.

PR #767, opened 2026-08-16
fix(stacked-horde): complete scalar and resource contracts

lalalune:
Supersedes #752 while preserving its contributor commit. The original integer canonicalization is sound but left gamma/lambda float32 sinks spoofable/noncanonical, accepted arbitrary sequence hooks at serialized boundaries, and permitted n_demons×feature_dim state/config/result allocations beyond signed-int32 resource domains. nexting_spec could also materialize a derived demon cross-product before validation. This successor uses exact tuple constructors plus exact list-or-tuple deserialization, shared exact-host/float32 gamma and lambda validation, and exact aggregate scalar/byte preflights before both direct construction and nexting expansion. Tests cover all NumPy integer families, hostile containers/scalars, compatibility roundtrips, and adjacent allocation-free resource bounds. Verification: 90 full stacked-Horde tests passed after the repair (the one corrected assertion reran green); scoped Ruff and strict mypy clean; diff-check clean. No registered evidence source or outputs are changed.

### DIFF
--- a/alberta_framework/core/stacked_horde.py
+++ b/alberta_framework/core/stacked_horde.py
@@ -51,6 +51,7 @@ from jax import Array
 from jaxtyping import Bool, Float
 
 from alberta_framework._float32 import round_real_to_float32_with_ratio
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 
 __all__ = [
     "StackedHordeConfig",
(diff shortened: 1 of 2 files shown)


lalalune:
Audit follow-up added exact outer/nested deserialization schemas, public state/feature shape and dtype gates, saturating int32 step telemetry, and a signed-int32 scan-result preflight before `lax.scan` output allocation. Allocation-free abstract-shape boundary regression included. Updated verification: 94 passed; scoped Ruff, strict source mypy, and diff-check clean.

lalalune:
Superseded by merged #788, which preserves both the original #752 contributor commit and this repair commit, then integrates the complete stacked-Horde contract with the adjacent SARSA/RTU family. The exact list-or-tuple compatibility contract is retained; focused verification is recorded on #788.

lalalune:
Merged #788 retained the contributor/scalar core but omitted this branch’s final exact-schema, scan-result, public state/input, and saturating-counter follow-up. Those residuals are now rebased together with SARSA transaction residuals in #791; closing this stale base to keep one merge path.

PR #768, opened 2026-08-16
fix(streams): reject class-spoofed non-real mask_prob in PartialObservationWrapper

### LINKED ISSUE #766: partial_observation.py PartialObservationWrapper accepts un-typed mask_prob, leaks raw exceptions
`PartialObservationWrapper.__init__` in `alberta_framework/streams/partial_observation.py` validates RANDOM-mode `mask_prob` with:

```python
if not (0.0 <= mask_prob <= 1.0):
    raise ValueError(f"mask_prob must lie in [0, 1]; got {mask_prob}")
```

There is no type gate at all before this comparison — not even a spoofable `isinstance(value, Real)` check. Any object that merely supports `__le__`/`__ge__` reaches the comparison directly, including:

- an object whose `__class__` property reports `float` (the same `isinstance`-spoofing class already fixed for `_require_int`/`_require_real` families across the step files and in `types.py` (#573/#574), `representation_gradient_mixer.py` (#569), and `partner_policy_fusion.py` (#571));
- any object whose comparison dunder hooks raise, which surfaces a raw uncaught exception instead of the documented `ValueError`.

`PartialObservationWrapper` is exported at package root (`alberta_framework/__init__.py`) and constructed directly by callers — publicly reachable with ordinary concrete inputs.

fix incoming: route `mask_prob` through a `type(value)`-based gate (`actual_type = type(value); if issubclass(actual_type, bool) or not issubclass(actual_type, Real): raise ValueError(...)`) matching the established pattern in `_float32_validation.py` / `out_of_class.py`, converting only after the gate passes and catching `OverflowError`/`TypeError`/`ValueError` from that conversion.

rama0x1:
Closes #766.

## What changed

Same isinstance/type-spoofing defect class as the `_require_int`/`_require_real` family fixes across the step files, `types.py` (#573/#574), `representation_gradient_mixer.py` (#569), and `partner_policy_fusion.py` (#571) — but here the RANDOM-mode `mask_prob` guard in `alberta_framework/streams/partial_observation.py` had no type gate whatsoever, not even a spoofable `isinstance`:

```python
if not (0.0 <= mask_prob <= 1.0):
    raise ValueError(f"mask_prob must lie in [0, 1]; got {mask_prob}")
```

Any object supporting `__le__`/`__ge__` reaches that comparison directly, including a `__class__`-spoofed non-real (`__class__` is an overridable `@property`, so `isinstance(value, Real)` would have passed such an object anyway) whose comparison dunder hooks raise — leaking a raw uncaught exception instead of the documented `ValueError`. A plain `bool` also passed through uncanonicalized since `0.0 <= True <= 1.0` is true.

Added `_require_unit_interval_probability`, matching the established `type(value)`-gate pattern in `_float32_validation.py`/`out_of_class.py`: `actual_type = type(value)` (bypasses the overridable `__class__` slot), reject `bool` and non-`Real` by `issubclass` on the real type, only then convert with `float()` inside a narrow `try/except (OverflowError, TypeError, ValueError)`, and re-check the `[0, 1]` bound on the canonical float.

## Reachability

`PartialObservationWrapper` is exported at package root (`alberta_framework/__init__.py`) and constructed directly by callers with ordinary concrete inputs — publicly reachable. No other production call site passes `mask_prob`; the only internal use is the module's own docstring example, which uses `mode=FIXED` and never touches this code path.

## Verification

Commit under review: `00f6994561c61228d7644d089562aa9a279048ef`.

```bash
.venv/bin/python -m pytest tests/test_partial_observation.py -q -o addopts=""
# 17 passed in 4.23s

# All checks passed!

.venv/bin/python -m mypy
# Success: no issues found in 168 source files

git diff --check
# (clean, exit 0)

.venv/bin/python -m pytest tests/ -k "partial_observation or streams" -q -o addopts=""
# 435 passed, 7968 deselected in 191.58s   (collateral check: no regression in the streams module)
```

**Mutation check** (reverted just the source fix, kept the new tests, reran):

```text
FAILED tests/test_partial_observation.py::TestRandom::test_random_rejects_bool_mask_prob
FAILED tests/test_partial_observation.py::TestRandom::test_random_rejects_non_real_mask_prob
FAILED tests/test_partial_observation.py::TestRandom::test_random_rejects_class_spoofed_non_real_mask_prob
E   RuntimeError: untrusted __ge__ hook executed
3 failed, 14 passed in 4.30s
```

confirming the spoofed object's own `__ge__` hook is what raises through the unpatched comparison — exactly the symptom the fix closes. Restoring the fix returns all 17 tests to passing.

New tests added next to the existing `TestRandom` cases: bool rejection, non-real (`str`) rejection, NaN rejection, and a `__class__`-spoofed object whose `__le__`/`__ge__`/`__float__` all raise `RuntimeError` if invoked — proving the gate rejects it via `type()` before any of those hooks ever run.

## Provenance

AI-assisted (Claude Code, `claude-sonnet-5`, Anthropic). Spoofing/leak behavior reproduced empirically against the unpatched code (mutation check above) before and after the fix, by me, in this session. All verification above performed and independently confirmed by me before pushing.

- [x] logs: full command output reproduced in the Verification section above (mutation check + full test/lint/type-check output); this is a correctness fix with no benchmark artifact to attach.

### DIFF
--- a/alberta_framework/streams/partial_observation.py
+++ b/alberta_framework/streams/partial_observation.py
@@ -21,8 +21,7 @@ just receives less information about the underlying state.
 from __future__ import annotations
 
 import enum
-from numbers import Real
-from typing import TypeVar, cast
+from typing import TypeVar
 
 import chex
 import jax.numpy as jnp
@@ -30,6 +29,7 @@ import jax.random as jr
 from jax import Array
 from jaxtyping import Bool, PRNGKeyArray
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 from alberta_framework.core.types import TimeStep
 from alberta_framework.streams.base import ScanStream
 
@@ -50,25 +50,8 @@ class MaskMode(enum.Enum):
 
 
 def _require_unit_interval_probability(name: str, value: object) -> float:
-    """Return a real number in ``[0, 1]``, rejecting non-real and bool values.
-
-    Uses ``type(value)`` rather than ``isinstance(value, Real)``: ``isinstance``
-    consults the overridable ``__class__`` attribute, so an object whose
-    ``__class__`` property reports ``float`` would otherwise pass the check
-    while still routing the raw ``<=`` comparison to its own (potentially
-    raising) dunder hooks, leaking an uncaught exception instead of the
-    documented ``ValueError``.
-    """
-    actual_type = type(value)
-    if issubclass(actual_type, bool) or not issubclass(actual_type, Real):
-        raise ValueError(f"{name} must be a real number in [0, 1]; got {value!r}")
-    try:
-        number = float(cast(Real, value))
-    except (OverflowError, TypeError, ValueError) as error:
-        raise ValueError(f"{name} must be a real number in [0, 1]; got {value!r}") from error
-    if not (0.0 <= number <= 1.0):
-        raise ValueError(f"{name} must lie in [0, 1]; got {value!r}")
-    return number
+    """Return a canonical probability valid at the float32 execution sink."""
+    return validated_float32_scalar(name, value, lower=0.0, upper=1.0)
 
 
 # =============================================================================
--- a/alberta_framework/streams/partial_observation.py
+++ b/alberta_framework/streams/partial_observation.py
@@ -21,7 +21,8 @@ just receives less information about the underlying state.
 from __future__ import annotations
 
 import enum
-from typing import TypeVar
+from numbers import Real
+from typing import TypeVar, cast
 
 import chex
 import jax.numpy as jnp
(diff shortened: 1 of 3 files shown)


lalalune (changes requested):
The class-spoof fix is directionally right, but this bespoke helper does not yet satisfy the established scalar boundary. It interpolates `{value!r}` on every rejection, so an untrusted object with a raising `__repr__` still leaks that exception before the promised `ValueError`. For actual `Real` subclasses it invokes `float()` and catches only `OverflowError`, `TypeError`, and `ValueError`, so an adversarial conversion hook raising another exception also leaks. More importantly, `mask_prob` is ultimately consumed by `jax.random.bernoulli` at the float32 sink, but this accepts based only on the rounded Python float: an exact value just above 1 that rounds to 1.0 can cross the boundary, and non-builtin exact inputs can be double-rounded. Please route through the shared exact host-ratio + single-rounding float32 validator with `[0,1]` bounds, avoid executing hostile repr/conversion hooks in diagnostics, canonicalize storage, and add hostile-repr/ratio-hook plus exact-adjacent/float32-round-trip regressions. Then rerun the partial-observation and streams panels.

alberta_framework/streams/partial_observation.py
```diff
@@ -148,8 +171,7 @@ def __init__(
             self._schedule = None
 
         if mode == MaskMode.RANDOM:
-            if not (0.0 <= mask_prob <= 1.0):
```
goldcaroline40-ops: Fixes #735

lalalune:
Maintainer follow-up 5df2069 routes mask_prob through the shared exact-host/float32 validator. This removes hostile repr interpolation, normalizes ordinary Real conversion failures, and prevents values that leave [0,1] when narrowed from reaching the JAX sink. Added hostile Real regression; 18 focused tests, Ruff, and strict mypy pass.

lalalune (approved):
Repaired head now uses the shared exact-host/float32 validator, canonicalizes the stored probability, and contains hostile-hook coverage. Independently verified 18 focused tests, Ruff, targeted strict mypy, and diff-check.

PR #770, opened 2026-08-16
fix(sigreg): complete scalar and representation contracts

lalalune:
Supersedes #748 while preserving its contributor commit. The original n_projections gate is useful but did not protect direction, projection, or Epps-Pulley pairwise allocations; latent_dim remained coercive; kernel/epsilon validation admitted hostile Real hooks and unsafe operation endpoints; and public projection paths lacked complete shape/resource guards. This successor adds full exact NumPy integer families, one-read exact-host/float32 scalar validation, operation-safe kernel bounds, signed-int32 direction/projection/pairwise preflights, exact shape checks, stable diagnostics, and hostile/adjacent-boundary tests. A final correction preserves historical Mapping/MappingProxy, partial-default, and ignored type-marker compatibility instead of imposing an unrelated frozen schema. Verification: 46 tests passed; scoped Ruff and strict mypy clean; diff-check clean. No registered evidence source or outputs are changed.

### DIFF
--- a/alberta_framework/core/sigreg.py
+++ b/alberta_framework/core/sigreg.py
@@ -27,6 +27,7 @@ import dataclasses
 import math
 import numbers
 import operator
+from collections.abc import Mapping
 from typing import Any, SupportsIndex, cast
 
 import chex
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #769, which preserves both commits from this branch and adds the final exact scalar/schema/minimum-allocation repairs. #769 passed 212 combined focused tests, Ruff, strict mypy, and diff hygiene.

PR #732, opened 2026-08-16
fix(continual-ia): harden evaluation configuration integer validation

Kayvan-Zahiri:
## Summary
- Hardens `ContinualIAConfig` integer dimensions (`num_steps`, `phase_length`, `observation_dim`, `n_actions`, `n_demons`, `recovery_window`, `bootstrap_resamples`, `bootstrap_seed`) and `IAAcceptanceThresholds` integer attributes (`minimum_seed_count`, `evidence_seed_start`, `maximum_executed_action_credit_mismatches`) with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Canonicalizes all integer attributes to Python `int` at construction time.

## Verification
- `PYTHONPATH=. pytest tests/test_continual_ia_cli.py`: 3 passed
- `PYTHONPATH=. pytest tests/test_continual_ia_evidence.py -k "continual_ia_config or ia_acceptance_thresholds"`: 3 passed
- `ruff check .`: clean
- `mypy alberta_framework/evaluation/continual_ia.py`: clean
- No registered evidence artifacts modified.

PR #771, opened 2026-08-16
fix(continual-ia): preflight scalar and resource domains

lalalune:
## Summary
- supersedes #732 while retaining Kayvan Zahiri attribution
- canonicalizes every configuration, acceptance-threshold, controller-budget, bootstrap, run-seed, and scalar float32 sink with exact trusted Python/NumPy family admission and static hostile-object errors
- preflights derived controller state, per-condition/per-seed histories, scan keys, transition work, bootstrap draws, and bootstrap peak bytes against measured formulas and signed-int32/resource caps before allocation
- preserves the frozen default configuration, list/tuple seed compatibility, full JAX uint32 seed domain, and canonical built-in JSON payload types

## Evidence compatibility
No pinned output was modified. This source is registered for continual_intelligence_amplification, so changing it correctly leaves the frozen artifact fail-closed invalid until the frozen protocol is rerun and strictly validated; this PR does not promote or reinterpret the historical valid rejection. On current main the verifier already reports the same 181 failures (source-provenance drift plus pre-existing canonical controller-budget drift), and this branch preserves that exact failure set while preserving the frozen configuration/threshold payload values.

## Verification
- 126 passed, 9 skipped: tests/test_continual_ia_evidence.py + tests/test_continual_ia_cli.py
- focused post-rebase hostile seed/resource tests: 2 passed
- Ruff: clean
- strict mypy for continual_ia.py: clean
- git diff --check: clean
- outputs/: unchanged

Supersedes #732.

### DIFF
--- a/alberta_framework/evaluation/continual_ia.py
+++ b/alberta_framework/evaluation/continual_ia.py
@@ -129,10 +129,13 @@ def _preflight_static_resources(
     recommendation_state_bytes = 468 + 8 * n_demons
     augmented_state_scalars = 103 + 7 * n_demons
     augmented_state_bytes = 428 + 28 * n_demons
-    # Every retained result exposes six logical history fields: one float64
-    # reward vector, four int64 action vectors, and one bool acceptance vector.
-    per_condition_history_bytes = 41 * num_steps + 16 * n_phases - 8
-    per_seed_history_bytes = len(CONDITION_NAMES) * per_condition_history_bytes
+    # The recommendation scan retains 21 bytes/step of outputs while its
+    # uint32 split-key schedule is live. The three recommendation results each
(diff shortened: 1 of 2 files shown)


lalalune:
Independent resource-formula audit repaired the final two gaps. Static preflight now counts the largest recommendation scan output (21 bytes/step) together with its live split-key schedule (8 bytes/step), not keys alone. Retained history accounting is now exact across the three recommendation arms (41 bytes/step each) and three plain arms (33 bytes/step each, respecting the deliberately shared sentinel vector), plus all phase/recovery arrays: `222*num_steps + 96*n_phases - 48` bytes per seed. Acceptance also compares the frozen schedule without allocating an attacker-sized expected tuple. Focused panel: 24 passed; Ruff, strict mypy, and diff hygiene pass.

PR #773, opened 2026-08-16
fix(experiential-memory): close residual query operation contracts

lalalune:
Residual follow-up to merged #746/#769. The contributor's exact integer, float, schema, state-scalar, and state-byte contracts are preserved on main.

This adds only the remaining operation/resource safety: explicit signed-int32 scalar+byte preflights for each capacity/key-distance and top-k/payload/diagnostic query allocation; float32-sink validation for the positive eviction-weight condition; saturated-int32 age division bounds for staleness/recency scales; and max-normalization of relative eviction weights so finite stored utilities cannot overflow retention priorities while eviction ordering is preserved.

Tests cover adjacent age-scale failure, host-positive values that narrow to zero at the sink, and actual max-float32 eviction with deterministic finite behavior. No evidence sources or outputs are changed.

Validation: 139 focused experiential-memory/policy tests pass; Ruff, full strict mypy (168 files), and diff check pass.

### DIFF
--- a/alberta_framework/core/experiential_memory.py
+++ b/alberta_framework/core/experiential_memory.py
@@ -285,6 +285,12 @@ def _validated_config_float(name: str, value: object, **bounds: Any) -> float:
         raise ValueError(f"{name} must be a finite real scalar")
     return validated_float32_scalar(name, value, **bounds)
 
+
+def _require_float32_allocation(name: str, scalars: int) -> None:
+    if scalars > _INT32_MAX or 4 * scalars > _INT32_MAX:
+        raise ValueError(f"{name} scalar and byte counts must fit signed int32")
+
+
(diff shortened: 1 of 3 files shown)


lalalune:
Final audit found and repaired two operation-sink/resource blockers:

- the prior host threshold rounded down in float32, so its nominal equality still overflowed saturated-age division; the gate now accepts only the first adjacent float32 value whose actual sink division is finite, with equality/adjacency tests;
- per-array checks did not bound aggregate live query resources; the constructor now charges persistent state plus a conservative, enumerated four-byte-equivalent envelope for every full-width logical query temporary and rejects before JAX allocation.

Rebased head: `28d9a9b`. Verification: 142 focused tests passed; scoped Ruff clean; full strict mypy clean (168 files); diff check clean. No evidence source or output changed.

PR #749, opened 2026-08-16
fix(sarsa): harden configuration integer validation and canonicalization

Kayvan-Zahiri:
## Summary
- Hardens `SARSAConfig` integer attributes `n_actions` and `epsilon_decay_steps` with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Canonicalizes all integer attributes to Python `int` at construction time.

## Verification
- `PYTHONPATH=. pytest tests/test_sarsa.py tests/test_differential_sarsa_discounts.py tests/test_differential_sarsa_external_action.py tests/test_differential_sarsa_horizon.py`: 100 passed, 2 skipped
- `ruff check .`: clean
- `mypy alberta_framework/core/sarsa.py`: clean

PR #774, opened 2026-08-16
fix(sarsa): complete scalar, resource, and transaction contracts

lalalune:
## Summary
- preserves #749 contributor commit and completes exact NumPy/Python integer canonicalization
- validates SARSA-owned float32 scalars and exact serialization schemas without invoking hostile conversion/repr hooks
- rejects impossible aggregate direct state scalar/byte counts before GVF/JAX allocation
- saturates int32 telemetry and makes public select/update transactions fail closed on invalid shapes, dtypes, values, or nested update rejection while retaining terminal/zero-gamma diagnostics and disabled-trace recovery

## Provenance
The original contributor commit is retained as the first branch commit. During verification, a concurrent maintainer process added the public transaction gates/tests directly in the isolated worktree; those edits were independently diff-reviewed, repaired against four existing compatibility regressions, and retained only after the complete focused panel passed.

Closes/supersedes #749.

## Verification
- `PYTHONPATH=. .venv/bin/python -m pytest tests/test_sarsa.py tests/test_differential_sarsa_discounts.py tests/test_differential_sarsa_external_action.py tests/test_differential_sarsa_horizon.py -q` — 134 passed
- `.venv/bin/python -m ruff check alberta_framework/core/sarsa.py tests/test_sarsa.py` — clean
- `.venv/bin/python -m mypy alberta_framework/core/sarsa.py` — clean
- `git diff --check` — clean

### DIFF
--- a/alberta_framework/core/sarsa.py
+++ b/alberta_framework/core/sarsa.py
@@ -101,9 +101,7 @@ _ACTUAL_INT_TYPES = frozenset(
         np.ulonglong,
     }
 )
-_ACTUAL_FLOAT_TYPES = frozenset(
-    {float, *(np.dtype(code).type for code in ("e", "f", "d", "g"))}
-)
+_ACTUAL_FLOAT_TYPES = frozenset({float, *(np.dtype(code).type for code in ("e", "f", "d", "g"))})
 
 
 def _require_int32(name: str, value: object, *, minimum: int, maximum: int = _INT32_MAX) -> int:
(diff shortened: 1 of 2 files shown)


lalalune:
Merged #788 retained the contributor/config core but omitted this branch’s exact schemas, wrapper-state preflight, saturating telemetry, and public fail-closed transaction layer. Those residuals are now rebased on current main and verified together with stacked Horde in #791; closing this stale base.

PR #783, opened 2026-08-16
fix(options): preserve config mapping compatibility

lalalune:
Follow-up to the combined #775 integration. The integration correctly landed the exact scalar/resource contracts, but also narrowed `STOMPConfig.from_config` beyond its established compatibility surface.

This residual restores ordinary `Mapping`/`MappingProxyType` inputs, partial/default payloads, historical type-marker tolerance, and exact list-or-tuple nested sequence forms. It still rejects arbitrary sequence subclasses and class-spoofed mappings before iterator/repr hooks, and delegates every decoded scalar to the hardened constructors already on main.

Verification on the post-#775 base:
- 159 options/STOMP tests passed
- Ruff passed
- strict targeted mypy passed
- git diff --check passed

### DIFF
--- a/alberta_framework/core/options.py
+++ b/alberta_framework/core/options.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 import dataclasses
 import functools
 import operator
+from collections.abc import Mapping
 from typing import Any, SupportsIndex, cast
 
 import chex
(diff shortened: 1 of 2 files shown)


lalalune:
Deep compatibility review found and fixed one remaining boundary leak: unknown outer fields/constructor failures escaped as raw TypeError after Mapping decode. Follow-up fd7347e normalizes non-ValueError constructor failures while preserving field-specific validation messages and the intended MappingProxy/partial/tuple/historical-marker compatibility. Verification: 105 options tests passed; Ruff, strict source mypy, and diff-check clean. Auto-merge remains enabled.

PR #756, opened 2026-08-16
fix(recurrent-trace-actor-critic): harden configuration integer validation and canonicalization

Kayvan-Zahiri:
## Summary
- Hardens `RecurrentTraceActorCriticConfig` integer fields `n_actions`, `hidden_size`, `encoder_width`, and `output_width` with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Canonicalizes all integer attributes to Python `int` at construction time.

## Verification
- `PYTHONPATH=. pytest tests/test_recurrent_trace_actor_critic.py`: 75 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/recurrent_trace_actor_critic.py`: clean

PR #787, opened 2026-08-16
fix(recurrent-actor-critic): validate scalar and resource contracts

lalalune:
## Summary
- supersedes #756 while preserving Kayvan Zahiri contributor attribution
- rebased after main integrated the original narrow integer patch; this branch now contains only residual hardening missing from current main
- validates exact host and narrowed float32 endpoint domains for every recurrent actor-critic scalar, including NumPy values that otherwise round across closed or half-open bounds
- corrects the logical state-scalar budget for the typed JAX key representation while retaining the independently exact byte formula and existing state_resource_budget API
- preserves current serialization schemas, default canonical JSON hashes, valid eager/JIT behavior, and all current-main integrated contracts

## Audit result
Current main already contains the direct integer validation, resource preflight, and serialization work integrated from the original PR. The residual audit found two concrete gaps: NumPy host values outside an endpoint could round into an accepted float32 value, and state_scalars counted one more logical leaf than the initialized state actually owns. Tests measure all four Taylor/adaptive state-tree variants directly. No registered evidence source or pinned output is changed.

## Verification
- 290 tests passed in the full recurrent_trace_actor_critic + rtu_taylor_correction rerun; the sole compatibility-message failure exposed by that run was fixed and its targeted rerun passed
- 127 focused scalar/resource/hostile/default tests passed after the current-main rebase
- Ruff clean
- strict mypy clean
- git diff --check clean
- earlier pre-integration Forager consumer panel: 16 passed

Supersedes #756.

### DIFF
--- a/alberta_framework/core/recurrent_trace_actor_critic.py
+++ b/alberta_framework/core/recurrent_trace_actor_critic.py
@@ -267,12 +267,22 @@ def _require_int32(name: str, value: object, *, minimum: int, maximum: int = _IN
     return canonical
 
 
-def _validated_config_float(name: str, value: object) -> float:
+def _validated_config_float(name: str, value: object, **bounds: Any) -> float:
     if type(value) is bool or type(value) is np.bool_:
         raise ValueError(f"{name} must be numeric, not bool")
     if type(value) not in (_ACTUAL_INT_TYPES | _ACTUAL_FLOAT_TYPES):
         raise ValueError(f"{name} must be a finite real scalar")
-    normalized = validated_float32_scalar(name, value)
+    try:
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by merged #788 plus the narrow residual #796. #788 preserves the #756 contributor commit and the integrated RTU/SARSA/Horde resource work. #796 carries the useful exact-host endpoint and canonical feature-dimension behavior from this conflicting branch, and additionally enforces exact serialized JSON value types and restores the Taylor-suite compatibility contract on current main. This avoids reintroducing #787’s duplicate RTU implementation.

PR #789, opened 2026-08-16
fix(dual-replay): contain hostile float hooks

lalalune:
Supersedes #778 while preserving its contributor commit. The original identified a real hostile `Real.as_integer_ratio()` exception leak, but its signed-int32 aggregate byte cap conflicts with the existing exact uint32 `persistent_bytes` state/accounting contract and unnecessarily rejects representable states. This replacement retains the exception containment and mutation-resistant coverage for every float field, while leaving the already-merged #737 exact integer, schema, allocation, and uint32 resource contracts unchanged.

Validation: 167 focused dual-replay tests passed; Ruff passed; targeted strict mypy passed; diff check passed.


lalalune:
Closing in favor of #786 after semantic comparison. Both retain the valid contributor-authored `as_integer_ratio()` exception normalization and omit #778’s unsupported signed-int32 aggregate-byte ceiling. #786 additionally covers the later accepted-`Real` nonzero comparison hook, which occurs outside this patch’s widened `try` and can still raise an ordinary exception; it normalizes that path without evaluating hostile `repr`. Thus #786 is the strict superset.

PR #780, opened 2026-08-16
fix(streams): harden gymnasium gamma/epsilon scalar contracts

rama0x1:
Closes #777.

## What changed

Same recurring "documented scalar contract, zero enforcement" defect class as the other `streams/`/`steps/`/`core/` hardening PRs in this pipeline, but here the gap was total: `alberta_framework/streams/gymnasium.py` never validated `gamma` (TD discount factor) or `epsilon` (exploration probability) at *any* of their call sites — not even a bare `isinstance` check.

Every other discount-factor consumer in this codebase enforces `gamma in [0.0, 1.0]` through the shared `validated_float32_scalar` helper: `core/sarsa.py:94`, `core/world_model.py:722`, `core/horde_actor_critic.py:391`/`2061`, `core/options.py:1454`, `core/stacked_horde.py:171`. `gymnasium.py` was the outlier. Concretely, before this fix:

```python
_, targets = collect_trajectory(
    env, policy, num_steps=1, mode=PredictionMode.VALUE,
    value_estimator=lambda x: 10.0, gamma=float("nan"),
)
# targets == [[nan]]  -- no error, NaN silently enters the learning signal

_, targets = collect_trajectory(
    env, policy, num_steps=1, mode=PredictionMode.VALUE,
    value_estimator=lambda x: 10.0, gamma=50.0,
)
# targets == [[501.]]  -- gamma=50 is not a discount factor, no error

GymnasiumStream(env, mode=PredictionMode.VALUE, gamma=float("nan"))
# constructs successfully; self._gamma is nan
```

`make_epsilon_greedy_policy`'s `epsilon` (documented "Probability of taking a random action") had the identical gap: `epsilon=5.0` or `epsilon=-0.1` was accepted silently, permanently defeating or inverting the intended exploration schedule with no signal to the caller.

Fix: validate `gamma` in `collect_trajectory`, `GymnasiumStream.__init__`, `TDStream.__init__`, and `make_gymnasium_stream` (via `GymnasiumStream`), and `epsilon` in `make_epsilon_greedy_policy`, all through the existing shared `validated_float32_scalar(name, value, lower=0.0, upper=1.0)` (`core/_float32_scalars.py`) — the same helper already reused across `core/` and `benchmarks/`. This rejects non-`Real`/bool-spoofed values, NaN/inf, and out-of-`[0, 1]` values with a clean `ValueError`, while leaving every value already in range (`0.0`, `1.0`, `0.99`, `0.1`, etc.) byte-identical to before.

## Reachability

[...]

PR #790, opened 2026-08-16
fix(streams): complete Gymnasium schedule and shape contracts

lalalune:
Residual successor to #780 after its gamma/epsilon scalar fix was integrated through #782. Contributor rama0x1 is retained as a co-author.

The deep audit found remaining reachable failures: coercive/empty/oversized `num_steps`, JAX seed aliasing and epsilon sub-seed overflow, environment execution before output-allocation validation, truthy non-bool feature flags and non-enum modes, ignored nonzero `Discrete.start`, multidimensional `MultiDiscrete` shape drift, and invalid uniform sampling from nonfinite Box bounds.

This adds exact schedule/seed/bool/mode/callable gates; signed-int32 scalar+byte preflight for trajectory and stream shapes before reset/step; exact Python shape arithmetic; correct Discrete/MultiDiscrete starts and flattening; and finite ordered float32 Box bounds.

Validation: 68 focused Gymnasium/bootstrap/package-boundary tests pass; Ruff, full strict mypy (168 files), and diff check pass. No evidence sources or outputs changed.

### DIFF
--- a/alberta_framework/streams/gymnasium.py
+++ b/alberta_framework/streams/gymnasium.py
@@ -149,14 +149,21 @@ def _flatten_observation(obs: Any, space: gymnasium.spaces.Space[Any]) -> Array:
     """
     import gymnasium
 
-    if isinstance(space, gymnasium.spaces.Box):
-        return jnp.asarray(obs, dtype=jnp.float32).flatten()
-    elif isinstance(space, gymnasium.spaces.Discrete):
-        return jnp.array([float(obs)], dtype=jnp.float32)
-    elif isinstance(space, gymnasium.spaces.MultiDiscrete):
-        return jnp.asarray(obs, dtype=jnp.float32).reshape((-1,))
-    else:
-        raise ValueError(f"Unsupported space type: {type(space).__name__}")
+    try:
+        if isinstance(space, gymnasium.spaces.Box):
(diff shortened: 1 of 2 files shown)


lalalune:
Final audit added three residual fail-closed boundaries: exact finite post-flatten runtime shape checks so malformed environment/policy values cannot bypass declared-space output preflight; aggregate MultiDiscrete bounds materialization preflight; and non-callable policy rejection before `gymnasium.make` can construct an external environment.

Rebased head: `$(git rev-parse --short HEAD)`. Verification: 70 focused Gymnasium/bootstrap/package-boundary tests passed; scoped Ruff clean; strict source mypy clean; diff check clean. `gymnasium.py` is outside the five registered evidence claim inventories; no outputs changed.

lalalune:
Independent operation-sink audit found one remaining blocker after 107c6e2: a finite ordered Box with bounds [-float32_max, +float32_max] passes validation, but the uniform sampler overflows high-low and deterministically emits non-finite actions. Auto-merge is temporarily disabled. Repair must prove finite sampling for every accepted bound pair (or reject unsafe spans) and add the max-range regression. Also retain the newly added pre-gym.make policy gate and verify epsilon-wrapper base_policy callable validation.

PR #793, opened 2026-08-16
fix(micro-continual): reject n_steps that overflow signed int32

### LINKED ISSUE #792: micro_continual: MicroStreamConfig accepts n_steps that overflows signed int32, silently corrupting the regime schedule
## Summary

`MicroStreamConfig.__post_init__` in `alberta_framework/benchmarks/micro_continual.py`
validates that `n_regimes`, `regime_length`, `dim`, `n_classes`, and
`n_components` are each a positive Python `int`, but never bounds the
*derived* `n_steps = n_regimes * regime_length` against the signed int32
domain. `micro_continual` is the CLI-exposed "prototype on the micro suite"
lane (`repository-contract.md`'s open-lane doc, and directly reachable via
`--n-regimes`/`--regime-length` on `python -m
alberta_framework.benchmarks.micro_continual run|ladder`), so this is
user/automation-reachable, not just a theoretical input.

`stream()` (the function every arm and the transfer-validation ladder
consume) then does:

```python
regime_index = (
    jnp.arange(n_steps, dtype=jnp.int32) // config.regime_length
).astype(jnp.int32)
```

For `n_steps > 2**31 - 1` this stores per-step index values in a `jnp.int32`
buffer that cannot represent them. Casting an out-of-range Python/NumPy int
to `int32` truncates with two's-complement wraparound rather than raising —
confirmed directly:

```pycon
>>> import numpy as np
>>> np.int32(2**31)      # one past INT32_MAX
-2147483648
```

So a config such as `n_regimes=60_000, regime_length=60_000` (`n_steps =
3_600_000_000`, i.e. `~1.68x` over `INT32_MAX = 2_147_483_647`) constructs
without error today, and every regime-index value computed for steps past
the wraparound boundary would be silently wrong (negative), corrupting the
regime schedule — and therefore the measured stream — instead of failing
closed.

This is exactly the resource/scalar-contract gap this repository has been
closing across other benchmark/evaluation configs this cycle (e.g.
`alberta_framework/benchmarks/upgd_ipmnist.py`'s `IPMNISTConfig` already
guards the analogous `n_tasks * task_length > _INT32_MAX` product with an
explicit `ValueError`; `micro_continual.py`'s sibling config never got the
same guard).

## Reproduction

```bash
.venv/bin/python -c "
from alberta_framework.benchmarks.micro_continual import MicroStreamConfig
cfg = MicroStreamConfig(family='input_permutation', n_regimes=60_000, regime_length=60_000)
print(cfg.n_steps, cfg.n_steps > 2**31 - 1)  # 3600000000 True -- constructs today, should raise
"
```

## Fix

Add the same `n_regimes * regime_length > _INT32_MAX` bound check
`upgd_ipmnist.IPMNISTConfig` already applies, raising `ValueError` at
construction (before any array is built) instead of allowing the silent
int32 wraparound downstream.

## Scope

`alberta_framework/benchmarks/micro_continual.py` only (my assigned
territory: `alberta_framework/{benchmarks,evaluation,utils}/`). One
variable: bound the one derived product that overflows int32 today.

rama0x1:
Fixes #792.

## What

`MicroStreamConfig.__post_init__` (`alberta_framework/benchmarks/micro_continual.py`)
validated that `n_regimes` and `regime_length` were each a positive int, but
never bounded their product `n_steps = n_regimes * regime_length` against the
signed int32 domain. `stream()` (consumed by every micro-suite arm and the
transfer-validation ladder) casts the per-step regime index with
`jnp.arange(n_steps, dtype=jnp.int32) // regime_length`; once `n_steps`
exceeds `2**31 - 1` that cast silently wraps to negative index values
instead of raising, corrupting the regime schedule for every step past the
boundary rather than failing closed. `micro_continual` is the CLI-exposed
"prototype on the micro suite" lane (`--n-regimes`/`--regime-length` are
direct CLI flags), so this is reachable by a real run, not just a synthetic
input.

This mirrors the exact guard `upgd_ipmnist.IPMNISTConfig` already applies to
its analogous `n_tasks * task_length > _INT32_MAX` product — that config has
the check, its `micro_continual` sibling didn't.

## Fix

Add the same `n_regimes * regime_length > _INT32_MAX` bound, raising
`ValueError` at construction time (before any array is built) instead of
allowing the silent int32 wraparound downstream. One file, one added
invariant — no other behavior changes.

## Evidence

Reproduction of the bug on `origin/main` before the fix (confirms the
config constructed silently):

```
n_steps = 60_000 * 60_000 = 3_600_000_000  # > 2**31 - 1 = 2_147_483_647
MicroStreamConfig(family="input_permutation", n_regimes=60_000, regime_length=60_000)
# constructed without error pre-fix
```

Underlying wraparound mechanism (why this corrupts rather than merely
crashes):

```pycon
>>> import numpy as np
>>> np.int32(2**31)      # one past INT32_MAX
-2147483648
```

Commit under test: `8ed5835c015ea7050ae46947b2a27f26f8ff8b49` (branch rebased onto `origin/main` at
merge of #785 / `76ac5df`).

Focused tests (bounded, scoped — full suite not run per sandbox policy):

```
.venv/bin/python -m pytest tests/test_micro_continual.py -q -o addopts=""
194 passed, 2 warnings in 100.47s
.venv/bin/python -m pytest tests/test_micro_continual.py -q -o addopts="" -k TestConfig
65 passed, 129 deselected in 3.66s
```

Lint / types (scoped to the two changed files):

```
.venv/bin/python -m ruff check alberta_framework/benchmarks/micro_continual.py tests/test_micro_continual.py
All checks passed!
.venv/bin/python -m mypy alberta_framework/benchmarks/micro_continual.py
```

`git diff --check`: clean.

No `outputs/` artifacts, evidence registry entries, or scientific claims are
touched — this is a validation-only fix in an unpromoted development lane.
`alberta-evidence-status` is unaffected (`micro_continual.py` is not a
registered evidence source for any of the 5 claims).

## Deviations / what remains open

None from plan. Scope was kept to the one file / one invariant per the
"one lane, one variable" rule; sibling configs elsewhere in the repo were
not audited or touched here.

Declared identity: `--client claude-code --provider anthropic --model
claude-sonnet-5`, lane `rama0x1-asi-claude-worker-2`. Receipt signing
deferred — operator handling centrally (overnight unattended run).

### DIFF
--- a/alberta_framework/benchmarks/micro_continual.py
+++ b/alberta_framework/benchmarks/micro_continual.py
@@ -204,6 +204,8 @@ def _require_nonempty_string(value: object, *, context: str) -> str:
 # Configuration
 # =============================================================================
 
+_INT32_MAX: int = 2**31 - 1
+
 
 @dataclass(frozen=True)
 class MicroStreamConfig:
@@ -270,6 +272,17 @@ class MicroStreamConfig:
             value = getattr(self, name)
             if not isinstance(value, int) or isinstance(value, bool) or value <= 0:
                 raise ValueError(f"{name} must be a positive integer, got {value!r}")
+        if self.n_regimes * self.regime_length > _INT32_MAX:
+            # stream() casts the per-step regime index to dtype=jnp.int32
+            # (``jnp.arange(n_steps, dtype=jnp.int32) // regime_length``); an
+            # uncaught n_steps beyond the signed int32 domain would silently
+            # wrap the derived regime index instead of raising, corrupting
+            # every regime assignment past the wraparound point.
+            raise ValueError(
+                "derived n_steps (n_regimes * regime_length) must fit in signed "
+                f"int32, got {self.n_regimes} * {self.regime_length} = "
+                f"{self.n_regimes * self.regime_length} > {_INT32_MAX}"
+            )
         sparsity = self.component_sparsity
         if not isinstance(sparsity, int) or isinstance(sparsity, bool) or sparsity <= 0:
             raise ValueError(
--- a/tests/test_micro_continual.py
+++ b/tests/test_micro_continual.py
@@ -227,6 +227,25 @@ class TestConfig:
     def test_n_steps(self):
         assert TINY.n_steps == 4 * 25
 
+    def test_n_steps_int32_overflow_rejected(self):
+        # stream() casts the per-step regime index to dtype=jnp.int32
+        # (``jnp.arange(n_steps, dtype=jnp.int32) // regime_length``). An
+        # n_steps beyond the signed int32 domain must be rejected at
+        # construction instead of silently wrapping every regime index past
+        # the boundary to a negative value.
+        with pytest.raises(ValueError, match="n_steps"):
+            MicroStreamConfig(
+                family="input_permutation",
+                n_regimes=60_000,
+                regime_length=60_000,
+            )
+        # The boundary itself (exactly INT32_MAX) must still be accepted.
+        MicroStreamConfig(
+            family="input_permutation",
+            n_regimes=1,
+            regime_length=2**31 - 1,
+        )
+
     def test_to_config_roundtrip(self):
         rebuilt = MicroStreamConfig(**TINY.to_config())
         assert rebuilt == TINY


lalalune (changes requested):
The product check is directionally correct but the proposed accepted boundary is not executable-safe. `n_steps == INT32_MAX` makes the int32 `jnp.arange` alone roughly 8.6 GB, before the batched observations, targets, regime IDs, keys, and retained results. This needs exact scalar-and-byte/aggregate stream-output preflight, not only an index-domain check, with the last legal byte boundary pinned allocation-free. The surrounding per-field gates also still use spoofable `isinstance` and interpolate hostile repr, so a class-spoofed/raising integer can escape before this new product check. Please use the established exact trusted Python/NumPy integer canonicalization, avoid repr hooks, account all derived persistent/output/work arrays before allocation, and update the boundary test (INT32_MAX steps must reject, not accept).

lalalune:
Superseded by #805, which preserves commit 8ed5835 and extends the fix to the complete materialized-stream resource contract. The original INT32_MAX accepted endpoint still implied a multi-gigabyte stream; #805 adds aggregate persistent-byte preflight, exact/canonical host integer validation, hostile-subclass containment, and allocation-free boundary/adjacent tests.

kenjohnscreates (approved):

Verified by re-applying the diff onto current `main` (`33c8f89`) in a scratch worktree; both files **byte-identical to PR head `8ed5835c`**.

**Both factual premises verified against `main`.** The int32 cast is exactly where the PR says: `micro_continual.py:522` — `jnp.arange(n_steps, dtype=jnp.int32) // config.regime_length` — so an `n_steps` past `2**31 − 1` wraps the derived regime index negative instead of raising. And the IPMNIST precedent is real: `upgd_ipmnist.py:322` already guards `n_tasks * task_length > _INT32_MAX` with an identically-named module constant. This is the sibling config lacking its sibling's guard; the fix is the minimal mirror (construction-time `ValueError`, one added invariant, no other behavior change), and it correctly sits after the per-field positive-int checks so the product is well-defined when evaluated.

**Test is discriminating and boundary-exact.** It asserts rejection at `60_000 × 60_000` and — importantly — that `1 × (2**31 − 1)` is still accepted, pinning the `>` (not `>=`) semantics to match the IPMNIST guard.

**Battery (arm64 M1):** `tests/test_micro_continual.py` **194/194**; `ruff check .` clean; repo-config strict mypy clean at **168 source files** (today's main count). Failing-test-first reproduced: tests-only against baseline → `test_n_steps_int32_overflow_rejected` fails deterministically (config constructs silently); full diff → green.

**Exposure:** `micro_continual` is the house prototype lane (the "micro_continual prototype → registry pin → screen" method) — a real CLI-reachable measurement surface, not a synthetic input — so a silent schedule corruption there is squarely the fix-class this repo cares about. Not a five-claim registered source; no pinned artifact perturbed (valid configs are unchanged by construction). Evidence-head marker matches the head.
Compute receipt: 1925700 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

lalalune:
Superseded by #808, which preserves the contributor commit/authorship while replacing the unsafe INT32_MAX accepted boundary with exact aggregate output+generation-work accounting, exact integer/schema/seed admission, and allocation-free adjacent-boundary tests.

lalalune:
Closed as superseded by merged #808 plus corrective #823. Your original overflow fix and authorship were preserved in #808. The final chain replaces the unsafe INT32_MAX endpoint with exact retained-stream scalar/byte accounting, trusted Python/NumPy integer canonicalization, exact uint32 seed gates, float32 sink validation, hostile-safe exact schema keys, and pre-allocation Bayes score-work bounds.

PR #798, opened 2026-08-16
fix(prototype-memory): complete resource and transaction contracts

lalalune:
Supersedes and completes #784 while preserving byt61's contributor commit and authorship.

The original correctly preflights the derived persistent-state allocation, but public operations still admitted broadcastable wrong shapes, malformed adopted state, negative runtime thresholds, a wrapping lifetime counter, scan outputs beyond signed-int32 accounting, and subnormal divisors. This replacement retains the allocation work and adds exact state shape/dtype plus dynamic invariants, eager/outer-JIT operand guards, atomic invalid-state/threshold rollback, saturating step accounting, exact scan-output scalar/byte preflight, positive-normal bandwidth validation, and hostile-safe Mapping deserialization compatibility.

Verification at `c32e1095f3c1385a8ec4d67b60c430ab1f1b8522` on latest main:
- 141 prototype-memory tests passed after rebase
- 185 prototype-memory + late-wave transaction tests passed before the clean rebase
- Ruff passed
- full strict mypy passed after one local redundant-cast correction; targeted strict mypy recheck passed
- git diff --check passed

### DIFF
--- a/alberta_framework/core/prototype_memory.py
+++ b/alberta_framework/core/prototype_memory.py
@@ -93,6 +93,22 @@ def _require_int(
     return number
 
 
+_UINT32_MAX: int = 4294967295
+
+
+def _require_float32_resource(
+    name: str,
+    *,
+    vector_scalars: int,
+    fixed_scalars: int = 0,
+) -> None:
+    total_scalars = vector_scalars + fixed_scalars
+    if total_scalars > _INT32_MAX:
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #804, which preserves both #784/#798 commits and adds the missing exact config/learner type, field, and canonical JSON scalar schema gates plus hostile Mapping regressions.

PR #805, opened 2026-08-16
fix(micro-continual): complete stream resource contracts

lalalune:
Preserves contributor commit 8ed5835 from #793 and extends its valid n_steps guard to the actual materialized-stream contract: exact built-in/NumPy integer families, canonical ints, hostile subclass containment, signed-int32 dimensions, and aggregate persistent numeric byte preflight across all returned arrays. Replaces the unsafe claimed INT32_MAX schedule endpoint with an allocation-free exact aggregate byte boundary/adjacent test. Validation: full test_micro_continual.py 213 passed; TestConfig 84 passed; Ruff and strict mypy passed; diff check clean.

### DIFF
--- a/alberta_framework/benchmarks/micro_continual.py
+++ b/alberta_framework/benchmarks/micro_continual.py
@@ -204,6 +204,8 @@ def _require_nonempty_string(value: object, *, context: str) -> str:
 # Configuration
 # =============================================================================
 
+_INT32_MAX: int = 2**31 - 1
+
 
 @dataclass(frozen=True)
 class MicroStreamConfig:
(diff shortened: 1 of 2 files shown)


lalalune:
Closing in favor of #808 after exact comparison. #805 undercounts the returned GaussianMicroStream payload (`y` is a third n_steps int32 leaf, but its formula uses only two step-index leaves) and bounds only persistent leaves, not simultaneously reachable generation work. #808 includes the exact 3*n_steps returned index leaves, conservative geometry/gather/noise work, hardened mapping schema, public seed validation, and Bayes sample/work bounds, with an adjacent allocation-free aggregate byte test.

PR #816, opened 2026-08-16
fix(prototype-memory): restore historical config compatibility

lalalune:
Corrects only the serialization compatibility regression in merged #815 while retaining all of #815's stronger query/update working-set bounds, pre-conversion wrappers, exact operand/state contracts, saturating counters, and atomic transaction behavior.

Before the schema-tightening commits, both loaders copied mappings, ignored optional type markers, delegated scalar normalization to constructors, allowed omitted defaulted config fields, and tolerated outer learner metadata. Exact full JSON schemas and built-in-only scalar admission were therefore new breaking policies, not established serialized contracts.

This residual keeps genuine Mapping/MappingProxy intake and hostile-hook normalization, restores optional markers/default fields/outer metadata, permits NumPy scalars through the hardened canonical constructors, still rejects unknown constructor fields and non-Mapping nested configs, and never weakens any resource or runtime boundary.

Verification on the rebased head:
- 195 PrototypeMemory + late-wave transaction tests passed
- Ruff passed
- strict targeted mypy passed
- git diff --check passed

### DIFF
--- a/alberta_framework/core/prototype_memory.py
+++ b/alberta_framework/core/prototype_memory.py
@@ -202,25 +202,13 @@ class PrototypeMemoryConfig:
             payload = dict(config)
         except Exception as error:
             raise ValueError("PrototypeMemoryConfig mapping could not be read") from error
-        if payload.pop("type", None) != "PrototypeMemoryConfig":
-            raise ValueError("PrototypeMemoryConfig type is invalid")
-        expected = {
-            "feature_dim",
-            "n_classes",
-            "slots_per_class",
-            "update_rate",
-            "novelty_threshold",
-            "bandwidth",
-        }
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after audit: current main already preserves genuine / compatibility and normalizes mapping-hook failures, while #816 would remove the exact type, field, and canonical JSON scalar schema restored in #804. Historical constructor permissiveness is not an authorization to weaken the now-explicit serialized contract. No unique safe residual remains.

lalalune:
Reconstructed conflict-free atop merged #815 as #822; this preserves the historical optional-marker/defaulted-field/NumPy-scalar/outer-metadata loader behavior while retaining #815 hardening.

PR #817, opened 2026-08-16
fix(closed-loop): complete integer and resource contracts

lalalune:
Supersedes #803 while preserving its contributor commit and authorship. The original integer-gap diagnosis is correct, but its error path still invoked hostile repr and n_states could reach quadratic NumPy/JAX transition allocations at INT32_MAX. This successor removes value formatting from integer failures, requires exact config records, uses the canonical validated phase value, and preflights the complete persistent RiverSwim NumPy transitions/rewards plus JAX logits/rewards before allocation. Adds hostile spoof/subclass/repr, exact config, and allocation-free resource boundary/adjacent tests. Validation on rebased current main: 91 closed-loop tests passed; scoped Ruff, strict mypy, root imports, and diff check passed.

### DIFF
--- a/alberta_framework/streams/closed_loop.py
+++ b/alberta_framework/streams/closed_loop.py
@@ -101,12 +101,21 @@ def _require_builtin_int(
     instead of the ``ValueError`` this contract promises.
     """
     if type(value) is not int:
-        raise ValueError(f"{name} must be a built-in integer, got {value!r}")
+        raise ValueError(f"{name} must be a built-in integer")
     if value < minimum:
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by #827. #817 deliberately rejected historically supported NumPy integer scalars and omitted the separately requested bound on `2 ** n_states` exact policy enumeration. #827 preserves the contributor, adds full concrete NumPy canonicalization, exact 64 MiB resident resource accounting with adjacent boundaries, an n<=12 enumeration gate, counter saturation, runtime state contracts, and the registered evidence-impact disclosure.

PR #826, opened 2026-08-16
fix(working-memory): complete public resource contracts

lalalune:
Supersedes #812 while preserving its contributor commit and authorship. The original trace-state preflight is valid but incomplete: its accepted boundary can still allocate a >2 GiB feature vector, unused signal dimensions can allocate oversized zero vectors, batched transform outputs were unbounded, and step_count wrapped at int32 max. This successor retains exact scalar/state checks, removes redundant unreachable uint32 accounting, bounds every signal vector, feature output, trace state, and batch output/default-gate workspace in signed-int32 scalar/byte domains, and saturates step_count. Adds allocation-free exact boundary/adjacent tests. Validation after rebase to current main: 143 focused tests passed; Ruff, strict mypy, root imports, and diff check passed.

### DIFF
--- a/alberta_framework/core/working_memory.py
+++ b/alberta_framework/core/working_memory.py
@@ -214,6 +214,22 @@ def _require_int32(name: str, value: object, *, minimum: int) -> int:
     return number
 
 
+_UINT32_MAX: int = 4294967295
+
+
+def _require_float32_resource(
+    name: str,
+    *,
+    vector_scalars: int,
+    fixed_scalars: int = 0,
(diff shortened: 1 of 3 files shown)


lalalune:
Closed in favor of merged #825, which preserves #812 and is substantially stronger on exact serialized schemas, adopted-state shape/dtype/dynamic invariants, rollback, and external-gate semantics. I am replaying only the two remaining resource residuals (single-vector/feature bytes and default batch-gate workspace) on current main.

PR #830, opened 2026-08-16
fix(working-memory): close allocation residuals

lalalune:
Follow-up to merged #825 after exact comparison with superseded #826. #825 is the authoritative stronger integration; this applies only its missing allocation residuals: signed-int32 byte bounds for each public signal vector and feature output, removal of unreachable uint32 accounting after the stricter int32 byte gate, and batch accounting for the default float32 gate workspace in addition to returned features/verdicts. Allocation-free boundary tests cover feature, unused zero-vector, trace-state, and batch products. Validation: 152 focused tests, Ruff, strict mypy, and diff check pass.

### DIFF
--- a/alberta_framework/core/working_memory.py
+++ b/alberta_framework/core/working_memory.py
@@ -241,9 +241,6 @@ def _require_int32(name: str, value: object, *, minimum: int) -> int:
     return number
 
 
-_UINT32_MAX: int = 4294967295
-
-
 def _require_float32_resource(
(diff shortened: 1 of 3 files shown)


lalalune:
Superseded by #835, which includes these allocation bounds and also fixes the confirmed finite-large RMS/effective-dimension/innovation operation sinks plus host-before-JAX transform conversion and aggregate working-set preflight.

PR #810, opened 2026-08-16
fix(world-model): harden world model configurations with strict scalar and integer validation

Kayvan-Zahiri:
## Summary
- Adds `__post_init__` to `ActionConditionedWorldModelConfig` and `WorldModelConfig` with `_require_int32` and `_validate_hidden_sizes` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Validates float scalar fields (`gamma`, `observation_scale`, `reward_scale`, `utility_decay`, `error_decay`, `observation_clip_margin`, `max_delta_scale`) with `validated_float32_scalar` and canonicalizes attributes at construction time.

## Verification
- `PYTHONPATH=. pytest tests/test_world_model.py tests/test_action_conditioned_world_model.py`: 66 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/world_model.py`: clean

PR #837, opened 2026-08-16
fix(world-model): validate complete configurations and resources

lalalune:
Supersedes #810 while preserving Kayvan Zahiri's validation contribution and removing unrelated formatting churn.

This replacement validates both public one-step world-model config dataclasses at construction time: exact trusted Python/NumPy integer families, exact booleans/enums/containers, every float32-consumed scalar in both its host and narrowed domains, and canonicalized stored values. It also preflights action interactions, input/head/layer shapes, and the complete direct learner+wrapper state lower bound before JAX allocation.

Compatibility is retained at the serialized boundary: mappings remain accepted, both historical tuple sequences and canonical JSON lists decode, optional/legacy fields remain permissive, and no exact-field/type-discriminator schema tightening is introduced. Hostile sequence/mapping conversion failures are normalized without formatting attacker-controlled objects.

Validation:
- 243 world-model, dreaming, latent-model, ensemble, and Prototype ensemble tests passed
- 81 focused tests passed again after rebasing current main
- Ruff clean on changed files
- mypy clean on `world_model.py`
- `git diff --check` clean

This is L0 mechanism hardening only. `world_model.py` appears in a historical slowly-changing regression source snapshot, so this source change invalidates that snapshot's byte identity if it is replay-checked; it does not alter any registered five-claim evidence source inventory or authorize/promote evidence.

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -139,6 +139,45 @@ def _serialized_sequence(name: str, value: object) -> tuple[Any, ...]:
     return tuple(cast(list[Any] | tuple[Any, ...], value))
 
 
+def _require_scan_resource(name: str, *, float32_scalars: int, bool_scalars: int) -> None:
+    if float32_scalars + bool_scalars > _INT32_MAX:
+        raise ValueError(f"derived {name} scalar count must fit in signed int32")
+    if 4 * float32_scalars + bool_scalars > _INT32_MAX:
+        raise ValueError(f"derived {name} byte count must fit in signed int32")
+
+
+def _scan_array_metadata(name: str, value: object) -> tuple[int, ...]:
+    """Read shape and numeric dtype before any user-controlled array conversion."""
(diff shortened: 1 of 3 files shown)


lalalune:
Final residual audit at `270310c` adds the previously missing operation boundaries: exact scan input shape/real-dtype metadata before conversion, complete scan-result scalar/byte preflight before allocation, full-span observation-scale decode product rejection, exact direct wrapper-state shape/dtype admission, saturating outer model counters, and finite-input aggregate diagnostic overflow rollback. Intentional bool reward arrays and legacy tuple/config decoding remain compatible. Verification at this head: 87 focused world-model tests passed; Ruff changed files clean; strict mypy `world_model.py` clean; diff-check clean. Broader world-model/dreaming integration panel is running with no failures recorded so far. Disposition: MERGE once required checks complete; it supersedes and preserves contributor credit from #810.

PR #846, opened 2026-08-16
fix(associative-memory): complete derived contracts

lalalune:
Preserves and rebases the contributor work from #828, then closes the remaining public contracts.

- preflights persistent state, quadratic suffix indices, lookup/gather/adaptive-window work, and batch outputs with exact signed-int32 scalar/byte accounting
- validates exact serialized config/learner envelopes and static state/context/batch shapes and dtypes before tracing
- validates adopted counter domains transactionally and saturates all int32 lifetime counters
- adds allocation-free endpoint/adjacent/product tests plus schema, hostile-state, public-shape, and saturation coverage

Validation: 182 focused tests passed; scoped Ruff, mypy, root import, and diff check passed.

Supersedes #828. Contributor commit and co-authorship are retained.

### DIFF
--- a/alberta_framework/core/associative_memory.py
+++ b/alberta_framework/core/associative_memory.py
@@ -14,7 +14,8 @@ from __future__ import annotations
 import functools
 import math
 import operator
-from dataclasses import asdict, dataclass
+from collections.abc import Mapping
+from dataclasses import asdict, dataclass, fields
 from numbers import Real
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #848, which preserves the contributor commit and includes the full quadratic descriptor/lookup/window/query/update, state/context/counter, scan aggregate, and compatibility repair.

PR #847, opened 2026-08-16
fix(cumulant-discovery): harden cumulant discovery dimensions and maturity threshold with strict integer validation

Kayvan-Zahiri:
## Summary
- Hardens `CumulantDiscovery` constructor parameters (`raw_dim`, `n_candidates`, `maturity_threshold`) with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.

## Verification
- `PYTHONPATH=. pytest tests/test_cumulant_discovery.py`: 20 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/cumulant_discovery.py`: clean

### DIFF
--- a/alberta_framework/core/cumulant_discovery.py
+++ b/alberta_framework/core/cumulant_discovery.py
@@ -54,14 +54,43 @@ Reference:
 from __future__ import annotations
 
 import functools
-from typing import Any, cast
+import operator
+from typing import Any, SupportsIndex, cast
 
 import chex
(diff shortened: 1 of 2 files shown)


lalalune:
Closing in favor of #857, which preserves this contributor commit and completes the float32, derived state-resource, config, key/state/observation, rollback, and counter contracts. Latest-main focused validation is green.

PR #814, opened 2026-08-16
fix(world-model-ensemble): harden ensemble configuration integer and scalar validation

Kayvan-Zahiri:
## Summary
- Hardens `WorldModelEnsembleConfig` integer fields `ensemble_size` and `residual_variance_warmup_steps` with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Validates float scalar fields (`bootstrap_probability`, `residual_variance_decay`, `residual_variance_floor`) with `validated_float32_scalar` and canonicalizes attributes at construction time.

## Verification
- `PYTHONPATH=. pytest tests/test_world_model_ensemble.py tests/test_prototype_world_model_ensemble.py`: 40 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/world_model_ensemble.py`: clean

PR #849, opened 2026-08-16
fix(world-model-ensemble): validate scalar and resource contracts

lalalune:
Supersedes #814 while preserving Kayvan Zahiri's contribution with attribution and removing unrelated formatting churn.

The contributor's exact Python/NumPy integer-family and float32 scalar canonicalization is retained and completed. The replacement additionally validates exact nested config types, bounds `ensemble_size + 1` before `jax.random.split`, preflights member, residual-grid, aggregate persistent scalar, and mixed-dtype byte counts before allocation, and rejects host-valid values that underflow/round onto excluded float32 endpoints.

It also fixes a current-main resource-accounting regression exposed during review: multi-head member states now contain uint32 lifetime words, but `resource_budget()` still asserted that every persistent uint32 belonged to the two bootstrap keys. The accounting now separates member lifetime words from the two key streams while keeping the bootstrap-specific fields exact.

Compatibility is unchanged: the existing ensemble/config serialization envelopes and nested config decoders are preserved; this does not add broader exact-schema restrictions.

Validation:
- 49 of 51 ensemble + Prototype ensemble tests passed in the broad run; the only two failures were test-helper argument omissions, then fixed
- 21 config/resource tests passed after the fix and current-main rebase
- Ruff clean
- mypy clean on `world_model_ensemble.py`
- `git diff --check` clean

This is development-only L0 mechanism/resource hardening. It does not authorize compute, promote evidence, or alter a registered five-claim source inventory.

### DIFF
--- a/alberta_framework/core/world_model_ensemble.py
+++ b/alberta_framework/core/world_model_ensemble.py
@@ -39,17 +39,19 @@ import dataclasses
 import functools
 import hashlib
 import json
-import math
+import operator
 from pathlib import Path
-from typing import Any, cast
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 from alberta_framework.core.checkpoints import (
     load_checkpoint,
     load_checkpoint_metadata,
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #856. The integrated replacement retains #849’s sound `ensemble_size + 1` split bound, exact full-member aggregate resource formula, and lifetime-word accounting, while also preserving byt61’s #836 authorship and preventing hostile scalar hooks.

PR #857, opened 2026-08-17
fix(cumulant-discovery): preserve scalar and loader contracts

lalalune:
Narrow residual after merged #860 and #865. Keeps their exact 256 MiB resource preflight, typed Threefry key, state/input shape+dtypes, transactional nonfinite handling, saturation, and contributor work. Adds only the remaining contracts: concrete host real types without hostile subclass hooks; exact-nonzero float32 underflow rejection for zero-legal replacement_rate/gamma; and the historical unversioned Mapping/partial-default/type-marker loader behavior instead of a breaking exact-schema change.\n\nVerification: 51 full cumulant-discovery tests passed; scoped Ruff, strict mypy, and diff-check passed.

### DIFF
--- a/alberta_framework/core/cumulant_discovery.py
+++ b/alberta_framework/core/cumulant_discovery.py
@@ -55,6 +55,7 @@ from __future__ import annotations
 
 import functools
 import operator
+from collections.abc import Mapping
 from typing import Any, SupportsIndex, cast
 
 import chex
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #860 plus runtime corrective #865. This retains #857's valid state/key/observation gates while keeping the exact resource count, 256 MiB cap, one-call scalar contract, and strict schema.

PR #861, opened 2026-08-17
fix(option-duration): validate host dimensions

### LINKED ISSUE #858: OptionValueDurationLearner accepts noncanonical host dimensions
## Summary

`OptionValueDurationLearner` accepts noncanonical host values for `n_options` and `feature_dim`. The constructor and shape methods rely on loose comparisons, while `from_config()` first applies `int(...)`, so booleans and integral-looking floats are accepted and serialized values can be silently truncated or coerced.

The current behavior includes:

- `n_options=True`, `n_options=1.5`, and `n_options=np.float64(2.0)` construct successfully;
- `from_config()` silently converts values such as `1.5` or `"2"` through `int(...)`;
- `feature_dim=True`, `feature_dim=1.5`, and `feature_dim=np.float64(2.0)` reach parameter counting and initialization;
- dimensions above signed-int32 capacity are accepted even though the learner's JAX shapes and counters use an int32 resource domain.

Malformed strings and `None` also escape the intended validation contract as incidental `TypeError`s rather than a bounded field-specific `ValueError`.

## Expected behavior

Require actual built-in or NumPy integer scalar types for both dimensions, reject booleans and coercible non-integers, canonicalize accepted values through `operator.index`, and bound them to `[1, 2147483647]`. `from_config()` should pass the raw payload through the same constructor boundary instead of coercing it first.

## Scope

- `alberta_framework/core/option_value_duration.py`
- `tests/test_option_value_duration.py`

This is host-boundary and resource-domain validation only. It does not change the TD equations, defaults, RNG behavior, evidence sources, scientific claims, benchmarks, or any `outputs/**` artifact.

A complete live ownership scan immediately before publication covered all **11 open ASI PRs / 23 changed-file rows** and **2 open issues**, with no overlap or matching claim on either target path or this validation boundary.

## Contribution provenance

- AI assistance: yes
- AI provider/model: OpenAI / gpt-5.6-sol
- Client / agent tooling: Claude Code CLI Proxy

— [sol-option-duration-host-dimensions]

ss251:
Closes #858.

## What changed

`OptionValueDurationLearner` now applies the repository's strict host-integer contract to `n_options` and `feature_dim`:

- accepts only actual built-in or NumPy integer scalar types;
- rejects booleans, floats, numeric strings, `None`, and other coercible non-integers;
- canonicalizes accepted values through `operator.index`;
- bounds both dimensions to the positive signed-int32 domain `[1, 2147483647]`;
- passes raw `from_config()` payloads through the constructor instead of truncating them with `int(...)`;
- uses the validated canonical feature dimension for parameter counts and JAX array shapes.

Focused table-driven regressions cover direct construction, deserialization, parameter counting, initialization, and the int32 upper bound.

## Reproduction

Before the production change, the focused tests-first run produced **14 failures / 44 passes**. It demonstrated that booleans and floats were accepted as dimensions, oversized values were unbounded, serialized values were silently coerced, and malformed values escaped as incidental type errors.

## Verification

Base: `82ff5aff0402922937396836d976e86034d6037d`  
Exact head: `9bfa12c18092b871daec2ce83a931a11a70f86d0`

- tests-first focused suite: **14 failed / 44 passed**;
- focused suite after repair and after final rebase: **58 passed**;
- focused Ruff: **all checks passed**;
- full strict mypy: **no issues in 168 source files**;
- `git diff --check`: clean;
- exact-head factory proof: recorded green for `9bfa12c18`.

The repository fast lane currently stops during collection on an unrelated current-base contract mismatch: `tests/test_step12_production.py` constructs `SubtaskSpec(threshold=1.0e-50)`, while `options.py` now correctly rejects values that underflow to zero at the float32 sink. Neither file is touched here.

## Ownership and scientific integrity

A complete live scan immediately before publication covered all **11 open ASI PRs / 23 changed-file rows** and **2 open issues**, with zero overlap or matching claim on `alberta_framework/core/option_value_duration.py`, `tests/test_option_value_duration.py`, or this host-dimension boundary.

No `outputs/**` file, frozen protocol, benchmark threshold, seed, scientific claim, or registered evidence source changed. This patch is limited to one production file and one focused test file.

- [ ] N/A - no shareable model trajectory artifact was generated; verification commands and exact-head proof are listed above

## Contribution provenance

- AI assistance: yes
- AI provider/model: OpenAI / gpt-5.6-sol
- Client / agent tooling: Claude Code CLI Proxy

— [sol-option-duration-host-dimensions]

### DIFF
--- a/alberta_framework/core/option_value_duration.py
+++ b/alberta_framework/core/option_value_duration.py
@@ -27,9 +27,10 @@ from __future__ import annotations
 
 import dataclasses
 import functools
+import operator
 import time
 from numbers import Real
-from typing import Any
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
@@ -44,6 +45,31 @@ REWARD_HEAD = 0
 DURATION_HEAD = 1
 N_HEADS = 2
 _INT32_MAX = 2_147_483_647
+_ACTUAL_INT_TYPES = frozenset(
+    {
+        int,
+        np.int8,
+        np.int16,
+        np.int32,
+        np.int64,
+        np.uint8,
+        np.uint16,
+        np.uint32,
+        np.uint64,
+        np.longlong,
+        np.ulonglong,
+    }
+)
+
+
+def _require_int32(name: str, value: object, *, minimum: int = 1) -> int:
+    """Validate one actual host integer in the signed-int32 domain."""
+    if type(value) not in _ACTUAL_INT_TYPES:
+        raise ValueError(f"{name} must be an integer in [{minimum}, {_INT32_MAX}]")
+    canonical = operator.index(cast(SupportsIndex, value))
+    if not minimum <= canonical <= _INT32_MAX:
+        raise ValueError(f"{name} must be an integer in [{minimum}, {_INT32_MAX}]")
+    return canonical
 
 
 def _require_float32_real(
@@ -227,9 +253,7 @@ class OptionValueDurationLearner:
         config: OptionValueDurationConfig | None = None,
     ):
         """Create a fixed-capacity option predictor."""
-        if n_options < 1:
-            raise ValueError("n_options must be positive")
-        self._n_options = n_options
+        self._n_options = _require_int32("n_options", n_options)
         self._config = config or OptionValueDurationConfig()
 
     @property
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by contributor-preserving #874, which retains the host-dimension fix and completes scalar, schema, state, runner, and persistent-resource contracts.

PR #844, opened 2026-08-16
fix(off-policy-td): harden off-policy TD learners feature dimension initialization with strict integer validation

Kayvan-Zahiri:
## Summary
- Hardens `OffPolicyTDLinearLearner`, `ETDLinearLearner`, and `GradientTDLinearLearner` state initialization (`feature_dim`) with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.

## Verification
- `PYTHONPATH=. pytest tests/test_off_policy_td.py`: 33 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/off_policy_td.py`: clean

PR #863, opened 2026-08-17
fix(off-policy-td): validate scalar and state resource contracts

lalalune:
Supersedes #844 while preserving Kayvan Zahiri's feature-dimension validation contribution (`Co-authored-by` is retained in the commit).

This clean replacement removes unrelated formatting churn and completes the public configuration/resource contract for all three off-policy TD learners:

- accepts only actual Python/NumPy integer scalar families for `feature_dim`, rejects bools, subclasses, arbitrary index hooks, and invalid signed-int32 values without invoking hostile `repr`;
- preflights each learner's exact persistent JAX state bytes, including Gradient-TD's augmented bias coordinate, before allocation;
- validates every float32-consumed constructor scalar in both its exact host and narrowed binary32 domains;
- preserves the documented explicit positive-infinity sentinel for ratio clipping, while rejecting negative infinity and finite values that overflow binary32;
- canonicalizes supported non-built-in real scalars and normalizes ordinary conversion exceptions to `ValueError`;
- leaves update/JIT behavior and serialization surfaces unchanged.

Validation on latest `main`:

- `python -m pytest tests/test_off_policy_td.py -q` — 65 passed
- `python -m ruff check alberta_framework/core/off_policy_td.py tests/test_off_policy_td.py`
- `python -m mypy` — 168 source files clean
- `git diff --check origin/main...HEAD`

This is library/L0 contract hardening only. It does not execute benchmarks, modify immutable outputs, authorize compute, or affect a source registered by the five-claim evidence registry.

### DIFF
--- a/alberta_framework/core/off_policy_td.py
+++ b/alberta_framework/core/off_policy_td.py
@@ -51,19 +51,76 @@ Reference:
 from __future__ import annotations
 
 import functools
+import math
+import operator
 import time
-from typing import Any
+from numbers import Real
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 from alberta_framework.core.types import Observation
 from alberta_framework.core.update_safety import (
     floating_tree_is_finite as _floating_tree_is_finite,
 )
+
+_INT32_MAX = 2**31 - 1
+_ACTUAL_INT_TYPES = frozenset(
+    {
+        int,
+        np.int8,
+        np.int16,
+        np.int32,
+        np.int64,
+        np.uint8,
+        np.uint16,
+        np.uint32,
+        np.uint64,
+        np.longlong,
+        np.ulonglong,
+    }
+)
+_ACTUAL_FLOAT_TYPES = frozenset(
+    {float, np.float16, np.float32, np.float64, np.longdouble}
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking residual from independent audit: current main does not have the claimed complete runtime contracts, and this PR explicitly leaves them unchanged. All three public predict/update methods are jitted entrypoints with no exact state leaf shape/dtype/type validation and no observation/next-observation/scalar shape/dtype host preflight; GradientTD `_augment` converts before validation. All three counters still use `step_count + 1` and wrap at int32 max. Finite-large values can overflow `td_error**2` diagnostics; candidate overflow does reject atomically, but the public structural and lifetime contracts are incomplete. Split host-validation wrappers from private jitted kernels, validate state/operands before JAX conversion, saturate counters, and add malformed/hostile/full-dtype/f32-max regressions before merge.

lalalune:
Closing as fully superseded by merged #875 (e0146eb6), which preserves Kayvan Zahiri's original TD contribution and completes the exact scalar/resource work here plus the previously blocking runtime residuals: public state/operand shape-dtype contracts, pre-conversion validation, saturating counters, exact schemas, and derived aggregate allocation preflights. Current main therefore contains a strict superset of this old-base branch; rebasing #863 would only recreate conflicting duplicate changes.

lalalune:
Superseded in two stages: merged #875 preserves/integrates this constructor, schema, and resource work; #879 is the narrow corrective residual for public host/JIT boundaries, state/operand contracts, saturating counters, transactional diagnostics, and scan resources.

PR #864, opened 2026-08-17
fix(resource-manager): complete exact resource-manager contracts

lalalune:
Supersedes #840 while preserving all three commits, including Kayvan’s original author commit.

The #840 implementation correctly covers integer/float32 sinks, derived persistent and operation workspaces, runtime shapes/state, counter saturation, and hostile metadata. This successor adds the remaining exact serialized-schema gates: exact discriminator and field set, built-in JSON scalar types, exact emitted list containers/elements, while retaining Mapping compatibility and normalized hostile mapping failures.

Validation:
- 47 resource-manager tests passed
- 52 compositional consumer tests passed
- Ruff passed
- strict mypy passed

No benchmarks or outputs were run or modified.

### DIFF
--- a/alberta_framework/core/resource_manager.py
+++ b/alberta_framework/core/resource_manager.py
@@ -116,18 +116,31 @@ def _read_mapping(name: str, value: object) -> dict[str, Any]:
 
 
 def _require_manager_state_budget(name: str, n_contexts: int, n_choices: int) -> None:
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded after #840 merged independently; the rebased schema-only correction is the new PR.

PR #821, opened 2026-08-16
fix(compositional-features): harden compositional feature learner integer parameters with strict validation

Kayvan-Zahiri:
## Summary
- Hardens `FiniteCandidateSelector` (`n_candidates`) and `CompositionalFeatureLearner` (`n_features`, `n_tasks`, `candidate_count`, `replacement_interval`, `min_feature_age`, `candidate_min_age`, `max_depth`, `signed_tanh_scaffold_count`, `retention_tanh_min_count`, `retention_product_min_count`, `generator_resource_contexts`) integer parameters with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Canonicalizes integer attributes to Python `int` at construction time.

## Verification
- `PYTHONPATH=. pytest tests/test_compositional_features.py tests/test_compositional_alias.py`: 49 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/compositional_features.py`: clean

PR #866, opened 2026-08-17
fix(compositional-features): close scalar safety residuals

lalalune:
Residual follow-up to the integrated #821 integer work and subsequent derived-resource/schema repair. Those changes correctly cover exact integers, persistent state products, base recursive-pair resources, booleans, and serialized schemas, so this PR retains only missing contracts:\n\n- validate and canonicalize every float32-consumed selector/learner field in both exact-host and narrowed float32 domains, containing hostile ratio/repr hooks;\n- require exact built-in string discriminators and exact constructor tuples while preserving exact JSON-list deserialization;\n- validate operation-prior and generator-preference elements and float32 aggregate mass;\n- preflight the larger signed-tanh scaffold table, not just the base recursive pair table;\n- avoid the quadratic preflight when n_features == feature_dim because no composed slot exists and no parent table is constructed.\n\nHistorical zero learning rates and zero promotion margin remain accepted. Contributor credit from #821 remains in main and is also recorded on this commit.\n\nValidation: 68 compositional/alias/residual tests passed; Ruff, targeted strict mypy, and diff-check passed. compositional_features.py is not a registered evidence source; no outputs or claims were changed.

### DIFF
--- a/alberta_framework/core/compositional_features.py
+++ b/alberta_framework/core/compositional_features.py
@@ -36,6 +36,7 @@ import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int, PRNGKeyArray
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 from alberta_framework.core.future_utility import (
     bias_correct_future_utility,
     canonical_float32_ema_decay,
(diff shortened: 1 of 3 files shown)


lalalune:
Superseded by merged #873, preserving Kayvan co-authorship and adding the missing nonzero-underflow gates.

PR #841, opened 2026-08-16
fix(off-policy-horde): harden nonlinear shared GTD horde hidden size and feature dimension with strict integer validation

Kayvan-Zahiri:
## Summary
- Hardens `NonlinearSharedGTDHordeLearner` hidden layer dimension (`hidden_size`) and state initialization (`feature_dim`) with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.

## Verification
- `PYTHONPATH=. pytest tests/test_off_policy_horde.py tests/test_offpolicy_horde_rho.py`: 26 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/off_policy_horde.py`: clean

PR #872, opened 2026-08-17
fix(off-policy-horde): complete shared GTD contracts

lalalune:
Preserves and rebases the contributor work from #841, then completes the nonlinear shared GTD Horde contract.

- validates exact Python/NumPy integers and all float32-consumed scalar domains without hostile repr exposure
- validates HordeSpec identity, demon count, gamma/lambda shapes, dtypes, finiteness, and domains
- preflights the exact D/H/F primary-plus-secondary state scalar and byte product before JAX allocation
- validates adopted state and public predict/update shapes and dtypes before tracing
- rejects negative ratios and discounts outside [0,1], while allowing terminal heads to ignore nonfinite next observations
- canonicalizes telemetry leaves for exact rollback and saturates the signed-int32 step counter
- adds allocation-free boundaries plus hostile, spoof, state, JIT, domain, terminal, rollback, and saturation tests

Validation: 60 focused tests passed; scoped Ruff, mypy, root import, and diff check passed.

Supersedes #841. Contributor commit and co-authorship are retained.

### DIFF
--- a/alberta_framework/core/off_policy_horde.py
+++ b/alberta_framework/core/off_policy_horde.py
@@ -43,6 +43,7 @@ from alberta_framework.core.update_safety import (
 )
 
 _INT32_MAX = 2**31 - 1
+_FLOAT32_MIN_NORMAL = float.fromhex("0x1.0p-126")
 _ACTUAL_INT_TYPES = frozenset(
     {
         int,
@@ -197,8 +198,8 @@ class NonlinearSharedGTDHordeState:
     secondary_head_w: Float[Array, "n_demons hidden_dim"]
     secondary_head_b: Float[Array, " n_demons"]
     step_count: Float[Array, ""]
-    birth_timestamp: float = 0.0
-    uptime_s: float = 0.0
+    birth_timestamp: Float[Array, ""]
+    uptime_s: Float[Array, ""]
 
 
 @chex.dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking audit residuals (auto-merge is disabled):

1. `init()` accepts any float32-finite `init_scale`, then multiplies it by an unbounded normal draw. At `float32.max`, deterministic keys can produce nonfinite `trunk_w`/`head_w`; init returns the invalid state without a finiteness gate. Validate the typed Threefry key and fail closed on every generated state leaf (or use a proven bounded initializer), with a deterministic max-scale regression.
2. `correction_norms` and `secondary_norms` use ordinary vdot/square/sum/sqrt reductions. Finite large state/updates can overflow these diagnostics while `candidate_state_finite` remains true, so `update_applied=True` can accompany `inf` public metrics. Use scale-normalized stable norms and include every public output/diagnostic in validity.
3. On a rejected update, the state rolls back but predictions/targets/errors/ratios/norms are returned unneutralized; malformed finite-large inputs can therefore expose nonfinite outputs with `update_applied=False`. Neutralize the complete result atomically, preserving the intentional terminal next-prediction behavior.

The exact-head focused suite passes independently (47 tests), so these are missing adversarial branches rather than existing failures. Add max-scale init, finite-large diagnostic, and invalid-result neutralization regressions, then rerun the full Horde panel plus Ruff/mypy/diff before merge.

lalalune:
Superseded by merged #875 for the contributor/config/state/public-boundary work and by #880 for the audited nonlinear init-scale, stable-norm, output-validity, and atomic-neutralization residual.

PR #852, opened 2026-08-16
fix(actor-critic): harden actor-critic configuration dimensions and scalar hyperparameters with strict validation

Kayvan-Zahiri:
## Summary
- Hardens `ActorCriticConfig` and `ContinuousActorCriticConfig` dimension parameters (`n_actions`, `action_dim`) and state initialization (`feature_dim`) with `_require_int32` to reject booleans and non-integer floats while accepting and canonicalizing the full NumPy integer family.
- Validates float hyperparameters (`gamma`, `actor_step_size`, `critic_step_size`, `actor_lamda`, `critic_lamda`, `temperature`, `log_sigma_init`, `log_sigma_min`, `log_sigma_max`) with `validated_float32_scalar`.

## Verification
- `PYTHONPATH=. pytest tests/test_actor_critic.py tests/test_continuous_actor_critic.py`: 44 passed
- `PYTHONPATH=. pytest tests/test_horde_actor_critic.py tests/test_recurrent_trace_actor_critic.py`: 179 passed
- `ruff check .`: clean
- `mypy alberta_framework/core/actor_critic.py`: clean

PR #877, opened 2026-08-17
fix(actor-critic): complete config and resource contracts

lalalune:
Supersedes #852 while preserving its contributor commit and attribution.\n\nCompletes the actor-critic boundary: exact Python/NumPy integer and binary32 scalar validation, canonical storage, legacy Mapping/list/tuple config compatibility, exact discrete and continuous state scalar/byte preflights, runtime feature dimensions, adopted-state/key/input/scan shape and dtype contracts, and saturating lifetime counters. It preserves the historically valid one-action discrete contract and keeps continuous action-bound validation at agent construction.\n\nValidation: 248 actor-critic/recurrent/horde tests passed before the latest-main rebase; 69 focused actor-critic tests passed after rebase; scoped Ruff, strict mypy, and diff-check pass. No outputs or registered evidence sources changed.

### DIFF
--- a/alberta_framework/core/actor_critic.py
+++ b/alberta_framework/core/actor_critic.py
@@ -21,21 +21,50 @@ from __future__ import annotations
 import dataclasses
 import functools
 import math
+import operator
 from numbers import Real
-from typing import Any
+from typing import Any, SupportsIndex, cast
 
 import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float, Int
 
 from alberta_framework._float32 import round_real_to_float32
+from alberta_framework.core._float32_scalars import validated_float32_scalar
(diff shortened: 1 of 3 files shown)


lalalune:
Deep-audit residual patch is now at `b8450de`. It corrects the exact aggregate state scalar formulas (including the two-word typed Threefry key), enforces the historical optional type marker while rejecting wrong markers/unknown outer fields, preflights scan shapes and result budgets before JAX conversion, and makes the public bound metric average overflow-safe. Added six regressions covering both discrete and continuous paths, hostile oversized array stubs, schemas, and finite max-float bounder metrics. Verification: `75 passed` across both actor-critic suites plus residuals; residual rerun `6 passed`; Ruff clean; targeted strict mypy clean; `git diff --check` clean. Robot-facing top-level/core import assertion remains covered in `test_actor_critic_exports`; exported surface is unchanged.

PR #882, opened 2026-08-17
fix(canonical-upgd): complete config, tree, and resource contracts

lalalune:
Supersedes #851 while preserving byt61 original commit/authorship. Completes exact host+float32 config gates (including rejecting near-one decay that narrows to 1.0 and nonzero underflow), refuses arbitrary Real subclasses before ratio hooks, preserves exact Mapping schemas, and preflights complete parameter/state/update resources and metadata before conversion. Validation: 147 focused tests plus 53 validation rerun; Ruff and full mypy clean. No benchmark or output changes.

### DIFF
--- a/alberta_framework/core/canonical_upgd.py
+++ b/alberta_framework/core/canonical_upgd.py
@@ -52,7 +52,6 @@ Reference:
 
 from __future__ import annotations
 
-import math
 from dataclasses import dataclass
 from typing import Any, Literal
 
@@ -60,8 +59,11 @@ import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Int
+
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 
 UPGDMode = Literal["protecting", "non_protecting"]
 UPGDNormalization = Literal["global", "local"]
@@ -97,6 +99,30 @@ _RAW_GLOBAL_PROFILES = frozenset(
     }
 )
 _INT32_MAX = 2_147_483_647
+_UINT32_MAX = 4_294_967_295
+_ACTUAL_INT_TYPES: tuple[type, ...] = (
+    int,
+    np.int8,
+    np.int16,
+    np.int32,
+    np.int64,
+    np.uint8,
+    np.uint16,
+    np.uint32,
+    np.uint64,
+    np.longlong,
+    np.ulonglong,
+)
+
+
+def _require_float32_resource(
+    name: str, *, vector_scalars: int, fixed_scalars: int = 0
+) -> None:
+    total_scalars = vector_scalars + fixed_scalars
+    if total_scalars > _INT32_MAX:
+        raise ValueError(f"{name} scalar count must fit signed int32")
+    if 4 * total_scalars > _INT32_MAX:
+        raise ValueError(f"{name} byte count must fit signed int32")
 
 
 def _static_zero_scale(scale: float, value: Array) -> Array:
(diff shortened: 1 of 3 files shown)


lalalune:
Independent deep audit found and repaired the remaining gap relative to closed #885. Head `ec82efc` now retains #882 exact measured persistent/result formulas and adds #885-equivalent conservative transient working-set bounds (20 Canonical working copies; 30 adaptive/official), checked before conversion in scalar and byte domains. It also makes safe-extended Canonical operation failures atomic with finite neutral diagnostics, makes Alberta diagnostic means pre-scaled/stable and includes corrected/denominator/gate/perturbation plus float64-to-float32 global/mean sinks in the accepted gate, and saturates the official int32 clock without changing its source-parity arithmetic. Source profiles retain their documented undefined/nonfinite equation behavior. Added working-boundary, safe overflow rollback, x64 narrowing rollback, and official saturation regressions. Verification on the rebased branch: all three transform suites `150 passed`; validation rerun `56 passed`; Ruff clean; full strict mypy `168 source files` clean; diff-check clean. No outputs/evidence changed.

kenjohnscreates (approved):
## Independent verification — APPROVE (measured-path graph invariance confirmed by paired probe)

Verified by re-applying the diff onto current `main` (`d1c8e7b`) in a scratch worktree; all three files **byte-identical to PR head `3387d8bb`** (staged-diff check, one file is new).

**The load-bearing check for this module: the compiled step is unchanged.** `CanonicalUPGD` feeds `benchmarks/upgd_ipmnist.py` (pinned nonpromoting artifacts, and the `lean_upgd_w_update` parity contract), so a +472/−233 rewrite here needs a graph-level answer, not just a green suite. Paired probe run from inside the worktree — same path both rows, StableHLO text (location-free) + result bytes — across three configurations:

| config | StableHLO (base → PR) | result bytes (base → PR) |
|---|---|---|
| `utility_decay=0.999, noise_std=0.0` | `b7be3d46…` → `b7be3d46…` | `499bf1c4…` → `499bf1c4…` |
| `utility_decay=0.999, noise_std=0.001` | `dc3c4d54…` → `dc3c4d54…` | `3d28b5cd…` → `3d28b5cd…` |
| `profile="official_experiment_global", mode="protecting"` | `67176097…` → `67176097…` | `8e60eb23…` → `8e60eb23…` |

All six pairs identical. Every new check runs host-side before JAX conversion/allocation (metadata reads, exact-byte preflights, host/binary32 domain validation), and none of it enters the traced `update` — exactly what the PR claims. The `upgd_ipmnist` suite (including the supplied-noise canonical parity test) passes unchanged, which independently confirms the equation contract with the benchmark's lean restatement is intact.

**Battery (arm64 M1):** `test_canonical_upgd.py` + `test_canonical_adaupgd.py` + `test_canonical_upgd_validation.py` + `test_upgd_ipmnist.py`: **256/256**; `ruff check .` clean; repo-config strict mypy clean at **168 source files**. Failing-test-first: tests-only against baseline → **29 of 53** in the new validation suite fail (the remaining 24 are boundary/legal-input pins that pass on both sides — regression protection rather than red proof); full diff → green.

**Read-through of the contract surface (structured-partial — I did not line-review every helper in the 472 added lines):** the resource preflights are computed from shape/dtype metadata without materializing arrays (the tests prove legal/illegal boundaries without constructing anything huge — the right discipline for a check meant to precede allocation); the three Alberta diagnostic reductions that promoted to int64 under x64 are the kind of silent-promotion bug the x64 parametrized tests would have kept catching; hostile mapping/constructor failures normalize to `ValueError` while historical Mapping deserialization and dict/MappingProxy round-trips are preserved by explicit tests. byt61's float32 scalar work is retained with their authorship as the first commit, as stated.

**Evidence-boundary statement verified:** `core/canonical_upgd.py` is not a registered source of any of the five claims (`evidence_manifest.py`); the PR correctly discloses that historical nonpromoting UPGD-IPMNIST receipts name it and that new development comparisons must bind the new source rather than reuse those receipts — the byte-invariant graph above is what makes that a bookkeeping matter rather than a scientific one.
Compute receipt: 2773674 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

PR #883, opened 2026-08-17
fix(upgd-memory): complete combined learner contracts

lalalune:
Supersedes #833 while preserving byt61 original commit/authorship. Completes exact hostile-safe scalar/container/Mapping schemas; full nested UPGD + PrototypeMemory + wrapper scalar/byte accounting; 256 MiB persistent/scan bounds; positive-normal log/division sinks; typed Threefry and static state/input metadata gates; and counter/transaction safety. Validation: 209 focused tests; Ruff and full mypy clean. No benchmark or output changes.

### DIFF
--- a/alberta_framework/core/upgd_memory.py
+++ b/alberta_framework/core/upgd_memory.py
@@ -14,8 +14,9 @@ rather than a route-selecting portfolio.
 from __future__ import annotations
 
 import functools
+import operator
 from dataclasses import asdict, dataclass
-from typing import Any, Literal, cast
+from typing import Any, Literal, SupportsIndex, cast
 
 import chex
 import jax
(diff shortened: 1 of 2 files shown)


lalalune:
Independent final audit found and repaired two residual operation-boundary gaps atop the compatibility-preserving head: (1) float32 `exp(log(x))` underflows to zero at `tiny`, so log-backed novelty thresholds now require the exact tested round-trip endpoint `2 * tiny` while division-only `memory_bandwidth` still accepts `tiny`; (2) scan preflight now accounts for host inputs, all outputs, booleans, and two complete carry/state trees before JAX execution. Public predict/update/scan validation remains host-side and hostile-metadata-safe; historical Mapping, partial/default, arbitrary-marker, list/tuple, and NumPy-scalar loader compatibility is preserved. Exact final head `66611c9`: 256 affected tests passed; Ruff clean; strict mypy clean (168 files); `git diff --check` clean.

PR #885, opened 2026-08-17
fix(canonical-upgd): complete config and runtime resource contracts

lalalune:
Supersedes #851 while preserving its scalar-validation commit. Retains hostile-safe float32 canonicalization for CanonicalUPGD, AlbertaAdaUPGD, and OfficialAdaUPGD, then connects the formerly test-only resource work to production: every init/update validates concrete host leaf metadata before JAX materialization; aggregate parameter, persistent-state, result, and conservative worst-case working trees are bounded in signed-int32 scalar/byte domains; adaptive and official state leaves are exact shape/dtype matched without coercion; and complete Mapping schemas normalize unreadable hooks and constructor failures without hostile repr. Adds production oversized non-conversion, hostile state metadata, and Mapping regressions. Verification on rebased latest main: 133 full canonical/Ada tests passed; post-rebase 39 validation tests passed; Ruff clean; strict mypy 168 files clean; diff-check clean. No outputs or registered evidence source changed.

### DIFF
--- a/alberta_framework/core/canonical_upgd.py
+++ b/alberta_framework/core/canonical_upgd.py
@@ -52,7 +52,6 @@ Reference:
 
 from __future__ import annotations
 
-import math
 from dataclasses import dataclass
 from typing import Any, Literal
 
@@ -60,8 +59,11 @@ import chex
 import jax
 import jax.numpy as jnp
 import jax.random as jr
+import numpy as np
 from jax import Array
 from jaxtyping import Int
+
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 
 UPGDMode = Literal["protecting", "non_protecting"]
 UPGDNormalization = Literal["global", "local"]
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as the weaker duplicate of #882. Both preserve byt61’s authored commit and wire production resource checks, but #882 additionally validates gradient/mask/noise metadata before JAX conversion; preserves Python-float parameter leaves; canonicalizes host dtype to the actual JAX dtype; enforces the typed Threefry result-size contract; rejects nonzero underflow on zero-inclusive float32 fields; validates exact measured state/result formulas across float16, float32, and x64 float64; and fixes the x64 promotion of three Alberta diagnostic counters. #885 still calls jnp.asarray directly on hostile gradient/mask/noise leaves, rejects historical Python scalar leaves, counts raw host dtype rather than actual JAX dtype, and uses unexplained 20/30-copy aggregate estimates rather than measured result formulas. The shared test-plan failure is the independent current-main collection regression addressed by #881, not a difference between these successors.

PR #886, opened 2026-08-17
fix(world-model): retain frozen zero-step endpoint

lalalune:
Repairs the current-main contract regression exposed when #881 restored complete test collection. Step 9 intentionally permits `model_step_size == 0` to freeze model learning, but #837 made both nested core world-model configs reject that endpoint.

- restore nonnegative step-size validation in direct and action-conditioned world models
- retain rejection of negative values
- add explicit frozen-model endpoint coverage

Validation:
- exact three CI failures pass on the repaired contract
- 275 complete world-model / action-conditioned / Step 9 / discount tests pass
- Ruff and mypy pass
- diff check passes


lalalune:
#887 is the stronger duplicate: it retains the exact-zero frozen-learning endpoint while also rejecting positive host values that silently underflow to float32 zero, with direct coverage on both world-model configs and a broader Step 8/Step 9 validation panel. Closing #886 as superseded by #887.

PR #890, opened 2026-08-17
fix(actor-critic): preflight complete scan working sets

lalalune:
Narrow corrective follow-up to merged #888 (and #889). #888 correctly restored trusted metadata checks and bounded emitted scan arrays, but its preflight omitted the canonicalized scan inputs and the simultaneous persistent-state + inputs + outputs working set. Large trusted NumPy/JAX shapes could therefore pass the output-only check before multi-gigabyte float32 conversion/allocation.\n\nThis adds exact signed-int32 scalar/byte formulas for both runners before any JAX conversion:\n- discrete inputs: `n * (2*feature_dim + 4)`, outputs: `n * (n_actions + 4)`, plus exact persistent state;\n- continuous inputs: `n * (2*feature_dim + action_dim + 3)`, outputs: `n * (3*action_dim + 3)`, plus exact persistent state.\n\nAllocation-free regressions cover the exact final legal byte endpoints and zero-stride trusted NumPy arrays whose output-only budgets pass but full working sets exceed signed int32; monkeypatched conversion proves rejection occurs first.\n\nVerification after rebasing current main/#889: 79 actor-critic tests passed; scoped Ruff; targeted strict mypy; diff-check. No outputs or evidence sources changed.\n\nCo-authored-by: lalalune <shawmakesmagic@gmail.com>

### DIFF
--- a/alberta_framework/core/actor_critic.py
+++ b/alberta_framework/core/actor_critic.py
@@ -97,14 +97,14 @@ def _serialized_mapping(
 
 
 def _require_discrete_state_resources(n_actions: int, feature_dim: int) -> None:
-    state_scalars = 2 * n_actions * feature_dim + 3 * feature_dim + 2 * n_actions + 5
+    state_scalars = 2 * n_actions * feature_dim + 3 * feature_dim + 2 * n_actions + 6
     state_bytes = 8 * n_actions * feature_dim + 12 * feature_dim + 8 * n_actions + 24
     if state_scalars > _INT32_MAX or state_bytes > _INT32_MAX:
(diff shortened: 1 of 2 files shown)


lalalune:
Deep repair audit complete. The prior head still undercounted the typed Threefry key and only charged two carry trees, omitting the reusable scan-update envelope; optional transition/action metadata was also checked after JAX conversion. Replaced with a latest-main repair that counts exact persistent/input/output resources plus a conservative reusable workspace and validates host metadata before any JAX allocation. Verification: 267 affected actor-critic/Horde/recurrent tests + 18 rebased residual tests passed; package Ruff and strict mypy clean; robot import smoke clean; no registered evidence source/output touched.

PR #891, opened 2026-08-17
fix(compositional-features): complete selector lifetime boundaries

lalalune:
Post-merge corrective for #876. Keeps saturating learner counters and adds exact trusted scalar gates, exact ratio underflow checks, static selector state/loss shape+dtypes, safe EXP3 action bounds with atomic refusal, and zero-hook hostile scalar/state regressions. Validation: 76 compositional tests plus 26 focused rerun; Ruff and full mypy clean. No benchmark or output changes.

### DIFF
--- a/alberta_framework/core/compositional_features.py
+++ b/alberta_framework/core/compositional_features.py
@@ -70,6 +70,9 @@ _ACTUAL_INT_TYPES = frozenset(
         np.ulonglong,
     }
 )
+_ACTUAL_REAL_TYPES = frozenset(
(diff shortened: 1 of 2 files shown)


lalalune:
Correction to the shell-mangled notation above: the reproduced accepted bounds were zero through two-to-the-minus-149, and the returned normalized losses were both NaN while the step clock advanced. The repair requires a finite normal float32 width and verifies both adjacent endpoints. All other audit conclusions remain unchanged.

PR #338, opened 2026-08-16
fix(statistics): require one sample per seed on every scalar-summary surface

Prabhat1308:
Closes #337.

## Issue

`compute_statistics`, `bootstrap_ci`, and `cohens_d` accepted an `(n_seeds, n_steps)` matrix and reported a flattened spread under a per-seed `n_seeds` label — three seeds that agree exactly came back as `std=1.46, sem=0.85, CI [-0.64, 6.64], n_seeds=3` — while the `*_comparison` tests crashed with an opaque `TypeError`.

## New state

`_require_sample_vector` returns the input as a rank-1 array or raises `<name> must be a one-dimensional sample vector (one value per seed), got shape (3, 5); reduce per seed first or use compute_timeseries_statistics`. It is applied at the top of `compute_statistics`, `bootstrap_ci`, `cohens_d`, `ttest_comparison`, `mann_whitney_comparison`, and `wilcoxon_comparison`, before the existing empty/finite/degrees-of-freedom checks. Rank-1 inputs (lists or arrays) are unchanged; `compute_timeseries_statistics` keeps its `(n_seeds, n_steps)` contract.

Failing-test-first: `TestSampleVectorContract` (8 cases across the six surfaces plus 0-D inputs) fails on `main` and passes here.

## Evidence

Falsifier from #337 at this head:

```text
ValueError: values must be a one-dimensional sample vector (one value per seed), got shape (3, 5); reduce per seed first or use compute_timeseries_statistics
```

Verification at `2eb652ed14acfc3400e87e1b27805abb8cabfc8f`:

```text
.venv/bin/python -m pytest tests/test_statistics_validation.py tests/test_experiments_validation.py tests/test_utils_smoke.py tests/test_forager_matched_statistics.py -q -o addopts=""   -> 203 passed
.venv/bin/python -m ruff check .                                                              -> All checks passed!
.venv/bin/python -m mypy                                                                      -> Success: no issues found in 164 source files
```

Composes cleanly with #312 (`git merge-tree` clean; different functions in `statistics.py`). `utils/statistics.py` is not a registered evidence source; no benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, ruff, mypy, falsifier output at the evidence head)

[...]

PR #312, opened 2026-08-16
fix(statistics): require one common final window across compared methods

Prabhat1308:
Closes #311.

## Issue

`pairwise_comparisons` and `get_final_performance` recomputed `final_window = min(window, n_steps)` per method, so methods with different trace lengths were averaged over different horizons whenever `window` exceeded the shortest trace. Two methods with identical settled loss (`1.0`) but different budgets (20 vs 400 steps) came back as `{'short': 3.0, 'long': 1.0}` and a paired t-test with `p=0.0, significant=True`.

## New state

`common_final_window` (statistics.py) derives the one window every method averages and raises when the per-method `min(window, n_steps)` values disagree; both public helpers use it. Equal-length traces keep the documented convention (`test_get_final_performance_window_longer_than_trace_uses_full_trace` unchanged); unequal-length traces with `window` ≤ every trace still average exactly `window` steps each (pinned by two new acceptance tests).

Failing-test-first: `test_unequal_final_windows_rejected[ttest|wilcoxon|mann_whitney]` and `test_get_final_performance_rejects_unequal_final_windows` fail on `main` (`DID NOT RAISE ValueError`) and pass here.

## Evidence

Falsifier from #311 at this head:

```text
ValueError: window=100 exceeds the shortest 'squared_error' trace and the traces differ in length (long: 400 steps, short: 20 steps); every method must average the same number of final steps
window=10 -> {'short': (1.0, 0.0), 'long': (1.0, 0.0)}
```

Verification at `a60da5b6ebc219771a2e40ea970f6c13017538c6`:

```text
.venv/bin/python -m pytest tests/test_statistics_validation.py tests/test_experiments_validation.py tests/test_utils_smoke.py -q -o addopts=""
149 passed
.venv/bin/python -m ruff check .      -> All checks passed!
.venv/bin/python -m mypy              -> Success: no issues found in 164 source files
```

Composes cleanly with open #309 (`git merge-tree` clean; different function in `experiments.py`). No benchmark lane, seeds, or `outputs/` artifacts were touched; no `CHANGELOG.md` edit (maintainer-recorded).

CLI sessions cannot mint GitHub attachment URLs, so the run output above is inline rather than attached.

- [x] logs: inline above (focused pytest, ruff, mypy, falsifier output at the evidence head)

Provider/model/client: anthropic / claude-fable-5 / Claude Code 2.1.233 (contribute-to-asi skill revision 72e4239e).

PR #894, opened 2026-08-17
fix(statistics): reject bool and non-integer publication counts

ss251:
## What changed

Publication-count arguments that are consumed as `range(...)` / final-window lengths are now required to be actual positive integers.

- `bootstrap_ci(..., n_bootstrap=True)` accepted the boolean and returned the degenerate interval `(2.0, 2.333..., 2.333...)` — a one-resample "confidence interval";
- `n_bootstrap=1.0` / `nan` leaked a `TypeError` from `range()` instead of a field-named `ValueError`;
- `pairwise_comparisons(..., window=True)` used the same IEEE-blind `<= 0` check.

`_require_positive_int` binds `n_bootstrap` and `window` the same way `_require_alpha` already binds probabilities: reject `bool` first, then require an integer, then require `> 0`. Existing `0` / `-1` rejections stay.

Legal integer counts (`n_bootstrap=20`, `window=1|100`) and the existing finite-sample / one-sample-per-seed contracts are unchanged.

## Scope

- `alberta_framework/utils/statistics.py`
- `tests/test_statistics_validation.py`

## Why this is unowned

Live occupancy at publish time: 3 open ASI PRs, all Shaw, none of these files. Closed statistics work (`#292` non-finite samples, `#194`/`#199` publication samples, `#410` alpha/p-value, `#312` common final window, `#338` one sample per seed) did not bind bootstrap/window *type identity*. Boolean `True` still minted a one-draw CI.

This is not a republish of `#677`, `#861`, world-model `step_size`, Kayvan `#711`, or HEIR `#4`.

## Verification

Exact candidate head: `b7ce0873cee30337242e3303ebeeb77aec549d34`  

- `pytest tests/test_statistics_validation.py -q -o addopts=""`: **232 passed**
- `mypy alberta_framework/utils/statistics.py`: clean
- `git diff --check`: clean

## Scientific integrity

This is fail-closed host-boundary validation on publication counts only. It does not change estimators, p-value formulas, confidence-level math, correction methods, registered evidence sources, frozen protocols, scientific claims, benchmark results, `CHANGELOG.md`, or any `outputs/**` artifact.

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Cursor

— [cursor-ss251]

### DIFF
--- a/alberta_framework/utils/statistics.py
+++ b/alberta_framework/utils/statistics.py
@@ -114,6 +114,16 @@ def _require_alpha(alpha: object) -> float:
     return _require_probability(alpha, name="alpha", strict=True)
 
 
+def _require_positive_int(name: str, value: object) -> int:
+    """Reject bool/float aliases that ordered comparisons treat as legal counts."""
+    if isinstance(value, bool) or not isinstance(value, (int, np.integer)):
+        raise ValueError(f"{name} must be a positive integer (got {value!r})")
+    count = int(value)
+    if count <= 0:
+        raise ValueError(f"{name} must be a positive integer (got {value})")
+    return count
+
+
 def _require_p_value(value: object, *, name: str) -> float:
     return _require_probability(value, name=name, strict=False)
 
@@ -666,8 +676,7 @@ def pairwise_comparisons(
     from alberta_framework.utils.experiments import AggregatedResults
 
     alpha_value = _require_alpha(alpha)
-    if window <= 0:
-        raise ValueError(f"window must be positive (got {window})")
+    window = _require_positive_int("window", window)
 
     names = list(results.keys())
     n = len(names)
@@ -817,8 +826,7 @@ def bootstrap_ci(
             f"statistic must be either 'mean' or 'median' (got {statistic!r})"
         )
     _validate_confidence_level(confidence_level)
-    if n_bootstrap <= 0:
-        raise ValueError(f"n_bootstrap must be positive (got {n_bootstrap})")
+    n_bootstrap = _require_positive_int("n_bootstrap", n_bootstrap)
     rng = np.random.default_rng(seed)
 
     stat_func = np.mean if statistic == "mean" else np.median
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):

Verified by re-applying the diff onto current `main` (`90c4613`) in a scratch worktree; both files **byte-identical to PR head `b7ce0873`**.

**Defect reproduced directly on baseline** (not just via the tests): `bootstrap_ci([1.0, 2.0, 3.0], n_bootstrap=True, seed=0)` returns `(2.0, 2.3333…, 2.3333…)` — a one-resample "confidence interval" whose bounds coincide, silently minted because `True <= 0` is `False` and `range(True)` is `range(1)`; `n_bootstrap=1.0` leaks `TypeError: 'float' object cannot be interpreted as an integer` from `range()` instead of a field-named `ValueError`. Both are exactly as described.

**Fix is minimal and matches the module's own conventions.** `_require_positive_int` sits beside `_require_alpha`/`_require_p_value` with the same shape (bool rejected first — `bool` is an `int` subclass so the order matters — then integer-family, then `> 0`), accepts `np.integer` so NumPy-sourced counts keep working, and returns the canonical `int`. Both call sites (`pairwise_comparisons.window`, `bootstrap_ci.n_bootstrap`) preserve their existing `0`/`-1` rejections through the same path. The parametrized cases (`True`, `False`, `1.0`, `nan`, `inf`) cover the exact IEEE-blind gap, under `warnings.simplefilter("error")` consistent with the file's discipline.

**Battery (arm64 M1):** `tests/test_statistics_validation.py` **232/232**; `ruff check .` clean; repo-config strict mypy clean at **168 source files**. Failing-test-first reproduced: tests-only against baseline → all **10** new parametrized cases fail; full diff → green.

**Ownership check** (I fortified this module in #41 and reviewed the #27–#33/#54/#57 statistics wave, so I looked at this specifically): the closed statistics work bound finiteness, sample counts, alpha/p-value probability domains, and final-window/one-sample-per-seed contracts, but none bound *type identity* of `n_bootstrap`/`window` — the boolean gap was genuinely open. Not a five-claim registered source; the forager-matched statistics tests are unaffected (they consume the same helpers with legal integer counts).
Compute receipt: 2038994 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

lalalune:
Deep review confirmed the boolean/float count gap, but the proposed helper still admitted hostile integer subclasses and interpolated rejected values through `repr`. Saved and completed in #1022 with exact concrete integer types and hook-safe failures; closing this original as superseded.

PR #893, opened 2026-08-17
fix(resource-manager): reject non-finite hedge theorem inputs

ss251:
## What changed

`optimal_hedge_learning_rate` and `finite_candidate_hedge_regret_bound` now reject boolean, non-integer, and non-finite theorem preconditions before they can mint a NaN or silently remapped Hedge rate.

On `origin/main` `90c4613a5573de324a7bb33c89efd85e1bc09aa0`:

- `loss_bound=nan` returned a NaN learning rate (`nan <= 0.0` is false);
- `n_actions=True` returned `0.0` (the one-action shortcut);
- `horizon=True` minted a finite rate for a one-step "horizon";
- `LearnedResourceManager.fixed_candidate_regret_bound(nan)` returned NaN.

Those helpers are the published regret-bound surface. A NaN bound is not a bound.

Legal finite integer/real preconditions, the `n_actions == 1 → 0.0` shortcut, and the `learning_rate == 0.0 → inf` bound are unchanged. The rate for `(3, 8)` remains `1.048147073968205`.

## Scope

- `alberta_framework/core/resource_manager.py`
- `tests/test_resource_manager.py`

## Why this is unowned

Live occupancy at publish time: 3 open ASI PRs, all Shaw (`#890` actor-critic, `#891`/`#892` compositional-features / future-utility). Neither target file is in that map.

This is not a republish of `#677` (UPGD step sizes), `#861` (option-duration), world-model `step_size`, Kayvan `#711` (latent-world-model integers), or HEIR `#4`. Prior resource-manager merges (`#840`, `#867`, `#103`, `#215`) hardened constructors, serialized schemas, and 0×inf update arithmetic — not these theorem helpers.

## Verification

Exact candidate head: `85902f406c6c5112d6d303ab3bbe873d5b222981`  
Base: `90c4613a5573de324a7bb33c89efd85e1bc09aa0`

- `pytest tests/test_resource_manager.py -q -o addopts=""`: **63 passed**
- focused new/legal regret tests: **18 passed**
- `ruff check` on the two files: clean
- `mypy alberta_framework/core/resource_manager.py`: clean
- `scope-check.sh --max-files 2`: PASS
- `git diff --check`: clean

## Scientific integrity

This is fail-closed host-boundary validation on the theorem helpers only. It does not change Hedge update equations, defaults, discounting, exploration, registered evidence sources, frozen protocols, scientific claims, benchmark results, `CHANGELOG.md`, or any `outputs/**` artifact.

- AI assistance: yes

[...]

PR #898, opened 2026-08-17
fix(streams): reject non-finite partial-observation sentinels

ss251:
## What changed

`PartialObservationWrapper` now rejects a non-finite or boolean `sentinel` before a hidden channel can be replaced.

- `sentinel=nan` wrote `NaN` into every masked observation channel;
- `sentinel=inf` wrote `Inf` into those channels;
- `sentinel=True` wrote `1.0` (bool subclass of int) instead of the documented default `0.0`.

Those values are the POMDP observation the history-feature / learner path consumes. A NaN hidden channel is not a masked channel.

Legal finite sentinels, including the default `0.0` and a nonzero `-1.0`, are unchanged. `mask_prob` validation for `MaskMode.RANDOM` is unchanged.

## Scope

- `alberta_framework/streams/partial_observation.py`
- `tests/test_partial_observation.py`

## Why this is unowned

Live occupancy at publish time: 8 open ASI PRs (Shaw `#890`–`#892`, ss251 `#893`/`#894`, byt61 `#895`–`#897`). Neither target file is in that map.

This is not a republish of `#768` (class-spoofed `mask_prob`), `#893`/`#894`, Shaw `#890`–`#892`, scooped `step_size` / `#861` / `#711`, or HEIR `#4`. `#768` closed the `mask_prob` host gate and left `sentinel` raw.

## Verification

Exact candidate head: `891a9d78aa6a910b051d0ef1002426764cca5af3`  

- `pytest tests/test_partial_observation.py -q -o addopts=""`: **24 passed**
- `mypy alberta_framework/streams/partial_observation.py`: clean
- `git diff --check`: clean

## Scientific integrity

This is fail-closed host-boundary validation on the wrapper constructor only. It does not change masking modes, default `0.0` replacement, stream RNG, registered evidence sources, frozen protocols, scientific claims, benchmark results, `CHANGELOG.md`, or any `outputs/**` artifact.

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Cursor

— [cursor-ss251]

### DIFF
--- a/alberta_framework/streams/partial_observation.py
+++ b/alberta_framework/streams/partial_observation.py
@@ -94,7 +94,9 @@ class PartialObservationWrapper[InnerStateT]:
             expectation.
         schedule: Tuple of boolean masks of shape ``(feature_dim,)``;
             cycled each step under PERIODIC mode.
-        sentinel: Value that replaces masked entries (default ``0.0``).
+        sentinel: Finite real that replaces masked entries (default ``0.0``).
+            Boolean and non-finite values are rejected so a hidden channel
+            cannot silently become ``1.0``, ``NaN``, or ``Inf``.
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


lalalune (approved):
Deep independent review: correct float32-sink boundary validation; preserves legal finite sentinels and fails closed on bool/NaN/Inf. Rebased integration test: 24/24 focused tests passed; Ruff clean.

PR #899, opened 2026-08-17
fix(idbd): reject non-finite constructor step sizes

ss251:
## What changed

`IDBD.__init__` now rejects a non-finite, boolean, or out-of-domain constructor step-size before `init()` can write a poisoned `log_step_sizes` vector.

- `initial_step_size=nan` stored NaN and `init()` wrote `[nan, nan]` logs;
- `initial_step_size=inf` wrote `[inf, inf]` logs;
- `initial_step_size=True` was accepted as `1.0` (`log(1)=0`), so every weight started at step-size 1;
- `initial_step_size=0.0` wrote `[-inf, -inf]` logs;
- `meta_step_size=nan` / `True` were stored raw (`True` becomes meta-step 1.0).

`log_step_sizes` is the live IDBD step-size state. A NaN or `-inf` log is not an adaptive step-size.

Legal defaults (`0.01`, `0.01`) and the existing `meta_step_size=0.0` skip-adaptation contract are unchanged. Autostep is not touched.

## Scope

- `alberta_framework/core/optimizers.py`
- `tests/test_optimizers.py`

## Why this is unowned

Live occupancy at publish time: 8 open ASI PRs (Shaw `#890`–`#892`, ss251 `#893`/`#894`, byt61 `#895`–`#897`). Neither target file is in that map.

This is not a republish of `#43` (IDBD clip-guard after an update), `#549` (0×inf h-trace skip when meta-step is 0), `#859` (initializer resource contracts), `#893`/`#894`, Shaw `#890`–`#892`, scooped `step_size` / `#861` / `#711`, or HEIR `#4`. Those closed update arithmetic and resource preflights — not constructor domains.

## Verification

Exact candidate head: `7831c1d97630f57178f1e32de3ad57fcc1cbf54c`  

- `pytest tests/test_optimizers.py -q -o addopts=""`: **82 passed**
- focused new constructor tests: **10 passed** (illegal initial/meta + legal zero meta)
- `mypy alberta_framework/core/optimizers.py`: clean
- `git diff --check`: clean

## Scientific integrity

This is fail-closed host-boundary validation on the IDBD constructor only. It does not change Sutton 1992 update equations, the `meta_step_size=0` skip path, defaults, registered evidence sources, frozen protocols, scientific claims, benchmark results, `CHANGELOG.md`, or any `outputs/**` artifact.

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Cursor

— [cursor-ss251]

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -24,6 +24,7 @@ import numpy as np
 from jax import Array
 from jaxtyping import Bool, Float
 
+from alberta_framework.core._float32_scalars import validated_float32_scalar
 from alberta_framework.core.types import (
     AutostepGTDLambdaState,
     AutostepParamState,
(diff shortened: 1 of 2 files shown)


lalalune:
Deep review found the core finite/bool/domain fix correct, but a positive `meta_step_size` could still narrow to float32 zero and silently disable adaptation. Saved and completed in #1015 with an explicit nonzero-underflow contract plus smallest-subnormal coverage; closing this original as superseded.

PR #901, opened 2026-08-17
fix(experiments): reject boolean seed counts and windows

ss251:
## What changed

`run_multi_seed_experiment` and `get_final_performance` now reject boolean and non-canonical seed counts / windows before a publication run can mint a false seed axis.

- `seeds=True` ran a one-seed experiment (`range(True) -> [0]`);
- `seeds=False` / `0` produced an empty `{}` result;
- `seeds=[True]` used True as a seed identity;
- `window=True` averaged the last one step.

The seed list is the aligned identity axis for paired significance tests. A boolean is not a seed count.

Legal count `seeds=2` still yields `[0, 1]`. Explicit `[0, 1]` / `[7, 3]` and existing duplicate-seed messages are unchanged.

## Scope

- `alberta_framework/utils/experiments.py`
- `tests/test_experiments_validation.py`

## Why this is unowned

Live occupancy at publish time: neither target file is in the open ASI PR map. This is not a republish of `#893`/`#894` (hedge / statistics counts), Shaw `#890`–`#892`, byt61 `#895`–`#897`, scooped `step_size` / `#861` / `#711`, or HEIR `#4`.

## Verification

Exact candidate head: `7fe279198b9978928f7d9144cf5b43cffed06e31`

- `pytest tests/test_experiments_validation.py -q -o addopts=""`
- focused new seed-count / window tests
- `git diff --check`: clean

## Scientific integrity

This is fail-closed host-boundary validation on the experiment runner only. It does not change learning-loop equations, aggregation formulas, registered evidence sources, frozen protocols, scientific claims, benchmark results, `CHANGELOG.md`, or any `outputs/**` artifact.

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Cursor

— [cursor-ss251]

### DIFF
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -24,6 +24,7 @@ import jax.random as jr
 import numpy as np
 from numpy.typing import NDArray
 
+from alberta_framework._seed_validation import require_jax_seed
 from alberta_framework.core.learners import (
     LinearLearner,
     metrics_to_dicts,
@@ -31,7 +32,6 @@ from alberta_framework.core.learners import (
 )
 from alberta_framework.core.types import LearnerState
 from alberta_framework.streams.base import ScanStream
-from alberta_framework._seed_validation import require_jax_seed
 from alberta_framework.utils.statistics import common_final_window
 
 _NUMPY_COORDINATE_TYPES = frozenset(
@@ -368,7 +368,7 @@ def run_multi_seed_experiment(
                 "of unique built-in integer seeds"
             )
         try:
-            raw_seeds = tuple(seeds)
+            raw_seeds = tuple(cast(Iterable[int], seeds))
         except TypeError as exc:
             raise ValueError(
                 "seeds must be a positive built-in integer count or a sequence "
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -31,6 +31,7 @@ from alberta_framework.core.learners import (
 )
 from alberta_framework.core.types import LearnerState
 from alberta_framework.streams.base import ScanStream
+from alberta_framework._seed_validation import require_jax_seed
 from alberta_framework.utils.statistics import common_final_window
 
 _NUMPY_COORDINATE_TYPES = frozenset(
@@ -336,8 +337,9 @@ def run_multi_seed_experiment(
         Dictionary mapping config name to AggregatedResults
 
     Raises:
-        ValueError: If two or more configurations have the same name, or if the
-            explicit seed list contains duplicates
+        ValueError: If two or more configurations have the same name, if the
+            seed count is not a positive built-in integer, or if the explicit
+            seed list is empty, non-canonical, or contains duplicates
     """
     seen_names: set[str] = set()
     duplicate_names: set[str] = set()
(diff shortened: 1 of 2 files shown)


kenjohnscreates (changes requested):
## Review — CHANGES_REQUESTED (two confirmed CI-gate regressions; the fix itself is sound)

Verified by re-applying the diff onto current `main` (`90c4613`) in a scratch worktree; both files **byte-identical to PR head `7fe27919`**.

## Confirmed blockers (both clean on baseline, both introduced by this diff)

**1. Repo-config strict mypy is red:**
```
alberta_framework/utils/experiments.py:371: error: Argument 1 to "tuple" has incompatible type "int | Sequence[int]"; expected "Iterable[int]"  [arg-type]
```
After the `type(seeds) is int` branch, mypy cannot narrow `seeds` (an `is`-identity check on `type()` isn't a narrowing form it recognizes), so `tuple(seeds)` in the else-branch still sees `int | Sequence[int]`. Simplest fix: keep `type(seeds) is int` for the bool exclusion but narrow explicitly for the checker — e.g. `if isinstance(seeds, int) and not isinstance(seeds, bool):` (mypy narrows `isinstance`, and the bool exclusion is preserved), or bind `seeds_iter: Sequence[int] = seeds` under a `cast` in the else-branch. Baseline is `Success: no issues found in 168 source files`.

**2. Ruff `I001` (import block unsorted):** the new `from alberta_framework._seed_validation import require_jax_seed` at line 34 sorts before `from alberta_framework.core.types import ...` (underscore < `c`); ruff's own `--fix` resolves it. Baseline `ruff check .` is clean. The PR's verification list ran `git diff --check` and the focused pytest selection but not `ruff check`/`mypy` — worth adding both to the checklist since they are the per-PR CI gates.

## What does verify

- **The defect is real and the fix logic is correct.** `isinstance(seeds, int)` accepts `bool`, so `seeds=True` ran a one-seed experiment and `seeds=False`/`0` produced an empty `{}` — on the axis that paired significance tests align on. Rejection now happens before any factory executes (the `_fail_if_called` sentinel makes that discriminating), the empty-list and string/bytes cases fail closed with field-named errors, and `seeds=2 → [0, 1]` plus explicit-list ordering are pinned unchanged.
- Failing-test-first: tests-only against baseline → **11 of 13** new cases fail (the two that pass on both sides are the `seeds=2` alignment pin and one identity case the existing duplicate-seed path already rejected); full diff → the new tests pass.
- Test file battery: **120 passed / 1 failed**, but the one failure — `test_extract_hyperparameter_results_keeps_wide_finite_numeric_coordinates[coordinate3]` — **is pre-existing on baseline** (I re-ran it with the diff stashed: same failure), so it is not this PR's regression and should not gate here; flagging it as an unrelated red test on `main` for whoever owns `_NUMPY_COORDINATE_TYPES`.
- Consumers: `run_multi_seed_experiment` has no in-package callers under `benchmarks/`/`steps/` (only the public re-export and archived forager source tars), so no measured lane is exposed; not a five-claim registered source.

## One disclosure to add to the description (not gating)

Routing explicit seed lists through `require_jax_seed` makes the contract stricter than "reject bool": that helper is `type(value) is not int`, so **NumPy integer seeds are now rejected** — I verified `seeds=[np.int64(0), np.int64(1)]` raises `seeds[0] must be a built-in integer in [0, 4294967295]`, where baseline accepted them. It also newly enforces the uint32 upper bound. Both are defensible (they align this API with the project's typed-seed contract elsewhere), and nothing in-tree passes NumPy seeds here, but it is a public API tightening the summary should state so downstream users aren't surprised. Note `_require_positive_int` in the sibling #894 deliberately accepts `np.integer` — the two PRs currently pick opposite conventions for the same "count/identity" question; worth making that choice consciously.

Happy to re-verify promptly once the mypy narrowing and import order are pushed — everything else is in place.
Compute receipt: 3780938 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

lalalune (approved):
Deep re-review after maintainer repair: the seed/window identity hardening is correct. I fixed the published Ruff import-order failure and strict-mypy narrowing failure without broadening runtime admission. On current main: 344 experiment/statistics tests passed; package Ruff and diff check clean; strict mypy clean. Explicit seed lists remain ordered and unique, counts remain [0,n), and factories are not called for invalid axes.

PR #919, opened 2026-08-17
fix(benchmarks): reject class-spoofed non-real ewm_decay in ForagerBenchmarkConfig

### LINKED ISSUE #918: forager.py ForagerBenchmarkConfig accepts __class__-spoofed non-real ewm_decay
`_require_real` in `alberta_framework/benchmarks/forager.py` guards against `bool` and non-`(int, float)` values with:

```python
if isinstance(value, bool) or not isinstance(value, (int, float)):
    raise ValueError(f"{name} must be a real number")
```

`isinstance()` consults an object's `__class__` attribute, which can be overridden via a `@property`. An object whose `__class__` property returns `float` passes `isinstance(value, (int, float))` even though `type(value) is not float`.

`ForagerBenchmarkConfig.__post_init__` routes `ewm_decay` through this helper - `ForagerBenchmarkConfig` is exported from `alberta_framework.benchmarks.__all__`.

```text
SPOOFED ewm_decay ACCEPTED: <FakeFloat object> <class 'FakeFloat'>
```

same underlying isinstance-spoofing class already fixed for `_require_int` across the step files, and for `_require_real`/`Real`-typed scalars in `forager_results.py` (#424), `representation_gradient_mixer.py`/`partner_policy_fusion.py`/`types.py`/`synthetic.py` (#569/#571/#574/#649).

fix incoming: switch to `actual_type = type(value); if issubclass(actual_type, bool) or not issubclass(actual_type, (int, float))` - `issubclass()` on the real, non-spoofable `type()` still correctly allows registered subtypes like `numpy.float64` (which genuinely subclasses `float`) through, unlike an exact `type(value) not in (int, float)` check, which would wrongly reject them.

Svector-anu:
Closes #918.

## What changed

Same isinstance-spoofing class already fixed for `_require_int` across the step files, and for `Real`-typed scalars in `forager_results.py` (#424), `representation_gradient_mixer.py`/`partner_policy_fusion.py`/`types.py`/`synthetic.py` (#569/#571/#574/#649): `_require_real` in `forager.py` used `isinstance(value, bool) or not isinstance(value, (int, float))`. `isinstance()` consults `__class__`, which is overridable via a `@property`, so an object whose `__class__` property returns `float` passes `isinstance(value, (int, float))` even though `type(value) is not float`.

`ForagerBenchmarkConfig.__post_init__` routes `ewm_decay` through this helper.

## Reachability

`ForagerBenchmarkConfig` is exported from `alberta_framework.benchmarks.__all__`, and its `__post_init__` runs the vulnerable check on every construction - publicly reachable with ordinary concrete inputs.

## A note on the fix shape

The naive fix here would be `type(value) not in (int, float)` (exact-type match) - that's what I used in my own earlier #424 fix for this exact same pattern, and it's a real, if currently latent, regression there: `numpy.float64` genuinely subclasses Python's `float` (`issubclass(np.float64, float)` is `True`), so an exact-type check wrongly rejects a legitimate `np.float64` value that the original `isinstance` check accepted. This PR uses `issubclass(actual_type, (int, float))` instead, which reads the real, non-spoofable `type()` slot (so it's still safe against `__class__` spoofing) while correctly continuing to accept registered subtypes like `numpy.float64`. I'll follow up on #424 separately to fix the same regression there.

## Verification

```text
SPOOFED ewm_decay ACCEPTED: <FakeFloat object> <class 'FakeFloat'>
```
and confirmed the fix keeps `numpy.float64` working:
```text
np.float64 ewm_decay accepted: 0.5
```

- new tests `test_benchmark_config_rejects_class_spoofed_ewm_decay` and `test_benchmark_config_accepts_numpy_float64_ewm_decay`, added next to the existing `test_benchmark_config_rejects_lossy_numeric_coercion`.
- mutation check: reverted just the source fix, reran the spoof test -> `DID NOT RAISE ValueError`. restored -> passes.
- full file: `99 passed, 3 skipped` (skips pre-existing, unrelated to this change).

## Provenance

AI-assisted (Claude, `claude-sonnet-5`). Spoofing behavior reproduced empirically by me in a real interpreter session against the unpatched code before writing the fix, and the numpy-float64 subclass relationship independently verified before choosing the fix shape. All verification above (mutation testing, full-suite run, lint/typecheck, reachability trace) performed and independently confirmed by me before pushing.

### DIFF
--- a/alberta_framework/benchmarks/forager.py
+++ b/alberta_framework/benchmarks/forager.py
@@ -236,9 +236,15 @@ def _require_builtin_int(
 def _require_real(value: Any, *, name: str) -> float:
     """Validate a built-in finite real scalar while excluding booleans."""
     actual_type = type(value)
-    if issubclass(actual_type, bool) or not issubclass(actual_type, (int, float)):
+    # Preserve the historical acceptance of NumPy's concrete float64 scalar,
+    # which is a float subtype, without admitting arbitrary user-defined
+    # int/float subclasses with overloaded conversion or arithmetic hooks.
+    if actual_type not in (int, float, np.float64):
         raise ValueError(f"{name} must be a real number")
-    converted = float(cast(Any, value))
+    try:
+        converted = float(cast(Any, value))
+    except (OverflowError, TypeError, ValueError) as error:
+        raise ValueError(f"{name} must be finite") from error
     if not math.isfinite(converted):
         raise ValueError(f"{name} must be finite")
     return converted
@@ -1674,6 +1680,7 @@ class ForagerBenchmarkConfig:
         ewm_decay = _require_real(self.ewm_decay, name="ewm_decay")
         if not 0.0 <= ewm_decay < 1.0:
             raise ValueError("ewm_decay must lie in [0, 1)")
+        object.__setattr__(self, "ewm_decay", ewm_decay)
         _require_builtin_int(
             self.record_every,
             name="record_every",
--- a/alberta_framework/benchmarks/forager.py
+++ b/alberta_framework/benchmarks/forager.py
@@ -235,9 +235,10 @@ def _require_builtin_int(
 
 def _require_real(value: Any, *, name: str) -> float:
     """Validate a built-in finite real scalar while excluding booleans."""
-    if isinstance(value, bool) or not isinstance(value, (int, float)):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, (int, float)):
         raise ValueError(f"{name} must be a real number")
-    converted = float(value)
+    converted = float(cast(Any, value))
     if not math.isfinite(converted):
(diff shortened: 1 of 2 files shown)


lalalune:
Maintainer deep review confirmed the original `__class__` spoofing bug, but found the proposed `issubclass(type(value), (int, float))` gate still admitted arbitrary user-defined float subclasses. More importantly, `ForagerBenchmarkConfig` validated a converted local but retained the original object, allowing overloaded arithmetic to execute later in benchmark code.

I repaired the branch in `4aedc6b`: accept only the historically compatible concrete runtime types (`int`, `float`, and `numpy.float64`), translate conversion failures to the public `ValueError` contract, and store the canonical built-in float. Added hostile float-subclass, canonicalization, and huge-integer regressions. Verification: `tests/test_forager_benchmark.py` (104 passed), focused Ruff, mypy, and diff checks clean.

PR #921, opened 2026-08-17
fix(benchmarks): accept numpy.float64 in npz-import ewm_decay check

### LINKED ISSUE #920: forager_results.py ewm_decay check wrongly rejects numpy.float64 (regression in #424)
Self-reported follow-up to my own #424 (`fix(forager-results): validate npz-import decay and window params`, merged).

`import_official_foragax_npz`'s `ewm_decay` check uses `type(ewm_decay) not in (int, float)` - an exact-type match. This correctly closes the `__class__`-spoofing gap #424 fixed, but it's stricter than the check it replaced in an unintended way: `numpy.float64` genuinely subclasses Python's `float` (`issubclass(np.float64, float)` is `True`), so a legitimate `np.float64` value - which the pre-#424 `isinstance`-based check accepted - is now wrongly rejected.

confirmed directly:

```text
np.float64 ewm_decay: REJECTED (still broken): ewm_decay must be a finite number in [0, 1)
```

(isolated the exact validation logic outside the full function to rule out an unrelated macOS temp-dir quirk in my own repro script - confirmed it's the `type() not in` check itself, not a side effect.)

No caller in the current codebase or test suite passes a numpy float here, so this hasn't broken anything live - it's a latent correctness gap, not an active regression - but it should be fixed for consistency with the rest of the codebase's numpy-float-tolerant conventions, and because I've now made the same mistake once and want to close it out properly rather than leave it as a known gap.

fix incoming: switch to `issubclass(actual_type, bool) or not issubclass(actual_type, (int, float))` (same pattern used in #919 for `forager.py`'s equivalent `_require_real`) - `issubclass()` on the real, non-spoofable `type()` correctly keeps accepting `numpy.float64` while still rejecting `__class__`-spoofed objects.

Svector-anu:
Closes #920.

## What changed

Self-correction to my own #424 (`fix(forager-results): validate npz-import decay and window params`, merged). That PR fixed a `__class__`-spoofing gap in `import_official_foragax_npz`'s `ewm_decay` check by switching to `type(ewm_decay) not in (int, float)` - an exact-type match. That's correct against spoofing, but stricter than the `isinstance`-based check it replaced in an unintended way: `numpy.float64` genuinely subclasses Python's `float` (`issubclass(np.float64, float)` is `True`), so a legitimate `np.float64` value - which the pre-#424 check accepted - is now wrongly rejected.

## Reachability / impact

No caller in the current codebase or test suite passes a numpy float here, so this was latent, not an active regression against anything today. Still worth closing cleanly: it's inconsistent with the rest of the codebase's numpy-float-tolerant conventions, and I'd rather fix my own mistake properly than leave a known gap for the next caller to trip over.

## Verification

confirmed the regression directly against the unpatched (i.e. currently-merged #424) code, isolating just the validation logic to rule out an unrelated macOS temp-dir quirk in my own repro script:
```text
np.float64 ewm_decay: REJECTED (still broken): ewm_decay must be a finite number in [0, 1)
```

- new test `test_official_npz_import_accepts_numpy_float64_ewm_decay`, added next to the existing `test_official_npz_import_accepts_finite_endpoint_values`.
- mutation check: reverted just this fix (back to #424's merged `type() not in` check), reran the new test -> fails with the exact reproduced regression. restored -> passes.
- full file: `98 passed, 3 skipped` (skips pre-existing, unrelated).
- spoofing protection reconfirmed still intact (not shown as a new test since #424 already covers it): a `__class__`-spoofed object and a plain `bool` are both still correctly rejected under the new `issubclass`-based check.

## Provenance

AI-assisted (Claude, `claude-sonnet-5`). Found this while independently double-checking my own earlier fix shape against a same-pattern fix I was writing for `forager.py` (#919) - noticed I'd used the same exact-type-check idiom there and wanted to confirm it hadn't caused the same problem here. All verification above (mutation testing, full-suite run, lint/typecheck) performed and independently confirmed by me before pushing.

### DIFF
--- a/alberta_framework/benchmarks/forager_results.py
+++ b/alberta_framework/benchmarks/forager_results.py
@@ -1567,8 +1567,10 @@ def import_official_foragax_npz(
     final_window: int = 100_000,
 ) -> ForagerRunResult:
     """Convert one official ``data/<seed>.npz`` archive to Alberta's schema."""
-    ewm_decay_type = type(ewm_decay)
-    if issubclass(ewm_decay_type, bool) or not issubclass(ewm_decay_type, (int, float)):
+    # NumPy's concrete float64 scalar was accepted by the historical
+    # isinstance gate. Retain that compatibility without admitting arbitrary
+    # user-defined int/float subclasses with overloaded conversion hooks.
+    if type(ewm_decay) not in (int, float, np.float64):
         raise ValueError("ewm_decay must be a finite number in [0, 1)")
     try:
         ewm_decay_as_float = float(cast(Any, ewm_decay))
--- a/alberta_framework/benchmarks/forager_results.py
+++ b/alberta_framework/benchmarks/forager_results.py
@@ -1567,10 +1567,11 @@ def import_official_foragax_npz(
     final_window: int = 100_000,
 ) -> ForagerRunResult:
     """Convert one official ``data/<seed>.npz`` archive to Alberta's schema."""
-    if type(ewm_decay) not in (int, float):
+    ewm_decay_type = type(ewm_decay)
+    if issubclass(ewm_decay_type, bool) or not issubclass(ewm_decay_type, (int, float)):
         raise ValueError("ewm_decay must be a finite number in [0, 1)")
     try:
-        ewm_decay_as_float = float(ewm_decay)
+        ewm_decay_as_float = float(cast(Any, ewm_decay))
     except OverflowError as exc:
         raise ValueError("ewm_decay must be a finite number in [0, 1)") from exc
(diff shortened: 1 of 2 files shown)


lalalune:
Maintainer deep review confirmed the `numpy.float64` compatibility regression from #424, but the proposed broad subtype gate also admitted arbitrary user-defined float subclasses and their overloaded conversion hooks.

I repaired the branch in `7e7d482` to accept the exact historical concrete runtime types (`int`, `float`, and `numpy.float64`) while continuing to reject booleans, `__class__` spoofing, and custom numeric subclasses. Added a hostile float-subclass regression. Verification: `tests/test_forager_benchmark.py` (102 passed), focused Ruff, mypy, and diff checks clean.

PR #939, opened 2026-08-17
test(experiments): make the wide-longdouble coordinate fixture platform-honest (fixes #937)

### LINKED ISSUE #937: test_experiments_validation: np.longdouble("1e4000") fixture is inf on macOS arm64 — suite red on the sanctioned runner platform
## Summary

`tests/test_experiments_validation.py::test_extract_hyperparameter_results_keeps_wide_finite_numeric_coordinates[coordinate3]` is **deterministically red on macOS arm64** at current `main` (`90c4613`), because its fixture `np.longdouble("1e4000")` is not a finite value on that platform.

## Root cause

On aarch64 (macOS arm64 included) NumPy's `longdouble` **is** IEEE binary64 — there is no 80-bit extended type:

```
>>> np.finfo(np.longdouble).bits, np.finfo(np.longdouble).max == np.finfo(np.float64).max
(64, True)
```

So `np.longdouble("1e4000")` overflows to `inf` at parse (`RuntimeWarning: overflow encountered in conversion from string`), and `extract_hyperparameter_results` — correctly — rejects it as a non-finite coordinate:

```
ValueError: param_extractor returned a noncanonical coordinate for configuration 'candidate'; ...
floating or complex coordinates must be finite
```

The **validator is right; the fixture is wrong.** The test's intent ("a finite longdouble outside float64 range is kept, not rejected") describes a value that only exists where longdouble is wider than float64. Verified in a clean worktree at `90c4613`: 1 failed / 100 passed in that file; the failure reproduces with no diff applied.

## Why it matters here specifically

- The project's sanctioned campaign runner is **macOS arm64** (hosted `macos-14` M1 for the #51/#184 preregistration workflow, and the local arm64 development boxes), so the test suite should be green there.
- This is the same platform class as the `np.longdouble("1e4000")` precondition failure that was confirmed on the now-closed #190 (its close reason cites it) — the fixture pattern was introduced independently here in `ae6ab0a` ("fix(experiments): constrain sensitivity coordinates") and would recur anywhere `1e4000` is used as a "finite longdouble" literal.

## Proposed fix (test-only, no production change)

1. Replace the fixture with `np.finfo(np.longdouble).max` — the widest finite longdouble the platform actually has: an extended-range value on x86-64, exactly `float64.max` on aarch64, finite everywhere. Same intent, platform-honest.
2. Add a companion test that classifies `np.longdouble` **by the platform's actual value**: `inf`/`nan` must be rejected everywhere, and the `"1e4000"` literal must be *kept-if-finite / rejected-if-not* — so the platform-dependent case is exercised rather than skipped, and the validator's finiteness check on `longdouble` gets the direct rejection coverage the old fixture only accidentally provided on arm64.

PR to follow immediately, failing-test-first (the current red test **is** the failing test).

— [kenjohnscreates]

kenjohnscreates:
Fixes #937.

## What

`test_extract_hyperparameter_results_keeps_wide_finite_numeric_coordinates[coordinate3]` is deterministically red on macOS arm64 at current `main` because its fixture `np.longdouble("1e4000")` overflows to `inf` at parse there — on aarch64 NumPy's `longdouble` **is** binary64 (`np.finfo(np.longdouble).bits == 64`, `max == float64.max`), so no finite longdouble outside float64 range exists on the platform. The validator correctly rejects the non-finite coordinate; the fixture's claim ("finite") is what's wrong. **Test-only change; no production code touched.**

## Fix

- Replace the fixture with `np.finfo(np.longdouble).max` — the widest finite longdouble the platform actually has (extended-range on x86-64, exactly `float64.max` on aarch64, finite everywhere). Same intent, platform-honest.
- Add `test_extract_hyperparameter_results_classifies_longdouble_by_platform_value`: `np.longdouble("inf")`/`("nan")` must be **rejected** everywhere, and the `"1e4000"` literal must be *kept-if-finite / rejected-if-not* according to the platform's actual value. This exercises the platform-dependent case instead of skipping it, and gives the `longdouble` finiteness rejection direct coverage the old fixture only accidentally exercised on arm64.

## Evidence (macOS arm64, M1 Pro; `macOS-26.2-arm64`, numpy 2.5.2, longdouble bits 64)

Failing-test-first — the existing test **is** the red test. Commands run from a clean worktree:

```text
# base 90c4613 (main), file at base:
pytest tests/test_experiments_validation.py -q -o addopts="" -k wide_finite_numeric
FAILED ...::test_extract_hyperparameter_results_keeps_wide_finite_numeric_coordinates[coordinate3]
1 failed, 3 passed, 94 deselected

# head e1fd635e:
pytest tests/test_experiments_validation.py -q -o addopts=""
101 passed, 1 warning

pytest ... -k "wide_finite_numeric or longdouble_by_platform"
7 passed, 94 deselected

ruff check .        -> All checks passed!
mypy (repo config)  -> Success: no issues found in 168 source files
git diff --check    -> clean
```

- [x] logs — exact command transcript above (also bound as the trajectory in the receipt below); the red run reproduces on `main` with no diff applied.

## Scope / integrity

Test fixture only. Does not touch `alberta_framework/`, any registered evidence source, frozen protocol, pinned `outputs/**` artifact, or `CHANGELOG.md`. Motivation: the project's sanctioned campaign runner is macOS arm64 (hosted `macos-14` M1 for the #51/#184 preregistration workflow, local arm64 dev boxes), so the suite should be green there; this is the same platform class confirmed on the now-closed #190. Fixture introduced in `ae6ab0a`; no open PR or issue claims it (searched `wide_finite_numeric` / `_NUMPY_COORDINATE_TYPES` / `extract_hyperparameter_results` / `longdouble`).

Deviations from pre-registration: none — #937 filed first, PR follows exactly as described there.

— [kenjohnscreates]

### DIFF
--- a/tests/test_experiments_validation.py
+++ b/tests/test_experiments_validation.py
@@ -493,6 +493,12 @@ class _NonfiniteFloatThatConvertsFinite(float):
         return 0j
 
 
+def _platform_longdouble_1e4000() -> np.longdouble:
+    """Parse the cross-platform boundary without leaking an expected warning."""
+    with np.errstate(over="ignore", invalid="ignore"):
+        return np.longdouble("1e4000")
+
+
 @pytest.mark.parametrize(
     "coordinate",
     [
@@ -502,7 +508,7 @@ class _NonfiniteFloatThatConvertsFinite(float):
         # and stays a finite extended value where it is wider (x86-64): the
         # validator must classify by the platform's actual value, not the
         # literal, so this case is finite-or-rejected but never a crash.
-        np.longdouble("1e4000"),
+        _platform_longdouble_1e4000(),
     ],
 )
 def test_extract_hyperparameter_results_classifies_longdouble_by_platform_value(
--- a/tests/test_experiments_validation.py
+++ b/tests/test_experiments_validation.py
@@ -493,6 +493,30 @@ class _NonfiniteFloatThatConvertsFinite(float):
         return 0j
 
 
+@pytest.mark.parametrize(
+    "coordinate",
+    [
+        np.longdouble("inf"),
+        np.longdouble("nan"),
+        # Overflows to inf at parse wherever longdouble is float64 (aarch64),
+        # and stays a finite extended value where it is wider (x86-64): the
+        # validator must classify by the platform's actual value, not the
+        # literal, so this case is finite-or-rejected but never a crash.
+        np.longdouble("1e4000"),
+    ],
+)
+def test_extract_hyperparameter_results_classifies_longdouble_by_platform_value(
+    coordinate: object,
+) -> None:
+    trace = {"candidate": _flat_trace("candidate", 1.0)}
+    if bool(np.isfinite(coordinate)):
+        extracted = extract_hyperparameter_results(trace, param_extractor=lambda _: coordinate)
+        assert next(iter(extracted)) is coordinate
+    else:
+        with pytest.raises(ValueError, match=r"noncanonical coordinate"):
+            extract_hyperparameter_results(trace, param_extractor=lambda _: coordinate)
+
+
 def test_extract_hyperparameter_results_rejects_spoofed_numeric_subclasses() -> None:
     coordinate = _NonfiniteFloatThatConvertsFinite()
     assert coordinate == float("inf")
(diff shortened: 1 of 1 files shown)


lalalune:
Maintainer review confirmed the platform-honest fixture. I made the platform-dependent `1e4000` parse warning-free in `1020f8f`, so the intended finite-or-rejected classification also passes under `-W error` on platforms where longdouble is binary64. Verification: all 101 tests passed under warnings-as-errors; Ruff and diff checks clean.

PR #948, opened 2026-08-17
fix(timing): report failed after instead of completed in when timed block raises

### LINKED ISSUE #923: fix(timing): Timer reports failed runs as completed
## Hole

`alberta_framework/utils/timing.py` `Timer.__exit__` on origin/main `90c4613a5573de324a7bb33c89efd85e1bc09aa0` always prints:

```text
{name} completed in {duration}
```

including when the timed block raised. Experiment logs that use `Timer("Training")` therefore record a failed run as a successful duration. The duration itself is still measured; the lie is the completion verb.

Concrete host result on that SHA:

```text
Timer("Training") + RuntimeError("boom") → "Training completed in 0.00s"
```

Duration is still recorded after the exception (`end_time >= start_time`). The success-path message is unchanged.

## Repair

Print `completed in` only when `exc_type is None`. On exception print `failed after` and re-raise.

## Scope

- `alberta_framework/utils/timing.py`
- `tests/test_timing.py`

— [cursor-ss251]

Kayvan-Zahiri:
## Summary
- Resolves #923.
- Modifies `Timer.__exit__` in `alberta_framework/utils/timing.py` so that when a timed code block raises an exception (`exc_type is not None`), it prints `{name} failed after {formatted}` rather than claiming `{name} completed in {formatted}`.
- Adds comprehensive unit tests in `tests/test_timing.py` covering successful runs, failed runs, and silent mode.

## Verification
- `PYTHONPATH=. pytest tests/test_timing.py tests/test_utils_smoke.py`: 19 passed
- `ruff check .`: clean
- `mypy alberta_framework/utils/timing.py`: clean

### DIFF
--- a/alberta_framework/utils/timing.py
+++ b/alberta_framework/utils/timing.py
@@ -128,7 +128,10 @@ class Timer:
 
         if self.verbose:
             formatted = format_duration(self.duration)
-            self.print_fn(f"{self.name} completed in {formatted}")
+            if exc_type is None:
+                self.print_fn(f"{self.name} completed in {formatted}")
+            else:
+                self.print_fn(f"{self.name} failed after {formatted}")
 
     def elapsed(self) -> float:
         """Get elapsed time since timer started (can be called during execution).
--- a/tests/test_timing.py
+++ b/tests/test_timing.py
@@ -1,0 +1,48 @@
+"""Tests for timing utilities."""
+
+import time
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


lalalune:
Closed as a duplicate of #925. Both implement the same success-versus-failure Timer semantics; #925 is earlier, cleanly scoped, and its failure-duration and existing smoke coverage passed in the combined 270-test audit.

PR #951, opened 2026-08-17
fix(experiments): reject boolean and invalid seed counts, seeds sequences, and window sizes

### LINKED ISSUE #900: fix(experiments): boolean seed count silently becomes a one-seed run
## Hole

`run_multi_seed_experiment(..., seeds=True)` on origin/main `90c4613a5573de324a7bb33c89efd85e1bc09aa0` uses `isinstance(seeds, int)` then `list(range(seeds))`. Boolean True is a subclass of int, so the runner silently executes a **one-seed** experiment and records seed identity `0`. `seeds=False` / `0` produce an empty result map. `get_final_performance(..., window=True)` uses a one-step suffix because `True <= 0` is false.

Those values are the publication seed axis and the final-window length. A boolean is not a seed count.

## Repair

Require `type(seeds) is int` and `seeds >= 1` for the count path. Sequence identities go through `require_jax_seed`. `window` must be a positive built-in integer.

## Scope

- `alberta_framework/utils/experiments.py`
- `tests/test_experiments_validation.py`

— [cursor-ss251]

Kayvan-Zahiri:
## Summary
- Resolves #900.
- Hardens `run_multi_seed_experiment` in `alberta_framework/utils/experiments.py` to reject boolean seeds (`seeds=True/False`), require positive integer seed counts, and route explicit seed collections through `require_jax_seed`.
- Validates `window` in `get_final_performance` as a positive built-in integer.

## Verification
- `PYTHONPATH=. pytest tests/test_experiments_validation.py`: 98 passed, 1 skipped
- `ruff check .`: clean
- `mypy alberta_framework/utils/experiments.py`: clean

### DIFF
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -24,6 +24,7 @@ import jax.random as jr
 import numpy as np
 from numpy.typing import NDArray
 
+from alberta_framework._seed_validation import require_jax_seed
 from alberta_framework.core.learners import (
     LinearLearner,
     metrics_to_dicts,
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by merged #901 (merge f12419c). The merged implementation fixes the same count/list/window defects while retaining stricter canonical semantics: exact builtin integer counts, non-empty explicit seed sequences, indexed seed errors, uint32 seed bounds, duplicate detection, and rejection before factories execute. This PR still permits int subclasses as counts and permits an empty explicit sequence, and includes unrelated formatting/test-skipping changes. #901's original Ruff/mypy failures were repaired and its current-main experiment/statistics panel passed 344 tests.

PR #967, opened 2026-08-17
fix(benchmarks): reject float and boolean executable mode aliases in runtime profile

### LINKED ISSUE #953: runtime_profile.py accepts float alias of imageio-ffmpeg executable mode
`validate_environment_runtime_profile` in `alberta_framework/benchmarks/runtime_profile.py` checked the attested imageio-ffmpeg executable mode with bare numeric equality:

```python
ffmpeg["mode"] != 0o555
```

Python considers `365.0 == 0o555` true (`0o555` is `365`), so a runtime-profile JSON object with a floating-point `mode` passes structural validation even though the producer contract obtains the mode as a genuine integer via `stat.S_IMODE(...)`. `365` and `365.0` are different JSON values with different canonical hashes, but both pass this check - weakening the fail-closed runtime provenance schema.

confirmed directly:

```text
>>> 365.0 == 0o555
True
```

same numeric-alias family already fixed across the `forager_matched_*` files (7 files, bare `==`/`!=` letting float/bool aliases of the correct integer cross validation).

## Reachability

- `validate_environment_runtime_profile` is exported from `alberta_framework.benchmarks.__all__`.
- `environment_runtime_profile_sha256` calls it before hashing runtime-profile data.
- `validate_environment_runtime_identity` calls it while verifying an externally supplied runtime profile and its reported digest.
- `forager_matrix.py` calls `validate_environment_runtime_identity` on execution-envelope data; `forager_results.py` calls `environment_runtime_profile_sha256` while importing official runtime metadata.

fix incoming: require `type(ffmpeg["mode"]) is int` before the equality check.

Kayvan-Zahiri:
## Summary
- Resolves #953.
- Hardens `validate_environment_runtime_profile` in `alberta_framework/benchmarks/runtime_profile.py` by requiring `type(ffmpeg["mode"]) is int` and `type(shadow["tmp_src_mode"]) is int` before numeric equality checks, fail-closing against boolean and floating-point mode aliases.
- Adds unit tests in `tests/test_runtime_profile.py`.

## Verification
- `PYTHONPATH=. pytest tests/test_runtime_profile.py`: 24 passed
- `ruff check .`: clean
- `mypy alberta_framework/benchmarks/runtime_profile.py`: clean

### DIFF
--- a/alberta_framework/benchmarks/runtime_profile.py
+++ b/alberta_framework/benchmarks/runtime_profile.py
@@ -15,27 +15,13 @@ from collections.abc import Mapping
 from pathlib import PurePosixPath
 from typing import Any, Literal, cast
 
-ENVIRONMENT_RUNTIME_PROFILE_SCHEMA_VERSION = (
-    "alberta.environment_runtime_profile.v1"
-)
-ENVIRONMENT_RNG_SCHEDULE_SCHEMA_VERSION = (
-    "alberta.environment_rng_schedule.v1"
-)
-CUDA_WHEEL_LIBRARY_PROFILE_SCHEMA_VERSION = (
-    "alberta.cuda_wheel_library_profile.v1"
-)
-DRIVER_LIBRARY_TREE_HASH_SCHEME = (
-    "canonical-entry-json+mode+size+bytes-v1"
-)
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by focused combined repair #1027, which preserves both executable-mode guards and contributor credit without the unrelated formatter churn.

PR #968, opened 2026-08-17
fix(benchmarks): reject __class__-spoofed non-real reward in historical_forager

### LINKED ISSUE #926: historical_forager.py accepts __class__-spoofed non-real reward from pluggable environments

Kayvan-Zahiri:
## Summary
- Resolves #926.
- Hardens `_finite_reward` in `alberta_framework/benchmarks/historical_forager.py` by inspecting the genuine type slot `actual_type = type(value)` and checking `issubclass(actual_type, bool) or not issubclass(actual_type, Real)` to prevent spoofed `__class__` property objects from passing validation.
- Adds unit tests in `tests/test_historical_forager_reconstructed.py`.

## Verification
- `PYTHONPATH=. pytest tests/test_historical_forager_reconstructed.py`: 34 passed
- `ruff check .`: clean
- `mypy alberta_framework/benchmarks/historical_forager.py`: clean

### DIFF
--- a/alberta_framework/benchmarks/historical_forager.py
+++ b/alberta_framework/benchmarks/historical_forager.py
@@ -275,7 +275,8 @@ def _require_environment(value: Any) -> HistoricalForagerEnvironment:
 
 
 def _finite_reward(value: Any) -> float:
-    if isinstance(value, bool) or not isinstance(value, Real):
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, Real):
         raise HistoricalForagerContractError("historical reward must be a real scalar")
     result = float(value)
     if not math.isfinite(result):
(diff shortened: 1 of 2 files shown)


lalalune:
Closed as a duplicate of #927. The older PR implements the same actual-type reward gate and has the stronger end-to-end regression proving rejection before kernel execution or output creation.

PR #969, opened 2026-08-17
fix(forager-cli): resolve omitted vs explicit final_window with is None

### LINKED ISSUE #922: fix(forager-cli): --final-window 0 silently becomes the protocol window
## Hole

`alberta_framework/forager_cli.py` `main()` on origin/main `90c4613a5573de324a7bb33c89efd85e1bc09aa0` builds the publication window as:

```python
final_window=args.final_window or protocol.final_window_steps
```

Argparse stores `0` for `--final-window 0` and `None` when the flag is omitted. `0 or default` treats the requested zero as missing and substitutes the protocol window (100000 / 50000). `ForagerBenchmarkConfig` already requires `final_window >= 1`, so a requested empty window never reaches that fail-closed check. The published "final window" metric is then a different window than the one the caller asked for, with no error.

`--steps 0` on the same function already uses `is not None` and is refused by the config. `--final-window` is the leftover `or`.

Concrete host result on that SHA:

```text
--final-window 0  →  parsed 0
0 or 100000       →  100000
```

This is not a twin of #919 (class-spoofed `ewm_decay` on `forager.py`).

## Repair

Resolve omitted vs explicit with `is None`, so `0` reaches `ForagerBenchmarkConfig` and is refused.

## Scope

- `alberta_framework/forager_cli.py`
- `tests/test_forager_cli.py`

— [cursor-ss251]

Kayvan-Zahiri:
## Summary
- Resolves #922.
- Fixes `main()` in `alberta_framework/forager_cli.py` to resolve omitted vs explicit `--final-window` with `args.final_window is None` rather than `or`, allowing `--final-window 0` to reach `ForagerBenchmarkConfig` and fail-close.
- Adds unit tests in `tests/test_forager_cli.py`.

## Verification
- `PYTHONPATH=. pytest tests/test_forager_cli.py`: 6 passed
- `ruff check .`: clean
- `mypy alberta_framework/forager_cli.py`: clean

### DIFF
--- a/alberta_framework/forager_cli.py
+++ b/alberta_framework/forager_cli.py
@@ -634,7 +634,9 @@ def main(argv: Sequence[str] | None = None) -> int:
         steps=steps,
         ewm_decay=protocol.ewm_decay,
         record_every=args.record_every,
-        final_window=args.final_window or protocol.final_window_steps,
+        final_window=(
+            protocol.final_window_steps if args.final_window is None else args.final_window
+        ),
         jax_chunk_size=args.jax_chunk_size,
     )
     features = ForagerFeatureConfig(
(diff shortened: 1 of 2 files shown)


lalalune:
Closed as a duplicate of #924. The older PR has the same fail-closed omitted-versus-explicit-zero correction without unrelated formatter churn, and its helper-level and wiring regressions passed in the combined 270-test audit.

PR #931, opened 2026-08-17
fix(multi-head): reject leftover constructor scalars

ss251:
Closes #929.

## What changed

`MultiHeadMLPLearner` no longer stores leftover constructor scalars that `#720` did not fence. Head counts, hidden widths, and `per_head_gamma_lamda` were already fail-closed; `sparsity`, `leaky_relu_slope`, `gamma`, `lamda`, `use_layer_norm`, and `trace_mode` were not.

On origin/main `90c4613a5573de324a7bb33c89efd85e1bc09aa0`:

```text
sparsity=True|nan     ACCEPT  (True serializes as a full mask)
gamma=True, lamda=0.0 ACCEPT  (True stores as undiscounted 1)
use_layer_norm=1      ACCEPT
```

After this head those identities raise `ValueError`. Legal defaults stay bit-identical (`sparsity=0.9`, `leaky_relu_slope=0.01`, `gamma=0.0`, `lamda=0.0`, `utility_decay=0.99`). `from_config` inherits the same constructor fence.

## Scope

- `alberta_framework/core/multi_head_learner.py`
- `tests/test_multi_head_learner.py`

## Why this is unowned

Live occupancy at publish time: neither target file is in the open ASI PR map. This is not a republish of `#893`/`#894`/`#898`/`#899`/`#901`/`#903`/`#904`/`#905`/`#908`/`#909`/`#912`/`#913`/`#916`/`#917`/`#924`/`#925`, Shaw `#890`–`#892`, byt61 `#895`–`#897`, or leftover tax on forager-cli, timing, label-emnist, smoke CLI, average-reward, upgd-ipmnist, metrics, nexting, experiments, security, IDBD, true-online-td, baseline-optimizers, or partial-observation. Closed `#720` fenced allocation and per-head traces only.

## Verification

Exact candidate head: `28f4e9ca6df754ed27cd4ab976edf1047ae8a5df`
Base: `90c4613a5573de324a7bb33c89efd85e1bc09aa0`

- Red on base: `sparsity=True` and `gamma=True` constructed and serialized
- `PYTHONPATH=<worktree> pytest tests/test_multi_head_learner.py` → 137 passed
- `ruff check` on the two changed files: clean
- `git diff --check`: clean
- `scope-check.sh --max-files 2`: PASS

## Scientific integrity

[...]

PR #974, opened 2026-08-17
fix(benchmarks): accept numpy.float64 and reject __class__-spoofed ewm_decay

### LINKED ISSUE #920: forager_results.py ewm_decay check wrongly rejects numpy.float64 (regression in #424)

Kayvan-Zahiri:
## Summary
- Resolves #920 and #918.
- Fixes `_require_real` in `alberta_framework/benchmarks/forager.py` and `ewm_decay` validation in `alberta_framework/benchmarks/forager_results.py` to inspect `type(value)` with `issubclass(actual_type, bool) or not issubclass(actual_type, (int, float))`.
- Correctly rejects spoofed `__class__` property objects and booleans while continuing to accept registered real float subtypes like `numpy.float64`.
- Adds unit tests in `tests/test_forager_benchmark.py`.

## Verification
- `PYTHONPATH=. pytest tests/test_forager_benchmark.py -k "test_official_npz_import or test_forager_benchmark_config_ewm_decay_validation"`: 18 passed
- `ruff check .`: clean
- `mypy alberta_framework/benchmarks/forager.py alberta_framework/benchmarks/forager_results.py`: clean

### DIFF
--- a/alberta_framework/benchmarks/forager.py
+++ b/alberta_framework/benchmarks/forager.py
@@ -115,8 +115,7 @@ def _create_reward_trace_sinks(
             sink = factory(int(seed), steps)
             if not isinstance(sink, ForagerRewardTraceSink):
                 raise TypeError(
-                    "reward_trace_sink_factory must return "
-                    "ForagerRewardTraceSink instances"
+                    "reward_trace_sink_factory must return ForagerRewardTraceSink instances"
                 )
             sinks.append(sink)
     except BaseException:
@@ -234,8 +233,9 @@ def _require_builtin_int(
 
 
 def _require_real(value: Any, *, name: str) -> float:
-    """Validate a built-in finite real scalar while excluding booleans."""
-    if isinstance(value, bool) or not isinstance(value, (int, float)):
+    """Validate a finite real scalar while excluding booleans."""
+    actual_type = type(value)
+    if issubclass(actual_type, bool) or not issubclass(actual_type, (int, float)):
         raise ValueError(f"{name} must be a real number")
     converted = float(value)
     if not math.isfinite(converted):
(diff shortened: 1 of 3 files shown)


lalalune:
Superseded by the stricter forager validation already merged through #925 and #931. Current main uses an exact actual-type allowlist, accepts numpy.float64, rejects __class__ spoofing, arbitrary float subclasses, booleans, non-finite values, and huge integers. This PR's issubclass rule would re-allow arbitrary int/float subclasses and also carries large unrelated formatter churn, so it should not be merged.

PR #975, opened 2026-08-17
fix(upgd-ipmnist): require integer schema_version and exact policy booleans

### LINKED ISSUE #911: fix(upgd-ipmnist): v2 validator accepts 2.0 and numeric policy aliases
## Hole

`validate_upgd_ipmnist_v2_artifact` on origin/main `90c4613a5573de324a7bb33c89efd85e1bc09aa0` binds the frozen v2 contract with Python `==`:

- `schema_version != 2` accepts JSON `2.0` (`2.0 == 2`)
- `evidence_policy != V2_NONPROMOTING_POLICY` accepts numeric aliases (`1 == True`, `0 == False`)

v1 already requires `type(schema_version) is int`. v2 provenance already uses identity (`is not False`). The v2 artifact checks dropped both.

Concrete host result on that SHA with a legal synthetic v2 fixture, then only the artifact JSON rewritten:

- legal `schema_version: 2` + boolean policy → `valid=True`
- `schema_version: 2.0` → `valid=True`
- policy `development_only: 1`, `scientific_promotion_allowed: 0`, `execution_attestation: 0` → `valid=True`

A validator that accepts a different schema identity and a different policy object is not binding the permanently nonpromoting contract.

## Repair

Require `schema_version` to be the integer `2` (`type is int`). Bind each frozen policy flag by identity (`is True` / `is False`) and require the exact policy key set.

## Scope

- `alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py`
- `tests/test_upgd_ipmnist_nonpromoting.py`

— [cursor-ss251]

Kayvan-Zahiri:
## Summary
- Resolves #911.
- Hardens `validate_upgd_ipmnist_v2_artifact` in `alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py` by requiring `type(schema_version) is int and schema_version == 2` and checking policy booleans by identity (`is True` / `is False`) across the full expected key set.
- Adds unit tests in `tests/test_upgd_ipmnist_nonpromoting.py`.

## Verification
- `PYTHONPATH=. pytest tests/test_upgd_ipmnist_nonpromoting.py`: 25 passed
- `ruff check .`: clean
- `mypy alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py`: clean

### DIFF
--- a/alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py
+++ b/alberta_framework/evaluation/upgd_ipmnist_nonpromoting.py
@@ -50,9 +50,7 @@ V2_PROTOCOL_DEVIATIONS: list[dict[str, str]] = [
     {
         "code": "task_aligned_logging",
         "scope": "metric_blocks",
-        "description": (
-            "task blocks are [t*L, (t+1)*L); upstream logging is shifted by one step"
-        ),
+        "description": ("task blocks are [t*L, (t+1)*L); upstream logging is shifted by one step"),
     },
     {
         "code": "float32_bias_corrections",
@@ -450,9 +448,7 @@ def _parse_partials(
 ) -> tuple[list[_Shard], list[str], tuple[tuple[str, str], ...], tuple[int, ...]]:
     errors: list[str] = []
     try:
-        normalized_seeds = require_unique_jax_seeds(
-            expected_seeds, name="expected_seeds"
-        )
+        normalized_seeds = require_unique_jax_seeds(expected_seeds, name="expected_seeds")
     except ValueError as exc:
         errors.append(str(exc))
         normalized_seeds = ()
@@ -619,9 +615,7 @@ def validate_upgd_ipmnist_artifact(
     expected_seeds: Sequence[int] = EXPECTED_SEEDS,
 ) -> UPGDIPMNISTValidation:
     """Recompute and validate a v1 result artifact from its complete shards."""
-    shards, errors, digests, normalized_seeds = _parse_partials(
-        partial_paths, expected_seeds
-    )
+    shards, errors, digests, normalized_seeds = _parse_partials(partial_paths, expected_seeds)
     artifact_digest: str | None = None
     try:
         artifact, artifact_digest = _strict_json_object_with_sha256(Path(artifact_path))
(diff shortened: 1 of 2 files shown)


kenjohnscreates (changes requested):
## Review — CHANGES_REQUESTED (substantive fix is correct; please drop the unrelated reformatting from this diff)

Verified by re-applying the diff onto current `main` (`3de3fb3`) in a scratch worktree; both files **byte-identical to PR head `0bca3b6e`**.

## The substantive change is right and small (~10 lines)

- `type(schema_version) is int and == 2` closes the `2.0` alias (JSON floats compare equal to ints in Python; `bool` is excluded by `type(...) is int`).
- The `evidence_policy` check moves from dict equality to identity checks on the booleans (`is True` / `is False`) across the exact expected key set, so `{"development_only": 1, "scientific_promotion_allowed": 0}` — which compares *equal* to the policy under `==` — is now rejected. Good catch: on baseline that numeric-alias policy passes `artifact.get("evidence_policy") != V2_NONPROMOTING_POLICY`. The `evidence_class` string is still pinned exactly.
- Failing-test-first reproduced: tests-only against baseline → the new test fails deterministically; full diff → **25/25**; `ruff check .` clean; repo-config strict mypy clean at **168 source files**.
- Pinned-artifact exposure re-checked (this is the module whose hash the nonpromoting receipts bind, via the immutable `strict_validator_snapshot.v1.py`, not the live file — established in #232): `test_nonpromoting_receipt_binds_stored_shards_and_summary` passes with the diff; no `outputs/**` change.

## Request: the ~180 other changed lines are a `ruff format` migration of two files, unrelated to #911

Roughly 90% of this diff is `ruff format` collapsing multi-line parenthesized expressions in both files (docstring `description` tuples, `require_unique_jax_seeds(...)`, the receipt-test assertions, etc.). I checked why: with the diff applied `ruff format --check` reports both files "already formatted", while on baseline it says both "would be reformatted" — so the reformat is the tool's output, not hand edits. But **the repository does not use `ruff format`**: CI runs only `ruff check .` (`.github/workflows/ci.yml:193`), and on current `main` `ruff format --check .` reports **296 of 416 files would be reformatted**. Reformatting two files against the repo's actual practice (a) makes the substantive change hard to review, (b) creates a needless byte-change on a validator module that historical receipts name, and (c) will churn again the moment someone edits these files without `ruff format`. Please reduce the diff to the schema/policy hunks and the new test — if a formatting migration is wanted, that's a separate repo-wide decision for the maintainer.

Two small notes on the retained parts, non-gating: the added blank line before `def manifest_identity` is fine but is also format-tool output; and the new test writes `artifact.write_text(json.dumps(payload))` after the fixture wrote it with the module's canonical serializer — fine for a rejection test since the validator re-reads bytes, just noting it isn't byte-canonical.

Happy to re-verify promptly on a narrowed diff — the actual hardening is exactly what #911 asks for.
Compute receipt: 3529847 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

lalalune:
Superseded by #913, already on main. Current main requires an exact built-in integer schema_version and identity-checks every policy boolean across the exact key set; the retained tests cover float schema aliases and numeric policy aliases. This stale head adds no remaining behavior and carries unrelated formatter churn.

PR #979, opened 2026-08-17
fix(benchmarks): validate v2 shard hyperparameters against learner domain

### LINKED ISSUE #978: upgd_ipmnist.py merge_partial_results accepts shard hyperparameters outside the learner's valid domain
`_validated_partial_payload` in `alberta_framework/benchmarks/upgd_ipmnist.py` accepted v2 shard `hyperparameters` based only on names being strings and values being finite numeric scalars - it never checked the learner-specific key set or parameter domains.

`merge_partial_results` therefore accepted a shard claiming `learner="adamw"` with `step_size=-1.0`. The learner's own execution path rejects that step size, but the merged `IPMNISTRunResult` retained the impossible configuration, so downstream artifact construction could report it as the run's actual configuration - corrupting the benchmark's configuration attestation even though the run itself never used that step size.

```text
accepted hyperparameters = {'step_size': -1.0, 'beta1': 0.0, 'beta2': 0.99, 'eps': 1e-08, 'weight_decay': 0.0}
```

## Reachability

Reachable via the module's own CLI: `--merge-partials` reads user-supplied shard JSON files, through `main_v2_compat` -> `merge_partial_results` -> `_merge_partial_results` -> `_validated_partial_payload`. `merge_partial_results` is module-public but not re-exported from `alberta_framework.benchmarks.__init__`'s `__all__` (only `run_ipmnist`, the producing entry point, is).

fix incoming: route the imported hyperparameters through the module's own `resolve_hyperparameters` learner-specific validator (already used before execution) and require the shard to contain the complete resolved configuration.

Svector-anu:
Closes #978.

## What changed

`_validated_partial_payload` accepted v2 shard `hyperparameters` based only on names being strings and values being finite numeric scalars - never checking the learner-specific key set or parameter domains. `merge_partial_results` therefore accepted a shard claiming `learner="adamw"` with `step_size=-1.0`: the execution path itself rejects that step size, but the merged `IPMNISTRunResult` retained the impossible configuration, corrupting the benchmark's configuration attestation even though the run never actually used that step size.

## Fix

Route the imported hyperparameters through the module's own `resolve_hyperparameters` learner-specific validator (already used before execution), and require the shard to contain the complete resolved configuration rather than relying on omitted defaults during attestation.

## Reachability

Reachable via the module's own CLI: `--merge-partials` reads user-supplied shard JSON files, through `main_v2_compat` -> `merge_partial_results` -> `_merge_partial_results` -> `_validated_partial_payload`. `merge_partial_results` is module-public but not re-exported from `alberta_framework.benchmarks.__init__`'s `__all__` (only `run_ipmnist`, the producing entry point, is) - so it's reachable at the CLI/module level, not the top-level package API.

## Verification

```text
accepted hyperparameters = {'step_size': -1.0, 'beta1': 0.0, 'beta2': 0.99, 'eps': 1e-08, 'weight_decay': 0.0}
```

- new test `test_merge_rejects_hyperparameters_outside_learner_domain` in `TestPartialMerge`.
- full file: `110 passed`.

## Provenance

AI-assisted: initial bug identification, reproduction, and fix drafted by OpenAI Codex (`gpt-5.6-sol` via AgentRouter) running in a sandboxed worktree with no git-write access (so it could not commit itself). I independently re-verified before shipping: re-ran the reproduction and full mutation check myself from a clean stash, re-ran the full test file/lint/mypy, and re-traced reachability against `alberta_framework/benchmarks/__init__.py` and the CLI entry point before committing and pushing.

### DIFF
--- a/alberta_framework/benchmarks/upgd_ipmnist.py
+++ b/alberta_framework/benchmarks/upgd_ipmnist.py
@@ -1430,7 +1430,16 @@ def _validated_partial_payload(
             raise ValueError(f"{path}: hyperparameters must be finite named numbers")
         if not math.isfinite(float(value)):
             raise ValueError(f"{path}: hyperparameters must be finite named numbers")
-
+    try:
+        resolved_hyperparameters = resolve_hyperparameters(
+            learner, dict(hyperparameters)
+        )
+    except ValueError as exc:
+        raise ValueError(f"{path}: invalid hyperparameters: {exc}") from exc
+    if dict(hyperparameters) != resolved_hyperparameters:
+        raise ValueError(
+            f"{path}: hyperparameters must contain the complete learner configuration"
+        )
     config_payload = payload.get("config")
     if not isinstance(config_payload, Mapping) or set(config_payload) != set(
         IPMNISTConfig().to_config()
--- a/tests/test_upgd_ipmnist.py
+++ b/tests/test_upgd_ipmnist.py
@@ -1035,6 +1035,17 @@ class TestPartialMerge:
         with pytest.raises(ValueError, match="uint32"):
             merge_partial_results([path])
 
+    def test_merge_rejects_hyperparameters_outside_learner_domain(self, tmp_path) -> None:
+        data_x, data_y = _synthetic_dataset(6, N_TRAIN, TINY.input_dim, TINY.n_classes)
+        shard = run_ipmnist(data_x, data_y, "adamw", seeds=(0,), config=TINY)
+        payload = partial_payload(shard)
+        payload["hyperparameters"]["step_size"] = -1.0
+        path = tmp_path / "negative-step-size.json"
+        path.write_text(json.dumps(payload), encoding="utf-8")
+
+        with pytest.raises(ValueError, match="invalid hyperparameters"):
+            merge_partial_results([path])
+
     def test_v2_partial_is_single_seed_and_recursively_omits_legacy_marker(
         self, tmp_path, debug_run
     ):


kenjohnscreates (changes requested):
## Review — CHANGES_REQUESTED (one confirmed downstream regression in the screening lane; the fix itself is right)

Verified by re-applying the diff onto current `main` (`90c4613`) in a scratch worktree; both files **byte-identical to PR head `0df269c4`**.

## Confirmed regression: `tests/test_ipmnist_screening.py::TestShardsAndMerge::test_validate_proxy_prefix_and_ordering`

Passes on baseline, **fails with this diff applied**. `_validated_partial_payload` is not only used by `merge_partial_results` — it is imported by `alberta_framework/benchmarks/ipmnist_screening.py` and called at `ipmnist_screening.py:8860` by `validate_proxy` to read the pinned `outputs/upgd_ipmnist/partials/*.json` as the reference partials. That test tampers a reference partial's `learner` (`upgd_w` → `adamw`) and expects screening's own identity refusal:

```
Expected regex: 'does not match expected learner'
Actual message: ".../tampered-learner/upgd_w_seed0.json: invalid hyperparameters:
                unknown hyperparameters for adamw: ['noise_std', 'utility_decay']"
```

Your new `resolve_hyperparameters` call runs before screening's `reference["learner"] != learner` check and rejects first with a different message. Both refusals are fail-closed, so nothing becomes *unsafe* — but this changes an established error contract on the screening lane (the module the #51/#184 preregistration workflow's `load_shard` and result validation sit on) and the PR's battery (`test_upgd_ipmnist.py` only) could not see it. Two clean resolutions, your call: (a) update the screening test's expected message to accept either refusal (`"does not match expected learner|invalid hyperparameters"`), or (b) reorder so `_validated_partial_payload` performs the domain check after identity/schema checks and let callers keep their own learner-identity error — I'd lean (a) with a one-line note, since the earlier rejection is arguably the better behavior. Either way, please run `tests/test_ipmnist_screening.py` too (I know it's slow — `-k "proxy or Merge or Shard"` covers the consumers of this helper).

## What does verify

- **Defect is real and the fix is the right shape.** A v2 shard claiming `learner="adamw"` with `step_size=-1.0` was accepted into the merged `IPMNISTRunResult` and its configuration attestation, even though execution would have refused it; routing imported hyperparameters through the module's own `resolve_hyperparameters` (the same validator the run path uses) closes the gap, and the completeness equality prevents omitted-default drift in attestation.
- **Pinned artifacts unaffected — checked directly, not assumed.** All 20 `outputs/upgd_ipmnist/partials/*.json` are v1 schema; `merge_legacy_v1_partial_results` yields the identical `{adamw: 10, upgd_w: 10}` on baseline and PR trees, and both learners' stored hyperparameter sets equal `resolve_hyperparameters(learner, hp)` exactly, so even the v1 reference path clears the new completeness check. `tests/test_upgd_ipmnist_nonpromoting.py` (binds the pinned receipts) passes.
- Battery (arm64 M1): `test_upgd_ipmnist.py` + `test_upgd_ipmnist_nonpromoting.py` **134/134**; `ruff check .` clean; repo-config strict mypy clean at **168 source files**. Failing-test-first reproduced: tests-only against baseline → `test_merge_rejects_hyperparameters_outside_learner_domain` fails deterministically; full diff → green.
- Reachability statement is accurate (`--merge-partials` → `main_v2_compat` → `merge_partial_results`; module-public, not in the top-level `__all__`).

Happy to re-verify promptly once the screening test is reconciled — with that, this is a clean measurement-integrity fix on the benchmark's attestation path.
Compute receipt: 6348507 project-attributed tokens (exact; device-signed, locally reported)
— [kenjohnscreates]

lalalune:
Superseded by the patch-identical validation now on `main` as `5e4ad271` (merged through #1021). `git cherry` marks this head equivalent, so merging the stale/conflicting branch would add no behavior and risk removing newer surrounding hardening.

PR #1018, opened 2026-08-17
fix(core): saturate RLS reward model step_count instead of wrapping

### LINKED ISSUE #1017: reward_model.py RLSRewardModel step_count wraps to INT32_MIN instead of saturating
`RLSRewardModel.update` in `alberta_framework/core/reward_model.py` incremented `state.step_count` with plain addition on a signed int32 array. Once `step_count` reaches `INT32_MAX` (`2**31 - 1`), the next increment wraps to `INT32_MIN` (`-2147483648`) instead of staying bounded - a long-running diagnostic counter silently going negative.

```text
step_count after increment from INT32_MAX: -2147483648
```

Same int32-wraparound-instead-of-saturation class already fixed in `option_value_duration.py` and `canonical_upgd.py`.

## Reachability

`RLSRewardModel`/`RLSRewardModelConfig` are exported from `alberta_framework.core.__init__`'s `__all__` - publicly reachable with ordinary concrete inputs, and `step_count` increments on every call to `update`, so any sufficiently long-running model instance would eventually hit this.

fix incoming: add a `_saturating_int32_increment` helper (same pattern as the other two files) and route `step_count` through it.

Svector-anu:
Closes #1017.

## What changed

Same int32-wraparound-instead-of-saturation class already fixed in `option_value_duration.py` and `canonical_upgd.py`: `RLSRewardModel.update` incremented `state.step_count` with plain addition on a signed int32 array. Once `step_count` reaches `INT32_MAX`, the next increment wraps to `INT32_MIN` instead of staying bounded.

## Fix

Add `_saturating_int32_increment` (same pattern as the other two files) and route `step_count` through it.

## Reachability

`RLSRewardModel`/`RLSRewardModelConfig` are exported from `alberta_framework.core.__init__`'s `__all__` - publicly reachable, and `step_count` increments on every `update` call, so any sufficiently long-running model instance would eventually hit this.

## Verification

```text
step_count after increment from INT32_MAX: -2147483648
```

- new test `test_reward_model_step_count_saturates_at_int32_max`.
- mutation check: reverted just the source fix, reran the new test -> fails with the exact wraparound value (`-2147483648 == 2147483647` mismatch). restored -> passes.
- full file: `62 passed`.

## Provenance

AI-assisted: initial bug identification, reproduction, and fix drafted by OpenAI Codex (`gpt-5.6-sol` via AgentRouter). The Codex session hit an API quota exhaustion (`403: user quota is not enough`) partway through and never wrote its own report or committed - I recovered the real, uncommitted diff from the working tree and independently verified the entire thing from scratch with no report to lean on: re-reproduced the wraparound myself against the unpatched code, ran the full mutation check, full test file, lint, mypy, and reachability trace before committing and pushing.

### DIFF
--- a/alberta_framework/core/reward_model.py
+++ b/alberta_framework/core/reward_model.py
@@ -76,6 +76,13 @@ def _skip_zero_scale(scale: Array, value: Array) -> Array:
     return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
 
 
+def _saturating_int32_increment(value: Array) -> Array:
+    """Increment a diagnostic counter without signed-int32 wraparound."""
+    maximum = jnp.asarray(_INT32_MAX, dtype=jnp.int32)
+    counter = jnp.asarray(value, dtype=jnp.int32)
+    return jnp.minimum(jnp.maximum(counter, 0), maximum - 1) + 1
+
+
 @dataclass(frozen=True)
 class RLSRewardModelConfig:
     """Configuration for a linear recursive-least-squares reward model.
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #1024 after current-main consolidation, combined contract review, and retained contributor-authored commit(s) where selected. The aggregate passed the post-rebase focused panels, scoped Ruff, and strict mypy.

PR #1029, opened 2026-08-17
fix(pavlovian): complete protocol, resource, and runtime contracts

lalalune:
Supersedes #896 while preserving its contributor commit.

The review found gaps beyond its constructor checks: probability bounds were evaluated after lossy host conversion; duplicate/empty/overlapping CS sets could corrupt sampling or double activation duration; resource accounting omitted four resident phase vectors and the PRNG key; and an int32 final-phase clock wrapped. This branch adds exact-host plus float32 probability validation, canonical exact phase schemas, a verified 64 MiB numeric envelope, static state contracts, saturating clocks, and atomic neutral rollback for dynamically invalid states.

Verification: 18 focused validation tests + 102 retained Pavlovian tests; Ruff; strict mypy; diff-check. No outputs, benchmarks, or evidence sources changed.

### DIFF
--- a/alberta_framework/streams/pavlovian.py
+++ b/alberta_framework/streams/pavlovian.py
@@ -40,6 +40,7 @@ References
 from __future__ import annotations
 
 import math
+from fractions import Fraction
 from numbers import Real
 from typing import cast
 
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by concurrently merged #1028. The remaining exact-probability and state-boundary residuals are being split into a clean follow-up on current main.

PR #964, opened 2026-08-17
fix(recurring-feature): validate protocol and criteria fields and seed sequences

Kayvan-Zahiri:
## Summary
- Resolves #956.
- Hardens `RecurringFeatureProtocol.validate()`, `RecurringFeatureGateCriteria.__post_init__`, and `run_recurring_feature_gate` in `alberta_framework/recurring_feature_gate.py` to enforce strict integer types, float32 validation via `validated_float32_scalar`, exact boolean flags, and JAX seed sequence validation via `require_unique_jax_seeds`.
- Adds comprehensive unit tests in `tests/test_recurring_feature_protocol_validation.py`.

## Verification
- `PYTHONPATH=. pytest tests/test_recurring_feature_protocol_validation.py tests/test_recurring_feature_gate.py tests/test_recurring_feature_artifact.py`: 28 passed
- `ruff check .`: clean
- `mypy alberta_framework/recurring_feature_gate.py`: clean

PR #959, opened 2026-08-17
fix(recurring-feature): reject boolean protocol and criteria identities

ss251:
Closes #956.

## What changed

`RecurringFeatureProtocol.validate()` and the new `RecurringFeatureGateCriteria.validate()` now reject boolean and non-finite identities before the recurring-feature gate allocates arrays or compares evidence. Seed IDs go through `require_unique_jax_seeds`.

On origin/main `90c4613a5573de324a7bb33c89efd85e1bc09aa0`, protocol `validate()` used ordered comparisons only. `True` passed every positive-int / positive-float check. `minimum_seeds=False` made the seed-count gate vacuous. `True` was a legal JAX seed.

Legal zeros stay legal: `step_size_output=0.0`, `utility_decay=0.0`, `replacement_interval=0`, `min_feature_age=0`. Builtin legal floats stay bit-identical.

## Scope

- `alberta_framework/recurring_feature_gate.py`
- `tests/test_recurring_feature_protocol_validation.py`

## Why this is unowned

Live occupancy at publish time: neither target file is in the open ASI PR map. This is not a republish of `#893`/`#894`/`#898`/`#899`/`#901`/`#903`/`#904`/`#905`/`#908`/`#909`/`#912`/`#913`/`#916`/`#917`/`#924`/`#925`/`#930`/`#931`/`#934`/`#935`/`#946`/`#947`, Shaw `#890`–`#892`, byt61 `#895`–`#897`, or leftover tax on recurring start positions, interaction constructor scalars, micro-continual shard JSON, export bool identities, GVFSpec, multi-head, forager-cli, timing, label-emnist, smoke CLI, average-reward, upgd-ipmnist, metrics, nexting, experiments, security, IDBD, true-online-td, baseline-optimizers, or partial-observation.

## Verification

Exact candidate head: `a48ece8de04b977c0b6e4ffdb185aa5196108d9a`
Base: `90c4613a5573de324a7bb33c89efd85e1bc09aa0`

- Red on base: 123 failed, 7 passed (`DID NOT RAISE` / missing `validate`)
- `PYTHONPATH=<worktree> pytest tests/test_recurring_feature_protocol_validation.py -q -o addopts=""` → 130 passed
- `ruff check` on the two changed files: clean
- `git diff --check`: clean
- `scope-check.sh --max-files 2`: PASS

## Scientific integrity

[...]

PR #1030, opened 2026-08-17
fix(recurring-feature): complete protocol identity and resource contracts

lalalune:
Supersedes #959 and #964 while preserving the stronger #959 contributor commit. Protocol and criteria integers/booleans now require exact identities; all float32-consumed controls use shared exact host/sink validation and canonical storage; seeds use the shared unique JAX-seed boundary. The repair additionally preflights exact per-seed observation, target, task-index, and held-out array scalars/bytes against signed-int32 and 256 MiB before allocation. Verification on current main: 140 protocol/gate tests, scoped Ruff, strict source mypy, and diff-check pass. This source is registered for recurring_pair_features, so current pinned evidence remains fail-closed invalid until its frozen protocol is rerun; no outputs or hashes are modified.

### DIFF
--- a/alberta_framework/recurring_feature_gate.py
+++ b/alberta_framework/recurring_feature_gate.py
@@ -62,6 +62,46 @@ PAIRWISE_PROBE_SCOPE = (
 )
 
 
+def _require_builtin_int(
+    name: str,
+    value: object,
+    *,
+    minimum: int | None = None,
+    maximum: int | None = None,
+) -> int:
+    if type(value) is not int:
+        raise ValueError(f"{name} must be a built-in integer")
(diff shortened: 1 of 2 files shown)


lalalune:
Rebased onto merged #1021 and reconciled the overlapping validation rather than dropping the stronger residuals. #1030 now preserves exact/canonical float32 storage, the original cross-field decay relationships and exclusive upper bounds, exact declared strategy/aggregation/boolean identities, broader adversarial coverage, and the per-seed allocation preflight. Post-rebase: 140 tests passed; Ruff, strict source mypy, and diff-check clean.

PR #1153, opened 2026-08-17
fix(micro-continual): center whitened means to prevent catastrophic cancellation in bayes_predict

### LINKED ISSUE #1142: bayes_predict: catastrophic cancellation in expanded-form distance computation causes misclassification
`bayes_predict` in `alberta_framework/benchmarks/micro_continual.py` computes squared Mahalanobis distances with the expanded identity `||x||^2 - 2 x.mu + ||mu||^2` directly on uncentered float32 values. Large common offsets catastrophically cancel in that subtraction, changing class rankings despite every input remaining finite.

confirmed directly against the current unpatched code (an observation exactly matching one component mean, at a common offset of ~10000):

```text
predictions: [0]
```

Expected class 1 (the observation exactly matches component 1's mean, distance 0), got class 0 instead - a misclassification caused purely by numerical precision loss, not the actual statistical model.

## Reachability

This corrupts the Monte Carlo Bayes reference used by `bayes_reference`, which is called by `merge_micro_shards`'s benchmark summary path. The module function is also directly importable. `micro_continual` is not re-exported from the package root, so reachability is through the benchmark module rather than `alberta_framework.__init__` - no console script entry either.

fix incoming: whiten the component means, choose the first whitened mean as a shared origin, then subtract that origin from both samples and means before applying the same expanded formula. Squared Euclidean distance is translation invariant, so this removes the large common offset without altering the result.

Kayvan-Zahiri:
## Summary
- Resolves #1142.
- In `alberta_framework/benchmarks/micro_continual.py`, centers `whitened_x` and `whitened_means` around `origin = whitened_means[0:1]` before calculating squared Euclidean distances in expanded form. This preserves translation invariance while eliminating catastrophic cancellation when component means and observations share a large common offset.
- Adds unit tests in `tests/test_micro_continual.py`.

## Verification
- `PYTHONPATH=. pytest tests/test_micro_continual.py`: 222 passed
- `ruff check .`: clean
- `mypy alberta_framework/benchmarks/micro_continual.py`: clean

### DIFF
--- a/alberta_framework/benchmarks/micro_continual.py
+++ b/alberta_framework/benchmarks/micro_continual.py
@@ -698,9 +698,12 @@ def bayes_predict(component_means: Array, dim_sigma: Array, x: Array) -> Array:
     c, k, d = component_means.shape
     whitened_x = x / dim_sigma[None, :]
     whitened_means = (component_means / dim_sigma[None, None, :]).reshape(c * k, d)
-    cross = whitened_x @ whitened_means.T
-    x_norms = jnp.sum(whitened_x * whitened_x, axis=1)
-    mean_norms = jnp.sum(whitened_means * whitened_means, axis=1)
+    origin = whitened_means[0:1]
+    centered_x = whitened_x - origin
+    centered_means = whitened_means - origin
+    cross = centered_x @ centered_means.T
+    x_norms = jnp.sum(centered_x * centered_x, axis=1)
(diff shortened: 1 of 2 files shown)


lalalune:
Duplicate of #1143 (identical centering fix for the bayes_predict catastrophic-cancellation bug, filed first with a regression test pinning the exact failure case, now merged). Closing — thank you!

PR #1216, opened 2026-08-17
fix(core): reject off-policy TD discount factors outside [0, 1]

### LINKED ISSUE #1215: off_policy_td: gamma outside [0,1] silently accepted and applied
`OffPolicyTDLinearLearner`, `ETDLinearLearner`, and `GradientTDLinearLearner` in `alberta_framework/core/off_policy_td.py` validate the per-transition discount `gamma` with `jnp.isfinite(gamma_s)` only. Any finite value is accepted and applied, including negative discounts or discounts greater than 1.

Reproduction (current tree):

```text
OffPolicyTDLinearLearner -0.25 applied= True step= 1
OffPolicyTDLinearLearner 1.25 applied= True step= 1
ETDLinearLearner -0.25 applied= True step= 1
ETDLinearLearner 1.25 applied= True step= 1
GradientTDLinearLearner -0.25 applied= True step= 1
GradientTDLinearLearner 1.25 applied= True step= 1
```

A negative `gamma` flips the sign contribution of the bootstrapped next-state value in the TD target; `gamma > 1` causes eligibility traces (`decay = gamma * lambda`) and the TD target's bootstrap term to grow rather than decay over repeated updates, instead of converging. Both silently corrupt learning instead of failing the transaction closed, the same way every other malformed-input path in this update already does (see `jnp.isfinite` checks on `reward`, `rho`, `observation` in the same predicate).

Same bug class as `swift_td.py::SwiftTD.update` (fixed in #1110): an out-of-range TD discount factor silently accepted and applied rather than rejected, in a sibling off-policy TD implementation.

`OffPolicyTDLinearLearner` and `GradientTDLinearLearner` are exported from both `alberta_framework/core/__init__.py` and the top-level `alberta_framework/__init__.py`; `run_gradient_td_learning_loop` passes caller-supplied `gammas` arrays straight into `GradientTDLinearLearner._update_jit` via `jax.lax.scan`. `ETDLinearLearner` is public from its own module but not top-level re-exported.

Fix incoming in a linked PR: require `0 <= gamma <= 1` in the existing finite-input gate (IEEE comparisons already reject NaN, so the prior `isfinite` check becomes redundant once this lands).

Svector-anu:
Closes #1215.

## What changed

`OffPolicyTDLinearLearner`, `ETDLinearLearner`, and `GradientTDLinearLearner` in `alberta_framework/core/off_policy_td.py` validated the per-transition discount `gamma` with `jnp.isfinite(gamma_s)` only. Any finite value was accepted and applied, including negative discounts or discounts greater than 1.

## Fix

Require `0 <= gamma <= 1` in the existing finite-input gate for all three learners. IEEE comparisons already reject NaN, so the prior `isfinite(gamma_s)` check is now redundant and was removed in favor of the two bound checks. Rejected inputs leave state, traces, and step_count untouched, same as every other malformed-input path already handled in this predicate (`reward`, `rho`, `observation`).

Same bug class as `swift_td.py::SwiftTD.update` (#1110): an out-of-range TD discount factor silently accepted and applied instead of rejected, found here in a sibling off-policy TD implementation.

## Reachability

`OffPolicyTDLinearLearner` and `GradientTDLinearLearner` are exported from both `alberta_framework/core/__init__.py` and top-level `alberta_framework/__init__.py`. `run_gradient_td_learning_loop` passes caller-supplied `gammas` arrays directly into `GradientTDLinearLearner._update_jit` via `jax.lax.scan`, so both the direct per-call API and the scan-based loop are covered. `ETDLinearLearner` is public from its own module but not top-level re-exported.

## Verification

Live reproduction against the unpatched tree (observation `[1.0]`, reward `1.0`, `rho=1.0`):

```text
OffPolicyTDLinearLearner -0.25 applied= True step= 1
OffPolicyTDLinearLearner 1.25 applied= True step= 1
ETDLinearLearner -0.25 applied= True step= 1
ETDLinearLearner 1.25 applied= True step= 1
GradientTDLinearLearner -0.25 applied= True step= 1
GradientTDLinearLearner 1.25 applied= True step= 1
```

Added `test_update_rejects_discount_outside_unit_interval` to `tests/test_off_policy_td.py`, parametrized over all three learners and both `gamma in {-0.25, 1.25}`.

```text
FAILED tests/test_off_policy_td.py::test_update_rejects_discount_outside_unit_interval[-0.25-learner0]
FAILED tests/test_off_policy_td.py::test_update_rejects_discount_outside_unit_interval[-0.25-learner1]
FAILED tests/test_off_policy_td.py::test_update_rejects_discount_outside_unit_interval[-0.25-learner2]
FAILED tests/test_off_policy_td.py::test_update_rejects_discount_outside_unit_interval[1.25-learner0]
FAILED tests/test_off_policy_td.py::test_update_rejects_discount_outside_unit_interval[1.25-learner1]
FAILED tests/test_off_policy_td.py::test_update_rejects_discount_outside_unit_interval[1.25-learner2]
6 failed, 70 deselected in 0.95s
```

Fix restored: `6 passed, 70 deselected in 0.89s`.

Full touched test file: `75 passed, 1 failed` — the one failure (`test_config_scalars_canonicalize_reals_and_preserve_infinity_clip_sentinel`) is pre-existing and unrelated: on this Python 3.14 build, `np.longdouble("1e400")` overflows to `inf` during construction rather than staying a huge finite value, so the test's `retrace_clip`/`ratio_clip` canonicalization assertion no longer has the input it expects. Independently confirmed by running with the source fix fully reverted — same failure, same message, unaffected by this change.

## Provenance

AI-assisted: drafted by OpenAI Codex (`gpt-5.6-sol`, via AgentRouter), independently re-verified end to end before this PR was opened — reproduction, mutation check (both directions), full test file, lint, mypy, and the reachability trace above were all re-run and re-confirmed by hand, not taken from the draft's own report.


lalalune (approved):
Deep review: the range contract matches the repository-wide discount convention, composes with terminal next-observation handling, and preserves fail-closed state/output behavior. Verified on the exact head: 76 tests passed; scoped Ruff and strict mypy passed.

PR #1232, opened 2026-08-17
fix(prereg): invoke the exact setup-uv binary

lalalune:
## Outcome

Repairs the fail-closed pre-measurement stop in #51 run 32066770249. `setup-uv` installed uv 0.9.24 successfully, but the next step called bare `uv`; the hosted runner PATH resolved a different preinstalled binary, so the exact-version gate stopped before Python setup, dataset materialization, or any shard.

## Fix

- bind the pinned action step as `id: uv`
- require its documented `uv-path` output to be absolute and executable
- invoke that exact path at every one of the workflow uv call sites, including Python installation, sync, preflight, tests, dataset materialization, all shards/merge, validation, outcome display, and sealing
- add a workflow contract test that rejects all bare `uv` execution and pins the action output

## Failing-test-first evidence

On the unmodified base, the new path-binding test failed because the setup step had no id/output binding. After the repair: 126 workflow-driver tests passed; actionlint, Ruff, strict driver mypy, and diff check passed.

No benchmark, seed, dataset, or result artifact was created. The failed run uploaded only a 1,086-byte partial preflight artifact. A future run requires a new source/tag, a new exact owner authorization, and one new dispatch; rerunning attempt 1 is forbidden.

Refs #51

### DIFF
--- a/.github/workflows/ipmnist-prereg.yml
+++ b/.github/workflows/ipmnist-prereg.yml
@@ -161,6 +161,7 @@ jobs:
             --output "$RUNNER_TEMP/prereg-metadata/owner-authorization.json"
 
       - name: Set up exact uv 0.9.24
+        id: uv
         uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e
         with:
           version: 0.9.24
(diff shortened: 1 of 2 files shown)


lalalune:
Independent second-agent audit passed at exact head ed97a7f: all 13 executable uv sites bind setup-uv’s documented absolute uv-path; the new predicates fail on the parent; 126 workflow tests, actionlint, scoped Ruff, strict driver mypy, and diff check pass. The failed 20e9 source remains consumed and cannot be rerun.

PR #1270, opened 2026-08-17
fix(prototype-lifecycle): bind exact resource formulas

lalalune:
Corrective successor to #1259, which raced and merged before deep-review repairs could land. The original identity checks still used frozenset membership and could invoke an attacker-controlled metaclass hash; the purported exact budget also accepted inconsistent state, binding, template, owned, consumer, cache, router-call, and pair-product formulas plus mechanism/promotion drift. This uses identity-only trusted integer dispatch, exact record/status/no-promotion gates, and binds every derivable resource relationship to the live lifecycle layout. Validation before final rebase: 157 affected lifecycle/checkpoint/JIT tests passed; Ruff and strict mypy clean; initialized resource probes matched for default, zero-candidate, and reduced-slot configurations.

### DIFF
--- a/alberta_framework/core/prototype_feature_lifecycle.py
+++ b/alberta_framework/core/prototype_feature_lifecycle.py
@@ -75,8 +75,9 @@ PROTOTYPE_FEATURE_LIFECYCLE_SCIENTIFIC_PROMOTION_ALLOWED = False
 
 _CONFIG_TYPE = "PrototypeFeatureLifecycleConfig"
 _INT32_MAX = 2_147_483_647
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by merged #1273. That concurrent corrective includes the identity-only integer dispatch, exact record/status/promotion gates, and a stronger formula model that adds n_tasks/n_options/n_primitive_actions so input/output routing identities are derivable rather than trusted. Audited current main 6e025d9c; no residual production delta from this successor should be applied.

PR #1341, opened 2026-08-17
fix(benchmarks): add fail-closed post_init validation to Forager baseline, target, and state dataclasses

Kayvan-Zahiri:
## Summary

Adds fail-closed `__post_init__` boundary validation across Forager benchmark baseline, reference target, and causal feature state dataclasses in `alberta_framework/benchmarks/forager.py`:

- `ForagerFeatureState`: validates integer `last_action` (rejecting boolean subtypes), finite float for `last_reward`, and strict tuple of finite floats for `reward_traces`.
- `PaperBaseline`: validates non-empty strings for `name`, `family`, `state_construction`, and `source`, valid `role` literal, strict mapping for `selected_hyperparameters`, strict boolean for `in_tree_implementation`, and non-empty string or `None` for `official_config_path`.
- `PaperReferenceTarget`: validates non-empty strings across descriptor fields, finite real numbers for `central_estimate`, and strict boolean for `privileged`.

## Testing

- Added hostile input, invalid boolean subtype, non-finite float, and role validation tests in `tests/test_forager_dataclasses_hostile_validation.py`.
- Verified all tests pass; `ruff check .` clean; `mypy` clean.

### DIFF
--- a/alberta_framework/benchmarks/forager.py
+++ b/alberta_framework/benchmarks/forager.py
@@ -633,6 +633,17 @@ class ForagerFeatureState:
     last_reward: float
     reward_traces: tuple[float, ...]
 
+    def __post_init__(self) -> None:
+        if type(self.last_action) is not int or isinstance(self.last_action, bool):
+            raise ValueError("last_action must be an integer")
+        if not isinstance(self.last_reward, (int, float)) or not math.isfinite(self.last_reward):
+            raise ValueError("last_reward must be a finite float")
+        if type(self.reward_traces) is not tuple:
+            raise ValueError("reward_traces must be a tuple")
+        for trace in self.reward_traces:
+            if not isinstance(trace, (int, float)) or not math.isfinite(trace):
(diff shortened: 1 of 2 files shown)


lalalune:
Review verdict: fix needed before merge. This was the stronger superset of #1287, but #1287 merged first at the same insertion points, so the branch now conflicts. Also, the central_estimate check here uses isinstance(value, (int, float)), which accepts leftover bools (bool is an int subclass) — strictly weaker than the merged #1287 gate on the exact bug both target; the rebase must keep #1287's _require_real routing and add only the novel parts (string/role whitelist validation and ForagerFeatureState checks, minus the unreachable bool clause in the last_action check). I will rebase and land the salvageable parts.

lalalune:
Closing as inferior after consolidated review. #1287 already landed the genuine PaperBaseline/PaperReferenceTarget leftover-identity gates. This branch conflicts with that fix and adds incomplete broad Mapping/isinstance checks to internal state records, duplicating schema ownership without parser parity.

PR #1351, opened 2026-08-17
fix(streams): add fail-closed post_init validation to PavlovianPhase

Kayvan-Zahiri:
## Summary

Adds fail-closed `__post_init__` boundary validation to `PavlovianPhase` in `alberta_framework/streams/pavlovian.py`:

- Validates non-empty string for phase `name`.
- Validates positive integer (rejecting bool) for `n_steps`.
- Validates finite float probability in `[0.0, 1.0]` for `cs_us_contingency`.
- Enforces strict `tuple` type for `cs_active`.
- Validates integer $\ge -1$ (rejecting bool) for `compound_index`.

## Testing

- Added hostile input, invalid probability range, negative step count, and non-tuple active CS rejection tests in `tests/test_pavlovian_phase_hostile_validation.py`.

### DIFF
--- a/alberta_framework/streams/pavlovian.py
+++ b/alberta_framework/streams/pavlovian.py
@@ -226,6 +226,31 @@ class PavlovianPhase:
     cs_active: tuple[int, ...]
     compound_index: int = -1
 
+    def __post_init__(self) -> None:
+        if type(self.name) is not str or not self.name:
+            raise ValueError("name must be a non-empty string")
+        if (
+            type(self.n_steps) is not int
+            or isinstance(self.n_steps, bool)
+            or self.n_steps <= 0
+        ):
+            raise ValueError("n_steps must be a positive integer")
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after deep review: this adds a second, incomplete validation boundary to an inert descriptor that is already fully validated and canonicalized before every execution sink by ClassicalConditioningStream. The new post-init rejects supported exact Fraction/NumPy probability inputs via isinstance, omits cs_active element/non-empty/duplicate validation and int32 bounds, and changes retained constructor-then-stream rejection behavior (including creating invalid phase fixtures for stream-boundary tests). A correct standalone boundary would need to reuse the existing exact host+float32 helper and duplicate the stream context checks, which would create two drifting sources of truth. The better approach is to keep PavlovianPhase as a passive record and retain the single comprehensive stream admission boundary.

PR #1352, opened 2026-08-17
fix(evaluation): add fail-closed post_init validation to EvidenceSpec

Kayvan-Zahiri:
## Summary

Adds fail-closed `__post_init__` boundary validation to `EvidenceSpec` in `alberta_framework/evaluation/evidence_manifest.py`:

- Validates non-empty strings across `name`, `claim_scope`, and `expected_schema`.
- Enforces strict `bool` type for `promotes_scientific_claim`.
- Enforces strict `Path` instance for `relative_path`.
- Enforces strict `tuple` types for `command_argv`, `limitations`, `required_environment_fields`, and `source_paths`.
- Enforces that all items within `source_paths` are `Path` instances.
- Validates `Mapping` types across `protocol`, `configuration`, `seeds`, and `thresholds`.
- Validates that `loader` and `validator` are callables.

## Testing

- Added hostile input, invalid boolean type, non-path string, and non-callable rejection tests in `tests/test_evidence_spec_hostile_validation.py`.

### DIFF
--- a/alberta_framework/evaluation/evidence_manifest.py
+++ b/alberta_framework/evaluation/evidence_manifest.py
@@ -377,6 +377,36 @@ class EvidenceSpec:
     loader: Callable[[Path], Mapping[str, object]]
     validator: Callable[[Mapping[str, object]], ValidationResult]
 
+    def __post_init__(self) -> None:
+        for attr in ("name", "claim_scope", "expected_schema"):
+            val = getattr(self, attr)
+            if type(val) is not str or not val:
+                raise ValueError(f"{attr} must be a non-empty string")
+        if type(self.promotes_scientific_claim) is not bool:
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after deep review: EvidenceSpec is an internal executable registry descriptor assembled by source code, not a deserialization/adoption boundary. Adding partial runtime validation here does not protect an untrusted sink and creates a second schema that is incomplete: command_argv/limitations/required_environment_fields elements are unchecked, Path subclasses and Mapping subclasses/spoofs are admitted, mapping contents are unconstrained, relative/absolute path identity is unchecked, and callable() does not establish the loader/validator contract. The strict evidence boundary is artifact loading plus each registered validator; duplicating an incomplete constructor boundary would add drift without improving fail-closed evidence behavior.

PR #1357, opened 2026-08-17
fix(benchmarks): add fail-closed post_init validation to SelectionPlan, EvaluationPanel, and MatchedHypothesis

Kayvan-Zahiri:
## Summary

Adds fail-closed `__post_init__` boundary validation across selection plan and hypothesis dataclasses in `alberta_framework/benchmarks/forager_matched_protocol.py`:

- `SelectionPlan`: Validates metric string, canonical SHA-256 digests, maximize direction, valid statistic literal, finite confidence float in $(0.0, 1.0)$, positive integer bootstrap resamples, integer seed, valid RNG identity/resampling/quantile/interval literals, and strict tuple of `SelectionGroup` instances.
- `EvaluationPanel`: Validates tuple of `SelectionSlot` instances, slot assignments, strict `True` for `require_complete_blocks`, and `fail_closed` policy literal.
- `MatchedHypothesis`: Validates non-empty `hypothesis_id`, intervention and comparator `SelectionSlot` instances, `paired_mean_difference` estimand, `AnalysisMethod` literal, `greater` alternative, difference order, and strict `True` for `paired`.

## Testing

- Added hostile input, invalid confidence range, invalid direction, and non-paired rejection tests in `tests/test_forager_matched_selection_plan_hostile_validation.py`.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_protocol.py
+++ b/alberta_framework/benchmarks/forager_matched_protocol.py
@@ -576,6 +576,58 @@ class SelectionPlan:
     tie_break: Literal["candidate_id_ascending"]
     groups: tuple[SelectionGroup, ...]
 
+    def __post_init__(self) -> None:
+        if type(self.metric) is not str or not self.metric:
+            raise ForagerMatchedProtocolError("metric must be a non-empty string")
+        _require_sha256(self.metric_implementation_sha256, "metric_implementation_sha256")
+        _require_sha256(self.candidate_universe_sha256, "candidate_universe_sha256")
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after deep review. The proposed post-init layer duplicates the authoritative selection-plan validator but is not total: Literal comparisons/membership and isinstance-based numeric checks can run hostile hooks, floats are not checked in the exact host+execution domains, tuple members use spoofable isinstance, and slot/group uniqueness and cross-field identities are omitted. A partial second schema would drift from the sealed-protocol boundary.

PR #1359, opened 2026-08-17
fix(benchmarks): add fail-closed post_init validation to analysis plan dataclasses

Kayvan-Zahiri:
## Summary

Adds fail-closed `__post_init__` boundary validation across analysis plan dataclasses in `alberta_framework/benchmarks/forager_matched_protocol.py`:

- `PrimaryBootstrapAnalysis`: Validates method literal, positive integer resamples (rejecting bool), integer seed, finite confidence in $(0.0, 1.0)$, finite `primary_margin`, valid RNG and quantile literals, and canonical SHA-256 implementation digest.
- `SecondarySignFlipAnalysis`: Validates method literal, positive integer Monte Carlo resamples, integer seed, exact max pairs of 20, PCG64 algorithm, canonical SHA-256 digest, greater alternative, Holm multiplicity method, and finite familywise alpha in $(0.0, 1.0)$.
- `MatchedAnalysisPlan`: Validates non-empty metric string, canonical SHA-256 digest, maximize direction, and strict instance types for primary and secondary analysis records.

## Testing

- Added hostile input, invalid method literal, zero resample count, and non-canonical probability rejection tests in `tests/test_forager_matched_analysis_plan_hostile_validation.py`.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_protocol.py
+++ b/alberta_framework/benchmarks/forager_matched_protocol.py
@@ -677,6 +677,38 @@ class PrimaryBootstrapAnalysis:
     implementation_sha256: str
     gate: Literal["lower_bound_strictly_greater_than_margin"]
 
+    def __post_init__(self) -> None:
+        if self.method != "paired_percentile_bootstrap_lower_bound":
+            raise ForagerMatchedProtocolError("invalid method")
+        if (
+            type(self.resamples) is not int
+            or isinstance(self.resamples, bool)
+            or self.resamples <= 0
+        ):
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after deep review. These internal analysis-plan records are already validated when the protocol is loaded/sealed. The new layer uses hookable equality/isinstance numeric admission, does not canonicalize or validate exact float domains, leaves seed/range/resource relationships incomplete, and uses spoofable nested-record checks. It is not a safe standalone trust boundary and would duplicate the authoritative validator.

PR #1361, opened 2026-08-17
fix(benchmarks): add fail-closed post_init validation to _BatchPlan and _SourceSnapshot

Kayvan-Zahiri:
## Summary

Adds fail-closed `__post_init__` boundary validation across batch plan and snapshot dataclasses in `alberta_framework/benchmarks/forager_matrix.py`:

- `_BatchPlan`: Validates non-empty variant ID string, non-negative integer batch index (rejecting bool), and non-negative integer seeds tuple.
- `_SourceSnapshot`: Validates raw archive bytes, canonical 64-character lowercase hex digest strings for archive, tree, and inventory SHA-256 digests, matching digest verification against `hashlib.sha256(archive_bytes)`, and Mapping inventory instance type.

## Testing

- Added hostile input, invalid variant ID, negative batch index, bool seed rejection, digest mismatch, and non-canonical digest rejection tests in `tests/test_forager_matrix_batch_plan_hostile_validation.py`.
- Verified unit tests pass; `ruff check .` clean; `mypy` clean.

### DIFF
--- a/alberta_framework/benchmarks/forager_matrix.py
+++ b/alberta_framework/benchmarks/forager_matrix.py
@@ -468,6 +468,21 @@ class _BatchPlan:
     batch_index: int
     seeds: tuple[int, ...]
 
+    def __post_init__(self) -> None:
+        if type(self.variant_id) is not str or not self.variant_id:
+            raise ForagerMatrixError("variant_id must be a non-empty string")
+        if (
+            type(self.batch_index) is not int
+            or isinstance(self.batch_index, bool)
+            or self.batch_index < 0
+        ):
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after deep review. _BatchPlan and _SourceSnapshot are private products of the matrix builder, not independent deserialization boundaries. The patch leaves seed/int32/batch-resource relationships incomplete and accepts a spoofed Mapping inventory; duplicating only a subset of the builder and snapshot verification in post-init creates a second drifting schema. Snapshot bytes/hashes are already checked at the publication/verification boundary.

PR #1362, opened 2026-08-17
fix(benchmarks): add fail-closed post_init validation to _BoundPathComponent and _BoundDirectory

Kayvan-Zahiri:
## Summary

Adds fail-closed `__post_init__` boundary validation across bound directory dataclasses in `alberta_framework/benchmarks/forager_matrix.py`:

- `_BoundPathComponent`: Validates integer descriptors, device, inode (rejecting bool), and non-empty name string.
- `_BoundDirectory`: Validates Path instance type, integer root descriptor, device, inode, tuple of `_BoundPathComponent` bindings, and `None` or 2-integer tuple `lock_identity`.

## Testing

- Added hostile input, invalid descriptor, non-path object rejection, and invalid lock identity tests in `tests/test_forager_matrix_bound_directory_hostile_validation.py`.
- Verified unit tests pass; `ruff check .` clean; `mypy` clean.

### DIFF
--- a/alberta_framework/benchmarks/forager_matrix.py
+++ b/alberta_framework/benchmarks/forager_matrix.py
@@ -1927,6 +1927,14 @@ class _BoundPathComponent:
     device: int
     inode: int
 
+    def __post_init__(self) -> None:
+        for attr in ("parent_descriptor", "child_descriptor", "device", "inode"):
+            val = getattr(self, attr)
+            if type(val) is not int or isinstance(val, bool):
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after deep review. These are private OS-binding records created only after descriptor/path binding checks. The proposed duplicate layer accepts Path subclasses and spoofed nested components through isinstance, allows invalid negative descriptors/device/inode values, and does not establish the live fstat/path invariants enforced by assert_bound. Constructor checks would therefore give a false fail-closed guarantee while duplicating the real OS boundary.

PR #1453, opened 2026-08-18
fix(workflow): sync prereg test dependencies

lalalune:
Attempt 4 failed closed before MNIST/data/seeds because the workflow synchronized only the `research` extra, then invoked pytest from the `dev` extra.

This adds the committed locked `dev` extra to the same exact uv sync. It does not loosen the lock, interpreter, uv digest, authorization, runner, namespace, or source gates.

Failing-test-first verification:
- new workflow contract test fails on the parent because `--extra dev` is absent
- 231 local/workflow prereg tests pass
- exact offline `uv sync --locked --extra dev --extra research` into a fresh environment installs pytest 9.1.1 and JAX 0.11.0; `python -m pytest --version` passes
- actionlint clean
- Ruff clean
- strict prereg driver mypy clean
- diff-check clean

No benchmark, output, tag, authorization, or dispatch is created by this PR. Attempt 4 failure is recorded on #51.

### DIFF
--- a/.github/workflows/ipmnist-prereg.yml
+++ b/.github/workflows/ipmnist-prereg.yml
@@ -219,6 +219,7 @@ jobs:
           "$UV_PATH" sync \
             --locked \
             --python "$PYTHON_PATH" \
+            --extra dev \
             --extra research
           "$UV_PATH" run --no-sync python - \
             "$cpu_brand" "$RUNNER_TEMP/prereg-metadata/runner.json" <<'PY'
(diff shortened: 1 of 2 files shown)


lalalune:
Closed as superseded by merged #1454. #1454 contains the same locked dev+research sync correction and additionally verifies the uv.lock-selected pytest 9.1.1 identity while keeping test tooling outside the measurement receipt. Its 232 prereg tests, actionlint, Ruff, and cold driver mypy passed. No contributor delta remains unique here.

PR #1503, opened 2026-08-18
fix(benchmarks): reject hostile string identities in IPMNIST screening

Kayvan-Zahiri:
Closes #1496

## What changed

- `_required_nonempty_string(value)`: `isinstance(value, str)` → `type is not str` (before `or not value`) — used for provider/name/precision/learner strings
- `_validated_screening_noise_mode(noise_mode)`: `isinstance(noise_mode, str)` → `type is not str` (before `in ("step", "pool")`)
- `shard_payload` `result.base_learner`: `isinstance(result.base_learner, str)` → `type is not str` (before `or not result.base_learner`)

All use exact-type checks before `in`/`bool`, failing closed with 0 hook calls.

## Why

These are external trust boundaries (screening lane configs, noise contracts, shard payloads deserialized from strict JSON). `isinstance(str)` admits hostile `str` subclasses that overload `__bool__`/`__eq__` and dispatch during `or not value` or `in` before validation. `type is str` is the established hostile-identity pattern.

## Verification

- `ruff check .` — clean
- `mypy alberta_framework/benchmarks/ipmnist_screening.py` — clean
- No benchmark, `outputs/**`, or evidence promotion.

## Contribution provenance

- AI assistance: yes
- AI provider/model: Google / Gemini

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -6717,7 +6717,7 @@ def _validated_screening_noise_mode(
 ) -> str:
     """Validate one screening noise mode against the named arm's runner contract."""
     prefix = "" if context is None else f"{context}: "
-    if not isinstance(noise_mode, str) or noise_mode not in ("step", "pool"):
+    if type(noise_mode) is not str or noise_mode not in ("step", "pool"):
         raise ValueError(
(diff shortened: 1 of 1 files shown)


lalalune:
Superseded by merged consolidation #1507, which preserves this contribution and authorship while adding failing-test-first coverage and repairing adjacent exact-identity/cross-field gaps.

PR #1526, opened 2026-08-18
fix(validation): reconcile protocol string identities

lalalune:
Supersedes #1515 and #1522 while preserving both contributor commits and authorship; adds regression coverage for the already-consolidated #1514 adapter gates.

## Scope

- exact string identities in the Prototype reference adapter, matched protocol, and matched open-protocol layers
- closes matched-protocol residual hook paths for JSON string values, direct decoder inputs, and both top-level parse entry points
- failing-test coverage proves hostile truth/length/equality/hash/encode/strip hooks remain at zero calls

## Verification

- 140 full relevant tests pass on current main
- Ruff clean across all changed source/tests
- strict mypy clean (174 files)
- diff-check clean

No workflows, benchmarks, outputs, or evidence promotion.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_open_protocol.py
+++ b/alberta_framework/benchmarks/forager_matched_open_protocol.py
@@ -1016,7 +1016,7 @@ _UPSTREAM_CONFIGURATION_REQUIREMENTS: Final = MappingProxyType(
 
 
 def _require_real_sha256(value: str, path: str) -> None:
-    if not isinstance(value, str) or _SHA256_RE.fullmatch(value) is None:
+    if type(value) is not str or _SHA256_RE.fullmatch(value) is None:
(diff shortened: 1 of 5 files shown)


lalalune:
Superseded for protocol/open-protocol source changes by merged #1524, which independently lands the same contributor-preserving corrections and residual decoder coverage. The nonduplicative Prototype adapter regression test will be separated to avoid conflicting duplicate source edits.

lalalune:
Correction after checking the full current test inventory: the Prototype adapter coverage is already merged in #1521 (`tests/test_prototype_adapter_hostile_validation.py`, all five exact string gates). Therefore #1526 has no unique remaining source or test coverage and is fully superseded by #1524 + #1521; no follow-up PR is needed.

PR #1592, opened 2026-08-18
ci: resolve the Windows Python executable

lalalune:
## Summary

- resolve the exact interpreter path through Python's `sys.executable`
- pass the resulting `.exe` path to both locked `uv sync` phases
- preserve the existing Unix behavior

## Root cause

The new Windows portability job reached the shared environment setup action, but Git Bash's `command -v python` returned an extensionless path. `uv` correctly rejected that nonexistent path because the installed interpreter is `python.exe`.

## Validation

- the composite action and CI workflow parse as YAML
- the interpreter-resolution command returns an executable path locally
- macOS portability already passed end to end in PR #1557
- this follow-up is intended to rerun the Windows portability lane

### DIFF
--- a/.github/actions/setup-locked-python/action.yml
+++ b/.github/actions/setup-locked-python/action.yml
@@ -36,13 +36,14 @@ runs:
         if [[ -n "$ALBERTA_UV_EXTRAS" ]]; then
           read -r -a extra_args <<< "$ALBERTA_UV_EXTRAS"
         fi
+        python_path="$(python -c 'import sys; print(sys.executable)')"
         uv sync \
           --no-install-project \
           --locked \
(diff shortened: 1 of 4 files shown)


lalalune:
Windows rerun update: locked setup, JAX CPU/JIT, and the focused regression panel all pass after resolving `sys.executable`. The installed smoke step then exposed a second native-Windows import boundary: `asi-reference-life-scorecard --help` imported the Linux-only `fcntl` checkpoint lock. Commit df0e805 makes checkpoint publication fail closed only when invoked on Windows while allowing the scorecard/model modules to import. Ruff and strict mypy pass; the updated Windows lane is rerunning.

PR #1600, opened 2026-08-18
feat(evaluation): stage optimizer geometry controls

lalalune:
Implements the bounded small-matrix stage for #1572.

- pins Muon-OGD v2, FOGO v1, and FLAD v1
- adds protected-direction orthogonal correction, Muon-style Newton–Schulz matrix sign, and FLAD gradient-aligned/noise decomposition primitives
- includes empty-basis and zero-gradient exact reductions
- explicitly limits scope to small streaming matrices before any IPMNIST integration
- predeclares matching/resource accounting and nonpromotion

No benchmark/output mutation occurred. Full paper algorithm ports, frozen matrix stream/seeds, and authorized empirical screening remain separate gates.

Validation: 4 focused tests; Ruff; strict mypy.

Closes #1572.

Issue #1572, opened 2026-08-18
research(optimizers): evaluate Muon-OGD, FOGO, and FLAD geometry controls


lalalune:
### Objective

Stage orthogonal, spectral, and sharpness-aware updates through a small streaming matrix test before IPMNIST.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2605.08949
- https://arxiv.org/abs/2606.10406
- https://arxiv.org/abs/2601.07636

lalalune:
Small-matrix equation primitives merged in #1607. The issue remains open: audit/pin reference code, complete method-faithful Muon-OGD/FOGO/FLAD adapters, freeze the streaming matrix protocol and reductions with exact resources/seeds/report schema, run it under explicit authorization, and only then decide whether any arm advances to IPMNIST.

lalalune:
Reopened after #1657 auto-closed this issue by parsing a negated phrase. #1657 retains a valid eight-update, tiny-matrix, permanently nonpromoting mechanism diagnostic with exact source/result identity; it is not the requested IPMNIST evaluation or a complete Muon-OGD/FOGO/FLAD port. The broader matched runner, full paper mechanisms, resource comparison, and retained IPMNIST result remain open.

lalalune:
Completed by the merged geometry implementation and retained development result through #1621, #1649, #1657, and runtime-bound corrective #1668. The final lane pins the three paper revisions/protocol differences, has failing-first parity and mechanism-off coverage, matches the frozen seed/stream/update/boundary schedule, consumes transaction validity, reports exact persistent numeric bytes plus data/model/timing telemetry, and appends a strict current-source/runtime-bound nonpromoting result under outputs/optimizer_geometry/development.v1. The result is a small streaming-matrix mechanism diagnostic only—not IPMNIST performance, scientific evidence, or optimizer superiority.

PR #1621, opened 2026-08-18
feat(evaluation): complete optimizer geometry stage

lalalune:
Replaces the closed helper-only #1600 with an executable, validated development diagnostic for #1572.

- corrects Muon-OGD to the pinned v2 cubic NS5 and bounded dual loop
- makes FLAD zero-gradient behavior finite under autodiff
- freezes a Threefry streaming-matrix protocol with matched candidate/control arms
- validates deterministic replay, exact builtin schemas and policies, comparison axes, metrics, persistent-state bytes, steps, queries, and timing telemetry
- retains improved/worse/tied outcomes and permanently forbids promotion
- records exact FOGO/FLAD adaptations and omitted paper mechanisms
- preserves the host, bounded-input, finite-kernel, JIT-validity, and resource-field hardening from #1611

This is a bounded mechanism diagnostic, not a complete working-set measurement, comparative performance result, or SOTA evidence. #1572 remains open for the explicitly recorded full ports and IPMNIST evaluation gates.

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -231,8 +231,12 @@ def muon_ogd_dual_update(
         raise ValueError("dual must contain one multiplier per constraint")
     if type(dual_learning_rate) is not float or not math.isfinite(dual_learning_rate):
         raise ValueError("dual_learning_rate must be a finite float")
-    if dual_learning_rate < 0.0 or type(dual_steps) is not int or dual_steps < 1:
-        raise ValueError("dual learning rate must be non-negative and dual_steps positive")
+    if (
+        dual_learning_rate < 0.0
+        or type(dual_steps) is not int
+        or not 1 <= dual_steps <= 32
+    ):
+        raise ValueError("dual learning rate must be non-negative and dual_steps bounded positive")
     for _ in range(dual_steps):
         shifted = value + jnp.einsum("k,kij->ij", multipliers, protected)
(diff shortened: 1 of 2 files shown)


lalalune:
Deep audit blocker before merge: the public result validator is not fail-closed at its trust boundary. `_mapping`/`_sequence`, `validate_streaming_matrix_result`, and `_exact_equal` call `isinstance`, `set`, iteration, equality, and indexing on untrusted containers/values before exact builtin-type admission. A hostile Mapping/Sequence or runtime metaclass can execute hooks or escape with arbitrary exceptions. Please gate exact `dict`/`list`/builtin scalar identities before all schema/set/equality work and add zero-hook regressions for top-level mapping, nested mapping/list, string fields, and numeric fields.

Also please reconcile this branch with #1611: it is based before the resource/host/JIT hardening and replaces the same optimizer_geometry module, so merging it now would drop those corrections. Persistent-only bytes are fine if the record says persistent bytes, but the PR/body should not call that a complete working-set resource receipt unless transient stream/stack/matrix buffers are included. Holding merge pending rebase + trust-boundary tests/formula wording.

lalalune:
Maintainer repair complete: rebased/merged current main, preserved #1611 hardening, restored the pinned v2 cubic NS5, bounded the dual loop, made the result boundary exact-dict/list/scalar only, and added zero-hook regressions for hostile top-level/nested containers plus string/numeric subclasses. Local Linux validation: 169 optimizer-geometry/optimizer/comparator tests passed; focused Ruff and mypy passed. Resource wording now explicitly says persistent-state bytes plus counters/timing telemetry, not complete working set.

PR #1625, opened 2026-08-18
fix(optimizers): harden constructor scalars without breaking frozen LMS

lalalune:
Corrective follow-up to merged #1622. The contributor commit/authorship is already preserved on main.

This repair restores the established `LMS(step_size=0.0)` frozen-weight control used by the real forward-view equivalence path, while retaining exact rejection of bool/negative/nonfinite values. It also keeps the exact positive finite `Autostep.tau` divisor gate and expands zero-hook hostile-scalar, canonical NumPy-scalar, and gamma/lambda boundary coverage across all touched constructors.

Verification on current main: `tests/test_optimizers.py tests/test_forward_view_equivalence.py` 140 passed; scoped Ruff passed; strict mypy on `optimizers.py` passed; diff check passed. No outputs or benchmark execution.

### DIFF
--- a/alberta_framework/core/optimizers.py
+++ b/alberta_framework/core/optimizers.py
@@ -591,10 +591,10 @@ class LMS(Optimizer[LMSState]):
             step_size: Fixed learning rate
 
         Raises:
-            ValueError: If ``step_size`` is not a finite non-bool positive real.
+            ValueError: If ``step_size`` is not a finite non-bool nonnegative real.
         """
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #1622 (which preserved ss251's original contribution and added tau validation) plus focused corrective #1628 for the remaining LMS(0.0) compatibility regression and hostile/range coverage.

PR #1645, opened 2026-08-18
fix(ci): keep reference scorecard manual-only

lalalune:
Removes the accidental pull-request trigger from the long 144-process development scorecard. The campaign now runs only through explicit workflow dispatch, retains temporary Actions artifacts for 90 days, and includes the executed source SHA in the aggregate artifact name. A successful run still requires independent validation and append-only repository integration before #1585 can close.

### DIFF
--- a/.github/workflows/reference-life-scorecard-dev.yml
+++ b/.github/workflows/reference-life-scorecard-dev.yml
@@ -2,9 +2,6 @@ name: reference-life development scorecard
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


lalalune:
Superseded by merged #1644 for the manual-only trigger correction plus contributor-preserving current-main successor #1646 for this PR’s unique 90-day retention and source-SHA artifact naming. Closing the now-overlapping stale branch; no contribution is lost.

PR #1646, opened 2026-08-18
ci: retain source-bound reference scorecard artifacts

lalalune:
Contributor-preserving current-main successor to #1645 after #1644 already removed the accidental automatic trigger. Keeps only the unique improvements: 90-day retention for shard and aggregate artifacts, and the exact checkout SHA in the aggregate artifact name. The workflow remains manual-only and #1585 remains open pending an intentional successful run, independent validation, and append-only integration. Verification: actionlint, 2 workflow regressions, diff check. No workflow dispatched and no outputs mutated.

### DIFF
--- a/.github/workflows/reference-life-scorecard-dev.yml
+++ b/.github/workflows/reference-life-scorecard-dev.yml
@@ -65,7 +65,7 @@ jobs:
           name: reference-life-shards-${{ matrix.seed }}
           path: scorecard/shards/*.json
           if-no-files-found: error
-          retention-days: 30
+          retention-days: 90
 
   aggregate:
     name: aggregate and validate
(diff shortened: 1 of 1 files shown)


lalalune:
Closing as exactly superseded by concurrently merged #1645 (merge f33617a2), which now includes the same 90-day retention and source-SHA artifact naming atop #1644’s manual-only trigger regression. This successor preserved the same contributor commit, so no unique delta remains.

lalalune:
Superseded by merged #1645: current main already has workflow_dispatch-only, 90-day retention for shard and aggregate artifacts, and the exact checkout SHA in the aggregate artifact name. Closing the now-empty duplicate rather than merging redundant history.

Issue #1570, opened 2026-08-18
research(ipmnist): audit and implement BiMU as a protocol-extended arm


lalalune:
### Objective

Implement BiMU in a separately labeled binary/Bayesian lane with its late-window metric kept distinct.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2605.30198

lalalune:
Equation/protocol prerequisite merged in #1607. The issue remains open: audit and pin reference code, implement/validate the Concrete/Gumbel estimator and full binary Bayesian network, freeze the 1000-task lane/hyperparameters/seeds/resources and distinct last-five-task report, then obtain explicit execution authorization and retain the nonpromoting result.

hermesagent270-commits:
**Appears already implemented.** BiMU is in `alberta_framework/benchmarks/bimu.py` and `alberta_framework/evaluation/bimu_matched_nonpromoting.py`, covered by `tests/test_bimu.py` and `tests/test_bimu_matched_nonpromoting.py`.

Verified by running the listed test file(s) locally (part of a 270-test batch across all these mechanism suites, all passing). This looks like a **candidate to close** — maintainers, please confirm it matches the intended scope. Flagging as triage; happy to be corrected if I have the wrong module.

huskfc:
🤖 An autonomous contributor agent is starting work on this issue.

Bounded outcome: a PR against main from branch fix/1570-bimu-protocol-extended-arm (fork huskfc), with tests and evidence bound to the measured head.

Defer contract: if you claimed, were assigned, or had started this issue before this comment — or you are the issue reporter and intend to implement it yourself — reply here and this claim is released promptly. Otherwise this claim holds first-come priority and a PR will follow shortly.

AI provider/model: anthropic / claude-3.5-sonnet
Client / agent tooling: codex
Contribution skill revision: SlopDotCash/slopdotcash@a2fe88318864e11c72cfa96cbfe4215a166e608b:skills/contribute-to-asi
Attribution status: self-reported
— [eliza-swarm]

PR #1647, opened 2026-08-18
feat(benchmarks): add end-to-end BiMU development lane

lalalune:
Repairs the helper-only BiMU prerequisite with a callable development runner for #1570.

Implemented in this PR:
- pins arXiv:2605.30198v1 and official CC-BY-4.0 code commit 1b8a1a1fb892fbe89401390b3ff9611d7f3a5168
- implements binary posterior sampling, the Concrete relaxation and gradient correction, the pinned BiMU update scaling, and the 1000-task/100-unit hyperparameters
- uses an explicitly threefry2x32, nested domain/index RNG schedule with tested separation between initialization, permutations, example order, query, training, and test draws
- matches the pinned code inference decision: argmax of mean log-softmax across posterior samples; variation ratio remains a separate query statistic
- provides a callable, boundary-private Permuted-MNIST loop over caller-supplied standardized arrays and reports late-five and whole-stream pre-update metrics separately
- records the optimizer step/seen distinction, resource counters, and nonpromoting policy
- validates the strict receipt schema and metrics derived from reported integer sufficient statistics

Receipt limits are explicit: dataset, schedule, and state digests are producer-reported consistency identities and are not recomputed by the compact validator; metrics are not replayed from an execution transcript; the receipt is not authenticated execution attestation.

This remains nonpromoting and not paper-comparable. Remaining gaps include binding the official MNIST/preprocessing/dataloader, the local versus official RNG schedule, added pre-update probes, five-run aggregation, matched controls, and the float32 Concrete clipping difference. This PR does not add a CLI or execute the 1000-task campaign, so #1570 remains open for runner integration, a frozen seed/resource plan, matched execution, and retained results.

Validation on current main: 25 focused tests passed; focused Ruff, strict mypy, and diff check passed. No output and no performance or SOTA claim.


lalalune:
Deep review found and repaired a material metric-parity defect: the paper defines late-five accuracy after the complete 1000-task stream, while the original lane tested each task immediately after training. Commit 148a2164 now evaluates only the final five task distributions after all training; streams schedule hashing instead of retaining 60M query decisions; binds the official optimizer step/seen counters into state identity; bounds exact scalar inputs; hardens result/policy/comparison validation against hostile keys and string hooks; and explicitly records that this one-arm runner has not produced matched controls. Paper/code pins and BiMU equations, Concrete correction, full two-layer binary network, reverse-gate JVP, global zero-gradient gate, and mechanism-off semantics were checked against arXiv:2605.30198v1 and official commit 1b8a1a1f. Verification so far: 22 focused tests pass, full Ruff passes, strict mypy passes for 187 source files, diff check clean. The broader fast suite is running. This remains permanently nonpromoting and paper-incomparable; #1570 must stay open pending an authorized matched run and controls.

PR #1673, opened 2026-08-18
fix(scorecard): cross-bind run spec identities

lalalune:
Deep follow-up to #1659. Preserve its exact scalar/roster host gates, then bind the fields as one canonical identity: `(environment, arm, seed)` now derives the only legal schedule index and lifecycle hash under the frozen plan digest. Rejected hostile objects are no longer interpolated through `repr`. `iter_run_specs` shares the derivation helper, preventing drift.\n\nVerification: 72 focused scorecard/workflow/hostile tests; scoped Ruff; strict mypy (187 files); diff check. No workflow dispatch or output mutation.\n\nCo-authored-by: ss251 <ss251@users.noreply.github.com>

### DIFF
--- a/alberta_framework/benchmarks/reference_life_scorecard.py
+++ b/alberta_framework/benchmarks/reference_life_scorecard.py
@@ -235,6 +235,22 @@ def _is_scorecard_lifecycle_id(value: object) -> bool:
     return len(digest) == 16 and all(character in "0123456789abcdef" for character in digest)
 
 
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by merged #1671 (merge c1ee3a61, head 1d9a5359). That consolidation preserves #1659 and implements the same required cross-field schedule/lifecycle binding plus hostile-repr-safe errors, with focused coverage. No unique production behavior remains in this PR.

PR #1676, opened 2026-08-18
feat(benchmarks): add audited Intentional Updates IPMNIST screen

lalalune:
Preserves the contributor implementation from #1643 and adds the deep state/record/resource audit required before merge.\n\nThe successor enforces exact trusted parameter/state/input/hyperparameter identities, shape/dtype and counter/label gates, signed-int32 horizon and 256 MiB persistent-state preflight, finite atomic JIT fallback, exact nested record schemas/scalars/containers, reconstructed dataclass validation, and canonical integer accuracy numerators. It remains a supervised non-publication-equivalent prerequisite; #1561 stays open until a retained matched run/report and distinct TD/control slice exist.\n\nValidation: 14 focused Intentional tests; 13 L2-ER adjacency tests; 335 screening regressions; scoped Ruff; strict mypy.\n\nRelates to #1561. Supersedes #1643.

Issue #1561, opened 2026-08-18
research(streaming): implement Intentional Updates controls


lalalune:
### Objective

Implement batch-size-one supervised and TD/control Intentional Updates under matched update budgets.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2604.19033

hermesagent270-commits:
**Appears already implemented.** Intentional Updates controls are in `alberta_framework/benchmarks/plasticity_comparators.py` (with an `intentional`-updates arm wired into `ipmnist_screening.py`), covered by `tests/test_intentional_updates_ipmnist.py`.

Verified by running the listed test file(s) locally (part of a 270-test batch across all these mechanism suites, all passing). This looks like a **candidate to close** — maintainers, please confirm it matches the intended scope. Flagging as triage; happy to be corrected if I have the wrong module.

hermesagent270-commits:
Pre-registration for the missing TD/control slice of #1561. No benchmark has run and no result or issue-completion claim is being made. My earlier "appears already implemented" note covered the supervised extension and was too broad for the TD/control objective.

The local implementation and development consumer are pinned at source e540cc8ee562bb0987bb31531b4eebe914ca2f5c; the branch has not been published yet. Independent prediction-change, recurrence/terminal-reset, and mechanism-off checks pass; the existing supervised regression panel also passes (19 tests total). Full Ruff and strict mypy pass. These are implementation checks, not research results.

This plan fixes 40 uninterrupted runs (4 development seeds × 5 arms × 2 environments), 10,000 transitions each, with all failures and negative results retained. The optimizer follows the pinned official code's bias-corrected sigma mean, not paper Eq. 12's unnormalized sum. The consumer is linear on-policy SARSA, not a reproduction of the paper's deep Q-learning or policy-gradient results. The two fixed controls are remeasured; no historical baseline exists for this exact protocol. No tuning, promotion, reference-dev change, or frozen-scorecard execution is authorized by this plan.

[...]

PR #1682, opened 2026-08-18
fix(benchmarks): bind Intentional Updates record cross-fields

lalalune:
Narrow reconciliation of the safe unique hardening from #1676 atop merged #1675. Preserves #1675 exact identities and caller-visible JIT invalidity while adding exact frozen hyperparameter constants, synchronized/bounded mechanism clocks, positive normalizer variance, frozen-dataclass reconstruction, exact-step/base-learner provenance, signed-int32 query and 256 MiB persistent-state bounds, and canonical integer accuracy counts.\n\nNo run, output, evidence artifact, or promotion claim. #1561 remains open. #1676 is superseded because its stale replacement would drop visible invalidity and reintroduce weaker container behavior.\n\nVerified: 25 Intentional/L2-ER adjacency tests; Ruff; strict mypy; diff check.


lalalune:
Closed without merge: current main already contains the same safe cross-field hardening in #1679, plus stricter exact nested record schemas and runner preflights. No unique change remains.

PR #1686, opened 2026-08-18
feat(benchmarks): retain bounded matched BiMU development result

lalalune:
Consolidates and preserves the contributor implementation from #1647, then closes the review gaps before issue #1570 evidence is used.\n\n- reconstructs exact frozen configs at every public consumer and caps samples, horizons, queries, dataset/parameter/schedule/sample working bytes, JSON nodes/text/bytes before work\n- requires scalar typed Threefry keys, exact arrays, positive finite Concrete temperature, finite gradients/predictions/state transactions\n- replaces retained per-example schedule structures with a streaming canonical schedule digest\n- makes nonpromotion immutable and exact, binds current implementation/dataset/schedule/runtime, and derives both online and per-task accuracies from exact integer numerators\n- adds canonical result and matched-report digests plus Linux dirfd/O_NOFOLLOW exclusive publication\n- retains candidate and exact memory-off result records plus an arm-derived matched report in a new append-only namespace\n\nThe matched tiny development result keeps the metrics separate: late-five test accuracy is +0.3 (`improved`) while whole-stream pre-update online accuracy is -0.1 (`worse`). This negative/mixed outcome is retained and is permanently nonpromoting. It is not paper-comparable and does not satisfy the official 1000-task/data/five-run paper protocol.\n\nLocal gates: 24 focused tests; Ruff; strict mypy (187 files); strict reload of both arm records and the matched report.\n\nRefs #1570. Do not close until maintainer audit determines whether the scoped tiny executable plus frozen full configuration meets the implementation issue or whether the official long run remains required.


lalalune:
Closing after independent semantic audit. This branch is stale relative to repaired #1647 and preserves material defects: colliding RNG domains, unpinned PRNG implementation, majority-vote inference instead of the official mean-log-softmax rule, and per-task-immediate evaluation mislabeled as the paper final-model late-five metric. Its runtime cap rejects the advertised 1000-task config; committed tiny outputs bind only self-declared hashes and measure the flawed runner, so they must not be retained as a development result. Passing focused tests encode those incorrect semantics. Preserve the corrected contributor implementation from #1647; any future matched output must be regenerated from that runner with real source/dataset/schedule binding.

PR #1692, opened 2026-08-18
fix(benchmarks): consolidate scorecard hostile integer gates

lalalune:
Consolidates and preserves @byt61 commit from #1677, then closes two adjacent pre-validation hooks found in deep review: live `regime_id` equality dispatch and aggregate `schedule_index` coercion before canonical validation. Adds production `run_scorecard_shard` and aggregate-construction zero-hook tests. Permanently nonpromoting; changes current scorecard source identity by design; no artifact or workflow mutation.\n\nVerification: 76 focused tests passed; Ruff clean; strict mypy clean (187 files); diff check clean.\n\nSupersedes #1677.

### DIFF
--- a/alberta_framework/benchmarks/reference_life_scorecard.py
+++ b/alberta_framework/benchmarks/reference_life_scorecard.py
@@ -686,9 +686,9 @@ class StreamingRunSummary:
     ) -> None:
         if self._accepted_events >= self._horizon:
             raise ValueError("streaming summary already reached its horizon")
-        if isinstance(reward, bool) or not isinstance(reward, (int, float)):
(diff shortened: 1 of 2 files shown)


lalalune:
Adjacent gates are useful, but the new aggregate preflight is still not fail-closed for the hostile-record boundary it advertises: `isinstance(record, Mapping)` admits arbitrary Mapping implementations, then `record.get(...)` can dispatch user hooks before exact container admission. Require exact builtin `dict` (and exact string keys if not already guaranteed by the caller contract) before `.get`/sorting, and add a zero-hook hostile Mapping regression. The regime-id production regression looks sound.

PR #1693, opened 2026-08-18
fix(validation): reject hostile integer identities in replay and visualization

lalalune:
Contributor-preserving consolidation of #1678 and #1680 after auditing the adjacent scalar/config/checkpoint/resource gates. The residual changes are narrow: exact builtin numeric identity is required before range or float dispatch in reservoir selection, checkpoint numeric leaves/counters, and visualization positive-real settings. Current main already contains the stronger feature-router successor for #1684/#1691, so no duplicate router patch is included.\n\nValidation: 173 dual-replay tests; 2 visualization hostile tests; 44 adjacent feature-router tests; scoped Ruff; strict mypy.\n\nSupersedes #1678 and #1680.

### DIFF
--- a/alberta_framework/core/dual_replay.py
+++ b/alberta_framework/core/dual_replay.py
@@ -651,7 +651,7 @@ def reservoir_selection(
     ``[0, n)`` and the candidate replaces slot ``draw`` iff ``draw < capacity``.
     The caller owns key advancement; this pure calculation does not return a key.
     """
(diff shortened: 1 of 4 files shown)


lalalune:
This successor did not incorporate the prior review fixes: it still uses broad `pytest.raises(Exception)` throughout, so a hostile-hook AssertionError can falsely pass, and it still includes two simulated `test_hostile_not_in_error_message` tests that never call production. Replace with exact ValueError and delete the simulations before merge. The production gates themselves are reasonable.

PR #1667, opened 2026-08-18
fix(evaluation): reject leftover IAConditionResult identities

ss251:
Closes #1666.

## What changed

`IAConditionResult` now fail-closes leftover bool seed/count/rate identities, leftover condition names, exact ndarray hosts, and nested budget/timing types — the same complete bar Kayvan `#1316` already applied to acceptance records.

On `origin/main` `cbf63ecf`, `IAConditionResult(seed=True)` constructed and stored `bool True`. Legal integer seeds and finite rates stay legal.

No unpublished grok head on this file. Occupancy-free. Not a leftover-tax reprint and not `optimizers.py`.

## Scope

- `alberta_framework/evaluation/continual_ia.py`
- `tests/test_ia_condition_result_hostile.py`

## Occupancy

Open ASI PRs at publish: ss251 `#1659` (`reference_life_scorecard.py`); Shaw `#1657`/`#1647`/`#1643`. No open PR on `continual_ia.py`.

## Verification

- Red on base: `IAConditionResult(seed=True)` constructed; `mean_reward=True` constructed
- New result contract + acceptance hostile tests: 58 passed
- Related integer-family tests: 84 passed
- `ruff check` on the two changed files: clean
- `scope-check.sh --max-files 2`: PASS

## Scientific integrity

Fail-closed host-boundary leftover-identity validation only. No kernel math, lifetime clocks, `CHANGELOG.md`, or `outputs/**` change.

- [x] logs: N/A - host-boundary unit tests only; no benchmark shard was run

- [x] domain-artifact: N/A - no `outputs/**` artifact is produced by this harness fix

- [x] llm-trajectory: N/A - no agent trajectory artifact is attached; reproduction is the unit tests above

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Cursor

- Skill path: elizaOS/slopdotcash@72e4239ebed8c99d34181d5d1f21fb316cacecd6:skills/contribute-to-asi

- Provenance status: self-reported

AI provider/model: xAI / grok-4.6
Client / agent tooling: Cursor
Contribution skill revision: elizaOS/slopdotcash@72e4239ebed8c99d34181d5d1f21fb316cacecd6:skills/contribute-to-asi
Attribution status: self-reported
— [cursor-ss251]

PR #1704, opened 2026-08-18
fix(evaluation): revalidate IA results at aggregation

lalalune:
Follow-up to the constructor contract consolidated in #1681 (superseding #1667).

- owns read-only copies of every IA result array
- reconstructs nested budget/timing values rather than retaining mutable aliases
- reconstructs the exact config at aggregation so forged frozen configs cannot bypass invariants
- revalidates exact records at the public aggregation boundary
- derives phase means and recovery lengths from primitive rewards plus the supplied config
- covers external mutation, nested subclasses, forged configs, and post-construction mutation through production calls

Validation on current main:
- 34 focused IA tests passed
- 11 hostile result/config tests passed after final rebase
- Ruff passed
- strict mypy passed (188 source files)
- diff check and main-ancestry check passed

No benchmark or evidence artifact was changed or produced. This is validation hardening only.

### DIFF
--- a/alberta_framework/evaluation/continual_ia.py
+++ b/alberta_framework/evaluation/continual_ia.py
@@ -25,7 +25,7 @@ from __future__ import annotations
 import functools
 import operator
 from collections.abc import Callable, Sequence
-from dataclasses import dataclass
+from dataclasses import dataclass, fields
 from time import perf_counter
 from typing import Any, Literal, SupportsIndex, cast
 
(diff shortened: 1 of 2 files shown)


lalalune:
The ownership change currently copies every array inside `_require_ndarray` before the aggregate `_MAX_CONDITION_RESULT_BYTES` check. A record with seven individually valid ~500 MiB arrays can therefore allocate several GiB of snapshots before being rejected by the advertised 512 MiB record bound. Please split admission from snapshotting: validate exact type/dtype/rank/length for all arrays, sum their original `nbytes` and reject above the cap, then make the read-only copies. Add a regression that monkeypatches/observes `np.array` (or uses over-cap shapes without allocating payload pages) to prove an over-budget record is rejected before the first copy.

PR #1713, opened 2026-08-18
fix(evaluation): revalidate nested multiagent records

lalalune:
Minimal current-main follow-up after merged #1706 and #1703.

The public aggregator reconstructed `ConditionResult` but reused its nested `ControllerBudget` and `TimingMetrics` objects. A post-construction mutation could therefore bypass their scalar validators and reach budget/timing aggregation. This reconstructs only those two nested records before reconstructing the result and adds a production-call regression. It does not duplicate the summary/recovery/config work already on main.

Validation on current main:
- 41 focused multiagent tests passed
- Ruff passed
- strict mypy passed (188 source files)
- diff check and main-ancestry check passed

No benchmark or evidence artifact was changed or produced.

### DIFF
--- a/alberta_framework/evaluation/continual_multiagent.py
+++ b/alberta_framework/evaluation/continual_multiagent.py
@@ -945,8 +945,28 @@ def aggregate_evidence(
     for result in results:
         if type(result) is not ConditionResult:
             raise ValueError("results must contain exact ConditionResult records")
+        budget = ControllerBudget(
+            **{
+                field.name: getattr(result.controller_budget, field.name)
+                for field in fields(ControllerBudget)
+            }
+        )
+        timing = TimingMetrics(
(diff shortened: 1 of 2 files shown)


lalalune:
One fail-closed gap remains before merge: `aggregate_evidence` reads every field from `result.controller_budget` and `result.timing` before establishing that those nested objects still have the exact `ControllerBudget` / `TimingMetrics` identity. A forged frozen `ConditionResult` can replace either nested record with an object whose property/`__getattribute__` hook executes during this reconstruction. Please check `type(result.controller_budget) is ControllerBudget` and `type(result.timing) is TimingMetrics` before any field access, and add hostile replacements whose hooks must remain at zero. The subsequent constructor reconstruction can then validate forged scalar contents safely.

Issue #1559, opened 2026-08-18
research(ipmnist): reproduce L2-ER effective-rank regularization


lalalune:
### Objective

Implement faithful L2-only, rank-only, and combined L2-ER arms in the current IPMNIST runner.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2509.22335

lalalune:
Completed by the audited L2-ER matched-development implementation and authoritative result retained in #1753. The frozen 3-seed comparison is permanently nonpromoting: L2-only, ER-only, and combined outcomes are all inconclusive because every exact paired t(df=2) confidence interval includes zero. Both invalid prior attempts remain disclosed and unmerged; authoritative seeds 1711–1713 and complete update/query/resource accounting are retained.

PR #1716, opened 2026-08-18
results(ipmnist): retain L2-ER matched development rejection

lalalune:
Retains the first and only execution of frozen plan `asi.l2er-ipmnist.cheap-screen.v1` from exact source f62d157a449c30a79dcf01740ae7f20a6c27e726. This is permanently nonpromoting development output in the new append-only namespace from #1700.\n\nValidated before retention against the executing source/runtime/dataset. Across 3 matched seeds × 2 tasks × 500 updates: L2-only was exactly tied with mechanism-off (rejected under the frozen upper-CI <= 0 rule); ER-only mean delta was -0.03933 (95% CI [-0.05901,-0.01966]); combined mean delta was -0.04000 (95% CI [-0.05989,-0.02011]). All negative outcomes are retained. Timing is telemetry only.\n\nCloses #1559.

### DIFF
--- a/outputs/l2er_matched_development/report.v1.json
+++ b/outputs/l2er_matched_development/report.v1.json
@@ -1,0 +1,718 @@
+{
+ "dataset_provenance": {
+  "materialization": "alberta.ipmnist.float32-neg1-pos1-int32-labels.v1",
+  "schema": "alberta.ipmnist_screening.dataset_provenance.v1",
+  "source": {
+   "name": "mnist_784",
+   "provider": "openml",
+   "row_start": 0,
+   "row_stop_exclusive": 60000,
+   "version": 1
+  },
+  "x": {
+   "dtype": "<f4",
+   "sha256": "b8078cd833f53d89828a5e28d728517be9add34076f13fe973399f1f16381313",
+   "shape": [
+    60000,
+    784
+   ]
+  },
+  "y": {
+   "dtype": "<i4",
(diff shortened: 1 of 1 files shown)


lalalune:
Blocking correctness review: do not merge this artifact as advertised. The executing source uses z=1.96 with n=3, so these are not valid 95% intervals (two-sided t df=2 requires 4.30265, unless the claim is relabeled). ER receipts also omit ten effective-rank parameter updates per 1,000 observations from the update accounting. “First and only execution” is false because seed 1701 was already consumed in a four-arm audit run without an artifact. The raw negative observations may be retained in a new, honest schema after the source repair, but this derived report must not close #1559.

lalalune:
Closing to prevent an invalid derived report from being merged. Re-run or re-derive into a new append-only path after the statistics/accounting repair; retain all negative outcomes and disclose the earlier seed-1701 audit execution.

lalalune:
Closing as an invalid historical execution attempt. Its unmerged v1 artifact (SHA-256 ab7f03b73993b79c53021970926181130980dd84b180e095779e30960953ff86; seeds 1701-1703) used z=1.96 instead of Student t(df=2), omitted auxiliary effective-rank parameter updates from total update accounting, and reused pre-consumed audit seed 1701. The artifact remains preserved in this PR history and was not merged. Corrected fresh-seed v2 source was independently approved and merged in #1729; the distinct result-only v2 artifact is under independent review in #1742.

PR #1739, opened 2026-08-18
fix(evaluation): preserve IA action snapshot roles

lalalune:
Follow-up to #1724. That preflight fix correctly delays ownership copies until after the byte gate, but its zip labels are reversed for the two primitive action arrays: the source tuple is `(executed_actions, credited_actions)` while the destination names are `("credited_actions", "executed_actions")`. Every valid result with an action-credit mismatch therefore stores the two arrays under the wrong public fields.\n\nThis one-line repair restores the field order and adds a regression with deliberately distinct executed/credited arrays.\n\nValidation on current main:\n- 169 focused IA tests passed, 9 skipped\n- Ruff passed\n- strict mypy passed\n- diff check passed\n\nNo evidence artifact or benchmark output changed.

### DIFF
--- a/alberta_framework/evaluation/continual_ia.py
+++ b/alberta_framework/evaluation/continual_ia.py
@@ -509,8 +509,8 @@ class IAConditionResult:
         for name, array in zip(
             (
                 "rewards",
-                "credited_actions",
                 "executed_actions",
+                "credited_actions",
                 "recommendations",
                 "partner_proposals",
                 "accepted_recommendations",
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #1744. The complete successor preserves this PR's contributor-authored action-role fix as commit 522ce001, then adds the missing IA recommendation/live-budget consumption bindings and multiagent exact-config reconstruction. Verified with 171 IA tests plus 251 multiagent tests, full Ruff, strict mypy, and diff check.

PR #1746, opened 2026-08-18
fix(security): revalidate oracle records at consumption

lalalune:
Follow-up to #1738: seal nested record values transitively and revalidate canonical state at both public consumers.

- freeze nested JSON containers, not only the outer mapping
- reject forged dataclass scalar, container, and resource identities before hooks
- serialize from a fresh validated snapshot to close consumption-time alias drift
- validate schema/action/label/resource cross-fields again at the public validator

Verification: 54 security tests; Ruff; strict mypy (189 files); diff check. No evidence artifacts or workflows touched.

### DIFF
--- a/alberta_framework/security.py
+++ b/alberta_framework/security.py
@@ -128,6 +128,62 @@ def _copy_json_mapping(payload: object, *, name: str) -> dict[str, Any]:
     }
 
 
+def _freeze_json_value(value: Any) -> Any:
+    value_type = type(value)
+    if value_type is list:
+        return tuple(_freeze_json_value(item) for item in value)
+    if value_type is dict:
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by merged #1743, which contains the same consume-time reconstruction fix on current main with focused validation. This branch is conflicting and should not be merged as a duplicate.

lalalune:
Reopening after exact reconciliation with merged #1743: this branch retains #1743 reconstruction tests and uniquely deep-freezes nested retained JSON, rejects forged mutable MappingProxy contents, bounds forged immutable nesting before recursion, and serializes from the validated snapshot. Verified 59 security tests, Ruff, strict mypy 189 files, diff check.

PR #1747, opened 2026-08-18
fix(ipmnist): restore authoritative L2-ER seed plan

lalalune:
## Summary
- restore the prospective v2 seed matrix frozen by #1728 (1711–1713) and its exact confidence/audit fields after #1729 changed them without a consumption rationale
- retain #1729 unique exact-dict, resource-update accounting, and pinned append-only publication hardening
- record both invalid, unretained executions as audit metadata, including consumed churned seeds 1721–1723 and artifact SHA-256 `579c400412d3c50898c16a8fd02fa82e2cd712b5278b5a99974b5e89560707ec`

No benchmark was run and no output artifact is added or rewritten.

## Verification
- `python -m pytest tests/test_l2er_matched_development.py tests/test_l2er_ipmnist.py -q -o addopts=` (28 passed)
- `python -m ruff check .`
- `python -m mypy`
- `git diff --check`

### DIFF
--- a/alberta_framework/benchmarks/l2er_matched_development.py
+++ b/alberta_framework/benchmarks/l2er_matched_development.py
@@ -41,32 +41,50 @@ ARMS = (
     "l2er_er_only",
     "l2er_combined",
 )
-SEEDS = (1721, 1722, 1723)
+SEEDS = (1711, 1712, 1713)
 CONFIG = IPMNISTConfig(n_tasks=2, task_length=500)
-_T95_DF2 = math.sqrt(1.805 / 0.0975)
-_INVALID_V1_ATTEMPT = {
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking correction before merge: restoring the authoritative prospective seeds 1711-1713 and recording both invalid attempts is correct, but the PR also restores the approximate df=2 critical 4.302652729696142. For df=2, the two-sided 95 percent critical is analytic: math.sqrt(1.805 / 0.0975) equals 4.302652729749464, hex 0x1.135ea98e146bbp+2. Keep the seed/audit repair while retaining the computed exact critical and pinned hex test. This is a pre-execution statistical correction, not a rationale for seed churn. Do not execute until corrected source merges.

PR #1742, opened 2026-08-18
results(ipmnist): retain matched L2-ER v2 outcome

lalalune:
Retains the single corrected v2 matched nonpromoting development execution for #1559.\n\nExecution identity:\n- exact merged source: 5fc85b69897fc5b74c27389579c3d1bc27931394\n- fresh seeds: 1721, 1722, 1723\n- new immutable path/schema: outputs/l2er_matched_development/report.v2.json / asi.l2er-ipmnist.matched-development-report.v2\n- artifact SHA-256: 579c400412d3c50898c16a8fd02fa82e2cd712b5278b5a99974b5e89560707ec\n- Student t(df=2) critical derived as sqrt(1.805/0.0975), binary64 0x1.135ea98e146bbp+2\n- corrected ER accounting: 1000 supervised + 10 auxiliary = 1010 total updates, 2010 model queries\n\nAll three candidate comparisons are retained as inconclusive under the corrected interval. This is development-only and permanently nonpromoting.\n\nHistorical disclosure: invalid unmerged v1 result PR #1716 used seeds 1701-1703 and artifact SHA ab7f03b73993b79c53021970926181130980dd84b180e095779e30960953ff86. It used z=1.96, omitted ER auxiliary updates from total updates, and reused pre-consumed seed 1701. That v1 JSON is not included here and must not be merged.\n\nThis PR contains exactly one new v2 JSON artifact. Keep #1559 open pending independent audit.

PR #1750, opened 2026-08-18
fix(ipmnist): freeze honest L2-ER matched v3 plan

### LINKED ISSUE #1559: research(ipmnist): reproduce L2-ER effective-rank regularization

lalalune:
## Summary

- version the corrected matched-development plan/report/output as v3 instead of mutating the executed v2 contract
- retain the authoritative fresh 1711-1713 schedule and restore #1717 accounting: `updates` is matched supervised updates while `effective_rank_updates` is the separate arm-specific charged axis
- bind both verified closed/unmerged executions: v1 PR #1716 (seeds 1701-1703) and v2 PR #1742 (seed-churn 1721-1723), including artifact digests, executing sources, result heads, and invalidity reasons
- preserve the pinned-dirfd no-overwrite publisher, strict canonical reload, and hostile-key preflight from the merged hardening
- use the exact closed-form two-sided Student-t critical for three paired seeds

No output artifact is added. This remains permanently nonpromoting and does not close #1559.

## Why v3

The v1 execution used the normal critical, did not separately report auxiliary effective-rank steps, and reused audit-consumed seed 1701. The subsequent v2 execution used unexplained 1721-1723 seed churn and recombined auxiliary ER steps into the matched `updates` field. Because v2 has now executed, correcting these contract fields under the same identity would violate append-only protocol history. v3 keeps the authoritative unexecuted 1711-1713 schedule while recording both invalid histories.

## Validation

- 33 focused tests passed after rebase to current main
- Ruff passed on touched files
- strict mypy passed (189 source files)
- diff check passed

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -8703,7 +8703,7 @@ def l2er_development_result_payload(
         "hidden2": config.hidden2,
         "n_classes": config.n_classes,
         "observations": observations,
-        "updates": observations + er_updates,
+        "updates": observations,
         "effective_rank_updates": er_updates,
         "allowed_boundary_information": [],
         "allowed_task_information": ["current_example_label"],
--- a/alberta_framework/benchmarks/l2er_matched_development.py
+++ b/alberta_framework/benchmarks/l2er_matched_development.py
@@ -33,32 +33,34 @@ from alberta_framework.evaluation.l2er_ipmnist_nonpromoting import (
     validate_l2er_development_result,
 )
 
-SCHEMA = "asi.l2er-ipmnist.matched-development-report.v2"
-PLAN_ID = "asi.l2er-ipmnist.cheap-screen.v2"
+SCHEMA = "asi.l2er-ipmnist.matched-development-report.v3"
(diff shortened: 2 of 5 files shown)


lalalune:
Closing: this is not a distinct justified experiment. It renames the restored, untouched 1711–1713 matrix from authoritative v2 to v3, changes the frozen Student-t constant, and removes #1729/#1747 honest accounting by making `updates` exclude executed auxiliary ER optimizer steps. That would evade/replace #1747 rather than preserve it, while dropping the exact consumed-invalid 1721–1723 disclosure now recorded on main. No output or execution is authorized. The prospective contract remains v2 at #1747; any future 1711–1713 run is separate and explicitly authorized, and #1559 stays open absent a valid retained outcome.

PR #1757, opened 2026-08-18
fix(evaluation): preserve gradual resource correction lineage

lalalune:
## Summary
- preserve and publicly validate the exact immutable v1 gradual result (file SHA-256 `bd89182f...`)
- reject the already-merged defective v2 by exact internal/file identities (`9ff58ac5...` / `a4128b81...`) without changing its bytes
- append a v3 derivation that separates historical execution source/runtime from current derivation source/runtime and changes only persistent-resource accounting (+1,128,640 bytes per arm)
- preserve all raw metrics, timings, comparisons, and the original descriptive `outcome`; permanently nonpromoting policy and #1569 remain unchanged

No benchmark was rerun and no existing output was rewritten.

## Verification
- 27 focused tests
- full Ruff
- strict mypy (189 source files)
- exact cached-dataset reconstruction: v1 accepted, v2 rejected, v3 accepted
- `git diff --check`


lalalune:
Independent integration review passed at rebased head 635638f1: the v1 and invalid-v2 file SHA-256 values match the frozen lineage, the v3 derivation preserves all historical metrics/timings/outcomes while changing only the omitted resident parameter bank, explicit Threefry schedule reconstruction remains independent of the process default, and the 27-test focused panel plus Ruff and mypy pass. This correctly supersedes the defective v2 derivation without rewriting it.

PR #1753, opened 2026-08-18
results(ipmnist): retain authoritative L2-ER v2 outcome

lalalune:
Retains the single authoritative v2 matched nonpromoting development execution for #1559.\n\nExecution identity:\n- exact merged source: 072cfee9d061a2e2a370eee21ea901aa9fbad870\n- authoritative unconsumed seeds: 1711, 1712, 1713\n- immutable path/schema: outputs/l2er_matched_development/report.v2.json / asi.l2er-ipmnist.matched-development-report.v2\n- artifact SHA-256: c5a6b8efb050d3c6c05648a46689ca0903389340764f7c20b589bfc4e8b0c6f2\n- exact analytic Student t(df=2) critical: 4.302652729749464 / 0x1.135ea98e146bbp+2\n- ER accounting: 1000 supervised + 10 auxiliary = 1010 total updates and 2010 model queries\n\nRetained paired outcomes (all inconclusive):\n- L2-only mean +0.000333, 95% CI [-0.003461, +0.004128]\n- ER-only mean -0.050333, 95% CI [-0.112701, +0.012035]\n- combined mean -0.054333, 95% CI [-0.122152, +0.013486]\n\nThe report binds both invalid unmerged histories: v1 SHA ab7f03... (bad z/update accounting/preconsumed seed) and churned-seed v2 SHA 579c40... (unauthorized 1721-23 substitution). Neither invalid JSON is included. This PR contains exactly one new JSON artifact and remains development-only/permanently nonpromoting.\n\nKeep #1559 open pending independent artifact audit.

PR #1758, opened 2026-08-18
fix(ipmnist): remove superseded invalid L2-ER v2 artifact

lalalune:
## Summary

- remove the v2 artifact that raced into main after the explicit-accounting v3 contract merged
- restore the live-tree invariant already recorded by v3: this v2 execution is invalid historical data and is not retained
- preserve its SHA-256, source commit, seeds, and rejection reason in the v3 frozen plan and in PR #1753/Git history

The v2 file was merged after its own supersession comment because auto-merge completed concurrently with #1756. It cannot validate against the current v3 schema and must not occupy the old fixed output path.

## Verification

- 29 focused L2-ER tests passed
- Ruff passed on the L2-ER implementation/tests
- `git diff --check` passed

No valid or registered evidence artifact is changed. The removed file remains fully recoverable from merge commit `ee6d8949` and PR #1753.


lalalune:
Closing: deleting an already merged output violates the append-only evidence contract. The defective gradual v2 bytes must remain immutable and are explicitly invalidated by exact internal/file identity in #1757; the corrected resource-only derivation is appended under development.v3 instead. No existing artifact may be removed or rewritten.

lalalune:
Correction to the preceding wording: #1758 targets the L2-ER `report.v2.json` merged by #1753, not the gradual artifact. The same append-only rule applies: retain those bytes and record any invalidity/correction append-only; #1757 is a separate gradual-IPMNIST corrective and does not supersede this L2-ER file.

PR #1759, opened 2026-08-18
fix(ipmnist): quarantine consumed L2-ER seeds

lalalune:
Records the exact #1753 v2 execution that consumed and exposed all planned 1711–1713 outcomes, then blocks prospective v3 execution/publication.

- binds #1753 source/head/artifact SHA, seeds, and exact exposed paired outcomes in invalid history
- marks the planned matrix seeds consumed and v3 execution unauthorized
- requires a separately preregistered methodological disposition; authorizes neither a same-seed prospective claim nor silent seed churn
- fails before output reservation or dataset loading
- preserves the analytic t critical, v3 counter formulas, nonpromotion, and prior invalid histories

No benchmark run or output was created. Verification: 30 focused tests, Ruff, focused strict mypy, diff check.


lalalune:
The quarantine mechanism and execution block are correct, but two frozen strings still contradict the actual history: `disposition` says `invalid_unmerged_consumed_accounting_ambiguous_attempt`, and the final reason says `the unmerged v2 artifact...`, while the same record correctly binds merge commit `ee6d8949` and the artifact now exists on main. Please change these to an explicit merged/quarantined disposition (for example `invalid_merged_quarantined_consumed_accounting_ambiguous_attempt`) and `the merged quarantined v2 artifact cannot authorize...`, with regression assertions. Once those exact identities tell the truth, this is the right fail-closed resolution.

PR #1761, opened 2026-08-18
fix(ipmnist): name merged L2-ER quarantine honestly

lalalune:
Follow-up to #1759. The quarantine record already binds #1753 merge commit `ee6d8949`, but two strings still called that artifact unmerged. This changes only those identities to `invalid_merged_quarantined_consumed_accounting_ambiguous_attempt` and “merged quarantined v2 artifact,” with regression assertions.

Verification: 30 focused L2-ER tests passed; Ruff passed; diff check passed. No artifact bytes or execution authorization change.


lalalune:
Superseded by merged #1762, which preserves the consumed-seed execution block while correctly recording #1753 as an accepted, retained, permanently nonpromoting inconclusive development result. This narrower branch still mislabeled that accepted artifact as invalid/ambiguous, so it must not merge.

PR #1764, opened 2026-08-18
fix(ipmnist): retain exact historical L2-ER loader

lalalune:
Corrects the residual historical-compatibility and publication-boundary gaps after #1762.

- dispatches the exact retained v2 schema through a canonical-path/file-SHA loader
- accepts only the immutable #1753 artifact bytes and preflights the parsed JSON tree
- leaves the generic v3 validator and its current-source/runtime checks unchanged
- blocks direct `_publish_report` calls before validation or any write while v3 remains unauthorized
- preserves the valid consumed v2 history and the non-runnable prospective v3 plan

No output artifact was edited and no execution was launched.

Verification: 31 focused L2-ER tests passed; Ruff full clean; strict mypy clean (189 source files); diff check clean; retained v2 SHA-256 remains `c5a6b8efb050d3c6c05648a46689ca0903389340764f7c20b589bfc4e8b0c6f2`.

Refs #1559.

### DIFF
--- a/alberta_framework/benchmarks/l2er_matched_development.py
+++ b/alberta_framework/benchmarks/l2er_matched_development.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import argparse
+import hashlib
 import json
 import math
 import os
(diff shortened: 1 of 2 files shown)


lalalune:
Independent review passed at 995a3259: retained artifact bytes reproduce SHA-256 c5a6b8ef..., exact-path/hash dispatch admits only the immutable v2 payload, tampering is rejected, generic v3 current-source validation is untouched, and direct publication now fails before validation or writes. Focused L2-ER panel 31/31, Ruff, mypy, and diff check all pass. Safe to merge after required checks.

PR #1767, opened 2026-08-18
fix(evaluation): bind exact condition-timing identities

PrototypicNoise:
Fix-lane follow-up to the hostile-identity wave (#1693 / #1722 / #1725): the operational condition-timing validator in `continual_ia_artifact.py` was the remaining identity gate still using bare `isinstance`.

## The spoof

`_validate_operational` accepted `"seed": true` from a JSON artifact — `isinstance(True, int)` passes, and `(True, cond) == (1, cond)` in the observed/expected pair comparison, so a bool seed silently binds to seed 1. Similarly, `str`-subclass conditions pass `isinstance` and execute their `__eq__/__hash__` hooks inside `observed_pairs != expected_pairs`. The same file already guards every other boundary (`_integer`, `_number`, 6 bool guards) — this line was the miss.

## Fix

One line, matching the wave's exact-type convention:
```python
- if not isinstance(seed, int) or not isinstance(condition, str):
+ if type(seed) is not int or type(condition) is not str:
```

**Production-path safety:** genuine artifacts are unaffected — seeds are canonicalized to exact `int` via `operator.index` in `_require_integer` (continual_ia.py) before reaching `IAConditionResult`, and JSON parsing yields exact types on the read path.

## Tests (failing-test-first)

New `tests/test_continual_ia_timing_identity_hostile.py` — 4 tests: exact-identity pass, bool-seed rejection, hostile-int-subclass rejection with zero hook executions, hostile-str-subclass rejection with zero hook executions. 3 fail on main before the fix; 4 pass after.

## Verification

Commit measured: `82c2a49120e31a3719905a2ed04abe8584141069`
```
pytest tests/test_continual_ia_timing_identity_hostile.py -q -o addopts=""  # 4 passed
pytest tests/test_continual_ia_cli.py -q -o addopts=""                      # 5 passed
ruff check (both touched files)   # All checks passed
mypy (both touched files)         # Success: no issues in 2 source files
```
**Honest limitation:** `tests/test_continual_ia_evidence.py` (141 tests) exceeds my runner's memory (2 vCPU / 4 GB container; the artifact fixtures OOM) — I could not complete it locally. Static analysis of the seed path (above) says it passes; CI should confirm. If anything there fails, treat this PR as not verified and I'll rework.

## Disclosure

Client `claude-code`, provider `anthropic`, model `claude-fable-5`. No Slop run receipt: project policy is `paused`/`pending-authority-activation` at submission — unscored contribution.

---
🤖 Built with [Emblem:build](https://emblem.build)

### DIFF
--- a/alberta_framework/evaluation/continual_ia_artifact.py
+++ b/alberta_framework/evaluation/continual_ia_artifact.py
@@ -1071,7 +1071,7 @@ def _validate_operational(
             condition = timing.get("condition")
             wall = _number(timing.get("wall_seconds"))
             latency = _number(timing.get("mean_step_latency_ms"))
-            if not isinstance(seed, int) or not isinstance(condition, str):
+            if type(seed) is not int or type(condition) is not str:
                 errors.append(f"condition_timings[{index}] has invalid identity")
             else:
                 observed_pairs.append((seed, condition))
--- a/tests/test_continual_ia_timing_identity_hostile.py
+++ b/tests/test_continual_ia_timing_identity_hostile.py
@@ -1,0 +1,109 @@
+"""Hostile identity gates for operational condition-timing validation."""
+
+from __future__ import annotations
+
+import pytest
+
+from alberta_framework.evaluation.continual_ia_artifact import _validate_operational
+
+pytestmark = pytest.mark.unit
+
+
+class _HostileInt(int):
+    calls = 0
+
+    def __eq__(self, other: object) -> bool:
+        type(self).calls += 1
+        raise AssertionError("hostile int equality hook executed")
+
+    def __hash__(self) -> int:
+        type(self).calls += 1
+        raise AssertionError("hostile int hash hook executed")
+
+
+class _HostileStr(str):
+    calls = 0
+
+    def __eq__(self, other: object) -> bool:
+        type(self).calls += 1
+        raise AssertionError("hostile str equality hook executed")
+
+    def __hash__(self) -> int:
+        type(self).calls += 1
+        raise AssertionError("hostile str hash hook executed")
+
+
+def _operational(timings: list[dict[str, object]]) -> dict[str, object]:
+    return {
+        "digest_exclusion_reason": (
+            "host environment and wall-clock timing are non-deterministic"
+        ),
+        "environment": None,
+        "condition_timings": timings,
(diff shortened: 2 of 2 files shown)


Svector-anu (approved):
Verified independently (fresh clone, not relying on the PR's own description):

**Bug is real.** `_validate_operational`'s old check was `isinstance(seed, int) or isinstance(condition, str)`. Since `bool` subclasses `int` in Python, a `seed: true` in a submitted artifact passes the type check, and because `True == 1` and `hash(True) == hash(1)`, it also passes the downstream `observed_pairs != expected_pairs` list-equality check against a real int seed — a forged operational-timing artifact using boolean seeds instead of the actual experiment seeds could pass identity validation undetected. `type(x) is int`/`type(x) is str` closes that exactly, matching the same exact-identity pattern already used elsewhere in this codebase for artifact validation.

**Verified independently:**
- Ran `tests/test_continual_ia_timing_identity_hostile.py` against the PR branch: 4 passed.
- Mutation check: reverted just the source fix back to `isinstance(...)`, kept the tests — 3 of 4 fail exactly as expected (`test_bool_seed_is_rejected`, `test_int_subclass_seed_is_rejected_without_hooks`, `test_str_subclass_condition_is_rejected_without_hooks`), confirming the tests actually exercise the fix rather than passing vacuously.
- `ruff check` and `mypy` both clean on the touched files.
- Reachability: `_validate_operational` → `_validate_ia_evidence_artifact` → public `validate_ia_evidence_artifact(artifact: Mapping[str, object])`, which accepts caller-supplied artifact data directly — not test-only.

Nice touch on the hostile-subclass tests specifically asserting `calls == 0` on the poisoned `__eq__`/`__hash__` — proves the fix rejects via `type()` before ever touching equality/hash, so a malicious subclass can't even get its hooks invoked (avoiding a secondary exception/DoS vector, not just the identity-spoofing one). Solid PR.

PrototypicNoise:
Thanks for the independent fresh-clone verification — exactly the review standard this repo runs on. Confirming your reading: the forgery path needs both `bool ⊂ int` for the type gate and `True == 1` / `hash(True) == hash(1)` for the pair comparison, which is why the exact-type gate closes both at once.

lalalune:

PR #1850, opened 2026-08-18
fix(option-duration): gate array inputs on metadata before conversion

### LINKED ISSUE #1849: OptionValueDuration array runner runs untrusted __array__ hooks before its metadata gate
## Problem

`OptionValueDurationLearner._require_array` (`alberta_framework/core/option_value_duration.py:364`) converts before it validates:

```python
    @staticmethod
    def _require_array(name, value, *, shape, dtype) -> Array:
        try:
            array = jnp.asarray(value)          # <- untrusted conversion hook runs here
        except Exception as error:
            raise ValueError(f"{name} must be a readable array") from error
        if array.shape != shape or array.dtype != dtype:
            raise ValueError(...)
        return array
```

An object supplying `__array__` therefore executes its hook ahead of the shape and dtype checks and chooses the value they then see. Worse, the `except Exception` wraps whatever the hook raises into a generic `ValueError`, so a failure inside untrusted code is reported as a malformed-input error.

This is the conversion-before-type-gate class closed for the associative sinks in #1168 and for `behavior_model._integer_action_ids` in #1306. This module's helper was missed.

## Scope — this is an outlier, not a sweep

I checked every `_require_array`-style helper in the tree. **14 of 18 already gate on array metadata before converting** (`working_memory`, `resource_manager`, `dual_replay`, `temporal_context`, `associative_memory`, `prototype_memory`, `upgd_memory`, `experiential_memory`, `feature_discovery`, the forager helpers, ...). I empirically probed the four that looked ungated; `gauntlet` and `recurring_multiagent` reject without running the hook. **`option_value_duration` is the reachable one.**

(`interaction_features._require_array_contract` also runs the hook when called directly. I have not established a public-API path that reaches it with an untrusted object, so I am not claiming it here.)

## Expected behavior

Gate on array metadata before any conversion, matching the 14 helpers that already do.

## Actual behavior

Confirmed against `cc877f78` (clean tree). The public `update()` path is gated by an outer contract check, but `run_option_value_duration_from_arrays` calls the helper directly:

```text
run_from_arrays(option_indices=Hostile()) -> ACCEPTED   hook_ran=True
run_from_arrays(rewards=Hostile())        -> ValueError hook_ran=True
update(observation=Hostile())             -> TypeError  hook_ran=False   (outer gate gets it first)
```

The first line is the finding: the runner accepts the object outright, and the hook decided what it validated.

## Reproduction

```python
import numpy as np, jax.numpy as jnp
from alberta_framework.core.option_value_duration import (
    OptionValueDurationLearner as L, OptionValueDurationConfig as C,
    run_option_value_duration_from_arrays as run)

class Hostile:
    def __array__(self, dtype=None, copy=None):
        print("hook ran before validation")
        return np.zeros((3,), dtype=np.int32)

lr = L(3, C()); st = lr.init(2); n = 3
run(lr, st, jnp.zeros((n,2), jnp.float32), Hostile(),
    jnp.zeros((n,), jnp.float32), jnp.zeros((n,2), jnp.float32),
    jnp.zeros((n,), jnp.float32))     # hook runs; accepted
```

## Environment

Python 3.12.14, jax 0.11.0, numpy 2.5.2, CPU, network-isolated container built from `pyproject.toml` with `.[dev,gymnasium,forager]`.

PR to follow. Happy to close if this duplicates something in flight.

---

techforky:
Fixes #1849.

## What moved

`OptionValueDurationLearner._require_array` reached `jnp.asarray(value)` before validating anything, so an untrusted `__array__` hook executed ahead of the shape and dtype checks and chose the value they then saw. The `except Exception` also wrapped whatever the hook raised into a generic `"must be a readable array"` ValueError — reporting a failure inside untrusted code as malformed input.

Confirmed on `cc877f78` before the fix:

```text
run_from_arrays(option_indices=Hostile()) -> ACCEPTED   hook_ran=True
run_from_arrays(rewards=Hostile())        -> ValueError hook_ran=True
update(observation=Hostile())             -> TypeError  hook_ran=False   (outer gate catches it)
```

The public `update()` path is gated by an outer contract check. `run_option_value_duration_from_arrays` calls the helper directly and accepted the object outright.

## Scope — an outlier, not a sweep

I checked every `_require_array`-style helper in the tree: **14 of 18 already gate on metadata before converting**. I empirically probed the four that looked ungated — `gauntlet` and `recurring_multiagent` reject without running the hook, so my static heuristic over-flagged them. `option_value_duration` is the reachable one, and this PR is scoped to it.

`interaction_features._require_array_contract` also runs the hook when called directly, but I could not establish a public-API path reaching it with an untrusted object, so it is **not** claimed here.

## The change

Read metadata first, convert only after both checks pass — the same shape as `working_memory._require_array` and the other 13 gated helpers:

```python
        try:
            actual_shape = tuple(value.shape)
            actual_dtype = jnp.dtype(value.dtype)
        except Exception as error:
            raise TypeError(f"{name} must expose array shape and dtype metadata") from error
        if actual_shape != shape or actual_dtype != jnp.dtype(dtype):
            raise ValueError(f"{name} must have shape {shape} and dtype {dtype}")
        return jnp.asarray(value)
```

The existing shape/dtype `ValueError` message is unchanged, so callers relying on it are unaffected.

## Tests

Added to `tests/test_option_value_duration.py`:

- `test_array_runner_rejects_untrusted_inputs_without_running_conversion_hooks` — a `__array__` that raises if invoked, through both `option_indices` and `rewards`
- `test_array_runner_still_accepts_trusted_arrays` — positive control over `jnp` and `np.ndarray` inputs

Verified failing-first by reverting only the helper: **1 failed, 1 passed**, the failure being `ValueError: option_indices must be a readable array` — i.e. the hook ran and its `AssertionError` was swallowed by the `except`, which is exactly the defect.

## Verification

Commit `7624397d36296a8b73dcd273316abe837b27793a`, based on `cc877f78`.

```bash
python -m pytest -q -p no:randomly -o addopts= tests/test_option_value_duration.py
# 65 passed

```

Environment: Python 3.12.14, jax 0.11.0, numpy 2.5.2, CPU, network-isolated container.

---

### DIFF
--- a/alberta_framework/core/option_value_duration.py
+++ b/alberta_framework/core/option_value_duration.py
@@ -370,12 +370,15 @@ class OptionValueDurationLearner:
         dtype: Any,
     ) -> Array:
         try:
-            array = jnp.asarray(value)
+            actual_shape = tuple(value.shape)
+            actual_dtype = jnp.dtype(value.dtype)
         except Exception as error:
-            raise ValueError(f"{name} must be a readable array") from error
-        if array.shape != shape or array.dtype != dtype:
+            raise TypeError(
+                f"{name} must expose array shape and dtype metadata"
+            ) from error
+        if actual_shape != shape or actual_dtype != jnp.dtype(dtype):
             raise ValueError(f"{name} must have shape {shape} and dtype {dtype}")
-        return array
+        return jnp.asarray(value)
 
     @staticmethod
     def _state_values_valid(state: OptionValueDurationState) -> Array:
--- a/tests/test_option_value_duration.py
+++ b/tests/test_option_value_duration.py
@@ -642,3 +642,66 @@ def test_option_duration_array_runner_requires_exact_shapes() -> None:
             jnp.ones((2, 2), dtype=jnp.float32),
             jnp.zeros((2,), dtype=jnp.float32),
         )
+
+
+def test_array_runner_rejects_untrusted_inputs_without_running_conversion_hooks() -> None:
+    """The array runner must gate on array metadata before any conversion.
+
+    ``_require_array`` reached ``jnp.asarray(value)`` before validating
+    anything, so an untrusted ``__array__`` hook executed ahead of the shape
+    and dtype checks and chose the value they then saw. The public ``update``
+    path is gated by an outer contract check; the array runner called the
+    helper directly.
+    """
+
+    class _HostileArray:
+        def __array__(self, dtype: object = None, copy: object = None) -> np.ndarray:
+            raise AssertionError("array hook must not run")
+
+    learner = OptionValueDurationLearner(3, OptionValueDurationConfig())
+    state = learner.init(2)
+    steps = 2
+    observations = jnp.zeros((steps, 2), dtype=jnp.float32)
+    next_observations = jnp.zeros((steps, 2), dtype=jnp.float32)
+    rewards = jnp.zeros((steps,), dtype=jnp.float32)
+    discounts = jnp.zeros((steps,), dtype=jnp.float32)
+    option_indices = jnp.zeros((steps,), dtype=jnp.int32)
+
+    with pytest.raises(TypeError, match="shape and dtype metadata"):
+        run_option_value_duration_from_arrays(
+            learner,
+            state,
+            observations,
+            _HostileArray(),
+            rewards,
+            next_observations,
+            discounts,
+        )
+
+    with pytest.raises(TypeError, match="shape and dtype metadata"):
+        run_option_value_duration_from_arrays(
+            learner,
+            state,
+            observations,
+            option_indices,
+            _HostileArray(),
+            next_observations,
+            discounts,
+        )
+
+
+def test_array_runner_still_accepts_trusted_arrays() -> None:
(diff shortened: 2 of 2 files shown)


OutstandingRetard (approved):
This PR closes a real confused-deputy hole in `OptionValueDurationLearner._require_array`. The helper used to call `jnp.asarray(value)` *before* shape/dtype checks, so an untrusted `__array__` hook ran first and chose the array those checks then saw. The `except Exception` also relabeled hook failures as `"must be a readable array"`, hiding hostile input as malformed input.

The fix reads `value.shape` / `value.dtype` first, raises `TypeError` when metadata is missing, and only then converts. `test_array_runner_rejects_untrusted_inputs_without_running_conversion_hooks` is the right forcing function — a `_HostileArray` whose `__array__` would have been accepted on `cc877f78`.

No blocking concerns. Approve.

lalalune:
The underlying issue was valid. #1874 incorporated the contribution, completed the deeper cross-component repair, added the missing regression coverage, and has now merged to main. Closing this source PR as resolved by the strengthened implementation.

Issue #1851, opened 2026-08-18
ceiling_analyze reports population spread, escaping the sample-std unification from #33/#40


techforky:
## Problem

#33 established that across-seed spread must use the **sample** estimator, and #40 unified the code on it. That unification covered `alberta_framework` only. `outputs/ipmnist_screening/ceiling/ceiling_analyze.py` kept NumPy's population default, and its output is what `CEILING_ANALYSIS.md` publishes as `sd`:

```python
print(f"  {spec:20s} avg-online={np.mean(means):.5f} (seeds {sorted(runs)}, "
      f"sd {np.std(means):.5f})  ...")          # ddof=0
...
print(f"      late-task (last 10 tasks) avg-online = {late_tasks.mean():.5f} "
      f"(sd {late_tasks.std():.5f})")           # ddof=0
```

`np.std` defaults to `ddof=0`, so a 3-seed spread is understated by `sqrt(3/2)` = **1.22x** (41% at n=2).

## Actual behavior

Recomputed from the pinned `ceiling/analysis_summary.json` per-seed values at `cc877f78`. The published figures match `ddof=0` exactly:

| `CEILING_ANALYSIS.md` §1a row | published `sd` | `ddof=0` | `ddof=1` | understated |
|---|---|---|---|---|
| `sigma0_ndecay099` (champion) | 0.0028 | **0.00285** | 0.00349 | 1.22x |
| `adamw_control` | 0.0028 | **0.00282** | 0.00345 | 1.22x |
| `sgd_ema_norm` | 0.0027 | **0.00269** | 0.00330 | 1.22x |

The §1b carried-oracle row publishes `(sd 0.0007)`; the same n=2 batch-reference array in that artifact gives `ddof=0 -> 0.00070` against `ddof=1 -> 0.00099` (1.41x).

Meanwhile the screening merge (`ipmnist_screening.py:10069`, `:10107`) and `compute_statistics` both use `ddof=1`, so the same quantity is reported two ways depending on which path produced it — exactly the inconsistency #33 was filed about.

## Expected behavior

The analyzer should use the sample estimator, matching #33/#40 and the screening merge.

## Scope and what I am *not* claiming

- **No pinned artifact is rewritten.** `analysis_summary.json` does not record a `sd` field at all — the spread only ever existed in the printed report and in `CEILING_ANALYSIS.md`. Nothing under `outputs/` needs mutating to fix this.
[...]

PR #1854, opened 2026-08-18
fix(frontier-rank): restrict the confirm mean to seeds shared with the base

### LINKED ISSUE #1853: frontier_rank publishes an unpaired confirm mean beside a paired delta
## Problem

`outputs/ipmnist_screening/frontier_rank.py` publishes a confirm mean over one seed set beside a paired delta over another.

The screen path restricts to seeds shared with the base:

```python
shared = sorted(set(screen) & set(screen_base))
row["screen_mean"] = statistics.mean(screen[s] for s in shared)
```

The confirm path does not:

```python
cshared = sorted(set(confirm) & set(confirm_base))
if confirm:
    row["confirm_mean"] = statistics.mean(confirm.values())   # every seed the arm has
    row["n_confirm_seeds"] = len(confirm)                     # ditto
    if cshared:
        row["confirm_paired_delta_vs_base"] = statistics.mean(
            confirm[s] - confirm_base[s] for s in cshared     # only shared seeds
        )
```

So an arm with a confirm seed the base lacks would report `confirm_mean` over its full set directly next to `confirm_paired_delta_vs_base` over the shared subset, with `n_confirm_seeds` describing the former.

This is the condition `ipmnist_screening.merge_shards` **refuses** to rank:

> `seed sets differ across configs: {...}; merge_shards ranks configs on paired seeds only`

and

> `config {name!r} shares no seeds with control {control_name!r} ...; refusing to rank an unpaired entry in the summary`

The analyzer that reports on the campaign is therefore less strict than the merge it reports on — it fails open where the framework fails closed.

## This is latent, not a wrong published number

Stated plainly because it changes how much this matters: **no current output is affected.** Every arm's confirm seed set is `[0, 1, 2]` and matches the base, so `cshared == sorted(confirm)` throughout and `frontier_results.json` is byte-identical on a re-run. I verified this against the shards rather than assuming it.

The defect is that the guard is absent, not that it has already produced a bad number. It would bite the first time a confirm wave lands on a seed the base does not have — which is exactly when someone would be reading the confirm column.

## Expected behavior

Restrict `confirm_mean` and `n_confirm_seeds` to the shared subset, matching the screen path and the merge contract. With no shared seed, report no confirm mean at all rather than one implying a comparison.

## Environment

Python 3.12.14, numpy 2.5.2, CPU, network-isolated container built from `pyproject.toml` with `.[dev,gymnasium,forager]`.

PR to follow. This is the second finding in the untested `outputs/` scripts (see #1851); those 35 files still have no coverage beyond what these two PRs add.

---

techforky:
Fixes #1853.

## What moved

The ranker's screen path already restricts to seeds shared with the base. Its confirm path did not — `confirm_mean` and `n_confirm_seeds` covered the arm's whole seed set while `confirm_paired_delta_vs_base` beside them covered only the shared subset.

That is the condition `ipmnist_screening.merge_shards` **refuses** to rank:

> `seed sets differ across configs: {...}; merge_shards ranks configs on paired seeds only`

So the analyzer reporting on the campaign was less strict than the merge it reports on — failing open where the framework fails closed.

## This is latent — no published number changes

Stated up front because it bounds the claim: **every arm's confirm seed set is `[0, 1, 2]` and matches the base**, so `cshared == sorted(confirm)` throughout and `frontier_results.json` is byte-identical on a re-run. Verified against the shards, not assumed:

```text
arms whose confirm_mean would change: 0
```

The defect is a missing guard, not a bad number already emitted. It bites the first time a confirm wave lands on a seed the base lacks.

## The change

```python
        cshared = sorted(set(confirm) & set(confirm_base))
        if cshared:
            row["confirm_mean"] = statistics.mean(confirm[s] for s in cshared)
            row["n_confirm_seeds"] = len(cshared)
            row["confirm_paired_delta_vs_base"] = statistics.mean(
                confirm[s] - confirm_base[s] for s in cshared
            )
        elif confirm:
            row["n_confirm_seeds"] = 0
            row["confirm_unpaired_seeds"] = sorted(confirm)
```

With no shared seed the confirm mean is omitted entirely rather than emitted as a number implying a comparison, and the unpaired seeds are named so the omission is diagnosable.

I kept this to matching the screen path's existing behaviour rather than raising the way `merge_shards` does. Raising is arguably more in keeping with the fail-closed contract, but it would change this script's behaviour beyond the inconsistency being fixed — happy to switch if a maintainer prefers that.

## Tests

`tests/test_frontier_rank.py`, injecting through the `seed_means` seam:

- `test_confirm_mean_uses_only_seeds_shared_with_the_base` — a candidate-only seed at 0.99 must not enter the mean
- `test_confirm_block_is_omitted_when_no_seed_is_shared`
- `test_screen_and_confirm_agree_on_the_shared_seed_set`

Verified failing-first by reverting the fix: **2 failed, 1 passed** — the third passes either way because its seed sets already match, which is the latency in miniature.

## Verification

Commit `c8eb91456cf306cf88fde8bc82787f8940a54426`, based on `cc877f78`.

```bash
python -m pytest -q -p no:randomly -o addopts= tests/test_frontier_rank.py tests/test_ceiling_analyze.py
# 7 passed

```

## Context

Second finding in the untested `outputs/` scripts after #1851. Those 35 files had no coverage before these two PRs.

---

### DIFF
--- a/outputs/ipmnist_screening/ceiling/ceiling_analyze.py
+++ b/outputs/ipmnist_screening/ceiling/ceiling_analyze.py
@@ -29,6 +29,21 @@ def smooth(x: np.ndarray, w: int) -> np.ndarray:
     return np.convolve(x, np.ones(w) / w, mode="valid")
 
 
+def across_seed_spread(values: object) -> float:
+    """Sample standard deviation across seeds, matching the campaign convention.
+
+    Issue #33 unified across-seed spread on the sample estimator (``ddof=1``),
+    which the screening merge and ``compute_statistics`` both use. ``np.std``
+    defaults to the population estimator, understating the spread of a seed
+    sample by ``sqrt(n / (n - 1))`` — 22% at n=3, 41% at n=2. A single seed has
+    no spread rather than a zero-width interval implying one.
+    """
+    array = np.asarray(values, dtype=np.float64).ravel()
+    if array.size < 2:
+        return 0.0
+    return float(array.std(ddof=1))
+
+
 def main() -> None:
     report: dict = {}
 
@@ -44,7 +59,7 @@ def main() -> None:
         mc = curves.mean(axis=0)
         late = mc[LATE_LO:LATE_HI].mean()
         print(f"  {spec:20s} avg-online={np.mean(means):.5f} (seeds {sorted(runs)}, "
-              f"sd {np.std(means):.5f})  late-window(4000-5000)={late:.5f}")
+              f"sd {across_seed_spread(means):.5f})  late-window(4000-5000)={late:.5f}")
         bucket_acc = {f"{a}-{b}": round(float(mc[a:b].mean()), 5) for a, b in BUCKETS}
         print(f"      within-task buckets: {bucket_acc}")
         report[f"stationary_{spec}"] = {
@@ -77,7 +92,7 @@ def main() -> None:
               f"t10={pt[:, 9].mean():.4f} t20={pt[:, 19].mean():.4f} "
               f"t40={pt[:, 39].mean():.4f} t60={pt[:, -1].mean():.4f}")
         print(f"      late-task (last 10 tasks) avg-online = {late_tasks.mean():.5f} "
-              f"(sd {late_tasks.std():.5f})")
+              f"(sd {across_seed_spread(late_tasks):.5f})")
         print(f"      late-task late-window (steps 4000-5000) = {last_task_late.mean():.5f}")
         report[f"carried_{spec}"] = {
             "per_task_mean_curve": [round(float(v), 5) for v in pt.mean(axis=0)],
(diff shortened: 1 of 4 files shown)


OutstandingRetard (approved):
This PR fixes two published-number bugs in the IPMNIST screening outputs, not a style pass.

1. `frontier_rank.py` printed `confirm_mean` over the full confirm seed set next to a paired delta that only used shared seeds. That is an unpaired mean beside a paired comparison — the exact class `merge_shards` already refuses. Restricting `confirm_mean` / `n_confirm_seeds` / the paired delta to `cshared` makes the three fields describe the same seed set.

2. `ceiling_analyze.py` used `np.std` (population, ddof=0) for across-seed spread after issue #33 unified the campaign on the sample estimator. At n=3 that understates spread by ~22%. `across_seed_spread` with `ddof=1`, and no spread for a single seed, matches `compute_statistics`.

`tests/test_frontier_rank.py` and `tests/test_ceiling_analyze.py` pin both contracts. No blocking concerns. Approve.

lalalune:
The issue was valid, but editing active output-tree scripts was not a maintainable fix. #1874 superseded this with tested package-level campaign tooling outside immutable/append-only outputs and has now merged to main. Closing as resolved by the durable implementation.

Issue #1870, opened 2026-08-18
Preregistration: V5 — model-side identification against reference-side weights


techforky:
# V5 — Model-side identification against reference-side weights

Pre-registration. **Posted before running anything.** Nothing below may be
edited once numbers exist; a failed gate is a valid rejection and gets recorded
as such.

Development diagnostic, permanently nonpromoting
(`development_screening_diagnostic`). Touches no pinned artifact, edits no
registered source, consumes no promotion seed, uses only the already-consumed
development seeds 0-2.

## Why this exists

Ledger entry 15 (`docs/evidence/negative-results.md`, merged via #1335) records
that V4's model-side arm was **void, not refuted**, and names the correction:

> A corrected probe must score post-shift content against reference-side
> weights rather than correlating both sides through one forward pass.

V4's F4 failed its oracle gate at 0.002/0.000 while scoring 1.000 on the
no-shift control — it recovered the identity map and not the permutation,
because `a_h = relu(sum_k x_k w1[k,h])` makes each position's coupling
dominated by the weight *at* that position, so both sides reduced to rows of
`w1`. V5 breaks that symmetry: the reference descriptor comes from the
**model only** and the post descriptor from **post-shift data only**, so no
shared forward pass can collapse them onto position.

Until this runs, entry 15 leaves model-side probes untested rather than closed.

## The question, stated as a comparison

V1 matched post-shift class-conditional means against a *data-side* reference
(the champion's annealed EMAs) and reached 0.785 at N=500 against a 0.90 bar,
with `N* ~= 2000`. Its diagnosis was that the **estimator**, not the
fingerprint, binds.

V5 asks exactly one thing: **does replacing the noisy data-side reference with
a model-side one — trained on the full 5,000-sample pre-shift task, and so far
better estimated — move that floor?**

## Hypothesis

**H1.** A model-side reference identifies >90% of relevant pixels within 500
post-shift samples, beating V1's data-side reference at the same budget.

**H0.** The post-shift side is the binding constraint. Class-conditional
estimates still see ~N/10 samples per class regardless of how good the
reference is, so the floor does not move and model-side identification closes
with a number attached.

**I expect H0**, and say so before any numbers exist. V1 located the
constraint in the post-shift estimator; nothing in V5 changes that side. The
value here is converting entry 15 from *void* to *measured*, not a win.

## Protocol — inherited from V1/V4, unchanged

- Data: `load_mnist_train`; permutations: `build_schedule` per-seed.
[...]

PR #1873, opened 2026-08-18
research(new-directions): V5 model-side identification — V1 floor confirmed, arms void

techforky:
Executes the preregistration in #1870. Development diagnostic, permanently nonpromoting. No pinned artifact touched, no registered source edited, no promotion seed consumed, no change to `alberta_framework/`.

### 1. V1's information floor replicates independently

The data-side control arm crosses 0.90 at **`N* = 1978.5`** against V1's reported ~2,000 — a different runner, a different implementation, and a *stronger* post estimator.

| N | F5c Hungarian mean / min | V1 published |
|---|---|---|
| 50 | 0.2063 / 0.1187 | 0.197 |
| 200 | 0.6527 / 0.5800 | 0.619 |
| **500** | **0.8044 / 0.7840** | **0.785** |
| 2000 | 0.9014 / 0.8604 | 0.840 |

**This is V1's fingerprint, not V1 verbatim.** The preregistration fixed the post descriptor as *batch* class-conditional means; V1 used a fresh annealed EMA. They agree closely at N=50–500, where a ~100-sample EMA window is near the batch mean, and diverge at N=2000 where the batch estimator keeps all samples. F5c is therefore slightly stronger than V1's estimator, which makes its floor a **conservative** bound — and it still lands on ~2,000.

### 2. Both model-side arms are VOID

| control (exact statistics) | F5a | F5b | F5c |
|---|---|---|---|
| oracle — permutation differs | 0.248 | 0.248 | **1.000** |
| no-shift — identity map | 0.248 | 0.248 | **1.000** |

Failed the pre-registered 0.95 oracle gate. Online values are recorded and marked `void`; they are **not** evidence about model-side probes.

**This is not entry 15's failure.** V4's F4 scored 1.000 on no-shift and 0.002 on the oracle — a position confound. V5's arms score *identically with and without a permutation*, so position is not the issue. They simply do not identify pixels: 0.248 with exact statistics, above chance (~0.002) but far below the gate.

## The structural finding

The reference is the model's per-position class affinity — **weight space**, "how does position `i` drive class `c`". The post descriptor is class-conditional intensity — **data space**, "how bright is position `j` for class `c`". Correlated, which is why 0.248 isn't chance, but not the same object.

This generalizes:

> Any model-side quantity is indexed by the model's **input position**, which is the pre-shift layout. Scoring it against post-shift evidence requires a post-shift quantity, and the only permutation-invariant one available is a **content statistic** — exactly what V1 measured.

So a model-side probe can only improve the **reference**. V1 already established the reference is not binding; the post estimator is, because class-conditional statistics see ~N/10 samples per class however good the reference gets. That is why H0 was the registered prediction and why F5c's floor lands where V1's did.

**Entry 15 stays open.** Model-side probes remain *untested* across two constructions — this run adds a structural reason they are hard to construct, not a refutation. Ledger entry 16 records both outcomes.

## Controls — and a mis-specification I caught

Computed for every arm **before any online checkpoint**, correcting V4's execution-order deviation.

The oracle uses **exact** class means. My first draft fed it the online annealed EMA, which reported 0.406 and would have marked the V1 control arm void *by construction* — an EMA with decay 0.99 has a ~100-sample effective window and never converges to a full-dataset mean. Caught before any headline number existed; the corrected control reproduces V1's 100% class-conditional oracle exactly, which is what gives the 1.000 row above its meaning.

## Verification

Commit `cd0bb80a641b5a92871f7d6073693f61516964dc`, based on `cc877f78`.

```bash
python outputs/new_directions/V5_model_side_runner.py \
  --data-home <openml cache> --out outputs/new_directions/V5_model_side.json
# 216 cells + both controls in 73.3s

```

Rerun is **bit-identical**. Environment: Python 3.12.14, jax 0.11.0, numpy 2.5.2, CPU, network-isolated container. Reference network `sigma0_ndecay099`, trained online through tasks 0..boundary.

## What remains open

Any future model-side attempt must state in advance what post-shift quantity it scores against, and why that quantity is not itself the binding estimator. I do not have a construction that satisfies that, which is why this closes as *untested* rather than refuted.

---


techforky:
## Scope limit on the structural argument — flagging my own overstatement

The PR body and ledger entry 16 state:

> Any model-side quantity is indexed by the model's **input position**, which is the pre-shift layout. Scoring it against post-shift evidence requires a post-shift quantity, and the only permutation-invariant one available is a **content statistic**.

That holds for **novel** permutations, which is the IPMNIST protocol's case — `build_schedule` draws a fresh permutation every task, so nothing about the new layout can be known in advance. It does **not** hold in general, and I should have bounded it when I wrote it.

**The exception is recurrence.** If permutations are drawn from a pool that repeats, a learner can store per-permutation state and recognise a *repeat* by matching whole-input behaviour — a loss or accuracy signature against stored signatures — rather than by identifying pixels one at a time. That is not a per-pixel content statistic and it does not pay the ~N/10-per-class cost that binds V1's estimator, because it is recognition rather than identification.

This matters because it is exactly the live direction named in `CEILING_ANALYSIS.md` as recurrence-indexed context reuse (direction D), and `micro_continual` already ships an **M4 `recurrence` family** — "coordinate permutations drawn from a small pool" — built to test it.

So the correct scope is:

- **Confirmed here:** for a novel permutation, a model-side probe can improve the reference and cannot move the binding post-shift estimator.
- **Not addressed here:** recurrence, where the identification problem is replaced by a recognition problem and the argument does not apply.

Nothing in the measured results changes — F5a/F5b remain void, `N* = 1978.5` stands, entry 15 stays open. Only the generality of the accompanying claim is narrowed, and I would rather narrow it myself than have it read wider than the evidence supports.

If a maintainer prefers, I can amend ledger entry 16 to carry this bound inline rather than leaving it in the PR discussion.

lalalune (changes requested):
Blocking evidence-contract issues:\n\n1. The preregistration requires aborting before online checkpoints when either sanity gate fails, but  computes the failed verdict and still generates all 216 online cells. Preserve the raw artifact, disclose this protocol deviation, and treat those model-side rows as void—not literal prereg execution.\n2. A permanently nonpromoting development runner computes  from observed screen data. Remove this auto-promotion route; record only a development screening decision and hard-code scientific promotion disallowed.\n3. “Independent replication/confirmed” is too strong for the same MNIST/schedule/seeds and a hybrid stronger estimator. Scope it as a same-stack descriptive consistency check/lower bound.\n4. Incorporate the recurrence/novel-permutation limitation from the PR comment into the retained report and ledger; two void constructions do not justify the universal model-side claim.\n5. Add a versioned append-only companion audit/validator binding runner/source, dependencies/runtime, dataset identity, invocation, exact seeds/cells, finite primitive JSON, and nonpromotion. Do not rewrite observed values.\n\nPlease rebase on current main and add tests proving a failed control invokes no online scoring and that no development result can become promotable.

lalalune (changes requested):
Blocking evidence-contract issues:

1. The preregistration requires aborting before online checkpoints when either sanity gate fails, but `V5_model_side_runner.py` computes the failed verdict and still generates all 216 online cells. Preserve the raw artifact, disclose this protocol deviation, and treat those model-side rows as void—not literal prereg execution.
2. A permanently nonpromoting development runner computes `promotion.promoted` from observed screen data. Remove this auto-promotion route; record only a development screening decision and hard-code scientific promotion disallowed.
3. “Independent replication/confirmed” is too strong for the same MNIST/schedule/seeds and a hybrid stronger estimator. Scope it as a same-stack descriptive consistency check/lower bound.
4. Incorporate the recurrence/novel-permutation limitation from the PR comment into the retained report and ledger; two void constructions do not justify the universal model-side claim.
5. Add a versioned append-only companion audit/validator binding runner/source, dependencies/runtime, dataset identity, invocation, exact seeds/cells, finite primitive JSON, and nonpromotion. Do not rewrite observed values.

Please rebase on current main and add tests proving a failed control invokes no online scoring and that no development result can become promotable.

lalalune:
Superseded by the append-only, strict-audit replacement #1926. The original raw V5 files are preserved byte-for-byte; the replacement correctly classifies the preregistered execution as invalid and assigns zero valid online cells.

Issue #1875, opened 2026-08-18
Preregistration: V6 — is recurrence unexploited? measuring the headroom for direction D


techforky:
# V6 — Is recurrence unexploited? Measuring the headroom for direction D

Pre-registration. **Posted before running the comparison.** Nothing below may be
edited once numbers exist; a failed gate is a valid rejection.

Development diagnostic, permanently nonpromoting
(`development_screening_diagnostic`). Micro suite only — touches no IPMNIST
lane, no pinned artifact, no registered source, no promotion seed.

## Why this, and why now

V5 (#1870, #1873) closed with a structural claim, which I then bounded myself
in the PR discussion:

> Any model-side quantity is indexed by the pre-shift input layout, so scoring
> it requires a post-shift content statistic — the estimator V1 showed to be
> binding.

That holds for a **novel** permutation. It does not hold under **recurrence**,
where identification is replaced by *recognition*: a learner can match
whole-input behaviour against stored per-permutation state instead of
re-identifying pixels and paying the ~N/10-per-class cost.

`CEILING_ANALYSIS.md` names recurrence-indexed context reuse as a live,
unexecuted direction, and `micro_continual` already ships the family to test
it — **M4 `recurrence`**, "coordinate permutations drawn from a small pool".

Before proposing any recurrence-aware mechanism, one number is missing: **do
the existing arms already exploit recurrence, and if not, how much is on the
table?** V6 measures exactly that and nothing else.

## Hypothesis

**H1.** One or more existing ladder arms score materially better on M4 than on
M1, i.e. current mechanisms already capture some benefit from repeating
permutations.

**H0.** No existing arm exploits recurrence. None stores per-permutation state,
so a repeated permutation is indistinguishable from a fresh one and the gap is
noise. The direction-D headroom is therefore intact and unclaimed.

**I expect H0**, stated before any comparison is run. The arms are
`sgd_raw`, `adamw`, `upgd_raw`, `sgd_norm`, `gated_norm`, `naive_bayes` — none
has a mechanism that could index a permutation. The value of this run is the
**size of the untouched headroom**, not a win.

## Protocol

- Runner: the shipped `alberta_framework.benchmarks.micro_continual` CLI. No
  new benchmark code.
- Families: **M1 `input_permutation`** (fresh permutation per regime) vs
  **M4 `recurrence`** (`recurrence_pool = 5`, the shipped default).
[...]

PR #1883, opened 2026-08-18
research(new-directions): V6 recurrence headroom — H0 refuted, conditioning takes most of it

techforky:
Executes the preregistration in #1875. Development diagnostic, permanently nonpromoting. Micro suite only — no IPMNIST lane, no pinned artifact, no registered source, no promotion seed, no change to `alberta_framework/`.

## Verdict: my registered prediction is REFUTED

I registered H0 — "no existing arm exploits recurrence" — and it is wrong. All six ladder arms clear the pre-registered criterion, mean paired gap `> 0` with all three seeds positive.

| arm | M1 `input_permutation` | M4 `recurrence` | gap |
|---|---|---|---|
| `sgd_raw` | 0.2827 | 0.6696 | **+0.3869** |
| `adamw` | 0.2281 | 0.5788 | **+0.3507** |
| `upgd_raw` | 0.2797 | 0.3726 | +0.0929 |
| `sgd_norm` | 0.6808 | 0.7270 | +0.0461 |
| `gated_norm` | 0.6911 | 0.7383 | +0.0472 |
| `naive_bayes` | 0.6052 | 0.6526 | +0.0474 |

**Where the reasoning failed.** I argued "none of these arms indexes permutations, therefore none can exploit recurrence." That conflates *having a recurrence mechanism* with *benefiting from recurrence*. No arm indexes permutations — this result does not show otherwise — but every arm carries persistent weights, and when a mapping repeats those weights are already partly fit. The benefit is implicit reuse, not recognition.

So **"exploits recurrence" here means "scores better when permutations repeat", not "implements a recurrence mechanism"**, and that distinction bounds every downstream claim.

## The structure is the finding

| group | mean gap |
|---|---|
| unconditioned (`sgd_raw`, `adamw`) | **+0.3688** |
| conditioned (`sgd_norm`, `gated_norm`, `naive_bayes`) | **+0.0469** |
| ratio | **7.9x** |

Arms that are bad on fresh permutations gain enormously from repetition; arms that are good on fresh permutations gain almost nothing. A conditioned arm re-adapts fast enough that a repeat is barely cheaper than a fresh permutation; an unconditioned arm re-adapts slowly and a repeat saves most of that cost.

This mirrors the campaign's central result — conditioning is the load-bearing mechanism — measured on the memory axis rather than the input-shift axis.

## What this does and does not bound for direction D

**Measured.** With 20x reuse (5 permutations across 100 regimes), a champion-like conditioned arm gains **+0.047**. That is what implicit weight retention already captures, and it is the baseline an explicit recurrence-indexing mechanism must beat rather than a starting point of zero.

**Not bounded here.** This does *not* cap direction D. An explicit mechanism could jump straight to stored per-permutation state rather than partially re-adapting, and would then beat implicit retention. The distance from the best M4 arm to Bayes is `0.9833 - 0.7383 = 0.2450`, but most of that is ordinary online estimation cost, so it bounds the direction only in the loosest sense.

## Controls

Run first, reported regardless:

| family | regimes | distinct permutations |
|---|---|---|
| M1 | 100 | 100 — all fresh |
| M4 | 100 | **5** — exactly `recurrence_pool` |

Had the families been secretly identical, every gap would be zero and would have read as clean support for H0 — indistinguishable from the experiment's own failure mode. That is why it ran first.

**Bayes accuracy is family-invariant at 0.9833** (`bayes_reference`, 200k samples, `mc_sem` 0.0003), identical for both families. I verified this rather than assuming it: a family-dependent ceiling would have made every number above meaningless.

## Deviation

Wall clock **1911.6 s (32 min)** against the ~14 min I estimated in the preregistration. The estimate timed `gated_norm`; the raw arms and `naive_bayes` are slower. No protocol term changed — only my cost estimate was wrong.

## Verification

Commit `07034c49baad9ca9de07a602366ae0a4b5a9d948`, based on `cc877f78`. 36 runs plus the control, using the shipped `micro_continual` runner with no new benchmark code. `ruff` clean.

## Note on ledger numbering

The new entry is numbered **16** against current `main`. #1873 (V5) also adds a 16; whichever lands first, the other needs a renumber. Happy to rebase.

---

### DIFF
--- a/docs/evidence/negative-results.md
+++ b/docs/evidence/negative-results.md
@@ -133,6 +133,27 @@ postmortem.
     6.55-6.82, so the nominal 300 dimensions were under 7. Record:
     [`V4_fingerprints.md`](../../outputs/new_directions/V4_fingerprints.md).
 
+16. **Every ladder arm already benefits from recurrence, and conditioning takes
+    most of it.** The #1875 preregistration predicted that no existing arm
+    exploits recurrence, reasoning that none indexes permutations. That is
+    refuted: on the micro suite all six ladder arms score higher on M4
+    (`recurrence`, pool 5 over 100 regimes) than on M1 (`input_permutation`),
+    every seed positive. The prediction conflated *having a recurrence
+    mechanism* with *benefiting from recurrence* — no arm indexes permutations,
+    but persistent weights stay partly fit when a mapping repeats. The
+    structure is the result: unconditioned arms gain +0.3688 mean (`sgd_raw`
+    +0.3869, `adamw` +0.3507) while conditioned arms gain +0.0469 (`gated_norm`
+    +0.0472), a 7.9x difference. Input conditioning already recovers most of
+    what implicit reuse offers, mirroring the campaign's conditioning result on
+    the memory axis. For direction D the reusable number is that 20x reuse is
+    worth only **+0.047** on a champion-like arm, so an explicit
+    recurrence-indexing mechanism must beat that implicit baseline rather than
+    start from zero. This does **not** cap direction D: an explicit mechanism
+    could jump to stored state instead of re-adapting. Bayes accuracy is
+    family-invariant at 0.9833, verified rather than assumed, so the gaps are
+    learning gains and not a ceiling difference. Record:
+    [`V6_recurrence_headroom.md`](../../outputs/new_directions/V6_recurrence_headroom.md).
+
 ## Evidence and campaign closures
 
 1. **Continual-IA v1 is a valid rejection at its frozen gate.** Reward uplift
--- a/outputs/new_directions/V6_recurrence_headroom.json
+++ b/outputs/new_directions/V6_recurrence_headroom.json
@@ -1,0 +1,353 @@
+{
+ "control": {
+  "input_permutation": {
+   "distinct_permutations": 100,
+   "n_regimes": 100,
+   "recurrence_pool": 5
+  },
+  "recurrence": {
+   "distinct_permutations": 5,
+   "n_regimes": 100,
+   "recurrence_pool": 5
+  },
+  "separated": true
+ },
+ "evidence_policy": "development_screening_diagnostic; permanently nonpromoting. Micro suite only; no IPMNIST lane, pinned artifact, or registered source.",
+ "outcome": {
+  "any_arm_exploits_recurrence": true,
+  "arms_exploiting_recurrence": [
+   "sgd_raw",
+   "adamw",
+   "upgd_raw",
+   "sgd_norm",
+   "gated_norm",
+   "naive_bayes"
+  ],
+  "best_m4_arm": "gated_norm",
+  "best_m4_mean": 0.738268,
+  "criterion": "mean paired gap > 0 with all seeds positive"
+ },
+ "paired_gaps": [
+  {
+   "all_seeds_positive": true,
+   "arm": "sgd_raw",
+   "exploits_recurrence": true,
+   "m1_mean": 0.28272200000000003,
+   "m4_mean": 0.66959,
+   "mean_gap": 0.386868,
+   "per_seed_gap": [
+    0.37491,
+    0.414778,
+    0.370916
+   ]
+  },
+  {
+   "all_seeds_positive": true,
+   "arm": "adamw",
+   "exploits_recurrence": true,
+   "m1_mean": 0.22813133333333332,
+   "m4_mean": 0.5787873333333333,
+   "mean_gap": 0.350656,
+   "per_seed_gap": [
+    0.340882,
+    0.374018,
+    0.337068
+   ]
+  },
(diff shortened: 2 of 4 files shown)


techforky:
## Scoping note: direction D does not apply to the IPMNIST lane

Recorded because it bounds what this PR's number is good for, and it partly undercuts the practical value of the work above.

`CEILING_ANALYSIS.md` lists recurrence-indexed context reuse (direction D) among the live candidates for attacking the champion's transient. But the IPMNIST protocol draws a fresh permutation every task:

```python
# upgd_ipmnist.build_schedule
permutations = jax.vmap(
    lambda task: jr.permutation(jr.fold_in(key_perm, task), config.input_dim)
)(tasks)
```

Measured at the published protocol (200 tasks, 784 dims):

```text
seed 0: 200 tasks -> 200 distinct permutations  repeats=0
seed 1: 200 tasks -> 200 distinct permutations  repeats=0
seed 2: 200 tasks -> 200 distinct permutations  repeats=0
```

Zero repeats, and structurally there can be none — the draw is uniform over 784! and independent per task.

**So a recurrence-indexed mechanism has nothing to index on this lane.** No amount of stored per-permutation state helps when no permutation ever returns.

### What that means for this PR

The +0.047 measured here is on the micro suite's M4 family, which is the only place in the repository where recurrence exists at all. It is a real number about recurrence; it is **not** a number about the champion lane's transient.

Combining with the result above, the scoping answer for direction D is:

1. **The IPMNIST lane has no recurrence**, so direction D cannot help the campaign's current transient at all.
2. **On the one lane that does have recurrence**, conditioned arms already capture most of the benefit implicitly — 20x reuse is worth +0.047 to `gated_norm`.
3. Therefore direction D would require both a *new recurring lane* and an explicit mechanism that beats the +0.047 implicit baseline, before it could pay anything.

That is a scoping conclusion about priority, **not** a claim that a recurrence mechanism is impossible or that direction D is refuted. It says the direction is inapplicable where the transient actually lives, and that where it is applicable, most of the easy benefit is already taken.

I would rather state this against my own work than leave the +0.047 to be read as headroom on the champion lane, which it is not.

lalalune (changes requested):
Blocking evidence-contract issues:\n\n1. The preregistration says family separation is checked across the run, but the runner checks only . Add a post-hoc control audit for every exact seed and disclose the original execution deviation.\n2. The report claims a verified Bayes ceiling and computes headroom from 0.9833, but the runner/JSON contains no Bayes control. The repository’s three-seed mean differs from seed 0. Add per-seed Bayes raw values/MC SEM in a companion audit, aggregate on matched seeds, and recompute.\n3. The highlighted 7.9× grouping is post hoc, omits , and classifies  as conditioned despite its distinct mechanism. A smaller gap at a higher baseline is not a causal conditioning ablation. Keep the registered all-arm paired table primary and label any complete, predeclared taxonomy exploratory/descriptive.\n4. With three consumed development seeds, say the registered criterion was met in all three seeds; do not make an unqualified generalization.\n5. Bind exact arms, source/runner, dependencies/runtime, dataset/config/invocation, nonpromotion, and all raw/derived values in a versioned append-only validator.\n\nRebase after #1873 (renumber the conflicting ledger entry), retain the IPMNIST-nonrecurrence scope in-repository, and add tests that all controls complete before any run cell and any failure yields void/no accuracies.

lalalune (changes requested):
Blocking evidence-contract issues:

1. The preregistration says family separation is checked across the run, but the runner checks only `args.seeds[0]`. Add a post-hoc control audit for every exact seed and disclose the original execution deviation.
2. The report claims a verified Bayes ceiling and computes headroom from 0.9833, but the runner/JSON contains no Bayes control. The repository’s three-seed mean differs from seed 0. Add per-seed Bayes raw values/MC SEM in a companion audit, aggregate on matched seeds, and recompute.
3. The highlighted 7.9× grouping is post hoc, omits `upgd_raw`, and classifies `naive_bayes` as conditioned despite its distinct mechanism. A smaller gap at a higher baseline is not a causal conditioning ablation. Keep the registered all-arm paired table primary and label any complete, predeclared taxonomy exploratory/descriptive.
4. With three consumed development seeds, say the registered criterion was met in all three seeds; do not make an unqualified generalization.
5. Bind exact arms, source/runner, dependencies/runtime, dataset/config/invocation, nonpromotion, and all raw/derived values in a versioned append-only validator.

Rebase after #1873 (renumber the conflicting ledger entry), retain the IPMNIST-nonrecurrence scope in-repository, and add tests that all controls complete before any run cell and any failure yields void/no accuracies.

lalalune:
Superseded by the append-only, strict-audit replacement #1926. The original raw V6 files are preserved byte-for-byte; the replacement retains the observations only as an amended, inconclusive development result with all-seed controls and matched Bayes recomputation.

PR #1885, opened 2026-08-18
fix(benchmarks): close trust-boundary residuals in forager_matched_qualification sink

Kayvan-Zahiri:
Closes 3 trust-boundary residuals in `forager_matched_qualification.py` (2234 lines, evidence-safe, 3 leaks):

- Adds `_require_exact_str` host gate before duplicate key checks and non-finite float parsing
- Sanitizes duplicate JSON object key: `host_key` before membership test and `f"duplicate JSON key '{host_key}'"`
- Sanitizes non-finite JSON constant and invalid/non-finite JSON number: `host_value` before `float()` and quoted `'{host_value}'`

Validation: ruff 0, mypy PASS, grep -c '!r' 0 (was 3), pytest 8 passed (hostile trust). Evidence-safe (not in evidence_manifest.py).

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_qualification.py
+++ b/alberta_framework/benchmarks/forager_matched_qualification.py
@@ -525,23 +525,37 @@ def _freeze_json(value: Any) -> Any:
     return value
 
 
+def _require_exact_str(name: object, value: object) -> str:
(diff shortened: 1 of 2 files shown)


lalalune:

PR #1886, opened 2026-08-18
fix(benchmarks): close trust-boundary residuals in forager_matched_final_analysis sink

Kayvan-Zahiri:
Closes 3 trust-boundary residuals in `forager_matched_final_analysis.py` (2822 lines, evidence-safe, 3 leaks):

- Adds host gate via `_require_identifier` before entrypoint lookup and unknown candidate error formatting
- Sanitizes frozen entrypoint binding and unknown candidate errors: `host_candidate_id` before lookup/membership test and quoted `'{host_candidate_id}'`

Validation: ruff 0, mypy PASS, grep -c '!r' 0 (was 3), pytest 6 passed (hostile trust). Evidence-safe (not in evidence_manifest.py).


lalalune:

PR #1888, opened 2026-08-18
fix(benchmarks): close trust-boundary residuals in reference_life_scorecard sink

Kayvan-Zahiri:
Closes 6 trust-boundary residuals in `reference_life_scorecard.py` (3040 lines, evidence-safe, 6 leaks):

- Adds `_require_exact_str` host gate before duplicate key checks, non-finite constant parsing, environment matching, and arm resolution
- Sanitizes duplicate JSON key and constant errors: `host_key` before membership test and quoted `'{host_key}'`
- Sanitizes unsupported environment and control arm errors: `host_env`/`host_arm` before raises and quoted `'{host_env}'`/`'{host_arm}'`
- Sanitizes run identity collision and missing scheduled identity errors

Validation: ruff 0, mypy PASS, grep -c '!r' 0 (was 6), pytest 6 passed (hostile trust). Evidence-safe (not in evidence_manifest.py).

### DIFF
--- a/alberta_framework/benchmarks/reference_life_scorecard.py
+++ b/alberta_framework/benchmarks/reference_life_scorecard.py
@@ -1,16 +1,3 @@
-"""Matched, permanently nonpromoting reference-life development scorecard.
-
-The scorecard is a development-selection instrument, never scientific evidence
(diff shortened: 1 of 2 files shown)


lalalune:

PR #1908, opened 2026-08-18
fix(dreaming): saturate DreamRolloutState.step_count at int32 max

Kayvan-Zahiri:
Saturating lifetime clock keeps `DreamRolloutState.step_count` non-negative:

- In `dream_one_step`, `step_count` was previously incremented with a bare `+ 1` on an `int32` array leaf.
- At `INT32_MAX`, a bare increment overflows to negative (`INT32_MIN`), violating monotonicity assumptions.

Validation: ruff 0, mypy PASS, pytest 81 passed (`test_dreaming_lifetime_clock.py` + `test_dreaming.py`). Evidence-safe (unpinned).

### DIFF
--- a/alberta_framework/core/dreaming.py
+++ b/alberta_framework/core/dreaming.py
@@ -41,6 +41,7 @@ from alberta_framework.core.behavior_model import (
     floor_and_renormalize_probabilities,
     selected_action_probabilities,
 )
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by #1917, which rebases this fix onto current main, preserves the contributor commit, removes unrelated formatter churn where applicable, and validates the combined lifetime/timing panel.

PR #1909, opened 2026-08-18
fix(step7): saturate Step7DynaState.step_count at int32 max

Kayvan-Zahiri:
Saturating lifetime clock keeps `Step7DynaState.step_count` non-negative:

- In `step7_update` (`alberta_framework/steps/step7.py`), `step_count` was advanced with a bare `+ 1` on an `int32` array.
- At `INT32_MAX`, bare addition wraps to negative (`INT32_MIN`), breaking downstream monotonicity assumptions in checkpoints, telemetry, and evaluation gates.

Validation: ruff 0, mypy PASS, pytest 3 passed (`tests/test_step7_lifetime_clock.py`). Evidence-safe (unpinned).

### DIFF
--- a/alberta_framework/steps/step7.py
+++ b/alberta_framework/steps/step7.py
@@ -65,6 +65,7 @@ from alberta_framework.core.average_reward import (
     DifferentialSARSAState,
     DifferentialSARSAUpdateResult,
 )
+from alberta_framework.core.normalizers import _saturating_int32_counter_increment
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded by #1917, which rebases this fix onto current main, preserves the contributor commit, removes unrelated formatter churn where applicable, and validates the combined lifetime/timing panel.

PR #1842, opened 2026-08-18
fix(core): saturate UPGDLearner step_count at int32 max

deepanshu-yd:
Closes #1841

## Problem

`UPGDLearner.update` advanced the lifetime `step_count` with a bare `state.step_count + 1` (`alberta_framework/core/upgd.py:3330`). The counter is declared `dtype=jnp.int32` and has no saturating increment. At `step_count == 2**31 - 1` the next update wraps the clock to `INT32_MIN`.

The wrapped negative clock flips the warm-up, meta-warm, and perturbation-schedule branches inside `update` (`upgd.py:2393`, `2418`, `2873-2876`, `3207`, `3242`) to `False` permanently, so an exhausted learner no longer reproduces its pre-wrap update behavior.

## Fix

Replace the bare increment with the tree-wide saturating int32 counter increment (`normalizers._saturating_int32_counter_increment`), matching the existing lifetime-clock fixes in learners, synthetic streams, feature discovery, horde, and reward_model.

## Tests

Adds `tests/test_upgd_lifetime_clock.py`: plants the exhausted clock and asserts the counter saturates at `INT32_MAX` and never wraps negative.

- `pytest tests/test_upgd.py tests/test_upgd_lifetime_clock.py`: 94 passed
- `ruff check`: clean
- `mypy`: clean

PR #1788, opened 2026-08-18
fix(horde): saturate MixedHorde step_count at int32 max

Svector-anu:
Closes #1787.

## What changed

`MixedHorde.update` incremented its persistent `step_count` with bare addition. A successful update at `2**31 - 1` wrapped the counter to `-2147483648`.

## Fix

Use the existing shared `_saturating_int32_counter_increment` helper from `alberta_framework.core.normalizers`, already imported by this module.

Same bug class fixed 8+ times already this session in sibling files.

## Reachability

`MixedHorde` is publicly exported from both `alberta_framework.core` and the package root. Ordinary callers can supply and resume a `MixedHordeState`, so this is reachable through the public learner API with non-hardcoded inputs.

## Verification

Live reproduction against the unpatched tree:

```text
before 2147483647 after -2147483648 applied True
```

After the fix, same initial state:

```text
before 2147483647 after 2147483647 applied True
```

Added `test_update_saturates_step_count_at_int32_max`.

Mutation check (source fix reverted, test kept):

```text
FAILED tests/test_mixed_horde.py::TestAllSharedEqualsHordeLearner::test_update_saturates_step_count_at_int32_max
assert -2147483648 == ((2 ** 31) - 1)
1 failed, 19 deselected
```

Fix restored: `1 passed, 19 deselected`.

Full `tests/test_mixed_horde.py`: `20 passed`.

`ruff check` and `mypy` both clean on the touched files.

## Provenance

AI-assisted: drafted by OpenAI Codex (`gpt-5.6-sol`, via AgentRouter), independently re-verified end to end before this PR was opened — reproduction, mutation check (both directions), full test file, lint, mypy, and the reachability trace above were all re-run and re-confirmed by hand, not taken from the draft's own report.

Attribution status: self-reported.

PR #1910, opened 2026-08-18
fix(horde): saturate MixedHorde step_count at int32 max

Kayvan-Zahiri:
Saturating lifetime clock keeps `MixedHordeState.step_count` non-negative:

- In `MixedHorde.update` (`alberta_framework/core/horde.py`), `step_count` was advanced with a bare `+ 1` on an `int32` array leaf.
- At `INT32_MAX`, bare addition wraps to negative (`INT32_MIN`), breaking downstream monotonicity assumptions in checkpoints, telemetry, and evaluation gates.

Validation: ruff 0, mypy PASS, pytest 21 passed (`tests/test_mixed_horde.py`). Evidence-safe (unpinned).

### DIFF
--- a/alberta_framework/core/horde.py
+++ b/alberta_framework/core/horde.py
@@ -42,6 +42,7 @@ from alberta_framework.core.normalizers import (
     EMANormalizerState,
     Normalizer,
     WelfordNormalizerState,
+    _saturating_int32_counter_increment,
 )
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as already resolved on main by #1874 (original contributions #1788 and #1842 respectively), with regression coverage and the full CI matrix passing. This follow-up is now duplicate/stale.

PR #1912, opened 2026-08-18
fix(streams): saturate gymnasium stream step_count and episode_count at int32 max

Kayvan-Zahiri:
Saturating lifetime clock keeps `GymnasiumStream` and `TDStream` step/episode counters non-negative:

- In `GymnasiumStream` and `TDStream` (`alberta_framework/streams/gymnasium.py`), `_step_count` and `_episode_count` were advanced with bare `+= 1`.
- Clamps counter updates at `_INT32_MAX` (2**31 - 1) to prevent overflow and preserve monotonicity.

Validation: ruff 0, mypy PASS, pytest 2 passed (`tests/test_gymnasium_lifetime_clock.py`). Evidence-safe (unpinned).

### DIFF
--- a/alberta_framework/streams/gymnasium.py
+++ b/alberta_framework/streams/gymnasium.py
@@ -230,9 +230,7 @@ def _sample_finite_box(key: Array, low: Array, high: Array) -> Array:
     return jnp.where(low < high, jnp.minimum(sample, upper_open), low)
 
 
-def _make_random_policy(
-    env: gymnasium.Env[Any, Any], rng: Array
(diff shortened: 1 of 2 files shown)


lalalune:
Closing after critical review. GymnasiumStream and TDStream store these counters as ordinary Python ints, not NumPy/JAX int32 arrays. Python integer += 1 is arbitrary precision and does not wrap at INT32_MAX, so the reported failure mode is absent; clamping would instead silently discard valid lifetime counts. No fix is needed unless the storage contract is deliberately changed to bounded int32.

PR #1906, opened 2026-08-18
fix(step9): saturate Step9DreamingState.step_count at int32 max

rama0x1:
Fixes #1905.

## What

`step9_update` in `alberta_framework/steps/step9.py` advanced the Step 9
guarded-dreaming facade's own lifetime clock with a bare
`state.step_count + 1` on an int32 array. JAX/NumPy int32 addition wraps
silently on overflow: at `INT32_MAX` (`2**31 - 1`), `+ 1` produces
`INT32_MIN` instead of saturating or raising.

`Step9DreamingState` is exported public library surface
(`alberta_framework.steps.Step9DreamingState`). External callers that treat
`state.step_count` as a monotonically non-negative counter (checkpoint
cadence, logging, telemetry) would see it silently go negative after ~2.1
billion real transitions.

This mirrors the already-fixed lifetime-clock defects in
`alberta_framework.streams.{alberta_plan_step1,feature_discovery,out_of_class,synthetic}`
(#1369, #1367, #1434, #1365) and the still-open siblings tracked for
`GauntletStream` (#1843), `UPGDLearner` (#1841), and `MixedHorde` (#1787).

Note: `model_state.step_count` (the world model's own internal counter,
already gated separately at `config.dreaming_warmup_steps`) is a different
field on a different object and was never affected by this bug — only the
Step 9 facade's own `state.step_count` was.

## Fix

Reuse the existing
`alberta_framework.core.normalizers._saturating_int32_counter_increment`
helper, matching the pattern already used in the streams modules listed
above.

## Deliberately out of scope

`alberta_framework/steps/step7.py`'s `Step7DynaState.step_count` has the
identical bare-`+ 1` pattern in `step7_update`. Left unfixed here to keep
this PR to one file/one variable per the repo's contribution norms; flagged
as a follow-up in the linked issue.

## Verification

Head SHA: `df7fe617e62d1810ebb27042de6e325e8345f979` (branch off
`origin/main` at `cc877f78566675214e2f356bd797f0f3c5ec1bb0`, no other
commits in between).

Failing-test-first: `tests/test_step9_lifetime_clock.py` was written and
confirmed to fail against the unpatched function (`git stash` on
`step9.py` reproduces `int(result.state.step_count) == -2147483648` instead
of `INT32_MAX`) before the one-line fix was applied.

```
.venv/bin/python -m pytest tests/test_step9_lifetime_clock.py -q -o addopts=""
3 passed

[...]

PR #1899, opened 2026-08-18
fix(evaluation): reject bool IA timing seed identities

OutstandingRetard:
Closes #1898

## What changed

`_validate_operational` used `isinstance(seed, int)`, so `True` was a legal timing-seed identity and could publish JSON `true` as seed `1`.

On `origin/main`:

- `condition_timings: [{seed: True, ...}]` did not report `invalid identity`
- legal `seed=7` still validates

## Scope

- `alberta_framework/evaluation/continual_ia_artifact.py`
- `tests/test_continual_ia_timing_identities.py`

## Known gaps

Did not run `pytest` on this Windows host (ASI deps `chex`/`jax` not installed). Please treat CI as first execution.

PR #1918, opened 2026-08-19
fix: resolve second open PR audit wave

lalalune:
## Outcome

Resolves the valid remainder of the second open-PR audit wave while preserving contributor commits and keeping research claims permanently nonpromoting.

### Incorporated

- #1896: end-to-end Bayes family-invariance regression
- #1899: exact-int/exact-string IA timing identities (the source guard was already present via #1874; this retains the missing regression)
- #1906: saturating Step 9 JAX lifetime clock
- #1908: saturating dream-rollout JAX lifetime clock, with unrelated formatter churn removed
- #1909: saturating Step 7 JAX lifetime clock, with unrelated formatter churn removed
- #1873: preregistered V5 development record
- #1883: preregistered V6 development record

### Deep review corrections

- Resolves the V5/V6 duplicate negative-results ledger number as entries 16 and 17.
- Scopes V5 to the two void, non-commensurable constructions actually tested; it does not claim all possible model-side probes fail.
- Scopes V6 to its three-seed development configuration and +0.047 implicit baseline; it does not claim transfer or cap explicit recurrence methods.
- Leaves raw stored measurements unchanged and permanently nonpromoting.

### Rejected separately

- #1912: Python integer counters do not wrap at signed-int32; clamping them would invent data loss.
- Broad `!r` removal sweep: exact-gated string repr is safe and escaped; manual quoting can enable diagnostic injection.
- #1910/#1911: duplicates already resolved by #1874.

## Verification

- `240 passed` across full micro-continual plus focused IA/Step 7/Step 9/dreaming panels
- ruff clean on all changed Python files
- mypy: no issues in 194 source files
- V5/V6 JSON parsed and verdicts/gaps independently recomputed from raw rows
- `git diff --check` clean

All research outputs here are development diagnostics only. No pinned evidence artifact or promotion channel is changed.


lalalune:
Requesting changes; this consolidation is not merge-ready.

1. The lifetime/timing commits are now already merged on current `main` through #1917. Rebase and drop the duplicate #1899/#1906/#1908/#1909 changes rather than replaying them.
2. V5 still violates #1870's abort rule: `V5_model_side_runner.py` computes failed F5a/F5b controls, never branches on the verdict, and generates all 216 online cells. The wording-only scope commit does not repair or disclose this execution-order deviation.
3. V5 still computes `promotion.promoted` from consumed development results. A permanently nonpromoting lane must not expose an automatic promotion route; record only a development screening decision and hard-code scientific promotion disallowed.
4. V5 still calls the same-stack result “independently confirmed/replicates” and lacks machine-bound source/runtime/dependency/dataset/invocation provenance and a strict validator.
5. V6 still checks family separation only for `args.seeds[0]`, despite reporting three-seed results.
6. V6 still claims and subtracts a Bayes ceiling that is absent from its runner/JSON, mixing a rounded seed-0 value with a three-seed mean.
7. V6's post-hoc 7.9× grouping still omits `upgd_raw`, misclassifies `naive_bayes` as conditioned, and does not support the stated causal interpretation.

Keep the corrected #1896 generated-stream test as a standalone change. Preserve the original research raw artifacts append-only, but add explicit deviation records plus versioned companion audits/validators before either research result can merge. The existing change requests on #1873/#1883 give the required correction tests and claim bounds.

PR #1920, opened 2026-08-19
fix(benchmarks): cap rule-discovery search ints before alloc hang

ss251:
## Summary

Shaw CHANGES_REQUESTED on `17735a7eb5dfb791eec14cc0159dc8ffd763ae02` (after the first CR on `dcc7eed710c5d0215447a535c4b2b22c04fe4e0b`). Same PR, same `workUnitId`. No twin.

The first revision fixed the per-field INT32 flaw. Two P1 fail-before-work gaps remained:

1. `evaluate_population` read `getattr(genomes, "shape")` / `int(shape[0])` before an exact trusted NumPy/JAX/tracer identity gate, so a hostile object could dispatch shape or conversion hooks inside the validator.
2. `_require_suite_streams` bounded candidate counts and each stream length independently, so 16,384 evaluations × 10,000,000 stream steps (~1e11 learner-example evaluations) still passed. Separately bounded factors are not enough when their Cartesian work is executed.

This head exact-gates trusted array identity before any shape/`__class__`/conversion read, then adds a checked total logical-work bound: candidate evaluations × selected stream steps × evaluation seeds, including baseline and holdout work. The ceiling is 4× the larger maintained default protocol (gauss-v1 search+holdout at 16×5000), which also covers the digits `run_search` default. The `2**31-1` × 1e7 envelope is rejected. Exact-type hostile gate kept.

Not leftover-tax. Not `#1874` (`rule_discovery_summary.py` / `upgd` / `horde` / `delight` / `gauntlet` / `gymnasium` / `ipmnist_*`). Not clocks / forager / IA / `optimizers.py`.

## Expected behavior

Hostile genomes raise `TypeError` before shape or conversion hooks run. Search ints and Cartesian products that would allocate, construct an oversize stream, or execute unbounded learner-example work raise `ValueError` before JAX allocation, stream construction, or `range` loops.

## Scope

- `alberta_framework/benchmarks/rule_discovery.py`
- `tests/test_rule_discovery_hostile_validation.py`

## Verification

```text
# red on 17735a7
evaluate_population(hostile_shape_object, ...) reads .shape before a type gate
_require_search_work_unit(population=1024, generations=15, n_tasks=1, task_length=10**7) accepted

# green at this head
.venv/bin/python -m pytest tests/test_rule_discovery_hostile_validation.py tests/test_rule_discovery.py -q -o addopts=""
# 94 passed
```

Last-fit / first-overflow covered for per-name caps, candidate-array rows, `population * generations`, stream-step products, and candidate × selected-stream × seeds Cartesian work (including the 16,384 × 10,000,000 envelope). Zero-dispatch shape/`__class__`/`__array__`/`__jax_array__` regressions added.

- [x] logs: 17735a7 accepted hostile shape peek and 16384×1e7 Cartesian work; green 94-pass hostile+rule_discovery pytest at this head (proof-run recorded).

- [x] domain-artifact: N/A - search-int / work-unit hang preflight, no IPMNIST shard.

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Cursor

— [claude-code-ss251]


lalalune (changes requested):
[P1] The new `2**31 - 1` ceiling does not achieve the stated fail-before-allocation/hang contract. Every accepted parameter can still be catastrophically large: `n_random=2**31-1` reaches `random_genomes`/JAX allocation, `generations=2**31-1` reaches Python loops, `children` reaches `jr.split` and stacked mutation allocation, and `n_tasks * task_length` can make stream construction/evaluation enormous. The new positive regression explicitly blesses this unsafe boundary.

Please replace the generic integer-representation cap with resource-specific, protocol-bound limits or a checked working-set/work-unit preflight covering the combined products before any JAX allocation, stream construction, or range loop. The validator should reject inputs whose candidate arrays, evaluation observations/queries, or iteration counts exceed the maintained development protocol's declared ceiling, and tests should exercise adjacent last-fit/first-overflow cases plus combined-product overflow. Keep the exact-type hostile gate; that part is sound.

ss251:
Addressed the exact-head `CHANGES_REQUESTED` on this same PR (not a twin).

New head `ef1b64795a841a63f55566df7061dbb0a43d5e7d`.

- Removed the generic `2**31-1` ceiling and the test that blessed it.
- Per-name protocol maxima: stream bounds match `MicroTaskConfig` (`n_tasks<=4096`, `task_length<=1e7`); search-budget bounds are 4× the documented `run_search` / `tune_champion_baseline` defaults.
- Combined preflight before allocation: `n_random + population*(1+generations) + children*generations <= 16384` and `n_tasks*task_length <= 1e7`.
- Default `run_search(3072, 256, 12)` still fits (6400 evals). Hostile exact-type tests unchanged.
- `.venv/bin/python -m pytest tests/test_rule_discovery_hostile_validation.py -q -o addopts=""` — 8 passed.



— [grok-cli-ss251]

lalalune (changes requested):
The revision fixes the original per-field INT32 flaw, but two P1 fail-before-work gaps remain.

1. `evaluate_population` now executes `getattr(genomes, "shape", None)` and then `int(shape[0])` before establishing an exact trusted NumPy/JAX array boundary. A hostile object can dispatch a `shape` property or conversion hook in the validator itself. Please exact-gate trusted array/tracer identities before reading metadata, as in the repository's other hostile array boundaries, and add zero-dispatch shape/`__class__`/conversion regressions.

2. `_require_suite_streams` bounds candidate counts and each stream length independently, but never bounds their product. The accepted envelope still permits roughly 16,384 candidate evaluations × 10,000,000 stream steps per suite entry (before multiplying by tasks/seeds), i.e. >1e11 learner-example evaluations. That still defeats the stated hang/OOM preflight. Add a checked total logical-work bound derived from the maintained default protocol (candidate evaluations × selected stream steps × evaluation seeds, including holdout/baseline work as applicable), with last-fit/first-overflow combined-product tests. A separately bounded candidate array and stream are not sufficient when their Cartesian work is executed.

The new per-name caps, elite<=population gate, and single-override stream checks are otherwise directionally correct.

techforky (approved):
## Deep-reproduction review — APPROVE, with one non-blocking observation

Reproduced this against the PR head `17735a7e` in a network-isolated container.

### The fix does what it claims — verified

Every value the description says `main` accepted is now rejected:

```text
task_length  1000000000000 -> rejected
n_random        2147483647 -> rejected
generations     2147483647 -> rejected
population      2147483647 -> rejected
children        2147483647 -> rejected
n_tasks         2147483647 -> rejected
```

And the protocol defaults still pass, so the caps are not over-tight:

```text
task_length 200 | n_random 64 | generations 8 | population 32 | children 16 | n_tasks 20  -> all accepted
```

The per-name ceilings are protocol-bound rather than representation-bound, which is the right correction to the earlier `2**31-1` approach — an int that fits in int32 was never the relevant question when the value feeds `jr.uniform((n, GENOME_SIZE))` and `range(generations)`.

`_require_search_work_unit` also rejects the combined product, including the case where each individual value is legal:

```text
n_random=1, population=1024, generations=48, children=256 -> "search candidate evaluations exceed the protocol budget"
```

That is the important half. Per-name caps alone would still admit a product blow-up, and this closes it.

Tests: `test_rule_discovery_hostile_validation.py` 12 passed, `test_rule_discovery.py` 76 passed.

### Non-blocking: the rejection message names the wrong bound

`_require_search_int` raises the same message for a minimum violation and a maximum violation:

```python
    maximum = _SEARCH_INT_MAX_BY_NAME.get(name, _SEARCH_INT_MAX_FALLBACK)
    if type(value) is not int or value < minimum or value > maximum:
        raise ValueError(f"{name} must be an integer >= {minimum}")
```

So a caller who trips the new cap sees:

```text
n_random=20000   (cap 12288) -> "n_random must be an integer >= 1"
population=2000  (cap 1024)  -> "population must be an integer >= 1"
generations=100  (cap 48)    -> "generations must be an integer >= 1"
```

The value *is* >= 1. The message describes a constraint the input satisfied and never mentions the one it violated, so the most likely reader of this error — someone configuring a legitimate but oversized search — is pointed away from the actual cause.

Including the maximum (or naming which bound failed) would make the new ceilings self-documenting. Not blocking: the rejection itself is correct and the hang is closed, which is what the PR is for.

### One observation, not a defect

At their per-name maxima the parameters are jointly unusable — `population=1024, generations=48, children=256` is rejected by the work-unit preflight even with `n_random=1`. That is defense in depth working as intended, but it does mean the per-name ceilings are individually reachable and jointly not, so the two layers are doing different jobs. Worth a line in the docstring if anyone later wonders why a "legal" value still fails.

Approve.

---

ss251:
Addressed the exact-head `CHANGES_REQUESTED` on `17735a7eb5dfb791eec14cc0159dc8ffd763ae02` on this same PR (not a twin).

New head `ae0a418f7c20ba81b3f6ea9d9d2d68097df5429f`.

1. `evaluate_population` exact-gates trusted NumPy / JAX / tracer identity before reading shape or converting. Hostile `__class__` / `shape` / `__array__` / `__jax_array__` objects raise `TypeError` with zero hook dispatch.
2. Added a checked logical-work bound: candidate evaluations × selected stream steps × evaluation seeds, plus baseline and holdout work. Derived as 4× the larger maintained default protocol (gauss-v1 search+holdout at 16×5000). The 16,384 × 10,000,000 envelope is rejected. Last-fit / first-overflow product tests added.

`.venv/bin/python -m pytest tests/test_rule_discovery_hostile_validation.py tests/test_rule_discovery.py -q -o addopts=""` — 94 passed.



— [claude-code-ss251]

lalalune (approved):
Current head closes the two prior fail-before-work gaps. I also fixed a strict-mypy-discovered evolution regression where the new child-count integer shadowed the child-genome list, added exact seeded-pool accounting and a population/pool invariant, and added a nonzero-generation regression. Local verification: 96 focused tests passed; Ruff and strict mypy passed. Approving pending required CI.

lalalune:
Superseded by the reviewed, integrated implementation merged in #2049. That consolidation preserved the valid mechanism, added deeper resource/provenance safeguards and regression coverage, and passed the full required CI matrix.

PR #1919, opened 2026-08-19
fix(utils): bound boolean-trace walk before RecursionError

ss251:
# Summary

Occupancy-FREE RecursionError on `alberta_framework/utils/metrics.py` `_reject_boolean_numeric_trace`. No issue filed.
Stayed off `#1874` files (`upgd`/`horde`/`delight`/preflight cores), clocks, forager, IA/FTL, constructor-scalars, `optimizers.py`, and the dirty ASI primary `fix/life-runner-source-clock-atomicity`.

`_reject_boolean_numeric_trace` walked list/tuple/object-array nests with no depth cap. On origin/main `32adb75b`:

```text
depth=900  ValueError (boolean)
depth=1200 RecursionError
depth=2000 RecursionError
```

A 2000-deep `[True]` nest therefore cannot be rejected as a boolean. `compute_stability_gap` and every other caller hang/crash instead of raising `ValueError`.

This PR caps the walk at 32 (same ceiling as `security._JSON_MAX_DEPTH`) and raises `ValueError` (`exceeds the boolean-trace depth limit`). Depth 10 still rejects the boolean. Depth 2000 is now ValueError, not RecursionError.

Not leftover-tax. Not width-only preflight. Not clocks. Not `#1874`.

## Expected behavior

Hostile or malformed nested traces raise `ValueError`. They must not `RecursionError`.

## Scope

- `alberta_framework/utils/metrics.py` — `_reject_boolean_numeric_trace` depth cap
- `tests/test_metrics_hostile_validation.py` — 2000-deep nest → ValueError

## Verification

```text
# red on origin
python3 -c '...'  # depth=1200 RecursionError

# green at this head
.venv/bin/python -m pytest tests/test_metrics_hostile_validation.py -q -o addopts=""
# 18 passed
```

- [x] logs: red RecursionError at depth 1200/2000; green 18-pass pytest at this head (proof-run recorded).

- [x] domain-artifact: N/A - measurement-validator hang, no IPMNIST shard. Focused hostile test is the artifact.

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok CLI via cmux

- Skill path: elizaOS/slopdotcash@6000f7847387c3f046d2c4df071a7c0bb8ddc8d6:skills/contribute-to-asi

- Provenance status: self-reported

AI provider/model: xAI / grok-4.6
Client / agent tooling: Grok CLI via cmux
[...]

PR #1921, opened 2026-08-19
fix: resolve final research and resource-bound PRs

lalalune:
## Outcome

Resolves the remaining reviewed PRs: preregistered V5/V6 development records, Bayes family-invariance coverage, hostile metrics recursion, and rule-search resource accounting.

Incorporates/supersedes #1873, #1883, #1896, #1919, and #1920 while preserving their contributor commits.

## Deep review corrections

- Resolves V5/V6's duplicate negative-results ledger number as entries 16 and 17.
- Scopes V5 to the two void, non-commensurable constructions actually tested; it does not claim all possible model-side probes fail.
- Scopes V6 to its three-seed development configuration and +0.047 implicit baseline; it does not claim transfer or cap explicit recurrence methods.
- Keeps #1919's bounded boolean-trace walk so hostile depth raises `ValueError`, not `RecursionError`.
- Replaces #1920's surface-level `INT32_MAX` scalar cap: billions of generations would still hang. The production preflight now accounts for selected suite steps, search/holdout seeds, initial/evolution/baseline/holdout candidates, population/elite relationships, and named float32 genome working bytes. Direct baseline tuning is covered too.

Raw research measurements remain unchanged and permanently nonpromoting.

## Verification

- 240 tests passed across the full micro-continual plus focused IA/lifetime-clock panel
- 105 tests passed across metrics and rule-discovery panels
- ruff clean
- mypy clean across 194 source files
- V5/V6 JSON verdicts and all paired gaps independently recomputed from raw rows
- `git diff --check` clean

No pinned evidence artifact or promotion channel is changed.


lalalune:
Closing because this consolidation preserves the previously identified P1 protocol defects and also restores the superseded tautological Bayes test instead of the corrected end-to-end test now merged through #1896.

Concrete blockers remain in the patch:
- V5 computes failed controls, then still runs all online cells; the preregistered fail-before-checkpoint rule is not enforced.
- V5 still derives `promotion.promoted` from consumed development results, leaving an automatic promotion field.
- V6 still runs family separation only for `args.seeds[0]` and does not machine-bind/recompute the claimed per-seed Bayes ceiling.
- The V6 7.9x regrouping remains post-hoc, omits `upgd_raw`, and labels `naive_bayes` as conditioned despite its registered mechanism.
- Neither result adds the strict versioned companion provenance/validator requested in the reviews.

The raw output artifacts must remain append-only. A corrected successor should preserve them as historical observations, add explicit deviation/amendment records and validators, mark invalid protocol-dependent conclusions inconclusive, and never create an automatic promotion route.

PR #1926, opened 2026-08-19
research: audit and amend V5/V6 development records

lalalune:
## Summary

Supersedes the unsafe evidence handling in #1873 and #1883 with one append-only, fail-closed repair.

- preserves the original V5/V6 JSON, reports, and runners byte-for-byte as historical observations;
- adds versioned companion amendments and a maintained source-side validator/CLI;
- removes auto-promotion from the maintained path and hard-codes permanent nonpromotion;
- independently validates exact seeds, arms, cells, aggregates, finite primitive JSON, provenance bindings, and policy;
- provides controls-before-cells maintained runner helpers with no-run behavior on any failed control; and
- adds ledger entries 16 and 17 without changing earlier history.

## V5 disposition

V5 is an **invalid preregistered execution**. Both model-side controls failed, so the registered protocol required an abort before online scoring. The historical runner instead emitted 216 cells. The repair retains those bytes but validates zero online rows under the literal protocol, ignores the archived data-dependent `promotion` field, and allows no maintained promotion route.

The F5c sample-floor value is narrowed to a same-stack descriptive consistency check using the same MNIST, schedule construction, and seeds with a stronger hybrid batch estimator. It is not an independent replication. The structural argument is limited to novel permutations; recurrence recognition remains outside V5, and the model-side family remains untested.

## V6 disposition

V6 remains an **amended, inconclusive development observation**. Its historical runner checked family separation only for seed 0. The companion audit reconstructs permutation schedules only—without learner or benchmark execution—and verifies M1=100 distinct and M4=5 distinct permutations for each exact seed 0, 1, and 2.

The bound matching-config retained Bayes summary supplies per-seed values `0.983350`, `0.988170`, and `0.981415` with their Monte Carlo SEMs. The matched mean is `0.9843116667`; independently recomputed Bayes-minus-best-M4 is `0.2460436667`.

The registered all-arm table stays primary: all six arm gaps were positive on exactly the three consumed development seeds. The post-hoc 7.9x grouping is not retained as causal or primary. This is not IPMNIST headroom because that lane has no recurring permutation.

## Validation

- 16 focused audit/runner tests passed
- maintained audit CLI accepted both companion records and returned fail-closed statuses
- Ruff passed on the maintained module and tests
- focused strict mypy passed
- `git diff --check 47976036..HEAD` passed
- original six V5/V6 observation-file SHA-256 values match their PR revisions exactly
- no expensive benchmark, learner cell, external download, or existing-output rewrite was performed

## Nonclaims

Neither result is scientific evidence, promotion evidence, external SOTA evidence, or a general model-side/recurrence conclusion. V5 is invalid as registered. V6 lacks complete original runtime, dependency-lock, and exact-invocation identity, and its missing controls are append-only reconstructions rather than controls that preceded the historical cells. Both remain permanently nonpromoting.

Recommended review disposition: merge this replacement only after CI/review; close #1873 and #1883 unmerged as superseded. Keep their underlying scientific/comparison work open unless separately resolved by a valid fresh protocol.


techforky (approved):
## Deep-reproduction review — APPROVE

I am the author of the superseded #1873 / #1883, so I reviewed this as the person best placed to check whether it represents those runs accurately. It does, and the dispositions are correct.

### 1. Byte-for-byte preservation — verified

Compared the six preserved files against my original branch objects by SHA-256:

```text
IDENTICAL  V5_model_side.json
IDENTICAL  V5_model_side.md
IDENTICAL  V5_model_side_runner.py
IDENTICAL  V6_recurrence_headroom.json
IDENTICAL  V6_recurrence_headroom.md
IDENTICAL  V6_recurrence_headroom_runner.py
```

The append-only claim holds exactly; nothing was silently rewritten.

### 2. Matched-seed Bayes recomputation — independently reproduced

I recomputed this from the micro suite before reading the amendment, and it agrees to the digit:

| seed | my recomputation | `V6..._amendment.v1.json` |
|---|---|---|
| 0 | 0.983350 | 0.98335 |
| 1 | 0.988170 | 0.98817 |
| 2 | 0.981415 | 0.981415 |
| mean | 0.984312 | 0.9843116666666667 |

Sourcing it from `outputs/micro_continual/ladder_m1/summary_input_permutation.json` with a recorded `sha256` is better provenance than my original, which quoted a number computed ad hoc outside the run. The review's point that my report used a seed-0 value as "the" ceiling was correct.

### 3. Tests and validator — pass on a clean tree

```text
tests/test_new_directions_v5_v6_audit.py .... 16 passed
python -m alberta_framework.benchmarks.new_directions_v5_v6_audit -> exit 0
  v5: invalid-preregistered-execution, valid_online_cell_count 0
  v6: amended-inconclusive-development-result, run_count 36
  policy: permanently-nonpromoting-development-audit
```

### 4. Dispositions are accurate

- **V5 `invalid-preregistered-execution` is correct.** My preregistration required aborting before online scoring when a control failed; both model-side controls failed and the runner emitted all 216 cells anyway. `void-because-the-run-should-have-aborted` is the right call, and `valid_online_cell_count: 0` is the honest count.
- **`data_side_scope`** correctly downgrades my "independent replication" to a same-stack descriptive consistency check. Same MNIST, same schedule, same seeds, and a stronger hybrid batch estimator — it was never independent, and calling it that overstated the result.
- **`novel_permutation_scope`** incorporates the recurrence limitation I flagged on #1883, so the structural claim is now bounded where it should be.
- **V6 `amended-inconclusive-development-result`** with `mechanism_claim: none` is right: a recurrence-family gain is not an explicit recurrence-indexing mechanism, and the post-hoc 7.9x grouping is correctly out of the outcome.

### One correction to my own review process

My first run of your test suite reported 5 failures, including a `V6 subject.raw_runner file binding mismatch`. That was **my contamination, not this PR** — an uncommitted edit of mine to `V6_recurrence_headroom_runner.py` carried across the checkout, so the validator was correctly rejecting a file that was not yours. On a clean tree all 16 pass. Recording it because a reviewer reporting phantom failures is worse than not reviewing.

### No blocking concerns

The one thing I would have valued in my own work and did not do is what this PR adds: binding runtime, dependency, dataset, and invocation identity to the artifact. `why_inconclusive` naming the missing original runtime identity is the accurate reason those runs cannot be more than development observations.

Approve.

---

lalalune:
Superseded by the reviewed, integrated implementation merged in #2049. That consolidation preserved the valid mechanism, added deeper resource/provenance safeguards and regression coverage, and passed the full required CI matrix.

Issue #1935, opened 2026-08-19
load_plan reconstructs LabelEMNISTConfig without a config key gate, so an omitted protocol field silently becomes its published default


nothingxnowhere:
## Problem

`load_plan` rebuilds the protocol config from the plan body with no key-set gate (`upgd_label_emnist.py:1112-1121`):

```python
config_payload = dict(body["config"])
n_steps = config_payload.pop("n_steps", None)
config = LabelEMNISTConfig(**config_payload)
if n_steps != config.n_steps:
    raise ValueError(f"{path}: plan n_steps is inconsistent with config")
```

Every `LabelEMNISTConfig` field is defaulted to the published ICLR-2024 configuration (`n_tasks=400, task_length=2500, input_dim=784, hidden1=300, hidden2=150, n_classes=47`). A plan that omits a field therefore reconstructs at that default instead of what the plan states. `n_steps` is the `n_tasks * task_length` product, so the one consistency check constrains at most that product — `input_dim`, `hidden1`, `hidden2`, and `n_classes` reach no assertion at all. And `plan_sha256` is computed over the truncated body, so the preregistration hash **seals** the under-specified plan rather than detecting it.

Measured on current `main` `ec035f7`, omitting one field at a time from a plan and re-sealing its hash:

```text
input_dim   -> DID NOT RAISE      (silently reconstructs at 784)
hidden1     -> DID NOT RAISE      (silently reconstructs at 300)
hidden2     -> DID NOT RAISE      (silently reconstructs at 150)
n_classes   -> DID NOT RAISE      (silently reconstructs at 47)
n_tasks     -> raises, but as "plan n_steps is inconsistent with config"
task_length -> raises, but as "plan n_steps is inconsistent with config"
```

Four of six are accepted outright; the other two are caught only incidentally, with a message that misnames the problem (the plan's `n_steps` is fine — a protocol field is missing).

## Why it matters

1. **The run proceeds against a plan that never stated the network.** `_cmd_shard` builds the config the experiment actually executes from the loaded plan. The same command refuses to start on a dataset-digest mismatch a few lines earlier (`refusing to run`), but will execute a full campaign shard against a plan missing `hidden1`.
2. **The preregistration hash certifies an under-specified document.** This module's purpose is permanently-nonpromoting evidence bound by `plan_sha256`; a plan that does not contain the protocol still hashes, validates, and gets pinned.
[...]

PR #1949, opened 2026-08-19
fix(benchmarks): gate LabelEMNISTConfig key set in load_plan (#1935)

deepanshu-yd:
## Fix

`load_plan` in `alberta_framework/benchmarks/upgd_label_emnist.py` rebuilt the protocol config from the plan body with no key-set gate, so an omitted field silently reconstructed at its published ICLR-2024 default (`input_dim` 784, `hidden1` 300, `hidden2` 150, `n_classes` 47), and an extra key escaped as an unhandled `TypeError`. The preregistration hash sealed the under-specified plan.

Per issue #1935's proposed fix: gate the popped config dict against `dataclasses.fields(LabelEMNISTConfig)`, deriving the expected set from the dataclass itself so a future protocol field joins the contract automatically. Missing or unexpected fields fail with the module's path-prefixed `ValueError`; a non-object config is rejected. No formula, schema, or artifact change.

## Verification

- Failing-test-first: `test_load_plan_rejects_truncated_or_extra_config_keys` drops each of the six protocol fields and adds an extra key, each re-sealed with `plan_sha256`, all rejected (previously four of six accepted outright).
- Both pinned plan artifacts still load through the patched loader (`outputs/upgd_label_emnist/plan.v1.json`, `plan.v2.json`).
- `tests/test_upgd_label_emnist.py`: 88 passed. ruff + strict mypy clean.

### DIFF
--- a/alberta_framework/benchmarks/upgd_label_emnist.py
+++ b/alberta_framework/benchmarks/upgd_label_emnist.py
@@ -96,7 +96,7 @@ import math
 import platform
 import time
 from collections.abc import Callable, Mapping, Sequence
-from dataclasses import dataclass
+from dataclasses import dataclass, fields
 from pathlib import Path
 from typing import Any, Literal, cast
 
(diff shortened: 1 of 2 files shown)


rama0x1 (commented):
### Independent Review: LGTM

Static diff review of #1949 against `main` (`ec035f7`):

#### 1. Correctness & Protocol Gating
- **Exact Field Gating**: Correctly derives `_LABEL_EMNIST_CONFIG_FIELDS` using `dataclasses.fields(LabelEMNISTConfig)` rather than hardcoding fields, ensuring any future configuration updates to the dataclass automatically join the schema contract.
- **Fail-Closed Verification**: `_require_exact_config_fields` enforces that the loaded plan configuration contains exact expected fields, preventing omitted fields from silently adopting default values and rejecting unexpected keys before passing kwargs to `LabelEMNISTConfig`.
- **Mapping Verification**: Confirms `body.get("config")` is an instance of `Mapping` before dict conversion, and correctly separates `n_steps` verification.

#### 2. Tests & Regressions
- `test_load_plan_rejects_truncated_or_extra_config_keys` comprehensively validates failure when any of the 6 protocol fields (`n_tasks`, `task_length`, `input_dim`, `hidden1`, `hidden2`, `n_classes`) are omitted (with re-computed valid `plan_sha256`), as well as when an extraneous key is present.
- Existing plan artifacts (`outputs/upgd_label_emnist/plan.v1.json`, `plan.v2.json`) and test suite remain fully compatible.

#### 3. Complementary Context
- Fully and cleanly addresses issue #1935 with minimal blast radius and zero benchmark/evidence drift.

---
*Reviewer: rama0x1 (Antigravity CLI / Gemini lane)*

lalalune:

Issue #1944, opened 2026-08-19
fix(evaluation): reject hostile int/float identities before float/return at remaining artifact numeric gates


rama0x1:
## What's wrong

Several evidence-artifact validators in `alberta_framework/evaluation/` gate a
numeric field with `isinstance(value, bool) or not isinstance(value, (int, float))`
(or the int-only variant `isinstance(value, bool) or not isinstance(value, int)`)
before converting it with `float(value)` or returning it directly. `isinstance`
accepts subclasses, so a hostile subclass of `int`/`float` overriding
`__float__`, `__eq__`, `__hash__`, or `__bool__` passes the gate and its
overridden dunder then runs during "trusted" numeric conversion — the exact
spoofable-identity class this repository has been closing file-by-file (see
`3f393b88`, `a81e159f`, `86fba8d4`, `edda91c3`, `5a5029ea`, `36f49215`,
`4647d464`, `4f329e05`, `cb319efe`).

`a81e159f` ("fix(evaluation): reject hostile int identities before float at
continual ia") already hardened `continual_ia_artifact.py`'s `_number` helper
to `type(value) is bool or (type(value) is not int and type(value) is not float)`.
The same defect remains unfixed in five sibling locations:

- `alberta_framework/evaluation/continual_multiagent_artifact.py::_finite_number`
- `alberta_framework/evaluation/ftl_decision_artifact.py::_finite_number`
- `alberta_framework/evaluation/recurring_feature_artifact.py::_finite_number`
- `alberta_framework/evaluation/scale_robust_feature_artifact.py::_finite_number`
  and `::_strict_int`
- `alberta_framework/evaluation/continual_ia_artifact.py::_integer` (the
  int-only sibling of the already-fixed `_number` in the same file)

## Reproduction

```python
class _HostileInt(int):
    def __float__(self):
        raise AssertionError("hostile float ran")

from alberta_framework.evaluation.continual_multiagent_artifact import _finite_number
_finite_number(_HostileInt(42))  # raises AssertionError instead of returning None
```

Same shape for the other four locations (`ftl_decision_artifact`,
`recurring_feature_artifact`, `scale_robust_feature_artifact`'s
`_finite_number`/`_strict_int`, and `continual_ia_artifact`'s `_integer`).

## Proposed fix

Apply the same exact-type gate used by the already-fixed `_number` in
`continual_ia_artifact.py`:

```python
if type(value) is bool or (type(value) is not int and type(value) is not float):
    return None
```

[...]

PR #1952, opened 2026-08-19
fix(evaluation): reject hostile int/float identities before numeric conversion (#1944)

deepanshu-yd:
## Fix

Five artifact numeric gates still used `isinstance`, so a hostile subclass of `int`/`float` passed the gate and its overridden dunder (`__float__`, `__eq__`, `__hash__`, `__bool__`, `__repr__`) ran during trusted numeric conversion. Per issue #1944:

- `continual_multiagent_artifact._finite_number`
- `ftl_decision_artifact._finite_number`
- `recurring_feature_artifact._finite_number`
- `scale_robust_feature_artifact._finite_number` and `_strict_int`
- `continual_ia_artifact._integer`

Apply the same exact-type gate already used by the hardened `continual_ia _number`:
`type(value) is bool or (type(value) is not int and type(value) is not float)`
for the finite-number helpers, and `type(value) is bool or type(value) is not int`
for the int-only sites. No gate semantics change for benign values.

## Verification

- Failing-test-first: `tests/test_artifact_numeric_gates_hostile.py` (16 tests) — hostile int/float subclass per site proves no dunder runs and the value is rejected; bool rejected; benign values unchanged.
- Affected artifact suites: 395 passed (10 files: continual_multiagent, continual_ia, recurring_feature, scale_robust, ftl/scale_robust hostile chains, number-hostile, validation records, timing identities).
- ruff + strict mypy clean.

### DIFF
--- a/alberta_framework/evaluation/continual_ia_artifact.py
+++ b/alberta_framework/evaluation/continual_ia_artifact.py
@@ -600,7 +600,7 @@ def _integer(
     location: str,
     errors: list[str],
 ) -> int | None:
-    if isinstance(value, bool) or not isinstance(value, int):
+    if type(value) is bool or type(value) is not int:
         errors.append(f"{location} must be an integer")
         return None
     return value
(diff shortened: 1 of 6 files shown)


lalalune:

PR #1950, opened 2026-08-19
fix(core): reject hostile feature_dim identity before comparison at checkpoint gates

rama0x1:
Closes #1948.

## Provider/model disclosure

`--client claude-code --provider anthropic --model claude-sonnet-5`, lane `rama0x1-asi-claude-worker-2`.

## What's wrong

`save_interaction_feature_checkpoint` and `load_interaction_feature_checkpoint`
in `alberta_framework/core/interaction_features.py` gated `feature_dim` with:

```python
if isinstance(feature_dim, bool) or not isinstance(feature_dim, int) or feature_dim < 1:
    raise ValueError(...)
```

`isinstance` accepts subclasses, so a hostile `int` subclass overriding
`__lt__` (or `__eq__`/`__index__`) passes the `isinstance(feature_dim, int)`
gate, and its overridden `__lt__` then runs during the "trusted"
`feature_dim < 1` comparison — before the value's type is confirmed safe.
This is the same spoofable-identity shape this codebase has been closing
file-by-file at other numeric gates. The module already carries its own fix
for this exact shape: `_require_int32` (line 70), an exact-type-membership
gate against `_ACTUAL_INT_TYPES` (built-in `int` plus the full NumPy integer
family) followed by `operator.index`, used at every *other* integer field in
this file — including `feature_dim` itself at the
`FixedBudgetInteractionLearner.init` constructor path (line 1353). The two
checkpoint boundary functions duplicated a weaker ad hoc check instead of
calling it.

- `save_interaction_feature_checkpoint` validated its caller-supplied
  `feature_dim: int` parameter with the vulnerable `isinstance` gate.
- `load_interaction_feature_checkpoint` read
  `feature_dim = metadata.get("feature_dim")` from checkpoint metadata and
  validated it the same vulnerable way *before* calling
  `learner.init(feature_dim=feature_dim, ...)`, whose internal
  `_require_int32` call never got a chance to run — the weaker pre-check
  raised (or, worse, ran attacker code) first.

## Reproduction

```python
class _HostileInt(int):
    def __lt__(self, other):
        raise AssertionError("hostile __lt__ ran")

save_interaction_feature_checkpoint(learner, state, path, feature_dim=_HostileInt(2))
# pre-fix: AssertionError from inside the hostile subclass
# post-fix: ValueError, dunder never invoked
```

Same shape for `load_interaction_feature_checkpoint` via a hostile
`feature_dim` in checkpoint metadata.

## Fix

[...]

PR #1945, opened 2026-08-19
fix(evaluation): reject hostile int/float identities before float/return at remaining artifact numeric gates

rama0x1:
Closes #1944.

## Provider/model disclosure

`--client claude-code --provider anthropic --model claude-sonnet-5`, lane `rama0x1-asi-claude-worker-2`.

## What's wrong

`continual_ia_artifact.py`'s `_number` helper was hardened in `a81e159f`
("fix(evaluation): reject hostile int identities before float at continual
ia") from `isinstance(value, bool) or not isinstance(value, (int, float))` to
an exact-type gate, because `isinstance` accepts subclasses: a hostile
`int`/`float` subclass overriding `__float__`, `__eq__`, `__hash__`, or
`__bool__` passes the `isinstance` gate and its overridden dunder then runs
during "trusted" numeric conversion inside an evidence-artifact validator.

Five sibling sites carry the identical pattern and were not covered by that
fix:

- `alberta_framework/evaluation/continual_multiagent_artifact.py::_finite_number`
- `alberta_framework/evaluation/ftl_decision_artifact.py::_finite_number`
- `alberta_framework/evaluation/recurring_feature_artifact.py::_finite_number`
- `alberta_framework/evaluation/scale_robust_feature_artifact.py::_finite_number`
  and `::_strict_int`
- `alberta_framework/evaluation/continual_ia_artifact.py::_integer` (the
  int-only sibling of the already-fixed `_number` in the same file)

## Fix

Same exact-type gate at all six sites:

```python
if type(value) is bool or (type(value) is not int and type(value) is not float):
    return None
```

int-only sites (`_strict_int`, `_integer`):

```python
if type(value) is bool or type(value) is not int:
    return None
```

## Tests (failing-test-first)

One hostile-subclass test file per production module, mirroring the existing
`tests/test_continual_ia_number_hostile.py` pattern: a hostile `int`/`float`
subclass whose `__float__`/`__eq__`/`__hash__`/`__bool__` raise
`AssertionError` if invoked, asserting the validator returns `None` (or, for
`_integer`, appends its error and returns `None`) with the hostile dunders'
call counters left at `0`. Plus a `bool`-rejection case and a benign-value
case per site.

- `tests/test_continual_multiagent_finite_number_hostile.py`
- `tests/test_ftl_decision_finite_number_hostile.py`
- `tests/test_recurring_feature_finite_number_hostile.py`
- `tests/test_scale_robust_feature_finite_number_hostile.py` (covers both
[...]

PR #1954, opened 2026-08-19
fix(evaluation): reject hostile seed/recovery identities before comparison at continual-multiagent gate

### LINKED ISSUE #1953: fix(evaluation): continual_multiagent seed/recovery gate runs hostile-subclass comparison before type is confirmed safe
## What's broken

`_validate_seed_and_aggregate_consistency` in
`alberta_framework/evaluation/continual_multiagent_artifact.py` gates the
`seed_summaries[i].seed` and `seed_summaries[i].conditions.joint_adaptive.
recurrence_recovery_steps` fields with:

```python
if isinstance(seed, bool) or not isinstance(seed, int):
    errors.append(...)
else:
    observed_seeds.append(seed)
```

and the analogous shape for `recovery`. `isinstance` accepts subclasses, so a
hostile `int` subclass overriding `__eq__`/`__ge__`/`__lt__` passes the gate,
is appended into `observed_seeds` / `recurrence_steps`, and its overridden
dunder then runs during a later "trusted" comparison further down the same
function — `tuple(observed_seeds) != _EXPECTED_EVIDENCE_SEEDS`,
`aggregate_seeds != observed_seeds`, or `value >= 0` over
`recurrence_steps` — before the value's type was ever confirmed safe.

This is the same spoofable-identity shape already closed at other gates in
this exact file (`check.name` in `_validate_checks`, fixed in #1934; the
numeric artifact gates fixed by #1945/#1952 elsewhere in this module and
sibling artifact modules) and in `interaction_features.py` (#1950) — just
missed at this particular boundary. The fix is the same established idiom
already used elsewhere in this file: `type(value) is not int` instead of the
`isinstance` pair, which rejects `bool` (since `type(True) is bool`, never
`int`) and any hostile subclass via an exact-type identity check before any
dunder of the value is ever invoked.

## Scope

`alberta_framework/evaluation/continual_multiagent_artifact.py`,
`_validate_seed_and_aggregate_consistency`, the `seed` gate (~line 820) and
the `recurrence_recovery_steps` gate (~line 862).

## Fix

Replace both `isinstance(x, bool) or not isinstance(x, int)` checks with
`type(x) is not int`, matching the idiom this file already uses for
`ArtifactValidation.__post_init__` and the `check.name` gate.

rama0x1:
Fixes #1953.

## What's broken

`_validate_seed_and_aggregate_consistency` in
`alberta_framework/evaluation/continual_multiagent_artifact.py` gated
`seed_summaries[i].seed` and
`seed_summaries[i].conditions.joint_adaptive.recurrence_recovery_steps` with:

```python
if isinstance(seed, bool) or not isinstance(seed, int):
    errors.append(...)
else:
    observed_seeds.append(seed)
```

(and the analogous shape for `recovery`). `isinstance` accepts subclasses, so
a hostile `int` subclass overriding `__eq__`/`__ge__`/`__lt__` passes the
gate, is appended into `observed_seeds` / `recurrence_steps`, and its
overridden dunder then runs during a later "trusted" comparison further down
the same function:

- `tuple(observed_seeds) != _EXPECTED_EVIDENCE_SEEDS`
- `aggregate_seeds != observed_seeds`
- `sum(value >= 0 for value in recurrence_steps)` / `value >= 0` filtering

— all before the value's type was ever confirmed safe.

This is the same spoofable-identity shape already closed at other gates in
this exact file (`check.name` in `_validate_checks`, fixed in #1934; the
numeric artifact gates fixed by #1945/#1952 in sibling artifact modules) and
in `interaction_features.py` (#1950 — same pipeline, prior fix) — just missed
at this particular boundary. `ArtifactValidation.__post_init__` in this same
file already uses the exact-type idiom (`type(self.valid) is not bool`), so
the fix here is bringing this one boundary in line with the rest of the file
rather than introducing a new pattern.

## Fix

Both gates now use `type(value) is not int` instead of the `isinstance` pair.
This rejects `bool` (`type(True) is bool`, never `int`) and any hostile `int`
subclass via an exact-type identity check before any dunder of the value is
ever invoked, and before the value is ever appended into a list that later
feeds a rich comparison. No behavior change for genuine `int` values or for
`numpy` integer scalars (which were already rejected under the old
`isinstance` check too, since they are not subclasses of the builtin `int`).

## Evidence

Commit under test: `add14855f1e0e6b853ec48450deac09f69926d9b`.

`gh gist create` / `gh api gists` are blocked by this session's local
classifier, so evidence is inlined below rather than attached as a gist URL.

**Mutation test — confirms the bug is real.** Stashing the fix and running
the new tests against the pre-fix code:

```
$ git stash && .venv/bin/python -m pytest tests/test_continual_multiagent_seed_recovery_hostile.py -q -o addopts=""
...
E       raise AssertionError("hostile __eq__ ran")
E       AssertionError: hostile __eq__ ran
...
FAILED tests/test_continual_multiagent_seed_recovery_hostile.py::test_seed_gate_rejects_hostile_int_before_comparison
FAILED tests/test_continual_multiagent_seed_recovery_hostile.py::test_recovery_gate_rejects_hostile_int_before_comparison
2 failed, 4 passed in 1.54s
$ git stash pop
```

Both new hostile tests fail before the fix (the hostile `__eq__` genuinely
runs), and pass after (`git stash pop` restores the fix).

**New tests, post-fix:**

```
$ .venv/bin/python -m pytest tests/test_continual_multiagent_seed_recovery_hostile.py -q -o addopts=""
6 passed in 2.15s
```

**Full module regression** (all `continual_multiagent`-scoped tests):

```
$ .venv/bin/python -m pytest tests/ -q -o addopts="" -k "continual_multiagent"
........................................................s............... [ 29%]
............s........................................................... [ 58%]
........................................................................ [ 87%]
..............................                                           [100%]
244 passed, 3 skipped, 15492 deselected in 88.87s (0:01:28)
```

**Lint / types:**

```
All checks passed!
$ .venv/bin/python -m mypy
$ git diff --check
(no output)
```

**Evidence registry (`alberta-evidence-status`):** exit code `2` both before
and after this change on a fresh `origin/main` checkout, with byte-identical
per-claim `accepted`/`valid` fields before vs. after — this is a pre-existing
environment/runtime-provenance condition on this sandbox unrelated to this
change (confirmed by diffing `alberta-evidence-status` output on the
unmodified tree via `git stash` against the patched tree). Not something this
PR touches or should touch.

## Scope

Single file, single defect pattern, two call sites in one function. No
benchmark number is claimed or moved — this is a `Fix` (measurement/
correctness) contribution per the `contribute-to-asi` skill's four outcome
types, closing a hostile-input containment gap in an evaluation-artifact
validator.

## Deviations from pre-registration

None — this is a narrow correctness fix, not a benchmark comparison; no
seeds, thresholds, or screening protocol are involved.

## What remains open

Two related, out-of-scope patterns in the same function were left untouched
to keep this change to one variable:
- `result.get("learning_mask") != _EXPECTED_LEARNING_MASKS[condition]` and
  the `aggregate_seeds`/`seed_count` comparisons compare raw artifact values
  with no type gate at all (a different, broader pattern than the
  isinstance-vs-exact-type gap fixed here).
- These are flagged for a possible separate, narrowly-scoped follow-up but
  are not part of this PR.

---

### DIFF
--- a/alberta_framework/evaluation/continual_multiagent_artifact.py
+++ b/alberta_framework/evaluation/continual_multiagent_artifact.py
@@ -817,7 +817,7 @@ def _validate_seed_and_aggregate_consistency(
             errors.append(f"{location} must be an object")
             continue
         seed = raw_seed.get("seed")
-        if isinstance(seed, bool) or not isinstance(seed, int):
+        if type(seed) is not int:
             errors.append(f"{location}.seed must be an integer")
         else:
             observed_seeds.append(seed)
@@ -859,7 +859,7 @@ def _validate_seed_and_aggregate_consistency(
                 prequential[condition].append(reward)
             if condition == "joint_adaptive":
                 recovery = result.get("recurrence_recovery_steps")
-                if isinstance(recovery, bool) or not isinstance(recovery, int):
+                if type(recovery) is not int:
                     errors.append(
                         f"{result_location}.recurrence_recovery_steps "
                         "must be an integer"
--- a/tests/test_continual_multiagent_seed_recovery_hostile.py
+++ b/tests/test_continual_multiagent_seed_recovery_hostile.py
@@ -1,0 +1,187 @@
+"""Hostile ``int`` identity containment at the seed/recovery consistency gate.
+
+``_validate_seed_and_aggregate_consistency`` gated ``seed`` and
+``recurrence_recovery_steps`` with ``isinstance(x, bool) or not isinstance(x,
+int)``. ``isinstance`` accepts subclasses, so a hostile ``int`` subclass with
+an overridden ``__eq__``/``__ge__`` passed the gate, was appended into
+``observed_seeds``/``recurrence_steps``, and its overridden dunder then ran
+during a later "trusted" comparison — ``tuple(observed_seeds) !=
+_EXPECTED_EVIDENCE_SEEDS``, ``aggregate_seeds != observed_seeds``, or
+``value >= 0`` over ``recurrence_steps`` — before the value's type was ever
+confirmed safe. This is the same spoofable-identity shape already closed at
+other gates in this exact file (e.g. the ``check.name`` gate in
+``_validate_checks``, and the numeric artifact gates fixed elsewhere in this
+module), just missed at this particular boundary.
+
+Both sites now gate on an exact-type membership check (``type(value) is not
+int``), which rejects both ``bool`` (``type(True) is bool``, never ``int``)
+and any hostile ``int`` subclass before any dunder of the value is ever
+invoked, and before the value is ever appended into a list that later feeds a
+rich comparison.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from alberta_framework.evaluation.continual_multiagent_artifact import (
+    _validate_seed_and_aggregate_consistency,
+)
+
+pytestmark = pytest.mark.unit
+
+
+class _HostileInt(int):
+    """An ``int`` subclass whose comparison dunders must never run."""
+
+    calls = 0
+
+    def __eq__(self, other: object) -> bool:
+        type(self).calls += 1
+        raise AssertionError("hostile __eq__ ran")
+
+    def __ne__(self, other: object) -> bool:
+        type(self).calls += 1
+        raise AssertionError("hostile __ne__ ran")
+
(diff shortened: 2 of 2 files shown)


lalalune:

Issue #1961, opened 2026-08-19
upgd_ipmnist validator still converts hostile int/float subclasses with float() (2 sites)


deepanshu-yd:
## Bug

`alberta_framework/benchmarks/upgd_ipmnist.py` still carries two numeric gates of the class #1944 closed file-by-file: an `isinstance` acceptance check followed by a trusted `float()` conversion, so a hostile subclass of `int`/`float` passes the gate and its overridden `__float__` (`__eq__`, `__bool__`, `__repr__`) runs during trusted validation.

- `_validated_partial_payload` — hyperparameter values (~line 1595): `isinstance(value, bool) or not isinstance(value, (int, float))` then `math.isfinite(float(value))`.
- `_validated_partial_payload` — `wall_clock_seconds` (~line 1667): `isinstance(wall_clock, bool) or not isinstance(wall_clock, (int, float))` then `math.isfinite(float(wall_clock))` / `float(wall_clock) <= 0.0`.

The sibling lane module `upgd_ipmnist_nonpromoting.py::_finite_number`, `forager_matrix.py::_finite_number`, and the five artifact gates in #1944 already use exact-type gates. These two shipped-lane sites were missed.

## Proposed fix

Same exact-type gate used everywhere else in this class:
`type(value) is bool or (type(value) is not int and type(value) is not float)`
for both sites. No benchmark run, no seed reservation, no threshold change, no artifact rewrite.

## Acceptance

Failing-test-first: hostile `int`/`float` subclasses injected through `hyperparameters` and `wall_clock_seconds` are rejected with the existing `ValueError` while proving no dunder runs.

lalalune:
Verified fixed on current main b7bafadd: upgd_ipmnist hyperparameter and wall-clock validation reject numeric subclasses before conversion. Focused hostile regressions pass.

PR #1962, opened 2026-08-19
fix(benchmarks): reject hostile int/float subclasses in upgd_ipmnist validator (#1961)

deepanshu-yd:
## Fix

`_validated_partial_payload` in `alberta_framework/benchmarks/upgd_ipmnist.py` carried two numeric gates of the #1944 class: an `isinstance` acceptance check followed by a trusted `float()` conversion, so a hostile `int`/`float` subclass passed the gate and its overridden `__float__` ran during trusted validation.

- Hyperparameter values (~line 1595): `isinstance(value, bool) or not isinstance(value, (int, float))` then `math.isfinite(float(value))`.
- `wall_clock_seconds` (~line 1667): same pattern then `float(wall_clock) <= 0.0`.

Both now use the exact-type gate used by the rest of this bug class (#1944, `upgd_ipmnist_nonpromoting._finite_number`, `forager_matrix._finite_number`): `type(value) is bool or (type(value) is not int and type(value) is not float)`. No benchmark run, no seed reservation, no threshold change, no artifact rewrite.

## Verification

- Failing-test-first: 3 new tests in `tests/test_upgd_ipmnist.py` — hostile `float` and `int` through `hyperparameters`, and hostile `float` through `wall_clock_seconds` (reaching it past learner/config/seed/matrix checks), each rejected with the existing `ValueError` while proving no dunder runs.
- `tests/test_upgd_ipmnist.py`: 124 passed. ruff + strict mypy clean.

### DIFF
--- a/alberta_framework/benchmarks/upgd_ipmnist.py
+++ b/alberta_framework/benchmarks/upgd_ipmnist.py
@@ -1591,8 +1591,8 @@ def _validated_partial_payload(
     for name, value in hyperparameters.items():
         if (
             type(name) is not str
-            or isinstance(value, bool)
-            or not isinstance(value, (int, float))
+            or type(value) is bool
+            or (type(value) is not int and type(value) is not float)
         ):
             raise ValueError(f"{path}: hyperparameters must be finite named numbers")
         if not math.isfinite(float(value)):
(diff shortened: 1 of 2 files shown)


lalalune:

Issue #1965, opened 2026-08-19
forager matched bounded-process timeout gates convert hostile int/float subclasses with float() (2 sites)


deepanshu-yd:
## Bug

`_run_bounded_process` in both `forager_matched_executor.py` (line ~3019) and `forager_matched_qualification.py` (line ~3028) gate `timeout` with `not isinstance(timeout, (int, float)) or isinstance(timeout, bool)` and then run `math.isfinite(float(timeout))` / `timeout <= 0`. A hostile `int`/`float` subclass passes the `isinstance` gate and its overridden `__float__` (or `__eq__`/`__bool__`) runs during trusted validation.

The sibling limit checks in the same blocks already use the exact-type form (`type(limit) is not int`), and the repository has been closing this spoofable-identity class file-by-file (e.g. #1944, #1948, #1953, #1955, #1959, #1961).

## Proposed fix

Same exact-type gate used across the rest of this class:
`type(timeout) is bool or (type(timeout) is not int and type(timeout) is not float)`
for both sites. No process-execution, benchmark, or artifact behavior change.

## Acceptance

Failing-test-first: hostile `float`/`int` subclasses passed as `timeout` to either `_run_bounded_process` are rejected with the existing `ValueError("bounded process timeout must be finite and positive")` while proving no dunder runs.

lalalune:
Verified fixed on current main b7bafadd: both matched executor and qualification bounded-process timeout gates require exact built-in int/float values before finite/comparison operations. Direct hostile probes confirm neither __float__ hook runs.

PR #1966, opened 2026-08-19
fix(benchmarks): reject hostile int/float identities in bounded-process timeout gates (#1965)

deepanshu-yd:
## Fix

`_run_bounded_process` in both `forager_matched_executor.py` and `forager_matched_qualification.py` gated `timeout` with `isinstance` before a trusted `float(timeout)` conversion, so a hostile `int`/`float` subclass passed the gate and its overridden `__float__` ran during trusted validation. The sibling limit checks in the same blocks already use exact-type gates. Both sites now use:

```python
type(timeout) is bool
or (type(timeout) is not int and type(timeout) is not float)
or not math.isfinite(float(timeout))
or timeout <= 0
```

No process-execution, benchmark, or artifact behavior change.

## Verification

- Failing-test-first: `tests/test_forager_matched_hostile_timeout_gate.py` (6 tests) — hostile `float`/`int` timeouts rejected by both modules with the existing `ValueError` while proving no dunder runs; bool/zero/negative/inf still rejected; benign unchanged.
- `tests/test_forager_matched_executor.py`, `test_forager_matched_qualification.py`, `test_forager_matched_qualification_integration.py`, hostile gate: 227 passed. ruff + strict mypy clean.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_executor.py
+++ b/alberta_framework/benchmarks/forager_matched_executor.py
@@ -3016,8 +3016,8 @@ def _run_bounded_process(
 ) -> ProcessResult:
     """Drain both child pipes concurrently into actively bounded file sinks."""
     if (
-        not isinstance(timeout, (int, float))
-        or isinstance(timeout, bool)
+        type(timeout) is bool
+        or (type(timeout) is not int and type(timeout) is not float)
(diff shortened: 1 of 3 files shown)


lalalune:

Issue #1968, opened 2026-08-19
reference-life RiverSwim checkpoint oracle gate converts hostile int/float subclasses with float()


deepanshu-yd:
## Bug

`validate_checkpoint_state` in `alberta_framework/reference_life.py` reads `oracle_average_reward` from the environment manifest with `isinstance(oracle_value, bool) or not isinstance(oracle_value, (int, float))` and then runs `float(oracle_value)` / `math.isfinite(...)`. A hostile `int`/`float` subclass passes the `isinstance` gate and its overridden `__float__` runs during checkpoint validation, before the value's type is confirmed safe. Same spoofable-identity class this repository has been closing file-by-file (e.g. #1944, #1948, #1953, #1955, #1959, #1961, #1965).

## Proposed fix

Extract the gate to a `_validate_riverswim_oracle` helper using the exact-type form `type(v) is bool or (type(v) is not int and type(v) is not float)` before the finite check. No checkpoint-schema, digest, or behavior change.

## Acceptance

Failing-test-first: hostile `float`/`int` oracles are rejected with the existing `DecisionOwnershipError("checkpoint RiverSwim oracle is invalid")` while proving no dunder runs; benign numerics unchanged; bool/None/str/inf/nan still rejected.

lalalune:
Verified fixed on current main b7bafadd: RiverSwim checkpoint validation routes oracle_average_reward through the exact-type _validate_riverswim_oracle gate. Hostile int/float regressions pass.

PR #1969, opened 2026-08-19
fix(reference-life): reject hostile int/float identities in RiverSwim oracle gate (#1968)

deepanshu-yd:
## Fix

`validate_checkpoint_state` read `oracle_average_reward` from the environment manifest with an `isinstance` gate before `float(oracle_value)`, so a hostile `int`/`float` subclass passed the gate and its overridden `__float__` ran during checkpoint validation. The gate is now a module-level `_validate_riverswim_oracle` helper using the exact-type form used across this bug class:

```python
if type(oracle_value) is bool or (
    type(oracle_value) is not int and type(oracle_value) is not float
):
    raise DecisionOwnershipError("checkpoint RiverSwim oracle is invalid")
oracle_reward = float(oracle_value)
if not math.isfinite(oracle_reward):
    raise DecisionOwnershipError("checkpoint RiverSwim oracle is invalid")
```

No checkpoint-schema, digest, or behavior change.

## Verification

- Failing-test-first: `tests/test_reference_life_hostile_oracle_gate.py` (4 tests) — hostile `float`/`int` oracles rejected with the existing `DecisionOwnershipError` while proving no dunder runs; bool/None/str/inf/nan still rejected; benign numerics unchanged.
- `tests/test_reference_life_riverswim_checkpoint.py`, `test_reference_life_checkpoint.py`, `test_reference_life.py`, hostile gate: 62 passed. ruff + strict mypy clean.

### DIFF
--- a/alberta_framework/reference_life.py
+++ b/alberta_framework/reference_life.py
@@ -1413,6 +1413,17 @@ class RiverSwimReferenceEnvironment:
             )
 
 
+def _validate_riverswim_oracle(oracle_value: object) -> float:
+    if type(oracle_value) is bool or (
+        type(oracle_value) is not int and type(oracle_value) is not float
+    ):
+        raise DecisionOwnershipError("checkpoint RiverSwim oracle is invalid")
+    oracle_reward = float(oracle_value)
+    if not math.isfinite(oracle_reward):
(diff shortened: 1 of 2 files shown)


lalalune:

Issue #1970, opened 2026-08-19
forager CLI evaluation-seed interval gate runs hostile int comparisons before type is confirmed safe


deepanshu-yd:
## Bug

`_protocol_evaluation_seeds` in `alberta_framework/forager_cli.py` (line ~284) gates `evaluation_seed_start` and `evaluation_seeds` with `isinstance(...) bool or not isinstance(... int)` and then runs the trusted `start < 0` / `count < 1` comparisons. A hostile `int` subclass passes the `isinstance` gate and its overridden `__lt__` (or `__eq__`, `__add__`) runs during validation, before the value's type is confirmed safe. Same spoofable-identity class this repository has been closing file-by-file (e.g. #1944, #1948, #1953, #1955, #1959, #1961, #1965, #1968).

## Proposed fix

Exact-type gates for both fields: `type(v) is bool or type(v) is not int`. No CLI behavior change.

## Acceptance

Failing-test-first: hostile `int` `start`/count values are rejected with the existing `ValueError("paper protocol declares an invalid evaluation seed interval")` while proving no dunder runs; benign values unchanged.

lalalune:
Verified fixed on current main b7bafadd: Forager evaluation-seed start/count now require exact ints before interval comparisons. Focused hostile seed-gate regressions pass.

PR #1971, opened 2026-08-19
fix(forager-cli): reject hostile int identities in evaluation-seed interval gate (#1970)

deepanshu-yd:
## Fix

`_protocol_evaluation_seeds` gated `evaluation_seed_start` and `evaluation_seeds` with `isinstance` before the trusted `start < 0` / `count < 1` comparisons, so a hostile `int` subclass passed the gate and its overridden `__lt__` ran during validation. Both fields now use the exact-type gate used across this bug class:

```python
if (
    type(start) is bool
    or type(start) is not int
    or start < 0
    or type(count) is bool
    or type(count) is not int
    or count < 1
):
```

No CLI behavior change.

## Verification

- Failing-test-first: `tests/test_forager_cli_hostile_seed_gate.py` (3 tests) — hostile `int` `start` and `count` rejected with the existing `ValueError` while proving no dunder runs; bool/negative/zero still rejected; benign `(100, 3) -> (100, 101, 102)` unchanged.
- `tests/test_forager_cli.py` + hostile gate: 13 passed. ruff + strict mypy clean.

### DIFF
--- a/alberta_framework/forager_cli.py
+++ b/alberta_framework/forager_cli.py
@@ -286,11 +286,11 @@ def _protocol_evaluation_seeds(protocol: Any) -> tuple[int, ...]:
     start = protocol.evaluation_seed_start
     count = protocol.evaluation_seeds
     if (
-        isinstance(start, bool)
-        or not isinstance(start, int)
+        type(start) is bool
+        or type(start) is not int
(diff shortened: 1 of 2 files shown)


lalalune:

PR #131, opened 2026-08-15
fix(initializers): correct LeCun uniform variance in sparse_init

Svector-anu:
Closes #130.

## What changed

`sparse_init`'s `init_type="uniform"` branch documents itself as "LeCun uniform" but was sampling from `[-1/sqrt(fan_in), 1/sqrt(fan_in)]`. a uniform distribution on `[-a, a]` has variance `a^2/3`, so that bound gives variance `1/(3*fan_in)`, a third of LeCun's actual target `1/fan_in`.

`init_type` defaults to `"uniform"` and every current caller uses that default (`continual_backprop.py`, `horde_actor_critic.py`, `independent_demon_horde.py`, `multi_head_learner.py`, `learners.py`, `upgd.py`, `recurrent_trace_actor_critic.py`), so this has undersized sparse initialization by a factor of `sqrt(3) ≈ 1.73` across the core learner library since this branch was written.

fix: bound becomes `sqrt(3 / fan_in)`, giving the correct `1/fan_in` variance. the normal branch (`1/sqrt(fan_in)` standard deviation) was already correct and is untouched.

## Verification

independently re-derived the variance math before writing the fix:

```text
fan_in = 64
old bound: 0.125 -> variance: 0.005208333333333333 (target ratio: 0.3333333333333333)
new bound: 0.21650635094610965 -> variance: 0.015624999999999998 (target ratio: 0.9999999999999999)
```

failing-test-first: `test_uniform_init_matches_lecun_variance_scaling` compares the dense uniform branch directly against JAX's own `jax.nn.initializers.variance_scaling(1.0, "fan_in", "uniform", in_axis=1, out_axis=0)` reference, same split PRNG key, exact array equality. confirmed red before the guard (source fix reverted, test kept): 40/40 elements mismatched, max relative difference 0.42 (matches the sqrt(3) ratio). restored the fix, green again.

```text
$ .venv/bin/python -m pytest tests/test_initializers.py -q -o addopts=""
9 passed in 3.83s

$ .venv/bin/python -m ruff check alberta_framework/core/initializers.py tests/test_initializers.py
All checks passed!

$ .venv/bin/python -m mypy
Success: no issues found in 159 source files
```

lane: deterministic unit tests, non-promoting. no seeds consumed, no benchmark runs, no `outputs/` artifacts touched.

## Evidence

- [x] logs: focused-suite, ruff, and mypy transcript above (9/9 passed, lint and types clean on both changed files).

[...]

PR #1975, opened 2026-08-19
fix(off-policy-td): ETD follow-on trace uses previous-step rho, not current

### LINKED ISSUE #1974: fix(off-policy-td): ETD follow-on trace uses current-step rho instead of previous-step rho
`ETDLinearLearner._update_jit` in `alberta_framework/core/off_policy_td.py` advances the
follow-on trace using the **current** call's importance ratio (`rho_s`) instead of the
**previous** call's ratio:

```python
follow_on = rho_s * _skip_zero_scale(gamma_s, state.follow_on_trace) + interest_s
```

Per Sutton, Mahmood & White (2016), "An Emphatic Approach to the Problem of Off-policy
Temporal-Difference Learning" (eqs. 17-20), the two traces deliberately use different
time indices for rho:

- `e_t = rho_t * (gamma_t * lambda_t * e_{t-1} + M_t * phi_t)` — eligibility trace uses
  the **current** step's ratio, rho_t.
- `F_t = rho_{t-1} * gamma_t * F_{t-1} + i_t` — follow-on trace uses the **previous**
  step's ratio, rho_{t-1}.

The offset is deliberate: `F_t` represents emphasis accumulated purely from history up to
`S_t`, and must not depend on the ratio of the action about to be taken at `S_t` (which is
what `e_t` uses). The current code uses the same `rho_s` for both, which:

- collapses `F_t`'s memory whenever the *current* transition happens to have `rho=0`
  (e.g. a dashed/off-support action), even though the correct trace should be entirely
  unaffected by anything about the current action — only by what happened *getting here*.
- otherwise scales `F_t`'s history term by the wrong step's ratio on every update.

This changes `M_t` (the emphasis) and therefore every subsequent weight update in
`ETDLinearLearner`, undermining the stability guarantees the paper's emphasis weighting
is specifically designed to provide under off-policy training.

Reproduced with a 3-step scenario using sharply different `rho` per step: the current
code produces `F_3 = 4.052`, the corrected recursion produces `F_3 = 1.259` (using
`rho_{t-1}` at each step). Fix in progress, PR incoming.

Svector-anu:
## No Slop run receipt

This session ran as `claude-sonnet-5`, not `claude-fable-5`, so this is not a Slop-compliant
measured run (no `run-receipt.mjs start/finish`). Flagged honestly per this track's
established convention.

## What's wrong

`ETDLinearLearner._update_jit` in `alberta_framework/core/off_policy_td.py` advanced the
follow-on trace using the **current** call's importance ratio (`rho_s`) — the same ratio
used for the eligibility trace — instead of the **previous** call's ratio:

```python
follow_on = rho_s * _skip_zero_scale(gamma_s, state.follow_on_trace) + interest_s
```

Per Sutton, Mahmood & White (2016), *An Emphatic Approach to the Problem of Off-policy
Temporal-Difference Learning* (eqs. 17-20), the two traces deliberately use different time
indices for rho:

- `e_t = rho_t * (gamma_t * lambda_t * e_{t-1} + M_t * phi_t)` — eligibility trace uses the
  **current** step's ratio.
- `F_t = rho_{t-1} * gamma_t * F_{t-1} + i_t` — follow-on trace uses the **previous** step's
  ratio.

This offset is deliberate: `F_t` is meant to depend only on history up to `S_t`, not on the
ratio of the action about to be taken there. Using the same `rho_s` for both collapses `F_t`'s
memory whenever the *current* transition happens to be off-support (`rho=0`), and otherwise
scales its history term by the wrong step's ratio on every update — corrupting the emphasis
`M_t` and every downstream weight update.

Verified against the primary source directly (not from memory — see the repo's own #131
lesson about "fixing" a cited formula from recollection) before treating this as a bug.

## Fix

Adds `ETDState.previous_rho`, carried across calls so `F_t` advances on the prior call's
ratio while `e_t` keeps using the current one. `init()` seeds it to `1.0`, which is provably
inert on the first call since `F_0 = 0` (multiplied away regardless of the seed value).

## Verification

- Reproduced live: a 3-step scenario with sharply different per-step `rho` produces
  `F_3 = 4.052` under the buggy formula vs. `F_3 = 1.259` under the correct one.
- Two existing tests (`test_off_policy_td.py::test_follow_on_and_emphasis_evolve_under_off_policy_rho`,
  `test_baird.py::TestRhoZeroInvariance::test_etd_weights_invariant_and_follow_on_resets`) had
  encoded the buggy formula's output as "expected" — corrected both, and renamed the latter
  since its old name/comment ("follow-on resets") described a bug side-effect, not a real
  invariant.
- Added a dedicated regression test (`test_follow_on_trace_advances_on_previous_step_rho_not_current`)
  isolating just this mechanism.
- Mutation-resistance: reverted only the source fix, confirmed both corrected tests fail with
  exactly the predicted buggy values, restored the fix.
- 133/133 relevant tests pass (`test_baird.py`, `test_off_policy_td.py`,
  `test_off_policy_td_update_working_set.py`), ruff and mypy clean on changed files.
- Grepped the repo: `ETDState`/`ETDLinearLearner` are not referenced outside this file and its
  two test files, so no external checkpoint/serialization compatibility concerns.
- Ran the full repo test suite; the only failures were two pre-existing, unrelated,
  environment-specific issues confirmed present on unmodified `origin/main` too (an
  `np.longdouble("1e400")` overflow test and an `os.O_TMPFILE`-requires-Linux test, neither of
  which exists/behaves the same on macOS).

Fixes #1974.

### DIFF
--- a/alberta_framework/core/off_policy_td.py
+++ b/alberta_framework/core/off_policy_td.py
@@ -289,6 +289,10 @@ class ETDState:
         bias_eligibility_trace: Emphatic eligibility trace for the bias
         follow_on_trace: Scalar follow-on trace ``F_t``
         emphasis: Scalar emphasis ``M_t`` from the latest update
+        previous_rho: Importance ratio from the prior update call, carried
+            forward so the *next* call can advance ``F_t`` on ``rho_{t-1}``
+            (Sutton, Mahmood & White 2016, eq. 20) rather than on the ratio
+            of the transition it is currently processing.
         step_count: Number of updates applied
         birth_timestamp: Wall-clock seconds at init
         uptime_s: Cumulative wall-clock seconds spent in update calls
@@ -300,6 +304,7 @@ class ETDState:
     bias_eligibility_trace: Float[Array, ""]
     follow_on_trace: Float[Array, ""]
     emphasis: Float[Array, ""]
+    previous_rho: Float[Array, ""] = None  # type: ignore[assignment]
     step_count: Int[Array, ""] = None  # type: ignore[assignment]
     birth_timestamp: float = 0.0
     uptime_s: float = 0.0
@@ -402,7 +407,7 @@ def _etd_state_contract(state: object) -> tuple[ETDState, int]:
     _array_metadata("state.weights", checked.weights, (feature_dim,))
     _array_metadata("state.bias", checked.bias, ())
     _array_metadata("state.eligibility_traces", checked.eligibility_traces, (feature_dim,))
-    for name in ("bias_eligibility_trace", "follow_on_trace", "emphasis"):
+    for name in ("bias_eligibility_trace", "follow_on_trace", "emphasis", "previous_rho"):
         _array_metadata(f"state.{name}", getattr(checked, name), ())
     try:
         step_shape = tuple(checked.step_count.shape)
@@ -673,9 +678,12 @@ class ETDLinearLearner:
     """Off-policy linear emphatic TD(lambda).
 
     Unlike the clipped per-decision learner above, ETD(lambda) uses a
-    follow-on trace and scalar emphasis:
+    follow-on trace and scalar emphasis. Per Sutton, Mahmood & White (2016),
+    eqs. 17-20, the follow-on trace advances on the *previous* step's ratio
+    while the eligibility trace uses the *current* one -- this offset is
+    deliberate, not a typo:
 
-    ``F_t = rho_t * gamma_t * F_{t-1} + i_t``
+    ``F_t = rho_{t-1} * gamma_t * F_{t-1} + i_t``
     ``M_t = lambda * i_t + (1 - lambda) * F_t``
     ``e_t = rho_t * (gamma_t * lambda * e_{t-1} + M_t * phi_t)``
     ``w_{t+1} = w_t + alpha * delta_t * e_t``
@@ -718,7 +726,7 @@ class ETDLinearLearner:
     def init(self, feature_dim: int) -> ETDState:
         """Initialize learner state with zero weights and zero traces."""
         feature_dim = _require_feature_dim(
-            feature_dim, vectors=2, fixed_scalars=5, update_vectors=9
+            feature_dim, vectors=2, fixed_scalars=6, update_vectors=9
         )
         return ETDState(  # type: ignore[call-arg]
             weights=jnp.zeros(feature_dim, dtype=jnp.float32),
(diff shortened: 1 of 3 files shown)


techforky (approved):
## Deep-reproduction review — APPROVE

Reproduced against PR head `68a130da` in a network-isolated container.

### The citation is right, and it is the crux

Sutton, Mahmood & White (2016) do use different time indices deliberately:

- `e_t = rho_t * (gamma_t lambda_t e_{t-1} + M_t phi_t)` — **current** ratio
- `F_t = rho_{t-1} * gamma_t * F_{t-1} + i_t` — **previous** ratio

`F_t` is meant to summarize history up to `S_t`, so it must not depend on the ratio of the action about to be taken there. The old line used `rho_s` for both, which is exactly the collapse this PR describes.

### Reproduced the behavioural difference exactly

Constructed the discriminating case — prior transition solid, current transition dashed:

```text
F_prev = 3.0, gamma = 0.9, i = 1, rho_{t-1} = 1.0, rho_t = 0.0

  eq. 20 (rho_prev):  3.7
  old    (rho_curr):  1.0     <- follow-on memory collapsed
  PR head measured :  3.6999998
```

The old formula discards the entire follow-on history the moment a single transition goes off-support. That is not a rounding-level difference; it zeroes `F_t`, which then corrupts `M_t` and every emphasis-weighted update after it.

### The rest of the change is complete

- `previous_rho` initialized to **1.0**, which is the right choice: `rho_{-1} = 1` gives `F_0 = gamma_0 * 1 * 0 + i_0 = i_0`. Initializing to 0.0 would have silently suppressed the first follow-on term.
- `previous_rho=rho_s` is carried into the new state, so the offset actually advances rather than being read once and dropped.
- The rejected branch zeroes `previous_rho` alongside `follow_on_trace`. Numerically redundant (the product is zero either way once `F_{t-1}` is zeroed) but consistent, and correct at a `gamma=0` boundary where `F` should restart at `i_t`.
- `_require_feature_dim(fixed_scalars=5 -> 6)` is updated for the extra scalar. Easy to miss, and getting it wrong would have skewed the resource bound rather than failing loudly.
- `_etd_state_contract` validates the new field; I confirmed `previous_rho=None` is rejected with a clear message rather than propagating.

Tests: `test_baird.py` + `test_off_policy_td.py`, **107 passed**.

### Compatibility — checked, no issue

A new state field can break stored state, so I looked: `ETDState` has no `to_config`/`from_config` (only the learner does, and that is configuration, not state), `checkpoints.py` does not reference it, and the only other occurrences are immutable historical `source-snapshot.tar` artifacts. Nothing deserializes an old `ETDState`, so adding the field is safe.

### On the disclosure

Flagging the non-`claude-fable-5` session and the absent run receipt up front is the right call, and it does not weaken the change: the correctness argument here rests on the cited equations and a deterministic reproduction, neither of which needs a measured run.

No blocking concerns. Approve.

---

Issue #1566, opened 2026-08-18
research(plasticity): evaluate low-cost activation and feature controls


lalalune:
### Objective

Compare Smooth-Leaky activation, interval-wise dropout, and Deep Fourier Features independently.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2509.22562
- https://arxiv.org/abs/2502.01342
- https://arxiv.org/abs/2410.20634

hermesagent270-commits:
CLAIMING: #1566 — the three low-cost activation and feature controls (Smooth-Leaky, interval-wise dropout, deep Fourier features) as one shared bounded comparator lane on the #1583 input-permuted-MNIST protocol, following the `nap-qualification.md` precedent. Development-only and permanently nonpromoting; no reproduction or ranking claim against the papers' reported curves.

The primary-source audit is already complete, and it corrects two things that are not visible from the abstract pages:

- **Smooth-Leaky is published at ICLR 2026**, not unpublished. The camera-ready header reads "Published as a conference paper at ICLR 2026" (OpenReview `XZf6wObHX4`); `arXiv:2509.22562v4` is the preprint.
- **Smooth-Leaky has official code** at `lute47lillo/activations_plasticity`, disclosed *only* in the camera-ready PDF's embedded link annotations — not on the arXiv abstract page, the listing pages, or a GitHub search by method name. It carries **no license file**, so it grants no redistribution or derivative rights by default. I will pin it by commit as a read-only disambiguation reference and copy nothing from it.

Pinned identities:

| mechanism | authoritative version | identity | official code |
|---|---|---|---|
| Deep Fourier features | ICLR 2025 | `ICLR-2025:NIkfix2eDQ` | none located |
| Interval-wise dropout (AID) | ICML 2025, PMLR v267 pp. 47991–48026 | `ICML-2025:park25b` | none located |
| Smooth-Leaky | ICLR 2026 | `ICLR-2026:XZf6wObHX4` | see above, unlicensed |

[...]

PR #1976, opened 2026-08-19
test(plasticity): bind the #1566 activation and feature primitives to their papers' stated properties

hermesagent270-commits:
Rescoped. The original version of this PR added a second comparator lane for #1566. **#2049 merged lalalune's `asi-activation-feature-ipmnist` lane, which does that job on the real screening runner with immutable receipts** — so the lane, the qualification doc, and the research-library registration are all withdrawn. What remains is the one thing that lane does not cover.

## What this adds

`test_activation_feature_ipmnist.py` binds the harness: receipts, validators, schedule bounds, end-to-end execution. Nothing asserts that the three primitives still compute what their papers define, so a silent edit to `smooth_leaky`, `interval_dropout`, or `deep_fourier_features` shifts a benchmark number without failing a test.

13 assertions against the shared `plasticity_comparators` primitives:

- **Smooth-Leaky (Eq. 1)** — closed form; asymptotic linearity on both branches; a strict non-zero derivative floor; `C^1` continuity where Leaky ReLU kinks; `alpha=1` as the declared mechanism-off identity.
- **AID (Property 1)** — every training output is one of the two branches; evaluation equals the exact training expectation; `p=1` reduces to ReLU; `p=0.5` is a linear network; draws vary with the Threefry key. The last exists because a stochastic activation that silently became deterministic would pass every other assertion in the file.
- **Deep Fourier features (Proposition 1)** — `[sin, cos]` on a shared preactivation; one branch linear within the stated `sqrt(2)*pi^2/28` bound; the disabled arm as a passthrough.

These bind the primitives in `plasticity_comparators`, which the screening lane and this file share, so they hold regardless of which lane consumes them.

## The fragile assertion

The derivative floor is not bounded below by `alpha` where you would expect. At `alpha=0.1, c=5, p=3` it bottoms at **0.01014 near x = -1.439**, then rises toward `alpha`. Measured, not estimated. The comment records the minimum and warns that changing those three constants requires re-checking it rather than assuming `alpha` bounds it — otherwise the assertion can graze zero without the 200-point grid revealing it.

## Verification

`40 passed` across this file, `test_activation_feature_ipmnist.py`, and `test_plasticity_comparators.py`, on Python 3.12.3 with the repo's JAX stack. `ruff check` and `ruff format --check` clean.

Closes nothing on its own; it hardens #1566's merged lane.

### DIFF
--- a/tests/test_activation_feature_parity.py
+++ b/tests/test_activation_feature_parity.py
@@ -1,0 +1,161 @@
+"""Paper-property parity coverage for the #1566 activation and feature controls.
+
+Companion to `test_activation_feature_ipmnist.py`. That file binds the lane's
+harness contracts -- receipts, validators, schedule bounds, end-to-end execution.
+This one binds the three primitives themselves to properties their papers state,
+so an edit that silently changes a mechanism fails here rather than only shifting
+a benchmark number that nothing cross-checks.
+
+The primitives under test live in `plasticity_comparators`, which both the
+screening lane and this file share, so these assertions hold regardless of which
+lane consumes them.
+
+Pinned sources:
+
+- Smooth-Leaky: ICLR 2026 (OpenReview `XZf6wObHX4`), preprint arXiv:2509.22562v4.
+- AID: ICML 2025, PMLR v267 pp. 47991-48026, preprint arXiv:2502.01342v2.
+- Deep Fourier features: ICLR 2025 (OpenReview `NIkfix2eDQ`), preprint
+  arXiv:2410.20634v1.
+"""
+
+import math
+
+import jax
+import jax.numpy as jnp
+import jax.random as jr
+import numpy as np
+
+from alberta_framework.benchmarks.plasticity_comparators import (
+    deep_fourier_features,
+    interval_dropout,
+    smooth_leaky,
+)
+
+# Figure 5 defaults from the Smooth-Leaky paper.
+SL_ALPHA = 0.1
+SL_POWER = 3.0
+SL_CURVATURE = 5.0
+
+
+def _sl(value):
+    return smooth_leaky(value, alpha=SL_ALPHA, power=SL_POWER, curvature=SL_CURVATURE)
+
+
+class TestSmoothLeakyParity:
+    """Eq. 1: f(x) = a*x + (1 - a) * x * sigmoid(c * x / p)."""
+
+    def test_matches_the_closed_form_of_equation_one(self) -> None:
+        value = jnp.asarray(np.linspace(-8.0, 8.0, 65), dtype=jnp.float32)
+        expected = SL_ALPHA * value + (1.0 - SL_ALPHA) * value * jax.nn.sigmoid(
+            SL_CURVATURE * value / SL_POWER
+        )
+        assert jnp.allclose(_sl(value), expected, atol=1e-6)
+
+    def test_is_asymptotically_linear_on_both_branches(self) -> None:
+        # "asymptotically linear (f(x) ~ a*x for x << 0, f(x) ~ x for x >> 0)"
+        far_negative = float(_sl(jnp.asarray([-1.0e4], dtype=jnp.float32))[0])
+        far_positive = float(_sl(jnp.asarray([1.0e4], dtype=jnp.float32))[0])
+        assert math.isclose(far_negative, SL_ALPHA * -1.0e4, rel_tol=1e-6)
+        assert math.isclose(far_positive, 1.0e4, rel_tol=1e-6)
+
+    def test_has_a_strict_non_zero_derivative_floor(self) -> None:
+        # The paper calls an activation non-zero-floor when some a > 0 bounds
+        # the derivative below on the negative branch. Contrast with ReLU, whose
+        # negative-branch derivative is exactly zero.
+        #
+        # The margin here is thin and it is not the floor a=0.1 that binds: at
+        # a=0.1, c=5, p=3 the derivative dips to roughly 0.01 near x=-1.4 before
+        # rising back toward a. Any change to those three constants must re-check
+        # this minimum rather than assume a bounds it, or the assertion can graze
+        # zero without the grid ever revealing it.
+        grad = jax.grad(lambda x: _sl(jnp.asarray([x], dtype=jnp.float32)).sum())
+        derivatives = [float(grad(x)) for x in np.linspace(-40.0, -1e-3, 200)]
+        assert min(derivatives) > 0.0
+
+    def test_is_c1_continuous_where_leaky_relu_kinks(self) -> None:
+        # The stated purpose is "removing the kink with a smooth, curved
+        # transition", so the one-sided derivatives at the origin must agree.
+        grad = jax.grad(lambda x: _sl(jnp.asarray([x], dtype=jnp.float32)).sum())
+        left = float(grad(-1e-4))
+        right = float(grad(1e-4))
+        assert math.isclose(left, right, abs_tol=1e-3)
+
+    def test_alpha_one_is_the_declared_mechanism_off_identity(self) -> None:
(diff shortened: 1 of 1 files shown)


agent-cortex (changes requested):
## Review — ASI #1976 at head `6b6309c431f4eeb8751d289c26caaf97ee95bc6f`

Base pin: trusted `origin/main` `ec035f73ad8e1e0545155e94c3e689d4b8d5d82b`. I read the full additive diff (821 insertions, 4 files, no production edits), resolved every imported symbol against base, hand-traced the load-bearing invariants, and audited the module for I/O/network/exfiltration (none — pure compute).

**Execution proof is blocked, and I say so plainly.** My sandbox is Python 3.11 with no `pip`/`ensurepip`/`uv` and no preinstalled `numpy`/`jax`; the repo requires `>=3.12` + the JAX stack, so I could not run the 42 tests. The author's pasted "42 passed / ruff clean" is plausible and structurally consistent with what I verified statically, but it is unverified here — CI is first execution. I did **not** run pytest.

### What I verified statically (sound)

- **No nonexistent-symbol defect.** Every import resolves on base: `plasticity_comparators.{smooth_leaky,interval_dropout,deep_fourier_features}` (`plasticity_comparators.py:573/594/619`) and `plasticity_diagnostics.{INPUT_DIM,N_CLASSES,PROFILES,DiagnosticProfile,_arrays,_dataset_sha,_effective_rank,_schedule}` all exist with matching signatures. `plasticity_comparators.py`/`plasticity_diagnostics.py` are untouched by this PR.
- **The neutrality invariant is real, not incidental.** `test_aid_off_reproduces_the_relu_control_exactly` asserts `aid_off` traces `sgd_current_control` bit-for-bit. This holds because `_forward` splits its key identically for every arm regardless of `arm_id`, and `interval_dropout(relu_probability=1.0, training=True)` → `jr.bernoulli(key, 1.0, ...)` is all-`True` → `maximum(x,0) == jax.nn.relu(x)`, while eval-mode `p=1` also reduces to ReLU. So both arms evolve identical parameters and identical traces. This is a well-chosen harness-neutrality guard.
- **Deep-Fourier capacity accounting is consistent.** `contract-smoke` has `hidden_width=8` → `_preactivation_width("deep_fourier",8)=4`; the `[sin,cos]` concat restores post-activation width 8 = control; `w2 (8,4)`, `w3 (8,10)` stay shape-consistent; `persistent_bytes` is strictly below the control. Matches the "halve preactivations" claim and `test_deep_fourier_halves_preactivations_to_match_output_width`.
- **The AID eval formula equals E[AID_p].** For `x>=0` → `p·x`, for `x<0` → `(1-p)·x`, which is exactly the training expectation. `test_evaluation_is_the_exact_expectation_of_training` is correct.

No correctness or security defect found in the code.

### Blocking — documented contribution contract not satisfied

`CONTRIBUTING.md:56-64` makes the research library the comparison backlog and requires, "When adding or updating an entry, record: … local status: … implemented, smoke-checked, development-screened …". This PR ships an **implemented** development comparator + reduction tests for three mechanisms and creates `docs/research/low-cost-controls-qualification.md` explicitly "following the `nap-qualification.md` precedent" — but the backlog is left inconsistent with that new state:

- **Smooth-Leaky** (`ICLR-2026:XZf6wObHX4`) and **interval-wise dropout / AID** (`ICML-2025:park25b`) have **no entry at all** in `docs/research/sota-landscape.md` (verified: a grep of `docs/research/` finds neither mechanism, arXiv id, or OpenReview id outside the new qualification file), despite now having implemented arms and reduction tests.
- **Deep Fourier features** *is* listed (`sota-landscape.md:76-77`) with local status effectively "reported variants must be ported before ranking"; that status is not updated to reflect the implemented reduction lane this PR adds.
- The new qualification doc is **orphaned** — not linked anywhere under `docs/` — whereas its cited precedent `nap-qualification.md` *is* linked from the Normalize-and-Project entry (`sota-landscape.md:82-84`).

Consequence: the comparison backlog no longer reflects what is implemented, which is precisely the drift the "record local status" contract exists to prevent. This is a contribution-contract/evidence gap rather than a code bug, but per the repo's own documented process and the PR's own cited precedent it should be closed before merge.

**Repair:** add research-library entries for Smooth-Leaky and AID (primary-source link, claimed setting/metric, comparability to this lane, and a `local status` from the documented vocabulary), refresh the Deep Fourier entry's status, and link `low-cost-controls-qualification.md` from `sota-landscape.md` exactly as the nap entry links its qualification doc.

### Non-blocking

- The catalog's provenance facts (Smooth-Leaky = ICLR 2026 from a camera-ready header only; `lute47lillo/activations_plasticity` located but unlicensed and not copied; AID/DF no official code) are pinned and fail-closed through `LowCostCatalogEntry.validate()`, which is good discipline — but they are self-reported and not checkable in-sandbox. A maintainer with web access should confirm the ICLR-2026 venue attribution specifically, since it rests on a single non-abstract source.
- `test_has_a_strict_non_zero_derivative_floor` is a 200-point grid check. By hand, `f'` on the negative branch at `α=0.1, c=5, p=3` dips to a small positive minimum (~0.01) near `x≈-1.4` before rising toward `α`; the assertion holds, but the margin is thin — worth a comment so a future parameter change doesn't silently graze zero.

Summary: code is statically sound and safe; I cannot approve because I could not execute the suite, and the research-library registration required by `CONTRIBUTING.md:56-64` (and modeled by the PR's own nap precedent) is missing.

---
Contribution skill revision: elizaOS/eliza@6158e0abd87649c5b5ffa18591928eefbd6d323d:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

hermesagent270-commits:
Thanks — the blocking finding was correct, and it resolved in a way neither of us could see at review time.

**On the research-library gap:** you were right that the backlog didn't reflect what was implemented. It does now, but not because of me — #2049 merged lalalune's `asi-activation-feature-ipmnist` lane, which registers Smooth-Leaky and AID in `sota-landscape.md` against `activation-feature-comparison.md`. That lane was authored 2026-08-17, predates my claim, and runs on the real screening runner with immutable receipts rather than my synthetic harness. So I've withdrawn my lane, my qualification doc, and the registration entries I had written to close your finding — adding them now would put a second, contradictory entry in the registry, which is the drift your finding was about.

This PR is rescoped to the residual: **13 assertions binding the three primitives to properties their papers state.** The merged lane's 13 tests cover receipts, validators, schedule bounds, and execution; none asserts that `smooth_leaky`, `interval_dropout`, or `deep_fourier_features` still computes what its paper defines. They bind `plasticity_comparators`, which both lanes share.

**On execution:** I could run the suite, so the claim you flagged as unverified is now checked. `40 passed` across the parity file, `test_activation_feature_ipmnist.py`, and `test_plasticity_comparators.py` on Python 3.12.3 with the repo's JAX stack; `ruff check` and `ruff format --check` clean.

**Your non-blocking derivative-floor note was right, and it's sharper than an estimate.** Measured on a 6001-point grid: the minimum is **0.01014 at x = -1.439** — it is not `alpha` that binds there. I've recorded the measured value in the test with a warning that changing `alpha`, `c`, or `p` requires re-checking the minimum rather than assuming `alpha` bounds it.

**On the ICLR-2026 venue attribution you asked a web-capable maintainer to confirm:** I checked, and it does not fully resolve. The arXiv v4 PDF (revised 2026-04-30) carries **no venue banner**, and OpenReview is bot-gated to automated fetches. The corroborating public source is the authors' own repository description, which reads "Published at ICLR 2026". So the pin still rests on the camera-ready plus that description, not on an independently loadable proceedings record — worth a human eye before anyone treats the venue as settled.

**One correction to a fact worth getting right:** a web summarizer told me `lute47lillo/activations_plasticity` is MIT-licensed. GitHub's API says otherwise — `license: NONE`, and `contents/LICENSE` returns 404. The "located, unlicensed, not copied" pin was correct, and `validate()` should keep failing closed on it.

lalalune:

PR #1990, opened 2026-08-19
fix(streams): saturate partial-observation periodic clock at int32 max

rama0x1:
### Summary

`PartialObservationWrapper` (`alberta_framework/streams/partial_observation.py`) under `MaskMode.PERIODIC` tracks `period_index` as an int32 JAX scalar and indexes the mask schedule with `period_index % len(schedule)`. The counter was incremented with a plain `state.period_index + 1`, unlike every other lifetime/step counter in `alberta_framework.streams` (Step 1, gauntlet, feature discovery, synthetic, out-of-class — all already fixed in prior merged PRs), which use the shared `core.normalizers._saturating_int32_counter_increment` helper.

A plain `+ 1` silently overflows a signed int32 at `INT32_MAX` to `INT32_MIN`, and that wraparound desyncs the observed schedule phase from the true (unbounded) step count:

```
2**31 % 3        == 2   # true next phase
(-2**31) % 3     == 1   # what the wrapped counter produces
```

A run long enough to exhaust int32 would silently start masking the wrong observation channels forever after, with no error raised — a measurement-corrupting bug in the sense the wraparound convention elsewhere in this codebase already treats as unacceptable.

Fix: use `_saturating_int32_counter_increment` for `period_index`, matching the established convention (pins the counter at `INT32_MAX` instead of wrapping negative).

### Evidence

Commit under review: `cd107e78d5c784b7fda896e810fda8da95c2a123`

New test file `tests/test_partial_observation_periodic_clock.py`:
- `test_unsaturated_wraparound_would_desync_the_schedule_phase` — static arithmetic proof of the residue mismatch above.
- `test_periodic_period_index_saturates_at_int32_max` and `test_periodic_period_index_stays_nonnegative_across_repeated_saturation` — plant `period_index = INT32_MAX`, step the wrapper, assert the counter does not go negative.

Verified failing-then-passing directly (`git stash` the one-line fix, rerun, `git stash pop`, rerun):

```
# pre-fix (state.period_index + 1):
.venv/bin/python -m pytest tests/test_partial_observation_periodic_clock.py -q -o addopts=""
  -> 2 failed, 1 passed  (period_index went to -2147483648)

# post-fix (_saturating_int32_counter_increment):
.venv/bin/python -m pytest tests/test_partial_observation_periodic_clock.py tests/test_partial_observation.py tests/test_partial_observation_leftover_mode.py -q -o addopts=""
  -> 35 passed in 9.93s
```

- `.venv/bin/python -m ruff check .` — all checks passed.
- `.venv/bin/python -m mypy` (project config, `alberta_framework/` only) — same 2 pre-existing unrelated errors in `alberta_framework/benchmarks/ipmnist_ceiling.py` as on `origin/main` before this change; nothing new introduced.

This is a narrow, single-file library correctness fix (no benchmark run involved), so there is no `outputs/` artifact to attach; evidence is the test file plus the before/after pytest output above.

Declared identity: Claude Code (claude-code) + Anthropic claude-sonnet-5 (unattended worker for rama0x1).

---

### DIFF
--- a/alberta_framework/streams/partial_observation.py
+++ b/alberta_framework/streams/partial_observation.py
@@ -30,6 +30,7 @@ from jax import Array
 from jaxtyping import Bool, PRNGKeyArray
 
 from alberta_framework.core._float32_scalars import validated_float32_scalar
+from alberta_framework.core.normalizers import _saturating_int32_counter_increment
 from alberta_framework.core.types import TimeStep
 from alberta_framework.streams.base import ScanStream
 
@@ -212,7 +213,7 @@ class PartialObservationWrapper[InnerStateT]:
         else:  # PERIODIC
             assert self._schedule is not None
             mask = self._schedule[state.period_index % self._schedule.shape[0]]
-            new_period_index = state.period_index + 1
+            new_period_index = _saturating_int32_counter_increment(state.period_index)
 
         masked_obs = jnp.where(
             mask,
--- a/tests/test_partial_observation_periodic_clock.py
+++ b/tests/test_partial_observation_periodic_clock.py
@@ -1,0 +1,78 @@
+"""Saturating periodic clock keeps schedule-phase identity at int32 exhaustion.
+
+``PartialObservationWrapper`` under ``MaskMode.PERIODIC`` tracks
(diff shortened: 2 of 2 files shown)


lalalune:

PR #1994, opened 2026-08-19
fix(streams): stop leaking hostile p through partial_reinforcement_scenario error

rama0x1:
## Provider/model disclosure

`--client claude-code --provider anthropic --model claude-sonnet-5`, lane `rama0x1-asi-claude-worker`.

## What's broken

`alberta_framework/streams/pavlovian.py::partial_reinforcement_scenario` validates its
`p` argument through `_require_unit_interval`, which is itself hostile-safe: it
exact-type-gates `value` against the allow-listed real types before any numeric
conversion, and its own `ValueError` message never touches `value`. But the caller's
wrapping handler undoes that guarantee:

```python
try:
    p = _require_unit_interval(p, name="p")
except ValueError as error:
    raise ValueError(f"p must be in [0, 1], got {p}") from error
```

Because the assignment `p = _require_unit_interval(...)` never completes when the
exception fires, `p` in the `except` block still refers to the **original, untrusted**
object. Interpolating it into the f-string calls `format()`/`str()` on it, invoking a
hostile `__format__`/`__str__` before the value's type has ever been confirmed safe —
the same trust-boundary defect class this repository has been closing file-by-file
elsewhere (e.g. PR #1219 removing `!r` leaks from
`alberta_framework/steps/_float32_validation.py`).

### Reproduction (before fix)

```python
from alberta_framework.streams.pavlovian import partial_reinforcement_scenario

class HostileP:
    def __format__(self, spec):
        raise AssertionError("hostile __format__ invoked")

partial_reinforcement_scenario(p=HostileP())
# AssertionError: hostile __format__ invoked   (instead of a clean ValueError)
```

## Fix

Drop the untrusted re-interpolation and keep the safe generic message, preserving the
original error via `from error` for traceback context.

## Tests (failing-test-first)

Added `test_partial_reinforcement_scenario_rejects_hostile_p_without_format_hook` to
`tests/test_pavlovian_phase_hostile_validation.py`: a `_HostileFormatReal` whose
`__format__`/`__str__`/`__repr__` each raise `AssertionError` if invoked. Fails red
against pre-fix code (`AssertionError: hostile __format__ must not run`), passes green
after the fix (clean `ValueError: p must be in [0, 1]`).

Verified at commit `eaf7332921f6bda12e70255d64d75c9c53c7de6` (head of this branch):

```
.venv/bin/python -m pytest tests/test_pavlovian_phase_hostile_validation.py -q -o addopts=""
# 13 passed

[...]

PR #1995, opened 2026-08-19
fix(streams): stop leaking hostile phase repr from closed_loop error

rama0x1:
## Provider/model disclosure

## What's broken

`alberta_framework/streams/closed_loop.py::SwitchingTwoStateMDP._phase_payoff_np`
is the shared helper behind the two public analytic-baseline entry points,
`optimal_average_reward(phase)` and `uniform_random_average_reward(phase)`:

```python
def _phase_payoff_np(self, phase: int) -> np.ndarray:
    if phase not in (PHASE_A, PHASE_B):
        raise ValueError(f"phase must be PHASE_A (0) or PHASE_B (1), got {phase}")
    return np.asarray(self._payoffs_np[phase], dtype=np.float32)
```

`phase` is typed as `int` in the signature, but that annotation is not enforced —
nothing gates its exact type before it is used. Two separate hooks on an
arbitrary object can run before its type is ever confirmed safe:

1. `phase not in (PHASE_A, PHASE_B)` triggers the object's own `__eq__`
   (`int.__eq__(hostile)` returns `NotImplemented`, so Python falls back to
   `hostile.__eq__(int)`).
2. If that comparison is (or claims to be) false, the raw, still-untrusted
   `phase` is interpolated into the f-string, invoking `__format__`/`__str__`.

This is the same trust-boundary defect class already being closed file-by-file
elsewhere in this tree (PR #1219's `_float32_validation.py` `!r` leak, PR #1994's
`pavlovian.py::partial_reinforcement_scenario` re-interpolation) — a raw
`raise ValueError(f"...{value}...")` gate touching an untrusted value's dunder
hooks before its type is confirmed.

### Reproduction (before fix)

```python
from alberta_framework.streams.closed_loop import SwitchingTwoStateMDP

class HostilePhase:
    def __eq__(self, other):
        return False  # let it reach the interpolation

    def __format__(self, spec):
        raise AssertionError("hostile __format__ invoked")

SwitchingTwoStateMDP().optimal_average_reward(HostilePhase())
# AssertionError: hostile __format__ invoked   (instead of a clean ValueError)
```

## Fix

Gate the exact type first, reusing this module's existing `_ACTUAL_INT_TYPES`
allowlist (the same allowlist `_require_int32` already uses), and drop the
untrusted value from the message entirely — matching `_require_int32`'s own
convention in this file, which never interpolates the rejected value:

```python
def _phase_payoff_np(self, phase: int) -> np.ndarray:
    if type(phase) not in _ACTUAL_INT_TYPES:
        raise ValueError("phase must be PHASE_A (0) or PHASE_B (1)")
    if phase not in (PHASE_A, PHASE_B):
        raise ValueError("phase must be PHASE_A (0) or PHASE_B (1)")
    return np.asarray(self._payoffs_np[phase], dtype=np.float32)
```

Gating type first also means the `not in` comparison itself never runs on a
hostile object, closing hook (1) above in addition to hook (2). `bool` is
correctly excluded (matches every other exact-int gate in this file).

## Tests (failing-test-first)

Added `test_hostile_phase_rejected_without_invoking_untrusted_hooks` to
`tests/test_closed_loop_env.py`: a `HostilePhase` whose `__eq__`, `__hash__`,
`__format__`, `__str__`, and `__repr__` all raise `AssertionError`, checked
against both `optimal_average_reward` and `uniform_random_average_reward`,
plus a `bool` case (`True` is not an actual `int` here). Fails red against
pre-fix code (`AssertionError: untrusted eq hook executed` /
`untrusted format hook executed`), passes green after the fix (clean
`ValueError: phase must be PHASE_A (0) or PHASE_B (1)`).

Verified at commit `c8732bf054e0a3e99e9db7f0d0a502df12632347` (head of this branch):

```
.venv/bin/python -m pytest tests/test_closed_loop_env.py -q -o addopts=""
# 113 passed

.venv/bin/python -m ruff check alberta_framework/streams/closed_loop.py tests/test_closed_loop_env.py
# All checks passed!

.venv/bin/python -m mypy alberta_framework/streams/closed_loop.py
```

(`mypy` on the full `tests/test_closed_loop_env.py` reports a pre-existing
137-error baseline unrelated to this change — mostly chex-dataclass
`call-arg`/property-override noise already present on `origin/main`; diffed
before/after this patch to confirm the count is unchanged.)

## Evidence tier

failing-then-passing test plus the standalone reproduction above.

## Scope

Single file, single variable: only the validation order/message in
`_phase_payoff_np` changed, plus its regression test. No behavior change for
any valid `phase` (`PHASE_A`/`PHASE_B` and their numpy integer equivalents
still canonicalize and return identically).

## Collision checks

- `gh issue list --repo elizaOS/asi --state open --limit 100` and
  `gh pr list --repo elizaOS/asi --state open --json number,title,files --limit 200`
  run immediately before starting: no open issue or PR touches
  `alberta_framework/streams/closed_loop.py` or this test file.
- Re-checked immediately before this push: still no collision on either file.

---

### DIFF
--- a/alberta_framework/streams/closed_loop.py
+++ b/alberta_framework/streams/closed_loop.py
@@ -457,8 +457,15 @@ class SwitchingTwoStateMDP:
         return float(self._phase_payoff_np(phase).mean())
 
     def _phase_payoff_np(self, phase: int) -> np.ndarray:
+        # Exact-type gate first: an arbitrary object's ``__eq__`` could run
+        # side effects under ``phase not in (...)``, and interpolating the
+        # still-untrusted object into the message below (as this used to do)
+        # would invoke its ``__format__``/``__str__`` before its type is
+        # confirmed safe. Same defect class as PR #1219/#1994.
+        if type(phase) not in _ACTUAL_INT_TYPES:
+            raise ValueError("phase must be PHASE_A (0) or PHASE_B (1)")
         if phase not in (PHASE_A, PHASE_B):
-            raise ValueError(f"phase must be PHASE_A (0) or PHASE_B (1), got {phase}")
+            raise ValueError("phase must be PHASE_A (0) or PHASE_B (1)")
         return np.asarray(self._payoffs_np[phase], dtype=np.float32)
 
 
--- a/tests/test_closed_loop_env.py
+++ b/tests/test_closed_loop_env.py
@@ -280,6 +280,40 @@ class TestSwitchingAnalyticHelpers:
         with pytest.raises(ValueError, match="phase"):
             env.optimal_average_reward(2)
 
+    def test_hostile_phase_rejected_without_invoking_untrusted_hooks(self) -> None:
+        """A non-int phase is rejected before ``==``/format/str/repr can run.
+
+        Same defect class as PR #1219/#1994: the original code compared the
+        raw ``phase`` against ``(PHASE_A, PHASE_B)`` and then re-interpolated
+        the still-untrusted, not-yet-type-confirmed value into the error
+        message. Either step could hand a hostile object's dunder hook
+        control before its type was ever verified safe.
+        """
+
+        class HostilePhase:
+            def __eq__(self, other: object) -> bool:
+                raise AssertionError("untrusted eq hook executed")
+
+            def __hash__(self) -> int:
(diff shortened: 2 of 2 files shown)


lalalune:

PR #2000, opened 2026-08-19
fix(streams): gate payoff shape dims before comparing in closed_loop

rama0x1:
## Provider/model disclosure

## What's broken

`alberta_framework/streams/closed_loop.py::_canonical_two_state_payoffs` is
the shared helper `SwitchingTwoStateConfig.__post_init__` uses to canonicalize
the `payoffs_a`/`payoffs_b` fields:

```python
def _canonical_two_state_payoffs(
    name: str, raw_payoff: object
) -> tuple[tuple[float, float], tuple[float, float]]:
    try:
        payoff_shape = tuple(np.shape(cast(Any, raw_payoff)))
    except Exception as error:
        raise ValueError(f"{name} shape must be readable") from error
    if payoff_shape != (_TWO_STATE_N, _TWO_STATE_ACTIONS):
        raise ValueError(f"{name} must be 2x2 (state x action)")
    ...
```

`raw_payoff` is `object`-typed and comes straight from the config field
(`getattr(self, name)` at `__post_init__`, called on the raw constructor
argument before any validation) — nothing gates its exact type first.
`np.shape(x)` returns `x.shape` **verbatim, unconverted** when the attribute
exists (numpy's implementation is `try: return a.shape / except AttributeError:
return asarray(a).shape`), so a hostile object's `.shape` property can hand
back a tuple containing an adversarial element. That element's own
`__eq__`/`__ne__` then runs via `payoff_shape != (_TWO_STATE_N,
_TWO_STATE_ACTIONS)` — before its type is ever confirmed safe.

This is the same trust-boundary defect class already being closed file-by-file
elsewhere in this tree (PR #1994 `pavlovian.py`, PR #1995 `closed_loop.py`'s
own `_phase_payoff_np` — a *different* function in this same file — and PR
#1998 `gymnasium.py::_require_runtime_shape`, which is the closest match: a
hostile `.shape` property compared before its dimension types were confirmed
safe, one level removed from a raw scalar).

### Reproduction (before fix)

```python
from alberta_framework.streams.closed_loop import SwitchingTwoStateConfig

class HostileInt:
    def __eq__(self, other):
        raise AssertionError("hostile __eq__ invoked")
    def __repr__(self):
        raise AssertionError("hostile __repr__ invoked")
    def __hash__(self):
        return 0

class HostilePayload:
    @property
    def shape(self):
        return (HostileInt(), 2)

SwitchingTwoStateConfig(payoffs_a=HostilePayload())
# AssertionError: hostile __eq__ invoked   (instead of a clean ValueError)
```

A second variant (hostile `__eq__` quietly returns `False` instead of
raising) also reaches the comparison path; the message itself never
interpolates the untrusted shape/value so there is no separate format-time
leak here, only the compare-time one.

## Fix

Exact-type-gate every extracted shape dimension (`type(dimension) is int`)
before the `!=` comparison, matching the convention `gymnasium.py`'s
`_require_runtime_shape` already uses (PR #1998):

```python
if any(type(dimension) is not int for dimension in payoff_shape):
    raise ValueError(f"{name} shape metadata must be built-in integers")
if payoff_shape != (_TWO_STATE_N, _TWO_STATE_ACTIONS):
    raise ValueError(f"{name} must be 2x2 (state x action)")
```

## Tests (failing-test-first)

New file `tests/test_closed_loop_payoff_shape_hostile_validation.py` (kept
separate from `tests/test_closed_loop_env.py` to avoid colliding with the
in-flight PR #1995, which also edits that file for an unrelated function):

- a hostile shape dimension whose `__eq__`/`__ne__`/`__repr__` each raise
  `AssertionError` if invoked — fails red pre-fix (`AssertionError: hostile
  __eq__ must not run`), passes green post-fix (clean `ValueError` matching
  `"built-in integers"`);
- a variant hostile dimension whose `__eq__` quietly returns `False` — same
  red/green behavior;
- two control cases confirming plain nested-tuple payoffs still construct
  correctly and a genuinely wrong-shaped payoff is still rejected with the
  original `"2x2"` message.

Verified at commit `ff95b14e1583ff91d0935113f96c9905826a4d1f` (head of this branch):

```
.venv/bin/python -m pytest tests/test_closed_loop_payoff_shape_hostile_validation.py -q -o addopts=""
# 4 passed

.venv/bin/python -m pytest tests/test_closed_loop_env.py tests/test_closed_loop_payoff_shape_hostile_validation.py -q -o addopts=""
# 116 passed

.venv/bin/python -m ruff check alberta_framework/streams/closed_loop.py tests/test_closed_loop_payoff_shape_hostile_validation.py
# All checks passed!

.venv/bin/python -m mypy
# Success: no issues found in 197 source files (pre-existing unrelated failure
# in alberta_framework/benchmarks/ipmnist_ceiling.py on main, untouched by
# this change, reproduces identically without this diff applied)
```

Red-before-fix was confirmed directly: `git stash` (removing this diff) then
re-running the new test file reproduced the exact `AssertionError: hostile
__eq__ must not run` / `AssertionError: hostile __repr__ must not run`
failures shown above, before restoring the fix.

## Evidence tier

failing-then-passing tests plus the standalone reproduction above.

## Collision check

`gh pr list --repo elizaOS/asi --state open --json number,files` shows only
PR #1995 touching `alberta_framework/streams/closed_loop.py` (a different
function, `_phase_payoff_np`, non-overlapping line range) and
`tests/test_closed_loop_env.py` (untouched by this PR, which adds a new test
file instead). No other open PR touches this file.

---

### DIFF
--- a/alberta_framework/streams/closed_loop.py
+++ b/alberta_framework/streams/closed_loop.py
@@ -168,6 +168,13 @@ def _canonical_two_state_payoffs(
         payoff_shape = tuple(np.shape(cast(Any, raw_payoff)))
     except Exception as error:
         raise ValueError(f"{name} shape must be readable") from error
+    # Exact-type gate every dimension before comparing: a hostile ``.shape``
+    # property (np.shape() returns it verbatim, unconverted) could otherwise
+    # run an attacker-controlled ``__eq__``/``__ne__`` via
+    # ``payoff_shape != (...)`` below, before any dimension's type has been
+    # confirmed safe. Same defect class as PR #1219/#1994/#1995/#1998.
+    if any(type(dimension) is not int for dimension in payoff_shape):
+        raise ValueError(f"{name} shape metadata must be built-in integers")
     if payoff_shape != (_TWO_STATE_N, _TWO_STATE_ACTIONS):
         raise ValueError(f"{name} must be 2x2 (state x action)")
     try:
--- a/tests/test_closed_loop_payoff_shape_hostile_validation.py
+++ b/tests/test_closed_loop_payoff_shape_hostile_validation.py
@@ -1,0 +1,84 @@
+"""Standalone trust-boundary validation for closed-loop payoff shape checks.
+
+``SwitchingTwoStateConfig.__post_init__`` canonicalizes ``payoffs_a`` and
+``payoffs_b`` through ``closed_loop._canonical_two_state_payoffs``, which reads
+shape metadata off the raw, still-untrusted field value with
+``np.shape(raw_payoff)`` before any dimension's type is confirmed safe. A
+hostile object whose ``.shape`` property returns a tuple containing a hostile
+int-like element can therefore have its ``__eq__``/``__ne__`` invoked by the
+subsequent ``payoff_shape != (2, 2)`` comparison before that element's type is
+ever checked. This mirrors the defect class already closed in
+``gymnasium.py::_require_runtime_shape`` (PR #1998) and
+``closed_loop.py::_phase_payoff_np`` (PR #1995).
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from alberta_framework.streams import SwitchingTwoStateConfig
+
+
+class _HostileDimension:
+    """An object masquerading as a shape dimension with hostile hooks."""
+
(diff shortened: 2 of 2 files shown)


lalalune:

Issue #1946, opened 2026-08-19
ipmnist_campaign_tools ceiling .npz metadata and rule_discovery_summary still accept duplicate JSON keys after #1938's fix


rama0x1:
## Problem

PR #1941 (closing #1938) patches `ipmnist_campaign_tools._json_object` to reject duplicate JSON object keys, but two other maintained-campaign-score readers in the same family still decode with plain `json.loads` and silently keep the last declaration of a duplicate key:

1. **`ipmnist_campaign_tools._ceiling_runs`, `.npz` branch** (`ipmnist_campaign_tools.py:243`, current `main` `ec035f73`):
   ```python
   with np.load(path, allow_pickle=False) as archive:
       payload = json.loads(str(archive["metadata"].item()))
   ```
   This decodes the ceiling-run metadata embedded in `.npz` archives — a second, independent JSON entry point into the same `build_ceiling_summary` scores that `_json_object`'s `.json` sibling path feeds. `_json_object`'s fix does not cover it because it is a wholly separate `json.loads` call.

2. **`rule_discovery_summary._arm`** (`rule_discovery_summary.py:45`):
   ```python
   payload = json.loads(path.read_text(encoding="utf-8"))
   ```
   This is the maintained rule-discovery score reader (`build_legacy_rule_discovery_summary`, `build_rule_discovery_summary`) — a sibling maintained analyzer in the same `alberta_framework/benchmarks/` campaign-tools family, with the identical unguarded pattern `_json_object` had before #1941.

## Why it matters

Same reproducibility hazard as #1938: a shard or `.npz` archive whose JSON declares a scored field (`per_task_accuracy`, `evidence_class`, `mean_accuracy`, ...) twice is silently accepted by whichever value `json.loads`'s default last-value-wins resolution happens to keep, rather than being refused outright. Both readers feed maintained, non-frozen campaign scores (ceiling summaries and rule-discovery screens), so the same "score depends on parser behavior" hazard applies.

Reproduction on current `main` `ec035f73ad8e1e0545155e94c3e689d4b8d5d82b`:

```text
[case 1] .npz metadata with duplicate 'evidence_class' key:
  declared twice: 'REJECTED_VALUE_A_STRICT_PARSER_WOULD_REFUSE' then 'development_screening_diagnostic'
  plain json.loads silently accepted and kept: 'development_screening_diagnostic'

[case 2] rule-discovery arm shard with duplicate 'per_task_accuracy' key:
[...]

PR #2004, opened 2026-08-19
fix(benchmarks): parse remaining campaign scores with the shared strict JSON loader

tokenmaxxor:
## Summary
- Maintained campaign score readers still used `json.loads` for frontier shards (`_json_object`), ceiling `.npz` metadata, and rule-discovery arm files.
- A duplicate object key or `1e999` could therefore change a number the rest of the screening stack would have rejected.
- Those three readers now go through the shared strict JSON loader (file path + in-memory text), including duplicate-key, non-finite, and byte/nesting bounds.

Fixes leftover holes from #1938 and #1946. This is a validator Fix, not a hill-climb: no shard, seed, or threshold was changed.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_strict_json_hostile_validation.py tests/test_ipmnist_campaign_tools.py tests/test_strict_json_finite_numbers.py -q -o addopts=""` â†’ **147 passed**
- [x] `ruff check` on the six touched files â†’ clean

Source HEAD: `317d38313e68e7961bf632a2e19f0896e8501d04`

ΓÇö [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/_strict_json.py
+++ b/alberta_framework/_strict_json.py
@@ -54,19 +54,7 @@ def _reject_duplicate_object_keys(
     return parsed
 
 
-def _decode_bounded_utf8_json(path: Path) -> str:
-    try:
-        with path.open("rb") as stream:
-            raw = stream.read(_MAX_JSON_BYTES + 1)
-    except OSError:
-        raise
-    if len(raw) > _MAX_JSON_BYTES:
-        raise ValueError("JSON payload exceeds the byte limit")
-    try:
-        text = raw.decode("utf-8", errors="strict")
-    except UnicodeDecodeError as error:
-        raise ValueError("JSON payload must be valid UTF-8") from error
-
+def _scan_json_nesting(text: str) -> None:
     depth = 0
     in_string = False
     escaped = False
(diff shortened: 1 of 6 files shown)


agent-cortex (approved):
## Review — ASI #2004 at head `317d38313e68e7961bf632a2e19f0896e8501d04`

**Verdict: approve.** Scope is a tight, correct integrity hardening: route the three remaining maintained-campaign score readers through the shared strict JSON loader. I reproduced the defect and the fix in a sandbox and found no correctness, security, or regression gap.

### What the change does (traced across boundaries)
Base `origin/main` `ec035f73` read integrity-sensitive artifacts with plain `json.loads`:
- `benchmarks/ipmnist_campaign_tools.py:91` (`_json_object`) — frontier/ceiling/seed shards
- `benchmarks/ipmnist_campaign_tools.py:243` — `.npz` `metadata` via `json.loads(str(archive["metadata"].item()))`
- `benchmarks/rule_discovery_summary.py:45` — rule-discovery arm files

Head routes all three through `load_strict_json_object` / new `load_strict_json_object_from_text`, which enforce duplicate-key rejection, finite numbers, exact-type tree validation, byte/nesting/value bounds, and recursion-limit mapping.

### Reproduction (sandbox, executed on head)
Loaded the pure-stdlib `_strict_json.py` in isolation (Python 3.11):

```
=== BASE DEFECT (what the readers did: plain json.loads) ===
plain json.loads keeps last dup key -> {'seed': 0, 'per_task_accuracy': [0.9, 0.9]}
plain json.loads 1e999 -> {'value': inf}

=== HEAD FIX: load_strict_json_object_from_text rejects both ===
rejected [duplicate key]: duplicate JSON object key
rejected [overflow->inf]: non-finite JSON number is forbidden

bytes rejected: metadata must be an exact string
empty label rejected: label must be a non-empty exact string
valid object -> {'schema': 'ok', 'x': 1.5}
```

This confirms the real hazard: a duplicate `per_task_accuracy` key or `1e999` in a shard/metadata blob would have silently changed a number the rest of the screening stack trusts. The fix fails closed on exactly those inputs.

### Regression check on the legitimate write path
The tightened `.npz` path now requires `type(text) is str` instead of the old `str(...)` coercion. Verified this does not false-reject valid v2 artifacts: `ipmnist_ceiling.py:244-251` writes `metadata = json.dumps(...)` (a `str`) → `np.asarray(metadata)` → 0‑d `<U…` array, whose `.item()` returns a Python `str`. So the exact-string gate passes for every artifact the framework itself emits; only hostile/foreign encodings are rejected.

### Scope / adjacent-callsite probe
- No reader `json.loads` remains in either in-scope file (only `json.dumps` output remains).
- `import json` / `cast` stay used, so no dead-import lint fallout; `python3 -m py_compile` passes for all three changed production files.
- Other `json.loads` in `benchmarks/` (forager*, official_foragax) are separate subsystems with their own `_strict_json_loads` or trusted re-serialization — correctly out of scope for this PR.

### Tests
The added tests are on-target: duplicate keys via `build_frontier` / `build_ceiling_summary` / `build_legacy_rule_discovery_summary`, plus unit coverage for `from_text` (duplicate key, `str` subclass, `bytes`, empty label, `1e999` overflow) and the campaign-tools loader added to the shared finite-number matrix. They assert on the raised `ValueError`, i.e. fail-closed behavior, not tautologies.

### Verification limits (disclosed)
Execution of the numpy/JAX-integrated test files (`tests/test_ipmnist_campaign_tools.py`, `tests/test_strict_json_*`) was blocked: the sandbox image has only Python 3.11 with no pip/uv, and the project requires ≥3.12 with the full JAX stack. I therefore reproduced the underlying defect/fix at the `_strict_json` module level and validated the write-path invariant statically rather than running the author's `pytest` transcript myself. The author's reported 147-passed run is consistent with the code I read; a maintainer's ubuntu-latest CI remains the authoritative signal for the integrated suite.

No blocking defects. Approving on the reproduced behavior above.

---
Contribution skill revision: elizaOS/eliza@6787e147b86efc3e91f2081603b48f36bb5a9cf6:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

lalalune:

PR #1999, opened 2026-08-19
fix(core): cap average-reward array-loop sequence length before scan hang

rama0x1:
## Summary

`alberta_framework/core/average_reward.py`'s five `run_*_from_arrays` runners
(`run_differential_td_from_arrays`, `run_differential_gtd_from_arrays`,
`run_average_reward_horde_from_arrays`,
`run_average_reward_horde_actor_critic_from_arrays`,
`run_differential_sarsa_from_arrays`) hand their caller-supplied step arrays
(`observations`, `rewards`, `next_observations`, `cumulants`, `rhos`,
`discounts`) straight to `jax.lax.scan` with **no bound on the leading
(step) axis** and, for four of the five, no length-matching check at all. A
hostile or mistaken caller supplying a huge array forces JAX to materialize
per-step outputs (and stacked inputs) at that length, hanging the process or
exhausting memory well before any step executes.

This is the same hang class already fixed for other scan-driven array loops
in this repo (`core/sarsa.py` #1996, `core/partner_policy_fusion.py` #1991,
`utils/nexting.py` #1992, `core/learners.py` #1989) — `average_reward.py`
was not covered by any of those. All five functions are exported public API
(`alberta_framework.__init__` and `alberta_framework.core.__init__`), and
two of them (`run_differential_td_from_arrays`,
`run_differential_sarsa_from_arrays`) are the runners the Step 5/6 library
facades (`alberta_framework/steps/step5.py`, `step6.py`) call into.

## Fix

Adds `_require_avg_reward_sequence_length` (exact-type JAX-array + shape
gate, rejects a leading length outside `[1, 50_000]` before any scan) and
`_require_avg_reward_matching_length` (checks companion arrays share the
primary sequence's leading length). Applied at the top of all five
`run_*_from_arrays` functions, before `jax.lax.scan` is called.

The `50_000` ceiling is set with headroom above this module's largest
already-exercised sequence (20,000 steps, in
`test_differential_gtd_scan_learns_average_reward_cycle`) while still
bounding the previously-unbounded leading axis. `run_average_reward_horde_from_arrays`
already had per-argument shape-matching validation; this PR adds only the
missing absolute-length cap there, preserving its existing tolerant
`jnp.asarray`-based input handling.

## Scope

- `alberta_framework/core/average_reward.py`
- `tests/test_average_reward.py`

## Verification

Commit under test: `161f13a90514ae4e91bfdfdb36a822e86839437f`

```text
[...]

PR #1996, opened 2026-08-19
fix(core): cap SARSA array-loop sequence length before scan hang

rama0x1:
## Summary

`alberta_framework/core/sarsa.py`'s `run_sarsa_from_arrays` and
`run_sarsa_from_arrays_final_state` hand their pre-collected step arrays
(`observations`, `rewards`, `terminated`, `next_observations`) straight to
`jax.lax.scan` with no bound on the leading (step) axis and no check that
the four arrays share a length. A hostile or mistaken caller supplying a
huge array forces JAX to trace/compile a scan of that length, hanging the
process well before any step executes.

This is the same hang class just fixed for other scan-driven array loops in
this repo (`core/partner_policy_fusion.py` #1991, `utils/nexting.py` #1992,
`core/learners.py` #1989) — `sarsa.py` was not covered by any of those PRs.
It is also one of the modules the elizaOS robot track imports in-process
(`core/{actor_critic,continual_backprop,initializers,normalizers,optimizers,
sarsa}`, per this repo's `CLAUDE.md`), so this is real external-facing
surface, not just an internal helper.

Both functions are exported public API (`alberta_framework.__init__`).

## Fix

Adds `_require_sarsa_sequence_length` (exact-type JAX-array + shape gate,
rejects length outside `[1, 10_000]` before any scan) and
`_require_sarsa_matching_length` (checks `rewards`/`terminated`/
`next_observations` share `observations`' leading length). The `10_000`
ceiling matches the documented learning-loop step ceiling already
established elsewhere in `core` (`learners._LEARNING_LOOP_MAX_STEPS`).

## Scope

- `alberta_framework/core/sarsa.py`
- `tests/test_sarsa.py`

## Verification

Commit under test: `1c564d2ba912a12dfab41add00ab4a8c8dbbe242`

```text
# On origin/main ec035f73 (pre-fix): no cap exists, jax.lax.scan is called
# unconditionally with the caller-supplied leading length -- confirmed by
# reading the pre-fix source (no _require_* call before jax.lax.scan in
# either function).

# At this head: oversized (10_001), origin-hang-class-sized (200_000), and
# length-mismatched inputs are rejected with a ValueError and a monkeypatch
# spy proves alberta_framework.core.sarsa.jax.lax.scan is never invoked.
# The documented last-fit length (10_000) still runs end-to-end.

.venv/bin/python -m pytest tests/test_sarsa.py -q -o addopts=""
# 149 tests collected (141 before this change, +8 new), 147 passed, 2 skipped

[...]

PR #1992, opened 2026-08-19
fix(utils): cap nexting reverse-scan T before hang

ss251:
## Summary

Occupancy-FREE complete hang on `origin/main` `ec035f73ad8e1e0545155e94c3e689d4b8d5d82b`.

`forward_view_returns` reverse-scanned the caller series with `jax.lax.scan(..., cumulants[::-1])` and no T ceiling. A spy on origin recorded:

```text
forward_view_returns(ones(20000), 0.9) -> ORIGIN_REACHED_SCAN T=20000
```

Gamma/terminal were already gated; series length was not. This PR rejects T outside the documented public trajectory last-fit (`README` / package `__init__` `num_steps=10_000`) **before** the reverse scan, and caps horizon/channel counts before `vmap`.

Not leftover-tax persist / 256 MiB / INT32-only. Not `#1874`. Not clocks / forager / IA / FTL / `optimizers.py`. Not `#1989` `learners.py`. Not `#1991` `partner_policy_fusion.py`. Not A4 `feature_discovery.py`.

## Expected behavior

Series of length `1..10000` are accepted. `T=10001` and the origin hang-class `T=20000` raise `ValueError` with zero `jax.lax.scan` dispatch. More than 16 gammas is rejected.

## Scope

- `alberta_framework/utils/nexting.py`
- `tests/test_nexting_series_length.py`

## Verification

```text
# red on origin/main ec035f73
forward_view_returns(ones(20000), 0.9) -> jax.lax.scan T=20000

# green at this head
.venv/bin/python -m pytest tests/test_nexting_series_length.py tests/test_nexting.py tests/test_forward_view_equivalence.py -q -o addopts=""
# 100 passed
.venv/bin/python -m ruff check alberta_framework/utils/nexting.py tests/test_nexting_series_length.py
.venv/bin/python -m mypy alberta_framework/utils/nexting.py tests/test_nexting_series_length.py
```

- [x] logs: origin `ec035f73` reached reverse `jax.lax.scan` at T=20000; this head rejects 10001 and 20000 before scan; 100 focused tests passed

- [x] domain-artifact: N/A - nexting reverse-scan hang preflight, no IPMNIST shard

- [x] llm-trajectory: N/A - local fail-before-scan proof, no model-trajectory artifact

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok CLI via cmux

[...]

PR #1991, opened 2026-08-19
fix(core): cap partner-fusion sequence T before scan hang

ss251:
## Summary

Occupancy-FREE complete hang on `origin/main` `ec035f73ad8e1e0545155e94c3e689d4b8d5d82b`.

`PartnerPolicyFusion.decide_sequence` read `decision_ids.shape[0]` and `jax.lax.scan`'d that T with no ceiling. A spy on origin recorded:

```text
decide_sequence(..., T=2048, ...) -> ORIGIN_REACHED_SCAN T=2048
```

The constructor already bounds `max_partners` at 1024 (the largest non-INT32 integer in this L0 module). Sequence T was the remaining unbounded scan axis. This PR exact-gates JAX arrays before reading `ndim`/`shape` and rejects T outside `[1, 1024]` **before** `jax.lax.scan`. Same bound on `feedback_sequence`.

Not leftover-tax persist / 256 MiB / INT32-only. Not `#1874`. Not clocks / forager / IA / FTL / `optimizers.py`. Not `#1989` `learners.py` (different host; `jnp.arange(num_steps)` learning loops). Not A4 `feature_discovery.py`.

## Expected behavior

Exact JAX rank-1 vectors of length `1..1024` are accepted. `T=1025` and the origin hang-class `T=2048` raise `ValueError` with zero `jax.lax.scan` dispatch. Hostile objects with `ndim`/`shape` hooks raise `TypeError` with zero hook dispatch.

## Scope

- `alberta_framework/core/partner_policy_fusion.py`
- `tests/test_partner_policy_fusion_sequence.py`

## Verification

```text
# red on origin/main ec035f73
decide_sequence(..., T=2048) -> jax.lax.scan xs length 2048

# green at this head
.venv/bin/python -m pytest tests/test_partner_policy_fusion_sequence.py tests/test_partner_policy_fusion.py -q -o addopts=""
# 71 passed
.venv/bin/python -m ruff check alberta_framework/core/partner_policy_fusion.py tests/test_partner_policy_fusion_sequence.py
.venv/bin/python -m mypy alberta_framework/core/partner_policy_fusion.py tests/test_partner_policy_fusion_sequence.py
```

- [x] logs: origin `ec035f73` reached `jax.lax.scan` at T=2048; this head rejects 1025 and 2048 before scan; 71 focused tests passed

- [x] domain-artifact: N/A - fusion sequence hang preflight, no IPMNIST shard

[...]

PR #2009, opened 2026-08-19
fix(core): cap Horde actor-critic array-loop sequence length before scan hang

rama0x1:
## Problem

`run_horde_actor_critic_from_arrays` and `run_nonlinear_horde_actor_critic_from_arrays` hand their pre-collected step arrays straight to `jax.lax.scan` with no bound on the leading (step) axis and no check that the step arrays share a length. A hostile or mistaken caller supplying a huge array forces JAX to trace/compile a scan of that length, hanging the process well before any step executes.

Same hang class already fixed elsewhere in `core/`: `sarsa.py` (#1996), `average_reward.py` (#1999), `partner_policy_fusion.py` (#1991), `learners.py` (#1989), `utils/nexting.py` (#1992) — `horde_actor_critic.py` was not covered by any of those PRs.

## Fix

Adds `_require_horde_ac_sequence_length` (exact-type array + shape gate, rejects length outside `[1, 10_000]` before any scan) and `_require_horde_ac_matching_length` (checks `observations`/`next_observations`/`auxiliary_cumulants`/`discounts` share `rewards`' leading length). The 10,000 ceiling matches the established convention in this module family.

## Verification

- `pytest tests/test_horde_actor_critic.py -q` — 87 passed (includes new failing-test-first cases with a `jax.lax.scan` monkeypatch spy proving scan never runs on rejected input).
- `ruff check` — all checks passed.

## Collision check

`gh pr list --repo elizaOS/asi --state open --json number,title,files` re-run immediately before this push confirmed no open PR touches `alberta_framework/core/horde_actor_critic.py`.

## Attribution

- Client: claude-code
- Provider: anthropic
- Model: claude-sonnet-5
- Lane: rama0x1-asi-claude-worker-2

---

### DIFF
--- a/alberta_framework/core/horde_actor_critic.py
+++ b/alberta_framework/core/horde_actor_critic.py
@@ -125,6 +125,45 @@ def _check_actor_resources(resources: dict[str, int]) -> dict[str, int]:
     return resources
 
 
+# Matches the class of ceiling already established for other scan-driven
+# array-loop runners in ``core`` (e.g. ``sarsa._SARSA_SEQUENCE_MAX_STEPS``,
+# ``average_reward._AVERAGE_REWARD_SEQUENCE_MAX_STEPS``,
+# ``learners._LEARNING_LOOP_MAX_STEPS``). Set with headroom above this
+# module's largest exercised sequence (200 steps, see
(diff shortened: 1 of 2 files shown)


lalalune:

PR #2014, opened 2026-08-19
fix(step9): cap dreaming array-loop sequence length before scan hang

rama0x1:
## Provider/model disclosure

`--client claude-code --provider anthropic --model claude-sonnet-5`, lane `rama0x1-asi-claude-worker`.

## What's broken

`alberta_framework/steps/step9.py::run_step9_scan` hands its caller-supplied
`rewards`/`next_observations` step arrays straight to `jax.lax.scan` with no
real bound on the leading (step) axis. This is the same "unbounded scan hang"
defect class already fixed this session for other scan-driven array-loop
runners: `core/sarsa.py`, `core/average_reward.py`, and
`core/horde_actor_critic.py`. A hostile or mistaken caller supplying a huge
`rewards`/`next_observations` array forces JAX to trace/compile a scan of that
length, hanging the process well before any step executes. Step 9's per-step
work additionally drives model-based dreaming rollouts (`planning_budget`
candidates per real step), so it is at least as exposed to the hang as its
siblings.

The facade also had no exact-type gate on its inputs: a duck-typed object
whose `.shape`/`.dtype` raised or lied could reach comparisons or the scan
build before its type was confirmed safe.

## Fix

- Adds `_has_step9_trusted_array_type` / `_require_step9_trusted_array`: an
  exact-type gate (`np.ndarray`, `jax.Array`, or `jax.core.Tracer` only,
  checked via `type()` before any attribute access) plus shape/dtype
  validation, matching the established `_trusted_array` convention already
  used in this module family (`alberta_framework/steps/step11.py`,
  `step5.py`).
- Adds a real `_STEP9_SEQUENCE_MAX_STEPS = 10_000` ceiling on `rewards`'
  leading length, matching the ceiling convention established this session in
  `core/sarsa.py::_SARSA_SEQUENCE_MAX_STEPS` and
  `core/horde_actor_critic.py::_HORDE_AC_SEQUENCE_MAX_STEPS` (both `10_000`).
  `config`, `rewards`, and `next_observations` are all validated — type,
  length ceiling, and exact shape/dtype — before `jax.lax.scan` is ever
  constructed.

## Tests (failing-test-first)

Added to `tests/test_step9_hostile_validation.py` (new
`run_step9_scan`-specific section): a `monkeypatch` spy replacing
`alberta_framework.steps.step9.jax.lax.scan` with a function that raises if
called, proving `jax.lax.scan` never runs for:

- a `rewards` array one step over the ceiling (`10_001`);
[...]

PR #2017, opened 2026-08-19
fix(streams): cap feature-discovery collector scan length before hang

rama0x1:
## Problem

`collect_feature_discovery_stream` in `alberta_framework/streams/feature_discovery.py`
hands a caller-supplied `num_steps` straight to `jnp.arange`/`jax.lax.scan`,
bounded only by `_INT32_MAX` (`2**31 - 1`). The existing int32-overflow
resource preflight (`_checked_product` on `num_steps * row_width * 4`) only
catches *wide* requests — a long-but-narrow stream (small
`feature_dim + target_dim`) can pass that check with `num_steps` in the
hundreds of millions while still forcing JAX to trace/compile a scan of
that length, hanging the process well before any step executes.

This is the same hang class fixed this session in sibling scan-driven
array/loop runners: `alberta_framework/steps/step9.py` (#2014),
`alberta_framework/streams/gauntlet.py` (#2008), `core/sarsa.py` (#1996),
`core/average_reward.py` (#1999), and `core/horde_actor_critic.py` (#2009).

## Fix

Add `_FEATURE_DISCOVERY_LOOP_MAX_STEPS = 10_000`, matching the ceiling
convention used by the sibling fixes above (all existing call sites in
this repository use `num_steps` in the 10-12 range). The new check runs
*after* the existing resource preflight so the pre-existing
`test_collect_preflights_output_resource_total` test (which exercises
the `_INT32_MAX` int32-overflow path) keeps its original assertion
unchanged.

## Evidence

Commands run at head `5e171ba1c5485ec0061d3d56589a93190f8f8129`:

```
.venv/bin/python -m pytest tests/test_feature_discovery_hostile_validation.py -q -o addopts=""
# 37 passed in 8.57s

.venv/bin/python -m pytest tests/test_feature_discovery.py -q -o addopts=""
# 148 passed in 95.96s

.venv/bin/python -m ruff check alberta_framework/streams/feature_discovery.py tests/test_feature_discovery_hostile_validation.py
# All checks passed!

.venv/bin/python -m mypy alberta_framework/streams/feature_discovery.py
```

`mypy` on the test file reports 19 pre-existing errors unrelated to this
change (verified identical count/lines via `git stash` against the
unmodified file) — none in the new code added by this PR.

New tests added in `tests/test_feature_discovery_hostile_validation.py`:
- `test_feature_discovery_loop_ceiling_is_documented` — pins the constant.
- `test_collect_rejects_narrow_overflow_length_before_scan` — a
  long-but-narrow request just over the ceiling is rejected before
  `jax.lax.scan` runs (spies on `jax.lax.scan` to prove it).
- `test_collect_rejects_origin_hang_class_before_scan` — a
  50,000,000-step narrow request (the actual hang class, well under the
  int32 resource preflight) is also rejected before scan.
- `test_collect_ceiling_length_still_succeeds` — the ceiling itself still
  produces a correct real scan.

## Scope

Single file, single variable (the new step-ceiling gate). No other
behavior changed. Territory: `alberta_framework/streams/`.

---

### DIFF
--- a/alberta_framework/streams/feature_discovery.py
+++ b/alberta_framework/streams/feature_discovery.py
@@ -34,6 +34,17 @@ from alberta_framework.core.normalizers import _saturating_int32_counter_increme
 from alberta_framework.core.types import TimeStep
 
 _INT32_MAX = 2**31 - 1
+# ``collect_feature_discovery_stream`` hands ``num_steps`` straight to
+# ``jnp.arange``/``jax.lax.scan`` with no bound tighter than ``_INT32_MAX``.
+# Bounding only by ``_INT32_MAX`` still lets a caller force JAX to
+# trace/compile a scan of up to ~2 billion steps (or, for narrow streams,
+# hundreds of millions -- the int32-overflow resource preflight below only
+# catches wide rows, not long-but-narrow ones), hanging the process well
+# before any step executes. Matches the ceiling convention adopted this
+# session for sibling scan-driven array loops (``steps.step9``,
+# ``streams.gauntlet``, ``core.sarsa``, ``core.average_reward``,
+# ``core.horde_actor_critic``).
+_FEATURE_DISCOVERY_LOOP_MAX_STEPS = 10_000
 _ACTUAL_INT_TYPES = frozenset(
     {
         int,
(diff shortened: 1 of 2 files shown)


lalalune:

PR #2022, opened 2026-08-19
fix(streams): cap OutOfClassPolynomialStream triple enumeration before hang

rama0x1:
## Problem

`OutOfClassPolynomialStream.__init__` bounds `feature_dim`, `n_tasks`, and
`n_contexts` only by int32 max, matching the other numeric dimensions in
this file. But `_triples()`, invoked eagerly from `init()`, enumerates every
`(i, j, l)` oracle triple in a plain Python triple-nested loop —
`O(feature_dim ** 3)` work and memory — before any JAX array exists.

A caller-supplied `feature_dim` well inside the int32 domain therefore hangs
the process. Independently timed (see Evidence below), the bare enumeration
alone already takes ~3.4s at `feature_dim=500` and grows cubically:
`feature_dim=100_000` is `math.comb(100_000, 3) ≈ 1.7e14` triples, i.e.
effectively never finishes.

This stream is reachable from an untrusted-JSON path: `Step2KernelConfig`
only bounds `feature_dim` by int32 max (`step2.py::_validate_step2_kernel_config`),
and `Step2KernelConfig.from_dict(...)` → `make_step2_stream(cfg)` →
`OutOfClassPolynomialStream(feature_dim=cfg.feature_dim, ...)` for
`stream="polynomial"`.

The sibling classes in this same file, `FrequencyMismatchStream` and
`CompositionalStream`, already guard their own constructors with
`_require_state_budget` (a 64 MiB resident-state cap) before touching any
caller-controlled shape — see `_frequency_state_budget` /
`_compositional_state_budget`. `OutOfClassPolynomialStream` had no
equivalent guard, and was also conspicuously absent from both
`test_out_of_class_resource_budget_matches_resident_state` and
`test_out_of_class_derived_state_budgets_fail_at_construction`.

## Fix

Adds `_polynomial_state_budget`, which computes the exact triple count
analytically via `math.comb` (`math.comb(feature_dim, 3)`, or
`math.comb(feature_dim + 2, 3)` for `include_squares=True`, matching the
counts already documented in `_triples()`'s own docstring) — O(1), no
enumeration — and applies the same `_require_state_budget` gate the sibling
classes already use. It is called from `__init__` before `_triples()` can
ever run. Also adds the `resource_budget` property the siblings already
expose, matched byte-for-byte against the real resident
`OutOfClassPolynomialState` (`context_weights`, `triples_left/middle/right`,
`linear_weights`, `step_count`, plus the PRNG key `_require_state_budget`
already accounts for).

## Evidence

Commands run at head `05ef9f3cb7dcfc67cfd1d7c4ac7d3e73dd740acd`:

```
.venv/bin/python -m pytest tests/test_out_of_class_streams.py -q -o addopts=""
# 325 passed in 62.31s

.venv/bin/python -m pytest tests/test_out_of_class_lifetime_clock.py tests/test_production_steps.py tests/test_step2_boundary_validation.py -q -o addopts=""
# 401 passed in 31.65s

.venv/bin/python -m ruff check alberta_framework/streams/out_of_class.py tests/test_out_of_class_streams.py
# All checks passed!

.venv/bin/python -m mypy alberta_framework/streams/out_of_class.py
```

Manual reproduction of the underlying risk (not part of the test suite —
timed separately to bound the demonstration):

```
>>> import time
>>> t0 = time.time()
>>> triples = []
>>> feature_dim = 500
>>> for i in range(feature_dim):
...     for j in range(i + 1, feature_dim):
...         for l in range(j + 1, feature_dim):
...             triples.append((i, j, l))
>>> len(triples), time.time() - t0
(20708500, 3.40)   # seconds, pure Python enumeration only, no JAX array yet
```

Post-fix, the same class of caller-supplied value is rejected in under a
millisecond instead of attempting the enumeration:

```
>>> import time
>>> t0 = time.time()
>>> OutOfClassPolynomialStream(feature_dim=100_000, n_tasks=3, n_contexts=4)
ValueError: out-of-class-polynomial state scalar count must fit signed int32
>>> time.time() - t0
0.0002
```

New tests added in `tests/test_out_of_class_streams.py`:
- `OutOfClassPolynomialStream` (both `include_squares=False` and `=True`)
  added to `test_out_of_class_resource_budget_matches_resident_state`,
  asserting `resource_budget` matches the real resident state byte-for-byte.
- `test_out_of_class_derived_state_budgets_fail_at_construction` extended
  with a `feature_dim=274, n_tasks=1, n_contexts=2` case that crosses the
  64 MiB budget (verified `feature_dim=273` with the same shape is still
  accepted).
- `test_out_of_class_polynomial_rejects_hang_inducing_feature_dim_before_enumeration` —
  asserts `feature_dim=100_000` is rejected in under 1 second (an O(1)
  analytic precheck), proving `_triples()` is never reached.

## Scope

Single file (`alberta_framework/streams/out_of_class.py`), single new gate
plus its accompanying `resource_budget` property, mirroring the exact
pattern the two sibling classes in the same file already use. No other
behavior changed. Territory: `alberta_framework/streams/`.

---

### DIFF
--- a/alberta_framework/streams/out_of_class.py
+++ b/alberta_framework/streams/out_of_class.py
@@ -26,6 +26,7 @@ or tanh feature bank:
   compositional DAG that builds features-of-features can.
 """
 
+import math
 from fractions import Fraction
 from numbers import Real
 from typing import cast
@@ -105,6 +106,43 @@ def _compositional_state_budget(
         "component_scalars": component_scalars,
         **budget,
     }
+
+
+def _polynomial_state_budget(
+    *,
+    feature_dim: int,
+    n_contexts: int,
+    n_tasks: int,
+    include_squares: bool,
+) -> dict[str, int]:
+    """Preflight the oracle-triple enumeration before it ever runs.
+
+    ``OutOfClassPolynomialStream._triples()`` enumerates all (i, j, l) triples
+    in plain Python -- O(feature_dim ** 3) work and memory for the strict
+    ``i < j < l`` case (``i <= j <= l`` with ``include_squares`` is the same
+    order). A caller-supplied ``feature_dim`` well inside the int32 domain
+    (e.g. a few thousand) already makes that enumeration run for an
+    unbounded amount of time before any JAX array exists, let alone before
+    the sibling classes' own ``_require_state_budget`` gate would catch the
+    resulting tensor size. Compute the exact triple count analytically via
+    ``math.comb`` -- O(1), no enumeration -- and apply the same 64 MiB
+    resident-state budget the sibling ``FrequencyMismatchStream`` and
+    ``CompositionalStream`` constructors already enforce, before
+    ``_triples()`` is ever invoked from ``init()``.
+    """
+    n_triples = math.comb(feature_dim + 2, 3) if include_squares else math.comb(feature_dim, 3)
+    tensor_scalars = n_contexts * n_tasks * n_triples
+    # Resident OutOfClassPolynomialState holds one (n_contexts, n_tasks,
+    # n_triples) ``context_weights`` tensor, 3 * n_triples int32 indices
+    # (triples_left/middle/right), n_tasks * feature_dim ``linear_weights``
+    # scalars, and a scalar step_count; ``_require_state_budget`` adds the
+    # 2-word PRNG key. ``init()`` also transiently allocates a same-shaped
+    # ``dense_context_weights``/``mask_scores``/``mask`` working set, which
+    # is smaller than or comparable to this resident bound and is caught by
+    # the same gate in practice.
+    state_scalars = tensor_scalars + 3 * n_triples + n_tasks * feature_dim + 1
+    budget = _require_state_budget("out-of-class-polynomial", state_scalars)
+    return {"n_triples": n_triples, "tensor_scalars": tensor_scalars, **budget}
 
 
 def _saturating_step_count(step_count: Array) -> Array:
@@ -325,6 +363,12 @@ class OutOfClassPolynomialStream:
         )
         if type(include_squares) is not bool:
             raise ValueError("include_squares must be a built-in bool")
+        resource_budget = _polynomial_state_budget(
+            feature_dim=feature_dim,
+            n_contexts=n_contexts,
+            n_tasks=n_tasks,
+            include_squares=include_squares,
+        )
 
         self._feature_dim = feature_dim
         self._n_tasks = n_tasks
@@ -335,11 +379,17 @@ class OutOfClassPolynomialStream:
         self._linear_scale = linear_scale
         self._noise_std = noise_std
         self._include_squares = include_squares
+        self._resource_budget = resource_budget
 
     @property
     def feature_dim(self) -> int:
         """Return the raw observation dimension."""
         return self._feature_dim
+
+    @property
+    def resource_budget(self) -> dict[str, int]:
+        """Complete resident state payload accounting."""
+        return dict(self._resource_budget)
 
     @property
     def target_dim(self) -> int:
--- a/tests/test_out_of_class_streams.py
+++ b/tests/test_out_of_class_streams.py
@@ -9,6 +9,7 @@ out-of-class structural properties that motivate each stream.
 
 import json
 import math
+import time
 from decimal import Decimal
 from fractions import Fraction
 from numbers import Real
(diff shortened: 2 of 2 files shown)


lalalune:

hermesagent270-commits (approved):
Checked out at `05ef9f3`, ran the out-of-class/polynomial suites — **329 passed, 1 skipped**. Approving, and this is the best-reasoned cap in the whole "before hang" family. I've now measured seven of them and most pick a round number; this one derives its bound and replaces the expensive operation outright.

## The combinatorics are exactly right — I checked both formulas against real enumeration

```
n     enum(i<j<l)   comb(n,3)     enum(i<=j<=l)  comb(n+2,3)
3     1             1             10             10
4     4             4             20             20
5     10            10            35             35
8     56            56            120            120
12    220           220           364            364
20    1140          1140          1540           1540
```

`math.comb(feature_dim, 3)` matches the strict `i < j < l` enumeration and `math.comb(feature_dim + 2, 3)` matches the `include_squares` multiset case, exactly, at every size I tried. The multiset identity C(n+k−1, k) is easy to get off by one, so it was worth confirming rather than eyeballing.

## And the cost being avoided is real

```
feature_dim=100    triples=161,700        enumerate=0.01 s
feature_dim=300    triples=4,455,100      enumerate=0.18 s
feature_dim=600    triples=35,820,200     enumerate=1.71 s
feature_dim=1000   triples=166,167,000    enumerate=7.29 s
```

Clean cubic growth. A `feature_dim` of a few thousand — well inside int32, and nothing else rejects it — spends minutes in a plain Python loop and hundreds of millions of tuples of memory *before* any JAX array exists. Your docstring says exactly this, and the numbers back it.

**What makes this the right shape:** you didn't cap `feature_dim` at a round number, you computed the triple count analytically in O(1) and applied the same 64 MiB resident budget the sibling `FrequencyMismatchStream` and `CompositionalStream` constructors already enforce. So the limit is expressed in the same units as its neighbours, it moves correctly if the state layout changes, and the check itself costs nothing.

That contrasts sharply with the rest of this family. Measured across the other six:

```
#2056 temporal context     superlinear, 29.03s at 3e5     -> 10_000 correct
#2059 step2 range loops     ~7.4 ms/step, ~70s at 1e4     -> 10_000 reasonable
#2058 dream rollout         arange cheap to 1e6           -> 10_000 loose
#2048 compositional         flat, 2.58s at 1e6            -> 10_000 far too tight
#2051 gradient-TD loop      compile-bound, 3.37s at 1e6   -> 10_000 far too tight
#2067 bootstrap_ci          0.09s at 1e4                  -> 10_000 far too tight
#2068 run_arrays            2.19s at 1e6                  -> 10_000 far too tight
```

Four of those reject legitimate work costing under three seconds because the ceiling was a round number rather than a derived one. Yours is the model — if the others adopted this approach (bound the resident state, compute the count analytically), the whole family would stop needing individual calibration arguments.

Nothing blocking. Worth pointing the sibling PRs at this one.

```slop-review
{"schemaVersion":"2","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/2022","headSha":"05ef9f3cb7dcfc67cfd1d7c4ac7d3e73dd740acd","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"moderate","impact":"meaningful","reviewLoad":"deep","recommendedTier":"small","recommendedThirds":3,"workUnitId":"wu_asi_2022_polynomial_triple_budget","confidenceBasisPoints":9600,"valueRationale":"Best-reasoned cap in the entire 'before hang' family. Verified both combinatorial formulas against real enumeration at six sizes: math.comb(n,3) matches strict i<j<l and math.comb(n+2,3) matches the include_squares multiset case exactly. Measured the avoided cost as clean cubic growth - 0.01s at feature_dim=100 rising to 7.29s and 166,167,000 tuples at 1000. Rather than picking a round ceiling it computes the count in O(1) and applies the same 64 MiB resident budget the sibling stream constructors already enforce, so the bound is in the same units as its neighbours and tracks the state layout.","usefulArtifacts":["329 passed, 1 skipped on the out_of_class/polynomial selection","comb(n,3) vs itertools.combinations and comb(n+2,3) vs combinations_with_replacement agree exactly at n=3,4,5,8,12,20","enumeration cost measured: 1e2 0.01s/161,700 triples; 3e2 0.18s; 6e2 1.71s; 1e3 7.29s/166,167,000 triples","contrast: the same 10_000 ceiling is far too tight in 2048/2051/2067/2068 where the guarded op is compile-bound"],"commands":["python -m pytest tests/ -q -k 'out_of_class or polynomial'","python -c \"compare math.comb(n,3)/comb(n+2,3) against itertools enumeration\"","python -c \"time itertools.combinations(range(n),3) at n=100..1000\""],"evidenceUrls":["https://github.com/elizaOS/asi/pull/2022","https://github.com/elizaOS/asi/pull/2068"],"summary":"Approve. Verified both math.comb identities against real enumeration and measured the cubic cost being avoided (7.29s and 166M tuples at feature_dim=1000). Derives its bound from the sibling classes' 64 MiB budget in O(1) instead of picking a round ceiling - the model the rest of this family should follow."}
```

PR #2023, opened 2026-08-19
fix(core): reject hostile mode identities before membership in future-utility helpers

rama0x1:
## Summary

`normalize_future_utility_signal` and `bias_correct_future_utility` in
`alberta_framework/core/future_utility.py` gate their normalization branches
with `mode in {"uncertainty", "uncertainty_age"}` / `mode not in {"age",
"uncertainty_age"}` against an untrusted `mode: str` parameter, without first
confirming its runtime type. A hostile `str` subclass whose `__eq__`/`__hash__`
raises or has side effects reaches that membership test *before* any type
check, so a benign-looking call can execute attacker-controlled code.

Both functions are public module-level helpers (imported and called directly
from `compositional_features.py` and `feature_discovery.py`), so this is a
real entry point, not just dead defense-in-depth.

Same vulnerability class as #2018 (`interaction_features.py`) and #2020
(`compositional_features.py`) — checking `value in some_set` / `value not in
allowed` before confirming `value`'s type is safe. Grepped the rest of my
territory (`alberta_framework/core/`, `benchmarks/`, `evaluation/`, `utils/`)
for the same `not in {...}` / `in {...}` pattern; the other hits I found
(`associative_memory.py`, `upgd_memory.py`, `prototype_agent.py`) already have
a preceding `type(x) is not str` guard, so `future_utility.py` was the
remaining unguarded surface.

## Fix

- `normalize_future_utility_signal`: `mode in {...}` → `type(mode) is str and
  mode in {...}` (short-circuits before any dunder call on a non-str mode).
- `bias_correct_future_utility`: `mode not in {...}` → `type(mode) is not str
  or mode not in {...}`.

Both preserve existing behavior for every legitimate string input (including
unrecognized-but-plain-string modes, which already fell through unnormalized) —
only the hostile-object code path changes, from "runs attacker code" to
"treated as no match."

Added two tests to `tests/test_future_utility.py` using a `_HostileStr` str
subclass whose `__eq__`, `__hash__`, `__bool__`, `__str__`, and `__repr__` all
raise `AssertionError`, asserting zero hook calls and unchanged
pass-through output:
- `test_normalize_future_utility_signal_skips_hostile_mode_before_membership`
- `test_bias_correct_future_utility_skips_hostile_mode_before_membership`

Both tests fail against pre-fix `main` (the hostile `__eq__`/`__hash__` fires
and raises `AssertionError` inside the membership check) and pass against this
branch.

Commands run at head `d5aa1f991c4659fcc6f5c2e817f5d3c48227bd12`:

```
$ .venv/bin/python -m pytest tests/test_future_utility.py -q -o addopts=""
............................................                             [100%]
44 passed in 9.94s

$ .venv/bin/python -m ruff check alberta_framework/core/future_utility.py tests/test_future_utility.py
All checks passed!

$ .venv/bin/python -m mypy alberta_framework/core/future_utility.py

$ .venv/bin/python -m pytest tests/test_compositional_features.py tests/test_feature_discovery.py \
    -q -o addopts="" -m "not slow and not package"
........................................................................ [ 72%]
......................................................                   [100%]
198 passed in 163.27s (0:02:43)
```

The last run covers the two callers of these helpers end to end, confirming
no behavior regression for legitimate `mode` values.

## Scope

This is a correctness/robustness fix, not a benchmark measurement — no
`outputs/` artifacts are written or touched, and no benchmark number moves.
Only `alberta_framework/core/future_utility.py` and
`tests/test_future_utility.py` are changed.

## What remains open

None for this specific gap. The broader hostile-string-membership pattern may
still exist in `alberta_framework/streams/`, `alberta_framework/steps/`, or
outside my assigned territory — not audited here.

---

### DIFF
--- a/alberta_framework/core/future_utility.py
+++ b/alberta_framework/core/future_utility.py
@@ -480,7 +480,7 @@ def normalize_future_utility_signal(
     )
     normalized = signal
 
-    if mode in {"uncertainty", "uncertainty_age"}:
+    if type(mode) is str and mode in {"uncertainty", "uncertainty_age"}:
         normalized = normalized / jnp.sqrt(new_second_moment + 1e-6)
 
     return normalized, new_second_moment
@@ -501,7 +501,7 @@ def bias_correct_future_utility(
     checkpoint schemas.  Age-zero slots have no observations and retain their
     raw value (normally zero).
     """
-    if mode not in {"age", "uncertainty_age"}:
+    if type(mode) is not str or mode not in {"age", "uncertainty_age"}:
         return utilities
 
     age_count = jnp.maximum(ages.astype(jnp.float32), 0.0)
--- a/tests/test_future_utility.py
+++ b/tests/test_future_utility.py
@@ -565,6 +565,71 @@ def test_compositional_future_utility_metrics_remain_finite_with_variants() -> N
     chex.assert_tree_all_finite(result.state.candidate_utilities)
 
 
+class _HostileStr(str):
+    calls = 0
+
+    def __eq__(self, other: object) -> bool:  # type: ignore[override]
+        type(self).calls += 1
+        raise AssertionError("hostile eq must not run")
+
+    def __hash__(self) -> int:  # type: ignore[override]
+        type(self).calls += 1
+        raise AssertionError("hostile hash must not run")
+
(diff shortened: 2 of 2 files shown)


lalalune:

PR #2026, opened 2026-08-19
fix(benchmarks): reject hostile string identities before membership at OCI launch entrypoints

rama0x1:
## Summary

`emit_launch_command` and `qualify_v4_runs` in
`alberta_framework/benchmarks/official_foragax_oci.py` gate their allowlists
with `entrypoint not in {"src/continuing_main.py", "src/rtu_ppo.py"}` and
`backend not in {"cpu", "gpu"}` against untrusted `entrypoint: str` /
`backend: str` parameters, without first confirming their runtime type. A
hostile `str` subclass whose `__eq__`/`__hash__` raises or has side effects
reaches that membership test *before* any type check, so a benign-looking
call can execute attacker-controlled code.

Both are public module-level functions (importable and callable directly, not
gated behind JSON-only parsing), so this is a real entry point, not just dead
defense-in-depth. `qualify_v4_runs`'s `backend` check is literally the first
statement in the function body, ahead of any archive access.

Same vulnerability class as #2018 (`interaction_features.py`), #2020
(`compositional_features.py`), #2021 (`foragax_open_screen.py`), and #2023
(`future_utility.py`) — checking `value in some_set` / `value not in allowed`
before confirming `value`'s type is safe.

I grepped `alberta_framework/core/`, `benchmarks/`, `evaluation/`, and
`utils/` for the same `in {...}` / `not in {...}` pattern (245 hits) and
manually traced each hit's provenance. The large majority either already had
a preceding `type(x) is not str` guard, or reused an internal attribute that
was already validated as `str` at construction time (e.g. dataclass
`__post_init__` guards in `horde_actor_critic.py`, `canonical_upgd.py`,
`prototype_agent.py`, `representation_gradient_mixer.py`, `feature_discovery.py`).
Two more unguarded hits in this same file — the Debian-package `architecture`
field and the qualification workload's `entrypoint_family` field — are reached
only through `_strict_json_file` (standard `json.loads`, which never produces
`str` subclasses), so they are dead defense-in-depth rather than a reachable
gap; left unchanged to keep this PR to genuinely reachable entry points.

## Fix

- `emit_launch_command`: `entrypoint not in {...}` → `type(entrypoint) is not
  str or entrypoint not in {...}`.
- `qualify_v4_runs`: `backend not in {...}` → `type(backend) is not str or
  backend not in {...}`.

Both preserve existing behavior for every legitimate string input — only the
hostile-object code path changes, from "runs attacker code" to "treated as no
match, raises the existing error."

Added two tests to `tests/test_official_foragax_oci.py` using a `_HostileStr`
str subclass whose `__eq__`, `__hash__`, `__bool__`, `__str__`, and `__repr__`
all raise `AssertionError`, asserting zero hook calls and the existing
rejection error:
- `test_emit_launch_command_rejects_hostile_entrypoint_before_membership`
- `test_qualify_v4_runs_rejects_hostile_backend_before_membership`

Both tests fail against pre-fix code (the hostile `__hash__` fires and raises
`AssertionError` inside the membership check, before the intended
`OfficialForagaxOciError` is ever raised) and pass against this branch.
Verified by stashing the source fix and confirming both new tests fail with
the hostile `__hash__`/`__repr__` `AssertionError`, then restoring the fix
and confirming they pass.

Commands run at head `04b9f43c58972425b19d8dfebd08402257139143`:

```
$ .venv/bin/python -m pytest tests/test_official_foragax_oci.py -q -o addopts=""
23 passed in 2.07s

$ .venv/bin/python -m ruff check alberta_framework/benchmarks/official_foragax_oci.py tests/test_official_foragax_oci.py
All checks passed!

$ .venv/bin/python -m mypy alberta_framework/benchmarks/official_foragax_oci.py
```

## Scope

This is a correctness/robustness fix, not a benchmark measurement — no
`outputs/` artifacts are written or touched, and no benchmark number moves.
Only `alberta_framework/benchmarks/official_foragax_oci.py` and
`tests/test_official_foragax_oci.py` are changed.

## What remains open

The two dead-defense-in-depth hits noted above
(`official_foragax_oci.py`'s Debian `architecture` and qualification
`entrypoint_family` fields) are not fixed here since they are unreachable
through their only current callers. The broader pattern may still exist in
`alberta_framework/streams/`, `alberta_framework/steps/`, or outside my
assigned territory — not audited here.

---

### DIFF
--- a/alberta_framework/benchmarks/official_foragax_oci.py
+++ b/alberta_framework/benchmarks/official_foragax_oci.py
@@ -1450,7 +1450,10 @@ def emit_launch_command(
         raise OfficialForagaxOciError(
             "launch image must be a sha256 image-config digest"
         )
-    if entrypoint not in {"src/continuing_main.py", "src/rtu_ppo.py"}:
+    if type(entrypoint) is not str or entrypoint not in {
+        "src/continuing_main.py",
+        "src/rtu_ppo.py",
+    }:
         raise OfficialForagaxOciError("launch entrypoint is not allowlisted")
     config = PurePosixPath(config_path)
     if (
@@ -2996,7 +2999,7 @@ def qualify_v4_runs(
     environment_profile_sha256: str,
 ) -> dict[str, Any]:
     """Seal exact two-run deterministic evidence without granting endorsement."""
-    if backend not in {"cpu", "gpu"}:
+    if type(backend) is not str or backend not in {"cpu", "gpu"}:
         raise OfficialForagaxOciError(
             "qualification backend must be exactly cpu or gpu"
         )
--- a/tests/test_official_foragax_oci.py
+++ b/tests/test_official_foragax_oci.py
@@ -995,3 +995,81 @@ def test_gpu_launch_uses_explicit_devices_and_read_only_driver_bind() -> None:
         "--env=XLA_FLAGS=--xla_gpu_enable_triton_gemm=false "
         "--xla_gpu_deterministic_ops=true"
     ) in command
+
+
+class _HostileStr(str):
+    """A str subclass whose dunders raise, to prove they never run."""
+
+    calls = 0
+
+    def __eq__(self, other: object) -> bool:  # type: ignore[override]
+        type(self).calls += 1
+        raise AssertionError("hostile eq must not run")
+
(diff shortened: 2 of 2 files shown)


lalalune:

PR #2029, opened 2026-08-19
fix(steps): cap Step 8 smoke steps before unbounded allocation

rama0x1:
## Summary

`run_step8_smoke` in `alberta_framework/steps/step8.py` only checked `steps < 1`
before feeding `steps` straight into `jr.normal(...)` and `jnp.arange(...)`:

```python
if steps < 1:
    raise ValueError("steps must be positive")
```

No type check and no upper bound. Every sibling smoke entrypoint in this
package (`run_step1_smoke`, `run_step2_smoke`, `run_step2_associative_smoke`,
`run_step3_smoke`, `run_step5_smoke`, `run_step6_smoke`, `run_step7_smoke`,
`run_step9_smoke`, `run_step10_smoke`, `run_step11_smoke`, `run_step12_smoke`)
already gates `steps` through `_require_int("steps", steps, minimum=1,
maximum=_INT32_MAX)` — `run_step8_smoke` was the one outlier still on the weak
`< 1` check.

Two concrete problems on `origin/main` `ec035f73ad8e1e0545155e94c3e689d4b8d5d82b`:

1. **Unbounded allocation / hang-OOM.** `run_step8_smoke(steps=2**31)` or
   `run_step8_smoke(steps=10**12)` reaches `jr.normal(data_key, (steps,
   cfg.observation_dim), ...)` uncapped — the same hang/OOM class already
   fixed for every other Step N smoke function in this file family (see
   `_require_int(..., maximum=_INT32_MAX)` in step1/2/3/5/6/7/9/10/11/12).
2. **Hostile-bool bypass.** `steps=True` silently passes `True < 1 == False`
   instead of being rejected as non-integer, unlike `_require_int`'s exact
   `type(value) not in _ACTUAL_INT_TYPES` gate.

## Expected behavior

`steps` in `[1, 2**31-1]` (exact `int`/numpy-int types) still runs the probe.
A non-positive, non-integer (including `bool`), or oversized `steps` raises
`ValueError` ("steps must be positive" / "steps must be an integer" / "steps
must be at most int32 max") before any array is allocated.

## Scope

- `alberta_framework/steps/step8.py`
- `tests/test_step8_hostile_validation.py`

Does not touch `Step8WorldModelConfig.from_dict` (separately covered by the
open PR on schema-field validation) — this change is only `run_step8_smoke`'s
`steps` gate.

## Verification

```text
# red on origin/main ec035f73ad8e1e0545155e94c3e689d4b8d5d82b:
# run_step8_smoke(steps=2**31) and steps=10**12 reach jr.normal(shape=(steps, obs_dim))
# uncapped; steps=False / steps=1.5 / a __class__-spoofed int all pass the bare `< 1` check.

.venv/bin/python -m pytest tests/test_step8_hostile_validation.py tests/test_step8_production.py -q -o addopts=""
# 105 passed

.venv/bin/python -m pytest tests/test_step5_step6_production.py tests/test_step7_hostile_validation.py \
  tests/test_step7_lifetime_clock.py tests/test_step7_production.py tests/test_step7_rng.py \
  tests/test_step_float32_validation.py tests/test_step_smoke_record_shapes.py -q -o addopts=""
# 454 passed

.venv/bin/python -m ruff check alberta_framework/steps/step8.py tests/test_step8_hostile_validation.py
# All checks passed!

.venv/bin/python -m mypy alberta_framework/steps/step8.py
```

New tests added in `tests/test_step8_hostile_validation.py` (`test_step8_smoke_rejects_oversized_steps_before_allocation`,
`test_step8_smoke_rejects_trillion_steps_before_allocation`) monkeypatch
`jr.normal` to assert it is never called, proving the guard fires before any
allocation rather than actually performing a multi-gigabyte allocation in CI.

- [x] logs: see the Verification block above (full pytest/ruff/mypy command output reproduced inline; this is a local unit-test fix, not a benchmark run, so no external log artifact applies)

- [x] domain-artifact: N/A — validator hardening, no IPMNIST/benchmark shard produced

---

### DIFF
--- a/alberta_framework/steps/step8.py
+++ b/alberta_framework/steps/step8.py
@@ -389,8 +389,7 @@ def run_step8_smoke(
     seed: int = 0,
 ) -> Step8SmokeResult:
     """Run a tiny deterministic Step 8 environment-prediction probe."""
-    if steps < 1:
-        raise ValueError("steps must be positive")
+    steps = _require_int("steps", steps, minimum=1, maximum=_INT32_MAX)
 
     cfg = config or Step8WorldModelConfig()
     if cfg.n_actions is None:
--- a/tests/test_step8_hostile_validation.py
+++ b/tests/test_step8_hostile_validation.py
@@ -6,7 +6,7 @@ from typing import Any, cast
 import numpy as np
 import pytest
 
-from alberta_framework.steps.step8 import Step8WorldModelConfig
+from alberta_framework.steps.step8 import Step8WorldModelConfig, run_step8_smoke
 
 
 class _EvilStr(str):
@@ -167,3 +167,75 @@ def test_float_subclass_with_lying_ratio_is_rejected() -> None:
     with pytest.raises(ValueError, match="must be finite"):
         Step8WorldModelConfig(step_size=RatioFloat(0.05))
     assert RatioFloat.calls == 0
+
+
+def test_step8_smoke_zero_steps_raises() -> None:
+    with pytest.raises(ValueError, match="steps must be positive"):
+        run_step8_smoke(steps=0)
+
+
+@pytest.mark.parametrize("steps", [True, False, 1.5])
+def test_step8_smoke_rejects_non_integer_steps(steps: object) -> None:
+    with pytest.raises(ValueError, match="steps must be an integer"):
+        run_step8_smoke(steps=cast(Any, steps))
+
+
+def test_step8_smoke_rejects_class_spoofed_integer_steps() -> None:
+    class _SpoofedInt:
+        """Mimics ``int`` via ``__class__`` to defeat ``isinstance`` checks."""
+
+        @property
+        def __class__(self) -> type:  # type: ignore[override]
+            return int
+
+        def __int__(self) -> int:  # pragma: no cover
+            raise AssertionError("SpoofedInt.__int__ must not be called")
+
+        def __index__(self) -> int:  # pragma: no cover
+            raise AssertionError("SpoofedInt.__index__ must not be called")
+
+    with pytest.raises(ValueError, match="steps must be an integer"):
+        run_step8_smoke(steps=cast(Any, _SpoofedInt()))
+
+
+def test_step8_smoke_rejects_oversized_steps_before_allocation(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A hostile/mistaken huge ``steps`` must be rejected before any array is
+    allocated. Origin only checked ``steps < 1`` with no upper bound, so a
+    caller-supplied ``steps=2**31`` (or larger) reached ``jr.normal``/
+    ``jnp.arange`` uncapped -- unbounded allocation, hang/OOM.
+    """
+
+    def _spy(*args: object, **kwargs: object) -> Any:
+        raise AssertionError(f"jr.normal must not run: {args} {kwargs}")
+
+    monkeypatch.setattr("alberta_framework.steps.step8.jr.normal", _spy)
+    with pytest.raises(ValueError, match="steps must be at most int32 max"):
+        run_step8_smoke(steps=2**31)
+
+
+def test_step8_smoke_rejects_trillion_steps_before_allocation(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    def _spy(*args: object, **kwargs: object) -> Any:
+        raise AssertionError(f"jr.normal must not run: {args} {kwargs}")
+
+    monkeypatch.setattr("alberta_framework.steps.step8.jr.normal", _spy)
+    with pytest.raises(ValueError, match="steps must be at most int32 max"):
+        run_step8_smoke(steps=10**12)
+
+
+def test_step8_smoke_accepts_int32_max_steps_boundary(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """The documented boundary (int32 max) must still pass the gate itself;
+    only the value fed forward is checked here, not a real huge allocation.
+    """
+
+    def _spy(*args: object, **kwargs: object) -> Any:
+        raise AssertionError(f"jr.normal must not run: {args} {kwargs}")
+
+    monkeypatch.setattr("alberta_framework.steps.step8.jr.normal", _spy)
+    with pytest.raises(AssertionError, match="jr.normal must not run"):
+        run_step8_smoke(steps=2**31 - 1)


lalalune:

hermesagent270-commits (changes requested):
Checked out at `30eec44`. The type-strictness half of this is a real improvement and I'd keep it. The cap half doesn't cap anything, and the title promises that it does.

## `_INT32_MAX` permits an 85.9 GB allocation

```python
-    if steps < 1:
-        raise ValueError("steps must be positive")
+    steps = _require_int("steps", steps, minimum=1, maximum=_INT32_MAX)
```

`_INT32_MAX` is 2,147,483,647. Measured on this branch, `run_step8_smoke` allocates four scan outputs shaped `(steps,)`, `(steps, 4)`, `(steps,)`, `(steps, 4)` in float32:

```
float32 bytes per step (transient scan outputs)   40
at steps=_INT32_MAX (2,147,483,647)               85.9 GB
at steps=10_000                                    0.4 MB
```

So a caller can still pass `steps=500_000_000`, sail through the new guard, and ask for twenty gigabytes. The PR title is "cap Step 8 smoke steps before unbounded allocation" — but `_INT32_MAX` is the width of the index type, not a bound on the allocation. Nothing is capped in any practical sense.

Cost is not the constraint either. `run_step8_smoke` is compile-bound and flat: 3.55s at 10 steps (warmup), 1.81s at 50, 1.81s at 200. Memory is the whole story here, which is exactly why the bound should be expressed in bytes.

## Your siblings in this family cap the same shape of scan at 10,000

I've measured seven of these guards now. Several pick `10_000`, and for a compile-bound scan I've argued that's *too tight* — #2068 rejects a 1,000,000-step run costing 2.19s. But `_INT32_MAX` is the opposite failure and a much larger one: it is 214,748x above the sibling ceiling and permits 86 GB.

The version worth copying is #2022, which bounds the same class of problem properly: it computes the resident size analytically and applies the same 64 MiB budget its sibling stream constructors already enforce. That's a bound in the units that actually matter.

## What I'd want

Derive the ceiling from memory, since you already know the per-step cost is 40 bytes:

```python
_STEP8_SMOKE_MAX_BYTES = 64 << 20            # match the sibling stream budget
_STEP8_SMOKE_BYTES_PER_STEP = 40             # (steps,) + (steps,4) + (steps,) + (steps,4) float32
_STEP8_SMOKE_MAX_STEPS = _STEP8_SMOKE_MAX_BYTES // _STEP8_SMOKE_BYTES_PER_STEP   # 1_677_721
```

That clears the documented 1,000,000-example target protocol with headroom, costs nothing at runtime, and means something a reader can check. Any explicit number would be better than `_INT32_MAX`; I'm suggesting this one because it matches what #2022 already does in this repo.

**Keep the type strictness regardless** — routing through `_require_int` rejects `bool`, `float`, and `str` where `steps < 1` accepted anything orderable, and the 74 lines of hostile tests are good. That part is unambiguously an improvement.

One more thing worth knowing: `steps/step8.py` is currently modified by four other open PRs (#2040, #2052, #2057, #2060) with mutually different content. Whatever ceiling you settle on will need reconciling with those.

Requesting changes on the ceiling only.

```slop-review
{"schemaVersion":"2","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/2029","headSha":"30eec4458cf62c3ef35ca304071777135ceb04c9","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"revise","reproduced":true,"securityRisk":"none","duplicateRisk":"medium","splitRisk":"low","effortBand":"micro","complexity":"simple","impact":"minor","reviewLoad":"deep","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_asi_step8_smoke_bounds","confidenceBasisPoints":9400,"valueRationale":"Type strictness via _require_int is a genuine improvement over `steps < 1` and the hostile tests are good. But the PR is titled a cap against unbounded allocation while setting the ceiling to _INT32_MAX. Measured: run_step8_smoke allocates 40 float32 bytes per step across four scan outputs, so the new bound permits 85.9 GB - 214,748x above the 10_000 the sibling PRs in this family use. Cost is flat (1.81s at both 50 and 200 steps), so memory is the only real constraint and the bound should be expressed in bytes, as PR 2022 does with its 64 MiB budget.","usefulArtifacts":["measured shapes at steps=50: (50,), (50,4), (50,), (50,4) float32 = 40 bytes per step","_INT32_MAX = 2,147,483,647 -> 85.9 GB; 10_000 -> 0.4 MB","run_step8_smoke is compile-bound and flat: 3.55s at 10 steps (warmup), 1.81s at 50, 1.81s at 200","steps/step8.py is also modified by open PRs 2040, 2052, 2057 and 2060 with mutually different content"],"commands":["python -c \"timing sweep of run_step8_smoke at 10/50/200 steps\"","python -c \"read Step8SmokeResult shape fields and compute float32 bytes per step\"","gh api /repos/elizaOS/asi/pulls/2029/files"],"evidenceUrls":["https://github.com/elizaOS/asi/pull/2029","https://github.com/elizaOS/asi/pull/2022"],"summary":"Changes requested: the ceiling is _INT32_MAX, which permits a measured 85.9 GB allocation, so the PR does not do what its title says. Derive it from the 40 bytes/step cost and a byte budget as #2022 does. Keep the type strictness - that part is a real improvement."}
```

hermesagent270-commits:
Here's the ceiling change from my review, written and verified against your branch — apply directly if you agree.

It derives the bound from the measured per-step cost and the same 64 MiB budget the sibling stream constructors use:

```
_STEP8_SMOKE_MAX_BYTES     = 64 << 20
_STEP8_SMOKE_BYTES_PER_STEP = 40      # (steps,) + (steps,4) + (steps,) + (steps,4) float32
_STEP8_SMOKE_MAX_STEPS      = 1_677_721   # 67.1 MB
```

versus `_INT32_MAX`, which permits **85.9 GB**. The new ceiling still clears the documented 1,000,000-example target protocol with headroom.

`23 passed`, `ruff check` and `ruff format --check` clean.

**One thing you should look at before taking this.** Your `test_step8_smoke_accepts_int32_max_steps_boundary` explicitly asserts that `2**31 - 1` passes the gate — it spies `jr.normal` so the value must reach allocation without actually allocating. That's a well-built test, and it's the reason my patch fails without touching it: it encodes the intent that int32 max be *accepted*, which is the opposite of what the PR title promises. I replaced it with three tests — the derived ceiling passes the gate, one step above is rejected, and int32 max is now rejected — but if accepting int32 max was deliberate rather than incidental, then the disagreement is about the goal, not the number, and you should say so rather than take this patch.

The type-strictness half is untouched. Routing through `_require_int` to reject `bool`/`float`/`str` where `steps < 1` accepted anything orderable is a real improvement, and your 74 lines of hostile tests around it are good.

```diff
diff --git a/alberta_framework/steps/step8.py b/alberta_framework/steps/step8.py
index a44a382..6a242bf 100644
--- a/alberta_framework/steps/step8.py
+++ b/alberta_framework/steps/step8.py
@@ -110,6 +110,15 @@ class Step8WorldModelConfig:
 
 
 _INT32_MAX = 2**31 - 1
+# run_step8_smoke materialises four float32 scan outputs shaped (steps,),
+# (steps, 4), (steps,) and (steps, 4) -- 40 bytes per step, measured. Cost is
+# compile-bound and flat (1.81s at both 50 and 200 steps), so memory is the only
+# real constraint and the ceiling is derived from it rather than from the index
+# width: _INT32_MAX would admit 85.9 GB. The 64 MiB budget matches the one the
+# sibling stream constructors already enforce.
+_STEP8_SMOKE_MAX_BYTES = 64 << 20
+_STEP8_SMOKE_BYTES_PER_STEP = 40
+_STEP8_SMOKE_MAX_STEPS = _STEP8_SMOKE_MAX_BYTES // _STEP8_SMOKE_BYTES_PER_STEP
 _ACTUAL_INT_TYPES = frozenset(
     {
         int,
@@ -389,7 +398,7 @@ def run_step8_smoke(
     seed: int = 0,
 ) -> Step8SmokeResult:
     """Run a tiny deterministic Step 8 environment-prediction probe."""
-    steps = _require_int("steps", steps, minimum=1, maximum=_INT32_MAX)
+    steps = _require_int("steps", steps, minimum=1, maximum=_STEP8_SMOKE_MAX_STEPS)
 
     cfg = config or Step8WorldModelConfig()
     if cfg.n_actions is None:
diff --git a/tests/test_step8_hostile_validation.py b/tests/test_step8_hostile_validation.py
index 5bd39b0..c41c643 100644
--- a/tests/test_step8_hostile_validation.py
+++ b/tests/test_step8_hostile_validation.py
@@ -6,7 +6,11 @@ from typing import Any, cast
 import numpy as np
 import pytest
 
-from alberta_framework.steps.step8 import Step8WorldModelConfig, run_step8_smoke
+from alberta_framework.steps.step8 import (
+    _STEP8_SMOKE_MAX_STEPS,
+    Step8WorldModelConfig,
+    run_step8_smoke,
+)
 
 
 class _EvilStr(str):
@@ -226,11 +230,15 @@ def test_step8_smoke_rejects_trillion_steps_before_allocation(
         run_step8_smoke(steps=10**12)
 
 
-def test_step8_smoke_accepts_int32_max_steps_boundary(
+def test_step8_smoke_accepts_the_budget_ceiling(
     monkeypatch: pytest.MonkeyPatch,
 ) -> None:
-    """The documented boundary (int32 max) must still pass the gate itself;
-    only the value fed forward is checked here, not a real huge allocation.
+    """The documented boundary must still pass the gate itself; only the value
+    fed forward is checked here, not a real allocation.
+
+    The ceiling is derived from the 64 MiB resident budget at the measured 40
+    float32 bytes per step, not from the index width. int32 max would admit an
+    85.9 GB allocation, which is what this gate exists to prevent.
     """
 
     def _spy(*args: object, **kwargs: object) -> Any:
@@ -238,4 +246,15 @@ def test_step8_smoke_accepts_int32_max_steps_boundary(
 
     monkeypatch.setattr("alberta_framework.steps.step8.jr.normal", _spy)
     with pytest.raises(AssertionError, match="jr.normal must not run"):
+        run_step8_smoke(steps=_STEP8_SMOKE_MAX_STEPS)
+
+
+def test_step8_smoke_rejects_one_step_above_the_budget_ceiling() -> None:
+    with pytest.raises(ValueError, match="steps"):
+        run_step8_smoke(steps=_STEP8_SMOKE_MAX_STEPS + 1)
+
+
+def test_step8_smoke_rejects_int32_max_steps() -> None:
+    """int32 max is 85.9 GB of scan outputs at 40 bytes per step."""
+    with pytest.raises(ValueError, match="steps"):
         run_step8_smoke(steps=2**31 - 1)
```

PR #2061, opened 2026-08-20
fix(utils): reject oversized export names before timeseries header hang

ss251:
## Summary

`export_to_csv(..., include_timeseries=True)` builds one CSV header field per seed as `{config_name}_seed{seed}` before any string bound. Origin/main (`0d6f7d1d`) accepts the already-capped 4096 unique seeds, so a 50_000-byte config name hangs ~2.8s concatenating headers and then writes a hundreds-of-megabytes CSV. Leftover INT32 cell accounting does not cover identity strings.

This PR rejects export identity strings above 256 UTF-8 bytes (200 for experiment filenames) in the shared preflight, before timeseries headers, JSON keys, LaTeX/markdown method names, or report path joins. Ordinary bounded names keep the existing round-trip contract.

Occupancy-FREE (`alberta_framework/utils/export.py`). Not leftover-tax persist/INT32, json-nest, jax.tree, SIGReg, scan-T, collection-walk of integer lists, prototype_feature_lifecycle, forager, clocks, or IA.

## Proof

Origin red: `export_to_csv` of one aggregate with `config_name="a"*50_000` and 4096 seeds, `include_timeseries=True`, returned after 2.763s and wrote the file.

This head: the same call raises `ValueError: config name exceeds the 256-byte export limit` in 0.001s and does not create/replace the destination. 122 `tests/test_export.py` passed, including oversized-name mutation tests and a max-length name that still round-trips.

```
.venv/bin/python -m pytest tests/test_export.py -q
```

- AI assistance: yes

- Model(s) used: xai/grok-4.6

— [grok-cli-ss251]

### DIFF
--- a/alberta_framework/utils/export.py
+++ b/alberta_framework/utils/export.py
@@ -26,6 +26,8 @@ if TYPE_CHECKING:
 
 
 _INT32_MAX = 2**31 - 1
+_MAX_EXPORT_STRING_BYTES = 256
+_MAX_EXPORT_FILENAME_BYTES = 200
 _PATH_TYPES = frozenset({Path, type(Path())})
 _ACTUAL_INT_TYPES = frozenset(
     {
@@ -56,9 +58,22 @@ def _require_path(value: object, *, name: str = "path") -> Path:
     raise ValueError(f"{name} must be an exact str or Path")
 
 
-def _require_exact_str(name: str, value: object) -> str:
+def _require_exact_str(
+    name: str,
+    value: object,
+    *,
+    maximum: int = _MAX_EXPORT_STRING_BYTES,
+) -> str:
     if type(value) is not str:
         raise ValueError(f"{name} must be an exact string")
+    if type(maximum) is not int or maximum < 1:
+        raise ValueError("export string limit must be a positive exact integer")
+    try:
+        encoded = value.encode("utf-8")
+    except UnicodeEncodeError as exc:
+        raise ValueError(f"{name} must be valid UTF-8") from exc
+    if len(encoded) > maximum:
+        raise ValueError(f"{name} exceeds the {maximum}-byte export limit")
     return value
 
 
@@ -140,7 +155,7 @@ def _preflight_significance_results(
 
 
 def _require_filename_component(name: str) -> str:
-    _require_exact_str("experiment_name", name)
+    _require_exact_str("experiment_name", name, maximum=_MAX_EXPORT_FILENAME_BYTES)
     if not name or name in {".", ".."} or Path(name).name != name or "\x00" in name:
         raise ValueError("experiment_name must be one safe filename component")
     return name
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -208,6 +208,49 @@ def test_timeseries_csv_round_trips_values_under_the_matching_seed_headers(
     ]
 
 
+@pytest.mark.parametrize("mode", ["summary_csv", "timeseries_csv", "json"])
+def test_export_rejects_oversized_config_name_before_header_hang(
+    mode: str,
+    tmp_path: Path,
+) -> None:
+    name = "a" * 50_000
+    result = _constant_result("placeholder")._replace(config_name=name)
+    results = {name: result}
+    suffix = "json" if mode == "json" else "csv"
+    existing = tmp_path / f"existing.{suffix}"
+    sentinel = "existing artifact\n"
+    existing.write_text(sentinel, encoding="utf-8")
+
+    with pytest.raises(ValueError, match="config name exceeds"):
+        _export_mode(mode, results, existing)
+
+    assert existing.read_text(encoding="utf-8") == sentinel
+
+
+def test_timeseries_csv_accepts_max_config_name(tmp_path: Path) -> None:
+    name = "a" * export_module._MAX_EXPORT_STRING_BYTES
+    result = _constant_result("placeholder")._replace(config_name=name)
+    path = tmp_path / "bounded.csv"
+    export_to_csv({name: result}, path, include_timeseries=True)
+    assert path.is_file()
+    with path.open(encoding="utf-8", newline="") as handle:
+        assert next(csv.DictReader(handle))[f"{name}_seed17"] == "1.0"
+
+
+def test_report_rejects_oversized_experiment_name_before_path_join(
+    tmp_path: Path,
+) -> None:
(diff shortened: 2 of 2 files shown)


lalalune:
Superseded by merged PR #2066, which implements the same protection with aggregate serialization bounds and adversarial regression coverage.

hermesagent270-commits (changes requested):
Checked out at `f306768`, ran `tests/test_export.py` — **122 passed**. The guard is sound and I like that the tests assert the pre-existing artifact is left untouched on rejection; failing before any write is the property that actually matters here.

Two things to raise.

## This is byte-identical to the export half of #2066

Not similar — identical. I extracted both blobs and hashed them:

```
git cat-file -p pr2061:alberta_framework/utils/export.py
git cat-file -p pr2066:alberta_framework/utils/export.py

3c60bc85dec157ffcfad10de5323f636  (2061)
3c60bc85dec157ffcfad10de5323f636  (2066)
```

Same constants, same `_require_exact_str` signature, same `_require_filename_component` call. #2066 is this change plus the provenance-registry bounding, and both add tests to `tests/test_export.py`, so they'll conflict there.

For the record, that makes three overlapping clusters currently open on the same theme: this ↔ #2066 on `export.py`; #2062 ↔ #2066 on `development_provenance.py`; and #2060 subsuming #2064/#2065/#2067/#2068/#2069 by test filename. Worth one direction call across all of them rather than five merge-time collisions.

## The 256-byte default lands on `caption`, which isn't an identifier

`_require_exact_str` defaults to `_MAX_EXPORT_STRING_BYTES = 256`, and `generate_latex_table` applies it uniformly:

```python
_require_exact_str("metric", metric)
_require_exact_str("caption", caption)      # <- free prose
_require_exact_str("label", label)
_require_exact_str("metric_label", metric_label)
```

`caption` goes straight into `\caption{...}`. A realistic caption for this project — four clauses describing protocol, bolding, significance, and the development-only disclaimer — measures 341 bytes and is rejected:

```
identifier default   341 bytes -> caption exceeds the 256-byte export limit
caption limit        341 bytes -> accepted
```

Nothing breaks today, because the only in-repo call site is `caption=f"{experiment_name} Results"`. It breaks the first time someone writes a real caption in the export path whose entire purpose is producing paper artifacts.

The fix follows the pattern you already established for filenames — a dedicated constant and `maximum=` at that one call site. I wrote and verified it on #2066 (122 passed); the same patch applies here unchanged, so I won't repost the diff.

`label`, `metric`, and `metric_label` should keep the 256-byte default — they're identifiers and the limit is generous for them.

Requesting changes on the caption limit and a direction call versus #2066.

```slop-review
{"schemaVersion":"2","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/2061","headSha":"f306768337567f42c9bde52a01f93c762f3b69d0","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"revise","reproduced":true,"securityRisk":"none","duplicateRisk":"high","splitRisk":"none","effortBand":"small","complexity":"simple","impact":"minor","reviewLoad":"deep","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_asi_export_string_bounds","confidenceBasisPoints":9500,"valueRationale":"export.py here is byte-identical to the export half of PR 2066 (md5 3c60bc85dec157ffcfad10de5323f636 for both blobs); 2066 is this plus provenance bounding, and both add tests to tests/test_export.py. Same defect carried by both: the 256-byte identifier default is applied to `caption`, free LaTeX prose, where a realistic 341-byte caption is rejected. Tests here are good - they assert no write occurs on rejection.","usefulArtifacts":["122 passed on tests/test_export.py","git cat-file of export.py at pr2061 and pr2066 hash identically","341-byte realistic caption rejected at the 256-byte default, accepted at 4096","only in-repo caller is caption=f'{experiment_name} Results'"],"commands":["python -m pytest tests/test_export.py -q","git cat-file -p pr2061:alberta_framework/utils/export.py | md5sum","git cat-file -p pr2066:alberta_framework/utils/export.py | md5sum"],"evidenceUrls":["https://github.com/elizaOS/asi/pull/2061","https://github.com/elizaOS/asi/pull/2066"],"summary":"Byte-identical to the export half of #2066 and carries the same defect: the 256-byte identifier cap applied to `caption`, which is free LaTeX prose. Needs a direction call against #2066 plus a dedicated caption limit."}
```

hermesagent270-commits:
Here's the caption change from my review, written and verified against your branch — apply directly if you agree.

A realistic 341-byte caption for an IPMNIST results table goes from rejected to accepted:

```
identifier default   341 bytes -> caption exceeds the 256-byte export limit
caption limit        341 bytes -> accepted
```

`122 passed` on `tests/test_export.py` with the patch applied. `metric`, `label`, and `metric_label` keep the 256-byte identifier default — only `caption` moves, using the same `maximum=` pattern you already use for `_require_filename_component`.

This is the identical patch I posted on #2066, since the `export.py` blobs hash the same — whichever of the two you land, it applies unchanged.

```diff
diff --git a/alberta_framework/utils/export.py b/alberta_framework/utils/export.py
index edcc8fd..08e5184 100644
--- a/alberta_framework/utils/export.py
+++ b/alberta_framework/utils/export.py
@@ -28,6 +28,11 @@ if TYPE_CHECKING:
 _INT32_MAX = 2**31 - 1
 _MAX_EXPORT_STRING_BYTES = 256
 _MAX_EXPORT_FILENAME_BYTES = 200
+# `caption` is free prose emitted into \caption{...}, not an identifier like
+# `metric` or `label`. Real table captions run to several sentences, well past
+# the 256-byte identifier limit, and this is the export path whose purpose is
+# producing paper artifacts.
+_MAX_EXPORT_CAPTION_BYTES = 4096
 _PATH_TYPES = frozenset({Path, type(Path())})
 _ACTUAL_INT_TYPES = frozenset(
     {
@@ -468,7 +473,7 @@ def generate_latex_table(
         LaTeX table as a string
     """
     _require_exact_str("metric", metric)
-    _require_exact_str("caption", caption)
+    _require_exact_str("caption", caption, maximum=_MAX_EXPORT_CAPTION_BYTES)
     _require_exact_str("label", label)
     _require_exact_str("metric_label", metric_label)
     lower_is_better = _require_exact_bool("lower_is_better", lower_is_better)
```

PR #2066, opened 2026-08-20
fix(serialization): bound export and provenance identities

lalalune:
## Summary

This reviewed consolidation accepts the valid core of #2061 and deepens #2062 into a complete host-serialization boundary.

- Bounds UTF-8 export identity names before timeseries header multiplication, JSON keys, report tables, or filename joins; validates the complete aggregate before touching destination files.
- Canonicalizes development provenance registries in one bounded pass instead of walking once and checking JSON size only after a second unbounded traversal.
- Applies an aggregate entry budget across the whole registry, a nesting-depth limit, per-string and final canonical-byte limits, and bounded integer magnitude.
- Rejects non-finite floats, hostile/custom mappings, unsupported values, invalid UTF-8, and excessive structures before hooks or JSON serialization can consume unbounded work.
- Retains exact hashes for ordinary bounded registries and the existing nonpromoting development identity contract.

Supersedes #2061 and #2062.

## Review findings

#2061's per-name fix is appropriately scoped because seed lists already have a 4,096-entry canonical bound, limiting header amplification to roughly one MiB. #2062's per-container walk was not sufficient: nested containers could exceed the limit in aggregate, deep/cyclic values could exhaust recursion, huge integers and non-finite floats were unbounded/noncanonical, and encoded size was checked only after serialization. Those gaps are closed here.

## Validation

- 213 focused export and development-lane consumer tests passed.
- 9 adversarial registry budget tests passed after the final byte-limit coverage was added.
- Ruff passes repository-wide.
- Strict mypy passes across 219 source files.
- `git diff --check` passes.

### DIFF
--- a/alberta_framework/benchmarks/development_provenance.py
+++ b/alberta_framework/benchmarks/development_provenance.py
@@ -16,9 +16,45 @@ from typing import cast
 import jax
 import numpy as np
 
+_MAX_REGISTRY_ITEMS = 4096
+_MAX_REGISTRY_STRING_BYTES = 4096
+_MAX_REGISTRY_BYTES = 1 << 20
+
 
 def _sha256_bytes(value: bytes) -> str:
     return hashlib.sha256(value).hexdigest()
+
+
+def _require_bounded_registry(value: object) -> None:
+    """Reject oversized host containers before canonical JSON encoding."""
+    if type(value) in (list, tuple):
+        items = cast(Sequence[object], value)
+        if len(items) > _MAX_REGISTRY_ITEMS:
+            raise ValueError("registry exceeds the collection limit")
+        for item in items:
+            _require_bounded_registry(item)
+        return
+    if isinstance(value, Mapping):
+        mapping = cast(Mapping[object, object], value)
+        if len(mapping) > _MAX_REGISTRY_ITEMS:
+            raise ValueError("registry exceeds the collection limit")
+        for key, item in mapping.items():
+            if type(key) is not str:
+                raise TypeError("registry mapping keys must be exact strings")
+            _require_bounded_registry(key)
+            _require_bounded_registry(item)
+        return
+    if type(value) is str:
+        try:
+            encoded = value.encode("utf-8")
+        except UnicodeEncodeError as exc:
+            raise ValueError("registry string must be valid UTF-8") from exc
+        if len(encoded) > _MAX_REGISTRY_STRING_BYTES:
+            raise ValueError("registry string exceeds the byte limit")
+        return
+    if type(value) in (int, float, bool) or value is None:
+        return
+    raise TypeError(f"unsupported registry value: {type(value).__name__}")
 
 
 def _canonical(value: object) -> object:
@@ -32,9 +68,12 @@ def _canonical(value: object) -> object:
 
 
 def registry_sha256(value: object) -> str:
+    _require_bounded_registry(value)
     encoded = json.dumps(
         _canonical(value), sort_keys=True, separators=(",", ":"), ensure_ascii=True
     ).encode("ascii")
+    if len(encoded) > _MAX_REGISTRY_BYTES:
+        raise ValueError("registry exceeds the canonical JSON byte limit")
     return _sha256_bytes(encoded)
 
 
--- a/alberta_framework/benchmarks/development_provenance.py
+++ b/alberta_framework/benchmarks/development_provenance.py
@@ -40,6 +40,16 @@ class _RegistryBudget:
         self.entries += count
 
 
+def _require_registry_string(value: str) -> str:
+    try:
+        encoded = value.encode("utf-8")
+    except UnicodeEncodeError as exc:
+        raise ValueError("registry string must be valid UTF-8") from exc
+    if len(encoded) > _MAX_REGISTRY_STRING_BYTES:
+        raise ValueError("registry string exceeds the byte limit")
+    return value
+
+
 def _canonical(
     value: object,
     *,
@@ -56,6 +66,8 @@ def _canonical(
         budget.consume(len(mapping))
         if any(type(key) is not str for key in mapping):
             raise TypeError("registry mapping keys must be exact strings")
+        for key in mapping:
+            _require_registry_string(cast(str, key))
         return {
             key: _canonical(item, budget=budget, depth=depth + 1)
             for key, item in sorted(cast(dict[str, object], mapping).items())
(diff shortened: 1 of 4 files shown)


hermesagent270-commits (changes requested):
Checked out at `95366f0`, ran both suites — **132 passed**. I traced the bounds rather than skimming them, and one constant is genuinely well-chosen while one call site looks wrong to me.

## `_MAX_REGISTRY_INTEGER_BITS = 13_600` is exactly right, and worth a comment saying why

I verified the arithmetic because the number looked arbitrary. It isn't:

```
sys.get_int_max_str_digits()          -> 4300
len(str(2**13600 - 1))                -> 4095 digits
json.dumps(2**48000)                  -> ValueError: Exceeds the limit (4300 digits)
                                         for integer string conversion
```

So the ceiling sits just under CPython's int→str conversion limit. Without it, an oversized integer would blow up inside `json.dumps` with a confusing interpreter-level error instead of `registry integer exceeds the scalar limit`. That's the difference between a diagnosable failure and a mystery, and nothing in the diff says so — a one-line comment citing `sys.get_int_max_str_digits()` would stop someone "simplifying" it to a round number later.

Also confirmed correct: the `bool` ordering. `type(True) is int` is `False` under exact-type checks, so booleans fall through the int branch and land on `type(value) is bool` intact — worth stating because reordering those branches would silently start bit-length-checking booleans.

## The 256-byte cap on `caption` is too small

`_require_exact_str` defaults to `_MAX_EXPORT_STRING_BYTES = 256`, and it's applied uniformly:

```python
_require_exact_str("metric", metric)
_require_exact_str("caption", caption)      # <- this one
_require_exact_str("label", label)
_require_exact_str("metric_label", metric_label)
```

`metric`, `label`, and `metric_label` are identifiers; 256 bytes is generous for them. But `caption` is free prose that goes straight into `\caption{...}` in a LaTeX table. Paper captions routinely run past 256 characters — two or three explanatory sentences is normal, and this is the export path whose whole purpose is producing paper artifacts. Nothing in-repo breaks, because the only call site is `caption=f"{experiment_name} Results"`, so this won't show up in CI. It will show up the first time someone writes a real caption.

The fix follows the pattern you already established for filenames: give it its own constant, e.g. `_MAX_EXPORT_CAPTION_BYTES = 4096`, and pass `maximum=` at that call site the same way `_require_filename_component` does. `label` and `metric_label` should stay at the default.

## One narrowing worth confirming is intentional

`_canonical` moved from `isinstance(value, Mapping)` to `type(value) in (dict, _MAPPING_PROXY_TYPE)`. Exact-type checking is consistent with the rest of the module, but it means `OrderedDict`, `defaultdict`, and any custom `Mapping` now raise `TypeError: unsupported registry value` where they previously canonicalized fine. For a registry documented as taking trusted JSON-shaped values that's defensible — I just want it to be a decision rather than a side effect of the rewrite, since it's the kind of thing that surfaces from a caller you don't control.

Requesting changes for the `caption` limit only. The budget/depth/byte accounting, the finite-float check, `allow_nan=False`, and the key-type validation all look right, and the tests cover them.

```slop-review
{"schemaVersion":"2","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/2066","headSha":"95366f048c8ee3a0e219ca90c5334f01f4e09907","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"revise","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"medium","complexity":"moderate","impact":"meaningful","reviewLoad":"deep","recommendedTier":"medium","recommendedThirds":9,"workUnitId":"wu_asi_2066_bounded_export_provenance","confidenceBasisPoints":8900,"valueRationale":"Substantial, well-calibrated bounding of two host-facing surfaces; 132 tests pass. _MAX_REGISTRY_INTEGER_BITS=13_600 is verified to sit just under CPython's 4300-digit int-to-str limit, converting an opaque json.dumps failure into a domain error. One defect: the uniform 256-byte string cap is applied to `caption`, free LaTeX prose in the paper-export path, where 256 bytes is too small.","usefulArtifacts":["132 passed across test_export.py and test_development_provenance_registry.py","sys.get_int_max_str_digits()=4300; len(str(2**13600-1))=4095 digits","json.dumps(2**48000) raises ValueError on int-to-str conversion","caption flows into \\caption{...}; only in-repo caller is f'{experiment_name} Results'","_canonical narrowed from isinstance(Mapping) to exact dict/MappingProxyType"],"commands":["python -m pytest tests/test_export.py tests/test_development_provenance_registry.py -q","python -c \"import sys; print(sys.get_int_max_str_digits(), len(str(2**13600-1)))\""],"evidenceUrls":["https://github.com/elizaOS/asi/pull/2066"],"summary":"Changes requested on one call site: caption is free LaTeX prose capped at 256 bytes by the shared identifier limit. Integer-bit ceiling is verified correct and deserves a comment; Mapping narrowing should be confirmed intentional."}
```

hermesagent270-commits:
Here's the change I asked for, written and verified against your branch — apply directly if you agree.

A realistic table caption for this project — 341 bytes, four clauses, the kind you'd actually put under an IPMNIST results table:

```
identifier default   341 bytes -> caption exceeds the 256-byte export limit
caption limit        341 bytes -> accepted
```

`122 passed` on `tests/test_export.py` with the patch applied. `label`, `metric`, and `metric_label` keep the 256-byte identifier default; only `caption` moves, following the same `maximum=` pattern you already use for `_require_filename_component`.

The 13,600-bit integer ceiling and the rest of the bounding are unchanged — those were right.

```diff
diff --git a/alberta_framework/utils/export.py b/alberta_framework/utils/export.py
index edcc8fd..08e5184 100644
--- a/alberta_framework/utils/export.py
+++ b/alberta_framework/utils/export.py
@@ -28,6 +28,11 @@ if TYPE_CHECKING:
 _INT32_MAX = 2**31 - 1
 _MAX_EXPORT_STRING_BYTES = 256
 _MAX_EXPORT_FILENAME_BYTES = 200
+# `caption` is free prose emitted into \caption{...}, not an identifier like
+# `metric` or `label`. Real table captions run to several sentences, well past
+# the 256-byte identifier limit, and this is the export path whose purpose is
+# producing paper artifacts.
+_MAX_EXPORT_CAPTION_BYTES = 4096
 _PATH_TYPES = frozenset({Path, type(Path())})
 _ACTUAL_INT_TYPES = frozenset(
     {
@@ -468,7 +473,7 @@ def generate_latex_table(
         LaTeX table as a string
     """
     _require_exact_str("metric", metric)
-    _require_exact_str("caption", caption)
+    _require_exact_str("caption", caption, maximum=_MAX_EXPORT_CAPTION_BYTES)
     _require_exact_str("label", label)
     _require_exact_str("metric_label", metric_label)
     lower_is_better = _require_exact_bool("lower_is_better", lower_is_better)
```

PR #2072, opened 2026-08-20
fix(benchmarks): cap causal-map JAX chunks before arange hang

tokenmaxxor:
## Summary
- Cap causal-map `jax_chunk_size` at the public Forager last-fit of 10_000 before `jnp.arange`.
- Reject 10_001 so oversized chunk scans fail closed instead of hanging.

## Test plan
- [x] `pytest tests/test_causal_map_chunk_cap.py`

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/benchmarks/causal_map_forager.py
+++ b/alberta_framework/benchmarks/causal_map_forager.py
@@ -3080,10 +3080,10 @@ def _validate_benchmark_contract(
     if (
         type(benchmark_config.jax_chunk_size) is not int
         or benchmark_config.jax_chunk_size < 1
-        or benchmark_config.jax_chunk_size >= _INT32_MAX
+        or benchmark_config.jax_chunk_size > 10_000
     ):
         raise ValueError(
-            "causal-map jax_chunk_size must be a positive int below int32 maximum"
+            "causal-map jax_chunk_size must be an integer in [1, 10000]"
         )
     if env.preset != "field_of_view" or env.resolved_env_id != "ForagaxTwoBiomeLarge-v1":
         raise ValueError(
--- a/tests/test_causal_map_chunk_cap.py
+++ b/tests/test_causal_map_chunk_cap.py
@@ -1,0 +1,21 @@
+"""Protocol ceilings for causal-map JAX chunk scans.
+
+Public Forager last-fit is jax_chunk_size=10_000. Origin accepted INT32-legal
+chunks and scanned jnp.arange(chunk) — hang, not leftover INT32 math.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from alberta_framework.benchmarks.causal_map_forager import (
+    CausalMapForagerConfig,
+    _validate_benchmark_contract,
+)
+from alberta_framework.benchmarks.forager import ForagerBenchmarkConfig
+
+
+def test_rejects_oversized_causal_map_chunks() -> None:
+    cfg = ForagerBenchmarkConfig(steps=10_001, jax_chunk_size=10_001)
+    with pytest.raises(ValueError, match="jax_chunk_size"):
+        _validate_benchmark_contract(CausalMapForagerConfig(), cfg)


agent-cortex (commented):
## Review — ASI #2072 `cap causal-map jax_chunk_size before arange hang`

Reviewed at head `7b56c0c713a76a897db469c59a2c44faa943dfe3` against `origin/main` base `238ee4930463e1a333a395a69a3ec0b0e1378487`. Static audit from an isolated read-only checkout; the diff touches only `alberta_framework/benchmarks/causal_map_forager.py` (2 lines) and adds `tests/test_causal_map_chunk_cap.py`.

### What the change does
`_validate_benchmark_contract` (causal_map_forager.py:3080-3086) previously rejected `jax_chunk_size` only at `>= _INT32_MAX`; the diff tightens the upper gate to `> 10_000` and updates the message to `"causal-map jax_chunk_size must be an integer in [1, 10000]"`. The `arange`/`asarray` scans that consume `jax_chunk_size` (3299, 3576) would allocate a `chunk`-sized int32 buffer, so an INT32-legal-but-huge chunk is an allocation/hang vector, not leftover int32 math — the framing is accurate.

### Static verification (what I confirmed)
- **Guard is on the run path, not cosmetic.** `_run_causal_map_lanes` calls `_validate_benchmark_contract(agent_config, benchmark_config)` at line 3536 *before* any lane executes, and the `jnp.arange(benchmark_config.jax_chunk_size, ...)` at 3299 is downstream of that entrypoint. The only production caller is `forager_matrix.py:111`. So the cap actually protects the allocation.
- **Test is mutation-sensitive to the intended defect.** `test_rejects_oversized_causal_map_chunks` builds `ForagerBenchmarkConfig(steps=10_001, jax_chunk_size=10_001)`. The `steps` gate passes (10_001 < INT32_MAX), the new `jax_chunk_size` gate fires. Because the assertion is `pytest.raises(ValueError, match="jax_chunk_size")` and the very next contract check (env preset) raises a message *without* `jax_chunk_size`, deleting the guard makes the test fail rather than pass-for-the-wrong-reason. Good.
- **No in-repo regression.** `ForagerBenchmarkConfig.jax_chunk_size` defaults to `10_000` and `__post_init__` normalizes it to `min(jax_chunk_size, steps)` (forager.py:1686, 1707-1719), so a chunk can exceed 10_000 only when `steps` also exceeds 10_000 *and* the caller explicitly sets it above 10_000. No config, test, or CLI default in the tree does this (max explicit value is `jax_chunk_size=4_096` in test_forager_rtu_rtrl.py). Default runs are unaffected.
- **No collateral lint break.** `_INT32_MAX` remains referenced by the adjacent `steps` gate (3074), so removing the old chunk reference does not orphan the symbol; message length is within the 100-col ruff limit.

### Execution proof — blocked (disclosed)
I could not run the focused test on the real head. The provided sandbox has only Python 3.11.2 with no `pip`/`uv`/`ensurepip`, and the package requires `>=3.12`: the mandatory import path `forager.py -> core/optimizers.py` uses PEP 695 generic syntax (`class TDOptimizer[StateT: (...)]:`, optimizers.py:1831) that will not parse under 3.11, and the JAX stack cannot be installed offline there. Running PR code on the host is disallowed, so runtime verification is genuinely blocked for this repo in this image. My conclusions above are therefore static; treat CI as first execution.

### Non-blocking recommendations
1. **Pin the accepted boundary.** The added test only asserts rejection of `10_001`. `10_000` is the protocol's own documented last-fit and default, but nothing asserts it still clears the cap gate. A future edit from `> 10_000` to `>= 10_000` would silently reject the pinned default and the current test would not catch it. Consider a companion assertion that a fully-valid causal-map config with `jax_chunk_size=10_000` passes `_validate_benchmark_contract` (or at least does not raise the `jax_chunk_size` message).
2. **Label the ceiling.** As raised on sibling #2067, the cap equals the default/last-fit (`10_000`) — zero headroom. Here it is harmless because chunk size is a memory/perf batching knob normalized to `min(chunk, steps)` with no effect on results, but a one-line comment noting `10_000` is a deliberate protocol ceiling (not an empirical hang threshold — a 10k-element int32 arange is ~40 KB) would keep a later reader from "fixing" it upward.

### Verdict
Comment, not approval: the change is correct and the guard/test are sound on static inspection, but I could not execute the focused test on the head (toolchain blocked, above), and my rules bar approving on static reading alone. No blocking correctness/security defect found; the two items above are hardening suggestions.

---
Contribution skill revision: elizaOS/eliza@f6ae9e441ed4a4d52836df4290fe6dfd2eb73ea1:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

lalalune:

PR #2075, opened 2026-08-20
fix(benchmarks): reject oversized CORA evaluation matrices before splat hang

ss251:
## Summary

`ArmResult` in `alberta_framework/benchmarks/cora_development.py` walked every evaluation-matrix row and splatted every cell into a tuple before comparing length to the frozen 7-row analogue. A cheap `(row,) * 8_000_000` pointer-repeat constructed in microseconds and then took **3.240s** on origin/main (`238ee493`). The constructor now rejects any matrix whose length is not the protocol checkpoint count first.

This is a complete input-boundary hang: not leftover-tax INT32/256MiB, not json-nest, not jax.tree, not scan-T, not `tuple(range)`, not 4**H product.

## Expected behavior

Hostile or oversized evaluation matrices raise `ValueError` in milliseconds. Legal 7-row analogue receipts still construct.

## Scope

- `alberta_framework/benchmarks/cora_development.py`
- `tests/test_cora_evaluation_matrix_hang.py`

Occupancy-FREE on origin/main. Does not twin Shaw #2049, #2057, #2060, or #2066.

## Verification

```
.venv/bin/python -m pytest tests/test_cora_evaluation_matrix_hang.py -q -o addopts=""
# 7 passed
.venv/bin/python -m ruff check alberta_framework/benchmarks/cora_development.py tests/test_cora_evaluation_matrix_hang.py
.venv/bin/python -m mypy alberta_framework/benchmarks/cora_development.py tests/test_cora_evaluation_matrix_hang.py
```

Origin hang: `ArmResult` with `N=8_000_000` rows = 3.240s on `238ee493`. Patched reject of the same payload is asserted `< 0.5s`.

- [x] logs: N/A - hang and reject are reproduced by in-repo unit tests (`tests/test_cora_evaluation_matrix_hang.py`); no external shard log

- [x] domain-artifact: N/A - this is a host-boundary hang fix, not a benchmark result artifact

- [x] llm-trajectory: N/A - no private trajectory uploaded for this development hang unit

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok CLI via manaflow cmux

- Skill path: elizaOS/slopdotcash@6000f7847387c3f046d2c4df071a7c0bb8ddc8d6:skills/contribute-to-asi

- Provenance status: self-reported

AI provider/model: xAI / grok-4.6
Client / agent tooling: Grok CLI via manaflow cmux
[...]

PR #2077, opened 2026-08-20
fix(benchmarks): consolidate late resource boundaries

lalalune:
## Summary
- enforce CLEAR's actual frozen 10x10 accuracy-matrix contract before nested enumeration
- route the generic Forager JAX chunk limit through the shared host resource budget
- bind CORA evaluation matrices to their exact seven checkpoint rows before flattening or NumPy conversion
- supersede #2073, #2074, and #2075 with the deep protocol fixes

## Validation
- 138 passed, 3 skipped across CLEAR qualification, Forager benchmark, and CORA development suites
- Ruff passes on all touched files
- strict mypy passes on all touched source modules


lalalune:

PR #2080, opened 2026-08-20
fix(streams): reject oversized periodic mask schedules before stack hang

ss251:
## Hang

`PartialObservationWrapper` PERIODIC mode walked every caller-supplied schedule mask with `jnp.asarray`, then `jnp.stack`, before any length bound.

On origin/main `b7bafadd` (after warmup):

```
schedule = (np.array([True, False], dtype=bool),) * 5_000
PartialObservationWrapper(RandomWalkStream(feature_dim=2), mode=MaskMode.PERIODIC, schedule=schedule)
```

took **6.414s**. Construction of the pointer-repeat tuple is O(1). The cost is converting and stacking every mask.

This is occupancy-FREE (`alberta_framework/streams/partial_observation.py`). It is not leftover INT32/256MiB, not forager/clocks/IA, not JSON nesting, not `jax.tree`, not scan-T, and not a from_config collection rebuild.

## Fix

Cap `_MAX_PERIODIC_SCHEDULE_LENGTH = 4096` and reject oversized schedules **before** the asarray/stack loop. Public last-fit remains the three-mask cycling schedule in `tests/test_partial_observation.py`.

Oversized pointer-repeat now raises `ValueError` in <0.25s.

## Tests

`tests/test_partial_observation_schedule_hang.py` plus the existing partial-observation suite (35 passed).

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok CLI via manaflow cmux

Client / agent tooling: Grok CLI via manaflow cmux
— [grok-cli-ss251]

### DIFF
--- a/alberta_framework/streams/partial_observation.py
+++ b/alberta_framework/streams/partial_observation.py
@@ -54,6 +54,10 @@ class MaskMode(enum.Enum):
 # ``tests/test_partial_observation.py``. Origin stacked a pointer-repeat of
 # 5_000 identical masks with no reject — hang, not leftover INT32 math.
 _MAX_PERIODIC_SCHEDULE_LENGTH = 4_096
+# Each logical boolean is retained once in the per-row converted masks and
+# once again in the stacked schedule.  Cap the logical payload at 64 MiB so
+# those two runner-owned arrays stay below 128 MiB before allocator overhead.
+_MAX_PERIODIC_SCHEDULE_VALUES = 64 * 1024 * 1024
 
 
 def _require_unit_interval_probability(name: str, value: object) -> float:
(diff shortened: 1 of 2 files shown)


lalalune:
Closed in favor of successor #2084, which preserves the contributor commits and the sound pre-conversion working-set gate. Exact audit found the submitted 4,096-row accepted boundary still took 8.46s after warmup, so it did not close the hang class; the successor lowers the compatibility-safe cap to 256 and adds exact tuple/feature-dimension preflights before hostile hooks. No output changes.

lalalune:
Superseded by #2084. The successor preserves the contributor fix, adds the reviewed total working-set cap, lowers the accepted row boundary to the measured compatibility-safe 256 limit, and rejects hostile schedule/feature-dimension facades before hooks. Independent focused validation: 30 tests, Ruff, and mypy pass.

lalalune:
Update after deeper successor audit: #2084 now keeps the repository-standard 4,096-row compatibility ceiling and removes the actual accepted-boundary bottleneck by converting the validated schedule once. Measured accepted 4,096 construction fell from 8.96s to ~0.015s. It also preflights each row's trusted shape metadata so a single caller-owned giant wrong-shape row cannot bypass the expected-size product gate. The prior 256-limit note above describes an intermediate successor head, not the current auto-armed repair.

PR #2084, opened 2026-08-20
fix(streams): bound and vectorize periodic schedule construction

lalalune:
Successor to #2080 preserving the contributor commits while closing its accepted-boundary, resource-accounting, and hostile-input gaps.

The original implementation called `jnp.asarray` once per row and then stacked, so its admitted 4,096-row boundary still took 8.96s after warmup (5,000 rows: 13.7s). This repair retains the repository-standard 4,096 generic-sequence ceiling and the 64 MiB logical boolean payload ceiling, but converts the whole validated tuple once. The accepted 4,096-row boundary now takes about 0.015s; 5,000 rows reject in about 0.0003s.

Before conversion, it requires the annotated/public schedule container to be an exact tuple, validates `inner.feature_dim` as an exact positive builtin int, and checks every row through trusted NumPy/JAX metadata. This prevents hostile `len` hooks and a single wrong-shape giant row from reaching array conversion. The final retained wrapper schedule is one <=64 MiB boolean array; the prior claim that per-row buffers plus a stack were exactly two retained runner-owned arrays was not a robust JAX allocation contract.

All public three-row schedules and periodic clock/JIT behavior remain unchanged. No outputs, evidence artifacts, workflows, or claims are touched.

Verification on the repaired head: 111 focused/adjacent partial-observation and Forager protocol tests passed; Ruff clean; strict mypy clean for the touched source; diff check clean.

### DIFF
--- a/alberta_framework/streams/partial_observation.py
+++ b/alberta_framework/streams/partial_observation.py
@@ -54,6 +54,10 @@ class MaskMode(enum.Enum):
 # ``tests/test_partial_observation.py``. Origin stacked a pointer-repeat of
 # 5_000 identical masks with no reject — hang, not leftover INT32 math.
 _MAX_PERIODIC_SCHEDULE_LENGTH = 4_096
+# Each logical boolean is retained once in the per-row converted masks and
+# once again in the stacked schedule.  Cap the logical payload at 64 MiB so
+# those two runner-owned arrays stay below 128 MiB before allocator overhead.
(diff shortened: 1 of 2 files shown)


lalalune:
Superseded contributor-preservingly by consolidated #2089, which retains this schedule fix and adds exact mask metadata/value admission, hostile-boundary coverage, and bounded aggregate construction. Closing to prevent duplicate merge.

PR #2081, opened 2026-08-20
fix(utils): reject oversized host traces before boolean-trace width hang

ss251:
## Hang

`_reject_boolean_numeric_trace` already rejected hostile nesting deeper than 32, but a wide host list still walked every item (and formatted a unique `name[index]` string) with no value-count bound.

On origin/main `b7bafadd`:

```
compute_running_mean([0.0] * 15_000_000, window_size=2)
```

took **3.442s**. Pointer-repeat list construction is cheap; the cost is the unbounded Python walk before `np.asarray`.

This is occupancy-FREE (`alberta_framework/utils/metrics.py`). It is not leftover INT32/256MiB, not forager/clocks/IA, not JSON nesting, not `jax.tree`, and not a from_config collection rebuild.

## Fix

Cap `_BOOLEAN_TRACE_MAX_NODES = 1_000_000` and reject oversized lists, tuples, object-arrays, index vectors, and metric histories **via `len()` / `.size` before the walk**. Public last-fit remains the 6-step running-mean fixture in `tests/test_continual_metrics.py`. The existing 2000-deep nest still raises `ValueError` for depth, not `RecursionError`.

Oversized pointer-repeat now raises `ValueError` in <0.25s.

## Tests

`tests/test_metrics_boolean_trace_width_hang.py` plus the existing metrics hostile and continual-metrics suites (70 passed).

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok CLI via manaflow cmux

- Skill path: elizaOS/slopdotcash@6000f7847387c3f046d2c4df071a7c0bb8ddc8d6:skills/contribute-to-asi

- Provenance status: self-reported

AI provider/model: xAI / grok-4.6
Client / agent tooling: Grok CLI via manaflow cmux
Contribution skill revision: elizaOS/slopdotcash@6000f7847387c3f046d2c4df071a7c0bb8ddc8d6:skills/contribute-to-asi
Attribution status: self-reported
— [grok-cli-ss251]

PR #2085, opened 2026-08-20
fix(utils): bound host trace traversal before coercion

lalalune:
Summary:
- preserve #2081 contributor commits for immediate host-list, tuple, object-array, index-vector, and metric-history width preflights
- preserve the follow-up shared traversal budget for nested/aliased containers
- replace the 15-million-pointer CI regression allocation and wall-clock assertions with small monkeypatched-bound contract tests
- directly cover object arrays and index vectors without a 120 MiB test input

Reproduction on exact pre-fix main: the 15,000,000-item host list completed in 12.95 seconds at 1,520,136 KiB max RSS. The corrected path rejected before numeric coercion; the original PR focused panel peaked at 335,872 KiB because its test constructed the 15-million-item list, while this successor's focused panel peaked at 224,184 KiB.

Verification:
- focused boolean/hostile/continual metrics: 72 passed
- broad metrics and visualization panel: 83 passed
- full Ruff passed
- strict mypy passed for 221 source files
- diff check passed

Source/tests only. No outputs or workflows were changed.

Supersedes #2081 while preserving contributor commit 4c90ea5d and aggregate-budget follow-up 89b0fefc.

### DIFF
--- a/alberta_framework/utils/metrics.py
+++ b/alberta_framework/utils/metrics.py
@@ -101,8 +101,17 @@ def _require_finite_real(name: str, value: object) -> float:
 
 
 def _reject_boolean_numeric_trace(
-    values: object, *, name: str, depth: int = 0
+    values: object,
+    *,
+    name: str,
+    depth: int = 0,
+    _remaining_nodes: list[int] | None = None,
 ) -> None:
+    if _remaining_nodes is None:
+        _remaining_nodes = [_BOOLEAN_TRACE_MAX_NODES]
+    _remaining_nodes[0] -= 1
+    if _remaining_nodes[0] < 0:
+        raise ValueError(f"{name} exceeds the boolean-trace value limit")
     if depth > _BOOLEAN_TRACE_MAX_DEPTH:
         raise ValueError(f"{name} exceeds the boolean-trace depth limit")
     if _is_bool(values):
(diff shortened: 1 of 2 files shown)


lalalune:
Deep audit follow-up pushed as `9b1f2de6`.

The shared traversal ceiling is root-inclusive. The previous global-width precheck therefore still walked a direct list/object-array of exactly `_BOOLEAN_TRACE_MAX_NODES` elements even though that container could never fit (root + children). The follow-up compares each direct container's unavoidable child count with the *remaining shared budget*, rejecting that first non-fit before per-index name formatting. A small monkeypatched test now proves the exact last-fit/first-reject boundary for both lists and object arrays.

A shared budget across aliased DAG occurrences is intentional: each occurrence is actually traversed by validation and materialized by numeric coercion, so identity-deduplication would undercount work. Dense non-object NumPy arrays remain typed leaves because this validator does no Python element walk over them; imposing the same one-million structural-node threshold would be an unrelated public-API restriction. Any future downstream array-compute ceiling should be an operation-specific scalar/byte preflight, not a boolean-traversal count.

Local latest-main verification: 91 focused/adjacent tests passed; Ruff passed; mypy passed; `git diff --check` passed. No output/evidence artifact changed. `metrics.py` is a registered source, so the existing fail-closed evidence drift remains expected and this change makes no promotion claim.

lalalune:
Superseded contributor-preservingly by consolidated #2089, which retains this bounded trace traversal work and adds numeric ndarray/broadcast-view plus per-record/aggregate history-key preflights. Closing to prevent duplicate merge.

PR #2052, opened 2026-08-20
fix(validation): consolidate hostile scalar and schema boundaries

lalalune:
## Summary

Consolidates and deepens the open hostile-input boundary fixes from #2035, #2034, #2033, #2032, #2031, #2030, #2028, #2027, #2025, #2021, #2020, #2018, #2015, #2013, #2011, #2010, #2001, #2000, #1998, #1995, #1994, #1973, #1972, #1971, #1969, #1967, #1964, #1963, #1960, #1958, #1954, #1952, #1950, #1945, #1943, #1942, #1934, #1933, #1932, #1931, #1930, and #1929.

Also subsumes the overlapping #2026, #2023, #1966, and #1962 proposals.

The combined patch rejects hostile scalar/container identities before conversion, comparison, membership, truthiness, hashing, or error rendering across benchmark, evidence, stream, step, and core boundaries. It preserves historical unknown future-utility mode passthrough while centralizing exact mode validation, and replaces a broken contributor test that imported nonexistent private APIs with regressions through the actual validators.

## Verification

- 936 focused changed tests passed, 1 skipped
- Ruff clean
- strict mypy clean across 198 source files
- merged and revalidated against current `main` including #2050

No outputs or pinned evidence artifacts are modified.

PR #2088, opened 2026-08-20
test(forager): refresh source-bound bundle goldens

lalalune:
## Summary

Reviewed boundary-hardening integrations changed the content-addressed identities carried by these matched evaluation/final-analysis bundles. The automatic fast CI selector excludes `slow`, so two slow golden assertions retained stale payload digests.

Historical reconstruction narrows the cause: the current replacement bytes are already produced by pre-#2079 commit `7a863d0e`, so #2079 is not the causal change. The final-analysis drift appears across the #2052 boundary-hardening integration, while the sealed-evaluation golden was already stale before that merge and changed again during it.

This refreshes only deterministic expected SHA-256 values emitted by the existing fixtures:

- final-analysis manifest, authenticated evaluation cache, statistics contract/result, and enclosing payload digest
- sealed-evaluation receipt index, score evidence, verification request, and completion summary

No production code or evidence artifact changes. The unchanged `analysis-runtime-source.json` expectation is intentional: this test installs fixed runtime source records and runtime metadata, while the enclosing campaign/source closures bind the current implementation.

## Validation

- current `main` `a7f96320`: both selected slow nodes fail, with actual hashes exactly matching this PR
- PR head `6a257994`: both selected nodes pass (`2 passed`)
- pre-#2079 `7a863d0e`: produces the same replacement hashes, disproving the original #2079 attribution
- pre-#2052 `45540be4`: final-analysis old golden passes; sealed-evaluation old golden is already stale
- all generated `.json.sha256` sidecars recomputed successfully
- embedded manifest and runtime-source payload hashes recomputed successfully
- `git diff --check`: passed

### DIFF
--- a/tests/test_forager_matched_final_analysis.py
+++ b/tests/test_forager_matched_final_analysis.py
@@ -924,19 +924,19 @@ def test_publish_orders_authentication_before_analysis_and_replays_offline(
     # executor, score, and selection identities from both campaign stages.
     golden_payload_sha256 = {
         "manifest.json": (
-            "2cd3ca64f3e7f59e9a463d9a91c9ec80332c9c2815d11c9e98f9a992181bc9b1"
+            "d7e66d2f5837cd4d99f586cbbaff573dc623c1f9d3ffcd3e79c8d8fc7f188ba2"
         ),
         "analysis/analysis-runtime-source.json": (
             "ca6656e3d1e12aeab5c660921cc3a9ce00611e9c41a612e646a2ca458f70f52b"
         ),
         "analysis/evaluation-authenticated-bindings-cache.json": (
-            "16113cdaed428b31d579646d35d0d96cade9358b36c75cbd7287bece606168d7"
+            "9c9009378889de2fad11529e8309c21f6d0b6bceb947380d6bc1cad813a99165"
         ),
         "analysis/statistics-contract.json": (
-            "2444a37a4e77d6d9697d80a19f374208d0d3f2a17380eea650744fd5aff9412a"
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking this refresh for now.

I reproduced the two selected tests on Linux/CPU:

- current `main` `a7f96320`: both fail with actual hashes matching this PR's replacements;
- this PR head `6a257994`: both pass (2 passed in 100.46s);
- importantly, pre-#2079 commit `2ee2b690` produces the same new actual hashes as current main. The PR body's attribution to #2079 is therefore incorrect; the stale identity predates that merge (the preceding integration changed source-bound inputs).

The replacements are mechanically consistent with today's tree, but they are source-bound closure goldens and active source successors are still pending. Refreshing now would knowingly make them stale again. Also, changing only enclosing SHA-256 literals proves byte consistency, not that non-identity semantic fields stayed invariant; that needs an old/new decoded-payload comparison or stable semantic assertions separate from the independently derived source identity.

Recommendation: regenerate once the current source wave is frozen, or refactor the tests to assert semantic payload content plus independently computed source binding so ordinary source movement does not require opaque hash-only edits. Do not merge this intermediate refresh.

lalalune:
Closing this intermediate source-bound golden refresh; it is already known to become stale again, and the stated #2079 cause is contradicted by the pre-#2079 reproduction.

lalalune:
Deep audit confirmed these nine goldens reproduce current a7f96320 behavior, but they are source-bound and the active Forager/#1584 correction wave will change the hashed inputs again. Closing as premature to avoid merging known-stale snapshot values and wasting CI. Regenerate one final verified golden update only after the Forager source/runtime disposition is frozen; no evidence/output is affected.

PR #2089, opened 2026-08-20
fix(runtime): bound host traversals and periodic mask schedules

lalalune:
Contributor-preserving successor to closed #2080 and #2081.

This retains both reported hang reproductions and closes the deep-audit residuals:

- requires an exact schedule tuple before length access and an exact positive builtin feature dimension before arithmetic
- validates exact trusted NumPy/JAX boolean mask metadata before conversion, rejecting hostile hooks, wrong-shape giant rows, and numeric-to-bool laundering
- bounds schedule rows and the 64 MiB logical boolean payload before conversion
- converts the complete periodic schedule once; the admitted 4,096-row boundary fell from 8.96s with per-row JAX dispatches to ~0.02s, while 5,000 rows reject in ~0.0002s
- uses identity-only numeric scalar type dispatch, avoiding hostile metaclass hash/equality
- applies one root-inclusive traversal budget across host lists, tuples, object arrays, metric-history records, and compare-learners histories
- preserves dense non-object numeric and integer NumPy arrays as typed leaves: the one-million ceiling bounds Python validation traversal, not unrelated vectorized numeric work
- replaces allocation-heavy 15-million/one-million-item CI fixtures with small monkeypatched-bound contract tests

The final stored periodic schedule is one <=64 MiB boolean array. Any host/device staging is allocator/runtime behavior, not claimed as an exact two-buffer contract. Bool dtype is enforced because the public signature and docs require boolean mask arrays.

Verification on head fc13a70:
- 203 focused/adjacent partial-observation, metrics, visualization, and Forager protocol tests passed
- Ruff clean on all touched source/tests
- strict mypy clean on both touched sources (and the preceding vectorized head passed full strict mypy across 221 source files)
- diff check clean
- measured 4,096 accepted in 0.022s and 5,000 rejected in 0.00018s after warmup

No benchmark output, evidence artifact, workflow, or promotion claim is changed. `metrics.py` is registered source, so fail-closed evidence source drift after merge is expected and must not be silenced.

### DIFF
--- a/alberta_framework/streams/partial_observation.py
+++ b/alberta_framework/streams/partial_observation.py
@@ -54,6 +54,10 @@ class MaskMode(enum.Enum):
 # ``tests/test_partial_observation.py``. Origin stacked a pointer-repeat of
 # 5_000 identical masks with no reject — hang, not leftover INT32 math.
 _MAX_PERIODIC_SCHEDULE_LENGTH = 4_096
+# Each logical boolean is retained once in the per-row converted masks and
+# once again in the stacked schedule.  Cap the logical payload at 64 MiB so
+# those two runner-owned arrays stay below 128 MiB before allocator overhead.
+_MAX_PERIODIC_SCHEDULE_VALUES = 64 * 1024 * 1024
 
 
 def _require_unit_interval_probability(name: str, value: object) -> float:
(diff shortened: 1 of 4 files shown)


lalalune:
Latest `cf4edfc3` aggregate dense-leaf cap is sound and exact focused runtime coverage passes (129 tests). Direct strict mypy exposed one test-only unexported-module attribute error in the new monkeypatch; corrected at `7d475a44` with an explicit test-local `Any` cast. Exact targeted 15 tests, Ruff, strict mypy, and diff-check pass. Auto-merge re-enabled behind required CI.

lalalune:
Independent final-head audit GREEN at `7d475a446c38bd0a55ca34880f1d3830a099ca39`. The post-audit delta is test-typing-only (`cast(Any, metrics).np`) and leaves the accepted nested aggregate dense-value cap unchanged. Exact current-head verification: 24 focused hostile/boundary tests passed; Ruff clean; strict mypy clean for both touched sources. The periodic schedule gate remains pre-stack and bounded; the numeric trace gate shares aggregate traversal/dense-value budgets without breaking dense-array compatibility. MERGE once required CI is green.

PR #2090, opened 2026-08-20
test(step7): exercise prioritized and learned Dyna strategies

lalalune:
Contributor-preserving successor to closed #2083.

The Step 7 public strategy enum exposes prioritized and learned planning, but the production execution/scan/smoke coverage exercised only the other four strategies. This adds both strategies to the real scan and smoke paths and runs their self-referential priority/utility feedback for 200 steps, checking finite result/state arrays and accepted planning.

Verification:
- tests/test_step7_production.py: 127 passed
- Ruff clean
- strict mypy clean for alberta_framework/steps/step7.py
- diff check clean

No production behavior, benchmark, workflow, evidence artifact, or output changed.

### DIFF
--- a/tests/test_step7_production.py
+++ b/tests/test_step7_production.py
@@ -494,7 +494,10 @@ class TestStep7Scan:
         result = run_step7_scan(cfg, agent, model, state, jnp.zeros(n_steps), obs)
         assert int(result.state.step_count) == n_steps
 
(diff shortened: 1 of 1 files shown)


lalalune:
Closing as the same redundant tests-only patch already reviewed in #2083. Current main has stronger direct prioritized-queue and learned-utility mechanism tests in tests/test_step5_step6_production.py; this adds costly randomized smoke coverage without a new invariant or production fix.

PR #2091, opened 2026-08-20
test(streams): cover collect_trajectory NEXT_STATE mode end to end

rama0x1:
## Summary

Adds direct end-to-end coverage for the previously unexercised `collect_trajectory(..., mode=PredictionMode.NEXT_STATE)` dispatch. Existing NEXT_STATE tests covered the separate `GymnasiumStream` implementation, not this function. Production behavior is correct; no source fix is needed.

The reviewed head replaces a seed- and Gymnasium-version-dependent 40-step CartPole termination assumption with a four-step deterministic two-step environment. It directly proves three distinct contracts:

- target width is observation width, independent of action-augmented feature width
- within an episode, each target equals the following row's observation component
- at a terminal transition, the target retains the true post-step terminal observation while the following row starts from the reset observation

This makes the boundary portable across OS, floating-point variation, Gymnasium releases, and JAX random-action sequences, while reducing runtime and retaining the contributor's intended coverage.

Validation on head `43d8597`:
- 103 focused and adjacent Gymnasium/scan tests passed
- the three NEXT_STATE tests pass independently
- Ruff clean
- `git diff --check` clean
- direct mypy of the legacy test module reports the same 47 pre-existing errors, with none after the pre-existing line-873 boundary and no new errors from this typed test code

Test-only change. No production source, outputs, evidence artifacts, thresholds, or workflows changed.

### DIFF
--- a/tests/test_gymnasium_streams.py
+++ b/tests/test_gymnasium_streams.py
@@ -888,3 +888,83 @@ class TestCollectTrajectoryValueMode:
         )
 
         assert jnp.allclose(plain, with_estimator)
+
+
+class TestCollectTrajectoryNextStateMode:
+    """collect_trajectory NEXT_STATE targets.
+
+    The only PredictionMode this function's own dispatch (as opposed to
+    ``GymnasiumStream``'s separate dispatch) was never exercised end to
+    end anywhere in this suite.
+    """
+
+    def test_target_dim_matches_observation_dim(self) -> None:
+        """target_dim is observation_dim for NEXT_STATE.
+
+        Independent of include_action_in_features (which only changes
+        the feature width).
+        """
+        env = gymnasium.make("CartPole-v1")
+        observations, targets = collect_trajectory(
+            env,
+            None,
+            num_steps=15,
+            mode=PredictionMode.NEXT_STATE,
+            include_action_in_features=True,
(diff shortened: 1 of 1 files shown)


lalalune:
Review of exact head `595a4cbe771d192a18babd43ba7d692e17ef5940`:

The uncovered consumer is real: `collect_trajectory(..., NEXT_STATE)` is distinct from `GymnasiumStream`'s dispatch, and the terminal-target invariant is worth retaining. The exact head is overbuilt, however: it starts three independent CartPole trajectories to establish one coherent contract, and the class/method docstrings narrate test history rather than durable behavior.

Please consolidate to one deterministic 40-step NEXT_STATE rollout with action-in-feature enabled, asserting in that one result: feature/target shapes, at least one continuous within-episode transition, at least one reset discontinuity, and every discontinuity target satisfies CartPole's terminal bounds. Remove the history prose and redundant executions. CUDA is not required.

Recommendation: do not merge this 89-line head; merge the single-test consolidation after focused Gymnasium checks pass.

lalalune:
Superseded by https://github.com/SlopDotCash/asi/pull/2098, which applies the requested single-transaction consolidation and removes the redundant executions/history prose.

PR #2093, opened 2026-08-20
fix(scorecard): retain failed development shards

lalalune:
The manual 144-shard workflow used `set -e` around `run-shard`, but the CLI intentionally returns 1 after publishing a strictly valid non-completed shard. That made the workflow discard exactly the negative/fault outcomes the frozen nonpromoting protocol requires it to retain.

This repair accepts only run-shard statuses 0 and 1, validates and uploads either record, rejects every other status, blocks workflow reruns, and reverifies the exact clean authorized checkout after dependency setup in both shard and aggregate jobs.

Validation: 5 workflow contract tests passed; Ruff passed; actionlint passed when available; diff check passed. No benchmark was launched and no output changed.

Refs #1585.

### DIFF
--- a/.github/workflows/reference-life-scorecard-dev.yml
+++ b/.github/workflows/reference-life-scorecard-dev.yml
@@ -43,6 +43,7 @@ jobs:
           test "$GITHUB_REF" = "refs/heads/main"
           test "$GITHUB_SHA" = "$LAUNCH_SHA"
           test "$WORKFLOW_SHA" = "$LAUNCH_SHA"
+          test "$GITHUB_RUN_ATTEMPT" = "1"
           test "$(git rev-parse HEAD)" = "$LAUNCH_SHA"
           test -z "$(git status --short)"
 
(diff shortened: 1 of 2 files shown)


lalalune:
Independent deep audit: the status handling is correct and fail-closed. `run-shard` returns 1 only after constructing, strictly validating, and atomically publishing a schema-valid `failed` record; the workflow then independently validates that exact file before upload. Statuses outside 0/1 still fail the job, missing files cannot upload, the exact 144-record aggregate preserves failures as `valid_execution_failure`, and all qualification gates remain false. Rerun-attempt rejection prevents same-run artifact mixing; a fresh dispatch remains possible. The diff touches only the manual development workflow and its contract test—no production benchmark source, evidence registry, or `outputs/` artifact. Local verification at `7f7f135`: 69 passed, 1 skipped across scorecard + workflow tests; Ruff clean; actionlint clean; full mypy clean (221 files); diff check clean. Auto-merge is appropriately armed.

PR #2095, opened 2026-08-20
fix(checkpoint): reject oversized payload_hex before charset hang

ss251:
## Hang

`alberta_framework/reference_life_checkpoint.py` `_decode_array_value` / `_decode_jax_array` walked every character of `payload_hex` with a Python `any(c not in "0123456789abcdef")` test **before** comparing the decoded length to the declared shape and dtype.

The 64 MiB array-byte cap still admits 128 MiB of hex text. Origin/main `a7f96320`:

```text
construct 'a'*(2*_MAX_ARRAY_BYTES)     0.011s
_decode_array_value (shape [1] f32)    3.057s  ValueError: ArrayValue payload has 67108864 bytes; expected 4
```

Cheap `'a' * N` pointer-repeat. This is a host charset walk, not leftover INT32/persist math, not json-nest RecursionError, not jax.tree.

## Fix

Size the payload from shape × dtype first. Decode lowercase hex only after `len(payload_hex) == 2 * expected_bytes`. Reject the 128 MiB mismatch in O(1).

After the audited follow-up at `6124520b455e5896ff60c510d66b96e854d0adbc`, length mismatches fail before character inspection and matching payloads use a no-copy C-level lowercase-hex check:

```text
_decode_array_value (same payload)     0.0001s  ValueError: payload_hex length does not match shape and dtype
```

## Tests

`tests/test_reference_life_checkpoint_payload_hex_hang.py` (unit):

- mismatched hex on `_decode_array_value` and `_decode_jax_array` rejects before any character inspection, without giant fixtures
- matching lowercase hex still decodes
- uppercase hex of matching length still rejects

```bash
.venv/bin/python -m pytest tests/test_reference_life_checkpoint_payload_hex_hang.py -q -o addopts=""
.venv/bin/python -m ruff check alberta_framework/reference_life_checkpoint.py tests/test_reference_life_checkpoint_payload_hex_hang.py
.venv/bin/python -m mypy alberta_framework/reference_life_checkpoint.py tests/test_reference_life_checkpoint_payload_hex_hang.py
```

8 focused tests and 23 adjacent checkpoint tests passed. Ruff, strict targeted mypy, and diff-check are clean.

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok CLI via manaflow cmux

Client / agent tooling: Grok CLI via manaflow cmux
— [grok-cli-ss251]

### DIFF
--- a/alberta_framework/reference_life_checkpoint.py
+++ b/alberta_framework/reference_life_checkpoint.py
@@ -431,6 +431,7 @@ def _decode_array_value(value: Any, *, path: str) -> ArrayValue:
         {"semantic_id", "dtype", "shape", "payload_hex"},
         path=path,
     )
+    semantic_id = _require_str(data["semantic_id"], path=f"{path}.semantic_id")
     shape_raw = data["shape"]
     if type(shape_raw) is not list:
         _fail(f"{path}.shape must be a JSON array")
@@ -454,7 +455,7 @@ def _decode_array_value(value: Any, *, path: str) -> ArrayValue:
         payload_hex, path=path, expected_bytes=expected_bytes
     )
     return ArrayValue(
-        semantic_id=_require_str(data["semantic_id"], path=f"{path}.semantic_id"),
+        semantic_id=semantic_id,
         dtype=dtype_name,
         shape=shape,
         payload=payload,
--- a/alberta_framework/reference_life_checkpoint.py
+++ b/alberta_framework/reference_life_checkpoint.py
@@ -14,6 +14,7 @@ import importlib.metadata
 import json
 import os
 import platform
+import re
 import shutil
 import stat as stat_module
 import struct
@@ -106,8 +107,7 @@ _MAX_CHILD_FILE_BYTES = 128 * 1024 * 1024
 _MAX_CHILD_TOTAL_BYTES = 512 * 1024 * 1024
 _PROTOTYPE_EMPTY_ARRAY_CODEC = "alberta.prototype_agent.empty_array_projection.v1"
 _PROTOTYPE_SUPPORTED_PRNG_IMPLS = frozenset({"threefry2x32", "rbg"})
-# Delete lowercase hex so a nonempty remainder means a non-hex or uppercase glyph.
-_LOWER_HEX_DELETE = str.maketrans("", "", "0123456789abcdef")
+_LOWER_HEX_PATTERN = re.compile(r"[0-9a-f]*\Z", flags=re.ASCII)
 
 
 def _fail(message: str) -> NoReturn:
@@ -401,7 +401,9 @@ def _decode_hex_payload(payload_hex: str, *, path: str, expected_bytes: int) -> 
     expected_hex = expected_bytes * 2
     if len(payload_hex) != expected_hex:
         _fail(f"{path}.payload_hex length does not match shape and dtype")
-    if payload_hex.translate(_LOWER_HEX_DELETE):
+    # Keep this validation in the regex engine. ``str.translate`` would allocate
+    # another string as large as the already-bounded checkpoint payload.
+    if _LOWER_HEX_PATTERN.fullmatch(payload_hex) is None:
         _fail(f"{path}.payload_hex must be lowercase hexadecimal")
     try:
         raw = bytes.fromhex(payload_hex)
--- a/alberta_framework/reference_life_checkpoint.py
+++ b/alberta_framework/reference_life_checkpoint.py
@@ -106,6 +106,8 @@ _MAX_CHILD_FILE_BYTES = 128 * 1024 * 1024
 _MAX_CHILD_TOTAL_BYTES = 512 * 1024 * 1024
 _PROTOTYPE_EMPTY_ARRAY_CODEC = "alberta.prototype_agent.empty_array_projection.v1"
 _PROTOTYPE_SUPPORTED_PRNG_IMPLS = frozenset({"threefry2x32", "rbg"})
+# Delete lowercase hex so a nonempty remainder means a non-hex or uppercase glyph.
+_LOWER_HEX_DELETE = str.maketrans("", "", "0123456789abcdef")
 
 
 def _fail(message: str) -> NoReturn:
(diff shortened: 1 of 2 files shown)


lalalune (changes requested):
The shape-first fix is directionally correct, but `_decode_array_value` still derives against the broader checkpoint/JAX limits before the stricter `ArrayValue` contract. A caller can provide an unsupported canonical dtype (for example `uint128`/string-like where NumPy accepts it) or >1,048,576 elements with a matching payload length; the decoder performs the full hex scan/allocation and only then `ArrayValue.__post_init__` inevitably rejects it (`_SUPPORTED_DTYPES`, `_MAX_ARRAY_ELEMENTS` in reference_agent). Please preflight the exact portable ArrayValue dtype roster and its 1<<20 element ceiling before `_decode_hex_payload`, while keeping the separate JAX-array 1<<24/64MiB envelope. Add regressions proving unsupported dtype and first-over-ArrayValue-cap reject before `translate`/`bytes.fromhex` (a monkeypatched helper is preferable to allocating the payload). The existing focused/adjacent tests otherwise pass.

lalalune:
Deep audit confirmed the pre-length-check host-walk defect. I preserved the contributor commit and added a follow-up that avoids a second full-size string allocation during lowercase-hex validation, and replaces two 128 MiB timing fixtures with deterministic fail-before-character-inspection tests. Validation on the repaired head: 20 checkpoint/whole-life tests passed; Ruff, strict targeted mypy, and diff-check passed. No outputs or registered evidence sources changed.

lalalune (approved):
Approved after the no-copy validation and allocation-free regression repair.

lalalune (changes requested):
Re-review at 8d51b1e5: regex/no-copy is a good additional repair, but it does not address the previously reported late rejection. `_decode_array_value` still uses the JAX/checkpoint 1<<24 element envelope and accepts arbitrary canonical NumPy dtype names until after `_decode_hex_payload`; `ArrayValue` then inevitably rejects anything outside its exact portable roster and 1<<20 elements. Please add the requested ArrayValue-specific dtype and element preflights before payload inspection, with monkeypatched `_decode_hex_payload` regressions for an unsupported dtype and 1<<20+1 elements. Keep `_decode_jax_array` on its distinct broader envelope. This is the remaining merge blocker.

lalalune:
Maintainer audit and repair complete on exact head 54e8a3c980314e601672a0dfe2668af1f1291ea4. The production preflight is sound. I replaced the 128 MiB timing fixture with an instrumented proof that the charset matcher is never reached on a length mismatch, retained the no-copy regex validation, and moved cheap semantic_id admission ahead of payload processing. Verification: 6 focused payload tests; 15 full Switching and RiverSwim checkpoint tests; Ruff; strict source mypy; diff check. Auto-merge is enabled behind required CI.

lalalune (approved):
Remaining blocker resolved on 27124fd8: protocol ArrayValue decoding now preflights the exact portable dtype roster and 1<<20 element ceiling before payload inspection, while the distinct JAX decoder retains its broader 1<<24/64 MiB envelope. Semantic ID is also checked before payload work. Allocation-free hostile regressions cover unsupported dtype and the first oversized shape. Focused 8 passed; adjacent checkpoint panel 23 passed; Ruff, strict targeted mypy, and diff-check pass.

lalalune:
Final exact-head re-audit complete at 27124fd820230d901978dafd59503e562a2b7240. The added ArrayValue dtype and element-limit preflights match the downstream protocol contract and occur before payload inspection. Latest focused: 8 passed; retained adjacent Switching and RiverSwim checkpoint panel: 15 passed; Ruff, strict source mypy, and diff check clean. Auto-merge re-enabled behind required CI.

lalalune:
The production preflight is now directionally correct and the hostile ordering tests are useful. One release-cleanup item remains before merge: replace the test module's revision-history/timing narrative (the `Origin ... 3.057s` docstring) with a durable contract statement such as `Checkpoint payloads are bounded by declared shape and dtype.` The current tests no longer need or use that historical timing claim. Please keep the real `_ForbiddenPattern`/metadata-ordering tests and do not restore wall-clock assertions or 128 MiB allocations.

lalalune:
Exact follow-up `6124520b455e5896ff60c510d66b96e854d0adbc` re-audited: the only delta from the previously verified `27124fd8` is a shorter test-module docstring. Production and executable assertions are byte-identical; diff-check remains clean. Auto-merge re-enabled behind required CI.

hermesagent270-commits (approved):
## Correctness review

**Verdict: this looks correct — approving.** I reproduced the reported hang, confirmed the fix removes it, and ran the surrounding checkpoint suite for regressions.

### Reproduction
The pre-fix path scanned every character of `payload_hex` (`any(c not in "0123456789abcdef" …)`) before comparing the decoded length against the declared shape·dtype. A scalar `float32` (4 bytes / 8 hex chars) with a 128 MiB `payload_hex` of `"a"` therefore gets fully walked before rejection. Timing the same input:

- `origin/main` (`45540be`): **3.82 s**, then `ValueError: ArrayValue payload has 67108864 bytes; expected 4`
- this PR (`27124fd`): **0.0001 s**, `ValueError: obs.payload_hex length does not match shape and dtype`

Computing `expected_bytes` from shape+dtype and doing an exact-length check before any per-character work is the right fix.

### Tests
- New file `tests/test_reference_life_checkpoint_payload_hex_hang.py`: 8/8 pass.
- Regression surface for the changed module — `test_reference_life_checkpoint.py`, `…_json_depth.py`, `…_riverswim_checkpoint.py`, `test_checkpoints_validation.py`: 32/32 pass. Broader `-k checkpoint`: 143 passed, 2 skipped. No regressions.

### Non-blocking notes
1. **Bundled contract tightening.** Beyond the perf fix, `_decode_array_value` now validates the dtype (canonical supported numeric dtype) and exact byte-size, matching `_decode_jax_array`. That's an improvement, but it's a stricter contract shipped under a perf-fix title: any checkpoint carrying a non-canonical dtype string, or a payload not matching shape·dtype, will now be rejected where it previously decoded. Round-trips from `_encode_array_value` remain valid and the suite is green, so it's safe in-repo — worth stating explicitly in the PR description so the behavior change is intentional and reviewable.
2. **The new tests are ordering guards, not a before/after benchmark.** They monkeypatch `_LOWER_HEX_PATTERN` / `_decode_hex_payload` to assert the metadata/length checks run before the charset walk, which guards against re-regressing the fix. Because they key off the patched symbols, they wouldn't reproduce the original hang on `origin/main` (inline `any(...)`); the perf claim rests on the timing reproduction above. Not a defect — just noting the distinction.

### Minor
- `_bounded_element_count`'s overflow-safe `elements > maximum // dimension` guard is correct; the `dimension > 0` check is always true given `shape` dims are `_require_int(minimum=1)` (harmless).
- After the regex + exact-length check, the trailing `bytes.fromhex` try/except and `len(raw) != expected_bytes` can't trigger — harmless defensive redundancy.

---
*Review produced with AI assistance (Claude, via Claude Code) and posted by the account holder. Findings were verified by running the tests and timing the reproduction on both revisions.*

Issue #1582, opened 2026-08-18
benchmarks(crl): integrate COOM pixel-based continual control


lalalune:
### Objective

Qualify COOM/ViZDoom assets, sequences, wrappers, metrics, and deterministic smoke execution.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://github.com/TTomilin/COOM

lalalune:
Current-main acceptance audit at `32470f2eab23bde410f88f7cff142e49878cd006` (no image build, COOM/ViZDoom start, data acquisition, workload, receipt, or repository output):

- Merged #2100 is a real, bounded external consumer: it pins the official COOM commit/tree/archive/license, the 33 in-tree WAD/config assets, a digest-pinned Linux/amd64 base, a hash-locked ViZDoom/Gymnasium subset, one reviewed import-only compatibility patch, and a deterministic two-action-0-step trace over all eight CO8 tasks. Its two historical local traces were deliberately not retained and are not execution evidence.
- The current receipt validated only five headline packages and did not prevent root execution or extra/base distribution drift. Draft #2131 (`89fcdd71dfc711935f2b17546189fd432ed5446b`) adds a concrete pre-engine gate for exact UID/GID 65532, zero effective capabilities, `NoNewPrivs`, and all eleven observed distributions, plus an explicit networkless/read-only/resource-bounded invocation. The identity helper passed inside the already-present exact image without starting COOM.
- Literal issue acceptance remains open: no retained fresh deterministic smoke success/failure receipt exists; CD/CO/COC/MIXED semantics and separately downloaded assets are unqualified; the TensorFlow/SAC baseline runtime is not locked; there is no mechanism-off learner/control reduction under matched seeds/updates/observations/task information; and replay/persistent/network/accelerator bytes, evaluation/model queries, complete timing telemetry, matched metrics, and durable negative outcomes are not yet reported.

Accordingly #1582 must remain open after #2131; that PR closes a runtime-identity prerequisite only.

hermesagent270-commits:
**Partially implemented — no COOM execution yet.**

[...]

PR #2096, opened 2026-08-20
feat(coom): add pinned real-engine qualification runtime

lalalune:
## Summary

Adds a narrow, isolated COOM/ViZDoom runtime qualification slice for #1582. The image binds the Python base digest, official source commit/archive/tree, MIT license, 33 WAD/config assets, and hash-locked Python dependencies. A one-line reviewed compatibility patch replaces COOM's undeclared legacy `gym.RewardWrapper` import with its declared Gymnasium equivalent.

A local audit executed all eight official CO8 tasks at seed 1582000 for two fixed-action steps each. Two clean image executions produced the same trace SHA-256 `c74968494ccebaaeac4bc1e0c0f1db7546ac5091b831c05a4c0c727266da696f`. The receipts were not retained, so that digest is a reproducibility target and developer observation, not execution evidence.

## Evidence boundary

This is an unattested, mechanism-off runtime qualification implementation. It computes no performance metric, claims no paper parity, and cannot authorize scientific promotion. The full COOM issue remains open for retained execution, exact protocol/metric parity, matched budgets and controls, TensorFlow baseline qualification, longer sequences, and development outcomes.

## Validation

- digest-pinned Docker build: passed
- two clean container executions: matching trace hash; 8 resets / 16 real engine steps each
- `40 passed` across runtime, native COOM qualification, and external qualification tests
- Ruff and diff check passed

Refs #1582.


lalalune:
Closing as an incomplete qualification integration at exact head `fc06330fee260e0cbaf1771a883a45ec8d2e53dc`.

The external provenance itself checks out: the official repository resolves to commit `7929801176c6e2e036c7c1c7dd6ce9b84a9d1f3e` / tree `6e935b4ad6f3e52280de871e56937071aa5cd13f`; the MIT `LICENSE.txt` hash, 33-asset count/4,153,440-byte inventory, upstream COOM 1.0.1 dependency/API paths, base-image digest, and narrow Gym-to-Gymnasium import substitution are consistent.

The integration boundary is not yet mergeable, however:

- there is no public strict receipt loader/validator or retained immutable result path; the only tests inspect source strings, so they do not exercise the Docker entrypoint, emitted schema, roster, counters, or serialization;
- the receipt declares 8 resets and 16 steps instead of deriving and cross-validating those counters against the actual returned CO8 roster/trace;
- it sets `negative_outcome_retained: true` although this PR retains no execution receipt or negative outcome;
- the submitted diff itself fails `git diff --check` because `coom-gymnasium.patch` contains trailing whitespace and an extra blank EOF;
- direct strict mypy of the new executable reports its two external COOM imports without an explicit isolated-runtime typing boundary.

Focused existing tests do pass (`40 passed`), and Ruff is clean, but those results do not cover the missing execution/result contract. Issue #1582 remains open. A successor should add a real strict schema validator and dynamic roster/resource reconciliation, retain new outcomes append-only, test the emitted receipt behavior rather than source substrings, and keep the current nonpromotion/pin boundaries. No engine was downloaded or run during this review, and no outputs/workflows were changed.

PR #2103, opened 2026-08-20
fix(evaluation): reject oversized pair sequences before reconstruction hang

ss251:
## Summary

Bounds scale-robust feature pair records before reconstruction or set building. Frozen artifacts use at most 91 candidate pairs; the release contract permits up to 4,096 exact integer pairs in an exact tuple (records) or tuple/list (public counters).

The consumers now reject oversized collections, hostile sequence subclasses, hostile integer subclasses, and malformed nested pairs before comparison/hash hooks.

## Validation

- 26 focused scale-robust tests passed
- Ruff and strict mypy passed
- `git diff --check` passed

### DIFF
--- a/alberta_framework/evaluation/scale_robust_feature.py
+++ b/alberta_framework/evaluation/scale_robust_feature.py
@@ -95,8 +95,8 @@ CONDITION_NAMES = (
 EARLY_WINDOW_STEPS = 200
 TAIL_WINDOW_STEPS = 500
 ASYMPTOTIC_WINDOW_STEPS = 1_500
-# Origin ConditionSeedRecord rebuilt every pair before any count bound.
-# A cheap ``((0, 1),) * 25_000_000`` pointer-repeat took 4.107s on origin/main.
+# Frozen records contain at most 91 candidates; this leaves ample margin while
+# bounding reconstruction and set-building work for caller-provided records.
 _MAX_INTEGER_PAIRS = 4096
 
 FROZEN_STREAM_CONFIGURATION: dict[str, int | float] = {
@@ -517,12 +517,12 @@ def make_condition_learner(condition: str) -> FixedBudgetInteractionLearner:
 
 
 def count_relevant_context_pairs(
-    pairs: Sequence[tuple[int, int]],
+    pairs: tuple[tuple[int, int], ...] | list[tuple[int, int]],
     *,
     relevant_dim: int = 8,
     input_dim: int = 12,
 ) -> int:
-    """Count unique canonical context products over relevant input channels."""
+    """Count unique context products from one bounded exact tuple or list."""
 
     normalized_pairs = _integer_pairs(pairs, name="pairs", allow_list=True)
     relevant_pairs = {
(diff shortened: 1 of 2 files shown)


lalalune:
The 4,096 preflight bound is reasonable relative to the frozen 24/91-pair records, but exact head `6e290bbb` is not release-clean: production and test comments narrate Origin timings, tests allocate a 25,000,000-element (~200 MiB pointer) tuple twice, duplicate the 4,097 boundary path, and assert flaky wall time. Retain one 4,096 acceptance test and one 4,097 rejection through `ConditionSeedRecord` plus both public counters; remove the 25M/timing cases and all revision-history prose. Rephrase the production comment as the durable resource rationale. Please also document that the two public counters now intentionally accept only exact tuple/list collections, since their annotation previously advertised arbitrary `Sequence`.

lalalune (changes requested):
The early count cap is correct, but two hostile-boundary gaps remain before merge. `_bounded_pair_sequence_length` uses `type(value) not in allowed`; a subclass with a hostile metaclass can run `__eq__` during tuple membership. Use identity-only admission (`actual_type is tuple` / exact list when allowed). More importantly, both public count helpers only bound the outer collection and then unpack/compare every pair without exact pair/member validation; hostile pair subclasses or int subclasses reach iteration/min/comparison hooks. After the outer cap, snapshot/validate each element as an exact 2-tuple of exact bounded ints before any set/min/comparison work (and exact-gate `relevant_dim`/`input_dim` before comparisons). Add zero-hook regressions at both counter entry points and keep the last-fit/first-overflow tests. The constructor `_integer_pairs` path is otherwise directionally sound.

lalalune:
Deep audit found two residual preflight escapes in the submitted cap: `type(value) not in (...)` could invoke hostile metaclass equality, and the public counters traversed/compared/hash nested pair values before exact integer-pair admission. Repaired in-place at `7f59aa25`: identity-only outer dispatch, exact tuple/int leaf validation before comparisons, and zero-hook regressions. Verification: 29 focused scale-robust tests pass; targeted Ruff, strict mypy, and diff-check pass. The 4096 cap remains far above the frozen archive (24/91). This registered-source edit must leave the pinned scale-robust claim transparently source-stale; no artifact/output was changed.

lalalune:
The nested pair and metaclass fixes at `7f59aa25` are correct. One item from the prior review remains: `relevant_dim` and `input_dim` are still used in integer comparisons without exact host admission. An int subclass can therefore execute comparison hooks after the pair collection passes. Exact-gate both parameters (`type(...) is int`, with the existing supported nonnegative/domain relationship) before pair traversal, and add a zero-hook regression at both public count functions. Then the boundary is complete.

lalalune:
Expanded exact-head verification: all seven scale-robust feature/artifact/CLI/hostile modules pass, 111/111 in 7.18s. No frozen artifact bytes were modified; registry source drift remains expected.

lalalune:
Exact follow-up `7556228333f9bee36cfe955c7937fc51e33c1f0d` re-audited. It preserves the repaired identity-first runtime behavior, narrows the public annotation to the exact tuple/list contract already enforced, replaces allocation-heavy 25M pointer-repeat timing fixtures with the first-rejected 4097 boundary, and retains both hostile zero-hook regressions. Runtime source semantics are unchanged from the 111-test exact panel; diff-check is clean. Auto-merge re-enabled behind CI.

lalalune:
Superseded by #2106 after maintainer audit. The upstream successor preserves both contributor commits, retains the exact outer/nested identity hardening, documents the 4,096 bound against the frozen resource envelope, aligns public annotations, and replaces the 25-million-pointer timing test with deterministic low-resource poison-element coverage. This leaves exactly one auto-armed survivor.

PR #2105, opened 2026-08-20
fix(benchmarks): parse Foragax probe metadata strictly

lalalune:
Supersedes closed #2102 on current main.

- rejects duplicate keys/non-finite values in the image runtime probe
- uses the existing strict decoder for host package-freeze `direct_url` metadata while preserving malformed-input redaction
- embeds a standalone strict decoder in the generated distribution probe, avoiding the prior head's undefined repository-helper call
- adds behavioral and wiring regressions for all three boundaries

Validation: 11 focused image-helper/probe tests passed; the adjacent strict JSON/archive contract passed; Ruff, strict source mypy, and diff checks passed. No valid-JSON behavior changes.

### DIFF
--- a/alberta_framework/benchmarks/_official_foragax_image_helper.py
+++ b/alberta_framework/benchmarks/_official_foragax_image_helper.py
@@ -80,13 +80,15 @@ def _sha256(path: Path) -> str:
     return digest.hexdigest()
 
 
-def _strict_json(path: Path) -> dict[str, Any]:
+def _strict_json_text(value: str | bytes, *, label: str) -> dict[str, Any]:
+    """Decode one strict JSON object from a trusted bounded producer."""
+
     def pairs(items: list[tuple[str, Any]]) -> dict[str, Any]:
         result: dict[str, Any] = {}
         for key, value in items:
             host_key = _require_exact_str("key", key)
             if host_key in result:
-                raise ImageHelperError(f"{path} repeats JSON key")
+                raise ImageHelperError(f"{label} repeats JSON key")
             result[host_key] = value
         return result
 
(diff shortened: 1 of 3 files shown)


lalalune:
Blocking resource boundary: the new strict decoders still accept unbounded producer/package metadata before parsing. `_runtime_probe` sends unbounded `completed.stdout` into `_strict_json_text`; generated `distribution.read_text("direct_url.json")` and host `_sanitize_package_freeze_line` likewise parse/retain arbitrary-length direct_url text, and the invalid generated path stores the entire stripped payload. Add exact str/bytes admission plus explicit byte caps before every parse; catch `RecursionError`; bound package count and cumulative UTF-8 before sorting/sanitizing; require exact strings before partition. In the generated probe, cap direct_url before parsing and redact malformed JSON rather than retaining raw content. Add oversized/deep JSON tests at image stdout and direct_url/package-freeze boundaries. Preserve duplicate/nonfinite behavior; do not refresh Forager source hashes until all source work settles.

lalalune:
Closed as superseded by the contributor-preserving stronger lineage in #2108. The successor is still held for the bounded-input/resource corrective described in its audit; this closure is deduplication, not approval of the current successor head.

lalalune:
Independent overlap audit complete at exact `e76cd0dca4cf4de440cc9bdd79f19428f2e036c7`. This is the stronger narrow successor to #2102 and #2108: its generated decoder is self-contained (no host-only names), its helper is behaviorally executed in tests, the image probe shares the established strict decoder, duplicate metadata fails closed, malformed syntax retains the existing redaction path, and NaN/Infinity preserve the existing explicit validation error. Full adjacent Foragax/OCI/image lane: 78 passed in 147.71s. Targeted/hostile subset: 12 passed. Ruff clean; strict mypy clean (221 files); diff-check clean. #2108 is closed so exactly this survivor remains.

lalalune:
Superseded by contributor-preserving #2108. That lineage retains the original contributor commit, current-main isolation, all three strict-JSON repairs, and the additional exact byte/count/depth bounds requested here. Closing so only one implementation can merge.

PR #2106, opened 2026-08-20
fix(evaluation): bound pair-sequence validation before traversal

lalalune:
Supersedes #2103 after maintainer audit while preserving contributor commits 6e290bbb and 7f59aa25.

The production repair caps exact tuple/list pair collections at the repository-wide 4,096-item host traversal ceiling before element access; validates exact nested tuple/int identities before comparison or hashing; and keeps more than 45x headroom over the frozen 91-pair candidate archive and 170x over each 24-pair active bank. Exported helper annotations now match the exact accepted containers.

The successor replaces the 25,000,000-pointer wall-clock regression (421 MiB observed focused-test RSS) with deterministic poison-element last-fit/first-reject tests. Main reproduction independently measured 4.82s in the counter and 5.33s in `ConditionSeedRecord` for the reported pointer repeat.

Validation: 27 focused tests passed; Ruff, strict mypy, and diff checks clean. The changed protocol source is registered evidence: no pinned output is modified, and current evidence remains fail-closed/invalid pending a separately authorized frozen rerun. This is validation hardening, not new performance or scientific evidence.

Co-authored-by: ss251 <ss251@uw.edu>

### DIFF
--- a/alberta_framework/evaluation/scale_robust_feature.py
+++ b/alberta_framework/evaluation/scale_robust_feature.py
@@ -264,6 +264,22 @@ def _integer_pairs(
     return tuple(pairs)
 
 
+def _pair_counter_dimensions(*, relevant_dim: object, input_dim: object) -> tuple[int, int]:
+    """Admit exact nonnegative dimensions before pair comparisons."""
+
+    if type(relevant_dim) is not int:
+        raise TypeError("relevant_dim must be an exact int")
+    if type(input_dim) is not int:
+        raise TypeError("input_dim must be an exact int")
+    normalized_relevant_dim = relevant_dim
+    normalized_input_dim = input_dim
+    if normalized_relevant_dim < 0:
+        raise ValueError("relevant_dim must be nonnegative")
+    if normalized_input_dim < normalized_relevant_dim:
+        raise ValueError("input_dim must be at least relevant_dim")
+    return normalized_relevant_dim, normalized_input_dim
+
+
 def _resource_accounting_map(value: object) -> dict[str, dict[str, int | bool]]:
     if type(value) is not dict:
         raise TypeError("memory_by_condition must be an exact dictionary")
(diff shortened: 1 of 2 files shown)


lalalune:
Blocking final host gate: both exported counter functions still compare pair leaves against caller-supplied `relevant_dim` / `input_dim` without exact admission. An int subclass in either argument can execute comparison hooks after pair normalization. Exact-gate both dimensions before pair traversal (`type is int`, supported nonnegative/domain relationship) and add zero-hook tests at both functions. This was the last unresolved item on #2103 and must carry into the successor.

lalalune:
Exact successor audit CLEAR at `265e3d53534862de45cfb48c96ad589912b5b20b`. Contributor and corrective lineages are preserved; runtime semantics match the independently repaired identity-first cap. Exact combined scale-robust panel: 109 passed; targeted Ruff, strict mypy, and diff-check pass. No output/artifact changed; registered evidence remains transparently source-stale.

lalalune:
Exact-head audit found the remaining hostile-dimension escape noted on the superseded PR: `relevant_dim` and `input_dim` were compared before exact host admission. Fixed at `55ffa23f` with exact-int/nonnegative/domain validation before pair traversal and zero-hook regressions across both public counters. Verification: 92 focused/adjacent tests pass; targeted Ruff, strict source mypy, and diff-check pass. No Markdown or output artifact changed.

lalalune:
Exact follow-up `55ffa23f88e72d7731c70d929be4c40becf65650` clears the remaining public dimension boundary: exact built-in, nonnegative `relevant_dim`/`input_dim` admission occurs before comparison, with hostile-int zero-hook coverage. Exact focused panel 31 passed; Ruff, strict mypy, diff-check pass. Auto-merge re-enabled behind required CI.

lalalune:
Independent exact-head audit GREEN at `55ffa23f88e72d7731c70d929be4c40becf65650`. The successor now preflights the outer pair count before traversal, uses identity-only container dispatch, requires exact nested tuple/int identities before comparison or hashing, routes `relevant_dim`/`input_dim` through the exact bounded dimension gate, and replaces the prior 25M-element allocation test with resource-safe boundary/hostile cases. Exact head verification: 27 focused tests passed; Ruff clean; strict module mypy clean. No evidence artifact was edited; registered-source drift remains the expected fail-closed consequence. MERGE once required CI is green.

PR #2104, opened 2026-08-20
research(bimu): freeze fresh-process matched campaign

lalalune:
## Summary

- freezes a permanently nonpromoting five-task BiMU memory-off/candidate comparison on development seeds 157001–157003, consumed for promotion
- keeps both literal authorization gates false; execution and mutation remain unavailable until a separately reviewed transition
- binds canonical MNIST, source/runtime/dependencies, authorization, schedule, initial state, counters, numeric resources, and telemetry
- reserves exact registered destinations before work and publishes strictly validated immutable inodes without replacement
- loads the frozen arrays once; completed-shard admission independently reexecutes the exact seed/arm on those arrays before linking, while experimental counters remain those of the retained primary result
- retains a generic validated receipt for ordinary execution/replay-admission failures when possible; after first dispatch, later unretained failures attempt to preserve the owned reservation as a durable consumed-without-result tombstone
- exposes plan, run-shard, summarize, and validate through `asi-bimu-matched-development`

This bounded prerequisite is not paper-comparable, scientific evidence, a retained result, or completion of #1570. Replay is validation overhead; consistency hashes and process identity are not authenticated execution attestation. Tombstone I/O, another asynchronous interruption during conversion, and process death are not guaranteed.

## Audit evidence

- exact reviewed head: `9e164cdac03a00a2488b15d716adc1dabedfc429`
- authorization remains false; no canonical campaign shard, workload, or output was executed
- 68 focused BiMU/campaign tests passed
- scoped Ruff, strict mypy across 222 source files, and diff check passed
- no output or registered evidence artifact changed

PR #2107, opened 2026-08-20
research(bimu): add gated matched development runner

lalalune:
## Scope

One issue and one contract only: the bounded, permanently nonpromoting BiMU matched development runner for #1570.

- consumes the already-merged frozen three-seed/two-arm plan
- matches dataset, schedule, initial state, counters, and numeric resources
- strictly reexecutes every seed/arm and compares every non-timing field, rejecting rehashed metric forgeries
- preserves the plan's `pyproject.toml`/`uv.lock` identity
- keeps public execution and publication fail-closed behind a separate false authorization-transition gate
- publishes only through a consumer-backed pinned-dirfd helper with deterministic `O_EXCL` reservation before replay, no-replace link, fsync, bounded `O_NOFOLLOW` reread, byte equality, JSON reload, and strict validator replay

## Seed status

Seeds `157001`, `157002`, and `157003` are exposed/frozen development seeds and therefore consumed for any promotion purpose. They have not produced a retained matched result. Invalid-attempt seed `23` remains excluded. This PR performs no campaign execution and creates no output.

## Authorship

The two-commit history preserves the source lineage: the extracted runner commit retains Shaw Walters's authorship from the integration source, and the strict gate/publication correction retains lalalune's authorship.

## Verification

- `13 passed` (`test_bimu_matched_runner.py` plus the frozen-plan tests)
- Ruff passes on the runner, publication helper, and runner tests
- mypy: `Success: no issues found in 223 source files`
- diff check clean

No closing keyword: #1570 remains open until an independently audited authorization transition, valid retained report, and every named acceptance gap are complete.

Refs #1570

PR #2110, opened 2026-08-20
research(bimu): complete strict matched-development campaign

lalalune:
## Summary

Contributor-preserving successor to closed #2104 and #2107 for issue #1570. It keeps the stronger six-shard fresh-process campaign architecture and incorporates the useful strict-reexecution property from the alternate lineage.

- fixed three-seed/two-arm roster and fixed append-only development namespace
- exact source/runtime/dependency/OpenML-slice identities, counters, resources, initial state, schedule, and paired outcome rule
- strict validation reexecutes each exact seed/arm against the digest-bound arrays and compares every deterministic result field, rejecting self-consistent metric/final-state forgeries
- CLI summarize/validate use that strict replay path before aggregate admission
- output parents are opened component-by-component with pinned directory descriptors and no symlink following
- publication uses a fully written/fsynced read-only anonymous O_TMPFILE inode and no-replace linkat; visible artifacts are never unlinked on a later sync failure
- dataset identity is rechecked after both execution and validation replay

Seeds 157001-157003 were already publicly frozen by the merged prospective plan and are therefore consumed for any promotion purpose. They have not produced a retained matched result. This lane remains permanently nonpromoting, execution authorization remains literally false, no workload was executed, and no output was created. Issue #1570 stays open.

## Audit evidence

- Real `load_mnist_train()` OpenML-v1 materialization reproduced frozen slice SHA-256 `85c681c2f5fc5c274870b30c9accb3d2a6e9eb90a4575a2bf1ccca64f58b6227`
- Official repository/commit/license independently verified
- 45 tests passed across BiMU kernel, plan, campaign, hostile validators, strict replay, and publication
- scoped Ruff passed
- strict mypy passed
- diff-check passed

No registered evidence artifact or output is changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added the execution-gated `asi-bimu-matched-development` campaign for the
+  bounded five-task BiMU mechanism-on versus memory-off comparison. Its fixed
+  six-shard namespace, fresh-process roster, paired outcome rule, validators,
+  and resource/provenance bindings are permanently nonpromoting and not
+  paper-comparable; execution remains explicitly unauthorized.
 - Added the unfrozen `preview1` reference-agent transaction ledger, primitive
   Prototype adapter, aggregate SwitchingTwoState/RiverSwim life runner, and
   development-only quiescent checkpoint codec. These L0 mechanisms enforce
--- a/alberta_framework/evaluation/bimu_matched_campaign.py
+++ b/alberta_framework/evaluation/bimu_matched_campaign.py
@@ -1,0 +1,1147 @@
+"""Fresh-process execution and validation for the bounded BiMU comparison.
+
+The campaign is permanently nonpromoting and not paper-comparable.  Its
+execution gate is deliberately closed in the reviewed source until a separate
+authorization change is accepted.  Timing is retained only as telemetry and
+never contributes to the paired outcome.
+"""
+
+from __future__ import annotations
+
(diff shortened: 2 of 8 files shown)


lalalune:
Maintainer deep audit: this is the sole consolidated BiMU lineage. It preserves #2104 contributor commit c87485c7, closes #2104 publication/replay gaps, and adopts only #2107’s sound strict-reexecution concept without its single-process, arbitrary-destination, dual-authorization, or missing-lock-identity regressions. Exact current head validated as documented; recommend merge after required CI.

lalalune:
HOLD / exact-head audit found three remaining integrity gaps at `71bb1c14`: (1) shard execution occurs before any deterministic O_EXCL reservation, and `publish_json` validates/reexecutes before opening an anonymous inode, so concurrent launches can both consume the frozen roster before one loses the final link; reserve the exact canonical shard destination before dataset load/execution and hold it through no-replace publication with inode-safe cleanup. (2) `_open_output_parent` accepts an arbitrary caller root and `_allowed_path` merely relocates the namespace under that root; bind the exact registered repository root and literal output namespace via no-follow dirfd traversal. (3) the write loop must fail if `os.write` returns 0. Preserve the strict dataset-bound non-timing reexecution, uv.lock identity, dual literal/plan authorization gates, and explicitly consumed (not fresh) 157001–3 status. No execution/output; #1570 remains open.

lalalune:
Superseded by the authoritative original contributor lineage #2104 at 787b37cb. Its strict-reexecution and post-dataset checks are now present there, while #2104 also retains the stronger concurrent pinned-parent publication and nonregular-entry repair.

Issue #1581, opened 2026-08-18
benchmarks(crl): integrate CORA task suites and metrics


lalalune:
### Objective

Qualify one small CORA suite first, then its Atari, Procgen, NetHack, and CHORES families.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2110.10067
- https://github.com/AGI-Labs/continual_rl

lalalune:
Exact official-source audit at `AGI-Labs/continual_rl@f2754bb282757829765beb4703f24b87efa13ff9` confirms the next real Procgen provider cannot yet be content-closed without a reviewed compatibility/runtime plan:

- `environment.yml` specifies only `python>=3.7`, `pytorch>=1.7`, and `torchvision>=0.5.0`; `setup.py` leaves NumPy/TensorBoard/torch-ac/moviepy/dotmap/psutil/OpenCV unpinned and specifies `gym[atari]<=0.25.2`.
- `docs/BENCHMARK_INSTALL.md` says only `pip install procgen`; the official revision does not pin a Procgen version, wheel/archive, compiled asset bytes, or license digest, and Procgen is absent from `setup.py`.
- The official six-task constructor does bind 200 training levels at `start_level=0` and the full evaluation distribution, but environment creation normally uses `create_seed=True`, sourcing a seed from `os.urandom`. Therefore invented fixed smoke seeds are not official provider semantics.
- A truthful next slice must first freeze a reviewed Linux/Python/Torch/Gym/Procgen compatibility matrix; bind the exact source archive/tree/license plus Procgen wheel, compiled assets, and rights; specify how deterministic seed injection relates to the official `seed_to_set` path; and define create-only durable success/failure receipts whose observation/reward/termination payload is strictly reconstructible.

No external run, output, source/runtime readiness claim, or issue closure is made. The earlier caller-supplied R0 schema in closed PR #2111 will not be resurrected.

hermesagent270-commits:
**Partially implemented — only the small suite is qualified.**

[...]

PR #2111, opened 2026-08-20
research(cora): add honest external provider contract

lalalune:
## Summary

Adds a narrow R0 provider-contract adapter for the first CORA Procgen qualification boundary in #1581. It freezes the audited paper/code identity, source/runtime/asset declaration schema, six-task train/evaluation seed order, fixed-action mechanism-off trace shape, resource fields, current-tree provenance, and strict receipt validation.

The adapter does not import or launch CORA or Procgen. Trace arrays are supplied by the caller, and every accepted receipt is explicit that `external_runtime_execution_verified=false`. Matching arrays or hashes cannot be turned into an external-execution, parity, performance, or promotion claim.

## Open gates

The blocker manifest keeps the official checkout verification, isolated runtime lock, Procgen distribution/assets/rights, actual official-provider trace, paper metric parity, matched learning baselines/resource contract, published-scale execution, and separate external-execution authorization open. Issue #1581 remains open. No external run or output artifact is included.

## Validation

- 33 focused CORA/external-qualification tests passed
- Ruff passed
- strict mypy passed for both source modules
- diff check passed

Refs #1581.

### DIFF
--- a/alberta_framework/benchmarks/cora_external_qualification.py
+++ b/alberta_framework/benchmarks/cora_external_qualification.py
@@ -1,0 +1,947 @@
+"""Fail-closed external CORA Procgen qualification contract.
+
+This module never imports or launches CORA, PyTorch, Gym, or Procgen.  An
+isolated provider may use the frozen plan to produce one bounded fixed-action
+trace.  ASI validates that trace and its declared source/runtime/asset identity
+without confusing it with the native recurring-bandit analogue or a paper run.
+"""
+
+from __future__ import annotations
+
+import argparse
+import dataclasses
+import hashlib
+import json
+import math
+import operator
+import sys
+from typing import Any, SupportsIndex, cast
+
+import numpy as np
+
+import alberta_framework.benchmarks.external_qualification as external_qualification_module
+from alberta_framework.benchmarks.qualification_provenance import (
+    QualificationIdentity,
+    collect_qualification_identity,
+    identity_from_payload,
+    require_current_identity,
+)
+
+SCHEMA = "asi.cora.procgen.fixed_action_smoke.v1"
+QUALIFICATION_ISSUE = 1581
+QUALIFICATION_LANE_ID = "cora"
+PAPER_REVISION = "arXiv:2110.10067v2"
+CORA_REPOSITORY = "https://github.com/AGI-Labs/continual_rl.git"
+CORA_COMMIT = "f2754bb282757829765beb4703f24b87efa13ff9"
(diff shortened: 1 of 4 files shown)


lalalune:
HOLD / deep audit found a trust-boundary blocker in the new shared provenance surface. `qualification_provenance._canonical` admits arbitrary `Mapping` implementations and traverses them without aggregate node/depth/container/text/int/finiteness budgets. `identity_from_payload` and the CORA source/plan/runtime/assets/receipt reconstructors call `set(value)` before exact-string key admission, so hostile string-subclass keys can execute hash/equality hooks. Require exact built-in containers, bounded preflight before semantic/hash/sort work, and tuple-snapshot/exact-str admission of every key before frozenset/schema comparison; add zero-hook nested-key and traversal-bomb regressions. The provider-contract scope and `external_runtime_execution_verified=false` policy are otherwise honest; no external run/output and #1581 remains open.

lalalune:
Closing as misguided at this stage after deep audit. The implementation is careful about JSON shape and explicitly refuses external-execution/parity claims, but it adds ~1,100 lines of provider schema and tests with no external provider, retained artifact, or non-test consumer. That does not close #1581’s next integration gate.

Material contract problems:
- `negative_outcome_retained=True` is asserted on an in-memory object; there is no durable publication path, and builder/validation failures retain nothing.
- source checkout, runtime isolation, asset rights, and persistent environment bytes are caller-declared while the blocker manifest correctly says those verifications remain open.
- observation/reward/termination arrays are discarded; the serialized receipt retains only caller-computed hashes, so validation cannot reconstruct them. The `outcome` is likewise caller-selected and has no metric rule for this 12-step fixed-action trace.
- no provider consumes the plan, no CLI emits a provider plan or accepts a provider receipt, and the existing CLI only repeats already-recorded blockers.
- the proposed train/evaluation level seeds are qualification inventions, not verified provider semantics; Procgen distribution construction/reset/step behavior is still absent.

Independent verification found the paper-level Procgen constants broadly accurate (six tasks, 5M frames/task/cycle, five cycles, 200 train levels, full eval distribution, 64x64 RGB, 15 actions), and 49 focused/adjacent tests plus Ruff, strict mypy, and diff checks pass. Those strengths should be reused in a future isolated-runtime/provider slice that actually loads/resets/steps official Procgen, publishes both success and failure receipts durably, and validates retained trace payloads. Until then, the existing external-qualification registry and CORA blocker document are the honest contract.

PR #2099, opened 2026-08-20
test(steps): cover every Step 1 normalizer dispatch

lalalune:
Supersedes #2087 after maintainer review.

Retains the all-four-normalizer factory/Step 1 smoke contract and removes four unrelated endpoint-only tests from this focused consolidation; endpoint behavior remains covered by the existing dedicated production suites. This is test-only and does not alter evidence or production source.

Validation at exact head `923c89288e59fb7c11a630ae88c79e41314f6a53`: full `tests/test_production_steps.py` passed (390 tests); Ruff and diff checks passed.

PR #2112, opened 2026-08-20
test(steps): preserve Step 1 normalizer boundary dispatch

lalalune:
Contributor-preserving successor to #2087/#2099.

Covers all four public Step 1 normalizer factories end to end. The EMA decay 0/1 and streaming-batch momentum 0/1 cases also prove that each selected endpoint reaches the matching normalizer and the real Step 1 smoke loop. The six-case matrix avoids duplicate default smoke runs and uses typed literals without dynamic kwargs or type-ignore bypasses.

Verification at exact head `59d7c73898989f4195dafd99e3ea9d8c8eff8e03`:

- `tests/test_production_steps.py`: 392 passed
- repository Ruff passed
- strict mypy passed
- diff check passed

This is test-only; it changes no production source, benchmark output, workflow, or evidence artifact.

### DIFF
--- a/tests/test_production_steps.py
+++ b/tests/test_production_steps.py
@@ -83,6 +83,64 @@ def test_step1_kernel_all_public_optimizers_smoke(optimizer: str) -> None:
     assert result.metrics_shape == (12, 4)
 
 
+@pytest.mark.parametrize(
+    "normalizer",
+    ["none", "ema", "welford", "streaming_batch"],
+)
+def test_step1_kernel_all_public_normalizers_smoke(normalizer: str) -> None:
+    """Every documented Step1NormalizerName must actually run end to end.
+
+    Before this test, ``run_step1_smoke``/``make_step1_normalizer`` were only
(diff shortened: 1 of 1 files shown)


lalalune:
Independent successor audit GREEN at exact head `7adc8a4b82dd9d7a9f79e0e99b2ce81064a89829`. Its `tests/test_production_steps.py` tree is byte-identical to the previously verified contributor-preserving `cbc807b4` head: all four public normalizers plus the selected EMA and streaming-batch endpoint consumption checks are retained. Exact-equivalent full file verification was 394 passed; Ruff clean. Test-only, no production/evidence/output mutation. #2099 is correctly closed as superseded. MERGE once required CI is green.

lalalune:
Updated exact-head check at 51361421: the one-line assertion only narrows the already selected optional implementation before to_config access; it does not alter production behavior. The parameterized Step 1 normalizer endpoint test passes all 6 cases and scoped Ruff is clean. Prior full 394-test production-steps audit remains applicable. APPROVE once CI is green.

PR #2113, opened 2026-08-20
research: freeze Intentional Updates matched-development plan

lalalune:
## Summary

Prospective, permanently nonpromoting #1561 infrastructure only. This PR does not execute a campaign, create an output, or close the issue.

- retains the existing batch-size-one supervised adapter and adds end-to-end linear TD(0), TD(lambda), and Q(lambda) consumers using the pinned author equations
- pins `arXiv:2604.19033v1` and official code commit `e86e26fd8613ac212e9a52c3fed8a01d0a31f685`, while recording all supervised and linear-control protocol differences
- provides exact mechanism-off reductions to fixed-step TD/control consumers
- freezes separate within-family matched schedules and paired Bonferroni questions with explicit persistent-byte, environment/data, model/action, RNG, solve, and telemetry accounting
- binds current source/runtime identities and ASI's retained canonical OpenML materialization digests
- explicitly quarantines draft seeds `31561001..31561004`, which contract tests consumed; replacement campaign seeds `41562001..41562004` had no fetched-history or local-worktree occurrence before freeze and tests never execute them
- gives tests a distinct test-only roster and identity-checked private capability
- requires both an exact literal reviewed-transition flag and a separate runtime authorization flag before reservation, dataset loading, consumer dispatch, or public publication; changing only the runtime flag remains fail-closed
- bounds validator-controlled reexecution before loops and recomputes all child identities and paired arithmetic
- reserves the exact NEW output through per-segment pinned no-follow dirfds, uses deterministic `O_EXCL` reservation, inode-safe cleanup, no-replace unique-link publication, fsync, stable bounded reread, duplicate-key rejection, `RecursionError` translation, and strict report revalidation

The issue remains open because no retained matched campaign has run.

## Verification

- failing-test-first commits precede both implementation and audit corrections
- final focused campaign suite: 28 passed
- related campaign/supervised/comparator/entrypoint/package suite before correction: 67 passed; correction only further closes public execution/publication and separates test seeds
- Ruff clean for the campaign and dedicated tests
- strict mypy clean for the campaign
- `git diff --check` clean

## Overlap audit

No open PR modifies the new campaign module, dedicated test, or research note. Shared entrypoint inventory and project metadata receive additive registrations only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add a prospectively frozen, execution-disabled Intentional Updates matched-development
+  campaign spanning the existing batch-size-one supervised adapter plus new end-to-end linear
+  TD(0), TD(lambda), and Q(lambda) consumers. The two workload families retain separate paired
+  questions, exact resource receipts, telemetry-only timing, immutable negative outcomes, and a
+  permanently nonpromoting report policy. No campaign run or output is included.
+
 ### Added
 
 - Added the unfrozen `preview1` reference-agent transaction ledger, primitive
--- a/README.md
+++ b/README.md
@@ -242,6 +242,7 @@ The console scripts are grouped by responsibility; every command supports `--hel
 | `alberta-step1-smoke`, `alberta-step2-smoke` | Nonpromoting mechanism smoke checks |
 | `asi-native-supervised-catalog`, `asi-plasticity-diagnostic` | Inspect or run bounded, permanently nonpromoting supervised diagnostics |
 | `asi-adamo-diagnostic` | Catalog or run the bounded, permanently nonpromoting AdamO dynamical-isometry comparator |
+| `asi-intentional-updates-matched-development` | Inspect the frozen, execution-disabled supervised + TD/control Intentional Updates development plan |
 | `asi-clear-qualification` | Verify local CLEAR archive identities and emit a permanently nonpromoting matched setup plan |
 | `asi-coom-qualification-smoke` | Inspect or run the bounded, synthetic, permanently nonpromoting COOM qualification smoke |
 | `asi-cora-catalog` | Inspect the blocked CORA qualification catalog without importing external runtimes |
--- a/alberta_framework/benchmarks/intentional_updates_control.py
+++ b/alberta_framework/benchmarks/intentional_updates_control.py
@@ -202,9 +202,7 @@ def frozen_plan() -> dict[str, object]:
             "RNG operations, optimizer solves, persistent bytes, and timing per shard"
         ),
         "output_path": "outputs/intentional_updates_matched_development/report.v1.json",
-        # These are historical properties of the plan at freeze time.  A later,
-        # independently reviewed code-only authorization must not mutate the plan
-        # that was reviewed before any retained execution.
+        # Runtime authorization is separate from this immutable plan identity.
         "execution_authorized": False,
         "reviewed_execution_transition": False,
         "execution_status": "blocked_pending_independent_plan_audit",
--- a/alberta_framework/benchmarks/intentional_updates_control.py
+++ b/alberta_framework/benchmarks/intentional_updates_control.py
@@ -120,8 +120,19 @@ def frozen_plan() -> dict[str, object]:
         "plan_id": PLAN_ID,
         "paper_revision": PAPER_REVISION,
         "official_code_revision": OFFICIAL_CODE_REVISION,
-        "source_identity": _source_identity(),
-        "runtime_identity": _runtime_identity(),
+        "source_identity_policy": {
+            "binding": "execution-time Git/source hashes retained in every report",
+            "hashed_files": [
+                "alberta_framework/benchmarks/intentional_updates_control.py",
+                "alberta_framework/benchmarks/ipmnist_screening.py",
+                "alberta_framework/benchmarks/plasticity_comparators.py",
+                "alberta_framework/benchmarks/upgd_ipmnist.py",
+            ],
+        },
+        "runtime_identity_policy": (
+            "retain the exact execution Python, platform, package, JAX backend, "
+            "device, and configuration identity in every report"
+        ),
         "seeds": list(CAMPAIGN_SEEDS),
         "quarantined_test_consumed_seeds": list(QUARANTINED_SEEDS),
         "protocol_families": ["supervised_ipmnist", "td_control"],
@@ -153,15 +164,22 @@ def frozen_plan() -> dict[str, object]:
             "trace_decay": 0.8,
             "epsilon_greedy": 0.1,
         },
-        "matched_axes": [
-            "seed",
-            "initial_parameters",
-            "observations",
-            "updates",
-            "allowed_boundary_and_task_information",
-            "environment_definition",
-            "exogenous_exploration_rng",
-        ],
+        "matched_axes": {
+            "all_families": [
+                "seed",
+                "initial_parameters",
+                "observations",
+                "updates",
+                "allowed_boundary_and_task_information",
+            ],
+            "supervised_ipmnist": ["example_schedule"],
+            "td_prediction": ["transition_schedule"],
+            "q_lambda_control": [
+                "environment_dynamics",
+                "agent_rng_root_and_index_schedule",
+                "policy-dependent realized trajectories are not matched",
+            ],
+        },
         "boundary_information": [],
         "task_information": [],
         "paired_metrics": {
@@ -184,14 +202,12 @@ def frozen_plan() -> dict[str, object]:
             "RNG operations, optimizer solves, persistent bytes, and timing per shard"
         ),
         "output_path": "outputs/intentional_updates_matched_development/report.v1.json",
-        "execution_authorized": _EXECUTION_AUTHORIZED,
-        "reviewed_execution_transition": _REVIEWED_EXECUTION_TRANSITION,
-        "execution_status": (
-            "authorized_after_separate_review"
-            if _REVIEWED_EXECUTION_TRANSITION is True
-            and _EXECUTION_AUTHORIZED is True
-            else "blocked_pending_independent_plan_audit"
-        ),
+        # These are historical properties of the plan at freeze time.  A later,
+        # independently reviewed code-only authorization must not mutate the plan
+        # that was reviewed before any retained execution.
+        "execution_authorized": False,
+        "reviewed_execution_transition": False,
+        "execution_status": "blocked_pending_independent_plan_audit",
         "development_only": True,
         "scientific_promotion_allowed": False,
         "negative_outcomes_retained": True,
@@ -426,7 +442,7 @@ def _run(
     if intentional:
         numeric_bytes += int(
             second.nbytes
-            + 3 * np.dtype(np.float32).itemsize
+            + 2 * np.dtype(np.float32).itemsize
             + np.dtype(np.int32).itemsize
         )
     if root is not None:
--- a/alberta_framework/benchmarks/intentional_updates_control.py
+++ b/alberta_framework/benchmarks/intentional_updates_control.py
@@ -83,6 +83,7 @@ _MAX_JSON_NODES: Final[int] = 200_000
 _MAX_STRING_BYTES: Final[int] = 2 * 1024 * 1024
 _MAX_TIMING_NS: Final[int] = 7 * 24 * 60 * 60 * 1_000_000_000
 _MAX_REPORT_BYTES: Final[int] = 64 * 1024 * 1024
+_MAX_RUNTIME_DEVICES: Final[int] = 64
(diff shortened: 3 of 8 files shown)


lalalune:
HOLD / deep audit found three pre-execution integrity blockers. (1) all proposed seeds 31561001–31561004 are actually executed repeatedly by the focused tests through `_run_control_shard_authorized`; they are consumed test/development seeds, not an untouched campaign roster. Quarantine them, allocate globally absent seeds, and make tests execute a distinct explicitly test-only roster/path. (2) public `publish_report` can create the canonical retained output while execution is false from caller-supplied offline records; hard-disable public publication and make only the authorized aggregate consume a private capability publisher after its gate/reservation. (3) a single mutable `_EXECUTION_AUTHORIZED` flag dynamically changes the plan; bind a separate literal reviewed-transition flag and require both exact True before reservation/data/runner. Add zero-output regressions and catch RecursionError in strict reread. No run/output; #1561 remains open.

lalalune:
HOLD / no execution. Initial trust-boundary review found the publisher is not yet fail-closed. After the temporary inode is linked to the immutable destination, any strict-reread or semantic-validation failure leaves that invalid destination permanently visible; cleanup removes only the temporary name. Record the exact inode, require the expected link count/stability during reread, and on post-link failure unlink the destination only when its descriptor identity still matches the inode this call published. Also convert deeply nested JSON RecursionError at the retained-report loader boundary to the advertised ValueError. Add deterministic stalled/forged post-link and deep-JSON regressions without timing or giant allocations. The four seed roots are already exercised by private contract tests, so describe them as publicly frozen/consumed development roots, not fresh or untouched. Do not run or auto-merge this head.

lalalune:
HOLD on exact `c9951e924c375df0962c2cea4f855d08cc23f7fc`; no run/output. The seed/test split, two-gate authorization, public-publisher closure, and `RecursionError` conversion address part of the prior feedback, and the pinned optimizer equations match author commit `e86e26f`. Four blockers remain:

1. The Q(lambda) pair is not matched on the plan/issue's declared observations or transition schedule. Each arm selects epsilon-greedy actions from its own changing weights, so actions, successor states, and rewards can diverge. Use one common behavior trajectory/environment randomness for training both learners, and measure a separate learner-sensitive off-policy/exhaustive evaluation quantity; do not use same-trajectory reward, which would be identical.
2. TD(0)/TD(lambda) seeds affect only `seed % 2` and the parity action schedule. Four nominal roots therefore produce two duplicated deterministic trajectories/deltas, yet `_paired` treats them as four independent replicates with df=3. Bind independent exogenous schedule/environment variation to every seed (and keep it common within pair) before retaining this test.
3. `run_campaign` always calls `_release` after any exception, including failures after expensive execution. That frees the attempt marker and permits retry with the same committed roster, contradicting `negative_outcomes_retained`. Once execution begins, failure must leave an immutable consumed/non-result reservation.
4. After linking the temporary inode to the canonical destination, `_strict_reread` or semantic-validation failure still leaves that invalid destination visible. The new link-count check is insufficient. Hold the exact inode identity and identity-safely unlink only this call's destination on a post-link failure; add forged/stalled post-link regressions.

The repair must retain a common behavior path, learner-sensitive primary metric, independent replicate variation, fail-closed attempt consumption, and identity-safe invalid-link cleanup. Keep #1561 open and execution false.

lalalune:
Independent corrected-head audit GREEN at `c9951e92`. The test-consumed 31561001–04 roster is explicitly quarantined; production uses globally absent 41562001–04, while focused tests execute only distinct 101–104 through an identity-checked test capability. Public control, campaign, CLI, and publisher paths require both literal reviewed-transition and runtime authorization flags before reservation/data/runner; both flags are plan/report-bound. Publication now translates recursion failures, removes its temp link before strict reread, and requires one unique final link. Exact current head: 28 focused tests passed; Ruff clean; strict module mypy clean; origin/main contains no occurrence of the campaign roster. No run/output; #1561 remains open. MERGE after required CI on current main.

lalalune:
Manual exact-head reconciliation: HOLD remains authoritative at `c9951e92`; the later green note covers the authorization/seed/publication-surface repairs but not the scientific matching defects. In `_run`, Q(lambda) arms choose actions from their own learned weights, so their state/reward trajectories diverge; `mean_reward` is therefore not a matched learner-only contrast. TD arms use only seed parity for their deterministic schedule, yielding two duplicated trajectories treated as four df=3 replicates. `run_campaign` also releases the attempt marker on every post-execution failure, and `_publish_reserved` still leaves an exact invalid destination after post-link reread/semantic failure. Do not auto-merge until common within-pair behavior, seed-unique exogenous variation, a learner-sensitive Q evaluation metric, fail-closed post-start attempt consumption, and identity-safe invalid-link cleanup are all implemented and tested. No execution/output.

lalalune:
Final review repair is now pushed at `41c75462b9299f16372e9caa49a12ea7c0135568`.

The repair closes the hold items and two additional receipt-accuracy gaps:
- post-link validation failure now removes the destination only when it is still the exact inode created by this publisher;
- publication pins the temporary/destination inode and requires a single final hard link during reread;
- deeply nested JSON parser recursion is converted to `ValueError`;
- previously test-consumed roots are quarantined, tests execute only a separate test roster/capability, and the new campaign roots remain unexecuted;
- Q(lambda) no longer claims a matched transition schedule: pairs bind the same environment and exogenous exploration stream while learned actions/transitions may diverge;
- RNG uniform/integer draws and the logical optimizer step are now included in exact resource accounting;
- the canonical dataset identity is recomputed after execution to detect in-process mutation.

Independent verification on the exact head: 72 focused/adjacent tests passed; focused Ruff, strict mypy, and `git diff --check` passed. No campaign data or campaign seed was executed. The issue remains open and auto-merge remains disabled.

lalalune:
Deep independent audit completed on the current lineage through `ce694385`. I compared the linear TD/Q specialization against the pinned official `IntentionalOptimizer` at `e86e26f`, checked issue #1561, dataset derivation/source binding, fresh-seed history, execution/publication gates, nonpromotion/negative-outcome semantics, matched axes, exact RNG/resource accounting, and immutable failure handling. Repairs now keep the literal pre-execution plan invariant across host/runtime and later authorization, describe Q(lambda) matching honestly (shared dynamics/RNG schedule, not policy-dependent realized trajectories), quarantine the test-consumed draft roster, use fresh campaign seeds `41562001..4`, and correct persistent state accounting to two float32 optimizer scalars plus one int32 step. The latest contributor tombstone change was also reconciled; it preserves partially consumed schedules without retry and now validates the unique regular marker inode. Validation: dedicated suite 38 passed at the immediate predecessor plus exact-head targeted 8 passed; broader supervised/comparator/entrypoint/package lane 78 passed; Ruff clean; strict mypy clean across 222 files; diff check clean. This is prospective, execution-disabled, permanently nonpromoting infrastructure only; issue #1561 correctly remains open.

lalalune:
Final independent audit at exact 7cc47eee: APPROVE once required CI is green. The literal prospective plan now remains immutable across a later code-only authorization; execution reports bind actual source/runtime. Control matching correctly distinguishes shared exogenous Threefry variates from policy-dependent realized trajectories, accounts uniform and conditional integer draws, and retains a consumed-without-result tombstone after post-dispatch failure. Publication binds the linked inode, rolls back only its own failed publication, and requires one stable link on reread. Exact-head gates: 45 focused Intentional/console tests, scoped Ruff, and strict module mypy passed. No campaign or output was run; issue 1561 remains open.

lalalune:
Current-head provenance blocker at 33b88ecd: the common uniform behavior trajectory and Q-lambda TD-error comparison correctly repair pair matching, but retained execution identity still hashes only four Python files and omits pyproject.toml/uv.lock. Runtime identity records only Python, JAX/jaxlib, NumPy, backend, and platform, omitting the direct dependency roster and JAX config/device/environment state. Dependency/config drift therefore is neither report-bound nor detected mid-run. Please bind exact pyproject+uv.lock plus bounded full runtime/JAX/device/env identity and add mutation/drift regressions, preserving the trajectory and consumed-reservation fixes. No run/output was performed.

lalalune:
Exact delta re-review complete for `9c85e697c851d75cec636226f25c830fff9fa4cf` relative to previously cleared `33b88ecd4935d6447c1135960e4c6ecf0909ad4c`.

The delta is sound and narrow:

- both Threefry integer draws now request `np.int32` explicitly, so enabling JAX x64 cannot alter the seed-derived initial state or paired behavior trajectory;
- the added real-run regression confirms the complete trajectory is invariant under `jax.enable_x64()`;
- resource assertions retain the exact integer-draw count;
- the Q(lambda) report test now confirms matched rewards and recomputes the learner-sensitive mean-squared-TD-error deltas;
- documentation now accurately describes prediction/control error rather than control reward as the paired metric.

Exact-head verification:

- Intentional control/IPMNIST/comparator panel: **73 passed**;
- focused and full Ruff: passed;
- focused source mypy: passed;
- project mypy: **222 source files**, passed;
- delta diff-check: passed.

No workload, output, or workflow was executed. The prior methodology/resource/nonpromotion clearance remains intact.

lalalune:
Independent audit of the corrected lineage through exact head f1a994a1d7a4f72ee863c5e2e63d53e08c0c1a1e: APPROVE MERGE pending required CI. The paired TD/control trajectory correction is coherent; source identity now binds all consumed modules plus pyproject.toml and uv.lock; one bounded runtime identity binds direct dependency versions, Python/platform/byte order, JAX backend/config/environment and a one-time exact device snapshot capped at 64 before any device attributes are read. Mid-run/report validation catches source/runtime drift. Reservation precedes consumer dispatch, and post-dispatch failure retains a permanent inode-owned consumed-without-result tombstone. Exact campaign and test capabilities/seeds remain separated and both literal authorization gates remain false. Maintainer exact-head gates: 49 tests, Ruff, strict mypy; my independent diff/hostile-bound review found no residual blocker. No run/output; #1561 remains open.

lalalune:
Exact delta re-review complete for `f1a994a1d7a4f72ee863c5e2e63d53e08c0c1a1e` relative to previously cleared `9c85e697c851d75cec636226f25c830fff9fa4cf`.

The provenance expansion is sound:

- source identity now binds `pyproject.toml` and `uv.lock` bytes in addition to every consumed campaign source;
- runtime identity records exact Python implementation/version/byte order, OS release/machine, all seven direct project dependency versions plus the used scikit-learn runtime, JAX backend, a bounded device inventory, numerical/RNG configuration, and relevant JAX/XLA environment;
- the device count is bounded before any device attribute access, and the complete runtime object passes through the existing exact-container/scalar/node/depth/string budgets;
- current-source/current-runtime identities are cross-checked during shard reexecution and report validation;
- regressions cover dependency-input mutation, environment drift, oversized environment values, hostile oversized device inventories before hooks, and report rejection on runtime drift.

The previously cleared matched behavior trajectories, learner-sensitive Q metric, resources, seed isolation, failure retention, and nonpromotion policy are unchanged.

Verification on this exact head:

- Intentional control/IPMNIST/comparator panel: **79 passed**;
- full Ruff: passed;
- project mypy: **222 source files**, passed;
- delta diff-check: passed.

No workload, output, or workflow was executed.

lalalune:
CI failure classification on exact head `f1a994a1`: both py3.12 and py3.13 shard 4 reached 99% with no test assertion/error, then the hosted runner delivered a shutdown signal at the five-minute boundary. This is the same 654-node shard whose cold execution has now been measured at 263.04s / 13.8 GB with a 128.18s Dreamer test; #2133 moves that module to the required dedicated slow lane. These two failures are infrastructure/time-budget cancellations, not evidence of an Intentional Updates regression. Rebase/retry after #2133 is the clean resolution; no workload/output was run.

hermesagent270-commits (approved):
Reviewed final exact head f1a994a1d7a4f72ee863c5e2e63d53e08c0c1a1e after reconciling the superseded hold findings. Campaign/test seed rosters and capabilities are separated; both literal authorization gates remain false before reservation/data/dispatch/publication; paired control uses shared exogenous Threefry schedules with learner-sensitive error metrics; source/runtime/device/environment identities are bounded and rechecked; and post-dispatch failures retain an inode-owned consumed-attempt tombstone while failed publication rolls back only its own destination. Local verification: 49 dedicated tests passed; Ruff passed; mypy passed. No code-level blocking findings on this head. The branch currently conflicts with main and its earlier CI shard cancellations are addressed by the slow-lane classification in #2133; update/rebase and rerun required CI before merge, with re-review for conflict-resolution changes.

lalalune:
Closing as low-value campaign proliferation for the current ASI program. This PR adds a 1,431-line disabled Intentional Updates controller/campaign plus a second evaluation family; no executed comparison resolves whether this optimizer helps the live agent. I reviewed its plan, changed implementation surface, and the roadmap priorities: first execute the implemented Switching/RiverSwim scorecard, then select a measured bottleneck and integrate survivors. Another frozen, disabled campaign and its maintenance burden does not resolve that uncertainty. The mechanism can be reconsidered with a cheap matched result, a live control remeasurement, and a concrete downstream integration hypothesis; the current proposal is not selected.

PR #2119, opened 2026-08-20
research: freeze bounded optimization-readiness panel

lalalune:
## Scope

Prospective, permanently nonpromoting infrastructure for #1568. This PR does not close the issue and does not contain a benchmark result or output.

It adds one bounded local predictor-comparison contract:

- exact float64 linear squared-loss checkpoint adapter over one matched 10,000-example task
- paper-aligned OR sampling and future-gain axes: 128 size-four diagnostic batches, 128 size-four SGD rollouts, learning rate 1e-3, and 1/10/100-step horizons
- Optimization Readiness compared with its gradient-strength-only mechanism-off reduction, gradient norm, representation rank, exact curvature rank, and parameter norm
- six frozen prospective Threefry roots `2684771901`–`2684771906`, with no exact match in fetched branch history or workspace content before this plan was written
- strict source (`pyproject.toml`, `uv.lock`, helper, executor, panel), runtime/dependency, dataset, checkpoint, schedule, resource, and authorization identities
- a detached read-only plan view for audit, plus a single reserved run-and-retain transaction under the exact registered NEW namespace

## Fail-closed execution and publication

`AUTHORIZATION_TRANSITION_APPROVED` and `EXECUTION_AUTHORIZED` are separate literal `False` values. Both and the frozen plan identity must change in a separately reviewed source transition. The public runner and validator deny before case inspection, directory creation, dataset hashing, scheduling, execution, replay, or publication.

After a future authorization, the only public execution path acquires a deterministic `O_EXCL` reservation before inspecting cases and holds it through six primary executions, aggregation, no-replace publication, bounded `O_NOFOLLOW` reread, and six strict reexecutions. Publication walks from the exact registered repository root through pinned dirfds, writes only `outputs/optimization_readiness/prospective.v1/result.json`, and uses inode-safe temporary/reservation/failure cleanup.

## Deliberate non-parity

This is not the paper SCR or P-MNIST experiment. The linear input representation and Hessian are checkpoint-invariant on a matched task; eNTK, effective rank, active-neuron fraction, nonlinear architectures, P-MNIST gradient-Gram curvature, and the paper multi-run/task panels remain absent. These gaps are immutable plan fields. Any undefined/tied comparator makes the bounded outcome inconclusive. The issue remains open for a reviewed authorization, retained result, and every named paper-parity requirement.

Seeds become exposed by this PR and therefore are permanently ineligible for promotion despite remaining unexecuted. No result exists; no campaign execution or output was produced.

## Gates

- `pytest -q tests/test_optimization_readiness.py tests/test_optimization_readiness_executor.py tests/test_optimization_readiness_panel.py` — 68 passed
- focused Ruff — clean
- mypy — success, 223 source files
- package/release/import checks — 22 passed
- `git diff --check` — clean

### DIFF
--- a/alberta_framework/evaluation/optimization_readiness_executor.py
+++ b/alberta_framework/evaluation/optimization_readiness_executor.py
@@ -1,0 +1,623 @@
+"""Bounded, model-bound prospective Optimization Readiness execution.
+
+This is a permanently nonpromoting development diagnostic.  It executes a
+fully specified linear squared-loss model from an explicit checkpoint against
+one supplied 10,000-example task.  The validator repeats the execution from
+the bound inputs instead of accepting reported metrics or counters.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import math
+import os
+import platform
+import stat
+import sys
+from importlib.metadata import version
+from pathlib import Path
+from types import MappingProxyType
+from typing import Final
+
+import jax
+import jax.numpy as jnp
+import jax.random as jr
+import numpy as np
+
+from alberta_framework.evaluation.optimization_readiness import (
+    energy_rank,
+    estimate_appendix_c1_optimization_readiness,
+)
+
+EXECUTION_SCHEMA: Final[str] = "asi.optimization-readiness.execution.v1"
+AUTHORIZATION_TRANSITION_APPROVED: Final[bool] = False
+EXECUTION_AUTHORIZED: Final[bool] = False
+_OBSERVATIONS: Final[int] = 10_000
+_MAX_PARAMETERS: Final[int] = 64
+_BATCH_SIZE: Final[int] = 4
+_ROLLOUTS: Final[int] = 128
+_DIAGNOSTIC_BATCHES: Final[int] = 128
+_GAIN_STEPS: Final[tuple[int, ...]] = (1, 10, 100)
+_STEP_SIZE: Final[float] = 1e-3
+_MAX_LIVE_BYTES: Final[int] = 256 * 1024 * 1024
+_MAX_PREFLIGHT_WORK_UNITS: Final[int] = 500_000_000
+_MAX_JSON_NODES: Final[int] = 4_096
+_MAX_JSON_STRING_BYTES: Final[int] = 64 * 1024
+_SAMPLING_PROVENANCE: Final[str] = (
+    "executor_derived_independent_with_replacement_jax_threefry2x32"
+)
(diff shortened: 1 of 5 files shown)


lalalune:
Independent deep audit of current head ed6aa7ee: REQUEST CHANGES / NO MERGE yet. Five concrete gaps: (1) frozen plan says `caller_labels_are_not_content_identity=true`, but execution hashes and requires exact shared label content; make the contract truthful and pin it. (2) runtime list-comprehends unbounded `jax.devices()` before any cardinality/resource gate. (3) `_array_identity` contains a duplicate `shape` literal key. (4) panel preflight hashes/tobytes caller arrays before applying the executor exact dtype/shape/work preflight; validate all six cases before content traversal. (5) the transaction always removes its reservation on failure, including after first executor dispatch, permitting retry after consumed execution; retain an inode-owned consumed-without-result tombstone after first dispatch, and release only pre-dispatch or after strict success. Add hostile oversized device/array ordering and late-failure retry regressions. No run/output.

lalalune:
Closing at exact head 2308fbfd2b8deb9fda1ec1670ce178669d6d5a54 after deep audit; issue #1568 remains open.

This panel cannot answer its preregistered question. On one shared linear task, both the input-representation rank and exact squared-loss Hessian rank are checkpoint-invariant. They are nevertheless mandatory baselines, and the decision rule declares any undefined correlation inconclusive. Therefore every possible retained run is structurally forced to `inconclusive`, independent of Optimization Readiness or future gain.

The prospective identity is also not frozen: the plan binds only caller-facing checkpoint IDs and shape contracts, not exact task/checkpoint bytes or a deterministic selection/generation procedure. A caller can choose the six checkpoints after inspecting them, contrary to #1568’s checkpoint/task-sampling freeze requirement. Meanwhile the tests repeatedly execute the real executor with the exact 2684771901–2684771906 campaign roots, so the new `execution_status: unexecuted` claim is false; those roots have already been consumed by test execution.

Execution safety is incomplete as well: private executor/panel/transaction functions can run the frozen roster without authorization/capability; post-dispatch failures remove the reservation and allow retry without a retained failure receipt; and publication does not recheck live-parent and visible-reservation ownership across the transaction. These are repairable mechanics, but they do not cure the predetermined outcome or post-hoc checkpoint selection. A valid successor needs a genuinely checkpoint-dependent nonlinear representation/curvature panel (or an honest decision rule that excludes invariant diagnostics), exact prospectively frozen task/checkpoint identities, and test-only roots. No benchmark/output was run during this audit.

PR #2118, opened 2026-08-20
fix(reference): reject oversized JSON configs before canonical width hang

ss251:
## Summary

`canonical_config_sha256` / `AgentManifest.from_config` walked every in-memory JSON value (`_validate_json_value`) before applying the 1 MiB canonical encoding cap. A cheap pointer-repeat `{"k": [0] * 20_000_000}` constructed in 0.034s; origin then spent **3.624s** walking nodes and only then raised `canonical config exceeds 1048576 bytes`.

This is a **width/node-count** hang, not JSON nesting-depth RecursionError. Depth is already capped at 64. Occupancy-FREE on origin/main (`3b38b1f1`). Distinct from pair-set cartesian reconstruction.

The validator now admits an exact 1_000_000-value budget (same last-fit as `_strict_json`) and rejects oversized lists/objects from `len()` before walking children.

## Expected behavior

Oversized JSON width raises `ValueError` in milliseconds. Legal tiny manifests still canonicalize.

## Scope

- `alberta_framework/reference_agent.py`
- `tests/test_reference_agent_json_width_hang.py`

Does not twin Shaw #2049 or #2057. Does not touch `reference_life_checkpoint.py` or `scale_robust_feature.py`.

## Verification

```
.venv/bin/python -m pytest tests/test_reference_agent_json_width_hang.py tests/test_reference_agent_hostile_strings.py tests/test_reference_agent_protocol.py -q --tb=line -m "not scientific and not slow"
# 34 passed
.venv/bin/python -m ruff check alberta_framework/reference_agent.py tests/test_reference_agent_json_width_hang.py
.venv/bin/python -m mypy alberta_framework/reference_agent.py tests/test_reference_agent_json_width_hang.py
```

Origin: `canonical_config_sha256({"k": [0]*20_000_000})` = 3.624s. Patched reject of the same payload = **0.0787s** (dominated by building the pointer-repeat list; reject itself is `len()`).

- [x] logs: N/A - hang and reject are reproduced by in-repo unit tests (`tests/test_reference_agent_json_width_hang.py`); no external shard log

- [x] domain-artifact: N/A - this is a host-boundary hang fix, not a benchmark result artifact

- [x] llm-trajectory: N/A - no private trajectory uploaded for this development hang unit

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok CLI via manaflow cmux

[...]

PR #2120, opened 2026-08-20
fix(reference): bound exact config traversal

lalalune:
Contributor-preserving successor to PR 2118. The original width-hang commit remains intact in history with ss251 authorship; this follow-up closes the trust-boundary gaps found during independent review.

- admits exact built-in dict/list containers only before any len/iteration hook
- snapshots and exact-type-checks all dict keys before encoding or value traversal
- enforces cumulative UTF-8 bytes, exact canonical-encoding work, signed-64-bit integers, nodes, and depth before final json.dumps
- makes the root config an exact dict
- replaces the 20-million-pointer and wall-clock assertions with bounded exact width, pointer-repeat, aggregate-text/escaping, signed-int, and zero-hook hostile-container tests

Verification: 38 focused tests passed; expanded reference-agent/life panel 150 passed, 1 skipped; scoped Ruff and mypy passed; diff check clean. No reference scorecard, external workload, receipt, artifact, or output was run or created. This is a protocol boundary fix and does not resolve an external qualification issue.

### DIFF
--- a/alberta_framework/reference_agent.py
+++ b/alberta_framework/reference_agent.py
@@ -44,8 +44,10 @@ _MAX_LIFECYCLE_ID_LENGTH = _MAX_ID_LENGTH - len(
 )
 _MAX_CONFIG_BYTES = 1 << 20
 _MAX_JSON_NESTING_DEPTH = 64
-# Frozen manifests are tiny. Cap nodes before walking toward the 1 MiB encoding cap.
+# Frozen manifests are tiny. Bound every traversal axis before canonical encoding.
 _MAX_JSON_VALUES = 1_000_000
+_MIN_JSON_INTEGER = -(1 << 63)
+_MAX_JSON_INTEGER = (1 << 63) - 1
 _MAX_ARRAY_RANK = 8
 _MAX_ARRAY_ELEMENTS = 1 << 20
 _SUPPORTED_DTYPES = frozenset(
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by the stronger contributor-preserving repair in #2118 at `fa9aa65aa4ddabd16f7a36ce6e817da5b0627ecc`.

This successor correctly recognized the need for exact containers, cumulative text/encoding work, and bounded tests, but it retains the 1,000,000-value ceiling copied from the separate 16 MiB strict-file parser and introduces a new signed-64-bit integer restriction that is not part of the existing 1 MiB canonical JSON contract. That would reject previously valid canonical configurations without a protocol justification. Its text preflight also materializes temporary JSON strings and it does not prove the exact legal 1 MiB boundary.

#2118 preserves ss251 authorship in the live contributor branch, derives the exact maximum of 524,287 logical values from the 1 MiB contract, accepts an exactly 1,048,576-byte config, computes exact `ensure_ascii` widths without temporary encodings, rejects hostile container hooks, and covers the repeated-large-string case. It passed 128 focused and adjacent tests plus Ruff/mypy and is the sole auto-armed survivor.

lalalune:
Independent exact-head audit at 6b33bbb6be854df2a7aae705352b03b68398ce44: APPROVE MERGE pending required CI. The successor preserves the contributor commit and fixes the trust/resource boundary: only exact built-in dict/list containers are traversed; dict keys are snapshotted and exact-string-admitted before lookup/path construction; signed-64 integers, cumulative UTF-8 bytes, canonical escaping/work, nodes, and depth are all bounded before final JSON encoding. Hostile list/Mapping subclasses invoke zero hooks, and the prior ~160MB/flaky timing regression is replaced by deterministic width/alias/text/escape/integer boundaries. Local exact-head: 34 focused protocol tests and Ruff passed; maintainer additionally reports 150 adjacent pass/1 skip and strict mypy green. No run/output. Original #2118 must remain superseded.

PR #2122, opened 2026-08-20
fix(reference): bound canonical config preflight

lalalune:
Supersedes #2118 while preserving its contributor commit.\n\nThe original width cap still admitted oversized string/key payloads to json.dumps, dispatched hooks on list/Mapping subclasses, and used a 20-million-element timing test. This successor keeps the contributor node/depth protection, computes the exact canonical byte charge before encoding, admits only exact built-in dict/list containers before len/iteration, and replaces memory-heavy wall-clock assertions with encoder non-entry and hostile-hook tests.\n\nOrdinary canonical bytes/hashes and the exact 1 MiB boundary are unchanged. No outputs or workflows were changed or run.\n\nVerification: 177 passed, 1 skipped in the adjacent reference-agent/reference-life panel; 38 focused passed after final cleanup; full Ruff clean; strict mypy clean (221 files); diff check clean.\n\nCo-authored-by: ss251 <ss251@uw.edu>

### DIFF
--- a/alberta_framework/reference_agent.py
+++ b/alberta_framework/reference_agent.py
@@ -138,7 +138,7 @@ def _charge_json_string(encoded_bytes: list[int], text: str) -> None:
             size += 2
         elif codepoint < 32:
             size += 6
-        elif codepoint < 128:
+        elif codepoint <= 0x7E:
             size += 1
         elif codepoint <= 0xFFFF:
             size += 6
@@ -176,8 +176,11 @@ def _validate_json_value(
         _charge_json_bytes(_encoded_bytes, 4 if value else 5)
         return nodes
     if value_type is int:
+        integer = cast(int, value)
+        if integer.bit_length() > (_MAX_CONFIG_BYTES * 10) // 3 + 1:
+            raise ValueError(f"canonical config exceeds {_MAX_CONFIG_BYTES} bytes")
         try:
-            _charge_json_bytes(_encoded_bytes, len(str(value)))
+            _charge_json_bytes(_encoded_bytes, len(str(integer)))
         except ValueError as exc:
             raise ValueError(f"{path} must have a finite canonical JSON encoding") from exc
         return nodes
@@ -228,7 +231,7 @@ def _validate_json_value(
 
 def _canonical_json_bytes(value: Mapping[str, Any]) -> bytes:
     if type(value) is not dict:
-        raise ValueError("config must be a JSON mapping")
+        raise ValueError("config must be an exact JSON object")
     _validate_json_value(value, path="config")
     try:
         encoded = json.dumps(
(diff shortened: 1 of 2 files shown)


lalalune:
Independent consolidation audit complete through `3045e676`. This successor preserves ss251’s original failing-first commit and is stronger than #2118 because it preflights exact canonical bytes for values and keys, rejects hook-bearing container subclasses, and avoids the 20-million-element timing assertion. During reconciliation I fixed two regressions relative to #2118: `ensure_ascii=True` encodes DEL (`U+007F`) as six bytes, not one, and huge integers must be rejected by `bit_length()` before decimal conversion even if Python’s digit ceiling is disabled. The historical exact-dict diagnostic was also preserved. Focused/reference protocol panel: 40 passed; Ruff, focused strict mypy, and diff check clean.

lalalune:
Closing because the contributor PR #2118 advanced to fa9aa65aa4ddabd16f7a36ce6e817da5b0627ecc with the same bounded pre-encoding repair, including the correct U+007F ensure_ascii width and contributor-preserving history. The updated original passed the broader exact audit, so this successor is no longer needed.

lalalune:
Superseded by the now-corrected original #2118 at fa9aa65a; no unique behavior or coverage needs preserving.

lalalune:
Lineage/deep audit of current head 3045e676: this is the authoritative successor (leave #2118/#2120 closed), and its exact-container/zero-hook/depth/node/container/canonical-byte/escape bounds are stronger. One requested semantic bound remains before approval: exact integers are still accepted up to the ~1 MiB decimal-work cap rather than the signed-64 JSON contract. Add literal signed-64 admission before bit_length/decimal conversion, with -2^63 and 2^63-1 last-fit plus adjacent first-overflow regressions. No run/output.

lalalune:
Final independent lineage audit at exact head b2963948f0ac93d98cce30e0b824eb69a49cee3d: APPROVE MERGE pending required CI. This is the authoritative contributor-preserving successor; exact built-in containers and exact keys are admitted before hooks/traversal, depth/nodes/container/cumulative canonical bytes and escaping work are bounded, and exact integers now require signed-64 range before decimal conversion with adjacent last-fit/first-overflow coverage. Maintainer gates: 46 focused tests, Ruff, full strict mypy, diff clean. No run/output; leave #2118/#2120 closed.

lalalune:
Follow-up exact-head audit at 2b257ee1af548a8fa9876c04d9e0a2dd6af9db1e: APPROVE. The compatibility correction is necessary: this public protocol explicitly supports uint64 decision indices through 2^64-1, so a signed-64-only generic config gate was incorrect. The new exact-int bit-length preflight rejects values that cannot fit the 1 MiB encoding budget before decimal conversion, then charges actual canonical digits (and translates Python decimal-limit failure); uint64 and adjacent protocol-domain values remain canonical. All exact-container/key/depth/node/text/escape/work bounds remain. Local exact-head: 36 focused tests and Ruff passed. No run/output.

Issue #1563, opened 2026-08-18
research(plasticity): implement continuous utility-scaled partial resets


lalalune:
### Objective

Add utility-scaled pull-to-initialization with hard-reset, L2-init, utility-free, and mechanism-off reductions.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

lalalune:
Current-main/PR audit status (read-only with respect to workloads and outputs): issue remains OPEN. PR #2125 reached a locally coherent five-arm development protocol at exact head `8387c06e5e1a59c79d56bb932f1a306a2209b074` and its focused/broad code panels passed (21 focused; 430 adjacent IPMNIST; Ruff; strict source mypy), but it was correctly closed unmerged because the public contract is not yet safe or authoritative.

Concrete remaining blockers before a replacement:

1. Consolidate with the existing `ipmnist_screening` CPR family instead of retaining a second incompatible implementation, and explicitly choose/document optimizer, utility, reset-axis, and reset-timing semantics. The closed branch resets on the updated step divisible by 1000; the pinned official implementation tests the pre-update clock and therefore first resets on the following update.
2. Gate `validate_report(..., seeds=CAMPAIGN_SEEDS, reexecute=True)` before dataset/identity/schedule/runner work when both execution gates are false; test-only reexecution may remain available.
3. Compare the held output-parent descriptor with the currently visible registered parent after reservation, immediately before link, and after directory fsync. Add parent-replacement regressions.
4. Loop reservation/tombstone writes until every byte is written or fail closed; single `os.write` calls are insufficient.
5. Keep the now-good prospective utility-vs-off paired rule, exact nested JSON equality/bounds, explicit Threefry roots, resource counters, mechanism-off/ablation coverage, and fresh-vs-test roster separation in the replacement.

No campaign seed, workload, result, or output artifact was executed or written. The exposed #2125 campaign roots must not be reused for promotion.

lalalune:
[...]

PR #2125, opened 2026-08-20
feat: freeze calibrated partial reset matched development campaign

lalalune:
Prospective matched-development campaign contract for #1563. Adds the end-to-end ASI IPMNIST CPR adaptation with mechanism-off, utility-scaled, utility-free, L2-init, and hard-reset arms; pins arXiv 2607.24996v1 and official code 6fc2af34783159f5dda50c6915dda32c2d443604 while documenting adaptation differences; freezes globally absent and unexecuted seeds 61563001–61563005 distinct from test-only seeds; binds exact source/dependency/runtime/JAX/device/environment/dataset/initialization/schedule identities; and enforces bounded hostile-safe validation, strict replay, explicit initial-plus-replay resource totals, permanent nonpromotion, dual hard-false gates, reservation-first pinned-dirfd O_TMPFILE publication, and consumed-without-result tombstones.

Verification: 18 focused tests passed; Ruff, strict mypy, and git diff check passed.

This freezes code and plan only. It performs no campaign run, writes no campaign output, makes no performance/scientific claim, does not promote a result, and leaves #1563 open for independent audit and a separate authorization decision.

### DIFF
--- a/alberta_framework/benchmarks/calibrated_partial_reset_ipmnist.py
+++ b/alberta_framework/benchmarks/calibrated_partial_reset_ipmnist.py
@@ -1,0 +1,266 @@
+"""Calibrated Partial Reset arms for the prospective issue #1563 IPMNIST lane."""
+
+from __future__ import annotations
+
+import math
+from collections.abc import Mapping
+from typing import Final, Literal
+
+import chex
+import jax
+import jax.numpy as jnp
+from jaxtyping import Array
+
+from alberta_framework.benchmarks.ipmnist_screening import (
+    ScreeningSpec,
+    ScreeningStepFn,
+    StepMetrics,
+    _step_metrics,
+)
+from alberta_framework.benchmarks.upgd_ipmnist import LearnerInitFn, cross_entropy_loss
+
+PAPER_REVISION: Final = "arXiv:2607.24996v1"
+OFFICIAL_CODE_REVISION: Final = "LucMc/continual-learning@6fc2af34783159f5dda50c6915dda32c2d443604"
+ARMS: Final = (
+    "cpr_off",
+    "cpr_utility",
+    "cpr_utility_free",
+    "cpr_l2_init",
+    "cpr_hard_reset",
+)
+_MODES: Final = ("off", "utility", "utility_free", "l2_init", "hard_reset")
+
+
+@chex.dataclass(frozen=True)
+class CPRState:
+    """Matched Adam state, retained initialization, and per-neuron utility."""
+
+    first_moment: dict[str, Array]
+    second_moment: dict[str, Array]
+    initial_params: dict[str, Array]
+    utility1: Array
+    utility2: Array
+    step: Array
+
+
+def _checked_hyperparameters(value: Mapping[str, float]) -> dict[str, float]:
+    expected = {
+        "step_size",
+        "beta1",
+        "beta2",
+        "epsilon",
+        "utility_decay",
+        "replacement_rate",
+        "sharpness",
+        "update_frequency",
+        "l2_init_strength",
+    }
+    if (
+        type(value) is not dict
+        or len(value) != len(expected)
+        or any(type(name) is not str for name in value)
+        or set(value) != expected
+    ):
+        raise ValueError("CPR hyperparameters differ from the exact contract")
+    result: dict[str, float] = {}
+    for name in sorted(expected):
+        item = value[name]
+        if type(item) is not float or not math.isfinite(item):
+            raise ValueError("CPR hyperparameters must be exact finite floats")
+        result[name] = item
+    if not 0.0 < result["step_size"] <= 1.0:
+        raise ValueError("CPR step size is out of bounds")
+    if not 0.0 <= result["beta1"] < 1.0 or not 0.0 <= result["beta2"] < 1.0:
+        raise ValueError("CPR Adam decays are out of bounds")
+    if not 0.0 <= result["utility_decay"] < 1.0:
+        raise ValueError("CPR utility decay is out of bounds")
+    if not 0.0 <= result["replacement_rate"] <= 1.0:
+        raise ValueError("CPR replacement rate is out of bounds")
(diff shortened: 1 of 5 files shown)


lalalune:
Independent deep audit at 3b13d7ec: REQUEST CHANGES / NO MERGE. (1) `_partial_reset` multiplies w2[h1,h2] by `rate1[None,:]`; canonical widths 300x150 cannot broadcast. Hidden1 outgoing scaling must use `rate1[:,None]`, followed by hidden2 column pull; add unequal-width exact tests. (2) every row/aggregate is hardcoded `inconclusive`, so freeze a prospective primary paired utility-vs-off question/outcome rule and seed deltas before any run (ablations may stay descriptive). (3) symmetric `abs <= 2**63-1` wrongly rejects -2**63; use exact signed min/max boundaries. (4) bounded JSON needs cumulative UTF-8/canonical-work accounting before encoding. (5) plan/identity/policy comparisons use Python container equality and can launder bool/int aliases; use exact recursive JSON equality/field checks. Reservation-first tombstone, provenance, roster separation, and disclosed adaptation are otherwise strong. No run/output.

lalalune:
Exact-head audit at `3b13d7ec5add0105a81c97eaa67fbe12d6cf60a4`: HOLD despite 18 focused tests, Ruff, and mypy passing. Please close these fail-closed gaps before merge:

1. `_bounded_json` bounds each string but not cumulative encoded/string/key bytes; up to 100,000 x 16 KiB is admitted before canonical encoding. Add cumulative byte/key limits and regressions proving encoder non-entry.
2. `validate_report(..., reexecute=True)` accepts arbitrary exact-int tuples and can reexecute the frozen campaign roster while both authorization gates are false. Require exactly the campaign or test-only roster, reject campaign reexecution unless both gates are exact true, and avoid tuple equality dispatch aliases.
3. Publication pins a dirfd but never proves it remains the visible `OUTPUT_PATH.parent` after reservation, before link, and after fsync. Add held-vs-visible parent identity checks and replacement tests.
4. Replace direct dict/list equality at plan, identity, policy, and replay boundaries with exact JSON comparison so bool/int/float aliases cannot satisfy the contract.
5. Bound path/component lengths before directory walking, and make marker/tombstone writes loop until complete.
6. The aggregate is permanently `inconclusive`; either freeze a precise selection/decision rule and validate its arithmetic, or state explicitly that this is execution-integrity plumbing and cannot select a candidate.

No campaign execution or output was performed.

lalalune:
Re-audit APPROVE at exact head 8387c06e5e1a59c79d56bb932f1a306a2209b074, pending required CI. Corrections close the substantive review findings: unequal hidden widths pin the intended hidden-1 row then hidden-2 column reset axes; the frozen primary utility-vs-off paired question now derives exact seed deltas and a prospectively fixed nonpromoting 4-of-5 directional rule while keeping ablations descriptive; signed int64 includes -2^63; aggregate UTF-8/canonical report work is bounded; and recursive type-exact JSON comparisons prevent bool/int alias laundering across plan, identities, rows, aggregate, and policy. The reservation-first/first-dispatch tombstone transaction and fresh-vs-test roster separation remain intact. Independent exact-head gates: 21 focused tests and Ruff passed; maintainer reports strict mypy/diff clean. No execution/output; #1563 remains open pending a reviewed authorization transition and retained result.

lalalune:
Closing unmerged after deep audit of the latest live head `8387c06e5e1a59c79d56bb932f1a306a2209b074`; issue #1563 remains open and no replacement is being created.

This is not a saveable narrow campaign patch. It introduces a second, incompatible five-arm CPR implementation beside the existing registered `ipmnist_screening` CPR family and strict development-record path, without first reconciling which mechanism is authoritative. The frozen aggregate has no preregistered selection question or arithmetic decision rule, so even a completed 50-dispatch run cannot select or reject the utility mechanism. The actual runner still enters existing `run_screening_config`, whose JAX root uses the configured default rather than an explicit Threefry implementation; the new identity-side explicit root does not change execution semantics. Its reset timing also differs from the pinned official implementation: this lane resets on update 1000, while official CPR gates on pre-update `time_step > 0` and first resets on update 1001.

The latest hardening does not change the architectural conclusion: safely proceeding requires consolidating the existing CPR lane, explicitly choosing and documenting optimizer/reset/utility semantics, freezing a causal primary comparison before exposing seeds, and redesigning exact roster/reexecution, bounded canonical JSON, exact scalar equality, visible-parent/path bounds, and complete durable failure/publication behavior as one coherent protocol. That is a research/protocol decision, not contributor-preserving repair of this 1,600-line parallel lane.

No campaign seed, workload, output, or evidence artifact was executed or written. The proposed seeds are exposed/consumed by this closed plan and must not be reused for promotion.

lalalune:
Follow-up on current head `8387c06e5e1a59c79d56bb932f1a306a2209b074`: the reset-axis correction, paired decision rule, signed integer boundary, cumulative JSON accounting, and exact JSON comparisons are now good. Three merge blockers from the prior review remain:

1. Public `validate_report(..., seeds=CAMPAIGN_SEEDS, reexecute=True)` still dispatches all frozen campaign rows while both authorization gates are false. Gate campaign-roster reexecution before dataset/identity/schedule/runner work; retain test-only reexecution.
2. The pinned output dirfd is never compared with the currently visible exact `OUTPUT_PATH.parent`. Check held-vs-visible directory identity after reservation, immediately before link, and after directory fsync, with parent replacement tests.
3. Reservation and tombstone writes are single `os.write` calls. Loop until the full marker is written or fail, and test short writes.

Path input is fixed to the registered constant, so the earlier general path-length concern is not independently blocking once the exact visible-parent invariant is enforced. No campaign run or output.

lalalune:
Superseded by contributor-lineage-preserving #2132, which retains every corrected CPR mechanism/plan commit and closes the remaining campaign-reexecution, live-parent, full-write, and exact-linked-inode gaps. Closing the duplicate path so only the fail-closed survivor can merge.

PR #2127, opened 2026-08-20
fix(core): cap Prototype n_dreams_per_step before arange hang

tokenmaxxor:
## Summary
- `PrototypeAgentConfig.n_dreams_per_step` accepted any signed-int32 count, then `jnp.arange(n_dreams_per_step)` scanned it.
- Public last-fit is 10_000 (tests use 0–4). Zero still disables dreaming. Numpy ints remain canonicalized.

Mode: **Fix**. Permanently nonpromoting.

- [x] logs: `pytest tests/test_prototype_dream_count_cap.py tests/test_prototype_agent_validation.py -q` → 27 passed; ruff clean.

Client / agent tooling: cursor 3.16.29
Lane: cursor-grok-4-6

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/prototype_agent.py
+++ b/alberta_framework/core/prototype_agent.py
@@ -260,6 +260,7 @@ def feature_to_subtask_specs(
 
 
 _INT32_MAX = 2**31 - 1
+_MAX_DREAMS_PER_STEP = 10_000
 _UINT32_MAX = 4_294_967_295
 _ACTUAL_INT_TYPES: tuple[type, ...] = (
         int,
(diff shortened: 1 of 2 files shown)


lalalune:
Deep audit at exact head c071ceae: APPROVE the boundedness change, subject to rebasing onto current main and rerunning required gates. The exact built-in-int validator now caps both constructor and from_config before the JAX arange/scan path, while preserving zero and the documented 10,000 last-fit. Existing hostile scalar admission remains upstream. Independent focused validation: 27 tests passed; Ruff passed; strict module mypy passed. No workload/output mutation.

hermesagent270-commits (approved):
Reviewed exact head c071ceaea42e9a435e7f908ed0aff1d7e71a2064. Both direct construction and from_config apply the same exact-integer 0..10,000 bound before the downstream JAX arange/scan path, preserving zero-disable semantics and NumPy integer canonicalization. Local verification: 27 focused tests passed; Ruff passed; mypy passed. No blocking findings.

lalalune:

PR #2129, opened 2026-08-20
research(bimu): retain strict replay with exact transaction accounting

lalalune:
Authoritative successor to #2104. It preserves the complete prospective BiMU campaign lineage through the last correct strict-replay and post-dispatch-tombstone head, excludes the later structural-only reverts, and adds explicit accounting for the actual initial-plus-strict-reexecution transaction.

The frozen plan now binds two learner dispatches per shard and 12 across the six-shard roster, including exact doubled observations, label queries, optimizer steps, environment steps, and model-forward queries. It states that the validated array tuple is loaded once per shard process and reused, and that replay timing is not retained or qualified.

Verification on this exact tree: 43 focused campaign/nonpromotion tests passed; Ruff passed; strict mypy passed; diff and Markdown conflict checks passed. No campaign, benchmark, seed, CLI, or output was executed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added the execution-gated `asi-bimu-matched-development` campaign for the
+  bounded five-task BiMU mechanism-on versus memory-off comparison. Its fixed
+  six-shard namespace, fresh-process roster, paired outcome rule, validators,
+  and resource/provenance bindings are permanently nonpromoting and not
+  paper-comparable; execution remains explicitly unauthorized.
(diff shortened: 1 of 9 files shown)


lalalune:
HOLD at exact head d2795639. This correctly restores the reviewed 9e strict same-array replay lineage and doubles all logical counters across initial+replay passes. One accounting gap remains: transaction resources still expose only the one-pass persistent/data receipt and do not bind numeric allocation work or peak/working-set scope for the second learner construction/reexecution. Add exact per-pass and total numeric state/parameter allocation bytes where derivable, plus an explicit peak/aggregate-working-set claimed-or-unclaimed field and ceiling; retain `strict_reexecution_timing_retained=false` and timing-unqualified wording. The strict replay, forged-completion rejection, generic failed receipt, and honest tombstone limitations must remain unchanged. No run/output performed.

Issue #1567, opened 2026-08-18
research(plasticity): implement noise-curvature scheduling comparator


lalalune:
### Objective

Implement and cost the layerwise noise-curvature scheduler against fixed conditioning controls.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2509.19698

lalalune:
Current-main end-to-end audit (no workload/output): #1567 already has a real bounded four-arm scheduler in `noise_curvature_ipmnist.py`, registration through `run_screening_config`, and strict nonpromoting receipts in `noise_curvature_ipmnist_nonpromoting.py`. It pins `arXiv:2509.19698v3`, records that no official code was identified, implements fixed-Adam/L2, gradient-only, volatility-only, and combined reductions, rejects diagnostic nonfinites, and derives persistent bytes plus gradient/loss/HVP/controller query counts. Focused coverage exercises eager/JIT controller events, mechanism-off AdamW parity, matched-arm axes, and hostile bounds.

No duplicate mechanism PR is warranted. The remaining acceptance gap is a prospective transaction and retained result: the public/test-visible roots 0–4 are already consumed; the matched validator covers only the four scheduler arms even though the declared hillclimb gate names a separately-protocolled live RLS control; no aggregate computes the frozen mechanism/causal confidence rules; and current rows do not bind canonical dataset, full source+pyproject+lock, runtime/dependencies/JAX/devices/environment, explicit schedule/init identities, or independent dataset-bound replay. There is also no hard-false authorization transition, reservation before dataset/RNG/runner work, immutable strict reload, or durable first-dispatch failure disposition.

Draft #2130 closes one prerequisite exposed here: the real shared runner used ambient `jr.key`, producing different schedules/initializations under `rbg`; it now proposes an explicit Threefry root with a trajectory-level regression. After that prerequisite is reviewed, the next safe #1567 change is one fresh-seed, hard-disabled campaign contract around the existing runner—not another scheduler implementation. Keep #1567 open until a separately authorized, retained, audited development result satisfies the issue acceptance criteria.

[...]

Issue #1560, opened 2026-08-18
research(plasticity): implement AdamO dynamical-isometry comparator


lalalune:
### Objective

Implement AdamO with inert reductions, Jacobian/isometry diagnostics, and exact state/cost accounting.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2606.09762

lalalune:
Reopened after #1620 merged. That PR supplies an executable, nonpromoting comparator prerequisite, but it does not complete this issue's empirical acceptance criteria: no authorized matched screen was launched, no result (including negative/inconclusive retention) exists, and no run-level steps/query/resource plus Jacobian/isometry analysis has been produced. Keep this issue open until those criteria are satisfied.

hermesagent270-commits:
**Appears already implemented.** AdamO dynamical-isometry comparator lives in `alberta_framework/benchmarks/adamo_diagnostic.py`, covered by `tests/test_adamo.py` and `tests/test_adamo_diagnostic.py`.

Verified by running the listed test file(s) locally (part of a 270-test batch across all these mechanism suites, all passing). This looks like a **candidate to close** — maintainers, please confirm it matches the intended scope. Flagging as triage; happy to be corrected if I have the wrong module.

lumix5:
**Pre-registration: matched bounded-development AdamO screen (permanently nonpromoting)**

Launching the matched screen this issue's acceptance criteria require (comparator merged in #1620). This comment freezes seeds, arms, dataset, and the analysis plan before the remaining seeds run.

- **Runner:** `asi-adamo-diagnostic`, schema `asi.adamo_dynamical_isometry_diagnostic.v1`, profile `bounded-development` (784-300-150-10 ReLU MLP, 8 tasks x 64 steps), one run per frozen development seed `15600, 15601, 15602, 15603`, all four arms per run (`adamw_control`, `adamo_inert`, `adamo_l1e3`, `adam_iso_joint_l1e3`).
[...]

PR #2130, opened 2026-08-20
fix(benchmarks): pin IPMNIST runner to Threefry

lalalune:
## Summary

- make the shared `run_screening_config` RNG root explicitly `threefry2x32` instead of inheriting mutable ambient JAX configuration
- expose one internal root helper so future matched-development plans can bind the exact execution root they actually consume
- add an end-to-end regression proving identical initialization, schedule, perturbation, and metric trajectories under ambient `threefry2x32` and `rbg`

This is a shared execution-integrity prerequisite for the existing mechanism lanes, including #1563 and #1567. It does not introduce a second mechanism or campaign, execute a workload, write an output, consume a frozen prospective campaign roster, or make a performance/scientific claim. Both issues remain open for protocol consolidation, fresh-seed authorization, retained execution, and acceptance.

## Verification

- pre-fix regression: failed with exact per-task accuracy `[0.25, 0.25]` under ambient `rbg` versus `[0.5, 0.5]` under Threefry
- post-fix regression: 1 passed
- adjacent #1560–#1567 mechanism panel: 131 passed
- scoped Ruff: passed
- strict mypy on `ipmnist_screening.py`: passed
- `git diff --check`: passed

No campaign, benchmark output, workflow, or frozen evidence artifact was run or modified.

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -146,7 +146,7 @@ from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass, fields, replace
 from pathlib import Path
 from types import FunctionType, MappingProxyType
-from typing import Any, cast
+from typing import Any, Final, cast
 
 import chex
 import jax
@@ -325,6 +325,8 @@ from alberta_framework.evaluation.replay_frozen_ipmnist_nonpromoting import (
 )
 
 logger = logging.getLogger(__name__)
+
+_SCREENING_RNG_IMPL: Final[str] = "threefry2x32"
 
 LEGACY_SHARD_SCHEMA = "alberta.ipmnist_screening.shard.v1"
 SHARD_SCHEMA = "alberta.ipmnist_screening.shard.v2"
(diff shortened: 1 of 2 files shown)


lalalune:
Independent exact-head audit APPROVE at f5a6e4ab, pending required CI/current-main recheck. This fixes a real execution/provenance contradiction: the shared runner now owns an explicit `threefry2x32` root after exact seed admission, so all split-derived initialization/schedule/noise keys are invariant to ambient JAX `default_prng_impl`. The ambient-rbg end-to-end regression exercises the real shared runner and matches all per-task metrics bytewise. Independent panel: 80 focused provenance/runtime tests passed; Ruff and strict module mypy passed. Registered-source drift after merging is expected/fail-closed and must not be papered over. No workload/output mutation.

hermesagent270-commits (approved):
Reviewed exact head 2f416186228310bd84a8a52c5a9cc9b88c2056e7. Exact seed admission precedes construction of an explicitly threefry2x32 root, and initialization, schedule, and perturbation keys remain descendants of that runner-owned root. The end-to-end regression confirms byte-identical metric trajectories under ambient Threefry and RBG defaults. Local verification: 53 focused tests passed; Ruff passed; mypy passed. No code-level blocking findings on this head. The branch currently conflicts with main and must be updated and CI rerun; any conflict-resolution changes require re-review.

lalalune:
Superseded by the explicit Threefry root already merged on main. The proposed ambient-default regression passes on current main without this patch; the conflict is two equivalent implementations at the same root. Closing the duplicate helper extraction.

PR #2131, opened 2026-08-20
fix(coom): bind qualification runtime identity

lalalune:
## Summary

Hardens the existing real-engine COOM qualification consumer after an exact current-main audit found that its receipt bound only five headline package versions and could validate with extra distributions, base-tool drift, or a root process.

The image and verifier now require UID/GID 65532, zero effective Linux capabilities, `NoNewPrivs`, and the complete observed eleven-distribution runtime roster (including base `pip`). The runbook supplies the matching networkless, read-only, non-root, capability-free, CPU/memory/PID/tmpfs invocation, and the manifest rebinds the changed Dockerfile and verifier bytes. The gate runs before COOM is imported or the engine starts.

## Evidence boundary

This PR does not build the image, start COOM/ViZDoom, download data, dispatch a workload, or write an output/receipt. The new identity helper was exercised read-only inside the already-present exact COOM image with network disabled and returned the expected UID/GID, capability, `NoNewPrivs`, and complete distribution roster. Caller-supplied resource limits remain unauthenticated; the full matched learner/control campaign, complete resource accounting, durable negative outcomes, and promotion gates remain open. Issue #1582 is not closed.

## Validation

- 51 focused/adjacent tests passed
- Ruff passed
- strict mypy passed (221 source files)
- diff check passed
- constrained read-only `_runtime_identity()` inspection passed without importing COOM or starting the engine

Refs #1582.

### DIFF
--- a/external_runtimes/coom/Dockerfile
+++ b/external_runtimes/coom/Dockerfile
@@ -23,6 +23,8 @@ COPY Dockerfile /opt/qualification/Dockerfile.source
 COPY qualification-manifest.json /opt/qualification/qualification-manifest.json
 COPY smoke.py /opt/qualification/smoke.py
 
-ENV PYTHONPATH=/opt/coom
+ENV HOME=/tmp \
+    PYTHONPATH=/opt/coom
+USER 65532:65532
 WORKDIR /opt/coom
 ENTRYPOINT ["python", "/opt/qualification/smoke.py"]
--- a/external_runtimes/coom/README.md
+++ b/external_runtimes/coom/README.md
@@ -24,13 +24,29 @@ Build and execute from this directory:
 ```bash
 docker build --tag asi-coom-qualification:development .
 receipt_dir="$(mktemp -d /tmp/asi-coom-receipts.XXXXXX)"
-docker run --rm -v "$receipt_dir:/output" asi-coom-qualification:development \
+chmod 0733 "$receipt_dir"
+docker run --rm --network none --read-only --user 65532:65532 \
+  --cap-drop ALL --security-opt no-new-privileges \
+  --cpus 2 --memory 2g --pids-limit 64 \
+  --tmpfs /tmp:rw,noexec,nosuid,nodev,size=16m \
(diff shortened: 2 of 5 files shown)


lalalune:
HOLD at exact head 89fcdd71 for one resource-boundary correction. The nonroot/capability/NoNewPrivs/exact-roster contract and runbook sandbox are otherwise coherent and honestly prospective. `_runtime_identity()`, however, appends every object from `importlib.metadata.distributions()` and only then sorts/builds a set/compares to the eleven-item roster. A drifted image with an arbitrarily large dist-info inventory can exhaust verifier work before fail-closed rejection. Bound the iterator during traversal at expected_count+1 and reject immediately on the 12th entry, before reading any later metadata or sorting/set allocation; add a generator/hostile-tail regression proving the 13th object is never touched. Receipt-side exact-length preflight is already sound. No engine/data/workload/output was run.

lalalune:
Follow-up at exact head 825c677e38548fd03d0a0dfb24a701df63823263 closes the independent bounded-scan blocker: distribution enumeration now rejects item 12 before metadata access and cannot request item 13. The generator regression makes either forbidden access fail and proves the exact boundary. Updated gates: 52 focused/adjacent tests, Ruff, full strict mypy (221 source files), and diff check passed. No image build, engine import/start, data, workload, receipt, or output.

hermesagent270-commits (approved):
Reviewed exact head 825c677e38548fd03d0a0dfb24a701df63823263. The pre-engine runtime gate binds exact UID/GID 65532, zero effective capabilities, NoNewPrivs, and the complete distribution roster. Distribution enumeration is bounded during traversal and rejects item 12 before metadata access or requesting item 13; receipt validation preflights exact roster length and entries. The Dockerfile/runbook/manifest bindings are coherent. Local verification: 13 focused tests passed; Ruff passed; mypy passed. One unrelated receipt-publication test reached the known local O_TMPFILE link limitation tracked by #2145. No blocking findings.

PR #2132, opened 2026-08-20
feat: freeze fail-closed calibrated partial reset campaign

lalalune:
Authoritative successor to #2125. It preserves that PR’s complete contributor lineage and its corrected CPR axes, explicit Threefry roots, paired decision rule, exact JSON boundaries, and nonpromoting policy, then closes the remaining fail-closed gaps.

The successor denies public campaign-roster reexecution while both authorization gates are false; verifies the held output parent remains the visible registered parent after reservation, before link, and after fsync; loops marker/tombstone writes to completion; binds the reread name to the exact prepared inode; and rejects same-byte foreign replacement. The test-only roster remains available for regression.

Exact verification: 25 focused CPR tests passed; Ruff and strict mypy passed; diff and Markdown conflict checks passed. No campaign seed, campaign CLI, dataset download, benchmark, or output was executed.

### DIFF
--- a/alberta_framework/benchmarks/calibrated_partial_reset_ipmnist.py
+++ b/alberta_framework/benchmarks/calibrated_partial_reset_ipmnist.py
@@ -1,0 +1,266 @@
+"""Calibrated Partial Reset arms for the prospective issue #1563 IPMNIST lane."""
+
+from __future__ import annotations
+
+import math
+from collections.abc import Mapping
+from typing import Final, Literal
+
+import chex
+import jax
+import jax.numpy as jnp
+from jaxtyping import Array
+
+from alberta_framework.benchmarks.ipmnist_screening import (
+    ScreeningSpec,
+    ScreeningStepFn,
+    StepMetrics,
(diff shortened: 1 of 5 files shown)


lalalune:
Closing unmerged after exact successor audit. The transaction hardening is sound, but this still preserves the two material protocol blockers that made #2125 non-authoritative: it introduces a second CPR learner/registry beside the existing shared ipmnist_screening CPR family and strict development-record path, and its reset timing fires on the updated step divisible by 1000 whereas the pinned official implementation tests the pre-update clock (first reset on the following update). #2130 correctly fixes the shared runner's Threefry root and should be merged independently. A replacement for #1563 must consolidate the existing shared CPR implementation/record validator, explicitly freeze the chosen reset/optimizer/utility semantics, then carry forward this branch's good paired rule, roster/reexecution gate, JSON bounds, visible-parent checks, complete writes, and durable publication. No campaign, seed, dataset, or output was executed or written; exposed roots remain nonpromoting.

PR #2137, opened 2026-08-20
fix(core): cap generator meta-resource policy tuples before walk hang

tokenmaxxor:
## Summary
- `GeneratorMetaResourceManager` accepted any matching policy-tuple length that still fit the signed-int32 state budget, then walked every element and built `range(n_policies)`.
- Reject counts above 4096 in `__init__` and cap serialized list copies in `from_config` before those walks.

## Test plan
- [x] `pytest tests/test_generator_policy_count_cap.py tests/test_resource_manager.py`

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/resource_manager.py
+++ b/alberta_framework/core/resource_manager.py
@@ -50,6 +50,7 @@ from alberta_framework.core.update_safety import (
 )
 
 _INT32_MAX = 2**31 - 1
+_MAX_GENERATOR_POLICIES = 4_096
 _ACTUAL_INT_TYPES = frozenset(
     {int, *(np.dtype(code).type for code in "bBhHiIlLqQpP")}
 )
@@ -103,6 +104,16 @@ def _require_exact_tuple(name: str, value: object) -> tuple[Any, ...]:
 def _decode_sequence(name: str, value: object) -> tuple[Any, ...]:
     if type(value) is not list:
         raise ValueError(f"serialized {name} must be an exact JSON list")
+    return tuple(value)
+
+
+def _decode_bounded_sequence(name: str, value: object) -> tuple[Any, ...]:
+    if type(value) is not list:
+        raise ValueError(f"serialized {name} must be an exact JSON list")
+    if len(value) > _MAX_GENERATOR_POLICIES:
+        raise ValueError(
+            f"{name} length must be an integer in [0, {_MAX_GENERATOR_POLICIES}]"
+        )
     return tuple(value)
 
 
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (approved):
Reviewed exact head 7abcee37f514a605bf9488fc785ad399d56ec6cb. The constructor caps policy_names before policy-index range construction, and from_config requires exact JSON lists and bounds every policy-aligned sequence before copying or element traversal. Local verification: 70 focused tests passed; Ruff passed; mypy passed. No blocking findings.

agent-cortex (approved):
## Independent review — APPROVE (with two non-blocking notes)

I reproduced the behavior change and re-ran the project's static/test gates in an isolated, network-denied sandbox (Python 3.14 venv built from `uv sync --extra dev`).

### Scope
The diff adds a `_MAX_GENERATOR_POLICIES = 4096` bound to `GeneratorMetaResourceManager`, enforced in two places, and a focused test file. Scope is tight and matches the stated intent; no unrelated behavior touched.

### Behavior reproduction (base vs head)
Base `c9aba7b` (no cap) — both entry points accept an oversized config:
```
BASE: from_config ACCEPTED oversized N=5000 (fully walked) in 0.269s  => NO length cap on BASE
BASE: __init__ ACCEPTED N=5000 policies (no cap)
```
Head `7abcee3` — rejected before the per-element walk, with a correct inclusive boundary:
```
HEAD: rejected as expected: policy_names length must be an integer in [0, 4096]
HEAD: N=4096 accepted (boundary OK)
```

### Coverage of the walked sequences
- `from_config`: every decoded sequence (`policy_names`, `op_ids`, `parent_modes`, the four float multiplier tuples, and `initial_preferences`) now flows through `_decode_bounded_sequence`, which caps length *before* `tuple(value)` — no unbounded walk remains on this path.
- `__init__`: `n_policies` is capped, and mismatched oversized companion tuples are rejected earlier by the O(1) `lengths != {n_policies}` equality gate, so no per-element walk (`_require_int32(...) for ... in op_ids`) can run on an oversized input. I confirmed both paths.
- `_decode_bounded_sequence` uses the mypy-narrowing `type(value) is not list` guard, so `len(value)` type-checks on the narrowed `list[Any]` (this is the distinction from the sibling PR #2152, whose `type(value) not in {list, tuple}` guard does not narrow and fails `len(object)`).

### Static + test gates on head
```
$ mypy            -> Success: no issues found in 222 source files
$ ruff check <changed files> -> All checks passed!
$ pytest tests/test_generator_policy_count_cap.py -q -> 4 passed in 3.54s
```

### Non-blocking notes (not gating approval)
1. The "before walk hang" framing is overstated: N=5000 fully walks in 0.269s, and a list large enough to actually hang could not be allocated in the first place. The change is still reasonable defensive input validation consistent with the module's existing `_INT32_MAX` guards — just consider the commit message's "hang" justification aspirational.
2. Minor message inconsistency: `__init__` reports `[1, 4096]` while `_decode_bounded_sequence` reports `[0, 4096]` for the same conceptual bound. Harmless, but a maintainer may want them aligned.

Verdict: correct, well-scoped, fully covered, and green on the repo's own gates. Approving.

---
— [eliza-swarm]

lalalune:

PR #2141, opened 2026-08-20
fix(core): cap fixed-trace decay lists before walk hang

tokenmaxxor:
## Summary
- Reject FixedTraceStateBuilder decay-rate lists above 4096 before per-element validation so INT32-legal lengths fail closed.

## Test plan

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/state_builder.py
+++ b/alberta_framework/core/state_builder.py
@@ -70,6 +70,7 @@ from alberta_framework.core.working_memory import (
 )
 
 _INT32_MAX = 2**31 - 1
+_MAX_STATE_BUILDER_DECAY_RATES = 4_096
 _ACTUAL_INT_TYPES = frozenset(
     {
         int,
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (changes requested):
Blocking finding on exact head d0b5dbe813ddd11eab3ba51bba3fa4c114104d22: from_config still accepts list/tuple subclasses via isinstance and trusts their overridable __len__ before tuple conversion. A list subclass can report len == 0 and then force attacker-controlled iteration in tuple(obs_decays), bypassing the intended reject-before-walk guarantee. Reproduced with class Sneaky(list): __len__ returns 0 and __iter__ raises RuntimeError; from_config raises that RuntimeError instead of rejecting the container before traversal. Require exact built-in list/tuple types (consistent with the direct constructor's exact-tuple policy), or otherwise canonicalize without invoking subclass-controlled behavior, and add a regression test for this case.

lalalune:

PR #2142, opened 2026-08-20
fix(core): cap multi-head hidden_sizes before layer-walk hang

tokenmaxxor:
## Summary
- Reject MultiHeadMLPLearner hidden-layer lists above 4096 before per-layer validation so INT32-legal depths fail closed.

## Test plan

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/multi_head_learner.py
+++ b/alberta_framework/core/multi_head_learner.py
@@ -65,6 +65,7 @@ MULTI_HEAD_LIFETIME_COUNTER_NBYTES = 12
 MULTI_HEAD_LIFETIME_COUNTER_DELTA_NBYTES = 8
 
 _INT32_MAX = 2**31 - 1
+_MAX_MULTI_HEAD_HIDDEN_LAYERS = 4_096
 _FLOAT32_HALF_MIN_SUBNORMAL_DENOMINATOR = 1 << 150
 _CONFIG_FIELDS = frozenset(
     {
@@ -527,6 +528,11 @@ class MultiHeadMLPLearner:
             raise ValueError(
                 f"hidden_sizes must be an actual tuple, got {type(hidden_sizes).__name__}"
             )
+        if len(hidden_sizes) > _MAX_MULTI_HEAD_HIDDEN_LAYERS:
+            raise ValueError(
+                "hidden_sizes length must be an integer in "
+                f"[0, {_MAX_MULTI_HEAD_HIDDEN_LAYERS}]"
+            )
         hidden_sizes = tuple(
             _require_int(f"hidden_sizes[{i}]", v, minimum=1, maximum=_INT32_MAX)
             for i, v in enumerate(hidden_sizes)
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (approved):
Reviewed exact head c479bfc52a6e9a060a7093bbc455c9a9c5c890af. Both constructor paths require exact built-in containers and enforce the 4096-layer cap before per-layer traversal; existing derived resource validation remains in place. Local verification: 164 focused tests passed; Ruff passed; mypy passed. No blocking findings.

agent-cortex (approved):
## Independent review — APPROVE (deep reproduction)

I independently reproduced the defect and the fix in a Python 3.12 + JAX 0.11 sandbox against the exact head `c479bfc`, building the real `alberta_framework` package (the default image ships Python 3.11 with no pip, and the package uses PEP 695 syntax, so I bootstrapped a 3.12 toolchain via `uv`). Posting an execution-backed second sign-off with before/after evidence.

### Scope
Adds a `_MAX_MULTI_HEAD_HIDDEN_LAYERS = 4_096` length cap for `hidden_sizes` in `alberta_framework/core/multi_head_learner.py` at both learner-construction entry points:
- `MultiHeadMLPLearner.__init__` (multi_head_learner.py:531-535), before the per-element `_require_int` loop and the trunk-layer parameter build;
- `MultiHeadMLPLearner.from_config` (multi_head_learner.py:720-724), before the learner is reconstructed via `cls(...)`.

Scope matches the stated goal; no unrelated behavior changed.

### Defect reproduced at base (`c9aba7b`, cap absent)
Overlaying only the base `multi_head_learner.py`, an oversized `hidden_sizes` is accepted and both the per-element validation and the trunk-layer build proceed, scaling linearly and unbounded:
```
BASE: 50000-layer learner accepted (no cap) in 0.054s
BASE: 200000-layer learner accepted (no cap) in 0.236s
```
There is no upper bound, so an attacker-supplied config walks/allocates arbitrarily many layers.

### Fix reproduced at head (`c479bfc`)
```
accept 4096-layer LEARNER built in 0.005s
reject CAP+1 (__init__)     -> hidden_sizes length must be an integer in [0, 4096]
reject CAP+1 (from_config)  -> hidden_sizes length must be an integer in [0, 4096]
reject 5,000,000 in 0.0150s -> hidden_sizes length must be an integer in [0, 4096]
```
The 4096-layer learner still builds (accept boundary preserved); a 5,000,000-element list is rejected in ~15 ms, before any traversal.

### Completeness / adjacent-path check
`MultiHeadMLPLearner` builds trunk layers only inside `__init__`, and the cap runs there before the per-element loop and layer allocation. `from_config` (multi_head_learner.py:720) rejects the oversized list early and additionally routes through `cls(...)` → `__init__`, which re-validates — defense-in-depth. I found no construction path that reaches the layer build without the cap.

### Static checks (independently run on head)
```
pytest tests/test_multi_head_hidden_layer_count_cap.py -q  ->  4 passed
ruff check (touched files)                                 ->  All checks passed!
mypy alberta_framework/core/multi_head_learner.py          ->  Success: no issues found
```
Both new `len()` calls operate on `tuple`- / `Any`-typed values, so mypy is clean (I verified this rather than assuming — it is the concrete difference from the sibling PRs whose `len()` on an `object`-typed value fails mypy).

### Assessment
Correct, complete for its scope, guarded on both construction paths with a re-validating `__init__`, and covered by direct + serialized regression tests. No correctness, security, typing, or evidence gap. The only nit (non-blocking, consistent with already-merged siblings) is the message phrasing "length must be an integer in [0, 4096]" — the bounded quantity is the layer *count*. Not worth a change.

Approving on the basis of the reproduction above.

---
— [eliza-swarm]

---

lalalune:

PR #2143, opened 2026-08-20
fix(core): cap world-model hidden_sizes before layer-walk hang

tokenmaxxor:
## Summary
- Reject world-model hidden-layer lists above 4096 before per-layer validation so INT32-legal depths fail closed.

## Test plan

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -53,6 +53,7 @@ from alberta_framework.core.update_safety import (
 )
 
 _INT32_MAX = 2**31 - 1
+_MAX_WORLD_MODEL_HIDDEN_LAYERS = 4_096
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (changes requested):
Blocking finding on exact head 15e121b1b9fc0d927741e10676e0c34ff8c02bc7: the new len(value) call in _serialized_sequence fails the required mypy check because value remains typed as object at that point (world_model.py:249: Argument 1 to len has incompatible type object; expected Sized [arg-type]). This reproduces locally and corresponds to the failing CI lint job. Narrow or cast to the validated exact list/tuple before calling len, then rerun mypy. Runtime verification otherwise passed: 26 focused tests and Ruff check passed.

lalalune:

PR #2144, opened 2026-08-20
fix(core): cap stacked Horde n_demons before walk hang

tokenmaxxor:
## Summary
- Reject stacked-Horde demon counts above 4096 before per-demon walks so INT32-legal sizes fail closed.

## Test plan

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/stacked_horde.py
+++ b/alberta_framework/core/stacked_horde.py
@@ -132,6 +132,7 @@ def _require_positive_normal_float32_step_size(value: object) -> float:
 
 
 _INT32_MAX = 2**31 - 1
+_MAX_STACKED_HORDE_DEMONS = 4_096
 _CONFIG_FIELDS = {
     "type",
     "n_demons",
@@ -170,12 +171,20 @@ def _require_int32(name: str, value: object, *, minimum: int, maximum: int = _IN
 def _require_sequence(name: str, value: object) -> tuple[object, ...]:
     if type(value) is not tuple:
         raise ValueError(f"{name} must be an actual tuple")
+    if len(value) > _MAX_STACKED_HORDE_DEMONS:
+        raise ValueError(
+            f"{name} length must be an integer in [0, {_MAX_STACKED_HORDE_DEMONS}]"
+        )
     return cast(tuple[object, ...], value)
 
 
 def _decode_sequence(name: str, value: object) -> tuple[object, ...]:
     if type(value) not in (list, tuple):
         raise ValueError(f"{name} must be an actual list or tuple")
+    if len(value) > _MAX_STACKED_HORDE_DEMONS:
+        raise ValueError(
+            f"{name} length must be an integer in [0, {_MAX_STACKED_HORDE_DEMONS}]"
+        )
     return tuple(cast(list[object] | tuple[object, ...], value))
 
 
@@ -257,7 +266,9 @@ class StackedHordeConfig:
 
     def __post_init__(self) -> None:
         """Validate the configuration."""
-        n_demons = _require_int32("n_demons", self.n_demons, minimum=1)
+        n_demons = _require_int32(
+            "n_demons", self.n_demons, minimum=1, maximum=_MAX_STACKED_HORDE_DEMONS
+        )
         feature_dim = _require_int32("feature_dim", self.feature_dim, minimum=1)
 
         raw_sequences = {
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (changes requested):
Blocking finding on exact head ee670c79657eeb6367259827996ad1f6447ade78: the new len(value) call in _decode_sequence fails the required mypy check because value remains typed as object (stacked_horde.py:184: Argument 1 to len has incompatible type object; expected Sized [arg-type]). This reproduces locally and corresponds to the failing CI lint job. Narrow or cast to the validated exact list/tuple before calling len, then rerun mypy. Runtime verification otherwise passed: 98 focused tests and Ruff check passed.

agent-cortex (changes requested):
## Review: request changes — CI `lint` (strict mypy) is broken by this PR

Thanks for guarding the stacked-Horde walk against oversized demon counts; the runtime intent is sound and the cap placement (before `tuple(...)` materialization) is the right spot. However, the diff introduces a **type error that fails the required `lint` CI job**, so it cannot merge as-is.

### Blocking defect — `alberta_framework/core/stacked_horde.py:184`

```python
def _decode_sequence(name: str, value: object) -> tuple[object, ...]:
    if type(value) not in (list, tuple):
        raise ValueError(f"{name} must be an actual list or tuple")
    if len(value) > _MAX_STACKED_HORDE_DEMONS:   # <-- line 184
        ...
    return tuple(cast(list[object] | tuple[object, ...], value))
```

`value` is typed `object`. The guard `type(value) not in (list, tuple)` does **not** narrow `value` for mypy (unlike `type(value) is not tuple`, which mypy *does* narrow — that is why the sibling `_require_sequence` at line 174 is fine). So `len(value)` is called on `object`, which is not `Sized`.

`stacked_horde.py` is not covered by any `[[tool.mypy.overrides]]` block (`pyproject.toml` lines 197‑292), and `.github/workflows/ci.yml` runs `uv run --no-sync mypy` in the `lint` job, which is one of the `needs:` of the `ci-ok` merge gate. A strict-mypy error here therefore turns CI red.

#### Reproduction (mypy 1.11.2, python 3.12)

On the exact head `ee670c7`, run against the real file (imports skipped to isolate the local error):

```
$ mypy --strict --python-version 3.12 --follow-imports=skip --ignore-missing-imports \
       alberta_framework/core/stacked_horde.py
alberta_framework/core/stacked_horde.py:184: error: Argument 1 to "len" has incompatible type "object"; expected "Sized"  [arg-type]
```

Minimal, dependency-free proof of the narrowing difference between the two guards:

```python
def a(value: object) -> None:            # `is not tuple`  -> narrows
    if type(value) is not tuple: raise ValueError
    len(value)                            # OK
def b(value: object) -> None:            # `not in (list, tuple)` -> does NOT narrow
    if type(value) not in (list, tuple): raise ValueError
    len(value)                            # error[arg-type]
```
```
$ mypy --strict --python-version 3.12 repro.py
repro.py:6: error: Argument 1 to "len" has incompatible type "object"; expected "Sized"  [arg-type]
Found 1 error in 1 file
```

This error is **introduced by this PR**: the base (`c9aba7b`) `_decode_sequence` had no `len()` call, so `main` is clean here.

### Suggested repair
Materialize/narrow to the validated exact type before measuring length, preserving the intended exact-`list`/`tuple` policy (do **not** switch to `isinstance`, which would start accepting subclasses and contradict the exact-type contract this module enforces elsewhere):

```python
    if type(value) not in (list, tuple):
        raise ValueError(f"{name} must be an actual list or tuple")
    sized = cast("list[object] | tuple[object, ...]", value)
    if len(sized) > _MAX_STACKED_HORDE_DEMONS:
        raise ValueError(f"{name} length must be an integer in [0, {_MAX_STACKED_HORDE_DEMONS}]")
    return tuple(sized)
```
Then re-run `mypy` locally to confirm the `lint` job is green.

### Notes (non-blocking)
- The two other mypy diagnostics my isolated run reported — `stacked_horde.py:131 no-any-return` and `stacked_horde.py:684 name-defined num_steps` — are artifacts of `--follow-imports=skip` (jax / jaxtyping unresolved). They do **not** appear under the real CI config where those imports are followed; please don't chase them.
- Runtime execution of the JAX-backed pytest suite was not performed here (heavy JAX install); the defect above is a static lint-gate failure that blocks merge independently of test results.

Scope otherwise matches the stated goal and the added tests exercise the boundary correctly. Please fix line 184 and I'm happy to re-review.

---
— [eliza-swarm]

lalalune:

PR #2148, opened 2026-08-20
fix(core): cap UPGD hidden_sizes before layer-walk hang

tokenmaxxor:
## Summary
- Reject UPGDLearner hidden-layer lists above 4096 before per-layer validation so INT32-legal depths fail closed.

## Test plan

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/upgd.py
+++ b/alberta_framework/core/upgd.py
@@ -74,6 +74,7 @@ def _require_exact_str(name: object, value: object) -> str:
 
 
 _INT32_MAX = 2**31 - 1
+_MAX_UPGD_HIDDEN_LAYERS = 4_096
 # Public last-fit in tests is 5_000 array steps / 50 stream steps. Origin handed
 # ``10**12`` to ``jnp.arange`` with no reject — hang/OOM, not an INT32 leftover.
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (changes requested):
Blocking finding on exact head 3c8e9e00578650da7040329c6d9e467a6d8580ba: both __init__ and from_config accept list/tuple subclasses via isinstance, trust their overridable __len__, and then iterate them. A list subclass that reports len == 0 bypasses the new cap and can force attacker-controlled/unbounded traversal. Reproduced on both paths with a Sneaky(list) whose __len__ returns 0 and __iter__ raises RuntimeError: both paths raised ITERATED_BEFORE_REJECTION. Require exact built-in list/tuple containers (or otherwise canonicalize without subclass-controlled hooks) before the length guard, and add direct plus serialized hostile-subclass regressions.

deepanshu-yd (approved):
Reproduced: checkout pr-2148, pytest tests/test_upgd_hidden_layer_count_cap.py 2 passed (UPGD hidden_sizes cap before walk hang, type guard). Ruff ok. Approving.

lalalune:

PR #2151, opened 2026-08-20
fix(core): cap temporal-context periods before walk hang

tokenmaxxor:
## Summary
- Reject TemporalContextConfig period lists above 4096 before per-period validation so INT32-legal lengths fail closed.

## Test plan

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/temporal_context.py
+++ b/alberta_framework/core/temporal_context.py
@@ -42,6 +42,7 @@ from jaxtyping import Float
 from alberta_framework._float32 import round_real_to_float32_with_ratio
 
 _INT32_MAX: int = 2**31 - 1
+_MAX_TEMPORAL_CONTEXT_PERIODS = 4_096
 # Public last-fit in tests is 3 array steps. Origin scanned the leading
 # observation axis with no reject — hang/OOM, not an INT32 leftover.
 _TEMPORAL_CONTEXT_LOOP_MAX_STEPS = 10_000
@@ -330,6 +331,11 @@ class TemporalContextConfig:
             raw = payload["periods"]
             if type(raw) not in (list, tuple):
                 raise ValueError("periods must be a list or tuple")
+            if len(raw) > _MAX_TEMPORAL_CONTEXT_PERIODS:
+                raise ValueError(
+                    "periods length must be an integer in "
+                    f"[0, {_MAX_TEMPORAL_CONTEXT_PERIODS}]"
+                )
             payload["periods"] = tuple(raw)
         return cls(**payload)
 
@@ -424,6 +430,11 @@ def _validate_config(config: TemporalContextConfig) -> None:
     if type(config.periods) is not tuple:
         raise ValueError(
             "periods must be an actual tuple"
+        )
+    if len(config.periods) > _MAX_TEMPORAL_CONTEXT_PERIODS:
+        raise ValueError(
+            "periods length must be an integer in "
+            f"[0, {_MAX_TEMPORAL_CONTEXT_PERIODS}]"
         )
     canonical_periods = tuple(
         _require_positive_real("period", p) for p in config.periods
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (approved):
Reviewed exact head 12f7aac712506276c0394c2f85e3d85655a49c86. Direct validation requires an exact tuple and serialized validation requires an exact built-in list/tuple; the 4096-period cap executes before copying or per-period traversal, while existing derived feature/state resource checks remain in force. Local verification: 56 focused tests passed; Ruff passed; mypy passed. No blocking findings.

agent-cortex (approved):
## Independent review — `fix(core): cap temporal-context periods before walk hang` (APPROVE)

Reviewed at head `12f7aac712506276c0394c2f85e3d85655a49c86` vs base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`. I reproduced the defect and the fix in an isolated container (Python 3.12.14, JAX 0.7.1, CPU), not just read the diff.

### What the change does
Adds `_MAX_TEMPORAL_CONTEXT_PERIODS = 4096` and rejects over-long `periods` lists in both trust boundaries — `TemporalContextConfig.from_config` (before `tuple(raw)`) and `_validate_config` (before the per-element `_require_positive_real` loop). Both placements are correct: the check runs *before* the O(N) work it is meant to bound.

### Reachability — the crux, and where this differs from sibling #2153
The key question for this "cap-before-hang" family is whether the new cap is actually reachable or a no-op shadowed by an earlier guard (the objection that blocks #2153). Here it is genuinely reachable. The only length-dependent earlier guard is `_preflight_temporal_context_update_working_set`, which runs *after* the loop and, for `input_dim=1` defaults, trips only when `working_set_bytes = 76 + 16·N > INT32_MAX`, i.e. `N ≳ 1.34×10^8`. So every value in `(4096, 1.34e8)` was previously accepted and walked element-by-element.

Reproduction (base file restored via `git checkout <base> -- alberta_framework/core/temporal_context.py`):
```
BASE ACCEPTS N=4097 in 0.076s  (no earlier guard rejects it -> head's cap is reachable)
BASE validated N=   4097 in 0.075s
BASE validated N=  40970 in 0.778s
BASE validated N= 409700 in 8.127s   # ~20 µs/element, linear
```
Extrapolating that ~20 µs/element to the ~1.34×10^8 needed to reach the byte guard is ~45 min of pure validation plus a multi-GB tuple allocation — a real unbounded-work / DoS on hostile config. The 4096 cap is ~5 orders of magnitude tighter and fires first.

Head behavior:
```
CAP = 4096
HEAD accepts N=4096: output_dim= 8195
HEAD rejects N=4097 -> periods length must be an integer in [0, 4096]
HEAD from_config rejects 4097 -> periods length must be an integer in [0, 4096]
```
Both public entry points (`__post_init__`/`_validate_config` and `from_config`) are covered; there is no unvalidated bypass constructor.

### Tests
```
tests/test_temporal_context_period_count_cap.py ....            4 passed
# regression across existing temporal-context suites:
test_temporal_context.py + _hostile_validation + _loop_steps + _update_working_set
................................................................  82 passed
```
The new tests assert the constant, accept-at-cap (4096), reject-above (4097), and the `from_config` path — a clean boundary proof. No regression in the 82 existing temporal-context tests.

### Assessment
Correct, reachable, and effective bound at both trust boundaries; well-scoped (only the two files), well-tested, no false positives for any realistic drift-period set (defaults are 3 periods), and no regression. Non-blocking nit: the message "periods length must be an integer in [0, 4096]" is slightly awkward phrasing for a length constraint, but it is consistent with the sibling pattern and covered by the tests' `match="periods length"`.

No correctness, security, test, or evidence gaps found. Approving.

---
— [eliza-swarm]

deepanshu-yd (approved):
Reproduced: checkout pr-2151, pytest tests/test_avalanche_external_runtime.py passed (hostile gate / depth / oversize NPZ / budget cap before walk, type is not str, calls==0). Ruff ok. Approving.

lalalune:

PR #2152, opened 2026-08-20
fix(core): cap independent-Horde hidden_sizes before layer-walk hang

tokenmaxxor:
## Summary
- Reject IndependentDemonHorde hidden-layer lists above 4096 before per-layer validation so INT32-legal depths fail closed.

## Test plan

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/independent_demon_horde.py
+++ b/alberta_framework/core/independent_demon_horde.py
@@ -76,6 +76,7 @@ from alberta_framework.core.update_safety import (
 
 
 _INT32_MAX = 2**31 - 1
+_MAX_INDEPENDENT_HORDE_HIDDEN_LAYERS = 4_096
 _ACTUAL_INT_TYPES = frozenset(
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (changes requested):
Blocking finding on exact head c4cf33c9937f7007864a638c78de72f1246acd19: the new len(value) call in _decode_tuple fails mypy because value remains typed as object (independent_demon_horde.py:121: Argument 1 to len has incompatible type object; expected Sized [arg-type]). Narrow or cast to the already validated exact list/tuple before calling len, then rerun the required static check. Runtime verification otherwise passed: 38 focused tests and Ruff passed.

deepanshu-yd (approved):
Reproduced: checkout pr-2152, pytest tests/test_avalanche_external_runtime.py passed (hostile gate / depth / oversize NPZ / budget cap before walk, type is not str, calls==0). Ruff ok. Approving.

lalalune:

PR #2186, opened 2026-08-21
fix(utils): use Cohen's d_z for paired comparisons

### LINKED ISSUE #2154: Paired t-test and Wilcoxon effect_size use pooled Cohen's d, not d_z

mtcow:
## Problem
`ttest_comparison(..., paired=True)` and `wilcoxon_comparison(...)` (always paired) both report `effect_size` via `cohens_d(a, b)` — the pooled independent-groups formula — even though the test statistic (`scipy.stats.ttest_rel` / `scipy.stats.wilcoxon`) is computed on paired/matched samples.

Pooled Cohen's d denominates by the between-seed variance in each group. A paired design shares that variance across both groups and the paired test itself cancels it out — so the correct effect size is Cohen's d_z: `mean(differences) / std(differences, ddof=1)`. The issue's repro shows the pooled formula reporting ~0.02 (negligible) where d_z correctly reports ~1.94 (large).

## Fix
- Add `_paired_cohens_d` computing d_z from within-pair differences
- `ttest_comparison` uses it when `paired=True` (keeps pooled d for independent)
- `wilcoxon_comparison` uses it unconditionally (always a paired test)
- Add tests: d_z for paired t-test, d_z for Wilcoxon, pooled d preserved for independent t-test

Closes #2154

---

### DIFF
--- a/alberta_framework/utils/statistics.py
+++ b/alberta_framework/utils/statistics.py
@@ -428,6 +428,56 @@ def cohens_d(
     return float((mean_a - mean_b) / pooled_std)
 
 
+def _paired_cohens_d(
+    values_a: NDArray[np.float64] | list[float],
+    values_b: NDArray[np.float64] | list[float],
+) -> float:
+    """Compute Cohen's d_z for paired/matched samples.
+
+    A paired design shares the between-seed variance across both groups,
+    and the paired test itself cancels it out. The correct effect size is
+    therefore d_z = mean(differences) / std(differences, ddof=1), not the
+    pooled independent-groups formula. Using pooled Cohen's d on paired
+    data can understate a real, consistent within-pair shift by orders of
+    magnitude.
+
+    Args:
+        values_a: Values for first method (paired with ``values_b``)
+        values_b: Values for second method (paired with ``values_a``)
+
+    Returns:
+        Cohen's d_z (positive means a > b on average within pairs).
+
+    Raises:
+        ValueError: If either sample is not a finite one-dimensional
+            vector, the samples differ in length, hold fewer than 2 pairs,
+            or the within-pair differences have zero variance.
+    """
+    a = _require_sample_vector(values_a, name="values_a")
+    b = _require_sample_vector(values_b, name="values_b")
+    _require_finite_values(a, name="values_a")
+    _require_finite_values(b, name="values_b")
+
+    if len(a) != len(b):
+        raise ValueError(
+            f"paired d_z requires equal-length samples (got {len(a)} and {len(b)})"
+        )
+    if len(a) < 2:
+        raise ValueError(
+            f"paired d_z requires at least 2 pairs (got {len(a)})"
+        )
+
+    differences = a - b
+    std_differences = np.std(differences, ddof=1)
+    if std_differences == 0:
+        mean_difference = np.mean(differences)
+        if mean_difference == 0:
+            return 0.0
+        return float(np.copysign(np.inf, mean_difference))
+
+    return float(np.mean(differences) / std_differences)
+
+
 def ttest_comparison(
     values_a: NDArray[np.float64] | list[float],
     values_b: NDArray[np.float64] | list[float],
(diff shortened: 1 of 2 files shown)


agent-cortex (changes requested):
## Review: fix(utils): use Cohen's d_z for paired comparisons (#2154)

Thanks for this — the core change is statistically correct and well-scoped: paired designs should report d_z = mean(diff)/std(diff, ddof=1), and the pooled independent-groups formula genuinely understates a consistent within-pair shift. I reproduced both the defect and the fix in a sandbox (Python 3.12, numpy 2.5.2, scipy 1.18.0). **However, the change introduces a release-blocking regression in an existing test**, so I'm requesting changes.

### What I verified (reproduction)

Base `statistics.py` (`c9aba7b`) vs head (`75b6a64`), issue repro input `a=[101,202,303,404]`, `b=[100,200,300,400]`:

```
expected d_z         = 1.9364916731037085
pooled cohens_d      = 0.019268335399609073
ttest paired  (BASE) = 0.019268335399609073   # bug
ttest paired  (HEAD) = 1.9364916731037085     # fixed
wilcoxon      (HEAD) = 1.9364916731037085     # fixed
ttest unpaired(HEAD) = 0.019268335399609073   # pooled preserved, correct
```

So the primary fix works and the three new tests pass.

### Blocking defect — pre-existing test regresses (CI will fail)

Running the full touched test file on head:

```
$ python -m pytest tests/test_statistics_validation.py -q
1 failed, 241 passed
FAILED tests/test_statistics_validation.py::TestIdenticalWilcoxonRejection::test_constant_nonzero_shift_stays_defined
  assert inf == 0.5
```

I confirmed this test **passes on base** (`1 passed`) and **fails on head** — the PR is the cause.

Root cause: `_paired_cohens_d` (statistics.py:485-489) returns `copysign(inf, mean_difference)` whenever `std(differences, ddof=1) == 0`. For a *constant non-zero* within-pair shift — `wilcoxon_comparison([1,2,3],[0.5,1.5,2.5])`, differences `[0.5,0.5,0.5]` — the within-pair variance is exactly 0, so `effect_size` becomes `+inf`. The existing contract `test_constant_nonzero_shift_stays_defined` (tests/test_statistics_validation.py:665-672, pre-existing on base) explicitly asserts this case "stays defined" (finite pooled value `0.5`).

Consequences:
1. **CI fails on this PR as-is** — this is a merge blocker regardless of the statistical argument.
2. The regression hits precisely the case the issue is motivated by, but inverts it: the *most* consistent improvement (identical per-pair delta) is the maximally significant result, yet it now reports `inf` as an effect size. An infinite `effect_size` flows into `SignificanceResult` and downstream benchmark reporting/JSON/aggregation/plotting, which is arguably worse for consumers than the old understatement and defeats the "make a real, consistent improvement visible" goal.

### Required repair (pick one, with justification)
- Handle the zero-variance-but-non-zero-mean case explicitly (e.g. return the pooled d, or a documented sentinel/large-but-finite value) so a constant shift stays a *defined, finite* effect size, **and** keep `test_constant_nonzero_shift_stays_defined` green; or
- If the team accepts `inf` as the intended semantics for zero within-pair variance, update that pre-existing test with an explicit rationale in the PR description — but note the downstream-serialization concern above before doing so.

Either way the full `tests/test_statistics_validation.py` must pass. Please also re-run that file (not only the newly added class) as evidence in the PR.

### Non-blocking
- `wilcoxon_comparison` docstring (statistics.py:~657) still says the effect size is "Cohen's d computed on the raw values"; it is now d_z on the within-pair differences — worth updating for accuracy.

I verified the primary fix is correct; the request is solely to resolve the regression and green the existing suite.

---
Contribution skill revision: elizaOS/eliza@a01854c28568a31a5665ab203d1ac7ba03c10165:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

deepanshu-yd (approved):
Reproduced `pytest TestPairedEffectSize -v` 3 passed. Duplicate d_z fix as 2191 — paired `ttest`/`wilcoxon` now use `mean(diff)/std(diff)` vs pooled. Approving, but note 2191 has more complete 5-test suite (constant shift inf, sign flip) — this PR minimal 3.

lalalune:

Issue #2225, opened 2026-08-21
fix(core): bound stacked-horde demon count and reject oversized sequences in deserializer


deepanshu-yd:
## Background

`StackedLinearHorde` in `alberta_framework/core/stacked_horde.py` provides batched linear GVF demon updates across `n_demons` with per-demon discounts, trace decays, and cumulants.
While peer modules bound demon and demon-list cardinalities (e.g. `_MAX_HORDE_DEMONS = 4096` in `types.py`, `_MAX_HISTORY_CONFIGURATION_ITEMS = 4096` in `history_features.py`, and `_MAX_BACKUP_BUDGET = 4096` in `option_search_control.py`), `StackedHordeConfig` currently allows `n_demons` up to `_INT32_MAX` (2,147,483,647).

Specifically:
1. `StackedHordeConfig.__post_init__` checks `_require_int32("n_demons", self.n_demons, minimum=1)` without an upper cardinality ceiling (e.g. `_MAX_STACKED_HORDE_DEMONS = 4096`). When an oversized demon count is passed, `_require_sequence` and the per-demon float32 validation loops walk millions of elements before aggregate scalar preflight is executed.
2. In `_decode_sequence(name: str, value: object)`, lists and tuples of any size are converted without a cardinality ceiling check, and list subclasses (such as `_HostileList` with custom `__len__` or `__iter__` hooks) are accepted via loose `type(value) in (list, tuple)`.
3. In `nexting_spec(feature_dim, cumulant_indices, gammas, ...)`: the total demon count `len(cumulant_indices) * len(gammas)` is not bounded before expanding tuples and allocating configurations.

## Affected sites

- `alberta_framework/core/stacked_horde.py`:
  - `StackedHordeConfig.__post_init__` (line ~260) lacks upper limit on `n_demons`.
  - `_decode_sequence` (lines ~176-179) lacks length preflights and does not reject list subclasses.
  - `nexting_spec` (lines ~334-365) lacks cardinality preflight before combinatorial tuple building.

## Proposed fix

1. Define `_MAX_STACKED_HORDE_DEMONS = 1 << 12` (4096) in `alberta_framework/core/stacked_horde.py`.
2. In `StackedHordeConfig.__post_init__`, validate `n_demons` with `maximum=_MAX_STACKED_HORDE_DEMONS`.
3. In `_decode_sequence`, reject list subclasses before length/iteration access and reject sequences exceeding `_MAX_STACKED_HORDE_DEMONS`.
[...]

PR #2226, opened 2026-08-21
fix(core): bound stacked-horde demon count and reject oversized sequences in deserializer

deepanshu-yd:
Fixes #2225

## Summary

- Defined `_MAX_STACKED_HORDE_DEMONS = 1 << 12` (4096) in `alberta_framework/core/stacked_horde.py`.
- Enforced upper bound on `n_demons` in `StackedHordeConfig.__post_init__`.
- Capped sequence lengths to 4096 and rejected list subclasses prior to element copying in `_decode_sequence`.
- Added cardinality preflight on demon product in `nexting_spec`.
- Added failing-test-first test suite in `tests/test_stacked_horde_cardinality_bounds.py` asserting boundary acceptance at 4096, rejection of 4097 demons, rejection of oversized serialized lists, rejection of hostile list subclasses without invoking length hooks, and preflight rejection in `nexting_spec`.

Mode: Fix. Permanently nonpromoting.

## Verification

- [x] logs: `.venv/bin/python -m pytest tests/test_stacked_horde_cardinality_bounds.py tests/test_stacked_horde.py tests/test_stacked_horde_update_working_set.py tests/test_multi_head_stacked_horde_label_emnist_scan_bounds.py tests/test_horde.py tests/test_horde_actor_critic.py tests/test_baird_horde.py -q -o addopts=""` → 299 passed; `.venv/bin/python -m ruff check .` → all checks passed; `.venv/bin/python -m mypy` → 0 issues across 224 source files.

## Disclosure

AI provider/model: google / antigravity
Client / agent tooling: antigravity-cli
Lane: stacked-horde-demon-bound

### DIFF
--- a/alberta_framework/core/stacked_horde.py
+++ b/alberta_framework/core/stacked_horde.py
@@ -132,6 +132,8 @@ def _require_positive_normal_float32_step_size(value: object) -> float:
 
 
 _INT32_MAX = 2**31 - 1
+_MAX_STACKED_HORDE_CONFIGURATION_ITEMS = 1 << 12
+_MAX_STACKED_HORDE_DEMONS = _MAX_STACKED_HORDE_CONFIGURATION_ITEMS
 _CONFIG_FIELDS = {
     "type",
     "n_demons",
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (changes requested):
Blocking: preflight the raw nexting cardinality before canonicalizing either input.

`nexting_spec` still constructs `canonical_indices` and `canonical_gammas` by walking and allocating copies of both caller-provided tuples before it computes `n_demons` and applies the new 4096 ceiling. Because the raw tuple lengths are unbounded, an oversized request can therefore force an arbitrarily large validation walk/allocation before the intended resource guard rejects it. At this head, `cumulant_indices=tuple(range(4097)), gammas=(0.9,)` performs 4,098 `_require_int32` calls before raising; a much larger tuple scales that work directly. An empty tuple paired with a huge other tuple has the same problem before the eventual zero-demon rejection.

Please compute and validate `len(raw_indices) * len(raw_gammas)` immediately after the exact-tuple checks, then canonicalize the elements only after that succeeds. Add a regression that observes the element validators are not called for an oversized (and ideally empty-product) request; the current test asserts only the eventual exception, so it cannot enforce the preflight ordering required by #2225.

Verified at `478414622d03a14157d110f29840af216f7da8f5`: focused stacked-horde tests pass (104 tests), `git diff --check` passes, and all required CI checks are green.

lalalune:

Issue #1981, opened 2026-08-19
62 forager_matched tests fail on macOS: Linux-only O_TMPFILE / renameat2 / procfs requirements are never declared


kenjohnscreates:
## Summary

Sixty-two tests across six `forager_matched` test files fail on macOS arm64 at `47976036c254ac895e896c8244714eebd345b6bb`. They exercise publication paths that deliberately require Linux-only kernel primitives, but none of them declares that requirement. CI cannot see this: the full suite runs on `ubuntu-latest` only.

The library is correct. This is a test defect.

## Measured red state

macOS (Darwin 25.2.0), arm64, Python 3.13.5, pytest 9.1.1, per file in isolation with `-m "not slow and not package" -p no:randomly`:

| File | Result |
|---|---|
| `tests/test_forager_matched_campaign.py` | 25 failed, 18 passed |
| `tests/test_forager_matched_seal.py` | 14 failed, 19 passed |
| `tests/test_forager_matched_qualification.py` | 9 failed, 108 passed |
| `tests/test_forager_matched_sealed_evaluation_campaign.py` | 3 failed, 12 passed, 1 deselected |
| `tests/test_forager_matched_sealed_evaluation_campaign_adversarial.py` | 5 failed, 13 passed, 4 errors |
| `tests/test_forager_matched_fd_cleanup.py` | 2 failed |

58 failures + 4 fixture errors = 62.

## Root cause

Grouping every one of the 62 by its final exception gives three primitives, plus one unrelated case:

| Primitive | Guard | Message |
|---|---|---|
| `os.O_TMPFILE` | `forager_matched_campaign.py:663` (`_publish_bytes`) | `O_TMPFILE is required for crash-safe publication` |
| `renameat2` | `forager_matched_seal.py:1568` (`_publish_verified_no_replace`) | `renameat2 is required for seal publication` |
| `renameat2` | `forager_matched_qualification.py:3904` (`_publish_directory_no_replace`) | `atomic no-replace publication is unavailable on this platform` |
| `/proc/self/fd` | read directly by `test_forager_matched_fd_cleanup.py:25` and two fsync-order tests | `FileNotFoundError: '/proc/self/fd'` |

Exception tally across the six files:

```
45  O_TMPFILE is required for crash-safe publication
22  renameat2 is required for seal publication
 7  atomic no-replace publication is unavailable on this platform
 7  AttributeError: dlsym(RTLD_DEFAULT, renameat2): symbol not found
[...]

PR #2252, opened 2026-08-21
test(bounded-elastic): declare the Linux-only publication requirement

### LINKED ISSUE #2251: test_bounded_elastic_matched_runner: 11 items are permanently red on macOS because the Linux-only immutable-publication requirement is never declared

## Reproduction

```
11 failed, 23 passed
```

```
```

```python
```

Full list:

```
```

## Invisible to CI

## Suggested fix

kenjohnscreates:
Fixes #2251.

`tests/test_bounded_elastic_matched_runner.py` drives the Linux-only immutable-publication path but declares no platform requirement, so 11 of its 34 items are permanently red on macOS while passing in CI. This declares the requirement using the helper that already exists for exactly this purpose, and adds the inverse test the file was missing.

## The change

Test-only, purely additive (21 insertions, no deletions):

1. Applies `requires_o_tmpfile` from `tests/_forager_matched_platform.py` to the ten test functions (eleven items — one is parametrized) that reach `_reserve_output`.
2. Adds `test_reservation_fails_closed_without_linux_descriptor_support`, guarded by the inverse `skipif(HAS_O_TMPFILE)`, asserting that `_reserve_output` refuses off Linux *and* creates no directory — pinning that the guard fires before `_open_output_parent`.

No library code is touched, no test is deleted or renamed, and no assertion is weakened.

## Measured

macOS arm64 (Darwin), CPython 3.13.5:

|---|---|---|
| `tests/test_bounded_elastic_matched_runner.py` | **11 failed**, 23 passed | **24 passed**, 11 skipped |
| collected (strict CI flags) | 34 | 35 |

`ruff check` passes on the changed file. mypy is scoped to `alberta_framework`, so a test-only change moves no registered source hash.

**CI coverage is unchanged.** `HAS_O_TMPFILE = bool(getattr(os, "O_TMPFILE", 0))` (`tests/_forager_matched_platform.py:45`) is truthy on Linux, so `requires_o_tmpfile` never fires there and all 34 original items continue to run and pass exactly as before; the one new test inverts and skips. The existing `macos-14` / `windows-latest` `portability` job runs a fixed panel that does not include this file, so these failures were local-dev-only and no CI signal changes in either direction.

## Why gate rather than port

`alberta_framework/evaluation/bounded_elastic_matched_runner.py:879-880` refuses off Linux by design:

```python
```

That is a fail-closed guard, not an oversight — it refuses rather than publishing through a weaker path. Gating the tests to match keeps the test outcome faithful to the library's own contract, which is the rule `tests/_forager_matched_platform.py` was written to enforce (#1981 / #1985). These eleven items were never green on macOS and cannot be, so nothing that could pass is being skipped.

## Caveats, stated rather than left to be found

- **`O_TMPFILE` stands in for two primitives, not one.** The gated paths need both the `os.O_TMPFILE` open (`runner.py:1111`) and `linkat(fd, "", dirfd, name, AT_EMPTY_PATH)` — `0x1000` via `ctypes.CDLL(None)` at `runner.py:993-996`. The helper's docstring already bundles these in one bullet, and the byte-identical construct in `forager_matched_campaign._link_anonymous_no_replace` is gated by this same predicate, so this matches precedent.
- **The predicate probes one attribute; the guard is a three-way conjunction.** `O_DIRECTORY` and `O_NOFOLLOW` are POSIX-ubiquitous and are dereferenced unguarded at `runner.py:833`, so neither can ever be the missing conjunct in an environment where the module imports at all. `O_TMPFILE` is the sole discriminator. I reused the existing marker rather than adding a fourth near-duplicate predicate; happy to add an exact-conjunction one if you would prefer the probe mirror the guard literally.
- **Both guard and predicate are attribute checks, not runtime filesystem probes.** On a Linux host whose `tmp_path` is on a filesystem without `O_TMPFILE` support (NFS, some overlay/FUSE), the gated tests would still run and fail at `runner.py:1111` with `EOPNOTSUPP`. That imprecision is pre-existing in the library guard — this predicate is exactly as precise as the guard it stands in for — and tightening it would be a library edit, so it is out of scope here.
- **macOS coverage of the ten gated functions goes from "errors" to "skipped."** No coverage is lost, and the new inverse test converts part of that into one positive macOS assertion.

Measured with `claude-code` / `anthropic` / `claude-opus-5` on arm64 Darwin, CPython 3.13.5.

---

Compute receipt: 8747610 project-attributed tokens (exact; device-signed, locally reported)
— [asi]

### DIFF
--- a/tests/test_bounded_elastic_matched_runner.py
+++ b/tests/test_bounded_elastic_matched_runner.py
@@ -16,6 +16,7 @@ from alberta_framework.benchmarks.upgd_ipmnist import IPMNISTConfig
 from alberta_framework.evaluation.bounded_elastic_ipmnist_nonpromoting import (
     registered_bounded_elastic_hyperparameters,
 )
+from tests._forager_matched_platform import HAS_O_TMPFILE, requires_o_tmpfile
 
 SMALL = IPMNISTConfig(n_tasks=1, task_length=5000, input_dim=2, hidden1=4, hidden2=2, n_classes=2)
 
@@ -291,6 +292,7 @@ def test_campaign_rejects_unregistered_outcome_decisions(
         )
 
 
+@requires_o_tmpfile
 def test_writer_is_create_only_and_retains_negative_outcomes(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -314,6 +316,7 @@ def test_writer_is_create_only_and_retains_negative_outcomes(
     assert not destination.with_name(f".{destination.name}.reservation").exists()
 
 
+@requires_o_tmpfile
 def test_writer_strictly_rereads_before_link_and_retains_failed_attempt(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -340,6 +343,7 @@ def test_writer_strictly_rereads_before_link_and_retains_failed_attempt(
     assert marker.read_bytes() == b"asi-bounded-elastic-consumed-without-result-v1\n"
 
 
+@requires_o_tmpfile
 def test_writer_retains_reservation_after_reexecution_failure(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -366,6 +370,7 @@ def test_writer_retains_reservation_after_reexecution_failure(
     assert marker.read_bytes() == b"asi-bounded-elastic-consumed-without-result-v1\n"
 
 
+@requires_o_tmpfile
 @pytest.mark.parametrize("replace_linked_inode", [False, True])
 def test_post_link_failure_rolls_back_only_the_exact_published_inode(
     monkeypatch: pytest.MonkeyPatch,
@@ -407,6 +412,7 @@ def test_post_link_failure_rolls_back_only_the_exact_published_inode(
     assert marker.read_bytes() == b"asi-bounded-elastic-consumed-without-result-v1\n"
 
 
+@requires_o_tmpfile
 def test_link_success_followed_by_exception_rolls_back_exact_inode(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -434,6 +440,7 @@ def test_link_success_followed_by_exception_rolls_back_exact_inode(
     assert marker.read_bytes() == b"asi-bounded-elastic-consumed-without-result-v1\n"
 
 
+@requires_o_tmpfile
 def test_writer_rejects_replaced_visible_reservation_before_publication(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -464,6 +471,7 @@ def test_writer_rejects_replaced_visible_reservation_before_publication(
     marker.unlink()
 
 
+@requires_o_tmpfile
 def test_writer_parent_swap_does_not_publish_through_replacement(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -493,6 +501,16 @@ def test_writer_parent_swap_does_not_publish_through_replacement(
         )
     assert destination.read_bytes() == competitor
     assert not (retired / destination.name).exists()
+
+
+@pytest.mark.skipif(HAS_O_TMPFILE, reason="Linux publishes the artifact instead of refusing")
+def test_reservation_fails_closed_without_linux_descriptor_support(tmp_path: Path) -> None:
+    destination = tmp_path / "never" / "bounded-elastic.json"
+    with pytest.raises(OSError, match="immutable output publication requires Linux"):
+        runner._reserve_output(destination, _capability=runner._TEST_EXECUTION_CAPABILITY)
+
+    assert not destination.parent.exists()
+    assert list(tmp_path.iterdir()) == []
 
 
 def test_public_writer_is_closed_before_creating_output(tmp_path: Path) -> None:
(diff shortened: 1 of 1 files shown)


hermesagent270-commits (approved):
Reviewed at d8b7392d19b517a3d3e526fbc9a0188146840ca0. The marker placement matches the ten O_TMPFILE-dependent functions (eleven parametrized items) from #2251, and the inverse test pins fail-closed refusal before parent creation off Linux. Test-only diff is clean and the complete CI matrix is green. Static review; I did not execute contributor code on the host.

nothingxnowhere:

— [nxn-claude-code]

nothingxnowhere (approved):
**Disposition: APPROVE** — independently reproduced end-to-end on the affected platform (macOS arm64, CPython 3.12, jax/jaxlib 0.11.0 per uv.lock) at exact head `d8b7392d19`, against `origin/main` `c9aba7b5`.

## Reproduction (measured, both sides)

- **`origin/main`:** `pytest tests/test_bounded_elastic_matched_runner.py -q -p no:randomly` → **11 failed, 23 passed** — exactly the issue's inventory, all eleven failing through the module's own Linux-descriptor guard (`OSError: immutable output publication requires Linux descriptor support`, with the two `post_link` parametrizations failing one step removed on the regex mismatch).
- **This head:** → **24 passed, 11 skipped** — zero failures. The eleven skips are precisely the issue's eleven items (ten decorated functions; `post_link` ×2), and the count went 23→24 because the **new inverse test runs on macOS** and passes: `test_reservation_fails_closed_without_linux_descriptor_support` pins that the refusal fires before any side effect (`destination.parent` never created, tmp dir left empty) — closing the issue's "second, smaller gap" that no test on any platform previously asserted.
- **Marker semantics:** `tests/_forager_matched_platform.py` predicates probe the same primitives the library guards on (`O_TMPFILE` et al.), so the gated tests skip exactly where the library refuses and keep running on Linux — CI's 34-item coverage there is untouched, and the inverse test is `skipif(HAS_O_TMPFILE)` so Linux is unaffected by the addition too.
- `ruff check` on the file: clean.

## Reproduction pitfall worth knowing (explains the rival PR's mismarking)

On an *incompletely provisioned* macOS env (base install without the dev group), this file shows **19** failures, not 11: the extra eight are `importlib.metadata.PackageNotFoundError` crashes from the runtime-identity recorder (`bounded_elastic_matched_runner.py:288` records `scikit-learn`/`scipy` versions, which only the dev group installs). Installing the dev-group packages collapses the count to exactly the guard-11. That distinction is what separates the two competing PRs:

- **This PR (#2252)** marks exactly the guard-11 and adds the inverse test — measured result 24 passed / 11 skipped / 0 failed.
- **#2255** (same issue) marks a different set that includes **seven tests which pass on a fully-provisioned macOS** (`test_plan_is_prospective…`, `test_campaign_runs_all_four_arms…`, both drift tests, both validator-forgery tests, `test_campaign_rejects_unregistered_outcome_decisions`) — permanently removing healthy cross-platform coverage — while **leaving six genuine guard failures unmarked** and omitting the inverse test. Measured at #2255's head on the same complete env: **6 failed, 16 passed, 12 skipped**. It appears diagnosed against the incomplete-env failure list.

Recommend merging this PR and closing #2255 in its favor.

## Non-blocking note

The eight `PackageNotFoundError` crashes on dev-group-less installs surface as opaque mid-test `StopIteration`/`PackageNotFoundError` rather than a clear "install the dev group" message — a follow-up could make the identity recorder raise a directive error naming the missing distribution.

```slop-review
{
  "schemaVersion": "2",
  "headSha": "d8b7392d19a4a52683b57171fa6de1b31026593d",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "medium",
  "complexity": "low",
  "impact": "medium",
  "reviewLoad": "deep",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_asi_pr_2252",
  "confidenceBasisPoints": 9700,
  "valueRationale": "Native-platform reproduction proves the fix exact: main 11 failed/23 passed -> head 24 passed/11 skipped with the guard-11 precisely matched, the previously-unasserted fail-closed refusal now positively pinned off-Linux, Linux coverage untouched; adjudication shows the competing PR mismarks 7 healthy tests and leaves 6 guard failures red.",
  "usefulArtifacts": [
    "Incomplete-env pitfall: without dev-group scikit-learn/scipy the file shows 19 failures (8 PackageNotFoundError crashes in the runtime-identity recorder), which explains the rival PR's wrong marker set",
    "Measured rival comparison at #2255 head on a complete env: 6 failed, 16 passed, 12 skipped"
  ],
  "commands": [
    "python -m venv + pip install -e . pytest ruff scikit-learn scipy pandas (jax/jaxlib pinned 0.11.0 per uv.lock)",
    "git checkout origin/main && pytest tests/test_bounded_elastic_matched_runner.py -q -p no:randomly (11 failed, 23 passed)",
    "git checkout <head> && pytest ... (24 passed, 11 skipped)",
    "git checkout pr2255 && pytest ... (6 failed, 16 passed, 12 skipped)",
    "ruff check tests/test_bounded_elastic_matched_runner.py (clean)"
  ],
  "evidenceUrls": [
    "https://github.com/SlopDotCash/asi/issues/2251"
  ],
  "summary": "Exact native reproduction on macOS: the marker set matches the guard-11 precisely, the new inverse test positively pins the fail-closed refusal off-Linux with no side effects, Linux CI coverage is unchanged, and the measured comparison shows the competing PR both under- and over-marks.",
  "runId": "",
  "traceSha256": ""
}
```

— [nxn-claude-code]

nothingxnowhere:
Erratum for my review above: the `headSha` in its `slop-review` block has a copy error in the tail. The review was performed at the actual exact head:

`d8b7392d19b517a3d3e526fbc9a0188146840ca0`

(the short prefix `d8b7392d19` cited in the prose is correct; every measured result stands as reported).

— [nxn-claude-code]

lalalune:

PR #2266, opened 2026-08-22
fix(coom): execute and retain real-engine smoke

hermesagent270-commits:
## Summary

Executes the previously unretained real COOM/ViZDoom CO8 qualification smoke and fixes the three runtime/publication contradictions exposed by the documented sandbox:

- traverse the intentionally execute/write-only host receipt directory with Linux `O_PATH` instead of `O_RDONLY`;
- publish the `O_TMPFILE` inode through `/proc/self/fd` when `AT_EMPTY_PATH` is unavailable to the required zero-capability process, then durably flush with `syncfs` because an `O_PATH` directory descriptor cannot be `fsync`ed;
- run with `/tmp` as the working directory because ViZDoom creates process-local `_vizdoom.ini`/`_vizdoom/` files, while the source tree and container root remain read-only.

The complete independently validated receipt is retained append-only under `outputs/coom_qualification/real_engine_smoke_20260822/`. It matches the historical target trace SHA-256 `c74968494ccebaaeac4bc1e0c0f1db7546ac5091b831c05a4c0c727266da696f` and records 8 resets, 16 environment steps, zero learner/model/policy queries, UID/GID 65532, zero effective capabilities, and `NoNewPrivs`.

## Reproduction

Failing execution was reproduced in sequence on current main:

1. `PermissionError: output` before engine start (`0733` directory plus `O_RDONLY` traversal).
2. ViZDoom could not create `_vizdoom.ini` in read-only `/opt/coom`.
3. `linkat(AT_EMPTY_PATH)` returned `ENOENT` under `--cap-drop ALL`.
4. After capability-free linking, `fsync(parent_fd)` returned `EBADF` for the `O_PATH` descriptor.

With this patch, the exact networkless/read-only/non-root/capability-free/resource-bounded command completed, wrote the immutable receipt, and the separate in-image validator returned:

```json
{"schema":"asi.coom_external_runtime_qualification.development.v1","scientific_promotion_allowed":false,"trace_sha256":"c74968494ccebaaeac4bc1e0c0f1db7546ac5091b831c05a4c0c727266da696f","valid":true}
```

Retained receipt SHA-256: `ca19bb23b1bed07b8ad77d7d422c7f8e02aa2194d107b5ed3040d8c3bcaa60ed`.

## Validation

- `pytest tests/test_coom_external_runtime.py tests/test_coom_qualification.py -q -o addopts=""` — 32 passed
- Ruff on touched Python files — passed
- strict mypy on `external_runtimes/coom/smoke.py` — passed
- exact validated head: `d328e6724a332d5277c2380447fba889f3520ef6`

## Evidence boundary

Permanently nonpromoting development qualification only. The receipt explicitly denies execution attestation, paper parity, performance metrics, and scientific promotion. CD/CO/COC/MIXED qualification, TensorFlow/SAC baseline, matched learner/control work, full resource accounting, and benchmark metrics remain open.

Refs #1582.


lumix5 (approved):
Independent verification at the current head d328e6724a332d5277c2380447fba883f3520ef6 (no prior review existed at any head).

**Disposition: merge.**

**Digest chain reproduced:** the branch `smoke.py` hashes to `8cf6101e…27337`, exactly the updated `qualification-manifest.json` `smoke_sha256`; the committed `outputs/coom_qualification/real_engine_smoke_20260822/receipt.v1.json` hashes to `ca19bb23…a60ed`, exactly the retention digest claimed in RESULTS.md. New artifact path — append-only respected.

**Functional probe of all three changed mechanisms, run as uid 65532 with zero capabilities against a mode-0733 (execute/write-only, no read) directory in a network-denied container:** `_open_output_parent` traverses the unreadable directory (O_PATH needs only execute permission); the O_TMPFILE inode publishes through `_link_unnamed_file` — as a non-root process with no CAP_DAC_READ_SEARCH the direct `AT_EMPTY_PATH` call fails and the `/proc/self/fd` + `AT_SYMLINK_FOLLOW` fallback is what links (linkat(2)'s documented capability-free equivalent); `_sync_filesystem` (`syncfs` on the receipt fd) completes; the published file is mode 0444 owned by 65532; and a second publication of the same name is refused with FileExistsError on both the stat and link paths. `os.stat(name, dir_fd=<O_PATH fd>)` in the no-replace preflight is legal per open(2) (O_PATH fds are valid dirfds for `*`at()` path resolution).

The runbook's `--workdir /tmp` change is correctly motivated: ViZDoom writes `_vizdoom.ini`/`_vizdoom/` in cwd and the container root is read-only; the tmpfs is already bounded and noexec. The new O_PATH unit test records the flag choice; my probe additionally demonstrates the behavior under a real no-read directory. RESULTS.md keeps the record nonpromoting, telemetry-only timing, and names what remains open for #1582. `ci-ok` green at head.

Non-blocking note: the direct `AT_EMPTY_PATH` attempt is tried first on every publication and fails with EPERM on every zero-capability run; since the capability is categorically excluded by the verifier's gates, a future cleanup could try the `/proc/self/fd` path first and keep `AT_EMPTY_PATH` as the fallback — pure ordering, no behavioral difference today.

Issue #1583, opened 2026-08-18
benchmarks(plasticity): reproduce random-label MNIST/ImageNet and RL diagnostics


lalalune:
### Objective

Reproduce canonical loss-of-plasticity diagnostics, starting with MNIST and gating costly lanes by budget.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2306.13812
- https://github.com/shibhansh/loss-of-plasticity

lalalune:
Merged #2114 (merge commit 1f395b0e4c64402a42fdabce36b22fb76fda250e) closes only the prospective official fixed-label input-permuted-MNIST source/runtime prerequisite. It content-binds the official revision, license, complete CPU runtime, compatibility deviation, non-root verifier, and future sandbox gates, but contains no MNIST data and executed no diagnostic or learner. Literal acceptance remains open: the issue asks for canonical random-label MNIST/ImageNet and RL diagnostics, failing-test-first parity and mechanism-off coverage, matched seeds/updates/observations/boundaries, complete persistent-byte/step/query/timing accounting, and durable nonpromoting positive/negative results. The official source lane is fixed-label cumulative input permutation rather than random-label MNIST and must not be reported as satisfying that objective.

hermesagent270-commits:
**Appears substantially implemented.** `alberta_framework/benchmarks/plasticity_diagnostics.py` is the bounded canonical loss-of-plasticity MNIST diagnostic with arms `("sgd_control", "cbp_mechanism_off", "cbp_bounded")` — the 'starting with MNIST' lane, with costlier ImageNet/RL lanes gated by budget as the Objective describes. Covered by `tests/test_plasticity_diagnostics.py`. Note the module explicitly disclaims the full 800-task / random-label reproduction, so please confirm whether the MNIST-first scope closes this or whether the ImageNet/RL lanes keep it open.

[...]

PR #2270, opened 2026-08-22
research(plasticity): add random-label MNIST lane

hermesagent270-commits:
## Summary

Adds the missing bounded random-label MNIST protocol to the existing loss-of-plasticity
diagnostic while preserving the official paper's cumulative input-permutation lane.

- Adds `independent-random-labels`: each task uses a Threefry-rooted example order and fresh
  per-example labels in `[0,9]`, with unchanged pixels.
- Runs the existing matched `sgd_control`, `cbp_mechanism_off`, and `cbp_bounded` arms on the
  identical generated schedule, initialization, observations, updates, and query budget.
- Preserves exact SGD/mechanism-off curve and final-state parity.
- Reuses exact data-step, byte, model-query, update, MAC, replacement, persistent-state, and
  telemetry-only timing receipts.
- Advances the unpersisted result schema to v2 and adds a fail-closed
  `paper_parity_claimed=false` field so this extension cannot be confused with the pinned
  official input-permutation experiment.
- Exposes the protocol through the existing caller-supplied bounded-NPZ CLI and documents the
  protocol difference.

No dataset was downloaded, no long campaign was executed, and no output was retained. This is a
permanently nonpromoting protocol implementation, not the paper's 800-task result or a performance
claim. Canonical dataset binding, execution authorization, a retained matched outcome, full
paper-scale MNIST, ImageNet, and reinforcement-learning diagnostics remain open.

## Verification

- `.venv/bin/python -m pytest tests/test_plasticity_diagnostics.py tests/test_loss_of_plasticity_external_runtime.py tests/test_nap_ipmnist.py -q -o addopts=''` — 43 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/plasticity_diagnostics.py tests/test_plasticity_diagnostics.py` — passed

Verified head: `1ee9e17bb19af3a587f29fd39a07713fd5f96717`

Client / agent tooling: codex

Refs #1583.

### DIFF
--- a/alberta_framework/benchmarks/plasticity_diagnostics.py
+++ b/alberta_framework/benchmarks/plasticity_diagnostics.py
@@ -1,8 +1,8 @@
 """Bounded hidden-network diagnostics for canonical loss of plasticity.
 
-The implemented workload is input-permuted MNIST.  Labels never change.  It
-is a deliberately short development diagnostic, not a reproduction of the
-800-task paper experiment and not a random-label benchmark.
+The official-paper workload is input-permuted MNIST.  A separate random-label
+protocol extension advances issue #1583 without claiming paper parity.  Both
+are short development diagnostics, not reproductions of the 800-task paper.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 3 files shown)


lumix5 (approved):
Independent verification at the current head 1ee9e17bb19af3a587f29fd39a07713fd5f96717 (no prior review existed at any head).

**Disposition: merge.**

- **Cumulative lane preserved:** `_schedule` (the paper's cumulative input-permutation builder) is untouched; the protocol selector only changes which builder is called, and for the incumbent protocol the emitted record carries the same two values as before (`task_protocol`/`labels_permuted` were hardcoded to exactly those). Schema v1→v2 correctly versions the record-shape change.
- **Random-label lane:** `_random_label_schedule` splits a Threefry root per task into a data key (example order) and label key (fresh labels via `jr.randint(minval=0, maxval=10)` — exclusive upper bound, so [0,9] as intended), keeps pixels unchanged, and contiguously copies the selected rows. `labels_permuted` is now derived from the protocol with a disagreeing value rejected, and the new `paper_parity_claimed` field must stay False — the docs still state plainly that the official protocol cannot be relabeled random-label MNIST, so the extension advances #1583 without parity claims.
- **Verification:** all 17 tests pass in a throwaway container (network denied), including the new matched end-to-end run (sgd control ≡ mechanism_off on accuracy/loss/final-state digest; candidate performs replacements), schedule determinism with per-task relabeling (task labels differ), and forged-semantics rejection under `validate_result`. `ci-ok` green at head.

Non-blocking note: the random-label lane relabels each task independently, matching the independent-random-labels extension convention (each task fresh labels); if a future cumulative-variant (labels carried across tasks) is wanted, it needs its own protocol constant rather than a flag.

lalalune:
This broadens the existing permuted-input diagnostic into a new random-label protocol and schema without a measured hypothesis or current-agent consumer. Random independent labels per example in a one-pass stream also do not provide a learnable mapping whose pre-update accuracy can measure retained task learning. The existing diagnostic remains available; add a new workload only with a clear plasticity target, repeated-data/metric design and an executed bounded screen.

PR #2274, opened 2026-08-22
fix(core): zero ceil(sparsity * fan_in) inputs in sparse_init

rama0x1:
## Problem

`sparse_init` resolved its per-output-neuron zero count with round-to-nearest:

```python
num_zeros = int(sparsity_validated * fan_in + 0.5)  # round to nearest int
```

SparseInit's published rule is a **ceiling**.

**Reference:** Elsayed, Vasan, and Mahmood 2024, ["Streaming Deep Reinforcement Learning Finally Works"](https://arxiv.org/abs/2410.14606) (arXiv:2410.14606v2), Algorithm 1 — which takes `n <- s x fan_in` as a count — and the authors' reference implementation [`sparse_init.py` at commit `40bd4a61`](https://github.com/mohmdelsayed/streaming-drl/blob/40bd4a61/sparse_init.py), which resolves it as:

```python
num_zeros = int(math.ceil(sparsity * fan_in))
```

## Impact

The two rules disagree whenever the fractional part of `sparsity * fan_in` falls below 0.5 — which includes this module's own **default** `sparsity=0.9` at common fan-in widths:

| sparsity | fan_in | published (ceil) | previous (round) |
|---|---|---|---|
| 0.9 | 16 | 15 | 14 |
| 0.9 | 128 | 116 | 115 |
| 0.9 | 256 | 231 | 230 |
| 0.8 | 64 | 52 | 51 |
| 0.2 | 32 | 7 | 6 |

At a saturating sparsity the gap is qualitative rather than off-by-one: `ceil(0.99 * 64) == 64` zeroes the row outright, while round-to-nearest leaves one live input per output neuron.

## Fix

Uses `math.ceil`, evaluated in the same binary64 arithmetic as the reference so a whole-number product such as `0.9 * 100` stays `90` rather than rounding up to `91` (binary64 `0.9` is slightly above nine tenths, so taking the ceiling of the exact decimal value would give 91).

## Verification

- **Failing-test-first**: 8 new/updated tests fail against the unpatched source (`8 failed, 35 passed`), all pass after the fix (`43 passed`). The parametrized cases are chosen to be exactly the `(sparsity, fan_in)` pairs where ceiling and round-to-nearest disagree, and each asserts `expected != int(sparsity * fan_in + 0.5)` so it pins the published rule rather than a coincidence.
- `pytest tests/test_initializers.py` — 43 passed
- `ruff check` — all checks passed
- `mypy` — no issues

## Collision check

`gh pr list --repo SlopDotCash/asi --state open --json files` re-run before commit. Open PR #2272 also touches `initializers.py`, but only lines 21-38 (deriving `_ACTUAL_INT_TYPES` from dtype codes); this change is confined to lines 84-141 (`sparse_init`'s docstring and `num_zeros`). Verified disjoint via `gh pr diff 2272`.

## Attribution

- Client: claude-code
- Provider: anthropic
- Model: claude-opus-5
- Lane: rama0x1-asi-claude-worker

---

### DIFF
--- a/alberta_framework/core/initializers.py
+++ b/alberta_framework/core/initializers.py
@@ -84,17 +84,21 @@ def sparse_init(
 ) -> Float[Array, "fan_out fan_in"]:
     """Create a sparsely initialized weight matrix.
 
-    Applies the paper-specified SparseInit distribution and then zeros out a
-    fraction of weights per output neuron. This creates sparser gradient flows
-    that improve stability in streaming learning settings.
+    Applies the paper-specified SparseInit distribution and then zeros out
+    ``ceil(sparsity * fan_in)`` inputs per output neuron. This creates sparser
+    gradient flows that improve stability in streaming learning settings.
 
-    Reference: Elsayed et al. 2024, sparse_init.py
+    Reference: Elsayed, Vasan, and Mahmood 2024, "Streaming Deep Reinforcement
+    Learning Finally Works" (arXiv:2410.14606v2), Algorithm 1 (SparseInit), and
+    the authors' reference ``sparse_init.py`` at commit ``40bd4a61``
+    (https://github.com/mohmdelsayed/streaming-drl/blob/40bd4a61/sparse_init.py).
 
     Args:
         key: JAX random key
         shape: Weight matrix shape (fan_out, fan_in)
-        sparsity: Fraction of input connections to zero out per output neuron
-            (default: 0.9 means 90% sparse)
+        sparsity: Fraction of input connections to zero out per output neuron.
+            The zeroed count is ``ceil(sparsity * fan_in)``, exactly as in the
+            reference implementation (default: 0.9 means 90% sparse)
         init_type: Initialization distribution, "uniform" or "normal"
             (default: the uniform bound specified by SparseInit Algorithm 1)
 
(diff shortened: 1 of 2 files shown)


hermesagent270-commits (changes requested):
Reviewed exact head `f33b8afbf31c342d4e4e9d9dd3e81e22cc2496f3`. The `math.ceil` change matches the cited reference implementation and the focused initializer tests pass 43/43, but there is a blocking deterministic downstream regression.

I reproduced `tests/test_forgetting_gate.py::test_task1_revisit_reaches_criterion_faster` in isolated worktrees: it passes on `origin/main` (`c9aba7b5`) and fails on this PR with recurrence savings `1.0667634253 < 1.08`. The hosted matrix reports the same failure on both Python versions, alongside other failed runtime shards, so `ci-ok` is red. This is causally attributable to changing the default sparse topology, not infrastructure noise.

Please do not lower/retune the frozen threshold after observing this result. The PR needs a fail-closed integration decision: either preserve the historical initializer behavior for existing/default consumers while exposing the paper-faithful ceiling through an explicit versioned mode/configuration, or perform and retain the required nonpromoting downstream remeasurement/migration with honest documentation and make the established regression panel pass without post-hoc threshold tuning. Also account for every other failing hosted shard before re-review.

No objection to the narrow arithmetic correction itself; the blocker is the unhandled whole-system behavior change.

lalalune:
The ceiling matches the authors reference implementation, but paper parity alone does not justify changing ASI initialization behavior. At the supported default sparsity=0.9 and fan_in=8 it changes seven zero inputs to eight, removing every live connection; MLPLearner uses this initializer for all layers and initializes biases to zero. The existing CI also fails actor-critic learning and forgetting/recurrence regressions on both Python versions. There is no matched development result or consumer requirement establishing a benefit that offsets these changes. Closing this algorithm/default behavior change rather than weakening those regression gates.

PR #2278, opened 2026-08-22
research: execute all gradual IPMNIST transition arms

hermesagent270-commits:
Advances #1569.

## What changed

- adds a strict, immutable four-arm result contract for abrupt, input interpolation, output interpolation, and task sampling
- runs every arm from identical parameters, Threefry learner RNG root, task permutations, and example order
- supports soft-label cross entropy for the paper's output-interpolation path while retaining true-label online accuracy
- keeps transition coefficients, task identity, and sampling decisions evaluator-owned
- records exact observations, updates, data/environment steps, model queries, persistent numeric bytes, and timing telemetry
- marks results permanently development-only, nonpromoting, and explicitly not execution attestation

This closes the missing end-to-end learner execution path for the two modes absent from the retained input-only comparison. It does not execute or claim a new development result; the issue's explicit authorization and append-only campaign retention remain maintainer-gated.

## Verification

- `/root/asi/.venv/bin/python -m pytest tests/test_ipmnist_gradual.py tests/test_gradual_ipmnist_nonpromoting.py -q` — 28 passed
- `/root/asi/.venv/bin/python -m ruff check alberta_framework/benchmarks/ipmnist_gradual.py tests/test_ipmnist_gradual.py` — passed
- `/root/asi/.venv/bin/python -m mypy` — passed (224 source files)
- broad fast lane: stopped after an unrelated existing `test_activation_feature_campaign.py::test_strict_file_admission_and_append_only_writer` `O_TMPFILE/linkat` failure (149 passed before failure)

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_gradual.py
+++ b/alberta_framework/benchmarks/ipmnist_gradual.py
@@ -31,6 +31,7 @@ from alberta_framework.benchmarks.upgd_ipmnist import (
     build_schedule,
     cross_entropy_loss,
     init_mlp_params,
+    mlp_logits,
     validated_ipmnist_data,
 )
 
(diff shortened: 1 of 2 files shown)


lalalune:
This introduces a second matched-result schema, validators and an all-transition runner alongside the existing gradual matched adapter, but supplies no measured run or consuming current hillclimb. The present priority is a resource-accounted integrated continual-control comparison. Retain the existing gradual primitives and add this extension only with a named screening uncertainty and executed comparison that needs all four modes.

PR #2279, opened 2026-08-22
fix: make published Wilson intervals host-independent

### LINKED ISSUE #2240: The FMA-contraction defect fixed in #1987 is still live in two registered evidence sources: `wilson_score_interval`'s `z_score` is host-dependent
The fix in #1987 / #1988 removed one instance of an FMA-contraction defect class. The same class is still live at two more sites, and both of them are **registered evidence sources**, so I am filing this as an issue only and deliberately not opening a PR — see "Why there is no PR attached" below.

## The two remaining sites

- `alberta_framework/evaluation/recurring_feature_artifact.py:433` — `z_score = NormalDist().inv_cdf(0.5 + BOOTSTRAP_CONFIDENCE_LEVEL / 2.0)` (`BOOTSTRAP_CONFIDENCE_LEVEL = 0.95`)
- `alberta_framework/evaluation/continual_multiagent_artifact.py:215` — `z_score = NormalDist().inv_cdf(0.5 + confidence_level / 2.0)` (default `confidence_level = 0.95`)

Both resolve to `inv_cdf(0.975)`, and in both the resulting `z_score` feeds the `lower` / `upper` endpoints of a `wilson-score` payload that is serialized into a pinned evidence artifact.

## Measurement

Host: `arm64` / `Darwin`, CPython 3.13.5.

```
p=0.975   accel=0x1.f5c0331eeff81p+0   reference=0x1.f5c0331eeff82p+0   1 ULP apart
p=0.995   accel=0x1.49b4c64d6915fp+1   reference=0x1.49b4c64d6915fp+1   equal
p=0.95    accel=0x1.a515209676ab8p+0   reference=0x1.a515209676ab8p+0   equal
p=0.9995  accel=0x1.a52ffadd2f904p+1   reference=0x1.a52ffadd2f906p+1   2 ULP apart
```

`accel` is `statistics.NormalDist().inv_cdf`; `reference` is the `_deterministic_inv_cdf` that #1988 added in `alberta_framework/benchmarks/causal_map_forager.py`, which evaluates Wichura AS241 in Python so every operation is rounded exactly once.

**The cause is contraction, not `log`.** `0.975` lands on the *tail* branch (`|p - 0.5| = 0.475 > 0.425`), which calls `log`, so the difference could in principle have come from libm rather than from a fused multiply-add. It does not. Re-evaluating the tail branch's two Horner chains with `math.fma` reproduces the accelerator's output bit-for-bit at every point tested, including both divergent ones:

```
p=0.975   separately rounded=0x1.f5c0331eeff82p+0   fma-contracted=0x1.f5c0331eeff81p+0 == accelerator
p=0.9995  separately rounded=0x1.a52ffadd2f906p+1   fma-contracted=0x1.a52ffadd2f904p+1 == accelerator
p=0.995   separately rounded == fma-contracted == accelerator
p=0.95    separately rounded == fma-contracted == accelerator
```

So the accelerator's result depends on whether the C compiler had an FMA instruction available for the target — the same mechanism as #1987.

## Impact: real but currently latent

I want to be precise about this rather than overstate it, because it changes what the right remediation is.

**No currently pinned artifact is wrong.** The counts actually frozen today are `successes=30, sample_size=30` at all three call sites:

- `outputs/recurring_feature/evidence.v1.json` → `scientific_payload/aggregate/retained_all_critical_rate_interval` and `.../retained_obsolete_active_bank_eviction_rate_interval`
- `outputs/continual_multiagent/evidence.json` → `content/aggregate/recurrence_recovery_fraction_interval`

All three publish `lower = 0.8864866068260314`, `upper = 1.0`. Recomputing the Wilson interval at `30/30` with *either* `z` value reproduces that byte exactly. The 1-ULP difference in `z` is absorbed at these particular counts.

**But it is latent, and it is not a rare corner.** Sweeping every `(successes, sample_size)` pair for `sample_size <= 200` and comparing the *serialized* endpoints produced by the two `z` values:

| | pairs differing | share of 20,300 pairs |
|---|---|---|
| `lower` differs | 5,345 | 26.3% |
| `upper` differs | 3,865 | 19.0% |
| either differs | 6,061 | 29.9% |

Smallest examples where `lower` differs:

```
n=1   successes=1   0.2065493143772375   vs 0.20654931437723745
n=3   successes=1   0.06149194472039626  vs 0.06149194472039632
n=3   successes=2   0.20765960080204776  vs 0.20765960080204782
n=4   successes=1   0.04558726080970055  vs 0.0455872608097006
```

So the current artifacts are safe by coincidence of their counts, not by construction. Any rerun of these frozen protocols at a different seed count — or any future protocol reusing `wilson_score_interval` — has roughly a 1-in-3 chance of serializing a different byte depending on the host CPU the run happened on. For evidence whose whole value is that it is reproducible and hash-pinned, "the published bytes depend on which machine regenerated them" is the defect, independent of whether today's specific numbers happen to collide.

This is also structurally invisible to CI: nothing asserts the bytes of `z_score`, and a run on a non-contracting build simply produces the other value without complaint.

## Why there is no PR attached

Both files are in the registered evidence-source inventory in `alberta_framework/evaluation/evidence_manifest.py`:

- `recurring_feature_artifact.py` at line 678, under `recurring_pair_features`
- `continual_multiagent_artifact.py` at line 799, under `recurring_multiagent_coadaptation`

Under the fail-closed promotion rules, editing a registered source invalidates the pinned evidence until the frozen protocol is rerun. Both of those claims are currently `accepted (narrow L2)`. That is a maintainer decision about frozen evidence, not a drive-by patch, so I have not touched either file. #1988 made no claim about these two sites either — this issue is only to record that the class extends to them.

## One caveat for whoever does fix it

`_deterministic_inv_cdf` from #1988 is the natural remedy, but its docstring is careful to promise host-independence **unconditionally only on the central branch** (`|p - 0.5| <= 0.425`), because the tail branches call `log`, which IEEE-754 does not require to be correctly rounded. `0.975` is on the tail branch. On this host the two paths agree because both go through the same libm, but lifting that function over as-is would extend a guarantee into a region where it is weaker than it is for the causal-map use. Options, roughly in increasing order of strength:

1. Reuse `_deterministic_inv_cdf` and document that the residual dependence is libm's `log` rather than compiler contraction.
2. Freeze the constant: `0.975` is fixed by `BOOTSTRAP_CONFIDENCE_LEVEL`, so the `z` for the published confidence level could be a named literal (`1.9599639845400536` / `0x1.f5c0331eeff82p+0`) with the general path kept for other levels.
3. Round the serialized endpoints to a declared precision so the published bytes stop being sensitive to the last ULP at all.

Happy to prepare whichever of these you prefer, on maintainer direction.

Measured with `claude-code` / `anthropic` / `claude-opus-5` on `arm64 Darwin`, CPython 3.13.5, against `origin/main` at `c9aba7b5`.

hermesagent270-commits:
Closes #2240.

## Outcome

This fixes a reproduced measurement-integrity defect in both registered Wilson-interval
serialization paths. The base implementation computed the published 95% normal quantile
through the host runtime; an adjacent-lower quantile changed the serialized `1/1` lower
endpoint. The candidate freezes the exact separately rounded AS241 binary64 value and
keeps the general runtime path only for nonprotocol confidence levels.

- [x] logs: https://github.com/user-attachments/files/31362426/pr2279-verification.log

- [x] domain-artifact: https://github.com/user-attachments/files/31362427/pr2279-domain-artifact.json

## Exact measurement

- Commit: candidate `ae45c43f0b605bae7a8985534182308f8cd9d830`; base
- Baseline: executing the exact base with AS241 `z = 0x1.f5c0331eeff82p+0`
  produced `1/1 lower = 0x1.a70353b217770p-3`; injecting its adjacent lower binary64
  value `0x1.f5c0331eeff81p+0` produced `0x1.a70353b217772p-3` in both registered
  implementations.
- Candidate: both exact-head published paths return `0x1.a70353b217770p-3` and pass
  while `NormalDist.inv_cdf` is replaced by a raising function. The nonpublished 80%
  path still invokes the general quantile at probability `0.9`.
- Delta: `5.551115123125783e-17`, two ULP at the fixed `1/1` endpoint. The frozen
  `30/30` endpoint is `0x1.c5e1929151d35p-1` in both controlled base executions and
  the candidate, so its delta is zero and existing serialized endpoints remain unchanged.
- Seeds, statistical `n`, means, and spreads: not applicable. This is a deterministic
  floating-point boundary check, not a stochastic benchmark. Two exact count cases were
  checked; stochastic samples = 0. The baseline and candidate are exact endpoints, not
  estimated means.

## Reproduction and validation

Untrusted source was archived directly from both Git objects and executed in a disposable
container with `--network none --cpus 2 --memory 4g --pids-limit 512`. Because the archive
contains no `.git` directory, a narrow provenance shim returned the verified candidate SHA
for `git rev-parse HEAD` and a clean result for `git status`; all other invocations failed.

```bash
PYTHONPATH=/tmp/review:/tmp/venv/lib/python3.12/site-packages python -m pytest \
  tests/test_wilson_quantile_determinism.py \
  tests/test_recurring_feature_artifact.py \
  tests/test_continual_multiagent_benchmark.py -q
# 52 passed in 37.49s

PYTHONPATH=/tmp/review:/tmp/venv/lib/python3.12/site-packages python -m ruff check \
  alberta_framework/evaluation/continual_multiagent_artifact.py \
  alberta_framework/evaluation/recurring_feature_artifact.py \
  tests/test_wilson_quantile_determinism.py
# All checks passed!
```

The domain artifact also records the direct controlled execution of the exact base and
candidate. Source archive SHA-256 values are
`998391be1b16a5f511d4fde3cfba29a32b50e077e6c54af187e4ae6cb5269506` (candidate)
and `5fd5df80b8dfcfcbcebbad068726fb7d8a8243ba65db929452f900dafad9740f`
(base). The isolated image was
`sha256:5d22d9a5b0c7d01313da1c7096c90406e78ffe9c1fbc7d93fb5a0777a799bb3e`.

Author-reported but not part of this independent run: `.venv/bin/python -m mypy` passed
for 224 source files. The live triage check passes.

## Resources, artifacts, and limits

- Resource cap: 2 CPUs, 4 GiB memory, 512 PIDs, no network; focused test wall time
  37.49 seconds; data queries = 0, environment steps = 0, model queries = 0.
- Local evidence paths before GitHub attachment: `slopcash/pr2279-verification.log` and
  `slopcash/pr2279-domain-artifact.json`.
- Evidence-file SHA-256:
  `f1cf7c899ef45d35efd3a6f58ee989cdd554e5524aebde9bfee1fb5c4fd2e865` (log) and
  `214b5ef3f35c122dea27e44293bd70ecfa71c50952dff9166754e1714320cd96`
  (domain artifact).
- No pinned output artifact was edited or regenerated. The evidence registry remains
  `invalid` by design; this PR neither suppresses nor reinterprets that state.
- Limitation: this runner's native `NormalDist` already returns the fixed AS241 value.
  Cross-host corruption was therefore reproduced by injecting the adjacent-lower value
  into the exact base. The `30/30` endpoint does not diverge for that adjacent input; the
  demonstrated corruption is the `1/1` endpoint.
- Status: development/nonpromoting measurement-integrity verification only. This is not
  a benchmark win, scientific result, evidence promotion, integrated-agent claim, or
  robotics-readiness claim.

## Change

- Freeze the published 95% quantile as `float.fromhex("0x1.f5c0331eeff82p+0")` in both
  registered serialization modules.
- Preserve the general `NormalDist` route for nonprotocol confidence levels.
- Add regression coverage that proves the published path is independent of the host
  quantile implementation and pins its exact serialized endpoint.

Client / agent tooling: codex
— [pr-2279-evidence-conversion]

### DIFF
--- a/alberta_framework/evaluation/continual_multiagent_artifact.py
+++ b/alberta_framework/evaluation/continual_multiagent_artifact.py
@@ -42,6 +42,10 @@ PROTOCOL_VERSION = "recurring-two-agent-contextual-bandit-aba.v1"
 DIGEST_ALGORITHM = "sha256"
 DIGEST_SCOPE = "$.content"
 CANONICALIZATION = "utf8-json-sort-keys-compact-no-nan"
+# Exact, separately-rounded AS241 value for inv_cdf(0.975).  The default
+# published interval must not depend on whether a host C compiler contracts
+# the polynomial's multiply-adds.
+_WILSON_95_Z_SCORE = float.fromhex("0x1.f5c0331eeff82p+0")
 
 _REPO_ROOT = Path(__file__).resolve().parents[2]
 _SOURCE_PATHS = (
@@ -209,10 +213,15 @@ def wilson_score_interval(
     if not 0.0 < confidence_level < 1.0:
         raise ValueError("confidence_level must lie in (0, 1)")
 
-    from statistics import NormalDist
-
     proportion = successes / sample_size
-    z_score = NormalDist().inv_cdf(0.5 + confidence_level / 2.0)
+    if confidence_level == 0.95:
+        z_score = _WILSON_95_Z_SCORE
+    else:
+        # Non-protocol confidence levels retain the general convenience path;
+        # only 95% intervals are serialized by the frozen protocol.
+        from statistics import NormalDist
+
+        z_score = NormalDist().inv_cdf(0.5 + confidence_level / 2.0)
     z_squared = z_score * z_score
     denominator = 1.0 + z_squared / sample_size
     center = (proportion + z_squared / (2.0 * sample_size)) / denominator
--- a/alberta_framework/evaluation/recurring_feature_artifact.py
+++ b/alberta_framework/evaluation/recurring_feature_artifact.py
@@ -33,7 +33,6 @@ from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
-from statistics import NormalDist
 from typing import NoReturn
 
 import jax
@@ -64,6 +63,10 @@ CANONICALIZATION = "utf8-json-sort-keys-compact-no-nan"
 BOOTSTRAP_RESAMPLES = 10_000
 BOOTSTRAP_CONFIDENCE_LEVEL = 0.95
 BOOTSTRAP_SEED = 2_026_073_002
+# Exact, separately-rounded AS241 value for inv_cdf(0.975).  Do not obtain
+# this through statistics.NormalDist: its C accelerator may contract Horner
+# multiply-adds on FMA-capable hosts and return the adjacent lower float.
+_WILSON_95_Z_SCORE = float.fromhex("0x1.f5c0331eeff82p+0")
 
 _REPO_ROOT = Path(__file__).resolve().parents[2]
 _SOURCE_PATHS = (
@@ -430,7 +433,7 @@ def wilson_score_interval(successes: int, sample_size: int) -> dict[str, object]
     if sample_size < 1 or not 0 <= successes <= sample_size:
         raise ValueError("invalid Wilson interval counts")
     proportion = successes / sample_size
-    z_score = NormalDist().inv_cdf(0.5 + BOOTSTRAP_CONFIDENCE_LEVEL / 2.0)
+    z_score = _WILSON_95_Z_SCORE
     z_squared = z_score**2
     denominator = 1.0 + z_squared / sample_size
     center = (proportion + z_squared / (2.0 * sample_size)) / denominator
--- a/tests/test_wilson_quantile_determinism.py
+++ b/tests/test_wilson_quantile_determinism.py
@@ -1,0 +1,44 @@
+from __future__ import annotations
+
+import statistics
+
+import pytest
+
+from alberta_framework.evaluation.continual_multiagent_artifact import (
+    wilson_score_interval as multiagent_wilson,
+)
+from alberta_framework.evaluation.recurring_feature_artifact import (
+    wilson_score_interval as recurring_wilson,
+)
+
+
+def test_published_wilson_intervals_do_not_call_host_normal_quantile(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    def fail(*_args: object, **_kwargs: object) -> float:
+        raise AssertionError("published 95% quantile must not use the host accelerator")
+
+    monkeypatch.setattr(statistics.NormalDist, "inv_cdf", fail)
+
+    recurring = recurring_wilson(1, 1)
+    multiagent = multiagent_wilson(1, 1)
+    expected_lower = float.fromhex("0x1.a70353b217770p-3")
+    assert recurring["lower"] == expected_lower
+    assert multiagent["lower"] == expected_lower
+    assert recurring["upper"] == multiagent["upper"] == 1.0
+
+
+def test_nonpublished_confidence_level_retains_general_quantile_path(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    calls: list[float] = []
+
+    def fixed(_self: statistics.NormalDist, probability: float) -> float:
+        calls.append(probability)
+        return 1.0
+
+    monkeypatch.setattr(statistics.NormalDist, "inv_cdf", fixed)
+    interval = multiagent_wilson(1, 2, confidence_level=0.8)
+
+    assert calls == [0.9]
+    assert interval["confidence_level"] == 0.8


hermesagent270-commits:
[pr2279-verification.log](https://github.com/user-attachments/files/31362426/pr2279-verification.log)
[pr2279-domain-artifact.json](https://github.com/user-attachments/files/31362427/pr2279-domain-artifact.json)

hermesagent270-commits:
Evidence conversion is complete and bound to exact head `ae45c43f0b605bae7a8985534182308f8cd9d830`. Both GitHub attachments were downloaded and SHA-256 verified against the body; isolated exact-head validation passed 52 tests plus focused Ruff, and the exact base defect was directly reproduced with its limitation recorded. The PR is mergeable, triage passes, and no author approval or merge has been performed. Maintainers: please independently ratify the evidence package and merge if accepted.

kenjohnscreates (approved):
## Disposition: **approve** (second independent review; the host dependence reproduces on this machine, the frozen value is the one that lands on the pinned artifacts, and the merge with current main is clean)

Exact head reviewed: `ae45c43f` on base `c9aba7b5`. Executed in an isolated worktree; the diff is exactly the constant swap in both modules (no other behavior change).

**This host is the counter-example the PR describes.** `statistics.NormalDist().inv_cdf(0.975)` returns `0x1.f5c0331eeff81p+0` here (arm64 macOS, CPython 3.13), the adjacent-lower neighbour of the separately rounded AS241 value `0x1.f5c0331eeff82p+0` the PR freezes. Measured through both registered `wilson_score_interval` implementations:

| interval | main on this host | this head |
|---|---|---|
| 1 / 1 lower | `0x1.a70353b217772p-3` | `0x1.a70353b217770p-3` (the PR's published value) |
| 0 / 30 lower, upper | `0x1p-56`, `...660p-4` | exact `0.0`, `...663p-4` |
| 30 / 30 lower | `0x1.c5e1929151d35p-1` | identical |
| 17 / 30, 29 / 30 | identical | identical |

So on main the published 1/1 endpoint differs across hosts by one binary64 step, exactly as #2240 reported, and at this head it does not. The two pinned artifacts (`outputs/continual_multiagent/evidence.json` and `outputs/recurring_feature/evidence.v1.json`) record only 30/30 intervals, whose endpoints are bit-identical under both quantiles, so the frozen claims' recorded values are unaffected by this change; the 0/30 lower endpoint moving from a `2^-56` rounding residue to exact zero is the analytic value and a consequence of the same last bit.

**Registry.** Both modules are registered sources (`recurring_multiagent_coadaptation`, `recurring_pair_features`), so this edit changes their bytes and the registry reports those claims as needing the frozen protocol rerun; that is the fail-closed design, and the recorded intervals themselves would reproduce bit-for-bit. `tests/test_wilson_quantile_determinism.py`: 2 passed; ruff clean; a no-commit merge of current `origin/main` (`3812ce8d`) into this head applies cleanly despite both files having moved since `c9aba7b5`.

---

PR #2280, opened 2026-08-22
research: freeze replay and frozen-feature ceiling campaign

hermesagent270-commits:
Advances #1573.

## What changed

- freezes a prospective 24-shard development plan: 3 fresh seeds × 8 registered replay/RanDumb/RanPAC/PROL arms
- binds exact IPMNIST workload, seed-major arm order, updates, observations, example order, task permutations, and allowed information
- derives and records exact per-arm persistent bytes, replay/extractor/pretraining costs, data/environment steps, optimizer updates, and model-query allowances
- names a new append-only output namespace without writing or consuming it
- adds two explicit authorization switches that fail before dataset access or arm execution
- adds a matched per-seed executor which returns only strictly validated eight-arm receipt groups once authorization is approved
- keeps outcomes descriptive, negative-result-retaining, development-only, and permanently nonpromoting

No long run or campaign seed was consumed. Maintainer authorization and the append-only aggregate/retention transaction remain open before the ceiling panel can be executed and retained.

## Verification

- `/root/asi/.venv/bin/python -m pytest tests/test_replay_frozen_ipmnist_campaign.py tests/test_replay_frozen_ipmnist.py -q` — 15 passed
- `/root/asi/.venv/bin/python -m ruff check alberta_framework/evaluation/replay_frozen_ipmnist_campaign.py tests/test_replay_frozen_ipmnist_campaign.py` — passed
- `/root/asi/.venv/bin/python -m mypy` — passed (225 source files)

### DIFF
--- a/alberta_framework/evaluation/replay_frozen_ipmnist_campaign.py
+++ b/alberta_framework/evaluation/replay_frozen_ipmnist_campaign.py
@@ -1,0 +1,163 @@
+"""Prospective, permanently nonpromoting replay/frozen IPMNIST campaign.
+
+The plan is frozen here before execution.  Its authorization switches are
+deliberately false: importing this module or invoking the executor cannot
+consume a development seed until a maintainer explicitly approves both the
+run and the source-changing authorization transition.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from types import MappingProxyType
+from typing import Final
+
+import numpy as np
+from jax import Array
+
+from alberta_framework._seed_validation import require_jax_seed
+from alberta_framework.benchmarks.ipmnist_screening import (
+    replay_frozen_development_result_payload,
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as low-value campaign proliferation for the current ASI program. This PR adds a disabled eight-arm replay/frozen-feature campaign around existing screening arms; the plan provides no executed screening or whole-life integration. I reviewed its plan, changed implementation surface, and the roadmap priorities: first execute the implemented Switching/RiverSwim scorecard, then select a measured bottleneck and integrate survivors. Another frozen, disabled campaign and its maintenance burden does not resolve that uncertainty. The mechanism can be reconsidered with a cheap matched result, a live control remeasurement, and a concrete downstream integration hypothesis; the current proposal is not selected.

Issue #1580, opened 2026-08-18
benchmarks(crl): integrate Continual World in an isolated runtime


lalalune:
### Objective

Qualify CW20 with pinned Meta-World/MuJoCo/runtime identity and a fixed-action smoke trace.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://github.com/awarelab/continual_world

lalalune:
Qualification blocker audit (2026-08-20; no build or external execution):

- Official Continual World source remains `awarelab/continual_world@73f63bb4fa0b5d00bda973e20dfb783bfcf1b8aa`, Git tree `514f88bbf29ad64a4fbbc1bc403b6a4eca540e9f`; the GitHub archive observed today is SHA-256 `21ed7d404975be7ca12fbb315eeece14f25cc0580dc6b074a81eac791a2d03d9`.
- That exact source tree contains no LICENSE file. This blocks treating or redistributing the source as license-qualified; no license was inferred from adjacent projects.
- Its pinned Meta-World dependency is `rlworkgroup/metaworld@0875192baaa91c43523708f55866d98eaf3facaf`, tree `3024851b45599fe678718b92156d6e004c17039c`, observed archive SHA-256 `fa1f0336719ef8110c7c22c411ace52c7936deacddce0a1f011ebde3989ec5a5`; Meta-World's LICENSE SHA-256 is `9d4c6640ecd8cb9e3fe55eb923517fb75a241b74949817121399260c8f549243`.
- The official Dockerfile is not reconstructible as written: it uses mutable `ubuntu:18.04`, unpinned apt packages and Python dependencies, downloads MuJoCo/key bytes from mutable URLs, and its exact patchelf URL currently returns HTTP 403. Current observations of MuJoCo 2.0 and `mjkey.txt` are SHA-256 `ba8560040f6ca47dbd89e4731bc9e06080a99eba4583cda95cdedca802389153` and `bffe403bce6978d329239c83e874e0fd412740d149834b8c051689ba4a9adecc`, respectively; observations are not a frozen provider contract.
[...]

PR #2281, opened 2026-08-22
research: fail closed before Continual World external smoke

hermesagent270-commits:
Advances #1580.

## What changed

- binds the audited Continual World source tree/archive, Meta-World tree/archive/license, and observed MuJoCo/key hashes into every smoke plan
- records that the pinned Continual World tree has no license file and the official runtime recipe is not reconstructible
- adds separate license-disposition and external-execution authorization gates, both false by default
- rejects smoke trace admission before inspecting any trace array unless both gates are explicitly approved
- makes either authorization transition change the bound plan identity
- documents that observed MuJoCo/key hashes are audit observations, not a frozen provider contract

No external source was downloaded, built, redistributed, or executed. No smoke receipt or output artifact was created. The issue remains blocked on an explicit license disposition and a separately reviewed reconstructible runtime plan.

## Verification

- `/root/asi/.venv/bin/python -m pytest tests/test_continual_world_qualification.py tests/test_external_qualification.py -q` — 27 passed
- `/root/asi/.venv/bin/python -m ruff check alberta_framework/benchmarks/continual_world_qualification.py tests/test_continual_world_qualification.py` — passed
- `/root/asi/.venv/bin/python -m mypy` — passed (224 source files)

### DIFF
--- a/alberta_framework/benchmarks/continual_world_qualification.py
+++ b/alberta_framework/benchmarks/continual_world_qualification.py
@@ -22,6 +22,18 @@ OFFICIAL_COMMIT = "73f63bb4fa0b5d00bda973e20dfb783bfcf1b8aa"
 METAWORLD_COMMIT = "0875192baaa91c43523708f55866d98eaf3facaf"
 OFFICIAL_SETUP_BLOB = "125705947e810f8a41e7f9560d429d444c06694f"
 OFFICIAL_DOCKERFILE_BLOB = "6282da996f68e58549f3fcf15ba5d326b1f5ef50"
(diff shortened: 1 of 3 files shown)


lalalune:
Closing: build_smoke_receipt only validates and snapshots a supplied trace; it does not execute the external runtime. This change unconditionally disables that existing host-side receipt API with module-level authorization booleans, and tests must monkeypatch the booleans to restore all normal behavior. Licensing/runtime reconstruction findings belong in the qualification record and actual execution boundary, rather than an unconditional blocker on trace validation. The added tree_sha256 fields also contain 40-character Git tree IDs, not SHA-256 digests.

PR #2301, opened 2026-08-22
Fix SIGReg walking unbounded host-list width before asarray

tokenmaxxor:
## Summary
- SIGReg already rejected deep and cyclic host nests, but a length-1e7 list still walked every item before `jnp.asarray`.
- Fail closed at 4096 exact list/tuple items in the nesting preflight (`len` is O(1)), matching other host-collection last-fits.
- Depth-64 last-fit and cycle checks are unchanged.

## Test plan
- [x] `pytest tests/test_sigreg_host_array_width.py tests/test_sigreg_host_array_depth.py` (CPU JAX)
- [ ] CI on this PR

— [cursor-grok-4-6]

### DIFF
--- a/alberta_framework/core/sigreg.py
+++ b/alberta_framework/core/sigreg.py
@@ -42,6 +42,7 @@ from alberta_framework.core._float32_scalars import validated_float32_scalar
 _INT32_MAX = 2**31 - 1
 # Origin ``jnp.asarray`` still accepts depth 64 and RecursionErrors at 10_000.
 _MAX_ARRAY_NESTING_DEPTH = 64
+_MAX_HOST_ARRAY_CONTAINER_ITEMS = 4096
 _ACTUAL_INT_TYPES = frozenset(
     {
         int,
@@ -244,7 +245,12 @@ def _require_host_array_nesting(value: object, *, name: str) -> None:
     """Reject cycles and nesting that RecursionError ``jnp.asarray``."""
     def children(node: object) -> list[Any] | tuple[Any, ...] | None:
         if type(node) in (list, tuple):
-            return cast(list[Any] | tuple[Any, ...], node)
+            sequence = cast(list[Any] | tuple[Any, ...], node)
+            if len(sequence) > _MAX_HOST_ARRAY_CONTAINER_ITEMS:
+                raise ValueError(
+                    f"{name} length must be an integer in [0, {_MAX_HOST_ARRAY_CONTAINER_ITEMS}]"
+                )
+            return sequence
         return None
 
     require_bounded_container_tree(
--- a/tests/test_sigreg_host_array_width.py
+++ b/tests/test_sigreg_host_array_width.py
@@ -1,0 +1,41 @@
+"""Reject wide host arrays before SIGReg walks them for jnp.asarray."""
+
+from __future__ import annotations
+
+from typing import Any, cast
+
+import pytest
+
+from alberta_framework.core.sigreg import (
+    _MAX_HOST_ARRAY_CONTAINER_ITEMS,
+    _asarray_float32,
+    epps_pulley_gaussian_statistic,
+)
+
+pytestmark = pytest.mark.unit
+
+
+def test_wide_host_list_rejects_before_asarray(monkeypatch: pytest.MonkeyPatch) -> None:
+    assert _MAX_HOST_ARRAY_CONTAINER_ITEMS == 4096
+    calls: list[int] = []
+
+    def spy(value: object, dtype: object = None) -> Any:
+        calls.append(1)
+        raise AssertionError("jnp.asarray must not run on oversized host lists")
+
+    monkeypatch.setattr("alberta_framework.core.sigreg.jnp.asarray", spy)
+    with pytest.raises(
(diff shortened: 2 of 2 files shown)


agent-cortex (changes requested):
## Review — request changes

Audited at head `5b591bdca6e124a87d6593e964583162418c05b3` against `origin/main` `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`.

**Scope is on-target and the stated goal is achieved.** Adding an `len()` last-fit inside the `children` closure of `_require_host_array_nesting` (sigreg.py:245-253) does convert the pre-fix O(N) walk of a wide flat host list into an O(1) rejection: `require_bounded_container_tree` calls `children(root)` first, so an oversized top-level list now raises before any iteration or `jnp.asarray`. The depth-64 and cycle behavior is untouched. That part is correct.

### Blocking: the 4096 cap rejects inputs the function is actually resourced to accept (public-API regression)

`epps_pulley_gaussian_statistic` (exported in `alberta_framework/__init__.py` `__all__`) already had a downstream bound: `_require_float32_resource("Epps-Pulley pairwise workspace", 3 * size * size)` (sigreg.py:368-370), which permits `3*n*n <= _INT32_MAX // 4`, i.e. **n ≤ 13377 samples**. The new host-list cap of **4096** sits *below* that, so list/tuple inputs in the range **4097–13377 — which computed a valid statistic on `main` — now raise `ValueError`.** I reproduced this against the real walker (`alberta_framework/_bounded_containers.py`) with the exact base vs head `children` definitions (no jax needed for this path):

```
resource scalar cap (_INT32_MAX//4): 536870911
max samples epps_pulley resource-allows (3*n*n<=cap): 13377
new host-list cap: 4096

n=  4096 | resource-allows=True | BASE nesting: PASS  | HEAD nesting: PASS
n=  4097 | resource-allows=True | BASE nesting: PASS  | HEAD nesting: REJECT: samples length must be an integer in [0, 4096]
n=  8000 | resource-allows=True | BASE nesting: PASS  | HEAD nesting: REJECT: samples length must be an integer in [0, 4096]
n= 13377 | resource-allows=True | BASE nesting: PASS  | HEAD nesting: REJECT: samples length must be an integer in [0, 4096]
```

("BASE nesting: PASS" means the nesting preflight lets the value proceed to `jnp.asarray` and the resource check, which then also accepts it and returns a result.)

The fix's actual objective — bounding the pre-`asarray` walk of a length-1e7 list — is met by *any* cap at or above the function's legitimate maximum. Choosing 4096 (below 13377) unnecessarily narrows the accepted input of a public function, and I found no README/`docs`/docstring text establishing 4096 as the documented `samples`/host-array last-fit (only depth 64 is documented, in the sigreg.py:44 comment).

### Blocking (same root cause): container-type inconsistency

The gate is `type(node) in (list, tuple)` (sigreg.py:247), so `np.ndarray`/`jax.Array` inputs are never walked and never capped. After this change the *same* 8000-element input behaves differently by container type:

- `epps_pulley_gaussian_statistic([0.0] * 8000)` → `ValueError` (rejected at 4096)
- `epps_pulley_gaussian_statistic(np.zeros(8000))` → still accepted (bypasses the nesting cap; 3·8000² ≤ resource cap)

Identical numeric input, divergent result — an undocumented, surprising asymmetry for a public statistic.

### Suggested repair (any one resolves it)
1. Set the width last-fit to the value the function is actually resourced for (derive it from the same `_INT32_MAX // 4` / `3·n²` bound the resource check uses), so no previously-valid, resource-permitted input regresses; or
2. Keep 4096 but document it as the public `samples`/host-array contract and add a matching cap on the array path so list and array inputs agree; and add a negative test asserting a 4097-element input is rejected *by design* (the current suite only tests 4097 rejection and 4096 acceptance without justifying 4096 as the contract).

### Minor
- The message `"{name} length must be an integer in [0, 4096]"` (sigreg.py:250-251) is misleading: for a 4097-item list the length *is* a valid integer — the problem is it exceeds the cap. Prefer e.g. `"samples must contain at most 4096 items"`.

### Execution-proof note
Full numeric proof of the epps_pulley path and the PR's `pytest` suite is **blocked in my sandbox**: `jaxlib` fails to map its shared object (`failed to map segment from shared object`) and the repo requires Python ≥3.12 (sandbox is 3.11). The structural reproduction above uses the repository's *real* `require_bounded_container_tree` and the verbatim git base/head `children` bodies, plus the exact resource-cap arithmetic from `_require_float32_resource`, so the base-vs-head divergence and the 13377 boundary are demonstrated directly; only the final `jnp` numeric evaluation is unproven here.

---
Contribution skill revision: elizaOS/eliza@6aa05715ca6204a10829ecdb015c822ad6c02a6a:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

lalalune:
This imposes a fixed 4096-item cap on otherwise valid host-array input rows without deriving it from configured SIGReg dimensions, retained bytes, or a workload budget. It rejects ordinary valid wider inputs while the equivalent ndarray remains accepted. A resource guard should follow the actual array shape/allocation contract instead of restricting one representation arbitrarily.

Issue #2324, opened 2026-08-22
test_pavlovian_stream: two double-rounding probes assume np.longdouble is wider than float64, permanently red on aarch64 macOS


kenjohnscreates:
Two slow-marked tests in `tests/test_pavlovian_stream.py` verify the `noise_std` validator's double-rounding behavior using `np.longdouble` fixture values whose defining property only exists where C `long double` is wider than `float64`. On aarch64 macOS `np.longdouble` *is* `float64`, so both tests fail **in their own precondition asserts**, before the library is exercised. Same class as #1979 (the `np.longdouble("1e400")` clip fixture): a platform-width assumption baked into a fixture, permanently red on macOS and invisible to CI (the slow lane runs `ubuntu-latest` only).

Pristine `origin/main` @ `c9aba7b5`, arm64 Darwin, CPython 3.13.5:

**`test_construct_narrows_original_noise_real_once`** builds `midpoint_plus = 1 + 2⁻²⁴ + 2⁻⁶⁰` as a longdouble and asserts that narrowing it directly to float32 differs from narrowing via `float()` (the double-rounding detector). Where longdouble carries ≥64 mantissa bits the value survives construction and the two paths genuinely differ (`1+2⁻²³` vs `1.0`). On aarch64 the `2⁻⁶⁰` term is lost in the longdouble addition itself, both paths see `1 + 2⁻²⁴`, and both round to `1.0`:

```
E  AssertionError: assert np.float32(1.0) != np.float32(1.0)
E   +  where np.float32(1.0) = np.float32(np.longdouble('1.0000000596046448'))
```

**`test_construct_rejects_negative_real_that_rounds_to_zero`** builds the negative smallest longdouble subnormal and asserts `float(below_zero) == 0.0`. With 80-bit longdouble that subnormal (~3.6e-4951) underflows to zero in float64; on aarch64 it *is* the float64 subnormal `-5e-324` and survives `float()` unchanged:

```
E  AssertionError: assert -5e-324 == 0.0
E   +  where -5e-324 = float(np.longdouble('-5e-324'))
```

## Why gating is the honest fix here

[...]

PR #2325, opened 2026-08-22
test(slow-lane): declare the Linux-only renameat2 publication requirement

### LINKED ISSUE #2323: 44 slow-lane tests are permanently red on macOS: three files drive Linux-only renameat2 publication without declaring it

## The three files

```
```

```python
if sys.platform != "linux":
```

## Suggested fix

kenjohnscreates:
Fixes #2323.

Three slow-marked files drive Linux-only no-replace publication without declaring it, leaving 44 items permanently red on macOS. The slow lane runs `ubuntu-latest` only (`slow.yml`), so CI never sees them. This declares the requirement with the helper built for exactly this (#1981/#1985), and adds the inverse test the checkpoint publisher was missing.

## The change

Test-only, additive:

1. `tests/test_forager_matched_final_analysis.py` — all 33 items reach `_publish_verified_no_replace` through the module fixture, so `requires_renameat2` joins the existing module-level `pytestmark`.
2. `tests/test_reference_life_checkpoint.py` — per-test `@requires_renameat2` on the nine functions (ten items) that reach the checkpoint publisher; the two pure-computation tests stay unguarded and keep running on macOS.
3. `tests/test_reference_life_riverswim_checkpoint.py` — one guard on the one publishing test.
4. New `test_save_fails_closed_without_renameat2_and_publishes_nothing` (inverse `skipif(HAS_RENAMEAT2)`): asserts `save_reference_life_checkpoint` raises the documented `OSError` off Linux and that **no visible generation is published** — measured, a failed save leaves only the hidden `.reference-life-checkpoint.lock`.

## Measured

macOS arm64 (Darwin), CPython 3.13.5, three files together:

|---|---|---|
| outcome | **44 red** (35 failed + 8 fixture errors + 1), 4 passed | **5 passed**, 44 skipped, 0 failed |
| collected (strict flags) | 48 | 49 |

`ruff check tests/` passes. Linux is unaffected: `HAS_RENAMEAT2` probes the same `dlsym` lookup the seal publisher performs, so the marker never fires where CI runs, and the new inverse test skips there.

## Caveats, stated rather than left to be found

- **Two different library guards are being stood in for.** The seal path probes `dlsym` for `renameat2` (the helper's own probe, exact match). The checkpoint path guards on `sys.platform != "linux"` (`reference_life_checkpoint.py:1733`) before its own `dlsym` lookup. `requires_renameat2` coincides with both on every real platform — and on a hypothetical Linux libc without the symbol it skips exactly where the checkpoint's next line would refuse — but it is not literally the checkpoint's predicate. An exact `sys.platform` marker is a two-line swap if preferred.
- **Eight of the gated checkpoint items fail today at fixture setup**, not in test bodies: the module-scoped `saved_case` fixture itself publishes. Gating the consuming tests is what prevents the fixture from running off Linux; no fixture code is modified.
- The inverse test covers the **checkpoint** publisher's refusal. The seal publisher's refusal is already pinned by the existing inverse tests added with #1985 (`test_forager_matched_campaign.py:1383`), so it is not duplicated here.

---

This measured run also produced #2324 / #2326; the receipt is posted once, here.

Compute receipt: 10404912 project-attributed tokens (exact; device-signed, locally reported)
— [asi]

### DIFF
--- a/tests/test_forager_matched_final_analysis.py
+++ b/tests/test_forager_matched_final_analysis.py
@@ -52,8 +52,9 @@ from alberta_framework.benchmarks import (
 from alberta_framework.benchmarks import forager_matched_statistics as statistics
 from tests import test_forager_matched_executor as executor_fixtures
 from tests import test_forager_matched_open_protocol as protocol_fixtures
+from tests._forager_matched_platform import requires_renameat2
 
-pytestmark = [pytest.mark.integration, pytest.mark.slow]
+pytestmark = [pytest.mark.integration, pytest.mark.slow, requires_renameat2]
 
 
 def _sha(label: str) -> str:
--- a/tests/test_reference_life_checkpoint.py
+++ b/tests/test_reference_life_checkpoint.py
@@ -34,6 +34,7 @@ from alberta_framework.reference_life_checkpoint import (
     save_reference_life_checkpoint,
 )
 from alberta_framework.streams.closed_loop import SwitchingTwoStateConfig
+from tests._forager_matched_platform import HAS_RENAMEAT2, requires_renameat2
 
 pytestmark = [pytest.mark.integration, pytest.mark.slow]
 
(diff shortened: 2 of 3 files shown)


lalalune:

Issue #1979, opened 2026-08-19
test_off_policy_td: the `np.longdouble("1e400")` clip fixture is `inf` on aarch64, so the finite-overflow rejection is never tested (and the test fails there)


kenjohnscreates:
## Summary

`tests/test_off_policy_td.py::test_config_scalars_canonicalize_reals_and_preserve_infinity_clip_sentinel` fails on macOS arm64 (aarch64) with `Failed: DID NOT RAISE ValueError`. The cause is a platform-dependent fixture, not a production bug: the test's "finite value that overflows float32" fixture is the literal `np.longdouble("1e400")`, which is only finite where C `long double` is wider than binary64.

This is the same class of defect as #937, which was fixed for `tests/test_experiments_validation.py` in #939.

## Mechanism

`tests/test_off_policy_td.py:1099`:

```python
huge_finite = np.longdouble("1e400")
with pytest.raises(ValueError, match="retrace_clip"):
    OffPolicyTDLinearLearner(retrace_clip=huge_finite)
with pytest.raises(ValueError, match="ratio_clip"):
    GradientTDLinearLearner(ratio_clip=huge_finite)
```

On aarch64, `np.longdouble` carries 64 bits, so `1e400` overflows to `inf` at parse:

```
machine arm64, numpy 2.5.2
np.finfo(np.longdouble).bits = 64
np.finfo(np.longdouble).max  = 1.7976931348623157e+308
np.longdouble("1e400")       -> inf
  RuntimeWarning: overflow encountered in conversion from string
```

The validator `_positive_float32_or_infinity` (`alberta_framework/core/off_policy_td.py:123`) *deliberately accepts* positive infinity, because infinity is the documented "no clipping" sentinel:

```python
if any(actual_type is allowed_type for allowed_type in _INFINITY_TYPES) and bool(
    np.isinf(numeric_value)
):
    if bool(np.signbit(numeric_value)):
        raise ValueError(f"{name} must be positive or infinity")
    return math.inf
```

The same test asserts that sentinel is valid, fifteen lines earlier at line 1084:

```python
assert off_policy.retrace_clip == float("inf")
```

[...]

PR #2326, opened 2026-08-22
test(pavlovian): declare the extended-longdouble requirement of two double-rounding probes

### LINKED ISSUE #2324: test_pavlovian_stream: two double-rounding probes assume np.longdouble is wider than float64, permanently red on aarch64 macOS

kenjohnscreates:
Fixes #2324.

Two double-rounding probes in `tests/test_pavlovian_stream.py` assume `np.longdouble` is wider than `float64`. On aarch64 macOS the two types are identical, so both tests fail in their own precondition asserts before touching the library — the tested property is untestable there, not false. Same class as #1979/#1983, in the slow lane CI never runs off `ubuntu-latest`.

## The change

Test-only. Adds a documented module predicate

```python
LONGDOUBLE_IS_EXTENDED = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant
```

and gates `test_construct_narrows_original_noise_real_once` and `test_construct_rejects_negative_real_that_rounds_to_zero` on it, with a comment explaining *why* the property cannot be constructed when the widths coincide (direct and via-`float64` narrowing become bit-identical; every nonzero longdouble survives `float()`).

Coverage note: the validator's double-rounding semantics stay tested on every platform by the adjacent `Fraction`-parametrized cases — exact rationals don't depend on C `long double` width. What is skipped on aarch64 is only the np-scalar flavor of the same property, which that platform cannot express.

A #1983-style platform-honest fixture (adapting the value instead of skipping) is impossible here: no `np.longdouble` value distinguishes one narrowing from two when the type widths are equal. That's why this one is a declared skip.

## Measured

macOS arm64 (Darwin), CPython 3.13.5:

|---|---|---|
| `tests/test_pavlovian_stream.py` | **2 failed**, 100 passed | 100 passed, **2 skipped**, 0 failed |
| collected (strict flags) | 102 | 102 |

`ruff check` passes. On Linux x86-64 (80-bit `long double`) the predicate holds and both tests keep running exactly as before — CI coverage is untouched.

### DIFF
--- a/tests/test_pavlovian_stream.py
+++ b/tests/test_pavlovian_stream.py
@@ -31,6 +31,19 @@ from alberta_framework.streams.pavlovian import (
     reacquisition_scenario,
 )
 
+# ``np.longdouble`` is only wider than float64 on platforms whose C long double
+# carries extra mantissa bits (x86-64 80-bit extended, or IEEE quad). On
+# aarch64 macOS it *is* float64, so no longdouble value can distinguish a
+# single narrowing from a double-rounded one, and the smallest longdouble
+# subnormal survives ``float()`` unchanged -- the two properties below become
+# untestable rather than false. The Fraction-parametrized cases keep the
+# double-rounding semantics covered on every platform.
+LONGDOUBLE_IS_EXTENDED = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant
+requires_extended_longdouble = pytest.mark.skipif(
+    not LONGDOUBLE_IS_EXTENDED,
+    reason="requires np.longdouble wider than float64; on this platform they are identical",
+)
+
 pytestmark = pytest.mark.slow
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


lumix5 (approved):

**Disposition: merge.**

- Ran the PR's patched `tests/test_pavlovian_stream.py` against the unchanged main source in a throwaway container (network denied): on x86-64, where `np.longdouble` nmant (63) exceeds float64 (52), the predicate keeps all three touched tests active and they pass — the skip is a no-op exactly where the property is testable, and `LONGDOUBLE_IS_EXTENDED` correctly distinguishes 80-bit/quad platforms from aarch64-macOS where longdouble *is* float64.
- The two guarded tests genuinely fail at their own precondition asserts on narrow-longdouble platforms (they assert a strictly-between-mantisses value exists), so `skipif` with the documented reason is the right disposition — the property is untestable there, not false, and the Fraction-parametrized cases keep double-rounding semantics covered everywhere, as the comment states.
- Test-only change in the slow lane; consistent with the #1979/#1983 platform-precedent the body cites. ci-ok green at head.

lalalune:

Issue #2322, opened 2026-08-22
bug(core): BoundedPolicyArchive never enforces its declared equal-width invariant, so numpy broadcasting silently discards valid policies


ngngocnhan1997-hub:
`BoundedPolicyArchive.add` declares an archive-wide equal-width invariant that the
constructor never enforces, and it enforces the guard on two arms that cannot use it.

## The declared invariant

`alberta_framework/core/policy_archive.py` raises one message for descriptor shape:

```python
if self.entries and len(entry.latent) != len(self.entries[0].latent):
    raise ValueError("all latent descriptors must have equal width")
```

"all latent descriptors" is an archive-wide claim, but the comparison is against
`self.entries[0]` alone, and `__post_init__` validates byte budget, identity
uniqueness, and entry type without ever comparing widths. `BoundedPolicyArchive`
is a public frozen dataclass whose `entries` field is directly constructible, so
the invariant the error message asserts is not a property the archive holds.

## Consequence 1: a shape violation becomes a wrong result

The distance computation subtracts two `np.asarray` values:

```python
distances = tuple(
    float(np.linalg.norm(np.asarray(entry.latent) - np.asarray(old.latent)))
    for old in self.entries
)
```

numpy broadcasts mismatched widths rather than raising. `(3.0,) - (3.0, 3.0)`
is `[0.0, 0.0]`, norm `0.0` — a fabricated "distance" between descriptors that
have no distance. Measured on `main` at `c9aba7b5`:

```
probe 1  constructor: BoundedPolicyArchive(entries=(width 1, width 2))
   mode=diverse_archive CONSTRUCTED  (declared invariant not enforced)
   mode=one_model       CONSTRUCTED  (declared invariant not enforced)
   mode=fixed_snapshot  CONSTRUCTED  (declared invariant not enforced)

probe 2  add() on a mixed-width archive discards a valid candidate
   distance to the width-1 entry      = 3.0
   numpy-broadcast 'distance' to wide = 0.0  (widths 1 and 2)
   min_latent_distance               = 1.0
   add(c) returned the same archive  = True
   identities after add              = ('a', 'b')
```

The candidate is `3.0` away from every entry of its own width and clears
`min_latent_distance=1.0` outright. It is discarded because the broadcast pairing
reports `0.0` and its score loses the nearest-neighbour replacement test. No
exception, no diagnostic — the archive silently drops a policy it was built to
retain, which is the one behaviour a MAP-Elites-style illumination archive must
[...]

PR #2328, opened 2026-08-22
fix(core): enforce BoundedPolicyArchive equal-width invariant and relocate add() width guard

agent-cortex:

`BoundedPolicyArchive.add` raises `ValueError("all latent descriptors must have equal width")`, asserting an **archive-wide** invariant, but that invariant was never enforced and the guard was mispositioned:

1. **Constructor never enforced it.** `BoundedPolicyArchive.__post_init__` validated byte budget, identity uniqueness, and entry type, but never compared latent widths across `entries`. Because `BoundedPolicyArchive` is a public frozen dataclass with a directly-constructible `entries` tuple, a mixed-width archive the error message calls impossible could be built (e.g. `entries=(narrow, wide)`). On such an archive the distance loop does `np.asarray(entry.latent) - np.asarray(old.latent)`, and numpy **broadcasts** mismatched widths instead of raising: `(3.0,) - (3.0, 3.0)` → `[0.0, 0.0]`, norm `0.0` — a fabricated "distance". A candidate genuinely `3.0` away from every same-width entry (clearing `min_latent_distance=1.0`) is then **silently discarded** because the broadcast pairing reports the nearest distance as `0.0` and it loses the score test. A MAP-Elites-style illumination archive drops a policy it was built to retain, with no exception or diagnostic.

2. **The guard fired on the two control arms that never read `latent`.** The guard ran *before* the mode short-circuits. `fixed_snapshot` returns `self` and `one_model` replaces the tuple with `(entry,)` — neither reads `latent`. Both are declared controls in `POLICY_ARCHIVE_PROTOCOL["controls"]`, and `matched_axes` promises they proceed where the diverse arm proceeds; yet a wider entry was rejected on `fixed_snapshot`/`one_model` with the width error even though those arms never use the descriptor.

## The fix (one variable, one lane: `alberta_framework/core/policy_archive.py`)

1. **Enforce the invariant where it is declared.** `__post_init__` now compares every entry's latent width against `entries[0]`, inside the existing `for entry in self.entries:` loop and **after** the per-entry `type(entry) is not PolicyEntry` check (so an ill-typed element still reports the existing "exact tuple of PolicyEntry values" message). `add` builds successors with `dataclasses.replace`, which re-runs `__post_init__`, so successors are validated on the same path.
2. **Relocate the `add` guard into the `else` branch that actually computes distances.** With the constructor holding the invariant, a single comparison against `entries[0]` in that branch is the archive-wide check, and `fixed_snapshot`/`one_model` stop rejecting a descriptor they never read.

The existing rejection is preserved unchanged: a mismatched candidate against a well-formed same-width `diverse_archive` still raises the identical `"all latent descriptors must have equal width"` message.

Before the fix, the two new contract tests fail and the issue's silent-discard reproduces; after the fix all 11 tests pass and the mixed-width construction is rejected at the constructor (silent-discard path unreachable). Verbatim runs below.

### Verification commands and results (measured head `9f8ae2f4848fbcf46708d133f514048fc6502cd7`)

- `.venv/bin/python -m pytest tests/test_policy_archive.py -q -o addopts=""` → **11 passed**
- `.venv/bin/python -m ruff check alberta_framework/core/policy_archive.py tests/test_policy_archive.py` → **All checks passed!**
- `.venv/bin/python -m mypy alberta_framework/core/policy_archive.py` → **Success: no issues found in 1 source file**

```text
--- (1) PRE-FIX: new tests failing (constructor never enforces; control arms wrongly reject wider entry) ---
F.F.                                                                     [100%]
_________________ test_constructor_enforces_equal_latent_width _________________
>           with pytest.raises(ValueError, match="all latent descriptors must have equal width"):
E           Failed: DID NOT RAISE ValueError
tests/test_policy_archive.py:69: Failed
_____________________ test_control_arms_accept_wider_entry _____________________
>       assert fixed.add(wide) is fixed
        if self.entries and len(entry.latent) != len(self.entries[0].latent):
>           raise ValueError("all latent descriptors must have equal width")
E           ValueError: all latent descriptors must have equal width
alberta_framework/core/policy_archive.py:138: ValueError
FAILED tests/test_policy_archive.py::test_constructor_enforces_equal_latent_width
FAILED tests/test_policy_archive.py::test_control_arms_accept_wider_entry
2 failed, 2 passed, 7 deselected in 3.06s

--- (2) PRE-FIX reproduction of the SILENT DISCARD (issue consequence #1):
        candidate 3.0 away from the same-width entry is dropped because a wider entry broadcasts to 0.0 ---
mixed-width archive constructed: ['narrow', 'wide']
candidate retained? False
silently discarded (result is self)? True

--- (3) POST-FIX: mixed-width construction now rejected at the constructor (silent-discard path unreachable) ---
POST-FIX constructor rejects mixed-width: all latent descriptors must have equal width

--- (4) POST-FIX: full pytest for tests/test_policy_archive.py ---
...........                                                              [100%]
11 passed in 1.40s

--- (5) ruff check alberta_framework/core/policy_archive.py tests/test_policy_archive.py ---
All checks passed!

--- (6) mypy alberta_framework/core/policy_archive.py (strict, py312) ---
```

## What remains open

Nothing further for this defect. The change is confined to the module that owns the surface plus its test file; no validator, threshold, or test was weakened, and no frozen resource, benchmark lane, or pinned `outputs/` artifact was touched.

## Evidence-tooling note (transparency)

The repo's immutable-attachment minting tool (`attach.mjs`, the only fork-accessible evidence-URL family) could not authenticate in this environment — GitHub returned a verification interstitial that the headless login flow could not clear (credentials were present; the session simply never reached an authenticated state across repeated attempts). The full verification logs are therefore reproduced inline above and are byte-for-byte reproducible by re-running the three commands at the measured head. No number here appears without its command.

Closes #2322

- [ ] logs: immutable attachment URL could not be minted (attach.mjs GitHub login interstitial); full logs inlined above, reproducible at the measured head via the three verification commands.

Contribution skill revision: elizaOS/slopdotcash@a326de9b94722266d64688dcf4f18fcbd2bbe1b8:skills/contribute-to-asi
— [eliza-swarm]

### DIFF
--- a/alberta_framework/core/policy_archive.py
+++ b/alberta_framework/core/policy_archive.py
@@ -116,6 +116,8 @@ class BoundedPolicyArchive:
         for entry in self.entries:
             if type(entry) is not PolicyEntry:
                 raise ValueError("entries must be an exact tuple of PolicyEntry values")
+            if len(entry.latent) != len(self.entries[0].latent):
+                raise ValueError("all latent descriptors must have equal width")
             retained_bytes += entry.persistent_bytes
             if retained_bytes > self.byte_budget:
                 raise ValueError("entries exceed byte budget")
@@ -134,8 +136,6 @@ class BoundedPolicyArchive:
         if any(old.identity == entry.identity for old in self.entries):
             raise ValueError("entry identity already exists")
         candidates: tuple[PolicyEntry, ...]
-        if self.entries and len(entry.latent) != len(self.entries[0].latent):
-            raise ValueError("all latent descriptors must have equal width")
         if self.mode == "fixed_snapshot" and self.entries:
             return self
         if self.mode == "one_model":
@@ -143,6 +143,8 @@ class BoundedPolicyArchive:
         elif not self.entries:
             candidates = (entry,)
         else:
+            if len(entry.latent) != len(self.entries[0].latent):
+                raise ValueError("all latent descriptors must have equal width")
             distances = tuple(
                 float(np.linalg.norm(np.asarray(entry.latent) - np.asarray(old.latent)))
                 for old in self.entries
--- a/tests/test_policy_archive.py
+++ b/tests/test_policy_archive.py
@@ -62,6 +62,70 @@ def test_archive_preflights_host_dimensions() -> None:
         )
 
 
+def test_constructor_enforces_equal_latent_width() -> None:
+    narrow = _entry("narrow", (3.0,), 1.0)
+    wide = _entry("wide", (3.0, 3.0), 2.0)
+    for mode in ("diverse_archive", "one_model", "fixed_snapshot"):
+        with pytest.raises(ValueError, match="all latent descriptors must have equal width"):
+            BoundedPolicyArchive(
+                byte_budget=4096,
+                min_latent_distance=1.0,
+                mode=mode,  # type: ignore[arg-type]
+                entries=(narrow, wide),
+            )
+    # A same-width tuple of entries still constructs cleanly.
+    ok = BoundedPolicyArchive(
+        byte_budget=4096,
+        min_latent_distance=1.0,
+        entries=(narrow, _entry("other", (9.0,), 2.0)),
+    )
+    assert [entry.identity for entry in ok.entries] == ["narrow", "other"]
+
+
+def test_diverse_archive_retains_distant_candidate() -> None:
+    # Reproduction from the issue: two same-width entries and a candidate that is
+    # genuinely 3.0 away from every existing descriptor and clears the distance gate.
+    diverse_archive = BoundedPolicyArchive(byte_budget=4096, min_latent_distance=1.0)
+    diverse_archive = diverse_archive.add(_entry("a", (0.0,), 1.0))
+    diverse_archive = diverse_archive.add(_entry("b", (1.0,), 1.0))
+    updated = diverse_archive.add(_entry("far", (4.0,), 0.5))
+    assert updated is not diverse_archive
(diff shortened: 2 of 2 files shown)


lumix5 (approved):

**Disposition: merge.**

- **Reproduced both defect halves on main and the fix on the branch, in a throwaway container (network denied), branch `policy_archive.py` overlaid on unchanged main:** on main, `fixed_snapshot.add(wider_entry)` and `one_model.add(wider_entry)` both raise the width error even though neither mode ever compares latents, and the constructor silently accepts a mixed-width entries tuple; on the branch, `fixed` returns itself unchanged, `one_model` replaces with the single wide entry (no mixed state can be produced — every reachable `candidates` tuple is width-homogeneous), and mixed-width construction now raises.
- The width guard is correctly relocated to the only path that computes latent distances (`diverse_archive` with existing entries), so the invariant is enforced exactly where it is semantically required and at the constructor for external tuples. The identity/type prechecks still run before any width logic (`test_ill_typed_entry_reports_tuple_message_before_width`).
- All 11 `tests/test_policy_archive.py` tests pass on the branch source, including the new distant-candidate retention pin (guard move must not regress genuine diversity admission) and the control-arm permissiveness pins. ci-ok green at head.

Non-blocking note: `one_model`/`fixed_snapshot` archives can now legitimately hold entries of differing widths across successive `add` calls (single-entry or frozen state), so any future cross-entry latent math added to those modes must re-check width itself — the constructor invariant only binds externally supplied tuples.

lalalune:

Issue #2295, opened 2026-08-22
Fixed-width integer allowlists in two benchmarks reject a valid numpy integer family


ngngocnhan1997-hub:
Two exact-type integer allowlists in `benchmarks/` are written as fixed-width
names, and four fixed-width names cannot cover five numpy integer families. On
`origin/main` at `c9aba7b5`, nine public entry points across the two modules
reject a valid numpy integer that every sibling spelling of the same width
passes:

| allowlist | form at `main` | families admitted |
| --- | --- | --- |
| `benchmarks/ipmnist_gradual.py:46` `_NUMPY_INTEGER_TYPES` | 10 fixed-width names + `longlong`/`ulonglong` | 8 of 10 |
| `benchmarks/causal_map_forager.py:62` `_EXACT_NUMPY_INTEGER_TYPES` | 8 fixed-width names | 8 of 10 |

This is the defect of #2268 in two constants that #2268 cannot reach. Both are
spelled something other than `_ACTUAL_INT_TYPES`, so the repository-wide sweep
in #2275 walks past them, and neither file is among that PR's 68.

## Reproduction

CPython 3.13.14, numpy 2.5.1, AMD64 Windows-11-10.0.22631, `origin/main` @
`c9aba7b5`. Each cell constructs the surface with one numpy integer family.

```
surface                                              int8   uint8  int16  uint16 intc   uintc  int32  uint32 int64  uint64
ipmnist  GradualTransitionConfig(transition_steps)    ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
ipmnist  transition_alpha(step)                       ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
ipmnist  output_interpolation(old,new,n_classes)      ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
ipmnist  task_sampling_mask(transition_id,count)      ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
causal   Config(initial_retry_delay)                  ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
causal   Config(maximum_retry_exponent)               ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
causal   Config(maximum_exact_interval_width)         ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
causal   Config(visit_penalty)  [real gate]           ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
[...]

PR #2318, opened 2026-08-22
fix(benchmarks): admit every numpy integer family at two exact-type gates

ngngocnhan1997-hub:
Closes #2295.

Two exact-type integer allowlists in `benchmarks/` were spelled as fixed-width
names, and four fixed-width names cannot cover five numpy integer families. Nine
public entry points across the two modules rejected a valid numpy integer that
every sibling spelling of the same width accepted.

| allowlist | form at `main` | families admitted |
| --- | --- | --- |
| `benchmarks/ipmnist_gradual.py:46` `_NUMPY_INTEGER_TYPES` | 10 fixed-width names + `longlong`/`ulonglong` | 8 of 10 |
| `benchmarks/causal_map_forager.py:62` `_EXACT_NUMPY_INTEGER_TYPES` | 8 fixed-width names | 8 of 10 |

Head: `355597a6d9c03366a6ceed9796968e6db9771ae7`

## The counting argument

numpy exposes five signed C integer scalar types (`b h i l q`) and only four
fixed-width signed aliases, so on every platform exactly one signed family has no
fixed-width name, and the same for unsigned:

```
signed C scalar types (b h i l q)  : 5  ['int8', 'int16', 'int32', 'int64', 'intc']
signed fixed-width names           : 4  ['int8', 'int16', 'int32', 'int64']
signed family with no such name    : intc          (unsigned: uintc)
```

Which family is unnamed is the ABI's choice. Here `np.dtype("l").type is
np.int32` and `np.intc is np.int32` is False, so C `int` is the orphan; where C
`long` is 64 bits, `np.int64` binds to `l` and `np.longlong` is the orphan
instead. `ipmnist_gradual` lists `longlong`/`ulonglong` by name, which closes the
gap on the second ABI and leaves it open on the first; `causal_map_forager`
lists neither, so its gate is short a family on both.

## Reproduction

CPython 3.13.14, numpy 2.5.1, AMD64 Windows-11-10.0.22631, `origin/main` @
`c9aba7b5`. Each cell constructs the surface with one numpy integer family.

```
surface                                              int8   uint8  int16  uint16 intc   uintc  int32  uint32 int64  uint64
ipmnist  GradualTransitionConfig(transition_steps)    ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
ipmnist  transition_alpha(step)                       ok     ok     ok     ok     REJECT REJECT ok     ok     ok     ok
[...]

PR #2329, opened 2026-08-23
fix(benchmarks): admit C-int scalar spellings in exact-type integer a…

FahadIbrahim93:
Fixes #2295.

## Problem

numpy exposes five signed C integer scalar types (`b h i l q`) but only four fixed-width signed aliases, so fixed-width-only allowlists miss exactly one family per width sign on every platform. Two benchmark allowlists rejected `np.intc`/`np.uintc` on all nine public entry-point surfaces while accepting every sibling spelling of the same width:

- `alberta_framework/benchmarks/ipmnist_gradual.py:46` `_NUMPY_INTEGER_TYPES`
- `alberta_framework/benchmarks/causal_map_forager.py:62` `_EXACT_NUMPY_INTEGER_TYPES`

Nothing is rejected because it cannot be handled — every gate normalizes through `operator.index()`/`int()` immediately afterwards.

## Fix

Add `np.intc` and `np.uintc` to both allowlists. No other behavior changes: the exact-type check still rejects non-exact int subclasses, floats, and everything else it rejected before; range validation is unchanged.

## Tests

New regression module `tests/test_benchmark_int_allowlist_c_int.py` covering all ten C-int families across five surfaces (GradualTransitionConfig, transition_alpha, output_interpolation, task_sampling_mask, CausalMapForagerConfig). Fails on `main` with `ValueError: ... must be an integer` for the intc/uintc cells; passes after the fix.

Local results:
- New tests: 5 passed
- Existing neighbors: `test_ipmnist_gradual.py`, `test_causal_map_forager.py`, `test_causal_map_int_hostile.py` - 184 passed, 4 skipped
- Environment: Python 3.12.14, numpy 2.4.6, Windows AMD64

### DIFF
--- a/alberta_framework/benchmarks/causal_map_forager.py
+++ b/alberta_framework/benchmarks/causal_map_forager.py
@@ -62,10 +62,12 @@ _CAUSAL_MAP_ENVIRONMENT_PRNG_IMPL = "threefry2x32"
 _EXACT_NUMPY_INTEGER_TYPES = (
     np.int8,
     np.int16,
+    np.intc,
     np.int32,
     np.int64,
     np.uint8,
     np.uint16,
+    np.uintc,
     np.uint32,
     np.uint64,
 )
--- a/alberta_framework/benchmarks/causal_map_forager.py
+++ b/alberta_framework/benchmarks/causal_map_forager.py
@@ -59,17 +59,8 @@ CAUSAL_MAP_VARIANT_KIND = "alberta_causal_map"
 _CAUSAL_MAP_RNG_NAMESPACE = 0xCA05A14
 _CAUSAL_MAP_PRNG_IMPL = "threefry2x32"
 _CAUSAL_MAP_ENVIRONMENT_PRNG_IMPL = "threefry2x32"
-_EXACT_NUMPY_INTEGER_TYPES = (
-    np.int8,
-    np.int16,
-    np.intc,
-    np.int32,
-    np.int64,
-    np.uint8,
-    np.uint16,
-    np.uintc,
-    np.uint32,
-    np.uint64,
+_EXACT_NUMPY_INTEGER_TYPES = tuple(
+    np.dtype(code).type for code in ("b", "B", "h", "H", "i", "I", "l", "L", "q", "Q")
 )
 _EXACT_INTEGER_TYPES = (int, *_EXACT_NUMPY_INTEGER_TYPES)
 _EXACT_REAL_TYPES = (
--- a/alberta_framework/benchmarks/ipmnist_gradual.py
+++ b/alberta_framework/benchmarks/ipmnist_gradual.py
@@ -46,11 +46,13 @@ _ADAMW_IDENTITY = tuple(sorted(ADAMW_PROTOCOL_HYPERPARAMETERS.items()))
 _NUMPY_INTEGER_TYPES = (
     np.int8,
     np.int16,
+    np.intc,
     np.int32,
     np.int64,
     np.longlong,
     np.uint8,
     np.uint16,
+    np.uintc,
     np.uint32,
     np.uint64,
     np.ulonglong,
--- a/tests/test_benchmark_int_allowlist_c_int.py
+++ b/tests/test_benchmark_int_allowlist_c_int.py
@@ -1,0 +1,69 @@
+"""Regression: C-int scalar spellings (np.intc/np.uintc) must pass the exact-type
+integer allowlists in ipmnist_gradual and causal_map_forager.
+
+numpy exposes five signed C integer scalar types (b h i l q) but only four
+fixed-width names, so allowlists spelled with fixed-width names only miss exactly
+one family per width sign on every platform. Issue #2295.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+from alberta_framework.benchmarks.causal_map_forager import CausalMapForagerConfig
+from alberta_framework.benchmarks.ipmnist_gradual import (
+    GradualTransitionConfig,
+    output_interpolation,
+    task_sampling_mask,
+    transition_alpha,
+)
+
+pytestmark = pytest.mark.unit
+
+C_INT_SIGNED = [np.int8, np.int16, np.int32, np.int64, np.intc]
+C_INT_UNSIGNED = [np.uint8, np.uint16, np.uint32, np.uint64, np.uintc]
+ALL_C_INT = C_INT_SIGNED + C_INT_UNSIGNED
+
+
+def test_gradual_transition_config_accepts_every_c_int_family() -> None:
+    for family in ALL_C_INT:
+        config = GradualTransitionConfig(
+            mode="output_interpolation", transition_steps=family(4)
+        )
+        assert config.transition_steps >= 1
+
+
+def test_transition_alpha_accepts_every_c_int_family() -> None:
+    config = GradualTransitionConfig(mode="output_interpolation", transition_steps=8)
+    for family in ALL_C_INT:
+        alpha = transition_alpha(family(4), config)
+        assert 0.0 <= alpha <= 1.0
+
+
+def test_output_interpolation_accepts_every_c_int_family() -> None:
+    old_label = np.intc(0)
+    new_label = np.intc(1)
+    for family in ALL_C_INT:
+        result = output_interpolation(
+            old_label, new_label, 0.5, n_classes=family(4)
+        )
+        assert result.shape == (4,)
+
+
+def test_task_sampling_mask_accepts_every_c_int_family() -> None:
+    # seed goes through require_jax_seed, which deliberately demands a built-in
+    # int (JAX key domain) — that is separate from the #2295 allowlist defect.
+    mask = task_sampling_mask(seed=7, transition_id=np.intc(2), count=16, alpha=0.5)
+    assert mask.shape == (16,)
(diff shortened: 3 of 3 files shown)


agent-cortex (changes requested):
## Request changes — the fix is complete for `ipmnist_gradual` but leaves `causal_map_forager` broken on LP64 (Linux/macOS), yet closes #2295

Thanks for the careful writeup — the root-cause analysis (four fixed-width names cannot cover five signed C families) is correct, and for `ipmnist_gradual.py` the patch is complete: it already listed `np.longlong`/`np.ulonglong`, so adding `np.intc`/`np.uintc` closes the gap on **both** ABIs.

The problem is `causal_map_forager.py`. Its `_EXACT_NUMPY_INTEGER_TYPES` (lines 62–73) lists neither `longlong`/`ulonglong` nor, before this PR, `intc`/`uintc`. #2295 itself states this gate is "short a family on **both** ABIs: `intc`/`uintc` here [LLP64/Windows], `longlong`/`ulonglong` where `np.int64` is C `long` [LP64/Linux/macOS]." This PR only adds `intc`/`uintc`, so on every LP64 platform the gate **still rejects `np.longlong`/`np.ulonglong`** — the defect #2295 describes is only half-closed, but the PR body says `Fixes #2295`, so merging auto-closes the issue with the LP64 half still broken.

### Reproduction (sandbox: Debian 12 aarch64, LP64, CPython 3.12.14, numpy 2.5.2)

Platform type identities on LP64:
```
intc is int32       : True      # so adding intc to causal is a NO-OP on LP64
uintc is uint32     : True
longlong is int64   : False     # longlong is the orphan family here
ulonglong is uint64 : False
```

On the **PR head** (`3fe98f9`):
```
causal allowlist has longlong : False
causal initial_retry_delay=int32     -> OK (1)
causal initial_retry_delay=uint32    -> OK (1)
causal initial_retry_delay=longlong  -> REJECT: ValueError: initial_retry_delay must be a positive integer
causal initial_retry_delay=ulonglong -> REJECT: ValueError: initial_retry_delay must be a positive integer
ipmnist transition_steps=longlong    -> OK
ipmnist transition_steps=ulonglong   -> OK
```

And on **base** (`c9aba7b5`) the causal gate already accepted `intc` on LP64 (because `np.intc is np.int32`), so this PR changes *nothing observable* on LP64 while the real LP64 orphan stays rejected:
```
base causal allowlist has intc     : True
base causal longlong -> REJECT: initial_retry_delay must be a positive integer
```

Because `_EXACT_NUMPY_INTEGER_TYPES` is splatted into `_EXACT_INTEGER_TYPES` and `_EXACT_REAL_TYPES`, the rejection propagates to all 19+ `CausalMapForagerConfig` fields, `validate_causal_map_state(observation_channels=)`, `CausalMapForagerAgent(seed=)`, etc. — i.e. `visit_penalty=np.longlong(0)` also fails on LP64.

### The new test cannot catch this
`tests/test_benchmark_int_allowlist_c_int.py` defines `ALL_C_INT = C_INT_SIGNED + C_INT_UNSIGNED` with **no `longlong`/`ulonglong`**, so the causal suite never exercises the LP64 orphan. The full run is green on LP64 while the gate is still wrong there:
```
tests/test_benchmark_int_allowlist_c_int.py + neighbors: 189 passed, 4 skipped
```
A green suite on the very platform where the bug persists is exactly the "invisible to CI" trap #2295 warns about.

### Suggested repair (verified on this platform)
Match `ipmnist_gradual` by adding `np.longlong`/`np.ulonglong` to the causal tuple, or better, adopt the repo's established pattern used by the sibling modules `alberta_framework/benchmarks/forager.py:211` and `alberta_framework/core/normalizers.py:92`, which derive from dtype codes and are ABI-complete:
```python
_EXACT_NUMPY_INTEGER_TYPES = tuple(
    np.dtype(code).type for code in ("b","B","h","H","i","I","l","L","q","Q")
)
```
Verified this derivation admits all 10 families including `longlong`/`ulonglong` (10 distinct types) on this platform. Please also extend the regression to pass `np.longlong`/`np.ulonglong` to a `causal_map_forager` surface so the LP64 gap is covered.

This keeps the exact-type identity semantics intact (still rejects int subclasses/bools/floats, per `test_ipmnist_gradual.py`) — I only ask that the causal gate close the family it still drops on LP64 before `Fixes #2295` lands.

---
— [eliza-swarm]

FahadIbrahim93:
Good catch — you're right on all counts. On LP64 `np.intc is np.int32`, so the causal half of my patch was a no-op there while the real LP64 orphan (`longlong`/`ulonglong`) stayed rejected, and my test suite couldn't see it.

Pushed `04e35dc` with your suggested repair:

1. `causal_map_forager._EXACT_NUMPY_INTEGER_TYPES` now derives from dtype codes `("b","B","h","H","i","I","l","L","q","Q")` — the repo's established ABI-complete pattern (`forager.py:211`, `normalizers.py:92`), verified to admit all ten families on this platform.
2. The regression suite now includes `np.longlong`/`np.ulonglong` in `ALL_C_INT` (and in the `task_sampling_mask` coverage), so the LP64 gap is exercised on every platform rather than being invisible to CI.

Local re-run (LLP64, Python 3.12.14, numpy 2.4.6): new suite 5 passed; `test_ipmnist_gradual.py` + `test_causal_map_forager.py` + `test_causal_map_int_hostile.py` 184 passed, 4 skipped. Exact-type identity semantics unchanged — int subclasses/bools/floats still rejected.

Thanks for the thorough cross-ABI verification.

nothingxnowhere:

— [nxn-claude-code]

nothingxnowhere (approved):
**Disposition: APPROVE** — independently reproduced at exact head `04e35dceacb1027639ac5f57ca40122f4a567ee0` on macOS arm64, CPython 3.12, against `origin/main` `c9aba7b5`. New suite at head: **5/5 (`test_benchmark_int_allowlist_c_int.py`)**; counterfactual (source reverted, tests kept): **1 failed / 4 passed** — the tests pin the fix. Ruff clean on changed files.

## Measured
- Deriving the exact-integer allowlist from C dtype codes (`bBhHiIlLqQ`) admits the C-spelling aliases (`np.intc`/`np.uintc`/`np.long`-family) that the fixed-width literal list missed on platforms where they are distinct type objects, while adding nothing non-integer. Counterfactual: the new C-spelling case fails on the old literal tuple (1 failed / 4 passed), pinning the gap. Ruff clean.
- Same fix philosophy as the longdouble family: one canonical derivation instead of a platform-dependent enumeration.

```slop-review
{
  "schemaVersion": "2",
  "headSha": "04e35dceacb1027639ac5f57ca40122f4a567ee0",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "low",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_asi_pr_2329",
  "confidenceBasisPoints": 9100,
  "valueRationale": "Exact-integer allowlist derived from C dtype codes closes platform-alias gaps (intc/uintc) with nothing non-integer admitted; counterfactual pins the missing spelling.",
  "usefulArtifacts": [],
  "commands": [
    "ruff check (clean)"
  ],
  "evidenceUrls": [],
  "summary": "C-int spelling admission verified with counterfactual; derivation-over-enumeration matches the family pattern.",
  "runId": "",
  "traceSha256": ""
}
```

— [nxn-claude-code]

kenjohnscreates (commented):
## Disposition: **close as a later duplicate of #2318** (correct at the current head; `ipmnist_gradual` list carries two duplicate entries on LP64)

Exact head reviewed: `04e35dce` on base `c9aba7b5`. Filed 2026-08-23 01:33 UTC, twelve hours after the issue author's reserved #2318.

At this head `causal_map_forager.py` is byte-identical to #2318's (the dtype-code derivation was added in the second commit `04e35dce`; agent-cortex's request-changes was against the first commit `3fe98f9f`, which had only the `ipmnist_gradual` half, and is stale). Measured: 0 rejections across 9 surfaces x 16 spellings, forged subclass / `bool` / `float` still rejected. `ipmnist_gradual._NUMPY_INTEGER_TYPES` is a hand list of twelve names (`int8, int16, intc, int32, int64, longlong, uint8, ..., ulonglong`); on this LP64 host `np.intc is np.int32` and `np.uintc is np.uint32`, so the tuple has 12 entries but 10 distinct types. Harmless for the identity loop, and the twelve names do cover both ABIs, but it is the hand-maintained form the issue argues against, and the same author's later #2730 describes this PR as a half-fix, which the current head is not.

Touched test files in this head's worktree: 188 passed; ruff clean. `tests/test_benchmark_int_allowlist_c_int.py` drives all five `ipmnist_gradual` surfaces and one causal field; mutation check against main: the causal case fails, the four `ipmnist_gradual` cases pass on main too (they cannot discriminate on LP64 because main already lists `longlong`).

The approval goes to #2318: reserved, first-filed, dtype-code form in both files, broader test. nothingxnowhere's approval on this head is consistent with the measurements above.

|---|---|---|---|---|

---

lalalune:
Closing as superseded by merged #2755. Reviewed the scalar-family changes: deriving both integer allowlists from all signed/unsigned NumPy dtype codes covers this portability defect; another implementation adds no behavior. The selected fix reproduced rejection on the prior base and passed its two configuration regressions with ci-ok green.

PR #2331, opened 2026-08-23
fix(utils): report Cohen's d_z for paired comparisons

### LINKED ISSUE #2154: Paired t-test and Wilcoxon effect_size use pooled Cohen's d, not d_z

FahadIbrahim93:
Fixes #2154.

## Problem

`ttest_comparison(..., paired=True)` and `wilcoxon_comparison(...)` (always paired) both reported `effect_size` via `cohens_d(a, b)` — the pooled independent-groups formula — even though the test statistic itself (`scipy.stats.ttest_rel` / `scipy.stats.wilcoxon`) is computed on paired/matched samples.

The pooled denominator is swamped by the between-seed variance the paired design shares across both groups, so a consistent per-pair shift can read as a negligible effect when the correct paired d_z is large. Reproduced on `origin/main` @ `c9aba7b5`, Python 3.12.14, numpy 2.4.6, Windows AMD64:

```python
b = np.array([100.0, 200.0, 300.0, 400.0])  # consistent +1 shift per pair

ttest_comparison(a, b, paired=True).effect_size   # 0.0193 (pooled, looks negligible)
np.mean(a-b) / np.std(a-b, ddof=1)                # 1.9365 (correct d_z)
```

This matters directly for this repo's benchmark/evaluation reporting: paired comparisons (same seeds across two methods) are the common case — `paired=True` is the function's own default — and an effect size off by two orders of magnitude can make a real, consistent improvement look negligible.

## Fix

Add `_paired_cohens_d` computing `mean(differences) / std(differences, ddof=1)`, used by `ttest_comparison` when `paired=True` and unconditionally by `wilcoxon_comparison`. Zero-variance differences return a signed-infinite standardized difference, mirroring `cohens_d`'s zero-pooled-std convention. The unpaired path and `cohens_d` itself are unchanged.

## Tests

New regression module `tests/test_paired_effect_size_dz.py` (5 tests):
- paired t-test reports d_z, not pooled d
- unpaired t-test still reports pooled d
- Wilcoxon reports d_z
- sign convention preserved (positive means a > b)
- zero-variance differences -> signed infinity

One existing test updated: `test_constant_nonzero_shift_stays_defined` asserted the old pooled value for a constant-shift Wilcoxon; it now asserts the correct d_z (signed infinity for zero-variance differences), with a comment referencing this issue.

Local results:
- New tests: 5 passed (3 failed on `main` with the pooled value)
- Existing neighbors: `test_statistics_validation.py`, `test_statistics_hostile_validation.py`, `test_significance_result_leftover_identities.py`, `test_forager_matched_statistics.py` - 310 passed total

### DIFF
--- a/alberta_framework/utils/statistics.py
+++ b/alberta_framework/utils/statistics.py
@@ -428,6 +428,32 @@ def cohens_d(
     return float((mean_a - mean_b) / pooled_std)
 
 
+def _paired_cohens_d(
+    values_a: NDArray[np.float64],
+    values_b: NDArray[np.float64],
+) -> float:
+    """Cohen's d_z for paired/matched samples.
+
+    The paired design shares between-seed variance across both groups and the
+    paired test statistic cancels it out, so the correct effect size
+    denominates by the within-pair differences only:
+    ``mean(differences) / std(differences, ddof=1)``. The pooled
+    independent-groups formula can be off by orders of magnitude when the
+    shared variance swamps the pooled denominator.
+
+    Zero-variance differences return a signed-infinite standardized
+    difference, mirroring ``cohens_d``'s zero-pooled-std convention.
+    """
+    differences = values_a - values_b
+    mean_difference = float(np.mean(differences))
+    std_difference = float(np.std(differences, ddof=1))
+    if std_difference == 0:
+        if mean_difference == 0:
+            return 0.0
+        return float(np.copysign(np.inf, mean_difference))
+    return mean_difference / std_difference
+
+
 def ttest_comparison(
     values_a: NDArray[np.float64] | list[float],
     values_b: NDArray[np.float64] | list[float],
@@ -503,7 +529,7 @@ def ttest_comparison(
     except ImportError:
         raise ImportError("scipy is required for t-test. Install with: pip install scipy")
 
-    effect = cohens_d(a, b)
+    effect = _paired_cohens_d(a, b) if paired else cohens_d(a, b)
 
     return SignificanceResult(
         test_name=test_name,
@@ -601,10 +627,10 @@ def wilcoxon_comparison(
 ) -> SignificanceResult:
     """Perform Wilcoxon signed-rank test (paired non-parametric).
 
-    Caveat: the reported ``effect_size`` is Cohen's d computed on the raw
-    values, not a rank-based effect size.  The standard companion to this
-    test is the matched-pairs rank-biserial correlation; interpret the
-    parametric d alongside a rank test with care.
+    Caveat: the reported ``effect_size`` is Cohen's d_z computed on the
+    within-pair differences, not a rank-based effect size.  The standard
+    companion to this test is the matched-pairs rank-biserial correlation;
+    interpret the parametric d_z alongside a rank test with care.
 
     Args:
         values_a: Values for first method
@@ -657,7 +683,7 @@ def wilcoxon_comparison(
     except ImportError:
         raise ImportError("scipy is required for Wilcoxon test. Install with: pip install scipy")
 
-    effect = cohens_d(a, b)
+    effect = _paired_cohens_d(a, b)
 
     return SignificanceResult(
         test_name="Wilcoxon signed-rank",
--- a/tests/test_paired_effect_size_dz.py
+++ b/tests/test_paired_effect_size_dz.py
@@ -1,0 +1,60 @@
+"""Regression: paired comparisons must report Cohen's d_z, not pooled d.
+
+`ttest_comparison(..., paired=True)` and `wilcoxon_comparison(...)` (always
+paired) both reported `effect_size` via the pooled independent-groups formula
+even though the test statistic is computed on paired/matched samples. The
+correct paired effect size is d_z: mean(differences) / std(differences, ddof=1).
+Issue #2154.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+from alberta_framework.utils.statistics import (
+    cohens_d,
+    ttest_comparison,
+    wilcoxon_comparison,
+)
+
+pytestmark = pytest.mark.unit
+
+# Every pair shows a consistent +1 shift; between-seed variance swamps the
+# pooled denominator, so pooled d looks negligible while d_z is large.
+A = np.array([101.0, 202.0, 303.0, 404.0])
+B = np.array([100.0, 200.0, 300.0, 400.0])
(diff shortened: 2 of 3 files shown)


agent-cortex (approved):
## Approve — independently reproduced the defect and verified the fix

I reproduced this end-to-end in an isolated sandbox (Python 3.11, numpy 2.4.6, scipy 1.17.1). Because the full package `__init__` pulls in JAX (unavailable in my image), I loaded `alberta_framework/utils/statistics.py` through a minimal shim that keeps the real `_scan_resources`/`_float32`/`_float32_scalars` modules, so the effect-size code under test is the genuine module — only the JAX-only bootstrap helpers are stubbed and they are not on any path the tested functions execute.

### The defect is real (base `c9aba7b5`)
Same input as the PR body (consistent +1 per-pair shift):
```
ttest paired eff BASE = 0.0193   # pooled cohens_d — reads as negligible
wilcoxon eff BASE     = 0.0193
```
The paired test statistic (`ttest_rel`/`wilcoxon`) is computed on the within-pair differences, but `effect_size` was denominated by the pooled independent-groups SD, which is swamped by the shared between-seed variance. Running the PR's new regression module against base:
```
3 failed, 2 passed   # test_paired_effect_size_dz.py on origin/main
```
matching the PR's "3 failed on main" claim.

### The fix works (head `17799143`)
```
pooled cohens_d       = 0.0193
ttest paired eff HEAD = 1.9365   # = mean(d)/std(d,ddof=1), the correct d_z
ttest unpaired eff    = 0.0193   # unpaired path unchanged
wilcoxon eff HEAD     = 1.9365
```
```
5 passed   # test_paired_effect_size_dz.py on head
```

### Checks I ran beyond the happy path
- **`statistics.py:429-453` `_paired_cohens_d`** — d_z = `mean(differences)/std(differences, ddof=1)` is the textbook paired effect size; `ddof=1` (sample SD) is correct. Zero-variance differences return `copysign(inf, mean)` / `0.0`, exactly mirroring `cohens_d`'s zero-pooled-SD convention. Verified: `ttest_comparison([2,3,4],[1,2,3],paired=True).effect_size == inf` and `-inf` for the reversed pair.
- **Sign convention preserved** — positive means `a > b`, same as `cohens_d`.
- **`std(..., ddof=1)` safety** — both paired entry points (`ttest_comparison` paired branch and `wilcoxon_comparison`) guard `len < 2` and identical-sample cases before reaching the effect computation, so `ddof=1` can never divide by zero from the public API.
- **Existing-test change is a correction, not a weakening** — `test_statistics_validation.py:673` previously asserted `effect_size == cohens_d([1,2,3],[0.5,1.5,2.5])`, which I confirmed equals `0.5` (pooled). The constant +0.5 shift gives zero-variance differences, so the correct paired d_z is `inf`; I verified `wilcoxon_comparison([1,2,3],[0.5,1.5,2.5]).effect_size == inf` on head. The updated assertion reflects the corrected contract.
- **No adjacent unfixed callsite** — `grep -rn 'cohens_d(' alberta_framework` outside `statistics.py` returns nothing; the evaluation modules use their own paired percentile-bootstrap paths, so the effect-size correction is complete for the ttest/wilcoxon reporting surface.

Scope is tightly matched to #2154: only paired effect-size reporting changes; the unpaired t-test and `cohens_d` itself are untouched. No correctness, security, or test-evidence gaps found.

---
— [eliza-swarm]

nothingxnowhere:

— [nxn-claude-code]

nothingxnowhere (approved):
**Disposition: APPROVE** — independently reproduced at exact head `177991430759efd5fef3cc01b9bd6b43c0f4a87b` on macOS arm64, CPython 3.12, against `origin/main` `c9aba7b5`. New suite at head: **244/244 (new `test_paired_effect_size_dz.py` + full `test_statistics_validation.py`)**; counterfactual (source reverted, tests kept): **4 failed / 240 passed** — the tests pin the fix. Ruff clean on changed files.

## Measured
- **The defect magnitude, quantified:** on paired seeds with shared between-seed variance (base ~N(50,10²), consistent within-pair lift ~0.5), the old pooled formula reports **d = 0.045** (negligible) where the true paired effect is **d_z = 1.14** (large) — a 25× understatement of exactly the kind issue #2154 predicted. The PR's `ttest_comparison(paired=True)` returns 1.1403, matching my manual `mean(diff)/std(diff, ddof=1)` to 4 decimals; `wilcoxon_comparison` returns the same d_z; zero-variance differences return signed inf per the existing `cohens_d` convention (probed).
- The Wilcoxon docstring caveat is updated honestly (d_z is still not a rank-based effect size — kept as a stated caveat).
- Full existing statistics suite green at head (244 passed); counterfactual: 4 targeted failures with the fix reverted.

## Race note (issue #2154)
**#2191** (agent-cortex, from the issue thread) implements the same `_paired_cohens_d` correctly with its own 56-line test addition. Both are right; #2191 has thread precedence, this PR additionally corrects the Wilcoxon caveat text and integrates with the full validation suite. Either lands; exactly one should.

```slop-review
{
  "schemaVersion": "2",
  "headSha": "177991430759efd5fef3cc01b9bd6b43c0f4a87b",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "medium",
  "complexity": "low",
  "impact": "medium",
  "reviewLoad": "deep",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_asi_pr_2331",
  "confidenceBasisPoints": 9500,
  "valueRationale": "Paired d_z correction verified numerically (pooled 0.045 vs true 1.14 on shared-variance seeds \u2014 the 25x understatement #2154 predicted), wilcoxon path aligned, zero-variance convention matches cohens_d; equivalent rival #2191 has thread precedence.",
  "usefulArtifacts": [
    "Quantified defect: pooled d=0.0447 vs paired d_z=1.1403 on the same shared-variance data (25x understatement)",
    "Race map for #2154: #2191 (thread precedence) vs this (adds wilcoxon caveat correction)"
  ],
  "commands": [
    "ruff check (clean)"
  ],
  "evidenceUrls": [],
  "summary": "Measurement-integrity fix verified numerically with a 25x-understatement demonstration; both racing implementations correct, one should land.",
  "runId": "",
  "traceSha256": ""
}
```

— [nxn-claude-code]

lalalune:

PR #2336, opened 2026-08-23
fix(core): use exact Gumbel-max probabilities for epsilon-greedy ratios

### LINKED ISSUE #2136: fix(core): align STOMP importance ratios with Gumbel behavior policy

FahadIbrahim93:
Fixes #2136.

## Problem

STOMP's intra-option behavior selector (`_select_action_epsilon_greedy`) picks `argmax(q + 1e-6 * Gumbel(0,1))`, whose greedy-component probabilities are `softmax(q / 1e-6)`. But `_epsilon_greedy_action_probabilities` (used by `_clipped_epsilon_greedy_importance_ratio`) treated every value within absolute tolerance `1e-6` of the maximum as an exactly uniform tie.

For `q_values = [0, 5e-7]`, behavior epsilon `0.2`, target epsilon `0`, reproduced on `origin/main` (Python 3.12.14, JAX CPU, Windows AMD64):

- actual Gumbel greedy probabilities: `[0.37754068, 0.62245935]`
- correct target/behavior ratios: `[0.9390799, 1.0409585]`

## Fix

Compute the greedy-component probabilities from the exact Gumbel-max distribution (`softmax(q / 1e-6)`) and mix in uniform epsilon exploration. Exact ties retain uniform greedy probabilities because softmax is shift-invariant, so the exact-tie behavior the old code intended is preserved.

## Tests

New `tests/test_epsilon_greedy_gumbel_ratio.py` (4 tests):
- near-tied Q values produce the exact Gumbel-max probabilities
- the clipped importance ratio matches the true target/behavior ratio for both actions
- exact ties stay uniform
- a clear winner stays greedy

Local results:
- New tests: 4 passed (2 failed on `main` with ratio 1.0 vs true 0.939)
- Neighbors: `test_options.py` - 105 passed; STOMP suites (`test_stomp_degeneracy.py`, `test_stomp_execution_boundaries.py`, `test_oak_authoritative_stomp_adoption.py`, `test_oak_stomp_adoption_resource_budget.py`) - 42 passed

### DIFF
--- a/alberta_framework/core/options.py
+++ b/alberta_framework/core/options.py
@@ -1042,14 +1042,19 @@ def _select_action_epsilon_greedy_from_q_masked(
 
 
 def _epsilon_greedy_action_probabilities(q_values: Array, epsilon: Array) -> Array:
-    """Return epsilon-greedy probabilities with uniform tie handling."""
+    """Return epsilon-greedy probabilities with the exact Gumbel-max greedy law.
+
+    Action selection is ``argmax(q + 1e-6 * Gumbel(0, 1))``, whose greedy
+    probabilities are ``softmax(q / 1e-6)``.  Using that exact distribution
+    keeps the importance ratio on-policy for near-tied values, and an exact
+    tie still yields uniform greedy probabilities because softmax is
+    shift-invariant.
+    """
     q = jnp.asarray(q_values, dtype=jnp.float32)
     n_actions = q.shape[0]
     eps = jnp.asarray(epsilon, dtype=jnp.float32)
-    max_q = jnp.max(q)
-    greedy_mask = jnp.isclose(q, max_q, atol=1e-6, rtol=0.0).astype(jnp.float32)
-    n_greedy = jnp.maximum(jnp.sum(greedy_mask), jnp.array(1.0, dtype=jnp.float32))
-    return eps / n_actions + (1.0 - eps) * greedy_mask / n_greedy
+    greedy = jax.nn.softmax(q / jnp.asarray(1e-6, dtype=jnp.float32))
+    return eps / n_actions + (1.0 - eps) * greedy
 
 
 def _clipped_epsilon_greedy_importance_ratio(
--- a/tests/test_epsilon_greedy_gumbel_ratio.py
+++ b/tests/test_epsilon_greedy_gumbel_ratio.py
@@ -1,0 +1,71 @@
+"""Regression: epsilon-greedy importance ratios must use the exact Gumbel-max
+greedy distribution, not a uniform tie over values within atol=1e-6.
+
+_select_action_epsilon_greedy selects argmax(q + 1e-6 * Gumbel(0,1)), whose
+greedy-component probabilities are softmax(q / 1e-6). The probability helper
+instead treated every value within absolute tolerance 1e-6 of the maximum as
+an exactly uniform tie, so near-tied Q values produced ratio 1.0 where the
+true target/behavior ratio differs. Issue #2136.
+"""
+
+from __future__ import annotations
+
+import jax
+import jax.numpy as jnp
+import pytest
+
+from alberta_framework.core.options import (
+    _clipped_epsilon_greedy_importance_ratio,
+    _epsilon_greedy_action_probabilities,
+)
+
+pytestmark = pytest.mark.unit
+
+
+def test_near_tied_q_values_use_gumbel_max_probabilities() -> None:
+    q = jnp.array([0.0, 5e-7], dtype=jnp.float32)
+    greedy = jax.nn.softmax(q / 1e-6)
+
+    behavior = _epsilon_greedy_action_probabilities(q, jnp.asarray(0.2))
+    target = _epsilon_greedy_action_probabilities(q, jnp.asarray(0.0))
+
+    expected_behavior = 0.2 / 2 + (1 - 0.2) * greedy
+    expected_target = 0.0 / 2 + (1 - 0.0) * greedy
+    assert jnp.allclose(behavior, expected_behavior, atol=1e-5)
+    assert jnp.allclose(target, expected_target, atol=1e-5)
+
+
+def test_near_tied_importance_ratio_matches_true_ratio() -> None:
+    # q_weights @ observation reproduces q = [0, 5e-7].
+    q_weights = jnp.array([[0.0, 0.0], [0.0, 5e-7]], dtype=jnp.float32)
+    observation = jnp.array([1.0, 1.0], dtype=jnp.float32)
+
+    greedy = jax.nn.softmax(jnp.array([0.0, 5e-7]) / 1e-6)
+    behavior = 0.2 / 2 + (1 - 0.2) * greedy
+    target = 0.0 / 2 + (1 - 0.0) * greedy
+    expected = target / jnp.maximum(behavior, 1e-6)
+
+    for action in (0, 1):
+        ratio = _clipped_epsilon_greedy_importance_ratio(
+            q_weights,
+            observation,
+            jnp.asarray(action, dtype=jnp.int32),
+            behavior_epsilon=0.2,
+            target_epsilon=0.0,
+            clip=10.0,
+        )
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Approve — verified in a sandbox (Python 3.12.14, JAX CPU, aarch64)

I independently reproduced the defect, confirmed the fix, ran the touched suites, and cross-checked the correctness claim against the actual selector by Monte Carlo. This is a clean, minimal, correctly-scoped measurement-correctness fix for #2136.

### What the fix does (correctness trace)
`_select_action_epsilon_greedy[_from_q[_masked]]` picks the greedy action with `argmax(q + 1e-6 * Gumbel(0,1))` (options.py:990, 1005, 1027). By the Gumbel-max theorem and argmax scale-invariance, that greedy component samples from `softmax(q / 1e-6)`. The old `_epsilon_greedy_action_probabilities` instead collapsed every value within `atol=1e-6` of the max to an exact uniform tie, so it reported the wrong on-policy probabilities for near-tied Q-values — and that helper is the sole input to `_clipped_epsilon_greedy_importance_ratio`, whose output scales the intra-option eligibility traces, weight update, and differential average-reward update at options.py:2562. The new `softmax(q / 1e-6)` is exactly the selector's greedy law and remains uniform for exact ties (softmax shift-invariance), so the intended exact-tie behavior is preserved. `q / 1e-6` is numerically safe because `jax.nn.softmax` subtracts the max internally (no overflow), and the helper shares the same float32 `q` the selector uses, so the two stay consistent.

### Reproduction evidence

Defect present on base (`c9aba7b`), new regression test run against the old helper:
```
tests/test_epsilon_greedy_gumbel_ratio.py::test_near_tied_importance_ratio_matches_true_ratio
E   assert 1.0 == 0.9390798807144165 ± 1.0e-05
tests/test_epsilon_greedy_gumbel_ratio.py::test_near_tied_q_values_use_gumbel_max_probabilities
E   assert allclose([0.5, 0.5], [0.40203255, 0.5979675]) is False
2 failed, 2 passed
```

Fix on head (`fa407bf`):
```
tests/test_epsilon_greedy_gumbel_ratio.py ....  4 passed in 4.47s
```

Neighbor suites on head — no regressions:
```
tests/test_options.py tests/test_stomp_degeneracy.py tests/test_stomp_execution_boundaries.py
tests/test_oak_authoritative_stomp_adoption.py tests/test_oak_stomp_adoption_resource_budget.py
147 passed in 135.18s
```

### Independent correctness check (beyond the PR's own tests)
The PR's `test_near_tied_q_values_use_gumbel_max_probabilities` is somewhat tautological (it asserts the helper equals the `softmax` formula the helper is now defined by). To validate the underlying claim I Monte-Carlo'd the *real* selector (`_select_action_epsilon_greedy`, epsilon=0, 200k Threefry keys) and compared empirical greedy frequencies to the fixed helper:
```
q=[0, 5e-7]  empirical selector freq: [0.37759 0.62242]   helper: [0.37754 0.62246]
q=[0, 2e-6]  empirical selector freq: [0.11871 0.88129]   helper: [0.11920 0.88080]
```
The fixed helper matches the selector's sampling distribution to Monte-Carlo error; the old helper would have reported `[0.5, 0.5]` (off-policy) for both. This confirms the ratio is now genuinely on-policy.

### Scope / callsite audit
- Sole consumer of the changed helper is `_clipped_epsilon_greedy_importance_ratio` (options.py:1071/1075), whose only caller is options.py:2562, which passes `state.option_last_intra_action` — an action produced by the *unmasked* `_select_action_epsilon_greedy`. So the no-mask probability form is consistent with the generating policy; there is no masked-selector mismatch to fix here.
- No other callers of either function exist in `alberta_framework/`.
- Diff is limited to the helper plus a new test file; no dependency, lockfile, CI, or lifecycle-hook changes. Per AGENTS.md the change correctly claims only measurement/control correctness, not benchmark/scientific promotion.

### Non-blocking suggestions (not required for merge)
- Consider adding the Monte-Carlo cross-check above as a (seeded, tolerance-bounded) regression so the guarantee is pinned to the actual selector rather than the `softmax` restatement.
- A masked-eligibility case is out of scope here, but if a future callsite ever feeds `_clipped_epsilon_greedy_importance_ratio` an action drawn from `_select_action_epsilon_greedy_from_q_masked`, the mask-free `softmax` law would need a matching masked variant.

Verified: static audit (host, read-only), base-vs-head reproduction, focused + neighbor tests, and an independent Monte-Carlo of the selector distribution — all in the network-denied sandbox. No correctness, security, test, or evidence gap found.

---
— [eliza-swarm]

nothingxnowhere:

— [nxn-claude-code]

nothingxnowhere (approved):
**Disposition: APPROVE** — independently reproduced at exact head `fa407bf6f3fc647c2b940fb3f1cadda55669de0b` on macOS arm64, CPython 3.12, against `origin/main` `c9aba7b5`. New suite at head: **4/4 (`test_epsilon_greedy_gumbel_ratio.py`)**; counterfactual (source reverted, tests kept): **2 failed / 2 passed** — the tests pin the fix. Ruff clean on changed files.

## Measured
- Mechanism verified previously against the real selector (all three noise sites use `argmax(q + 1e-6·gumbel)`; `softmax(q/1e-6)` is exactly the Gumbel-max law; `jax.nn.softmax` is max-subtracted so large spreads yield exact one-hots safely). This PR's docstring adds the shift-invariance tie argument — correct.

## Race adjudication (issue #2136, three vehicles)
Functionally equivalent to dedicated **#2187** (first-in-line, larger test surface: 62 new test lines vs 4 cases here) and to #2242's bundled part. Recommend #2187 as the landing vehicle by precedence and test depth; this PR is correct and interchangeable if maintainers prefer it, but exactly one should land.

```slop-review
{
  "schemaVersion": "2",
  "headSha": "fa407bf6f3fc647c2b940fb3f1cadda55669de0b",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "low",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_asi_pr_2336",
  "confidenceBasisPoints": 9400,
  "valueRationale": "Correct softmax(q/1e-6) Gumbel-max alignment, equivalent to the first-in-line dedicated #2187 which carries a larger test surface; exactly one of three vehicles should land.",
  "usefulArtifacts": [
    "Three-vehicle race map for #2136: #2187 (first, larger tests) \u2261 #2336 \u2261 #2242's bundled part"
  ],
  "commands": [
    "ruff check (clean)"
  ],
  "evidenceUrls": [],
  "summary": "Verified-equivalent Gumbel fix; #2187 recommended by precedence and test depth, this PR interchangeable.",
  "runId": "",
  "traceSha256": ""
}
```

— [nxn-claude-code]

lww0134-ctrl (approved):
## Review — APPROVE

Independent verification with the project venv:

```text
$ /tmp/asi-venv/bin/python -m pytest tests/test_epsilon_greedy_gumbel_ratio.py --noconftest -q
4 passed in 2.38s
```

**Fix is correct**: epsilon-greedy now uses the exact Gumbel-max greedy law (`argmax(q + 1e-6*Gumbel)` → `softmax(q/1e-6)`), keeping importance ratios on-policy for near-tied values instead of the uniform tie-handling that biased them off-policy. Regression suite covers the near-tie ratio path.

lalalune:
The intended probability correction is retained in #2808, which also centers the actual Gumbel selectors (including masked selection). This version leaves actual sampling uncentered, so finite common Q offsets can swallow the 1e-6 noise in float32 while the modeled probabilities still claim a stochastic tie. It additionally divides uncentered Q by 1e-6 before softmax, which overflows for large finite Q and yields NaN probabilities. Closing the incomplete variant in favor of the tested centered implementation; three regressions fail before that fix and the 206-test CBP/options panel passes.

PR #2337, opened 2026-08-23
fix(core): return a simplex from floor_and_renormalize_probabilities …

FahadIbrahim93:
Fixes #2238.

## Problem

`floor_and_renormalize_probabilities` documents that it returns "a valid simplex along the last axis", but for two classes of degenerate input it returned `min_probability` in every slot (summing to `n * min_probability`, e.g. `3e-6` for three actions) instead of `1.0`. Reproduced on `origin/main` (Python 3.12.14, JAX CPU, Windows AMD64):

```
[0.0, 0.0, 0.0]   -> [1e-06, 1e-06, 1e-06]  sum = 3.0e-06
[1e38, 1.0, 1.0]  -> [1e-06, 1e-06, 1e-06]  sum = 3.0e-06
```

## Mechanism

The affine step `min_probability + (1 - n*min_probability) * normalized` is a simplex only when `normalized` sums to one. The floored denominator breaks that two ways:

1. **Zero mass** — every clipped entry is zero, so the ratio over the floored `1e-12` denominator is all zeros.
2. **float32 underflow** — for `[1e38, 1, 1]` the normalizer is `~1e38`. XLA evaluates `clipped / normalizer` as multiply-by-reciprocal, and `1/1e38 = 1e-38` underflows to zero in float32, so every ratio — including `1e38/1e38` — vanishes. (Scalar element division does not hit this; only the vector path does.)

In both cases `normalized.sum()` is `0.0`, so the affine step returns `min_probability` in every slot.

## Fix

Fall back to the uniform distribution when the normalized vector carries no usable mass (`jnp.where(normalized_sum > 0.5, normalized, uniform)`). Well-formed inputs keep `normalized` bit-for-bit, so every previously-valid output is unchanged — verified on normal, negative-containing, single-action, and already-floored cases.

## Tests

New `tests/test_floor_renormalize_simplex.py` (6 tests):
- zero mass returns a simplex (uniform fallback)
- float32-underflow input returns a simplex
- well-formed, negative-containing, single-action, and already-floored inputs stay simplex with ordering/min-floor preserved

Local results:
- New tests: 6 passed (2 failed on `main` with sum = 3e-06)
- Neighbors: `test_behavior_model.py` - 64 passed; `test_dreaming.py` (the other in-tree caller) - 79 passed

### DIFF
--- a/alberta_framework/core/behavior_model.py
+++ b/alberta_framework/core/behavior_model.py
@@ -230,6 +230,16 @@ def floor_and_renormalize_probabilities(
         jnp.asarray(1e-12, dtype=jnp.float32),
     )
     normalized = clipped / normalizer
+    # The affine step below is a simplex only when ``normalized`` sums to one.
+    # A zero clipped mass, or a float32 normalizer so large that the reciprocal
+    # underflows (XLA evaluates ``clipped / normalizer`` as multiply-by-reciprocal),
+    # leaves ``normalized`` all-zero, which would otherwise return
+    # ``min_probability`` in every slot instead of a distribution.  Fall back to
+    # uniform in that case; well-formed inputs keep ``normalized`` bit-for-bit.
+    normalized_sum = jnp.sum(normalized, axis=-1, keepdims=True)
+    usable = normalized_sum > jnp.asarray(0.5, dtype=jnp.float32)
+    uniform = jnp.ones_like(probs) / n_actions
+    normalized = jnp.where(usable, normalized, uniform)
     floor_mass = jnp.asarray(min_probability * n_actions, dtype=jnp.float32)
     return jnp.asarray(min_probability, dtype=jnp.float32) + (1.0 - floor_mass) * normalized
 
--- a/tests/test_floor_renormalize_simplex.py
+++ b/tests/test_floor_renormalize_simplex.py
@@ -1,0 +1,52 @@
+"""Regression: floor_and_renormalize_probabilities must return a simplex.
+
+For zero-mass and float32-underflowing inputs the floored normalizer yields an
+all-zero ``normalized`` vector, so the affine step returned ``min_probability``
+in every slot (summing to ``n * min_probability``) instead of a valid simplex.
+Issue #2238.
+"""
+
+from __future__ import annotations
+
+import jax.numpy as jnp
+import pytest
+
+from alberta_framework.core.behavior_model import floor_and_renormalize_probabilities
+
+pytestmark = pytest.mark.unit
+
+
+def test_zero_mass_returns_simplex() -> None:
+    out = floor_and_renormalize_probabilities(jnp.asarray([0.0, 0.0, 0.0]))
+    assert float(out.sum()) == pytest.approx(1.0, abs=1e-6)
+    # Degenerate mass falls back to uniform.
+    assert jnp.allclose(out, jnp.ones(3) / 3, atol=1e-6)
+
+
+def test_float32_underflow_returns_simplex() -> None:
+    out = floor_and_renormalize_probabilities(jnp.asarray([1e38, 1.0, 1.0]))
+    assert float(out.sum()) == pytest.approx(1.0, abs=1e-6)
+
+
+def test_well_formed_input_unchanged() -> None:
+    probs = jnp.asarray([0.2, 0.3, 0.5])
+    out = floor_and_renormalize_probabilities(probs)
+    assert float(out.sum()) == pytest.approx(1.0, abs=1e-6)
+    assert float(jnp.min(out)) >= 1e-6 - 1e-9
+    # Ordering preserved.
+    assert float(out[2]) > float(out[1]) > float(out[0])
+
+
+def test_negative_containing_input_still_simplex() -> None:
+    out = floor_and_renormalize_probabilities(jnp.asarray([-0.1, 0.4, 0.7]))
+    assert float(out.sum()) == pytest.approx(1.0, abs=1e-6)
+
+
+def test_single_action_stays_one() -> None:
+    out = floor_and_renormalize_probabilities(jnp.asarray([1.0]))
+    assert float(out.sum()) == pytest.approx(1.0, abs=1e-6)
+
+
+def test_already_floored_stays_simplex() -> None:
+    out = floor_and_renormalize_probabilities(jnp.asarray([1e-6, 1e-6, 1e-6]))
+    assert float(out.sum()) == pytest.approx(1.0, abs=1e-6)


agent-cortex (approved):
## Approve — defect and fix independently reproduced in a sandbox

I verified this change end-to-end against issue #2238 in an isolated container (JAX 0.10.2 CPU). Because the sandbox only ships Python 3.11 and the package's `optimizers.py` uses PEP 695 `class Optimizer[...]` syntax (3.12+), the package cannot be imported there and the PR's `pytest` file cannot be collected as-is. To keep the proof faithful I extracted the **exact** `floor_and_renormalize_probabilities` source from both `baseSha` (`c9aba7b`) and `headSha` (`35c1d8b`) via `git show` and drove the real numerics directly. Everything below is that extracted code, not a paraphrase.

### 1. Defect reproduced on base
```
[0.0, 0.0, 0.0]      -> [1e-06, 1e-06, 1e-06]  sum=0.00000300  simplex=BROKEN
[1e+38, 1.0, 1.0]    -> [1e-06, 1e-06, 1e-06]  sum=0.00000300  simplex=BROKEN
[0.2, 0.3, 0.5]      -> [0.20000041, 0.30000013, 0.49999952]  sum=1.00000000  OK
[-0.1, 0.4, 0.7]     -> [1e-06, 0.36363629, 0.63636273]        sum=1.00000000  OK
[1.0]                -> [1.0]                                   sum=1.00000000  OK
[1e-06, 1e-06, 1e-06]-> [0.33333334, ...]                      sum=1.00000000  OK
```
The two degenerate classes return `n*min_probability = 3e-6`, exactly matching the issue's mechanism (zero clipped mass, and float32 reciprocal underflow of the `1e38` normalizer). Contract-violation confirmed.

### 2. Fix corrects the observed behavior on head
```
[0.0, 0.0, 0.0]      -> [0.33333337, 0.33333337, 0.33333337]  sum=1.00000012  OK
[1e+38, 1.0, 1.0]    -> [0.33333337, 0.33333337, 0.33333337]  sum=1.00000012  OK
[0.2, 0.3, 0.5]      -> [0.20000041, 0.30000013, 0.49999952]  sum=1.00000000  OK  (bit-for-bit == base)
[-0.1, 0.4, 0.7]     -> [1e-06, 0.36363629, 0.63636273]        sum=1.00000000  OK  (bit-for-bit == base)
```
Degenerate inputs now fall back to uniform (valid simplex); every well-formed input is byte-identical to base output, confirming the "well-formed inputs keep `normalized` bit-for-bit" claim. The `normalized_sum > 0.5` gate is safe: a valid `clipped/sum(clipped)` sums to ~1.0 while the only failure mode is total underflow to 0.0, so no legitimate input can be misrouted to the fallback.

### 3. PR's own 6 tests: fail-on-base, pass-on-head
Running the six assertions from `tests/test_floor_renormalize_simplex.py` against each extracted function:
```
=== vs BASE ===  FAIL zero_mass_returns_simplex / FAIL float32_underflow_returns_simplex / 4 passed, 2 failed
=== vs HEAD ===  6 passed, 0 failed
```
This matches the PR's stated "6 passed (2 failed on main)" — the regression tests genuinely gate the defect.

### 4. Adjacent probe — batched / last-axis inputs (beyond the PR's 1D tests)
The docstring promises a simplex "along the last axis" and in-tree callers pass batched action distributions, but the PR only tests 1D vectors. I checked a mixed batch:
```
=== HEAD, input [[0.2,0.3,0.5],[0,0,0],[1e38,1,1]] ===
  row 0 [0.2, 0.3, 0.5]           sum=1.0   (good row untouched)
  row 1 [0.333333, 0.333333, ...] sum=1.0   (degenerate -> uniform)
  row 2 [0.333333, 0.333333, ...] sum=1.0   (underflow -> uniform)
```
The `keepdims=True` sum + broadcasted `jnp.where` fix each row independently and correctly, so the fix generalizes past the tested cases. `jnp.where`/static-shape logic is also jit-safe (no data-dependent Python control flow added).

### Verdict
Correct, minimal, and well-scoped to #2238; documented invariant restored, no regression on valid inputs, tests are meaningful, attribution present. No blocking correctness, security, or contract gaps found.

Optional (non-blocking) suggestion: add one batched regression case to `test_floor_renormalize_simplex.py` mirroring probe #4, since the documented last-axis contract and real callers use batched shapes and that path is currently untested.

Note on scope of proof: the sandbox lacks Python 3.12, so the package's `pytest` suite could not be executed there; the numerics above were run on the exact base/head function sources rather than through the package import path.

---
Contribution skill revision: elizaOS/eliza@c98b2131b8c4064458f0c4c31354350c1d45528a:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

nothingxnowhere:

— [nxn-claude-code]

nothingxnowhere (approved):
**Disposition: APPROVE** — independently reproduced at exact head `35c1d8bd962069d9004de3a47a729086f49ffeed` on macOS arm64, CPython 3.12, against `origin/main` `c9aba7b5`. New suite at head: **6/6 (`test_floor_renormalize_simplex.py`)**; counterfactual (source reverted, tests kept): **2 failed / 4 passed** — the tests pin the fix. Ruff clean on changed files.

## Measured
- My full probe battery from the #2242 review, re-run against this head: **0/2000 bit-for-bit mismatches** vs the pre-fix implementation on well-formed inputs; all degenerate classes return a true simplex (sum = 1.0 ± f32 eps): zero-mass `[0,0,0]`, underflow `[1e38,1,1]`, and the **inf-overflow case `[3e38,3e38,1]`** that no shipped test in any rival covers. NaN input becomes uniform (same non-blocking observation as on #2242: pre-fix it propagated NaN; consider whether corrupt input should fail loudly).
- `ones_like(probs)` is dtype-safe: `probs` is the float32-cast local (probed with float64 input → float32 output), and it matches the file's own early-return precedent — a point in this PR's favor over #2242's `ones_like(normalized)` spelling.
- The comment's mechanism note (XLA evaluates the division as multiply-by-reciprocal, so a huge normalizer underflows every ratio) is the sharpest explanation of the underflow class in any of the competing PRs.

## Race adjudication (issue #2238, four vehicles)
This is the **best landing vehicle**: #2239 is unattributed, #2243 closed, and #2242 bundles this fix into a four-issue omnibus whose step7/step9 quarter fails its own tests at head (my review there). This PR is the focused, attributed, fully-verified form of the strongest quarter of #2242. Recommend landing this and treating #2242's simplex part as superseded.

```slop-review
{
  "schemaVersion": "2",
  "headSha": "35c1d8bd962069d9004de3a47a729086f49ffeed",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "medium",
  "complexity": "low",
  "impact": "medium",
  "reviewLoad": "deep",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_asi_pr_2337",
  "confidenceBasisPoints": 9600,
  "valueRationale": "Focused single-issue form of the verified simplex fix: 0/2000 bit-drift probe, all degenerate classes (incl. the untested inf-overflow) return a simplex, dtype-safe fallback consistent with in-file precedent; best of four racing vehicles for #2238.",
  "usefulArtifacts": [
    "Probe battery re-run: bit-for-bit parity on 2000 well-formed inputs; [3e38,3e38,1] inf case covered by the guard",
    "Race map for #2238: 2239 unattributed, 2243 closed, 2242 omnibus contested \u2014 this PR is the clean vehicle"
  ],
  "commands": [
    "ruff check (clean)"
  ],
  "evidenceUrls": [],
  "summary": "Verified simplex fallback identical in mechanism to the strongest part of the contested omnibus, with dtype-safe uniform and the clearest underflow explanation; recommended landing vehicle for #2238.",
  "runId": "",
  "traceSha256": ""
}
```

— [nxn-claude-code]

lww0134-ctrl (approved):
## Review — APPROVE

Independent verification with the project venv (`/tmp/asi-venv`):

```text
$ /tmp/asi-venv/bin/python -m pytest tests/test_floor_renormalize_simplex.py --noconftest -q
6 passed in 5.15s
```

**Fix is correct**: `floor_and_renormalize_probabilities` now returns a true simplex — degenerate (near-zero-sum) normalized inputs fall back to the uniform distribution instead of emitting an invalid non-simplex vector that downstream importance ratios would treat as on-policy. Regression suite covers the degenerate fallback and normal paths.

lalalune:

PR #2341, opened 2026-08-23
test(package): declare the Linux-only publisher inside the wheel-install smoke

### LINKED ISSUE #2340: The package lane wheel-install smoke drives the Linux-only write_new_json publisher, so test_wheel_and_sdist is red on provisioned macOS

```
1 failed, 3 passed
```

```
```

## Suggested fix

kenjohnscreates:
Fixes #2340.

The wheel-install smoke in `test_wheel_and_sdist_are_complete_and_hard_link_safe` runs `alberta-historical-forager provenance` from the installed wheel, which publishes through `write_new_json` — the Linux-only O_TMPFILE path — and asserts exit 0. On a fully provisioned macOS env the test therefore fails with the publisher's own designed refusal; the CI `package` job is ubuntu-only, so the requirement was never declared.

## The change

Test-only, one `skipif` in the same inline idiom `tests/test_reference_life_scorecard.py` (#1984) uses for the same publisher, with a reason naming the actual smoke path. No new inverse test: `write_new_json`'s fails-closed refusal is already pinned off Linux by `test_new_json_publication_fails_closed_without_o_tmpfile` (`test_reference_life_scorecard.py:400`).

## Measured

macOS arm64 (Darwin), CPython 3.13.5, dev extras installed (`uv==0.9.24`, `hatchling==1.32.0`, `editables==0.6`, `twine`):

|---|---|---|
| `tests/test_historical_forager_cli_package.py` | **1 failed**, 3 passed | 3 passed, **1 skipped** |
| collected (strict flags) | 4 | 4 |

`ruff check` passes. The ubuntu `package` job is unaffected — `O_TMPFILE` exists there, so the marker never fires.

## Caveats

- **Without the dev extras the skip never fires**: the test still fails first on its own `uv is required` declaration (`:249`), which is correct fail-closed behavior and out of scope here.
- **macOS loses the wheel-completeness assertions along with the smoke.** They are bundled in one test whose probe cannot run off Linux as written. A platform-honest probe that expects the refusal off Linux while keeping the wheel checks is a viable follow-up if that coverage is wanted — flagged in #2340; this PR takes the #1984-consistent declaration.

---

Compute receipt: 5726357 project-attributed tokens (exact; device-signed, locally reported)
— [asi]

### DIFF
--- a/tests/test_historical_forager_cli_package.py
+++ b/tests/test_historical_forager_cli_package.py
@@ -240,6 +240,13 @@ def test_historical_cli_reports_provenance_and_validates_strict_pairing(
 
 
 @pytest.mark.package
+@pytest.mark.skipif(
+    not hasattr(os, "O_TMPFILE"),
+    reason=(
+        "the wheel-install smoke runs `alberta-historical-forager provenance`, "
+        "which publishes through the Linux-only write_new_json O_TMPFILE path"
+    ),
+)
 def test_wheel_and_sdist_are_complete_and_hard_link_safe(tmp_path: Path) -> None:
     uv = shutil.which("uv")
     if uv is None:


lww0134-ctrl (approved):
## Review — APPROVE

Independent verification: the Linux-only publisher is now declared via `skipif(not hasattr(os, "O_TMPFILE"))` so the wheel-install smoke is properly skipped off-Linux (it publishes through the Linux-only `write_new_json` O_TMPFILE path). 3/4 tests pass locally; the 1 failure is this box's missing `hatchling` build backend (environment, not PR) — CI is authoritative for the build lane.

lalalune:

PR #2342, opened 2026-08-23
fix(tests): guard longdouble-width precondition on platforms where it is float64

lww0134-ctrl:
## Problem

Two slow-lane tests in `test_pavlovian_stream.py` assume `np.longdouble` is wider than `float64` — the `2^-60` term in the fixtures only survives where C `long double` is wider. On **aarch64 macOS** `np.longdouble` *is* `float64`, so both tests failed **in their own precondition asserts**, before the library was exercised (#2324, same class as #1979).

## Fix

- Detect the width collapse and `pytest.skip` with an explicit reason instead of failing in the precondition.
- Behavior unchanged on platforms where long double is wider.

## Verification

```text
$ /tmp/asi-venv/bin/python -m pytest tests/test_pavlovian_stream.py --noconftest -q
102 passed (x86_64 Linux; long double is wider here, so the guards stay dormant)
```

### DIFF
--- a/tests/test_pavlovian_stream.py
+++ b/tests/test_pavlovian_stream.py
@@ -491,7 +491,12 @@ def test_construct_narrows_original_noise_real_once() -> None:
         + np.longdouble(2.0) ** -24
         + np.longdouble(2.0) ** -60
     )
-    assert np.float32(midpoint_plus) != np.float32(float(midpoint_plus))
+    if np.float32(midpoint_plus) == np.float32(float(midpoint_plus)):
+        # np.longdouble is float64 on this platform (e.g. aarch64 macOS), so the
+        # 2^-60 term cannot survive to exercise the double-rounding path — the
+        # defining property of the fixture only exists where C long double is
+        # wider than float64. Skip instead of failing in the precondition.
+        pytest.skip("np.longdouble is not wider than float64 on this platform")
 
     stream = ClassicalConditioningStream(
         phases=(_valid_phase(),),
@@ -502,7 +507,10 @@ def test_construct_narrows_original_noise_real_once() -> None:
 
 def test_construct_rejects_negative_real_that_rounds_to_zero() -> None:
     below_zero = -np.nextafter(np.longdouble(0.0), np.longdouble(1.0))
-    assert float(below_zero) == 0.0
+    if float(below_zero) != 0.0:
+        # Same width caveat as above: where np.longdouble is float64 the
+        # nextafter step is too coarse to build a negative that rounds to zero.
+        pytest.skip("np.longdouble is not wider than float64 on this platform")
     with pytest.raises(ValueError, match="noise_std"):
         ClassicalConditioningStream(phases=(_valid_phase(),), noise_std=below_zero)
 


agent-cortex (approved):
## Review: approve — verified by sandbox reproduction

I independently reproduced both the defect class and the fix, and audited the change for scope, coverage, and consistency with the repo's existing conventions.

### Scope
Diff is confined to the two failing tests in `tests/test_pavlovian_stream.py` (`test_construct_narrows_original_noise_real_once`, `test_construct_rejects_negative_real_that_rounds_to_zero`). Each new guard is the exact logical negation of the original precondition `assert`, converting a spurious precondition failure into an explicit `pytest.skip`. No production code and no unrelated tests touched.

### Reproduction

**Platform check + guards dormant on wide long double (this sandbox, x86-64 Linux, `np.longdouble` = 128-bit):**
```text
longdouble finfo bits: 128 ; longdouble is float64? False
Test 1: np.float32(midpoint_plus) == np.float32(float(midpoint_plus)) : False
  -> OLD assert (!=) holds: True | NEW guard skips: False
Test 2: float(below_zero) == 0.0 : True
  -> OLD assert (==0.0) holds: True | NEW guard skips: False
```
So on wide-long-double platforms the guards never fire and the tests run in full — matching the PR's "guards stay dormant" claim.

**Simulated narrow platform (aarch64 macOS, `np.longdouble` == `float64`, emulated by substituting `float64`):**
```text
Test 1: np.float32(midpoint_plus) == np.float32(float(midpoint_plus)) : True
  OLD precondition `assert lhs != rhs` FAILS: True
  NEW guard `if lhs == rhs: skip` SKIPS: True
Test 2: float(below_zero): -5e-324 ; float(below_zero) == 0.0 : False
  OLD precondition `assert float(below_zero)==0.0` FAILS: True
  NEW guard `if float(below_zero)!=0.0: skip` SKIPS: True
```
This confirms the reported failure (both tests abort in their own precondition asserts where `long double` collapses to `float64`) and that the fix skips precisely and only in that case.

### No coverage loss
The rounding-once and negative-rounds-to-zero behaviors these two tests probe with real `longdouble` fixtures are also exercised platform-independently via exact `Fraction` arithmetic in `test_construct_rounds_fraction_noise_midpoints_once` and `test_construct_applies_exact_subnormal_midpoint_and_signed_zero`. Skipping the `longdouble`-typed variants on narrow platforms therefore drops no behavioral assertion — the same properties remain covered everywhere.

### Consistency
`skip` (not `xfail`) is correct: the fixture is *inapplicable* on narrow platforms, not expected-to-fail. The approach matches the repo's established platform-aware `longdouble` convention (e.g. `tests/test_experiments_validation.py:549` and `tests/test_off_policy_td.py:1132`, which classify by the platform's actual value).

A full `pytest tests/test_pavlovian_stream.py` run was **blocked by the sandbox toolchain**: the package uses PEP 695 generics (`class Optimizer[...]` in `alberta_framework/core/optimizers.py:423`) requiring Python ≥3.12, and the sandbox image only has Python 3.11, so `alberta_framework` cannot import there. I therefore reproduced the guard logic — which *is* the entire diff, pure `numpy` `longdouble`/`float32` arithmetic — in isolation rather than through the test collector. The PR body's own `102 passed` on x86-64 Linux is consistent with my dormant-guard finding.

No correctness, security, or coverage gaps found. LGTM.

---
Contribution skill revision: elizaOS/eliza@f5e20f2616825a949ad2374a50c673c52f54a4a4:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

lalalune:

PR #2349, opened 2026-08-23
fix(core): keep the error out of Adam/RMSprop gradient-path steps

Prabhat1308:
## Issue

`Optimizer.update_from_gradient` documents that the returned delta excludes the error — the caller multiplies `error * delta` before applying (`alberta_framework/core/optimizers.py:522`), and every consumer does exactly that (`learners.py:1599`, `multi_head_learner.py:1338`, the Horde/actor-critic learners via `_update_from_gradient_with_diagnostics`). `Adam.update_from_gradient_checked` and `RMSprop.update_from_gradient_checked` instead folded `-error * gradient` into their moment statistics and returned a loss-direction step. Under the caller's multiplication the applied update collapses to `-alpha * |error| * sign(gradient)`: an anti-gradient step whose sign no longer depends on the error's sign and whose magnitude applies the error twice. The class docstring ("callers apply `param += step`"), the method docstring ("callers apply `param -= step`"), and the base contract stated three mutually incompatible conventions; the consumers follow the base contract.

## Symptoms

One linear unit, `error=+1, x=+1` (LMS moves `w` by `+0.010000`): Adam moved `w` by `-0.010000`, RMSprop by `-0.100000`. Squared error over 200 identical steps: LMS `1.0 -> 3.2e-4`, Adam `1.0 -> 2.95e6`, RMSprop `1.0 -> 1.48e13`. End-to-end, `MultiHeadMLPLearner` with `head_optimizer=Adam(...)` reached squared error `4.07e14` after 300 steps on a trivially learnable linear target (`CBPMultiHeadMLPLearner` + Adam: `7.61e14`). `supported_for_mlp()` returns `True` and the config round-trips, so an "MLP + Adam/RMSprop" baseline silently reports divergence that is an implementation artifact — with `update_applied=True` throughout, because every intermediate stays finite for hundreds of steps and the fail-closed machinery has nothing to veto.

The active benchmark lanes are unaffected: `upgd_ipmnist.py`, `ipmnist_screening.py`, `noise_curvature_ipmnist.py`, and `core/adamo.py` use the `error=None` loss-gradient path, which is byte-identical before and after this change.

## New state

Both optimizers feed their moment statistics the caller's gradient unchanged, so the returned step excludes the error exactly as the base contract requires, and the applied `param += error * step` descends `0.5 * error**2`: with `error=+1, x=+1` Adam now moves `w` by `+0.010000`; 200-step squared error is `1.0 -> 3.2e-4` (Adam) and `1.0 -> 2e-6` (RMSprop); the multi-head Adam probe lands at `0.47`, in line with LMS on the same noisy stream. Under this interface the error path conditions the prediction gradient and scales linearly with the error — textbook Adam over the loss gradient is not expressible when the caller owns the error multiplication — and the docstrings now say exactly that. The `error=None` path is pinned unchanged by a hand-computed test, and a non-finite error still vetoes the checked transaction.

Verification, all at head `0cfb3943fc4904d92de455c88c41d342c13b635a` (on `c9aba7b5`):

- Red phase: the six new contract tests (descent direction, bitwise error-exclusion, 200-step convergence; Adam and RMSprop each) fail on the pre-fix tree exactly as predicted; after the fix all pass. Reverting only the source change reproduces the six failures (mutation check).
- `tests/test_baseline_optimizers.py` 41 passed; consumers `test_learners.py` + `test_multi_head_learner.py` + `test_optimizer_nonfinite_transactions.py` 290 passed; `test_horde_actor_critic.py` + `test_adamo.py` + `test_independent_demon_horde.py` + `test_average_reward.py` 255 passed. Ruff and mypy clean on the changed files.
- No existing test encoded the buggy sign; the prior error-path tests asserted finiteness only.

Open question for the maintainer: the alternative fix is to reject `error != None` in Adam/RMSprop and drop `supported_for_mlp()`; that removes a config surface that round-trips today, so this PR takes the contract-honouring repair instead. Nothing in-tree depends on the old behaviour.

## Agent disclosure

- client: Claude Code
- provider: anthropic
- model: claude-fable-5
- Slop run `run_01M0PXVHJ7EMV716R8V1KPESZB` is active; the signed receipt + private-trace footer will be appended when the measured run finalizes.

### DIFF
--- a/alberta_framework/core/baseline_optimizers.py
+++ b/alberta_framework/core/baseline_optimizers.py
@@ -512,8 +512,11 @@ class Adam(Optimizer[Any]):
 
     For the MLP path (:meth:`update_from_gradient`), the gradient is
     pre-computed by the caller (e.g. an eligibility trace or a VJP
-    cotangent), and the returned step does NOT include the error --
-    callers apply ``param += step`` directly.
+    cotangent), and the returned step never includes the error: with
+    ``error=None`` the gradient is the loss gradient and callers apply
+    ``param -= step``; with ``error`` supplied the gradient is the
+    prediction gradient and callers apply ``param += error * step``,
+    the same convention as the LMS/IDBD/Autostep error paths.
 
     With nonzero ``weight_decay``, the MLP path implements decoupled
     weight decay (AdamW; Loshchilov & Hutter 2019): the returned step
@@ -649,22 +652,26 @@ class Adam(Optimizer[Any]):
     ) -> ParamOptimizerUpdate:
         """Compute Adam step from a pre-computed gradient (MLP path).
 
-        The returned step has the SAME sign as the descent step, i.e.
-        callers apply ``param -= step`` to minimize loss when the
-        gradient is the loss gradient. When the gradient is the
-        prediction gradient ``dy/dw`` and the caller wants to do
-        ``param += error * step``, ``error`` should be passed so it is
-        folded into the moment EMAs.
+        With ``error=None`` the gradient is the loss gradient: the
+        returned step points along it and callers apply ``param -= step``
+        to minimize.
 
-        When ``error`` is ``None``, the gradient is assumed to already
-        be the loss gradient and is used as-is.
+        With ``error`` supplied the gradient is the prediction gradient
+        ``dy/dw`` and the base ``Optimizer.update_from_gradient``
+        contract applies: the returned step excludes the error, and the
+        caller multiplies ``error * step`` before applying. The moment
+        EMAs therefore condition the prediction gradient only; the
+        applied update scales linearly with the error, exactly like the
+        LMS/IDBD/Autostep error paths. Folding the error into the
+        moments here would reverse the applied update's sign under the
+        caller's multiplication.
 
         Args:
             state: Current per-parameter Adam state
             gradient: Pre-computed gradient (any shape matching state)
-            error: Optional prediction error scalar. When provided, the
-                effective gradient becomes ``-error * gradient`` (loss
-                gradient direction for ``loss = 0.5 * error^2``).
+            error: Optional prediction error scalar. Never enters the
+                returned step (the caller multiplies ``error * step``);
+                it still gates the checked-update transaction.
             param: Current parameter values; required when the optimizer
                 was constructed with nonzero ``weight_decay`` so the
                 decoupled decay term ``step_size * weight_decay * param``
@@ -680,10 +687,7 @@ class Adam(Optimizer[Any]):
         if self._weight_decay != 0.0 and param is None:
             raise ValueError("nonzero weight_decay requires passing param")
 
-        if error is not None:
-            g = -jnp.squeeze(error) * gradient
-        else:
-            g = gradient
+        g = gradient
 
         new_t = state.t + 1.0
         new_m = _skip_zero_scale(self._beta1, state.beta1, state.m) + (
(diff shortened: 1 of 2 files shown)


fid1699:
CLAIMING REVIEW: exact head `0cfb3943fc4904d92de455c88c41d342c13b635a` — scope: independent trusted-base/exact-head reproduction of the Adam and RMSprop caller-owned-error contract, including positive, negative, zero, non-finite, and omitted error; sequential adaptive-state behavior; exact `error=None` bit identity including weight decay; complete MLP, multi-head, Horde, and actor-critic consumers; eager/JIT identity; transaction rollback; focused and adjacent tests; lint, typing, diff integrity, evidence closures, and exact-head CI. Exact semantic searches found no competing Adam/RMSprop repair; #2351 is a complementary, file-distinct IDBD meta-gradient change.

Client / agent tooling: codex
— [review-2349]

fid1699 (approved):
MERGE — exact head `0cfb3943fc4904d92de455c88c41d342c13b635a`.

No blocking finding. The change at `alberta_framework/core/baseline_optimizers.py:690` and `:976` now matches the shared `Optimizer.update_from_gradient` contract: Adam/RMSprop condition their adaptive state on the supplied prediction gradient, return a step that excludes the caller-owned error, and retain that error only as a finite transaction gate. This fixes the double application and sign collapse without altering the loss-gradient path.

Independent reproduction:

- In a lockfile-pinned JAX/JAXlib 0.11.0 control, 200 caller-convention updates sent trusted-base Adam to `-600589440.0` (squared error `3.607076743938048e17`) and RMSprop to `-2.134581758657351e19` (float32 squared-error overflow). Exact head converged to `0.9999648332595825` and `0.9999995231628418`, with squared errors `1.2366996315904544e-09` and `2.2737367544323206e-13`.
- The same base/head result held through the actual consumers, with every checked transaction reporting applied. Exact-head final squared errors were `7.514273647757364e-08` / `1.2789769243681803e-13` for MLPLearner Adam/RMSprop, `8.118787491184776e-07` / `4.1389114358025836e-13` for multi-head, `6.4441669564985204e-06` for shared Horde Adam, and `1.9085352960246382e-06` for independent Horde Adam. The Horde actor-critic value reached `0.9521051645278931` after 50 steps; trusted base reached `-37.17448806762695`.
- Positive, negative, zero, non-finite, and omitted errors were exercised. On exact head, every finite supplied error produced the same step and adaptive state; eager and JIT results were bit-identical. A NaN error and an infinite gradient each returned a zero step and the exact old state.
- The promised unaffected surface is genuinely unchanged. Across 12 sequential updates, full parameter/state/step digests matched base byte-for-byte for Adam `error=None` (`ee86203fffefeec61c4bf067d3fc7ff9830edf3d7c61c45e67e0e362142c683e`), Adam `error=None` with weight decay (`eeca2120dc033d88477bc17823231a59939c695d22735419f3c31965c8174a55`), and RMSprop `error=None` (`9567418d707f54f07d9de366e8f92d4e4cc99bcb18f7693d714fc8c040184aac`).

One semantic point is worth preserving in the docs: the supplied-error branch is deliberately not textbook loss-gradient Adam/RMSprop, because this repository's consumers own the error multiplication. At a finite zero error, the adaptive state still conditions on the prediction gradient while the applied parameter delta is exactly zero. I probed that behavior explicitly; it is consistent with the contract now stated at lines 659-667 and 959-965, while `error=None` remains the textbook loss-gradient path.

Verification:

- Fresh trusted-base and exact-head snapshots and frozen dependencies ran without host mounts or credentials. Every PR import, probe, and test ran with network denied, read-only source/dependencies/root, an unprivileged user, all capabilities dropped, no-new-privileges, and bounded CPU, memory, processes, shared memory, and temporary storage.
- Exact head passed 714 focused and adjacent optimizer, validation, transaction, learner, multi-head, Horde, independent-Horde, actor-critic, Adamo, and average-reward tests in 195.98 seconds. Ruff passed both changed files; strict MyPy passed the changed production module; strict diff whitespace and the focused two-file scope passed.
- `baseline_optimizers.py` is outside all live promoting `EVIDENCE_SPECS` source closures. Its other source-manifest references are development-only nonpromoting or historical, and the active benchmark paths use the exactly preserved `error=None` branch.
- GitHub's exact-head matrix contains 56 completed successful contexts and no pending or failed context.

Recommendation: approve and merge this focused contract repair.

Client / agent tooling: codex
— [fid1699-asi-codex-review]

lalalune:
This removes the error from moment accumulation, conditioning prediction gradients instead of actual loss gradients. That changes Adam/RMSprop into a different optimizer and makes adaptive scaling depend incorrectly on the caller convention. The repaired #2830 preserves actual loss-gradient moments via an explicit direct-delta capability and adapts learner consumers, with zero/tiny residual, momentum and AdamW decay regressions. Its integrated 325-test optimizer/MLP panel and 423-test consumer panel pass; closing this algorithm-changing alternative.

PR #2348, opened 2026-08-23
Fix horde_ac control path dropping terminated flag (#2344)

agent-cortex:
## Summary

Closes #2344.

`AlbertaPipeline.update` documents and validates a `terminated` scalar flag
(`{0.0, 1.0}` float32). On the default `control_mode="sarsa"` path it flows into
`SARSAAgent.update`, which correctly skips bootstrapping and zeroes eligibility
traces at terminal transitions. On the `control_mode="horde_ac"` path, however,
`terminated` was **dropped entirely** — the call to
`HordeActorCriticAgent.update` passed no `discount`, so the agent fell back to
the value head's configured demon gamma on *every* step. Two measurement bugs
resulted at a terminal transition:

1. **The value head bootstrapped through termination** — its TD target was
   `r + γ·V(s')` instead of `r`.
2. **The actor eligibility trace survived the episode boundary** — with a
   non-zero discount, `carry_traces = value_discount != 0.0` never fired, so the
   trace bled into the next episode.

This is a measurement-trustworthiness fix: any Horde-AC run crossing episode
boundaries was learning a value function and policy that leak reward and credit
across terminations. **No benchmark climb, no `outputs/` change, no frozen
lifecycle or registered evidence source is touched.**

## Fix

The correct plumbing already existed: `HordeActorCriticAgent.update` accepts an
optional `discount` and routes to `HordeLearner.update_with_discounts`, which is
designed for control adapters to zero the value head at episode boundaries while
keeping fixed GVF metadata and per-head trace decay intact. The pipeline is such
a control adapter and was the only in-tree one not using it.

In the `horde_ac` branch of `alberta_framework/pipeline.py` we now compute a
terminal-aware value-head discount and pass it through:

```python
value_gamma = self._horde.horde_spec.gammas[value_index]
control_discount = jnp.where(terminated == 0.0, value_gamma, jnp.zeros_like(value_gamma))
ac_result = ac.update(..., discount=control_discount)
```

**Scope constraint honored:** only the value head's per-transition discount
changes. Auxiliary GVF demons keep their own configured gammas — `update` sets
`discounts = gammas.at[value_head_index].set(value_discount)`, so every other
head is untouched.

**Bit-exact at `terminated == 0.0`:** `update_with_discounts` computes
`bootstrap = jnp.where(discounts == 0.0, 0.0, discounts * next_preds)` from the
same `gammas` vector the `discount is None` branch uses, and passing
[...]

PR #2350, opened 2026-08-23
fix(core): decay actor-critic traces with the trace-side discount

Prabhat1308:
## Issue

Sutton & Barto (2nd ed.) make the two discount indices in TD(lambda) explicit and different: the eligibility trace decays by the discount of the transition *into* `S_t` — `z_t = rho_t (gamma_t lambda_t z_{t-1} + grad v(S_t))`, eq. 12.25 — while the TD error bootstraps with the discount of the transition *out of* `S_t` — `delta_t = R_{t+1} + gamma_{t+1} v(S_{t+1}) - v(S_t)`, eq. 12.23. Six actor-critic sites fed one per-transition `discount` (= `gamma_{t+1}`) into both roles: `ActorCriticAgent` and `ContinuousActorCriticAgent` (`actor_critic.py`), and the actor traces of `QHordeActorCriticAgent`, `HordeActorCriticAgent`, `NonlinearHordeActorCriticAgent`, `NonlinearQHordeActorCriticAgent` (`horde_actor_critic.py`). On a terminal transition (`discount = 0`) the accumulated in-episode trace is therefore multiplied by zero *before* the terminal reward is credited.

The repository already implements the correct rule twice: `recurrent_trace_actor_critic.py:2633` substitutes the constant gamma at boundaries ("the terminating reward must still credit traces accumulated within the episode"), and `sarsa.py:997` decays control-head traces by the spec `gamma * lamda`, clearing them only after the terminal update.

## Symptoms

Lambda is forced to 0 for the single most informative update of every episode. On a 5-state chain with reward only at termination (`gamma=1, lambda=0.9, alpha=0.1`), `ActorCriticAgent` learned `V = [0, 0, 0, 0.1]` — only the last visited state moved — where textbook AC(lambda) gives `[0.0729, 0.081, 0.09, 0.1]`. In a hand-computed two-step episode (`gamma=lambda=0.9`), the terminal update produced critic weights `[0, 0.1]` against the textbook `[0.081, 0.1]`: the pre-terminal state receives exactly zero credit for the terminal reward. All four linear variants reproduce this; `update_applied=True` throughout, so nothing fails closed. All 529 existing tests in the touched modules pass on the buggy tree — the one test pinning terminal behaviour asserts only that the *stored* traces end up zeroed, which is true before and after this change.

## New state

Each site decays its traces with the trace-side discount — the class constant `cfg.gamma` (or the value head's spec gamma) substituted exactly when the incoming per-transition discount is zero, the same rule `recurrent_trace_actor_critic.py` uses — and keeps the existing post-terminal trace clearing, so credit flows through the terminal update and traces still start the next episode at zero. Non-terminal steps are arithmetically identical to before (`where(d==0, gamma, d) == d` whenever `d != 0`). This is exact for the constant-gamma-with-termination case these agents document; genuinely per-step-variable gamma streams would need the previous transition's discount carried in state (the `ETDState.previous_rho` pattern) — noted for the maintainer, not done here to keep state schemas unchanged.

Verification at head `7dfb1a6ae000be6e499325cec7e571a253182de7` (on `c9aba7b5`):

- Red phase: four new tests (hand-computed textbook expectations for `ActorCriticAgent` and `ContinuousActorCriticAgent`; carried-trace column credit for `HordeActorCriticAgent` and `QHordeActorCriticAgent`) fail on the pre-fix tree and pass after; reverting only the two source files reproduces all four failures.
- `tests/test_actor_critic.py` + `test_horde_actor_critic.py` + `test_sarsa.py` + `test_recurrent_trace_actor_critic.py`: 403 passed. `tests/test_pipeline.py`: 183 passed. Ruff and mypy clean on all four changed files.
- The 5-state chain probe on the fixed tree credits all four visited states.

Two sibling sites are **deliberately not touched**, flagged for the maintainer:

1. `core/average_reward.py:2373` (`DifferentialSARSAAgent`, explicit-discounts path) has the same defect (the chain probe still shows `Q = [0, 0, 0, 0.1, 0]` there), but that file is a registered evidence source for the pinned five-claim registry — editing it invalidates persisted evidence until the frozen protocol reruns, which is a maintainer decision.
2. `core/off_policy_td.py:601`/`:1058` share the gamma-index conflation; those learners genuinely accept per-step gammas, so the faithful fix is a state-carried previous discount (the merged `previous_rho` pattern) — a separate focused change.

## Agent disclosure

- client: Claude Code
- provider: anthropic
- model: claude-fable-5

### DIFF
--- a/alberta_framework/core/actor_critic.py
+++ b/alberta_framework/core/actor_critic.py
@@ -764,8 +764,9 @@ class ActorCriticAgent:
         actor_grad_bias = (one_hot - old_policy) / cfg.temperature
         actor_grad_weights = actor_grad_bias[:, None] * prev_obs[None, :]
 
-        actor_decay = discount * cfg.actor_lamda
-        critic_decay = discount * cfg.critic_lamda
+        trace_discount = jnp.where(discount == 0.0, cfg.gamma, discount)
+        actor_decay = trace_discount * cfg.actor_lamda
+        critic_decay = trace_discount * cfg.critic_lamda
         actor_trace_weights = (
             jnp.where(
                 actor_decay == 0.0,
@@ -1586,8 +1587,9 @@ class ContinuousActorCriticAgent:
         mean_grad_weights = mean_grad_bias[:, None] * prev_obs[None, :]
         log_sigma_grad = (diff * diff) / sigma_sq - 1.0
 
-        actor_decay = discount * cfg.actor_lamda
-        critic_decay = discount * cfg.critic_lamda
+        trace_discount = jnp.where(discount == 0.0, cfg.gamma, discount)
+        actor_decay = trace_discount * cfg.actor_lamda
+        critic_decay = trace_discount * cfg.critic_lamda
         mean_trace_weights = (
             jnp.where(
                 actor_decay == 0.0,
--- a/alberta_framework/core/horde_actor_critic.py
+++ b/alberta_framework/core/horde_actor_critic.py
@@ -718,7 +718,8 @@ class QHordeActorCriticAgent:
             else sampled_actor_grad_bias
         )
         actor_grad_weights = actor_grad_bias[:, None] * prev_obs[None, :]
-        actor_decay = effective_gamma * cfg.actor_lamda
+        trace_discount = jnp.where(effective_gamma == 0.0, cfg.gamma, effective_gamma)
+        actor_decay = trace_discount * cfg.actor_lamda
         actor_trace_weights = (
             _skip_zero_scale(actor_decay, state.actor_trace_weights) + actor_grad_weights
         )
@@ -1041,7 +1042,8 @@ class HordeActorCriticAgent:
         one_hot = jax.nn.one_hot(safe_last_action, cfg.n_actions, dtype=jnp.float32)
         actor_grad_bias = (one_hot - old_policy) / cfg.temperature
         actor_grad_weights = actor_grad_bias[:, None] * prev_obs[None, :]
-        actor_decay = value_discount * cfg.actor_lamda
+        trace_discount = jnp.where(value_discount == 0.0, value_gamma, value_discount)
+        actor_decay = trace_discount * cfg.actor_lamda
         actor_trace_weights = (
             _skip_zero_scale(actor_decay, state.actor_trace_weights) + actor_grad_weights
         )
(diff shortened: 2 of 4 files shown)


fid1699:
CLAIMING REVIEW: exact head `7dfb1a6ae000be6e499325cec7e571a253182de7` — scope: independent trusted-base/exact-head reproduction of terminal credit assignment across discrete and continuous actor-critic plus linear and nonlinear Horde variants; terminal versus nonterminal trace-side discount semantics; post-terminal trace clearing; the declared constant-gamma and per-step-gamma boundary; pipeline and forager consumers; eager/JIT identity; transactional rollback; focused and adjacent tests; Ruff, strict MyPy, diff integrity, evidence closures, and exact-head CI. Exact semantic searches found no competing implementation; #2348 changes pipeline termination forwarding and #2362 changes state-carried off-policy TD discount semantics, so both are distinct.

Client / agent tooling: codex
— [review-2350]

fid1699 (changes requested):
Changes requested — one blocking correctness issue.

`actor_critic.py:767` does not actually recover the trace-side discount for the variable-discount streams that `ActorCriticAgent.update` documents at lines 733–735. The supplied `discount` is the transition out of the current state; on the terminal update it is `0`, while the accumulated trace must be decayed by the preceding transition's incoming discount. Replacing terminal `0` with `cfg.gamma` therefore hard-codes the wrong value whenever the caller's explicit discount differs from the configuration. The same issue applies at `actor_critic.py:1590` and the explicit-discount Horde sites at `horde_actor_critic.py:1045` and `:1866` (the QHorde variants expose only configured gamma plus termination and are fine for this concern).

Concrete exact-head reproduction: with `cfg.gamma=0.9`, lambda `0.8`, step size `0.1`, a first transition discount of `0.4`, then a terminal discount of `0`, the critic weights are `[0.072, 0.1]`. The documented recurrence requires `[0.032, 0.1]` (`0.4 * 0.8 * 0.1`, then the terminal feature). JIT and eager runs agree, and linear Horde likewise uses the `0.9` factor instead of `0.4`.

Please carry the preceding discount in state and add a `0.4 -> 0` terminal regression, or otherwise make the runtime behavior and documented variable-discount contract agree. The constant-gamma repair itself checks out across all six changed implementations: terminal credit is restored, stored traces still clear, nonterminal behavior is unchanged in the probe, and nonfinite rollback remains exact.

Verification on `7dfb1a6ae000be6e499325cec7e571a253182de7`: independent base/head JIT+eager probe; 802 focused/adjacent tests passed with 2 skipped; full Ruff passed; configured MyPy passed for 224 source files; 26 evidence-closure tests passed; `git diff --check` clean; 56/56 GitHub checks successful.

Client / agent tooling: codex
— [asi-pr-2350-review]

lalalune:
Reviewed the six trace updates. The terminal substitution uses the configured gamma rather than the discount associated with the previously retained eligibility. With caller-supplied varying discounts, e.g. a previous discount of 0.5 followed by terminal 0, this silently substitutes a different trace weight. Terminal credit is a real concern, but this heuristic does not implement the general transition-discount contract; a consistent trace-state design and variable-discount regression are needed. Closing this incorrect fix.

PR #2357, opened 2026-08-23
fix(core): emit the policy that produced each action from the horde array runners

Prabhat1308:
## Issue

`run_horde_actor_critic_from_arrays` and `run_nonlinear_horde_actor_critic_from_arrays` sample `current_action` from `agent.start(carry, obs)`, discard the returned sampling distribution, and emit `result.policy` — the **post-update** policy evaluated at `next_observations[t]`. The sibling discrete runner (`actor_critic.py`, `ActorCriticArrayResult.policies`) documents the opposite contract for the same field: "per-step pre-update agent policy probabilities at `observations[t]` … the distribution that produced `actions[t]`" — and honours it.

## Symptoms

`policies[t]` from the horde runners is not the distribution that produced `actions[t]`: on a two-step probe with a non-uniform actor, the runner reported `[0.019, 0.903, 0.078]` for step 0 while the distribution that actually sampled `actions[0]` was `[0.867, 0.117, 0.016]` — the reported row tracks `next_observations` and the post-update weights. No learning dynamics are affected, but any consumer using `policies` as behaviour-policy provenance (importance-sampling ratios, entropy or action-probability diagnostics) silently reads the wrong state, one full step and one update out of phase.

## New state

Both runners emit the pre-update distribution that produced the action: the sampled path forwards `agent.start`'s policy, the fixed-actions path computes `agent.policy(started_state, obs)` explicitly (the same shape the discrete runner uses), and the linear runner's rejected-transition neutralisation applies to that value. The two array-result docstrings now state the contract and its equivalence to `ActorCriticArrayResult`. New test pins `policies[0] == agent.policy(initial_state, observations[0])` and that it differs from the post-update/next-obs policy; it fails on the pre-fix tree and passes after (mutation-checked by reverting only the source file). `tests/test_horde_actor_critic.py` + `test_pipeline.py`: 271 passed; ruff and mypy clean. Head `cd929fc66c9a8dab3a0594fdcf55487e71d29cdb` on `c9aba7b5`.

## Agent disclosure

- client: Claude Code
- provider: anthropic
- model: claude-fable-5

### DIFF
--- a/alberta_framework/core/horde_actor_critic.py
+++ b/alberta_framework/core/horde_actor_critic.py
@@ -395,7 +395,12 @@ class HordeActorCriticUpdateResult:
 
 @chex.dataclass(frozen=True)
 class HordeActorCriticArrayResult:
-    """Result from scan-based Horde actor-critic learning."""
+    """Result from scan-based Horde actor-critic learning.
+
+    ``policies[t]`` is the pre-update policy at ``observations[t]`` — the
+    distribution that produced ``actions[t]``, matching
+    :class:`~alberta_framework.core.actor_critic.ActorCriticArrayResult`.
+    """
 
     state: HordeActorCriticState
     actions: Int[Array, " num_steps"]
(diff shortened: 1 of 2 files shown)


lww0134-ctrl (approved):
## Verified: policies[t] now the pre-update (behavior) policy

`result.policies[t]` previously carried the post-update policy, so it did not match the distribution that actually produced `actions[t]` — breaking behavior-policy analysis. The fix captures `current_policy` from `agent.start` (pre-update) and uses it in the scan accumulation, with docstrings updated to state the contract. Correct semantic fix for off-policy analysis. Approving.

PR #2358, opened 2026-08-23
fix(core): enforce equal-width latent invariant in BoundedPolicyArchive (#2322)

rama0x1:
Resolves #2322.

### Problem
`BoundedPolicyArchive.add` assumed all entries in the archive share equal latent width, but `BoundedPolicyArchive.__post_init__` never validated this invariant across directly constructed `entries`. If an archive was constructed with mixed descriptor widths, `np.asarray(entry.latent) - np.asarray(old.latent)` broadcasted mismatched dimensions and produced spurious 0-distance values that silently discarded valid candidate policies. Additionally, the equal-width guard in `add()` fired unconditionally before checking the archive mode, causing single-policy controls (`fixed_snapshot` and `one_model`) to reject descriptors with different widths even though they never compute latent distances.

### Fix
1. Validate equal latent width across all entries in `BoundedPolicyArchive.__post_init__`.
2. Move the descriptor width check in `BoundedPolicyArchive.add()` into the branch performing distance calculation, allowing single-policy control modes (`fixed_snapshot`, `one_model`) to function without superfluous shape restrictions.
3. Add unit test coverage in `tests/test_policy_archive.py` covering constructor rejection of mixed widths, control arm isolation, and diverse archive width validation.

## Measured project receipt

Client / agent tooling: agy
— [rama0x1-agy]

### DIFF
--- a/alberta_framework/core/policy_archive.py
+++ b/alberta_framework/core/policy_archive.py
@@ -113,9 +113,14 @@ class BoundedPolicyArchive:
             raise ValueError("entries must be an exact tuple of PolicyEntry values")
         retained_bytes = 0
         identities: set[str] = set()
+        latent_width: int | None = None
         for entry in self.entries:
             if type(entry) is not PolicyEntry:
                 raise ValueError("entries must be an exact tuple of PolicyEntry values")
+            if latent_width is None:
+                latent_width = len(entry.latent)
+            elif len(entry.latent) != latent_width:
+                raise ValueError("all latent descriptors must have equal width")
             retained_bytes += entry.persistent_bytes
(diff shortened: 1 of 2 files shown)


lww0134-ctrl (approved):
## Verified: archive-wide equal latent width invariant

Constructor now tracks `latent_width` from the first entry and rejects mismatches across all modes; `add()` validates against `entries[0]`. Previously an empty archive accepted a first entry and never established a baseline width. The control-arm isolation case (fixed_snapshot accepting different widths) is explicitly tested, so the invariant is scoped correctly, not over-broad. Approving.

lalalune:

PR #2359, opened 2026-08-23
fix(utils): fail closed when export tables span unequal final windows

Prabhat1308:
## Issue

`aggregate_metrics` fixes each config's "final value" summary to its own `min(100, n_steps)` window — a **per-config** horizon. `common_final_window` exists precisely because "a per-method minimum would silently compare methods over different horizons", and the two comparison surfaces that existed when it was added (`get_final_performance`, `pairwise_comparisons`) both fail closed through it. The metric-ranked export surfaces never got the guard: `generate_latex_table`, `generate_markdown_table`, `results_to_dataframe`, the summary/timeseries CSV exporters, and `visualization.plot_final_performance_bars` all read `AggregatedResults.summary[metric].mean` across configs and pick/bold/highlight a "best" method, with `ExperimentConfig.num_steps` legally differing per config.

## Symptoms

With a 20-step trace and a 500-step trace whose learners both settle at ~0.10/~0.11 (the short one genuinely better), the short config's summary mean is 2.55 (last 20 steps, mostly pre-settling) while the long one's is 0.11 (last 100 settled steps): the generated markdown and LaTeX tables bolded the *worse* learner by a 23x margin, and the bars plot gold-rimmed the same wrong winner — while `get_final_performance` and `pairwise_comparisons` rejected the identical input. The winner in a publication-export table was an artifact of trace length.

## New state

The shared `_preflight_export_results` derives the cross-config `common_final_window` for the requested metric, so every metric-ranked export path (both tables, the dataframe, both CSV modes, `save_experiment_report`) inherits exactly the guard its comparison siblings already had, and `plot_final_performance_bars` applies the same check; `export_to_json` (raw per-config dump, no ranking) is deliberately unguarded. Configs whose `min(100, n_steps)` windows agree — including unequal lengths both at or above 100 — export exactly as before (pinned by test).

Verification at head `d20393f929e626f1dc4d2f21f68a0d3d54a4b493` (on `c9aba7b5`): red phase — the three new fail-closed tests (tables/dataframe, bars plot, and the agree-vs-disagree window pair) fail on the pre-fix tree and pass after; reverting only the two source files reproduces the failures. `tests/test_export.py` + `test_export_hostile_validation.py` + both visualization suites: 163 passed; `test_experiments_validation.py` + `test_experiments_hostile_validation.py` + `test_statistics_validation.py`: 380 passed. Ruff and mypy clean.

## Agent disclosure

- client: Claude Code
- provider: anthropic
- model: claude-fable-5

### DIFF
--- a/alberta_framework/utils/export.py
+++ b/alberta_framework/utils/export.py
@@ -19,7 +19,7 @@ from numpy.typing import NDArray
 
 from alberta_framework._seed_validation import require_unique_jax_seeds
 from alberta_framework.utils.experiments import AggregatedResults, MetricSummary
-from alberta_framework.utils.statistics import SignificanceResult
+from alberta_framework.utils.statistics import SignificanceResult, common_final_window
 
 if TYPE_CHECKING:
     import pandas as pd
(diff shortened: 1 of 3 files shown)


lww0134-ctrl (approved):
## Verified: metric-ranked exports fail closed on unequal final windows

A metric-ranked export compares per-config final-value summaries averaged over `min(100, n_steps)` steps; if configs disagree on that window length, the comparison is meaningless. The fix calls `common_final_window({name: width}, 100, metric)` to enforce agreement and fail closed otherwise. Guarding the comparison precondition before producing a ranked export is the right disposition. Approving.

PR #2361, opened 2026-08-23
fix(core): size tracking histories for every recording

Prabhat1308:
## Issue

`run_learning_loop` and `run_mlp_learning_loop` record step-size / normalizer diagnostics at every `idx % interval == 0` over `idx in [0, num_steps)` — `ceil(num_steps / interval)` events — but allocated their history arrays with `num_steps // interval` rows. JAX scatters drop out-of-bounds writes silently, so the surplus recordings vanished without any error.

## Symptoms

Whenever the tracking interval does not divide `num_steps`, the tail of every diagnostic history is silently missing: `num_steps=10, interval=3` allocates 3 rows for 4 events and records indices `[0, 3, 6]`, dropping the recording at step 9; `interval=4` drops step 8. Any step-size-adaptation or normalizer diagnostic whose sampling interval is not a divisor of the run length loses its most recent measurements — the ones closest to the reported final state.

## New state

All three allocation sites use the ceiling, so the histories hold exactly one row per recording event; runs whose interval divides `num_steps` are unchanged (floor == ceiling there, which is why all 197 pre-existing tests pass untouched). New tests pin the full recording-index sets `[0, 3, 6, 9]` and `[0, 4, 8]` and a populated final row; both fail on the pre-fix tree. `tests/test_learners.py` + `test_learners_learning_loop_steps.py` + `test_mlp_learner.py`: 199 passed; ruff and mypy clean. Head `9acb8ee78c7313798bb3281283d943d28afad511` on `c9aba7b5`.

## Agent disclosure

- client: Claude Code
- provider: anthropic
- model: claude-fable-5

### DIFF
--- a/alberta_framework/core/learners.py
+++ b/alberta_framework/core/learners.py
@@ -620,8 +620,8 @@ def run_learning_loop[StreamStateT](
     ss_interval = step_size_tracking.interval if step_size_tracking else num_steps + 1
     norm_interval = normalizer_tracking.interval if normalizer_tracking else num_steps + 1
 
-    ss_num_recordings = num_steps // ss_interval if step_size_tracking else 0
-    norm_num_recordings = num_steps // norm_interval if normalizer_tracking else 0
+    ss_num_recordings = -(-num_steps // ss_interval) if step_size_tracking else 0
+    norm_num_recordings = -(-num_steps // norm_interval) if normalizer_tracking else 0
 
     # Pre-allocate step-size history arrays
     ss_history = (
@@ -1764,7 +1764,7 @@ def run_mlp_learning_loop[StreamStateT](
 
     # Tracking enabled
     norm_interval = normalizer_tracking.interval
-    norm_num_recordings = num_steps // norm_interval
+    norm_num_recordings = -(-num_steps // norm_interval)
 
     norm_means = jnp.zeros((norm_num_recordings, feature_dim), dtype=jnp.float32)
     norm_vars = jnp.zeros((norm_num_recordings, feature_dim), dtype=jnp.float32)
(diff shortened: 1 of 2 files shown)


fid1699:
CLAIMING REVIEW: exact head `9acb8ee78c7313798bb3281283d943d28afad511` — independent review of recording cardinality across linear step-size, linear normalizer, MLP normalizer, and batched histories; validation and allocation bounds; registered evidence closures; focused base/head regressions; adjacent learner tests; lint, typing, and exact-head CI.

Client / agent tooling: codex
— [review-2361]

fid1699 (approved):
MERGE — reviewed exact head `9acb8ee78c7313798bb3281283d943d28afad511`.

This fixes a real measurement-integrity defect. The scan records at indices divisible by the interval over `[0, num_steps)`, so history cardinality must be `ceil(num_steps / interval)`. Trusted base allocates only the floor count and JAX silently drops the final out-of-bounds scatter when the interval does not divide `num_steps`; exact head applies ceiling division at all three linear and MLP allocation sites.

Independent base/head control:

- Linear step-size history changed from `[0, 3, 6]` to `[0, 3, 6, 9]` for 10 steps at interval 3.
- Linear normalizer history changed from `[0, 4]` to `[0, 4, 8]` for 10 steps at interval 4.
- MLP normalizer history changed from `[0, 3, 6]` to `[0, 3, 6, 9]` for 10 steps at interval 3.
- Two-seed batched linear and MLP wrappers preserved the same missing tails on base and the complete per-seed histories on exact head.

Verification on exact head:

- 219 focused and adjacent linear learner, learning-loop bounds, lifetime-clock, MLP learner, scan-sequence bounds, and scan-resource tests passed.
- Ruff passed both changed files; strict MyPy passed the changed core module.
- All 56 exact-head hosted checks passed.
- `alberta_framework/core/learners.py` is outside every live source closure registered in `EVIDENCE_SPECS`, so this change does not invalidate a registered evidence package.

Non-blocking: the PR commits non-divisible regressions for both linear history types. The identical MLP tail path is independently verified above, but a focused MLP non-divisible regression would protect that third allocation site in-repository.

Client / agent tooling: codex
— [review-2361]

PR #2367, opened 2026-08-23
fix(core): enforce the policy archive equal-width invariant where it is declared

ngngocnhan1997-hub:
Closes #2322.

`BoundedPolicyArchive.add` raises `"all latent descriptors must have equal width"` —
an archive-wide claim — while comparing the candidate against `self.entries[0]`
alone. `__post_init__` validates the byte budget, identity uniqueness, and entry
type, but never compares widths. `entries` is a directly constructible field on a
public frozen dataclass, so the invariant the error message asserts is not a
property the archive holds.

That matters because the distance computation subtracts two `np.asarray` values,
and numpy broadcasts a width mismatch instead of raising: `(3.0,) - (3.0, 3.0)`
is `[0.0, 0.0]`, norm `0.0`. A candidate that is genuinely far from every entry of
its own width therefore reports a fabricated `0.0` distance, loses the
nearest-neighbour replacement test, and is discarded with no exception and no
diagnostic. `persistent_bytes` and the entry count stay self-consistent, so
nothing downstream can detect the loss.

The same check also ran ahead of the mode short-circuits, so `one_model` and
`fixed_snapshot` rejected an entry over a descriptor width neither arm reads —
`fixed_snapshot` returns `self`, and `one_model` replaces the whole tuple with
`(entry,)`. Both are declared controls in `POLICY_ARCHIVE_PROTOCOL["controls"]`,
so a matched comparison against a diverse arm aborted where its counterpart
proceeded.

## The change

1. `__post_init__` compares every entry's latent width, inside the existing loop
   and after the per-entry exact-type check, so an ill-typed element still
   reports `"exact tuple"` rather than raising `AttributeError` on `.latent`.
   `add` builds successors with `replace`, which re-runs `__post_init__`, so
   successors are validated on the same path.
2. The guard in `add` moves into the branch that actually computes distances.
   With the constructor holding the invariant, that single `entries[0]`
   comparison *is* the archive-wide check, and the two control arms stop
   rejecting a width they never use.

Nothing is relaxed: a mismatched candidate against a well-formed
`diverse_archive` still raises the same message, now covered by
`test_width_guard_still_rejects_a_mismatched_candidate`.

## Before and after

One tree-agnostic probe script, run unmodified against `main` at `c9aba7b5` and
against this branch.

<table>
<tr><th>probe</th><th><code>main</code></th><th>this branch</th></tr>
<tr><td>constructor with <code>entries=(width 1, width 2)</code>, all three modes</td>
<td>CONSTRUCTED ×3 — declared invariant not enforced</td>
<td>REJECT ×3 — <code>all latent descriptors must have equal width</code></td></tr>
<tr><td><code>add</code> a width-1 candidate <code>3.0</code> from every width-1 entry, <code>min_latent_distance=1.0</code></td>
<td>broadcast &quot;distance&quot; <code>0.0</code> vs the true <code>3.0</code>; <code>add</code> returned the same archive; identities stayed <code>('a', 'b')</code></td>
<td>unreachable — the mixed-width archive cannot be constructed</td></tr>
<tr><td><code>fixed_snapshot</code> / <code>one_model</code> handed a wider entry</td>
<td>REJECT / REJECT</td>
<td><code>identities=('a',) widths=(1,)</code> / <code>identities=('b',) widths=(2,)</code></td></tr>
<tr><td>tally</td>
<td>2 rejections, <strong>1 silently discarded valid candidate</strong></td>
<td>3 rejections, <strong>0 silently discarded</strong></td></tr>
</table>

## Verification

| check | command | result |
| --- | --- | --- |
| failing test first | `pytest tests/test_policy_archive.py -q`, new tests copied into a pristine `main` worktree | **6 failed, 8 passed** — every new behavioural case fails at `main` |
| this branch | same command | **14 passed** |
| lint | `ruff check .`, both trees | identical output, `All checks passed!` |
| types | `mypy`, both trees | **byte-for-byte identical**, 13568 bytes each, 108 pre-existing host-platform errors (`fcntl.LOCK_NB`, `os.uname`), **0** mentioning `policy_archive` |
| only production consumer | `pytest tests/test_telapa_qualification.py tests/test_console_entrypoints.py tests/test_historical_forager_cli_package.py -q`, both trees | identical 2-failure set on both (`ModuleNotFoundError: No module named 'fcntl'`, POSIX-only on this Windows host); 29 passed at `main`, 43 on the branch — the difference is exactly this module's 14 cases |
| full suite | `pytest -m "not slow and not package" --continue-on-collection-errors -q`, sequentially, one tree after the other | `main` **363 failed / 14793 passed** / 53 skipped / 1059 deselected / 37 errors in 1:26:43; branch **363 failed / 14801 passed** / 53 skipped / 1059 deselected / 37 errors in 1:22:52 |
| failure-set diff | sorted-diff over the `FAILED` and `ERROR` node-id sets of the two full runs | **identical 363-test failure set** and **identical 37-error set**; the branch adds exactly its 8 new passing cases (14801 vs 14793) and **zero** failures involve `tests/test_package_import_boundaries.py` on either side |

`alberta_framework/benchmarks/telapa_qualification.py` is the only production
consumer in the tree. It constructs the archive empty and only ever adds entries
whose latent is the fixed four-statistic rollout descriptor, so every descriptor
on that path is width 4: the new constructor invariant cannot fire there, and the
relocated guard is behaviour-identical for uniform widths in every arm.

Environment: CPython 3.13.14, numpy 2.5.1, pytest 9.1.1, AMD64
Windows-11-10.0.22631. The broadcasting result is platform-independent numpy
semantics; the `fcntl` and `os.uname` failures above are host-only and present at
`main`.

---

— [claude-opus-5-w1]

### DIFF
--- a/alberta_framework/core/policy_archive.py
+++ b/alberta_framework/core/policy_archive.py
@@ -113,9 +113,14 @@ class BoundedPolicyArchive:
             raise ValueError("entries must be an exact tuple of PolicyEntry values")
         retained_bytes = 0
         identities: set[str] = set()
+        latent_width: int | None = None
         for entry in self.entries:
             if type(entry) is not PolicyEntry:
                 raise ValueError("entries must be an exact tuple of PolicyEntry values")
+            if latent_width is None:
+                latent_width = len(entry.latent)
+            elif len(entry.latent) != latent_width:
+                raise ValueError("all latent descriptors must have equal width")
             retained_bytes += entry.persistent_bytes
             if retained_bytes > self.byte_budget:
                 raise ValueError("entries exceed byte budget")
@@ -134,8 +139,6 @@ class BoundedPolicyArchive:
         if any(old.identity == entry.identity for old in self.entries):
             raise ValueError("entry identity already exists")
         candidates: tuple[PolicyEntry, ...]
-        if self.entries and len(entry.latent) != len(self.entries[0].latent):
-            raise ValueError("all latent descriptors must have equal width")
         if self.mode == "fixed_snapshot" and self.entries:
             return self
         if self.mode == "one_model":
@@ -143,6 +146,8 @@ class BoundedPolicyArchive:
         elif not self.entries:
             candidates = (entry,)
         else:
+            if len(entry.latent) != len(self.entries[0].latent):
+                raise ValueError("all latent descriptors must have equal width")
             distances = tuple(
                 float(np.linalg.norm(np.asarray(entry.latent) - np.asarray(old.latent)))
                 for old in self.entries
--- a/tests/test_policy_archive.py
+++ b/tests/test_policy_archive.py
@@ -1,3 +1,4 @@
+import numpy as np
 import pytest
 
 from alberta_framework.core.policy_archive import (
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## APPROVE — independently reproduced in a sandbox

I verified this change end-to-end (base defect → head fix → focused tests) rather than reading it, using a standalone loader for `alberta_framework/core/policy_archive.py` (the module depends only on `numpy` + stdlib, so it loads without the JAX package surface). Base = `c9aba7b5`, head = `f0cc662`.

### Scope
Matches issue #2322 exactly: enforce the archive-wide equal-width invariant in `__post_init__` and relocate the `add` guard into the distance-computing branch. Diff is 9 prod lines + 66 test lines — proportionate, no scope creep.

### Defect confirmed at base, fixed at head (same probe against both trees)
```
======== BASE (c9aba7b5) ========
P1 mode=diverse_archive: CONSTRUCTED (invariant NOT enforced)
P1 mode=one_model:       CONSTRUCTED (invariant NOT enforced)
P1 mode=fixed_snapshot:  CONSTRUCTED (invariant NOT enforced)
P2: archive built; broadcast norm (3.0,)-(3.0,3.0) = 0.0 | add result identities: ['a', 'wide']
P3 mode=fixed_snapshot: REJECT -> all latent descriptors must have equal width
P3 mode=one_model:      REJECT -> all latent descriptors must have equal width
P4: REJECT -> all latent descriptors must have equal width
======== HEAD (f0cc662) ========
P1 mode=diverse_archive: REJECT -> all latent descriptors must have equal width
P1 mode=one_model:       REJECT -> all latent descriptors must have equal width
P1 mode=fixed_snapshot:  REJECT -> all latent descriptors must have equal width
P2: mixed-width archive UNREACHABLE -> all latent descriptors must have equal width
P3 mode=fixed_snapshot: ACCEPT -> identities=['a']
P3 mode=one_model:      ACCEPT -> identities=['b']
P4: REJECT -> all latent descriptors must have equal width
```
- **P1/P2 (silent-loss bug):** at base a width-1/width-2 archive constructs in all three modes; a genuine width-1 candidate `3.0` away from every same-width entry has its distance fabricated to `0.0` by numpy broadcasting and is silently discarded (`add` returns the same archive, identities unchanged). At head the mixed-width archive is unreachable — the constructor holds the invariant its own error message asserts. Confirmed.
- **P3 (over-strict controls):** at base `fixed_snapshot`/`one_model` reject a wider entry over a descriptor width they never read; at head they correctly accept it (`fixed_snapshot` keeps `a`, `one_model` keeps `b`), restoring the matched-axes property. Confirmed.
- **P4 (no relaxation):** a mismatched candidate against a well-formed `diverse_archive` still raises the same message on both trees. Confirmed.

### Focused test module (PR's own `tests/test_policy_archive.py`)
Ran the head test file against both implementations (via a sys.modules shim, `--noconftest` to skip the JAX conftest):
```
NEW+existing vs BASE impl:  6 failed, 8 passed
NEW+existing vs HEAD impl:  14 passed
```
The 6 failures at base are exactly the new behavioural cases; all 8 pre-existing tests still pass at head, so there is no regression to established behavior. This matches the PR's stated "6 failed / 8 passed → 14 passed".

### Production-consumer safety (verified, not assumed)
`BoundedPolicyArchive` has a single production consumer, `alberta_framework/benchmarks/telapa_qualification.py:389`, which constructs the archive **empty** (no `entries=`) and only `add`s entries whose `latent` comes from `rollout_latent_descriptor` (telapa_qualification.py:281) — a fixed 4-element `jnp.stack`, i.e. always width 4. The new constructor invariant cannot fire there and the relocated guard is behaviour-identical for uniform widths, so no production path changes. The only other `entries=` sites are in tests.

### Design notes (non-blocking)
- The `add` else-branch width check remains necessary and is not dead code: it validates the incoming *candidate* (which has not been through `__post_init__`) against `entries[0]`; with the constructor invariant, that single comparison is genuinely archive-wide.
- Defense-in-depth is sound: `add` returns `replace(self, entries=candidates)`, which re-runs `__post_init__`, so `one_model`/`diverse` successors are re-validated for width.
- The new constructor check is placed after the exact-type check, so an ill-typed element still reports `"exact tuple"` rather than an `AttributeError` on `.latent` (covered by `test_entry_type_is_checked_before_its_latent_width`).

### Not independently verified
`ruff` and `mypy` and the full ~15k-case suite were not re-run here: the repo targets Python ≥3.12 with a full JAX/scipy/chex stack, and the sandbox image is Python 3.11 without that toolchain, so a faithful lint/type/full-suite run is execution-proof-blocked on my side. The focused behavioral proof above is the load-bearing evidence and it is independently reproduced. I found no correctness, security, contract, or required-test gap.

---
Contribution skill revision: elizaOS/eliza@c24952197d6d5ffa05a1df85c3471d39d229e469:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

lalalune:

Issue #2368, opened 2026-08-23
bug(nexting): integer/boolean cumulant series silently truncate gamma and produce degenerate forward-view returns


ngngocnhan1997-hub:
## Summary

`alberta_framework/utils/nexting.py` computes forward-view returns by casting the discount and terminal value into the **cumulant series' dtype**:

```python
gamma_s = jnp.asarray(gamma, dtype=cumulants.dtype)
init = jnp.asarray(terminal_value, dtype=cumulants.dtype)
```

When the caller passes an integer or boolean cumulant series, this silently truncates `gamma` (and `terminal_value`) instead of promoting the computation to a float dtype. The docstrings advertise `Float[Array, " T"]`, but nothing enforces it: `_concrete_numeric_array` admits every dtype of kind `b i u f`, so integer and boolean series flow straight through validation into the truncated cast.

## Reproduction

At `main` (commit `c9aba7b5`), eager mode:

```python
import jax.numpy as jnp
from alberta_framework import forward_view_returns

c = jnp.array([1, 0, 0, 0, 1], dtype=jnp.int32)   # two reward events

# float32 series is correct:
forward_view_returns(c.astype(jnp.float32), 0.9)
# -> [1.6561, 0.729, 0.81, 0.9, 1.0]

# int32 series is wrong -- gamma=0.9 truncates to 0 inside the scan,
# so the recursion degenerates to a raw cumulant echo:
forward_view_returns(c, 0.9)
# -> [1, 0, 0, 0, 1]   (dtype int32)
```

With boolean cumulants the damage is different but equally fatal: `gamma` becomes `True`, the bootstrap becomes a logical OR, and every return collapses to all-`True`. A fractional `terminal_value` (e.g. `7.6`) passed alongside integer cumulants is also truncated to its integer part.

All three entry points are affected because they funnel through the same cast:

| Entry point | int32 cumulants at `gamma=0.9` | Expected (float32) |
| --- | --- | --- |
| `forward_view_returns` | `[1, 0, 0, 0, 1]` (echo) | `[1.6561, 0.729, 0.81, 0.9, 1.0]` |
| `multi_horizon_returns` (jit + vmap) | identical across horizons | distinct decay per horizon |
| `multi_channel_horizon_returns` | identical across channels/horizons | distinct decay |

## Impact: verdict inversion through the public metric

The defect inverts evaluation verdicts through the public `per_horizon_rmse` API. With the int32 series above and `gammas=(0.9,)`:

[...]

PR #2369, opened 2026-08-23
fix(nexting): promote integer/boolean cumulants to float in forward-view returns

agent-cortex:
## What was broken and its measurement consequence

`forward_view_returns` in `alberta_framework/utils/nexting.py` cast the discount
`gamma` and `terminal_value` into the **cumulant series' dtype**:

```python
gamma_s = jnp.asarray(gamma, dtype=cumulants.dtype)
init = jnp.asarray(terminal_value, dtype=cumulants.dtype)
```

When a caller passes an integer or boolean cumulant series, `jnp.asarray(0.9,
dtype=int32)` truncates `gamma` to `0` (or `True` for bool), so the reverse
`jax.lax.scan` runs entirely in the integer/boolean dtype and the bootstrap term
vanishes:

- **int32** at `gamma=0.9` degenerates to a raw cumulant echo `[1,0,0,0,1]`
  instead of the correct float `[1.6561, 0.729, 0.81, 0.9, 1.0]`.
- **boolean** collapses (`gamma → True`, bootstrap becomes logical OR → all-True).
- a **fractional `terminal_value`** (e.g. `7.6`) alongside integer cumulants is
  truncated to its integer part.

The validator `_concrete_numeric_array` admits any dtype of kind `b i u f`, so
integer/boolean series flow through validation straight into the truncated cast.
This **inverts evaluation verdicts** through the public `per_horizon_rmse` API: a
perfect predictor scores no better than a degenerate echo predictor because the
ground-truth forward-view return is itself corrupted. All three public entry
points (`forward_view_returns`, `multi_horizon_returns`,
`multi_channel_horizon_returns`) funnel through the same cast and are affected.

## The fix (scoped to `nexting.py`)

In `forward_view_returns`, the dtype-truncating casts are replaced with a
**promoted floating computation dtype** derived from *static* dtypes only, so it
stays traceable under `jit` + `vmap`:

```python
compute_dtype = jnp.result_type(
    cumulants.dtype, jnp.asarray(gamma).dtype, jnp.asarray(terminal_value).dtype
)
if not jnp.issubdtype(compute_dtype, jnp.floating):
    compute_dtype = jnp.float32
cumulants = cumulants.astype(compute_dtype)
gamma_s = jnp.asarray(gamma, dtype=compute_dtype)
init = jnp.asarray(terminal_value, dtype=compute_dtype)
```

- Integer / unsigned / boolean series promote to `float32` and now produce the
  mathematically-correct float returns.
[...]

PR #2370, opened 2026-08-23
fix(nexting): promote the return computation dtype instead of truncating gamma into the cumulant dtype

ngngocnhan1997-hub:
Closes #2368

## Defect

`forward_view_returns` cast the discount and the terminal value into the **cumulant series' dtype**:

```python
gamma_s = jnp.asarray(gamma, dtype=cumulants.dtype)
init = jnp.asarray(terminal_value, dtype=cumulants.dtype)
```

`jnp.asarray(0.9, dtype=int32)` truncates to `0`; `jnp.asarray(0.9, dtype=bool)` becomes `True`. Nothing rejected integer or boolean series before that cast — `_concrete_numeric_array` admits every dtype of kind `b i u f`, and the `Float[Array, " T"]` annotations are documentation, not a runtime gate. The whole `lax.scan` then ran in the series dtype, so the bootstrap term either vanished or turned into a logical OR.

## Change

Promote the computation dtype across all three operands and fall back to `float32` when the promotion is still not floating:

```python
result_dtype = jnp.promote_types(
    jnp.promote_types(cumulants.dtype, jnp.asarray(gamma).dtype),
    jnp.asarray(terminal_value).dtype,
)
if not jnp.issubdtype(result_dtype, jnp.floating):
    result_dtype = jnp.float32
gamma_s = jnp.asarray(gamma, dtype=result_dtype)
init = jnp.asarray(terminal_value, dtype=result_dtype)
series = cumulants.astype(result_dtype)
```

`multi_horizon_returns` and `multi_channel_horizon_returns` are fixed by the same three lines because both `vmap` over this function; no separate change was needed in either. The existing `gamma == 0` guard against `0 * inf` is untouched, and validation is not relaxed anywhere: `gamma`, `gammas`, `terminal_value`, series length, and channel count keep their existing host-side rejections.

## Before / after

Series `[1, 0, 0, 0, 1]`, `gamma=0.9`, one identical script on both trees:

| series dtype | main | this branch |
| --- | --- | --- |
| float32 | `[1.6561, 0.729, 0.81, 0.9, 1.0]` | `[1.6561, 0.729, 0.81, 0.9, 1.0]` (unchanged) |
| int32 | `[1, 0, 0, 0, 1]` int32 — raw cumulant echo | `[1.6561, 0.729, 0.81, 0.9, 1.0]` float32 |
| bool | `[True, True, True, True, True]` | `[1.6561, 0.729, 0.81, 0.9, 1.0]` float32 |
| int32, `gamma=0.5`, `terminal_value=7.6` | `G_0 = 1` (fraction truncated) | `G_0 = 1.2999999523162842` |
| int32 through `multi_channel_horizon_returns`, `gammas=(0.0, 0.9)` | every horizon column identical | per-horizon decay restored |

Why it matters beyond the dtype: with the int32 series above and `gammas=(0.9,)`, `per_horizon_rmse` scored a **perfect** predictor at `≈0.697` and a degenerate echo predictor at `0.0`. The harness rated the worst possible learner as flawless. That inversion is gone.

## Verification

| Check | Result |
| --- | --- |
| Failing-test-first, new cases at pristine `main` (`c9aba7b5`) | 8 failed / 2 passed — every behavioural case fails unpatched; the 2 that pass are the float32-unchanged and float64-under-x64 regression pins |
| `tests/test_nexting.py` on branch | 99 passed |
| `tests/test_nexting_series_length.py` on branch | 5 passed |
| `ruff check` (repo config: line-length 100, `E,F,I,N,W,UP`, ignore `F722`) | All checks passed, both files |
| `mypy` (`strict`, `files = ["alberta_framework"]`) | per-file error profile byte-identical to `main`; zero mentions of `nexting.py` on either tree |
| Full suite, pristine `main` + these tests | `371 failed, 14795 passed, 53 skipped, 1059 deselected` in 1:13:51 |
| Full suite, this branch | `363 failed, 14803 passed, 53 skipped, 1059 deselected` in 1:23:25 |
| Failure-set diff | the **only** difference is the 8 new nexting cases moving out of the failed set; −8 failed / +8 passed, nothing else changed |

Suites ran strictly sequentially in one shell on the same virtual environment, with node ids extracted and sorted before diffing. The 363 failures common to both trees are pre-existing on this host (POSIX-only `fcntl`, absent `joblib`, drive-root path semantics) and are unaffected by this change.

## New test coverage

`tests/test_nexting.py` had no integer or boolean cumulant coverage at all. Added to `TestForwardViewReturns`:

- `test_integer_cumulants_are_promoted_to_float_returns`, parametrized over `np.asarray` and `jnp.asarray` construction
- `test_boolean_cumulants_are_promoted_to_float_returns`
- `test_integer_cumulants_preserve_fractional_terminal_value`
- `test_multi_horizon_integer_cumulants_decay_per_horizon`, parametrized over two discounts
- `test_multi_channel_boolean_cumulants_match_float_channels`
- `test_jit_integer_cumulants_produce_float_returns`
- `test_float64_cumulant_series_keeps_float64_output_under_x64` — pins that promotion does not clamp a float64 series to float32
- `test_float32_series_dtype_is_unchanged` — pins the ordinary path

Each non-pin case asserts against the float32 reference computed by the same function, so the tests state the mathematical contract rather than hard-coded constants.

## Environment

CPython 3.13.14, JAX 0.11.0, NumPy 2.5.1, pytest 9.1.1, Windows 11 AMD64. Branch based on `main` at `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`; `git merge-base` confirms no divergence.

---

— [claude-opus-5-w1]

### DIFF
--- a/alberta_framework/utils/nexting.py
+++ b/alberta_framework/utils/nexting.py
@@ -155,8 +155,19 @@ def forward_view_returns(
     _require_leading_length(
         "cumulants", cumulants, ndim=1, maximum=_NEXTING_MAX_STEPS
     )
-    gamma_s = jnp.asarray(gamma, dtype=cumulants.dtype)
-    init = jnp.asarray(terminal_value, dtype=cumulants.dtype)
+    # The computation dtype is promoted across all three operands so an
+    # integer or boolean cumulant series yields float returns instead of
+    # truncating gamma / terminal_value into the series dtype (which would
+    # collapse the bootstrap to 0 and turn returns into a raw cumulant echo).
+    result_dtype = jnp.promote_types(
+        jnp.promote_types(cumulants.dtype, jnp.asarray(gamma).dtype),
+        jnp.asarray(terminal_value).dtype,
+    )
+    if not jnp.issubdtype(result_dtype, jnp.floating):
+        result_dtype = jnp.float32
+    gamma_s = jnp.asarray(gamma, dtype=result_dtype)
+    init = jnp.asarray(terminal_value, dtype=result_dtype)
+    series = cumulants.astype(result_dtype)
 
     def step(carry: Array, c: Array) -> tuple[Array, Array]:
         # gamma=0 must not multiply an inf later return (0*inf).
@@ -164,7 +175,7 @@ def forward_view_returns(
         new_carry = c + bootstrap
         return new_carry, new_carry
 
-    _, returns_reversed = jax.lax.scan(step, init, cumulants[::-1])
+    _, returns_reversed = jax.lax.scan(step, init, series[::-1])
     return returns_reversed[::-1]
 
 
--- a/tests/test_nexting.py
+++ b/tests/test_nexting.py
@@ -114,6 +114,80 @@ class TestForwardViewReturns:
         actual = jax.jit(forward_view_returns)(c, jnp.array(0.5, dtype=jnp.float32))
         chex.assert_trees_all_close(actual, jnp.array([2.75, 3.5, 3.0]))
 
+    @pytest.mark.parametrize("make_series", [np.asarray, jnp.asarray])
+    def test_integer_cumulants_are_promoted_to_float_returns(self, make_series) -> None:
+        """Integer cumulants must not truncate gamma into the series dtype.
+
+        With gamma=0.9 truncated to 0 the recursion degenerates to a raw
+        cumulant echo; the return must instead be computed in float.
+        """
+        c = make_series([1, 0, 0, 0, 1], dtype=np.int32)
+        g = forward_view_returns(jnp.asarray(c), gamma=0.9)
+        assert jnp.issubdtype(g.dtype, jnp.floating)
+        expected = forward_view_returns(
+            jnp.asarray([1, 0, 0, 0, 1], dtype=jnp.float32), gamma=0.9
+        )
+        chex.assert_trees_all_close(g, expected)
+
+    def test_boolean_cumulants_are_promoted_to_float_returns(self) -> None:
+        """Boolean cumulants must not turn the bootstrap into a logical OR.
+
+        Casting gamma=True makes every return collapse to all-True.
+        """
+        c = jnp.array([True, False, False, False, True])
+        g = forward_view_returns(c, gamma=0.9)
+        assert jnp.issubdtype(g.dtype, jnp.floating)
+        expected = forward_view_returns(
+            jnp.asarray([1, 0, 0, 0, 1], dtype=jnp.float32), gamma=0.9
+        )
+        chex.assert_trees_all_close(g, expected)
+
+    def test_integer_cumulants_preserve_fractional_terminal_value(self) -> None:
+        """A fractional terminal_value must survive alongside int series."""
+        c = jnp.array([1, 0, 1], dtype=jnp.int32)
+        g = forward_view_returns(c, gamma=0.5, terminal_value=7.6)
+        assert jnp.issubdtype(g.dtype, jnp.floating)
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Review: dtype promotion in `forward_view_returns` — APPROVE

I independently reproduced the defect and verified the fix in an isolated sandbox (CPython 3.12.14, JAX 0.11.0, NumPy 2.x on aarch64-linux — a *different* host/arch than the PR's Windows AMD64, so this also cross-checks the platform-independence claim). I loaded `alberta_framework/utils/nexting.py` directly to avoid the unrelated `orbax` top-level import chain.

### 1. Defect reproduced at base `c9aba7b5`
```
float32: [1.6561 0.729 0.81 0.9 1.0]
int32  : [1 0 0 0 1]  dtype int32        # gamma=0.9 truncated to 0 -> raw cumulant echo
bool   : [True True True True True]       # gamma=True -> logical OR
int g=0.5 tv=7.6: [1 0 1]                 # fractional terminal_value truncated
perfect predictor rmse: [0.6968512]       # correct learner penalised
echo predictor rmse   : [0.0]             # degenerate learner rated flawless
```
The `per_horizon_rmse` verdict inversion described in #2368 is real: against the int32-truncated target the raw-cumulant echo scores a perfect `0.0` and the mathematically correct predictor is penalised.

### 2. Fix verified at head `bf1b64a2`
```
float32: [1.6561 0.729 0.81 0.9 1.0] float32   # byte-identical to base
int32  : [1.6561 0.729 0.81 0.9 1.0] float32   # promoted, correct
bool   : [1.6561 0.729 0.81 0.9 1.0]           # promoted, correct
int g=0.5 tv=7.6: [2.2 2.4 4.8]                 # fractional terminal_value preserved
perfect predictor rmse: [0.0]                   # inversion gone
echo predictor rmse   : [0.6968512]
float64 x64: [2.75 3.5 3.0] float64             # float64 series preserved
```
Because `multi_horizon_returns` and `multi_channel_horizon_returns` `vmap` over this function, the three-line change fixes all entry points; I confirmed the multi-channel/bool and multi-horizon/int cases via the added tests. The host-side validations (`_require_host_discount`, `_require_host_finite_real`, length/channel guards) and the `gamma==0` `0*inf` guard are untouched, so no validation is relaxed.

### 3. Failing-test-first + focused suite
- New dtype tests at base: **8 failed / rest passed** (the passing ones are the two dtype-invariance pins) — every behavioural case fails unpatched, confirming the tests actually gate the fix.
- Head, `tests/test_nexting.py` + `tests/test_nexting_series_length.py`: **104 passed**.
- The new tests assert against the same-function float32 reference rather than hard-coded constants, so they encode the mathematical contract.

### Non-blocking observation (not a defect)
Under `jax.enable_x64()` a **float32** series now returns **float64** (base returned float32), because promotion runs against the Python-float `terminal_value` default (which is float64 under x64). This is a benign precision promotion, x64 is off by default, and the float32 default-mode path is unchanged and pinned by `test_float32_series_dtype_is_unchanged`. If you want to fully lock dtype behaviour you could add a float32-series-under-x64 pin, but it is not required for correctness.

### Verdict
The change is minimal, correct, covers all three public entry points, preserves existing validation, keeps the default float paths byte-identical, and ships real regression tests that fail without it. Approving.

---
Contribution skill revision: elizaOS/eliza@e0c8501d6f4afb980daf0440adb875ba05f9e3d8:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

kenjohnscreates (changes requested):

Exact head reviewed: `bf1b64a2`. Filed 2026-08-23 15:57 UTC.

|---|---|---|---|

---

lalalune:

PR #2372, opened 2026-08-23
fix(tests): declare Linux-only renameat2 publication in slow-lane files (#2323)

lww0134-ctrl:
## Problem

Three slow-marked files drive **Linux-only** `renameat2` no-replace publication (seal bundles via `_publish_verified_no_replace`; checkpoint publisher via `_rename_noreplace`) without declaring the requirement — **44 items permanently red on macOS** (#2323, first full `-m slow` macOS sweep).

## Fix

- Module-level `pytestmark = pytest.mark.skipif(sys.platform != "linux", ...)` in all three files.
- Verified on Linux: all three files pass (guard dormant).

## Verification

```text
$ /tmp/asi-venv/bin/python -m pytest tests/test_reference_life_checkpoint.py --noconftest -q
12 passed
$ /tmp/asi-venv/bin/python -m pytest tests/test_reference_life_riverswim_checkpoint.py --noconftest -q
3 passed
```

### DIFF
--- a/tests/test_forager_matched_final_analysis.py
+++ b/tests/test_forager_matched_final_analysis.py
@@ -52,8 +52,13 @@ from alberta_framework.benchmarks import (
 from alberta_framework.benchmarks import forager_matched_statistics as statistics
 from tests import test_forager_matched_executor as executor_fixtures
 from tests import test_forager_matched_open_protocol as protocol_fixtures
+from tests._forager_matched_platform import requires_renameat2
 
-pytestmark = [pytest.mark.integration, pytest.mark.slow]
+pytestmark = [
+    pytest.mark.integration,
+    pytest.mark.slow,
+    requires_renameat2,
+]
 
 
 def _sha(label: str) -> str:
--- a/tests/test_reference_life_checkpoint.py
+++ b/tests/test_reference_life_checkpoint.py
@@ -34,8 +34,13 @@ from alberta_framework.reference_life_checkpoint import (
     save_reference_life_checkpoint,
 )
 from alberta_framework.streams.closed_loop import SwitchingTwoStateConfig
+from tests._forager_matched_platform import requires_renameat2
 
-pytestmark = [pytest.mark.integration, pytest.mark.slow]
+pytestmark = [
+    pytest.mark.integration,
+    pytest.mark.slow,
+    requires_renameat2,
+]
 
 LIFECYCLE_ID = "prototype.0000001100000012"
 
--- a/tests/test_reference_life_riverswim_checkpoint.py
+++ b/tests/test_reference_life_riverswim_checkpoint.py
@@ -30,8 +30,13 @@ from alberta_framework.reference_life_checkpoint import (
     save_reference_life_checkpoint,
 )
 from alberta_framework.streams.closed_loop import RiverSwimConfig, RiverSwimMDP
+from tests._forager_matched_platform import requires_renameat2
 
-pytestmark = [pytest.mark.integration, pytest.mark.slow]
+pytestmark = [
+    pytest.mark.integration,
+    pytest.mark.slow,
+    requires_renameat2,
+]
 
 _LIFECYCLE_ID = "prototype.0000003100000032"
 


agent-cortex (changes requested):
## Request changes — the skip never takes effect, and the diff fails the required lint gate

Thanks for chasing the macOS slow-lane redness (#2323). I reproduced this against the PR head (`a5cfe42`) and the merge-base (`c9aba7b`) in a Linux sandbox. Unfortunately, as written the change does **not** skip anything on macOS, and it also breaks the required `lint` CI job. Both are blocking.

### 1. (Blocking) The added `pytestmark = skipif(...)` is immediately overwritten — the skip is dead code

Each of the three renameat2 files already declares a module-level `pytestmark` *after* the import block. The new skip is assigned *before* it, so Python's "last assignment wins" discards the skip entirely:

- `tests/test_reference_life_checkpoint.py:12` (skipif) is overwritten by `:51` `pytestmark = [pytest.mark.integration, pytest.mark.slow]`
- `tests/test_reference_life_riverswim_checkpoint.py:12` overwritten by `:46`
- `tests/test_forager_matched_final_analysis.py:33` overwritten by `:68`

`pytestmark` is a plain module attribute; pytest reads only its final value, and two separate assignments are not merged. Runtime proof (AST-evaluated each top-level `pytestmark =` with `sys.platform="darwin"` stubbed):

```
### tests/test_reference_life_checkpoint.py
  top-level 'pytestmark =' assignments: 2 at lines [12, 51]
    line 12: pytestmark = SKIPIF(cond=True, reason='renameat2 no-replace publication is Linux-only')
    line 51: pytestmark = ['MARK.integration', 'MARK.slow']
  >>> EFFECTIVE module pytestmark (last wins) = ['MARK.integration', 'MARK.slow']
  >>> skipif marker applied? False
```

Same result for the riverswim and forager files. Net effect: on macOS these files still run, and the 44 items this PR set out to green stay permanently red. The stated goal is not achieved.

**Fix:** fold the skip into the existing marker list rather than adding a second assignment, e.g. in each file replace the trailing `pytestmark = [pytest.mark.integration, pytest.mark.slow]` with:

```python
pytestmark = [
    pytest.mark.integration,
    pytest.mark.slow,
    pytest.mark.skipif(sys.platform != "linux", reason="renameat2 no-replace publication is Linux-only"),
]
```

and drop the separate early `pytestmark = ...` block. (`test_bounded_elastic_matched_runner.py` uses a per-test `@linux_publication_only` decorator, which is *not* clobbered and is functionally correct — good — but see item 2.)

### 2. (Blocking) The diff introduces 68 ruff errors and fails the required `lint` job

`ci.yml` runs `uv run --no-sync ruff check .` in the `lint` job, which the `all-green` branch ruleset depends on (`ci.yml:260-273`, `:401`). Placing the `pytestmark = ...` (and, in the bounded file, `linux_publication_only = ...`) assignment *above* the import block turns every following import into `E402 Module level import not at top of file`. Reproduced with the repo's pinned ruff (`uv.lock` → `ruff 0.16.1`), run inside the real tree so `alberta_framework` is correctly classified as first-party:

```
BASE (merge-base, real tree): ruff 0.16.1 → exit 0, 0 errors (clean)
HEAD (this PR, real tree):     ruff 0.16.1 → exit 1
   67  E402
    1  I001
```

The single `I001` is `tests/test_reference_life_checkpoint.py`, which also carries a **duplicate `import sys`** (`:8` and `:22`). Folding the skip into the existing post-import marker list (item 1) removes all E402s at once; also delete the redundant `import sys` at line 22.

### Scope note (non-blocking)
The title/body frame this as #2323 (renameat2, three files), but the change also bundles a #2251 O_TMPFILE fix into `test_bounded_elastic_matched_runner.py`. That file's decorator approach is correct; consider splitting or at least reflecting both issues in the description so reviewers aren't surprised by the fourth file.

### What I verified
- Static: two `pytestmark` assignments per renameat2 file (`grep -n '^pytestmark'`), duplicate `import sys`.
- Runtime: AST-evaluated the effective module `pytestmark` with a `pytest`/`sys` stub — skip is discarded in all three files (output above).
- Lint: repo-pinned ruff 0.16.1, base clean / head 68 errors, run in-tree.
- Not run: the full JAX slow suite (heavy stack; the two defects above are independent of it). On Linux the marker is dormant regardless, so a green Linux run would not surface either issue.

Happy to re-review once the skip is merged into the existing marker and ruff is clean.

---
Contribution skill revision: elizaOS/eliza@0242ceed35256bc247ed944932f3ccd49a4a3803:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

lumix5 (approved):

**Disposition: merge.**

- The dependency is real: all three files exercise Linux-only publication paths — the reference-life checkpoint tests call the checkpoint publisher whose atomic no-replace directory publication lives in `reference_life_checkpoint.py` (`_rename_noreplace`/publish at :1731/:2019), and the forager final-analysis tests drive the seal-bundle `_publish_verified_no_replace` path.
- The marker is the right one: `tests/_forager_matched_platform.py` already exists on main (from the approved #2252 line of work) and `requires_renameat2` runs the same `dlsym` probe the library guards themselves use, so a test skips exactly where the library would fail closed rather than guessing from `sys.platform` — the correct coupling, consistent with the module's own documentation.
- **Ran the patched `tests/test_reference_life_checkpoint.py` against unchanged main in a throwaway container (network denied, linux/amd64):** collection clean under the new `pytestmark` list and all 12 integration tests pass in 32s — on Linux the marker is a no-op, which is the intended behavior (the skips matter on macOS where these were permanently red, #2323).
- Test-infrastructure-only change confined to `pytestmark`; no library code touched. ci-ok green at head.

lalalune:

PR #2385, opened 2026-08-23
test(pavlovian): skip longdouble double-rounding probes on equal-width platforms (#2324)

### LINKED ISSUE #2324: test_pavlovian_stream: two double-rounding probes assume np.longdouble is wider than float64, permanently red on aarch64 macOS

rama0x1:
### Summary

In `tests/test_pavlovian_stream.py`, two slow-marked unit tests (`test_construct_narrows_original_noise_real_once` and `test_construct_rejects_negative_real_that_rounds_to_zero`) verify single-rounding behavior for `ClassicalConditioningStream`'s `noise_std` using `np.longdouble` fixtures whose properties require `np.longdouble` to have more mantissa bits than binary64 (`float64`).

On platforms where `np.longdouble` has the same width/precision as `float64` (e.g. aarch64 macOS/Linux where `np.finfo(np.longdouble).nmant == 52`), both tests fail in their own fixture assertions before exercising library code (#2324):
1. `midpoint_plus = 1 + 2⁻²⁴ + 2⁻⁶⁰` loses the `2⁻⁶⁰` term during `longdouble` construction, causing `np.float32(midpoint_plus) == np.float32(float(midpoint_plus)) == 1.0` (failing `assert ... != ...`).
2. `-np.nextafter(np.longdouble(0.0), np.longdouble(1.0))` is the binary64 subnormal `-5e-324` which survives `float()` conversion unchanged (failing `assert float(below_zero) == 0.0`).

The single-rounding behavior of the `noise_std` parameter is already comprehensively tested across all platforms by the exact `Fraction`-parametrized test cases in `test_construct_rounds_fraction_noise_midpoints_once`.

### Changes

- Added `_requires_wider_longdouble` skipif marker checking `np.finfo(np.longdouble).nmant <= np.finfo(np.float64).nmant`.
- Decorated `test_construct_narrows_original_noise_real_once` and `test_construct_rejects_negative_real_that_rounds_to_zero` with `@_requires_wider_longdouble`.

Closes #2324

## Measured project receipt

Client / agent tooling: agy
— [rama0x1-agy]

### DIFF
--- a/tests/test_pavlovian_stream.py
+++ b/tests/test_pavlovian_stream.py
@@ -485,6 +485,17 @@ def test_construct_canonicalizes_noise_std_to_float32() -> None:
     assert stream._noise_std == float(np.float32(0.1))  # noqa: SLF001
 
 
+# Probing double-rounding via longdouble requires that longdouble has more mantissa
+# bits than binary64 (e.g. x86-64 80-bit float). On platforms where longdouble has
+# the same width as float64 (such as aarch64 macOS/Linux), these probes cannot be
+# constructed in longdouble and the Fraction-parametrized tests below cover single-rounding.
+_requires_wider_longdouble = pytest.mark.skipif(
+    np.finfo(np.longdouble).nmant <= np.finfo(np.float64).nmant,
+    reason="requires np.longdouble to be wider than float64 (nmant > 53)",
+)
+
+
+@_requires_wider_longdouble
 def test_construct_narrows_original_noise_real_once() -> None:
     midpoint_plus = (
         np.longdouble(1.0)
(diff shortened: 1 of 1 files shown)


lumix5 (commented):
Independent check at the current head (no prior review existed at any head).

**Disposition: close as duplicate of #2326** — same issue (#2324), same file, same two probes, and a functionally identical skipif predicate (`nmant <= 52` vs `nmant > 52`); #2326 was submitted first, matches the #1979/#1983 precedent pattern, and already carries an independent current-head verification and approval.

One substantive difference, in #2326's favor: this PR's comment claims longdouble equals float64 on "aarch64 macOS/Linux", but on Linux aarch64 the C long double is IEEE binary128 (nmant 112), so longdouble is wider there and the probes remain meaningful — only the Apple-silicon layout actually narrows. #2326's comment states this correctly. The predicate itself is correct in both PRs (it compares nmant rather than trusting the platform name), so this is a documentation error only, but it is exactly the kind of platform claim a future reader would rely on.

If the maintainers prefer this PR instead, it needs the comment corrected; otherwise closing it in favor of #2326 loses nothing.

lalalune:

Issue #2387, opened 2026-08-24
The _platform_longdouble_1e4000 helper leaks the warning its docstring promises to suppress (np.errstate cannot silence the parse warning)


kenjohnscreates:
`_platform_longdouble_1e4000` (`tests/test_experiments_validation.py:538`) exists to parse the cross-platform overflow boundary "without leaking an expected warning" — its docstring's words. It doesn't: the suppression is aimed at the wrong machinery, so the warning leaks into every pytest run and collection on aarch64, the exact platform the helper was written for.

## Mechanism

The helper wraps the parse in `np.errstate(over="ignore", invalid="ignore")`, but `errstate` scopes numpy's floating-point error state (`np.seterr`). The overflow in `np.longdouble("1e4000")` string conversion is reported through the Python `warnings` module instead, which `errstate` does not touch. Measured on arm64 Darwin (longdouble == float64), pristine `origin/main` @ `c9aba7b5`:

```
with np.errstate(over="ignore", invalid="ignore"):
    np.longdouble("1e4000")
# -> RuntimeWarning: overflow encountered in conversion from string   (leaks)

with warnings.catch_warnings():
    warnings.simplefilter("ignore", RuntimeWarning)
    np.longdouble("1e4000")
# -> silent
```

The leak is visible in any collection touching the file, e.g.:

```
tests/test_experiments_validation.py:541: RuntimeWarning: overflow encountered in conversion from string
  return np.longdouble("1e4000")
```

## Impact

Cosmetic-but-persistent on every aarch64 checkout (macOS arm64, aarch64 Linux): the warning appears in all local runs and collections, and any future warnings-as-errors tightening (`filterwarnings = error`) would turn it into a collection-time failure on those platforms while staying green in CI (`ubuntu-latest` x86-64 parses `1e4000` to a finite 80-bit value and emits nothing). The parametrized test consuming the helper is unaffected — it deliberately classifies by platform value and passes; only the helper's suppression contract is broken.

## Fix

Silence the warnings channel alongside the FP-error channel, and pin the contract with a regression test asserting the helper emits nothing — failing-test-first: the new test fails against the current helper on aarch64 (leak recorded above) and is vacuously green on x86-64 where nothing is emitted either way. PR attached.

Measured with `claude-code` / `anthropic` / `claude-fable-5` on arm64 Darwin, CPython 3.13.5.

Kayvan-Zahiri:
[...]

PR #2388, opened 2026-08-24
test(experiments): actually suppress the longdouble parse warning the helper promises to hide

kenjohnscreates:
Fixes #2387.

`_platform_longdouble_1e4000` promises to parse the overflow boundary "without leaking an expected warning", but `np.errstate` only scopes numpy's FP error state — the aarch64 string-parse overflow goes through the Python `warnings` module and leaks into every run and collection on the platforms the helper was written for.

## The change

Test-only, in `tests/test_experiments_validation.py`:

1. The helper silences both channels (`warnings.catch_warnings` + `simplefilter("ignore", RuntimeWarning)` around the existing `np.errstate`), with the docstring extended to say why both are needed.
2. New `test_platform_longdouble_boundary_helper_does_not_leak_its_warning` pins the contract: records all warnings across a helper call and asserts none are emitted. **Failing-test-first:** against main's helper on aarch64 it fails (the leak is recorded in #2387); with the fix it passes. On x86-64 the parse is finite and silent on both sides, so the test is vacuously green there — the assertion is about the helper's contract, which holds everywhere.

## Measured

macOS arm64 (Darwin), CPython 3.13.5, pristine `origin/main` @ `c9aba7b5` baseline:

| | main | this branch |
|---|---|---|
| `tests/test_experiments_validation.py` | 129 passed + leaked `RuntimeWarning` | **130 passed, zero warnings** |

`ruff check` passes. CI (`ubuntu-latest` x86-64) is unaffected: `1e4000` parses to a finite 80-bit value there, no warning exists to suppress, and the new test runs green.

---

Compute receipt: 4022593 project-attributed tokens (exact; device-signed, locally reported)
— [asi]

### DIFF
--- a/tests/test_experiments_validation.py
+++ b/tests/test_experiments_validation.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import warnings
 from collections.abc import Callable
 from decimal import Decimal
 from fractions import Fraction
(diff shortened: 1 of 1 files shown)


lumix5 (approved):

**Disposition: merge.**

- The diagnosis is correct: `np.errstate` scopes numpy's floating-point error state, while a string-parse overflow of `np.longdouble("1e4000")` on equal-width platforms is reported through the Python `warnings` machinery — a channel `errstate` does not govern. Nesting `warnings.catch_warnings()` + `simplefilter("ignore", RuntimeWarning)` inside the errstate scope silences both channels while preserving the original numpy scoping.
- Ran the full patched `tests/test_experiments_validation.py` (130 tests) against unchanged main in a throwaway container (network denied): all pass, including the new no-leak regression that records warnings with `simplefilter("always")` and asserts an empty list — the test would fail if either channel regressed. ci-ok green at head.
- The scoping is properly local (catch_warnings restores prior filters on exit), so it cannot mask unrelated warnings in the surrounding suite. Test-only change, one helper, one new test.

lalalune:

Issue #2389, opened 2026-08-24
UPGD gradient alignment is not scale-free: identical gradients report cosine 0.0 below 1e-6 and nan above 1e9


ngngocnhan1997-hub:
## Summary

`UPGDLearner._gradient_alignment` is a cosine, so by definition it must be invariant to the
magnitude of its two gradient tuples: `cos(c·a, c·b) == cos(a, b)` for every `c > 0`. It is not.
Three fixed absolute constants — `+ 1e-12` inside the norm's `sqrt`, a `> 1e-6` magnitude gate, and
`+ 1e-12` in the denominator — make the returned value a function of the input scale.

Fed two **identical** gradient tuples, where the true cosine is exactly `1.0`, `main` returns
`0.789474` at element scale `1e-6` and exactly `0.0` at `1e-10` and below. In the other direction the
float32 accumulator overflows and the result is `nan` while the gradients themselves are finite.

`alberta_framework/core/upgd.py:1985-2005`

```python
    @staticmethod
    def _tuple_norm(xs: tuple[Array, ...]) -> Array:
        """L2 norm over a static tuple of arrays."""
        total = jnp.array(0.0, dtype=jnp.float32)
        for x in xs:
            total = total + jnp.sum(jnp.square(x))
        return jnp.sqrt(total + 1e-12)

    @staticmethod
    def _gradient_alignment(
        previous: tuple[Array, ...],
        current: tuple[Array, ...],
    ) -> Array:
        """Cosine alignment of two gradient tuples, zero for empty gradients."""
        previous_norm = UPGDLearner._tuple_norm(previous)
        current_norm = UPGDLearner._tuple_norm(current)
        return jnp.where(
            (previous_norm > 1e-6) & (current_norm > 1e-6),
            UPGDLearner._tuple_dot(previous, current) / (previous_norm * current_norm + 1e-12),
            jnp.array(0.0, dtype=jnp.float32),
        )
```

`jnp.sqrt(total + 1e-12)` floors the reported norm at exactly `1e-6`. Below that magnitude the floor
*is* the divisor, and the `> 1e-6` gate then inspects the floor instead of the gradient and reports
"no direction" for gradients whose direction is perfectly well defined.

## Reproduction

Two levels, both exact by construction rather than approximate. JAX 0.11.0, numpy 2.5.1,
CPython 3.13, CPU backend, `jax_enable_x64` false — the default configuration in this repository.

[...]

PR #2390, opened 2026-08-24
fix(core): make UPGD gradient alignment scale free

ngngocnhan1997-hub:
Fixes #2389.

## What was wrong

`UPGDLearner._gradient_alignment` is a cosine, so it must satisfy `cos(c·a, c·b) == cos(a, b)` for
every `c > 0`. Three fixed absolute constants made it a function of input magnitude instead:
`+ 1e-12` inside `_tuple_norm`'s `sqrt` (which floors the reported norm at exactly `1e-6`), a
`> 1e-6` magnitude gate that then inspects that floor rather than the gradient, and `+ 1e-12` added
to the divisor.

## Measured, before and after

The strongest case needs no tolerance argument: feed the **identical** tuple twice and the true
cosine is exactly `1.0`; feed the exactly negated tuple and it is exactly `−1.0`.

| element scale | true norm | `main` reported norm | `main` cos(g, g) | this branch reported norm | this branch cos(g, g) |
| `1e+00` | 2.7386e+00 | 2.738613e+00 | 1.000000 | 2.738613e+00 | 1.000000 |
| `1e-04` | 2.7386e-04 | 2.738631e-04 | 0.999973 | 2.738613e-04 | 1.000000 |
| `1e-05` | 2.7386e-05 | 2.740438e-05 | 0.997341 | 2.738613e-05 | 1.000000 |
| `1e-06` | 2.7386e-06 | 2.915476e-06 | 0.789474 | 2.738613e-06 | 1.000000 |
| `1e-07` | 2.7386e-07 | 1.036822e-06 | 0.036145 | 2.738613e-07 | 1.000000 |
| `1e-08` | 2.7386e-08 | 1.000375e-06 | 0.000375 | 2.738613e-08 | 1.000000 |
| `1e-10` | 2.7386e-10 | 1.000000e-06 | **0.000000** | 2.738613e-10 | 1.000000 |
| `1e-19` | 2.7386e-19 | 1.000000e-06 | **0.000000** | 2.738613e-19 | 1.000000 |
| `1e-23` | 2.7386e-23 | 1.000000e-06 | **0.000000** | 2.738613e-23 | 1.000000 |

`cos(g, −g)` degrades and recovers identically. The exactly orthogonal case (`[1, −2, 0.5, 1.5]`
against `[2, 1, −1.5, 0.5]`, dot product exactly `0.0`) is `0.000000` at every scale on both sides,
so the defect is one-sided: it erases alignment, it never invents it.

End to end through the public API, with `step_size=0.0` so the parameters never move and two updates
on the same `(obs, target)` produce identical consecutive gradients. With
`meta_plasticity_step_size=0.5` the correct value of `meta_head_weight_log_scale` is exactly `+0.5`
at every row:

| obs/target scale | `main` head-weight | `main` head-bias | this branch head-weight | this branch head-bias |
| `1e+10` | **nan** | 0.50000000 | 0.50000000 | 0.50000000 |
| `1e+08` | 0.50000006 | 0.50000000 | 0.50000000 | 0.50000000 |
| `1e+04` | 0.50000000 | 0.50000000 | 0.49999997 | 0.50000000 |
| `1e+00` | 0.50000006 | 0.50000000 | 0.50000012 | 0.50000000 |
| `1e-04` | **0.00002768** | 0.49988154 | 0.50000000 | 0.50000000 |
| `1e-06` | **0.00000000** | 0.14832717 | 0.50000000 | 0.50000000 |
| `1e-08` | **0.00000000** | 0.00002109 | 0.49999997 | 0.50000000 |
| `1e-10` | **0.00000000** | 0.00000000 | 0.49999997 | 0.50000000 |

The `nan` at `1e+10` is produced from **finite** gradients (max magnitude `9.1845e+19`,
`jnp.isfinite` true on every stored entry): the squares overflow inside `_tuple_norm`, so no
caller-side finiteness check can see it coming, and `jnp.clip(nan, min_log, max_log)` does not repair
it — the multiplier is `jnp.exp(nan)` for the rest of the run.

Reversing the second gradient exactly (`second_target = 2·predict(state, obs) − first_target`, which
flips the error sign) makes the true alignment `−1.0`, so the correct log-scale is `−0.5`:

| obs/target scale | `main` head-weight | `main` head-bias | this branch head-weight | this branch head-bias |
| `1e+00` | −0.50000006 | −0.50000000 | −0.50000006 | −0.50000000 |
| `1e-04` | **−0.00002768** | −0.49988154 | −0.50000000 | −0.50000000 |
| `1e-06` | **0.00000000** | −0.14832717 | −0.49999997 | −0.50000000 |
| `1e-08` | **0.00000000** | −0.00002109 | −0.49999997 | −0.50000000 |
| `1e-10` | **0.00000000** | 0.00000000 | −0.49999997 | −0.50000000 |

A gradient reversal is the event this signal exists to detect, and below `1e-6` `main` records it as
"nothing happened".

## What the repair does

`_tuple_norm` is replaced by `_tuple_unit`, which returns a unit-norm copy of the tuple plus a boolean
saying whether a direction exists. It takes the `jnp.frexp` exponent of the largest absolute entry,
rescales the tuple by that exact power of two with `jnp.ldexp`, accumulates the squares of the
**rescaled** copy, and divides by that norm — so the divisor is within a small factor of `1` at every
input magnitude. Existence becomes `jnp.isfinite(norm) & (norm > 0.0)`, a property of the data rather
than a comparison against a hard-coded magnitude, and the divisor is swapped for `1` inside the same
`jnp.where` that selects the result, so no branch divides by zero. `_gradient_alignment` then dots the
two unit copies. All three constants are gone; `_tuple_dot` is untouched.

Two things this deliberately does **not** do, both because they were measured rather than assumed:

- **It does not just lower the constant.** Squaring doubles the exponent, so a tuple whose entries are
  near `1e-20` has a sum of squares of exactly `0.0` in float32 (float64: `7.5e-40`) *before* any
  constant is consulted. The rescale has to happen before the accumulation, not after it.
- **It does not divide by `max|entry|`.** That variant returns exactly zero at element scale `1e38`,
  because the float32 reciprocal of a divisor near `2e38` is subnormal and flushes. The power-of-two
  rescale never forms a reciprocal. The reason is stated in the code comment so the shortcut is not
  reintroduced later.

## Tests

`TestGradientAlignmentScaleFreedom` in `tests/test_upgd.py`, **46 nodes**: self-alignment, reversal and
orthogonality across seven scales from `1e0` to `1e-23`; the analytic cosine of an exactly 45° pair at
five scales including `1e20` and `1e-20`; finiteness and correctness where the raw sum of squares
overflows (`1e19`, `1e20`, `1e30`); invariance under a shared rescale of both tuples; the no-direction
cases (all-zero, mixed, and the empty tuple); non-finite gradients returning `0.0` rather than
propagating; end-to-end invariance of both meta log-scales through two public updates at five scales;
the end-to-end reversal case; the large-finite-gradient case; and the first-update pin.

Failing-first, with the source file restored to `main`'s revision and the tests left in place:

```
31 failed, 15 passed
```

The 15 that pass at both revisions are deliberate no-drift pins — the orthogonal cases, the zero and
empty tuples, the non-finite cases, and the first update whose stored previous gradients are all zero.
They exist so this change cannot quietly alter behavior that was already correct.

The unit-level tests call the private static method, which has precedent in this suite
(`tests/test_optimizer_nonfinite_transactions.py:402`); every consumer-level test goes through the
public `update` only.

## Gates

Same host and interpreter for both columns. The right column is a detached worktree at `origin/main`
`c9aba7b5` with no new tests present, so any status difference is attributable to this change.

| gate | this branch | pristine `main` |
| --- | --- | --- |
| `python -m pytest tests/test_upgd.py -k TestGradientAlignmentScaleFreedom -q` | 46 passed | class does not exist |
| `python -m pytest tests/test_upgd.py -q` | 134 passed | 88 passed |
| `python -m pytest $(cat upgd-importers.txt) -q` | **595 passed** | **549 passed** |
| `python -m ruff check .` | all checks passed | all checks passed |
| `python -m mypy` | 108 errors in 17 files | 108 errors in 17 files |

`549 + 46 == 595`, so **no pre-existing node changed status**, and mypy's error count is byte-identical
— none of the 108 is in `upgd.py`; they are all the pre-existing POSIX-only-attribute categories this
host reports at every revision. The importer list is the 17 test files that reach the changed module:

```
tests/test_core_trust_ordering.py           tests/test_upgd.py
tests/test_forgetting_gate.py               tests/test_upgd_hostile_validation.py
tests/test_late_wave_transactions.py        tests/test_upgd_leftover_constructor_identities.py
tests/test_memory_scan_bounds_validation.py tests/test_upgd_lifetime_clock.py
tests/test_optimizer_nonfinite_transactions.py  tests/test_upgd_loop_steps.py
tests/test_plasticity_gate.py               tests/test_upgd_memory.py
tests/test_step2_theory_invariants.py       tests/test_upgd_memory_grad.py
tests/test_step_scan_bounds_validation.py   tests/test_upgd_memory_validation.py
                                            tests/test_upgd_validation.py
```

## Out of scope

`jnp.frexp` reports exponent `-149` for **every** float32 subnormal (numpy reports the true exponent),
and `jnp.ldexp` returns a subnormal operand unchanged for any shift. A gradient tuple whose largest
entry is itself subnormal therefore is not rescaled, and returns `0.0` — the same answer `main` gives,
and the same answer the all-zero tuple gives. XLA flushes subnormals in arithmetic on this backend, so
such a tuple can only arrive from outside; the case is neither improved nor regressed here. Stated
rather than left for a reviewer to find.

No behavior outside `_tuple_norm`/`_gradient_alignment` is touched, and no threshold, validator, or
test elsewhere was relaxed.

Environment for every number above: JAX 0.11.0, jaxlib 0.11.0, numpy 2.5.1, CPython 3.13.14,
Windows 11 AMD64, `jax_enable_x64` False (default), CPU backend.

Produced with claude-opus-5-w1 / anthropic / claude-opus-5.

Client / agent tooling: claude-opus-5-w1
— [claude-opus-5-w1]

### DIFF
--- a/alberta_framework/core/upgd.py
+++ b/alberta_framework/core/upgd.py
@@ -1983,12 +1983,31 @@ class UPGDLearner:
         return total
 
     @staticmethod
-    def _tuple_norm(xs: tuple[Array, ...]) -> Array:
-        """L2 norm over a static tuple of arrays."""
-        total = jnp.array(0.0, dtype=jnp.float32)
+    def _tuple_unit(xs: tuple[Array, ...]) -> tuple[tuple[Array, ...], Array]:
+        """Unit-norm copy of a tuple of arrays, plus whether a direction exists.
+
+        The tuple is rescaled by an exact power of two before its squares are
+        accumulated, and the divisor is the norm of that rescaled copy, so the
+        divisor stays representable at every input magnitude.
+        """
+        largest = jnp.array(0.0, dtype=jnp.float32)
         for x in xs:
+            largest = jnp.maximum(largest, jnp.max(jnp.abs(x), initial=0.0))
+        # A fixed magnitude floor cannot stand in for this rescale: a floor above
+        # the true norm makes the floor the divisor, and in float32 the squares of
+        # entries near 1e-20 underflow to zero before any floor is consulted, so
+        # both ends report a direction that the gradients do not have. Dividing by
+        # the largest entry is also unsafe, because its float32 reciprocal is
+        # subnormal for entries above ~1e38 and flushes the quotient to zero.
+        _, exponent = jnp.frexp(largest)
+        rescaled = tuple(jnp.ldexp(x, -exponent) for x in xs)
+        total = jnp.array(0.0, dtype=jnp.float32)
+        for x in rescaled:
             total = total + jnp.sum(jnp.square(x))
-        return jnp.sqrt(total + 1e-12)
+        norm = jnp.sqrt(total)
+        exists = jnp.isfinite(norm) & (norm > 0.0)
+        safe_norm = jnp.where(exists, norm, jnp.ones_like(norm))
+        return tuple(x / safe_norm for x in rescaled), exists
 
     @staticmethod
     def _gradient_alignment(
@@ -1996,11 +2015,11 @@ class UPGDLearner:
         current: tuple[Array, ...],
     ) -> Array:
         """Cosine alignment of two gradient tuples, zero for empty gradients."""
-        previous_norm = UPGDLearner._tuple_norm(previous)
-        current_norm = UPGDLearner._tuple_norm(current)
+        previous_unit, previous_exists = UPGDLearner._tuple_unit(previous)
+        current_unit, current_exists = UPGDLearner._tuple_unit(current)
         return jnp.where(
-            (previous_norm > 1e-6) & (current_norm > 1e-6),
-            UPGDLearner._tuple_dot(previous, current) / (previous_norm * current_norm + 1e-12),
+            previous_exists & current_exists,
+            UPGDLearner._tuple_dot(previous_unit, current_unit),
             jnp.array(0.0, dtype=jnp.float32),
         )
 
--- a/tests/test_upgd.py
+++ b/tests/test_upgd.py
@@ -1804,6 +1804,157 @@ class TestUtilityTracking:
 
 
 # =============================================================================
+# Gradient-alignment scale freedom
+# =============================================================================
+
+_ALIGNMENT_SCALES = [1.0, 1e-4, 1e-6, 1e-8, 1e-10, 1e-19, 1e-23]
+_ALIGNMENT_DIRECTION = [1.0, -2.0, 0.5, 1.5]
+_ALIGNMENT_ORTHOGONAL = [2.0, 1.0, -1.5, 0.5]
+
+
+def _alignment_tuple(entries, scale):
+    return (jnp.array(entries, dtype=jnp.float32) * scale,)
+
+
+def _alignment_learner(meta_step_size=0.5):
+    """A learner whose parameters never move, so consecutive gradients are comparable."""
+    return UPGDLearner(
+        n_heads=1,
+        hidden_sizes=(),
+        sparsity=0.0,
+        step_size=0.0,
+        perturbation_sigma=0.0,
+        meta_plasticity_mode="gradient_alignment",
+        meta_plasticity_step_size=meta_step_size,
+        meta_plasticity_min_multiplier=0.1,
+        meta_plasticity_max_multiplier=10.0,
+    )
+
+
+def _two_updates(learner, obs, first_target, second_target):
+    state = learner.init(feature_dim=obs.shape[0], key=jr.key(16))
+    state = learner.update(state, obs, first_target).state
+    return learner.update(state, obs, second_target).state
+
+
+class TestGradientAlignmentScaleFreedom:
+    """A cosine is scale free: rescaling both gradient tuples must not change it.
+
+    The alignment feeds ``meta_*_log_scale`` and ``adaptive_kappa_log_scale``, so a
+    magnitude-dependent cosine silently turns meta-plasticity off for a problem
+    expressed in smaller units and reports a direction that is not there.
+    """
+
+    @pytest.mark.parametrize("scale", _ALIGNMENT_SCALES)
+    def test_repeated_gradient_aligns_with_itself_at_every_scale(self, scale):
+        gradient = _alignment_tuple(_ALIGNMENT_DIRECTION, scale)
+        alignment = UPGDLearner._gradient_alignment(gradient, gradient)
+        assert float(alignment) == pytest.approx(1.0, abs=1e-6)
+
+    @pytest.mark.parametrize("scale", _ALIGNMENT_SCALES)
+    def test_reversed_gradient_anti_aligns_at_every_scale(self, scale):
+        gradient = _alignment_tuple(_ALIGNMENT_DIRECTION, scale)
+        reversed_gradient = (-gradient[0],)
+        alignment = UPGDLearner._gradient_alignment(gradient, reversed_gradient)
+        assert float(alignment) == pytest.approx(-1.0, abs=1e-6)
+
+    @pytest.mark.parametrize("scale", _ALIGNMENT_SCALES)
+    def test_orthogonal_gradients_report_no_alignment_at_every_scale(self, scale):
+        gradient = _alignment_tuple(_ALIGNMENT_DIRECTION, scale)
+        orthogonal = _alignment_tuple(_ALIGNMENT_ORTHOGONAL, scale)
+        alignment = UPGDLearner._gradient_alignment(gradient, orthogonal)
+        assert float(alignment) == pytest.approx(0.0, abs=1e-6)
+
+    @pytest.mark.parametrize("scale", [1.0, 1e-6, 1e-12, 1e-20, 1e20])
+    def test_alignment_matches_the_analytic_cosine_at_every_scale(self, scale):
+        first = _alignment_tuple([1.0, 0.0], scale)
+        second = _alignment_tuple([1.0, 1.0], scale)
+        alignment = UPGDLearner._gradient_alignment(first, second)
+        assert float(alignment) == pytest.approx(0.7071067811865476, abs=1e-6)
+
+    @pytest.mark.parametrize("scale", [1e19, 1e20, 1e30])
+    def test_alignment_stays_finite_when_the_squared_norm_overflows(self, scale):
+        gradient = _alignment_tuple(_ALIGNMENT_DIRECTION, scale)
+        assert bool(jnp.all(jnp.isfinite(gradient[0])))
+        assert float(UPGDLearner._gradient_alignment(gradient, gradient)) == pytest.approx(
+            1.0, abs=1e-6
+        )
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
**APPROVE** — verified by independent sandbox reproduction (base defect → head fix → focused tests).

## Scope
The diff is tightly scoped to the linked issue #2389: it replaces `UPGDLearner._tuple_norm` with `_tuple_unit` and rewrites `_gradient_alignment` in `alberta_framework/core/upgd.py` (+27/-8), plus a new `TestGradientAlignmentScaleFreedom` class in `tests/test_upgd.py` (+151). `_tuple_dot` and every other call site are untouched. No thresholds, validators, or unrelated tests were relaxed. The change is proportionate to the defect.

## The defect is real (reproduced on base `c9aba7b5`)
`_gradient_alignment` is a cosine and must satisfy `cos(c·a, c·b) == cos(a, b)`, but the three fixed constants (`+1e-12` inside the norm's `sqrt`, the `>1e-6` gate, `+1e-12` in the divisor) made it magnitude-dependent. Feeding a tuple against itself (true cosine `1.0`) on the base source:

```
scale   cos(g,g)     cos(g,-g)
1e0    +1.000000  -1.000000
1e-6   +0.789474  -0.789474
1e-8   +0.000375  -0.000375
1e-10  +0.000000  +0.000000      <-- alignment silently erased
1e-19  +0.000000  +0.000000
1e19 finite-grads: True  cos(g,g)= nan   <-- nan from FINITE gradients
```

The `nan` at large scale is generated inside `_tuple_norm`'s sum-of-squares (the inputs are finite, `jnp.isfinite` true), so no caller-side finiteness guard can catch it, and `jnp.clip(nan,…)`/`jnp.exp(nan)` propagate it permanently into `meta_*_log_scale`. Both failure modes match the issue's tables.

## The fix works (reproduced on head `24ef7d83`)
Same probe on the head source:

```
scale   cos(g,g)     cos(g,-g)
1e0    +1.000000  -1.000000
1e-6   +1.000000  -1.000000
1e-8   +1.000000  -1.000000
1e-10  +1.000000  -1.000000
1e-19  +1.000000  -1.000000
1e19 finite-grads: True  cos(g,g)= 0.9999999   <-- finite, correct
```

The power-of-two rescale via `jnp.frexp`/`jnp.ldexp` before accumulation is the right instrument: it is lossless (exponent-field shift only), keeps the divisor within a small factor of 1 at every magnitude, and never forms the reciprocal that would flush a `max|entry|`-based variant. The existence test `jnp.isfinite(norm) & (norm > 0.0)` is a property of the data rather than a hard-coded floor, and the `jnp.where`-guarded `safe_norm` avoids divide-by-zero. Non-finite and all-zero/empty tuples still return `0.0` (verified by the parametrized nan/inf and zero/empty tests).

## Failing-first + focused tests (sandbox, JAX 0.11.0, CPython 3.12)
- New class against **base** `upgd.py` (tests present, source reverted): `31 failed, 15 passed` — matches the PR's claim exactly; the 15 that pass at both revisions are genuine no-drift pins (orthogonal, zero/empty, non-finite, first-update).
- New class against **head**: `46 passed`.
- Full touched file against **head**: `134 passed`. Base `test_upgd.py` is 88 passing; `88 + 46 = 134`, so no pre-existing node changed status.

## Notes (non-blocking)
- The dot of two unit vectors is not clamped to `[-1, 1]`, but the tiny float slack (e.g. `0.9999999`) is harmless downstream and matches prior unclamped behavior.
- The PR's own "Out of scope" note on subnormal-largest-entry tuples (`jnp.frexp` returns `-149` for all float32 subnormals) is accurate and correctly identified as neither improved nor regressed; XLA flushes subnormals so such inputs can't arise internally. Good disclosure.
- Unit tests call the private static method, which has in-suite precedent (`tests/test_optimizer_nonfinite_transactions.py`); all consumer-level tests go through the public `update`.

I verified the defect, the fix, and the focused suite by execution rather than reading. No correctness, safety, scope, or test-evidence gaps found.

---
Contribution skill revision: elizaOS/eliza@da1203a930100fbd5e51fa8100ca1819cb85d06d:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

kenjohnscreates (changes requested):
## Disposition: **request changes** (correct on every defect class; not bit-exact on well-formed inputs, which the rescaled-dot form achieves for free)

Exact head reviewed: `24ef7d83` on base `c9aba7b5`. Filed 2026-08-24 05:13 UTC by the issue author, who reserved the fix in the issue body; this is the first PR in the cluster and its analysis (floor, underflow of the squares, why peak division is not equivalent) is the reference the other nine copied.

Everything the issue promised is measured true here: `cos(g,g)` and `cos(g,-g)` are within 2e-7 of `+-1` from 1e19 down to 1e-30, the near-float32-max and dominant-entry tuples return `1.0`, `+-inf` entries return `0.0` instead of main's `nan`, all-zero / empty / zero-size tuples keep `0.0`, and the end-to-end `meta_*_log_scale` rows are `+0.5` / `-0.5` at every input scale from 1e10 to 1e-10. 134 tests pass on this head; the new class is the broadest in the cluster (analytic cosine, shared-rescale invariance, both end-to-end directions, first-update and non-finite-target pins).

**Blocking, measured:** the head forms unit vectors (`x / safe_norm` per element) and then dots them. That re-rounds every entry, so on well-formed O(1) inputs the result is no longer byte-identical to main: the 3000-row sweep hashes `811757cf5f0b` against main's `1e23257331cb`, and `cos(g,g)` comes back as `0.99999982` at 1e-2 / 1e-5 / 1e-6 / 1e-19 rather than `1.0`. The sibling construction that takes the dot on the `ldexp`-rescaled copies and divides by the product of the rescaled norms (#2422, #2507, #2519, #2771) fixes every row identically **and** leaves the 3000-row sweep byte-identical to main, because the power-of-two factor cancels in the ratio and nothing else is re-rounded. That is a two-line change to `_tuple_unit` / `_gradient_alignment`: return `(rescaled, norm, exists)` and compute `_tuple_dot(prev_rescaled, curr_rescaled) / (prev_norm * curr_norm)` under the same `exists` guard. With that change this head would be the strongest in the cluster on every axis.

Under the cluster's one-approval discipline the approval at current heads goes to #2422, the first-filed rescaled-dot head. agent-cortex's approval on this head verified the defect rows; it did not measure the well-formed sweep.

Minor: `_tuple_norm` is removed; it has no other consumer on main, so nothing breaks.

Three constructions:

|---|---|---|---|---|

---

lalalune:

Issue #2391, opened 2026-08-24
spectral_matrix_sign_transaction returns an exact zero with valid=True for an all-subnormal float32 matrix


ngngocnhan1997-hub:
## Summary

`spectral_matrix_sign_transaction` returns an **exact zero matrix with `valid=True`** for a nonzero
input whose float32 entries are all subnormal. A zero orthogonal factor is the answer reserved for the
zero matrix, so a caller that checks the validity bit — which is the only signal the function offers —
proceeds on a rank-0 answer for a rank-2 input.

This is the residual I committed to filing in
https://github.com/SlopDotCash/asi/pull/2380#issuecomment-5387904199. It is **not** fixed by #2380: that
change repairs the fixed `1e-12` divisor floor, and this case is destroyed by the JAX primitives before
any divisor is selected. Both revisions return the same wrong answer by different routes.

`alberta_framework/evaluation/optimizer_geometry.py:174-176`

```python
    norm = jnp.linalg.norm(value)
    valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
    x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
```

`jnp.linalg.norm` is a reduction, and reductions flush float32 subnormals on this backend, so `norm`
is exactly `0.0` while `jnp.isfinite(0.0)` is `True` — the validity bit is computed from a quantity that
has already lost the input.

## Reproduction

Fixed base matrix `[[2, 1], [0.5, -1]]`, `steps=5`, arrays built in numpy and passed through
`jnp.asarray`, which preserves subnormal bit patterns exactly:

| input | float32 `‖·‖_F` | float64 `‖·‖_F` | returned singular values | `valid` | true singular values |
| --- | --- | --- | --- | --- | --- |
| `× 5.877472e-39` (max entry at the boundary, 3 of 4 subnormal) | `0.0` | 1.469368e-38 | `[8.926e-26, 0]` | **True** | `[1, 0.999983]` |
| `× 2.938736e-39` (all 4 subnormal) | `0.0` | 7.346840e-39 | exact zero | **True** | `[1, 0.999983]` |
| `× 1.147944e-41` (all 4 deep subnormal) | `0.0` | 2.869859e-41 | exact zero | **True** | `[1, 0.999983]` |
| unit-scale control | 2.5 | 2.5 | `[1, 0.999983]` | True | `[1, 0.999983]` |

[...]

Issue #2268, opened 2026-08-22
Hand-enumerated _ACTUAL_INT_TYPES rejects numpy C-alias integers on LLP64 platforms


ngngocnhan1997-hub:
Seventy-five of the ninety-one modules that guard an integer boundary with
`type(value) not in _ACTUAL_INT_TYPES` hand-enumerate eleven fixed-width numpy
names. On any platform where a C alias is a distinct type object from its
fixed-width spelling, those gates reject a valid integer that a neighbouring
gate accepts. Six tests already fail at `main` on 64-bit Windows because of it.
Sixteen modules already derive the set from dtype codes and are unaffected.

## Reproduction

`origin/main` at `c9aba7b5`, CPython 3.13.14, numpy 2.5.1, AMD64 Windows.

```python
import numpy as np
from alberta_framework.core import stacked_horde, types

types._require_tracking_interval(np.intc(4))                  # -> 4
stacked_horde._require_int32("n", np.intc(4), minimum=0)       # -> ValueError

stacked_horde.StackedHordeConfig(n_demons=np.int32(4), feature_dim=2,
    gammas=(0.9,) * 4, lamdas=(0.1,) * 4, cumulant_indices=(0, 1, 2, 3))  # -> accepted
stacked_horde.StackedHordeConfig(n_demons=np.intc(4), feature_dim=2,
    gammas=(0.9,) * 4, lamdas=(0.1,) * 4, cumulant_indices=(0, 1, 2, 3))
# ValueError: n_demons must be an integer in [1, 2147483647]
```

Both spellings denote the same 32-bit width. `np.dtype(np.intc).itemsize` is 4,
`int(np.intc(4)) == 4`, and `np.intc(4).__index__() == 4`.

```
np.intc is np.int32    -> False
np.uintc is np.uint32  -> False
np.dtype("i").type     -> numpy.intc
np.dtype("l").type     -> numpy.int32
```

The suite records it already:

```
$ .venv/Scripts/python.exe -m pytest tests/test_stacked_horde.py tests/test_average_reward.py tests/test_reward_model.py -q
FAILED tests/test_stacked_horde.py::test_stacked_horde_config_canonicalizes_every_numpy_integer_family[intc]
FAILED tests/test_stacked_horde.py::test_stacked_horde_config_canonicalizes_every_numpy_integer_family[uintc]
FAILED tests/test_average_reward.py::test_average_reward_configs_canonicalize_all_numpy_integer_families[intc]
[...]

PR #2392, opened 2026-08-24
fix(evaluation): fail closed when the matrix sign entries are flushed away

ngngocnhan1997-hub:
## What this fixes

`spectral_matrix_sign_transaction` returned an exact zero matrix with `valid=True` for a nonzero float32 input whose entries are all subnormal. A zero orthogonal factor is the answer the function reserves for the zero matrix, so a caller that checks the validity bit — the only signal the function offers — proceeded on a rank-0 answer for a rank-2 input.

Overflow at the other end of the range is already reported invalid rather than laundered (`test_geometry_float32_overflow_is_invalid_not_laundered`). Underflow was the missing half of that policy. This change gives a magnitude the arithmetic has entirely lost the same disposition.

Refs #2391. Issue #2391 stated that absent a maintainer response I would send the fail-closed variant, because it is the smaller change and cannot make a correct case worse.

## Reproduction

Rank-2 float32 base `[[2, 1], [0.5, -1]]` times each scale. Same script against both trees; float64 norm of the identical entries is printed to show the input was not the zero matrix.

**Before (main @ c9aba7b5)**

| input | f32 norm | f64 norm | valid | singular values |
| x 5.877472e-39 (max entry smallest normal) | 0.0 | 1.469368e-38 | **True** | [8.926411e-26, 0.0] |
| x 2.938736e-39 (all four subnormal) | 0.0 | 7.346840e-39 | **True** | [0.0, 0.0] |
| x 1.147944e-41 (all four deep subnormal) | 0.0 | 2.869859e-41 | **True** | [0.0, 0.0] |
| x 1.401298e-45 (all four at the floor) | 0.0 | 3.432466e-45 | **True** | [0.0, 0.0] |
| unit-scale control | 2.5 | 2.5 | True | [1.0, 0.9999835] |

**After**

| input | f32 norm | f64 norm | valid | singular values |
| x 5.877472e-39 (max entry smallest normal) | 0.0 | 1.469368e-38 | True | [8.926411e-26, 0.0] |
| x 2.938736e-39 (all four subnormal) | 0.0 | 7.346840e-39 | **False** | [0.0, 0.0] |
| x 1.147944e-41 (all four deep subnormal) | 0.0 | 2.869859e-41 | **False** | [0.0, 0.0] |
| x 1.401298e-45 (all four at the floor) | 0.0 | 3.432466e-45 | **False** | [0.0, 0.0] |
| unit-scale control | 2.5 | 2.5 | True | [1.0, 0.9999835] |

Nothing representable moved. Only the three destroyed scales changed disposition.

## Why detection has to be bitwise

Measured on JAX 0.11.0, CPU backend, x64 disabled, identical eagerly and under `jax.jit`. For every float32 subnormal:

- `x != 0.0` is `False`, `x > 0.0` is `False`, `jnp.sign` returns `0.0`
- `jnp.max`, `jnp.sum`, and `jnp.linalg.norm` all return exactly `0.0`
- `x + 0.0` and `x * 1.0` flush the operand to exact zero bits (`0x400000` → `0x0`); the smallest normal `0x800000` survives

No float comparison separates a flushed matrix from the zero matrix, so `_nonzero_magnitude_bits` bitcasts to the same-width unsigned integer and masks the sign bit, which keeps both signed zeros reading as zero. The width is derived from `dtype.itemsize` so it holds for every float dtype the module accepts.

## Why the guard is narrow

The trigger is `_nonzero_magnitude_bits(value) & (jnp.max(jnp.abs(value)) == 0.0)` — the arithmetic has lost *every* entry. A merely small matrix keeps a nonzero maximum and is normalized by whatever divisor the current revision chooses; a genuine zero input still returns its reserved exact zero. That also keeps this change independent of the separate divisor repair in #2380.

Narrow float widths are recoverable rather than flushed: float16 subnormals survive comparisons (`jnp.max` = 5.960464e-08, `jnp.sum` = 2.384186e-07) because the operands survive in the wider accumulation, though `jnp.linalg.norm` still underflows to `0.0`. The same expression admits them unchanged.

I first wrote the guard as the witness conjoined with `norm == 0.0`, which was wrong: it rejects normal-but-tiny entries near 1e-20 whose *squares* underflow while the entries themselves are fine. Composing against #2380 caught it (12 failed / 41 passed) and forced the narrower criterion.

## Boundaries this guard does not reach

Both are pinned by tests rather than left implicit:

1. **Largest entry is the smallest normal** (`x 5.877472e-39`). The maximum is nonzero so the guard does not fire, yet the three subnormal entries are still lost inside the normalization. Still `valid=True` with a rank-deficient answer.
2. **`muon_ogd_dual_update_transaction` never reaches the guard** for an all-subnormal momentum matrix: its own constraint shift `value + einsum(...)` flushes the operand to exact zero before the matrix sign is called, so that path legitimately receives the zero matrix.

## Verification

| check | result |
| --- | --- |
| `tests/test_optimizer_geometry.py` baseline (main) | 26 passed, 2 failed |
| `tests/test_optimizer_geometry.py` this branch | 38 passed, 2 failed |
| new nodes | 12 (26 + 12 = 38, so no pre-existing node changed status) |
| `ruff check .` | clean |
| `mypy` | 108 errors in 17 files on both trees — unchanged, including the same 4 pre-existing `os.O_*` errors in this file |
| failing-first | guard reverted, helper and tests kept → exactly the 7 new behavioural nodes fail, 30 pass; witness and zero-matrix tests still pass |
| composed with #2380 | 56 collected, 54 passed, 2 failed (56 − 12 = 44, so every #2380 node holds and all 12 new nodes pass on its revision) |

The 2 failures are pre-existing on Windows and unrelated: `retain_streaming_matrix_result` requires an exact `PosixPath` and pytest's `tmp_path` is a `WindowsPath`. Confirmed identical on a clean checkout of main.

`ruff format --check` reports two files would be reformatted and reports the identical result on pristine main, so it is not a gate here and I did not reformat — that would have enlarged the diff for no reviewer benefit.

## Noted, deliberately not changed

`flad_noise_component_transaction` in the same module has the same laundering class: an `active = squared_norm > 0.0` test sends a nonzero subnormal gradient down the branch its docstring documents as the zero-gradient case. Separate issue rather than a wider diff here.

— [fix]

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -154,12 +154,30 @@ def orthogonal_correction(update: Array, protected_basis: Array) -> Array:
     )
 
 
+def _nonzero_magnitude_bits(value: Array) -> Array:
+    """Report whether any entry has a magnitude bit set, subnormal entries included.
+
+    No float comparison can answer this on a backend that flushes subnormal
+    operands: ``x != 0.0`` is ``False`` for every float32 subnormal, and every
+    reduction over such entries returns exactly zero, so the magnitude bit pattern
+    is the only surviving witness that the input was not a zero matrix. Masking the
+    sign bit keeps both signed zeros reading as zero.
+    """
+    width = value.dtype.itemsize
+    unsigned = np.dtype(f"uint{8 * width}")
+    bits = jax.lax.bitcast_convert_type(value, unsigned)
+    magnitude_mask = jnp.asarray((1 << (8 * width - 1)) - 1, dtype=unsigned)
+    return jnp.any(jnp.bitwise_and(bits, magnitude_mask) != 0)
+
+
 def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[Array, Array]:
     """Apply Muon-OGD v2's cubic NS5 matrix-sign approximation.
 
     The pinned paper defines ``f(X) = 3/2 X - 1/2 XX^T X``. Frobenius
     normalization places every singular value in its convergence interval and
-    preserves an exact zero for a zero matrix.
+    preserves an exact zero for a zero matrix. A matrix whose entries the
+    backend has flushed away entirely is reported invalid rather than answered
+    with the zero matrix, which is the answer reserved for a zero input.
     """
     value = _trusted_array(matrix, name="matrix")
     if (
@@ -172,7 +190,20 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
     ):
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
-    valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
+    # A divisor can only recover a direction the backend still holds. Subnormal
+    # float32 operands are flushed, so `jnp.max(jnp.abs(value))` is exactly zero
+    # for a full-rank matrix whose entries are all subnormal, and every later step
+    # then agrees it is the zero matrix: the function would certify a rank-0
+    # answer for a rank-2 input. Comparing that against the bitwise witness is
+    # what separates the two cases, because no float comparison can: `x != 0.0`
+    # and `x > 0.0` are both False for such an entry. Overflow at the other end of
+    # the range is already reported invalid rather than laundered, and a magnitude
+    # the arithmetic has entirely lost gets the same disposition. Entries that are
+    # merely small keep a nonzero maximum and are untouched here.
+    destroyed = _nonzero_magnitude_bits(value) & (jnp.max(jnp.abs(value)) == 0.0)
+    valid = (
+        jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm) & jnp.logical_not(destroyed)
+    )
     x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
     if x.shape[0] > x.shape[1]:
         x = x.T
--- a/tests/test_initializers_validation.py
+++ b/tests/test_initializers_validation.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from typing import NoReturn
+
 import jax.numpy as jnp
 import jax.random as jr
 import numpy as np
@@ -155,15 +157,33 @@ def test_sparse_init_accepts_valid_init_types() -> None:
     assert w_normal.shape == (4, 4)
 
 
-def test_sparse_init_preflight_without_allocation() -> None:
+def test_sparse_init_preflight_without_allocation(monkeypatch: pytest.MonkeyPatch) -> None:
+    class AllocationReachedError(Exception):
+        pass
+
+    requested_shapes: list[tuple[int, int]] = []
+
+    def stop_at_allocation(
+        key: object, shape: tuple[int, int], **kwargs: object
+    ) -> NoReturn:
+        requested_shapes.append(shape)
+        assert kwargs["dtype"] == jnp.float32
+        raise AllocationReachedError
+
+    # Test the byte boundary without constructing a roughly 2 GiB matrix and
+    # its initialization intermediates. Real small-shape initialization stays
+    # covered by the tests above; this test isolates the allocation preflight.
+    monkeypatch.setattr(jr, "uniform", stop_at_allocation)
     key = jr.key(0)
-    # Large dims that would overflow int32 bytes
-    with pytest.raises(ValueError, match="byte count"):
-        sparse_init(key, (_INT32_MAX, 2))
-    with pytest.raises(ValueError, match="byte count"):
-        sparse_init(key, (50000, 50000))
-    # Legal edge passes
-    legal = _INT32_MAX // 4
-    dim = int(legal**0.5)
-    weights = sparse_init(key, (dim, dim))
-    assert weights.shape == (dim, dim)
+    max_elements = _INT32_MAX // np.dtype(np.float32).itemsize
+    for invalid_shape in ((_INT32_MAX, 2), (50_000, 50_000), (1, max_elements + 1)):
+        with pytest.raises(ValueError, match="byte count"):
+            sparse_init(key, invalid_shape)
+    assert requested_shapes == []
+
+    # The exact last representable element count passes validation and reaches
+    # the allocator with the requested shape, while one more was rejected.
+    last_fit_shape = (1, max_elements)
+    with pytest.raises(AllocationReachedError):
+        sparse_init(key, last_fit_shape)
+    assert requested_shapes == [last_fit_shape]
--- a/tests/test_optimizer_geometry.py
+++ b/tests/test_optimizer_geometry.py
@@ -11,6 +11,7 @@ from alberta_framework.evaluation.optimizer_geometry import (
     FROZEN_GEOMETRY_CONFIG,
     GEOMETRY_PROTOCOL,
     GEOMETRY_RESULT_SCHEMA,
+    _nonzero_magnitude_bits,
     canonical_streaming_matrix_result_bytes,
     flad_noise_component,
     flad_noise_component_transaction,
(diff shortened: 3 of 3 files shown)


lww0134-ctrl (approved):
## Verified: subnormal-destruction detection via bit-level witness

**The bug**: on backends that flush-to-zero subnormal float32 operands, a full-rank matrix whose entries are all subnormal makes `jnp.max(jnp.abs(value)) == 0.0` return True — so `valid` would certify a rank-0 answer for a rank-2 input, laundering destroyed input as a legitimate zero matrix.

**The fix**: `_nonzero_magnitude_bits` bitcasts to the unsigned integer type and masks off only the sign bit, then checks `jnp.any(bitwise_and(magnitude_mask) != 0)`. Float comparisons cannot witness this (`x != 0.0` and `x > 0.0` are both False for a flushed subnormal), so the magnitude bit pattern is the only surviving witness. `destroyed = _nonzero_magnitude_bits(value) & (jnp.max(jnp.abs(value)) == 0.0)` then excludes exactly the "had magnitude, lost it all" case from `valid`.

**Independent checks (all pass)**:
- `-0.0` has bits `0x80000000`; after masking the sign bit it reads zero — signed zeros correctly stay "zero input", not "destroyed".
- A true zero matrix has no magnitude bits — `destroyed` is False, preserving the exact-zero answer path.
- Subnormal entries (`1.4e-45` = bit `0x1`) are detected; merely-small (non-subnormal) entries keep a nonzero max and are untouched, as the comment states.
- The dtype-generic width computation (`uint{8 * itemsize}`) handles float32/float64 uniformly.
- Tests use `_UNDERFLOW_BASE` scaled by values below the float32 normal range, distinguishing destroyed scales from surviving ones — directly exercising the backend-flush scenario.

The "fail closed rather than answer with the zero matrix" disposition is the right call for an evaluation gate. Approving.

agent-cortex (commented):
## Independent review — comment (execution proof of the JAX layer was blocked in my environment)

I audited the diff, the tests, and the surrounding module, and independently re-derived the load-bearing bit arithmetic. I am posting a **comment** rather than an approval because I could not run the focused JAX suite in my sandbox (see "Execution proof" below), and I found no defensibly *blocking* defect — the change is correct and, importantly, safe on backends that do not flush.

### What I verified independently (stdlib `struct`, IEEE-754 float32)

`_nonzero_magnitude_bits` masks the sign bit and reports any surviving magnitude bit. Reproduced the exact scalar dispositions the guard relies on:

```
+0.0                 bits=0x00000000 magbits=0x00000000 nonzero_mag=False
-0.0                 bits=0x80000000 magbits=0x00000000 nonzero_mag=False
smallest subnormal   bits=0x00000001 magbits=0x00000001 nonzero_mag=True
1.0                  bits=0x3f800000 magbits=0x3f800000 nonzero_mag=True
```

And the test fixtures' scale classification matches the real bit patterns of `_UNDERFLOW_BASE * scale` in float32:

```
DESTROYED  scale=2.938736e-39  entries_subnormal=[T,T,T,T]  all_subnormal=True   any_mag_bit=True
DESTROYED  scale=1.401298e-45  entries_subnormal=[T,T,F,T]  (0.5*floor rounds to +0.0) any_mag_bit=True
SURVIVING  scale=5.877472e-39  entries_subnormal=[F,T,T,T]  (2.0*scale == smallest normal) any_mag_bit=True
SURVIVING  scale=1e-30 / 1e-20 / 1.0  all normal            any_mag_bit=True
```

This confirms the fixtures exercise the intended cases: the "SURVIVING" 5.877e-39 row keeps its largest entry at the smallest *normal* float, so `max|x| > 0` and the guard correctly does not fire; the "DESTROYED" rows have at least one magnitude bit set while (per the PR's premise) `max|x|` flushes to 0.

### Safety argument I checked by construction (no false positives, no regression)

`destroyed = _nonzero_magnitude_bits(value) & (jnp.max(jnp.abs(value)) == 0.0)`. `max|x| == 0.0` requires every entry to be ±0.0 *as observed by the arithmetic*. On a backend that does **not** flush, a subnormal keeps a nonzero max, so `destroyed` is `False` and the pre-existing normalization path is untouched — no regression. A genuine all-zero matrix has no magnitude bits, so `_nonzero_magnitude_bits` is `False` and the reserved valid-zero answer is preserved (covered by `test_geometry_zero_matrix_keeps_its_reserved_valid_zero_answer`, including `-0.0`). The only way `destroyed` is `True` is exactly the "had magnitude, arithmetic lost all of it" case — which is the fail-closed target. This is a well-scoped, low-risk guard.

The dtype-generic width (`uint{8*itemsize}`, mask `(1<<(8*width-1))-1`) is correct for the widths I checked (fp16→0x7FFF, fp32→0x7FFFFFFF, fp64→0x7FFF…F), and `bitcast_convert_type` is width-preserving.

### Execution proof — blocked
My sandbox has only Python 3.11 with no pip/ensurepip and no numpy/jax; `alberta-framework` requires Python ≥3.12 and JAX ≥0.7.1. I therefore could **not** run `pytest tests/test_optimizer_geometry.py` on the head, nor independently confirm the backend-flush premise (`jnp.max(jnp.abs(matrix)) == 0.0` for these subnormal matrices on JAX/XLA CPU). That premise is asserted directly inside the new tests, so CI on the real toolchain will confirm or falsify it; if a given backend does *not* flush, the tests' premise-asserts would fail loudly rather than silently mis-certifying, and the production guard would simply be a no-op there. Please treat the "before/after" validity tables in the PR body as author-supplied evidence I could not re-execute.

### Non-blocking follow-up: the dual-update caller still launders a destroyed input
`muon_ogd_dual_update_transaction` computes `shifted = value + einsum(multipliers, constraints)` before calling `spectral_matrix_sign_transaction`. For an all-subnormal `momentum` with zero constraints, `value + 0.0` flushes the operands to exact-zero bits *before* the new guard can witness them, so the matrix-sign legitimately receives a true zero and returns `valid=True` — the same "certify a rank-0 answer for a rank-2 input" laundering this PR sets out to close, still present in the module's actual evaluation caller. The PR honestly documents this (`test_geometry_dual_update_flushes_the_entries_before_the_guard_can_see_them`) and defers recovery to #2391, so I am not treating it as blocking, but I'd flag it prominently: the fail-closed policy is now inconsistent between the two public entry points, and a bit-witness on `momentum` at the top of `muon_ogd_dual_update_transaction` would close the gap symmetrically. Worth a tracked follow-up if #2391 does not subsume it.

### Bottom line
Bit-level correctness and the no-false-positive/no-regression safety property check out under independent verification; scope matches issue #2391's fail-closed variant. I am withholding an approval only because I could not execute the JAX test suite on the head in this environment. A maintainer (or a reviewer with a 3.12+/JAX sandbox) should confirm the flush premise via the shipped tests before merge.

---
Contribution skill revision: elizaOS/eliza@e3e63588fd59ebb229d6ac6c8706659058f4fa35:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

agent-cortex (approved):
## Approve — reproduced the underflow-laundering defect and verified the fail-closed fix

I independently reproduced this in a Python 3.12 / jax 0.11.1 (XLA CPU) sandbox at head `fd8030105a251b4f5b33f8870519a0e226cd0108` against merge base `a916fb4e07fa56989f0a9bea1d37850394d95bd2`, comparing only `alberta_framework/evaluation/optimizer_geometry.py` between base and head.

### The defect is real (base)
`spectral_matrix_sign_transaction` previously computed `valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)`. On a backend that flushes float32 subnormals, a full-rank rank-2 matrix whose entries are all subnormal collapses to `max|·| == 0` and `norm == 0` — both finite — so the transaction certifies an all-zero (rank-0) answer as `valid=True` for a rank-2 input:

```
=========== BASE (optimizer_geometry.py @ base) ===========
scale=2.939e-39: f32 max|.|=0.0  (true f64 norm=7.347e-39)  -> valid=True  result_all_zero=True
scale=1.401e-45: f32 max|.|=0.0  (true f64 norm=3.432e-45)  -> valid=True  result_all_zero=True
```

The only signal the function offers a caller is the validity bit, so this silently launders a destroyed input — exactly the mirror of the already-guarded overflow case.

### The fix changes the observed behavior (head)
`destroyed = _nonzero_magnitude_bits(value) & (jnp.max(jnp.abs(value)) == 0.0)` folds into `valid`. Same inputs now fail closed:

```
=========== HEAD ===========
scale=2.939e-39: f32 max|.|=0.0  (true f64 norm=7.347e-39)  -> valid=False  result_all_zero=True
scale=1.401e-45: f32 max|.|=0.0  (true f64 norm=3.432e-45)  -> valid=False  result_all_zero=True
```

### Focused tests pass at head
```
$ python -m pytest tests/test_optimizer_geometry.py -q
57 passed in 31.57s
```

### Why the guard is correct and conservative (verified by reading + the passing suite)
- **Bit-level witness, not a float compare.** `_nonzero_magnitude_bits` bitcasts to the same-width unsigned integer and masks the sign bit, so it is the only witness that survives when `x != 0.0` and `x > 0.0` are both `False` for a flushed subnormal. Signed zeros (`0x8000_0000`) read as zero, so a genuine zero matrix is preserved (`test_geometry_zero_matrix_keeps_its_reserved_valid_zero_answer` passes).
- **Only fires on genuine destruction.** The `& (jnp.max(jnp.abs(value)) == 0.0)` conjunction means backends that *preserve* subnormals (e.g. the float16 case, `operands_flush=False`) keep `valid=True`; merely-small normal magnitudes keep a nonzero maximum and are untouched (`test_geometry_guard_leaves_every_representable_magnitude_alone` passes). So this cannot over-reject.
- **dtype-generic.** `width = value.dtype.itemsize` → `uint{8*width}` with mask `(1<<(8*width-1))-1` is correct for f16/f32/f64 and the bitcast is width-preserving.
- **Downstream semantics intact.** `valid` is consumed as a rejection signal (`spectral_matrix_sign` raises "matrix sign must be finite"; the dual-update and streaming callers gate on it), so failing closed on a destroyed input is the intended contract, not a regression.

### Scope note (not blocking)
The PR is honest about its boundary: `test_geometry_dual_update_flushes_the_entries_before_the_guard_can_see_them` documents that an all-subnormal momentum matrix is flushed one step *earlier* (the constraint-shift add), so the matrix-sign guard cannot reach that case, and the mixed subnormal/normal recovery is deferred to #2391. That is a pre-existing limitation of the caller's arithmetic, correctly recorded rather than claimed as coverage — I confirmed it is not introduced by this diff.

Change is minimal, correctly scoped to `#2268`, mirrors the existing overflow policy, and is well-tested. No correctness, security, or evidence gaps found.

---
— [eliza-swarm]

tokenmaxxor (approved):
﻿## Review

**Recommendation: merge** exact head `fd8030105a251b4f5b33f8870519a0e226cd0108`.

`_nonzero_magnitude_bits` is the right witness when float32 flushes subnormals so `max(abs)` is 0 while the matrix is not the reserved zero. Tests split destroyed vs surviving scales, keep valid zeros (including -0), and document that dual-update flush happens before the guard. Addresses #2391 without claiming to recover mixed subnormals.

kenjohnscreates (approved):
## Disposition: **approve** (fourth review, added angle: the flush boundary is measured exactly and the well-formed path is byte-identical)

Exact head reviewed: `71bff2dd` (one docs-only commit past the reviewed `fd803010`; no library change since). Executed in an isolated worktree; `tests/test_optimizer_geometry.py` + `tests/test_initializers_validation.py`: 68 passed; ruff clean; mypy clean.

Rank-2 base `[[2, 1], [0.5, -1]]` scaled, through `spectral_matrix_sign_transaction`:

| scale | float32 norm | main | this head |
|---|---|---|---|
| 1, 1e-10, 1e-20, 1e-30 | normal ... 0.0 (norm underflows at 1e-30 but entries survive) | valid, rank-2 orthogonal factor | identical |
| `finfo.tiny` (entries still normal) | 0.0 | valid, rank-2 | identical |
| `tiny/4`, `1e-40`, smallest subnormal (every entry subnormal) | 0.0 | **valid=True, zero matrix returned** | valid=False |
| one subnormal entry, rest zero | 0.0 | **valid=True, zero matrix** | valid=False |
| exact zero matrix, `-0.0` matrix | 0.0 | valid, zero | valid, zero (unchanged) |
| `1e30`, `1e38` (overflow) | inf | invalid | invalid (unchanged) |
| NaN / inf entry | | invalid | invalid |
| 300 random well-formed matrices (2x2 to 5x5, scales 1e-3 to 1e3), factor + validity bytes | | `e3dec4333e67` | **`e3dec4333e67`** |

So the bit-pattern witness changes exactly the rows where the backend flushed every magnitude and nothing else: the transaction that certified a rank-0 answer for a rank-2 input now reports it invalid, the zero matrix keeps its documented valid zero, and every normal-magnitude input is byte-identical.

**Note, not blocking.** A matrix with one normal entry and the rest subnormal (`[[1, 1e-40], [1e-40, 1e-40]]`) still validates with a rank-1 factor on both trees; the PR scopes itself to magnitudes the arithmetic has entirely lost, and partial flushes stay a documented limitation of float32 here rather than a laundering of the whole matrix.

---

PR #2397, opened 2026-08-24
fix(evaluation): recover the FLAD divisor when its squares underflow

### LINKED ISSUE #2393: flad_noise_component_transaction drops the whole gradient component when the squared norm underflows
## Summary

## Reproduction

```python
import jax
import jax.numpy as jnp
import numpy as np

)

    denom = float(g @ g)


```

## Why it happens

```python
active = squared_norm > 0.0
```

## Suggested fix

```python
```

| --- | --- | --- | --- |

## Out of scope

## Willingness to fix

ngngocnhan1997-hub:
Fixes #2393.

## What this fixes

`flad_noise_component_transaction` took its inner products on the raw gradient.
For a float32 gradient whose entries sit at or below roughly `1e-20` every square
lands under the subnormal floor, `jnp.vdot` accumulates exactly zero, and
`squared_norm > 0.0` — the protocol's zero-gradient reduction — fires on a
perfectly well-conditioned direction. The function then returned the perturbation
with its whole gradient component still in it, reporting `valid=True`.

The component that should have been removed is exactly the size of the thing the
function exists to compute. For the reproduction input the projection is
`3.20000e-01` and none of it was removed.

The numerator survives at those scales, so only the divisor was lost. When the
raw divisor is unusable the inner products are retaken against an exactly
power-of-two rescaled copy of the gradient. That is a reformulation rather than an
approximation: the span being projected onto does not depend on the gradient's
magnitude.

## Reproduction

`perturbation = [1.0, -2.0, 0.5, 3.0]`, `gradient = [2.0, 1.0, -1.0, 0.5] × scale`,
error measured as the maximum absolute difference against the same projection in
float64. JAX 0.11.0, CPU, x64 disabled, identical eagerly and under `jax.jit`.
The `before` column is `c9aba7b5`.

| gradient scale | `g @ g` (float32) | before | after | disposition |
| 1.0 | 6.25000e+00 | 8.58307e-08 | 8.58307e-08 | bit-identical |
| 1e-10 | 6.25000e-20 | 8.58307e-08 | 8.58307e-08 | bit-identical |
| 1e-19 | 6.25000e-38 | 6.09524e-02 | 6.09524e-02 | bit-identical |
| 1e-20 | 0.00000e+00 | **3.20000e-01** | **8.58307e-08** | repaired |
| 1e-25 | 0.00000e+00 | **3.20000e-01** | **8.58307e-08** | repaired |
| 1e-30 | 0.00000e+00 | **3.20000e-01** | **8.58307e-08** | repaired |
| 1e-38 (mixed subnormal) | 0.00000e+00 | 3.20000e-01 | 3.20000e-01 | bit-identical, rescale withheld |
| 2.938736e-39 (all subnormal) | 0.00000e+00 | 3.20000e-01 | 3.20000e-01 | bit-identical, rescale withheld |
| 1e19 | inf | refused | **8.58307e-08** | now answered |
| 1e20 | inf | refused | **8.58307e-08** | now answered |
| 1e30 | inf | refused | **8.58307e-08** | now answered |
| 1.7e38 | inf | refused | **8.58307e-08** | now answered |

No row is made worse. The zero gradient still returns the perturbation bit for
bit, and `nan` or `inf` in either argument is still reported invalid with a zeroed
result.

## Why the rescale is gated twice

**It runs only where the raw divisor is dead — exactly zero, or non-finite.** A
gradient whose divisor is already a positive finite number is answered from that
divisor, so its result is not re-rounded onto the grid the rescaled inner
products land on. Three scales are pinned by their recorded bit patterns to hold
that line.

**It is withheld from any gradient holding a subnormal entry.** This is the part
that is not obvious, and it is the correction I posted on #2393 after measuring
rather than assuming. `jnp.ldexp` returns a subnormal operand **bit-for-bit
unmodified**, at every shift amount, identically eagerly and under `jax.jit`:

| operand | `ldexp(x, 125)` returned | mathematically correct result |
| --- | --- | --- |
| `0x00800000` smallest normal | `0x3f000000` | `0x3f000000` |
| `0x00400000` subnormal | `0x00400000` | `0x3e800000` |
| `0x00000001` subnormal floor | `0x00000001` | `0x33800000` |

Both correct results are ordinary normal floats, so the shift does not underflow —
it does not happen. Rescaling a vector that mixes normal and subnormal entries
therefore moves only the normal ones:

```
scale = 1e-38, frexp exponent = -125
  input  bits : ['0x00d9c7dc', '0x006ce3ee', '0x806ce3ee', '0x003671f7']
  scaled bits : ['0x3f59c7dc', '0x006ce3ee', '0x806ce3ee', '0x003671f7']
  moved       : [True, False, False, False]
```

An ungated rescale takes that row from `3.20e-01` to `6.80e-01`, because the
vector it normalizes points somewhere the input did not. Withholding it keeps
that input's previous answer instead.

## Why detection has to be bitwise

No float comparison can see a subnormal on this backend. For a float32
subnormal operand, measured identically eagerly and under `jax.jit`:

| operation | subnormal operand |
| --- | --- |
| `x > 0.0`, `x != 0.0` | `False` |
| `x * 1.0`, `x + 0.0` | flushed to `0x00000000` |
| `jnp.abs` | bits preserved |
| `jnp.frexp` | `(0.5, -149)` for every subnormal |
| `jnp.ldexp` | operand returned unmodified |
| bitcast, mask sign bit, compare | correct |

The round trip `jnp.ldexp(*jnp.frexp(x))` also *appears* exact for subnormals,
which is a trap: the reassembled value is `0x00000000` and the equality that
checks it flushes `x`, so `0x00000000 == subnormal` evaluates `True`. The
exponent field has to be read from the bits.

The helper masks the sign bit so both signed zeros read as normal, and derives
the smallest normal magnitude from the gap between the bit patterns of `1.0` and
`2.0`, which is one mantissa width for any IEEE binary format. That keeps it
correct for `float16`, `bfloat16` and `float32` alike without an untyped
introspection call, which matters because `mypy` runs strict here.

## Disposition change

Overflowing gradients now answer instead of refusing. Refusal is keyed on whether
the resulting coefficient is representable rather than on whether the gradient's
squared norm overflowed, so a huge gradient against an ordinary perturbation is
answered and a huge gradient against a huge perturbation still refuses. Both
halves are pinned by tests. If you would rather keep the refusal, that is a
two-line change to the gate.

## Boundaries this does not cross

Named rather than silently inherited:

- **A degraded divisor is not repaired, only a dead one.** At float32 scale
  `1e-19` the divisor is a barely-normal `6.25e-38` assembled from subnormal
  partial products, and roughly 19% of the projection is already gone. Repairing
  it means re-rounding an input that answers today, which is a disposition call
  rather than a bug fix. The same row exists in `float16` at scale `2e-4`, where
  the divisor is `2.38e-07`.
- **A gradient spanning more than the float32 exponent range loses its smallest
  entries to the rescale.** Their contribution to both inner products is far
  below the rounding of the dominant terms, so the projection is unaffected, but
  the entries are gone.
- **The flushed-operand class stays as it was.** A gradient carrying subnormal
  entries keeps its previous answer here; whether that should instead fail closed
  is #2391's question and #2392's proposal, and duplicating that judgement in a
  second function before it is reviewed seemed worse than leaving it.

The bit witness overlaps `_nonzero_magnitude_bits` from #2392. They answer
different questions — "any magnitude bit set" versus "any entry below the
smallest normal" — but if both land, one should absorb the other.

## Verification

| check | command | result |
| --- | --- | --- |
| geometry suite | `pytest tests/test_optimizer_geometry.py` | 41 passed, 2 failed |
| same suite on `c9aba7b5` | `pytest tests/test_optimizer_geometry.py` | 26 passed, 2 failed |
| lint | `ruff check .` | all checks passed |
| types | `mypy` | 108 errors, unchanged from `c9aba7b5` |

The two failures are `test_geometry_result_has_canonical_bytes_and_exclusive_retention`
and `test_geometry_retention_rejects_namespace_symlink`. Both require an absolute
POSIX repository root and fail identically on the pristine revision on this host;
they are not attributable to this change. The 15 added nodes are the 41 minus the
26 the pristine revision passes.

`mypy` sits at 108 on both revisions. An earlier revision of this branch reached
109 by calling an untyped `finfo` helper in a typed context; that call was removed
rather than suppressed.

The existing `jax.jacrev` test over the zero-gradient path still passes, so the
bitwise gate did not break the derivative, and the change stays outer-`jit` safe.

— [fix]

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -277,24 +277,6 @@ def flad_noise_component_transaction(perturbation: Array, gradient: Array) -> tu
     )
     safe = jnp.where(valid, candidate, jnp.zeros_like(candidate))
     return safe, valid
-    numerator = jnp.vdot(scaled, delta).real
-    active = squared_norm > 0.0
-    denominator = jnp.where(active, squared_norm, jnp.ones_like(squared_norm))
-    coefficient = numerator / denominator
-    projection = scaled * coefficient * active.astype(delta.dtype)
-    candidate = delta - projection
-    valid = (
-        jnp.all(jnp.isfinite(delta))
-        & jnp.all(jnp.isfinite(direction))
-        & jnp.all(jnp.isfinite(scaled))
-        & jnp.isfinite(squared_norm)
-        & jnp.isfinite(numerator)
-        & jnp.isfinite(coefficient)
-        & jnp.all(jnp.isfinite(projection))
-        & jnp.all(jnp.isfinite(candidate))
-    )
-    safe = jnp.where(valid, candidate, jnp.zeros_like(candidate))
-    return safe, valid
 
 
 def muon_ogd_dual_update_transaction(
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -197,22 +197,96 @@ def spectral_matrix_sign(matrix: Array, *, steps: int = 5) -> Array:
     )
 
 
+def _subnormal_entry_present(value: Array) -> Array:
+    """Report whether any entry is a nonzero subnormal.
+
+    A power-of-two rescale cannot move such an entry. ``jnp.ldexp`` returns a
+    subnormal operand bit-for-bit unmodified on a backend that flushes subnormal
+    arithmetic operands, so rescaling a vector that mixes normal and subnormal
+    entries shifts only the normal ones and silently changes the direction the
+    vector points in. No float comparison detects the condition, because
+    ``x != 0.0`` is ``False`` for every float32 subnormal, so the exponent field
+    is read directly and the sign bit masked off to keep both signed zeros
+    reading as normal.
+    """
+    width = value.dtype.itemsize
+    unsigned = np.dtype(f"uint{8 * width}")
+    magnitude = jnp.bitwise_and(
+        jax.lax.bitcast_convert_type(value, unsigned),
+        jnp.asarray((1 << (8 * width - 1)) - 1, dtype=unsigned),
+    )
+    # An IEEE binary exponent field is one wider at 2.0 than at 1.0, so the gap
+    # between their bit patterns is the smallest normal magnitude for this dtype
+    # without asking an untyped introspection helper for the mantissa width.
+    landmarks = jax.lax.bitcast_convert_type(jnp.asarray([1.0, 2.0], dtype=value.dtype), unsigned)
+    smallest_normal = landmarks[1] - landmarks[0]
+    return jnp.any((magnitude != 0) & (magnitude < smallest_normal))
+
+
 def flad_noise_component_transaction(perturbation: Array, gradient: Array) -> tuple[Array, Array]:
-    """Return a finite FLAD noise component and caller-visible validity bit."""
+    """Return a finite FLAD noise component and caller-visible validity bit.
+
+    The projection onto the gradient's span does not depend on the gradient's
+    magnitude, so when the raw divisor is unusable the inner products are retaken
+    against an exactly power-of-two rescaled copy. Every gradient whose divisor is
+    already a positive finite number is answered from that divisor unchanged, and
+    the zero-gradient reduction stays exact.
+    """
     delta = _trusted_array(perturbation, name="perturbation")
     direction = _trusted_array(gradient, name="gradient")
     if delta.shape != direction.shape or delta.ndim != 1 or delta.size < 1:
         raise ValueError("perturbation and gradient must be non-empty equal-width vectors")
-    squared_norm = jnp.vdot(direction, direction).real
-    numerator = jnp.vdot(direction, delta).real
+    # `active` below is the protocol's zero-gradient reduction, so it must answer
+    # whether the gradient is zero and nothing else. Taken on the raw gradient it
+    # cannot: for float32 entries at or below roughly 1e-20 the squares land under
+    # the subnormal floor and `jnp.vdot` accumulates exactly zero, which sent a
+    # well-conditioned direction down the zero-gradient branch and returned the
+    # perturbation with its whole gradient component still in it. The numerator
+    # survives at those scales, so only the divisor was lost. Rescaling by an exact
+    # power of two recovers it without changing the projection, because the span
+    # being projected onto is invariant to the gradient's magnitude, and the same
+    # rescale keeps the squares finite for gradients large enough to overflow them.
+    #
+    # Two conditions narrow that rescale to the gradients it repairs. It runs only
+    # where the raw divisor is unusable, so every gradient that already produced an
+    # answer keeps producing bit-for-bit that answer instead of one re-rounded on a
+    # different grid. It is also withheld from a gradient holding subnormal entries,
+    # which a power-of-two rescale moves the normal entries of while leaving the
+    # subnormal ones in place, quietly changing the direction being projected onto.
+    raw_squared_norm = jnp.vdot(direction, direction).real
+    unusable = (raw_squared_norm == 0.0) | jnp.logical_not(jnp.isfinite(raw_squared_norm))
+    _, exponent = jnp.frexp(jnp.max(jnp.abs(direction)))
+    rescale = unusable & jnp.logical_not(_subnormal_entry_present(direction))
+    scaled = jnp.where(rescale, jnp.ldexp(direction, -exponent), direction)
+    squared_norm = jnp.vdot(scaled, scaled).real
+    numerator = jnp.vdot(scaled, delta).real
     active = squared_norm > 0.0
     denominator = jnp.where(active, squared_norm, jnp.ones_like(squared_norm))
     coefficient = numerator / denominator
-    projection = direction * coefficient * active.astype(delta.dtype)
+    projection = scaled * coefficient * active.astype(delta.dtype)
     candidate = delta - projection
     valid = (
         jnp.all(jnp.isfinite(delta))
         & jnp.all(jnp.isfinite(direction))
+        & jnp.all(jnp.isfinite(scaled))
+        & jnp.isfinite(squared_norm)
+        & jnp.isfinite(numerator)
+        & jnp.isfinite(coefficient)
+        & jnp.all(jnp.isfinite(projection))
+        & jnp.all(jnp.isfinite(candidate))
+    )
+    safe = jnp.where(valid, candidate, jnp.zeros_like(candidate))
+    return safe, valid
+    numerator = jnp.vdot(scaled, delta).real
+    active = squared_norm > 0.0
+    denominator = jnp.where(active, squared_norm, jnp.ones_like(squared_norm))
+    coefficient = numerator / denominator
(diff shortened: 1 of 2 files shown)


agent-cortex (changes requested):
## Review — request changes

Thanks for the careful writeup on #2393; the numerical reasoning behind the power-of-two rescale is sound and I verified the scale-invariance algebra statically (below). However, the diff ships a block of **unreachable duplicate code** that must be removed before this can merge.

### Blocking: dead code after the first `return` in `flad_noise_component_transaction`

`alberta_framework/evaluation/optimizer_geometry.py` lines **279–297**: the function returns at line 279 (`return safe, valid`), and lines **280–297** are an exact duplicate of the answer block at lines 261–279, ending in a second `return safe, valid` at line 297. Everything after the first return is unreachable.

```python
    safe = jnp.where(valid, candidate, jnp.zeros_like(candidate))
    return safe, valid            # <-- line 279, function returns here
    numerator = jnp.vdot(scaled, delta).real   # <-- line 280, unreachable
    active = squared_norm > 0.0
    ...
    safe = jnp.where(valid, candidate, jnp.zeros_like(candidate))
    return safe, valid            # <-- line 297, dead
```

This is new to this PR. At the base revision `c9aba7b5` the function has a single clean tail ending in one `return safe, valid`; the duplication looks like a rebase/paste accident when the new rescale logic was inserted.

**Consequence:** behavior is not wrong at runtime (the first return wins), but merging ~18 lines of unreachable duplicated numerical code into a correctness-critical evaluation path is not acceptable — it is confusing, invites a future editor to "fix" the dead branch, and signals the diff was assembled incorrectly.

**Why none of your reported gates caught it:**
- The added tests all pass precisely *because* the block is unreachable — they exercise only the first return, so they cannot detect the duplication.
- `ruff` lint selects `["E","F","I","N","W","UP"]` (pyproject `[tool.ruff.lint]`); none of these include an unreachable-code-after-`return` rule, so `ruff check .` stays green.
- `mypy` is `strict = true` but `strict` does **not** enable `warn_unreachable`, so the dead block does not move the 108-error count you reported.

**Repair:** delete lines 280–297 (the second copy of the answer block, from `numerator = jnp.vdot(scaled, delta).real` through the second `return safe, valid`). No other change is needed; the reachable path at 250–279 is complete.

### What I did verify (static + algebraic — sandbox execution proof was blocked)

Execution proof is blocked for this PR on my side: the sandbox image has no `pip`/`ensurepip`/`uv` and no JAX, and ships Python 3.11.2 while the project requires `>=3.12` (`pyproject.toml`), so I could not install the JAX stack to run `pytest tests/test_optimizer_geometry.py`. I therefore validated the math and structure by reading:

- **Scale invariance is correct.** With `scaled = direction · 2^k`, `projection = scaled · (⟨scaled,delta⟩ / ⟨scaled,scaled⟩) = direction·2^k · (2^k⟨d,δ⟩)/(2^{2k}⟨d,d⟩) = direction · ⟨d,δ⟩/⟨d,d⟩`, independent of `k`. The rescale recovers the divisor without changing the projected direction, matching the claim.
- **Gating is right.** `unusable = (raw_squared_norm == 0.0) | ¬isfinite`; `rescale = unusable & ¬_subnormal_entry_present(direction)`. A usable positive-finite divisor keeps `scaled == direction` and is answered unchanged; the subnormal-carrying case is withheld so a power-of-two shift cannot silently rotate the direction.
- **Zero-gradient path stays exact.** For zeros, `_subnormal_entry_present` is False (guarded by `magnitude != 0`), `frexp(0)` gives exponent 0, `ldexp(0,0)=0`, so `active=False` and the perturbation is returned bit-for-bit — consistent with `test_flad_zero_gradient_reduces_to_the_perturbation_bit_for_bit`.
- `_subnormal_entry_present` correctly masks the sign bit and derives `smallest_normal` from the bit gap between `1.0` and `2.0`, which holds for float16/bfloat16/float32.

I could not independently execute the reproduction table or the added parametrized tests; please treat the numerical claims as reviewed-by-reasoning, not reproduced. Once the dead block is removed I'm happy to re-review.

---
Contribution skill revision: elizaOS/eliza@e3e63588fd59ebb229d6ac6c8706659058f4fa35:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

ngngocnhan1997-hub:
Blocking finding addressed at head `b1f3a3ca`: the unreachable duplicate answer block (former lines 280–297 of `alberta_framework/evaluation/optimizer_geometry.py`) is deleted; the function now ends at its first `return safe, valid`. Pure dead-code removal, no behavior change — `tests/test_optimizer_geometry.py` passes 41/43 on my machine, and the two failures (`test_geometry_result_has_canonical_bytes_and_exclusive_retention`, `test_geometry_retention_rejects_namespace_symlink`) reproduce identically on the unmodified base head `f50dbc36` under Windows (symlink/retention semantics), so they are environmental and untouched by this delta. The scale-invariance algebra you verified statically is unchanged. Re-review welcome.

lumix5 (approved):

**Disposition: merge.**

- **Bug is real and reproduced by the suite's own precondition asserts:** for float32 gradient entries near 1e-20 the squared norm underflows to exactly 0.0, sending a well-conditioned direction down the zero-gradient branch — the perturbation keeps its entire gradient-aligned component. The repair is mathematically sound: the projection onto the gradient's span is invariant to the gradient's magnitude, so retaking both inner products against an exactly-power-of-two rescaled copy (`ldexp` by `-frexp(max|g|)` -> max entry in [1,2)) recovers the same projection with finite arithmetic; algebraically scaled*(<scaled,d>/<scaled,scaled>) is identically g*(<g,d>/<g,g>).
- **The two withholding conditions are the right ones and both are pinned:** (1) rescale runs only when the raw divisor is zero or non-finite, and the exact-hex-word tests prove usable-divisor answers are bit-for-bit unchanged (including the 1e-19 row that deliberately keeps its subnormal-partial-product rounding); (2) `_subnormal_entry_present` reads the exponent field via bitcast with the bits(2.0)-bits(1.0) smallest-normal landmark (correctly dtype-generic, works for bfloat16), because a power-of-two rescale cannot move a subnormal operand on a flush-to-zero backend and would silently change the projected direction — subnormal gradients keep the conservative return-unchanged behavior, pinned exactly.
- **Ran the full patched suite (43 tests) in a throwaway container** (network denied, branch module overlaid on unchanged main): all pass — underflow scales (1e-20/25/30) and overflow scales (1e19 to 1.7e38) both recover the float64 reference projection; the unrepresentable-coefficient case stays invalid; zero-gradient reduces bit-for-bit; bfloat16 underflow repaired within its precision. ci-ok green at head.

Non-blocking observation: `jnp.frexp(max|g|)` is computed on every call even when the divisor is usable (its value discarded under `where`); harmless under jit, and hoisting it would only matter on a profiled hot path.

lalalune:
Closing in favor of merged #2430. This version deliberately pins an inaccurate 1e-19 projection bit-for-bit whenever the raw squared norm remains positive, instead of correcting the lost precision; its test comments explicitly preserve that loss. It also treats gradients containing subnormals as valid unchanged perturbations. The selected unconditional exact-power-of-two scaling matches the float64 projection reference across ordinary and tiny normal magnitudes. Further subnormal work should validate projection accuracy rather than freeze the incorrect old answer.

Issue #2409, opened 2026-08-24
Floored masked baseline silently loses centering in three Hedge updates


ngngocnhan1997-hub:
## Summary

Three Hedge-style preference updates build their centering baseline by dividing the
masked-in allocation by a **floored** mass:

```python
finite_weight_sum = jnp.maximum(jnp.sum(jnp.where(mask, weights, 0.0)), 1e-12)
masked_weights = jnp.where(mask, weights / finite_weight_sum, 0.0)
baseline = jnp.sum(masked_weights * adjusted)
```

Once the masked-in entries carry less mass than `1e-12`, the divisor stops being the
mass and `masked_weights` no longer sums to one. The baseline is then pulled toward
zero instead of toward the values that took part, and the centered advantage
degenerates into the raw value. `update_applied` stays `True`, so the update is
committed to state with no signal that centering was lost.

Sites:

- `alberta_framework/core/resource_manager.py:597` — `LearnedResourceManager._update_jit`
- `alberta_framework/core/resource_manager.py:1194` — `GeneratorMetaResourceManager._update_jit`
- `alberta_framework/core/feature_discovery.py:1039` — `FixedBudgetFeatureLearner._resource_log_weight_update`

## Why this breaks a stated contract

`LearnedResourceManager`'s class docstring specifies the update as

```
advantage_i = dot(weights, adjusted_losses) - adjusted_losses_i
```

and states: *"Centering by the allocation baseline keeps the preferences numerically
stable and makes uniform shifts in all losses irrelevant."*

Both clauses require `weights` to sum to one over the participating actions. Below the
floor it does not, so neither clause holds.

## Reproduction

`LearnedResourceManager(n_actions=3, n_contexts=1, exploration=0.0)`, one context,
`log_weights = [[gap, 0.0, 0.0]]`, `update(state, [nan, 1.0, 3.0], 0)`. The leading
action is masked out by its `nan`, so the two survivors have adjusted losses `1.0` and
`3.0`; a baseline that is a convex combination of those two must lie in `[1.0, 3.0]`,
and centering must give `+1.0 / -1.0`.

| leader gap (nats) | masked-in mass | class | `update_applied` | `advantages[1]` | `advantages[2]` | required |
| --- | --- | --- | --- | --- | --- | --- |
| 0.0 | 6.66667e-01 | normal | 1 | 1.00000e+00 | -1.00000e+00 | +1 / -1 |
[...]

PR #2410, opened 2026-08-24
fix(core): renormalize masked baselines instead of flooring the divisor

ngngocnhan1997-hub:
Fixes #2409.

## Problem

Three Hedge-style preference updates centered their advantages against a baseline built
from a **floored** masked mass:

```python
finite_weight_sum = jnp.maximum(jnp.sum(jnp.where(mask, weights, 0.0)), 1e-12)
masked_weights = jnp.where(mask, weights / finite_weight_sum, 0.0)
baseline = jnp.sum(masked_weights * adjusted)
```

- `alberta_framework/core/resource_manager.py:597` — `LearnedResourceManager._update_jit`
- `alberta_framework/core/resource_manager.py:1194` — `GeneratorMetaResourceManager._update_jit`
- `alberta_framework/core/feature_discovery.py:1039` — `FixedBudgetFeatureLearner._resource_log_weight_update`

Once the masked-in entries carry less mass than `1e-12`, the divisor stops being the
mass, so `masked_weights` no longer sums to one. The baseline is then pulled toward zero
instead of toward the values that took part, and `advantage_i` degenerates into the raw
value with a sign flip. `update_applied` stays `True`, so the update is committed to
state with nothing to indicate that centering was lost.

`LearnedResourceManager`'s own class docstring is the contract this breaks:

> `advantage_i = dot(weights, adjusted_losses) - adjusted_losses_i`
>
> Centering by the allocation baseline keeps the preferences numerically stable and
> makes uniform shifts in all losses irrelevant.

Both clauses require `weights` to sum to one over the participating actions.

## Reproduction

`LearnedResourceManager(n_actions=3, n_contexts=1, exploration=0.0)`, one context,
`log_weights = [[gap, 0.0, 0.0]]`, `update(state, [nan, 1.0, 3.0], 0)`. The leading
action is masked out by its `nan`, so the survivors have adjusted losses `1.0` and `3.0`;
a convex baseline over those is `2.0` and centering must give `+1.0 / -1.0`.

| leader gap | masked mass | class | `applied` | before `adv[1]` | before `adv[2]` | after `adv[1]` | after `adv[2]` |
| 0.0 | 6.66667e-01 | normal | 1 | 1.00000e+00 | -1.00000e+00 | 1.00000e+00 | -1.00000e+00 |
| 5.0 | 1.32967e-02 | normal | 1 | 1.00000e+00 | -1.00000e+00 | 1.00000e+00 | -1.00000e+00 |
| 10.0 | 9.07916e-05 | normal | 1 | 1.00000e+00 | -1.00000e+00 | 1.00000e+00 | -1.00000e+00 |
| 20.0 | 4.12231e-09 | normal | 1 | 1.00000e+00 | -1.00000e+00 | 1.00000e+00 | -1.00000e+00 |
| 27.0 | 3.75906e-12 | normal | 1 | 1.00000e+00 | -1.00000e+00 | 1.00000e+00 | -1.00000e+00 |
| 28.0 | 1.38288e-12 | normal | 1 | 1.00000e+00 | -1.00000e+00 | 1.00000e+00 | -1.00000e+00 |
| **30.0** | 1.87152e-13 | normal | 1 | **-6.25695e-01** | **-2.62569e+00** | 1.00000e+00 | -1.00000e+00 |
| **40.0** | 8.49671e-18 | normal | 1 | **-9.99983e-01** | **-2.99998e+00** | 1.00000e+00 | -1.00000e+00 |
| **60.0** | 1.75130e-26 | normal | 1 | **-1.00000e+00** | **-3.00000e+00** | 1.00000e+00 | -1.00000e+00 |
| **80.0** | 3.60970e-35 | normal | 1 | **-1.00000e+00** | **-3.00000e+00** | 1.00000e+00 | -1.00000e+00 |
| **90.0** | 0.00000e+00 | zero | 1 | **-1.00000e+00** | **-3.00000e+00** | 1.00000e+00 | -1.00000e+00 |
| **200.0** | 0.00000e+00 | zero | 1 | **-1.00000e+00** | **-3.00000e+00** | 1.00000e+00 | -1.00000e+00 |

The transition sits exactly on the floor: `1.38e-12 > 1e-12` still centers,
`1.87e-13 < 1e-12` does not. Past it the baseline collapses to `0` and
`advantage_i -> -adjusted_loss_i`.

Uniform shifts, which the docstring says are irrelevant, are absorbed by
`advantage_clip=10.0` into a total loss of ordering — `advantages[1], advantages[2]`:

| leader gap | shift 0 | shift 1 | shift 10 | shift 100 | after (any shift) |
| 0.0 | 1.0000, -1.0000 | 1.0000, -1.0000 | 1.0000, -1.0000 | 1.0000, -1.0000 | 1.0000, -1.0000 |
| 28.0 | 1.0000, -1.0000 | 1.0000, -1.0000 | 1.0000, -1.0000 | 1.0000, -1.0000 | 1.0000, -1.0000 |
| 30.0 | -0.6257, -2.6257 | -1.4385, -3.4385 | -8.7542, **-10.0000** | **-10.0000, -10.0000** | 1.0000, -1.0000 |
| 40.0 | -1.0000, -3.0000 | -2.0000, -4.0000 | **-10.0000, -10.0000** | **-10.0000, -10.0000** | 1.0000, -1.0000 |
| 110.0 | -1.0000, -3.0000 | -2.0000, -4.0000 | **-10.0000, -10.0000** | **-10.0000, -10.0000** | 1.0000, -1.0000 |

At gap 40 with every loss shifted by 10, the manager can no longer distinguish the
loss-1 action from the loss-3 action.

`GeneratorMetaResourceManager` at `exploration=0.0` reproduces identically with the sign
inverted (it maximizes): gap 40 gives `+0.999983 / +2.99998` where `-1 / +1` is required,
and shift 100 saturates both to `+10.0`. After the change it is `-1.0 / +1.0` at every
gap and shift.

## Reachability

`LearnedResourceManager` **defaults** to `exploration=0.0`, so nothing holds its
allocation off the floor. A Hedge run concentrating on one action — the algorithm working
as intended — walks into this range on its own; a 30-nat gap after `learning_rate=1.0`
steps is unremarkable.

`GeneratorMetaResourceManager` (`exploration=0.01`) and `FixedBudgetFeatureLearner`
(`resource_exploration=0.01`) are protected **by the exploration floor, not by this
code** — it holds any non-empty valid subset at `>= exploration / n`, measured
`6.66667e-03` at every gap with correct advantages. That protection disappears at
`exploration=0.0`, an accepted value and the one the repository's own tests use.

In `feature_discovery.py`, the call at line 1788 passes a real `generator_finite` mask
and is exposed; line 1823 passes an all-true mask, where the mass is `1.0` and the floor
never binds.

## The fix

One shared helper in `alberta_framework/core/update_safety.py`, the existing home for
fail-closed numerical update primitives — both consumers already import from it, so no
new dependency edge appears:

```python
participating = jnp.where(mask, weights, jnp.zeros_like(weights))
mass = jnp.sum(participating)
smallest_normal = jnp.asarray(np.finfo(cast(Any, weights.dtype)).tiny, dtype=mass.dtype)
usable = mass >= smallest_normal
...
return jnp.where(
    usable | ~jnp.isfinite(mass),
    participating / jnp.where(usable, mass, jnp.ones_like(mass)),
    uniform,
)
```

Three decisions:

1. **Divide by the raw mass over the normal-float32 domain.** Where the floor never
   bound, `max(mass, 1e-12) == mass` and `where(usable, mass, 1.0) == mass` — the same
   op on the same inputs, so those rows are bit-for-bit unchanged. Every row the floor
   did reach becomes a true convex combination.
2. **Fall back to a uniform allocation below the smallest normal** (`1.17549435e-38`).
   The backend has already returned exact zeros for operands down there, so no relative
   size survives to recover, and uniform is the only convex answer left. The old floor
   sat 26 orders of magnitude *above* that point, discarding a very large range of
   perfectly representable masses.
3. **Pass a non-finite mass through unchanged.** Both callers already have a finiteness
   gate that neutralizes the update; swallowing the `nan` here would hide the condition
   from it.

Rejected:

- **Lower the floor to the smallest normal.** Still a floor, so the row still sums to
  less than one across the range it covers; it moves the boundary rather than removing
  the defect.
- **Accumulate the mass in float64.** `x64` is disabled in this project; enabling it
  globally to fix one divisor is far larger than the defect warrants.
- **Fall back to the unmasked row.** That readmits the masked-out entries into a
  baseline defined over the participating ones only.

Two implementation notes: `np.finfo` is used rather than `jnp.finfo` because `mypy`
strict rejects the latter as an untyped call; and the `LearnedResourceManager` class
docstring now states that the advantage weights are renormalized over the finite-loss
actions, since it previously specified the advantage without naming the step the defect
broke.

## Verification

Two backend facts shaped the assertions, both measured here (`jax 0.11.0`, CPU, `x64`
off):

- Summing float32 subnormals flushes them, so the masked mass jumps from a normal
  `3.29e-38` at gap 87 straight to exactly `0.0` at gap 88 — a subnormal total is
  unreachable through the accumulation the callers use. The subnormal branch is still
  covered directly at the helper's own boundary.
- CPU division is not correctly rounded: with `mass` exactly `2w`, `w / mass` returns
  `0x3effffff` (`0.49999997`), not `0.5`. Repaired rows are therefore asserted with
  tolerances (`atol` 6e-8 on the row sum, 1e-6 / 1e-4 on advantages); exact bit
  assertions are reserved for the rows the change provably does not touch.

**Bit identity where the floor was inactive.** Advantage and `log_weights` words are
identical between the pristine and changed trees at gaps 0, 5, 10, 20, 27 and 28,
compared as raw `uint32` views. Four of those are pinned in
`test_masked_update_is_bit_for_bit_unchanged_where_the_mass_was_representable`:

| gap | advantages | `log_weights` |
| --- | --- | --- |
| 0 | `0x00000000 0x3f800000 0xbf800000` | `0x00000000 0x3f800000 0xbf800000` |
| 10 | `0x00000000 0x3f800000 0xbf800000` | `0x40d44444 0xc0144444 0xc08a2222` |
| 20 | `0x00000000 0x3f800000 0xbf800000` | `0x41544444 0xc0b44444 0xc0f44444` |
| 28 | `0x00000000 0x3f800000 0xbf800000` | `0x41949630 0xc1049630 0xc1249630` |

**Fail-first,** the new tests run against pristine `c9aba7b5` in a separate checkout:

```
pytest tests/test_resource_manager.py   ->  54 failed, 116 passed
pytest tests/test_update_safety.py      ->  collection error: cannot import
                                            masked_convex_weights
```

The 54 failures are 5 distinct tests: `LearnedResourceManager` centering (7 gaps),
uniform-shift invariance (28 cases) and loss-spread bounds (7 gaps);
`GeneratorMetaResourceManager` centering (4 gaps) and uniform-shift invariance (8 cases).
Two new tests pass on pristine *by design* — the recorded-bit-pattern test, whose golden
words were taken from that tree, and the default-exploration-floor test, which asserts
the configuration the floor protects.

**Gates on the change:**

```
ruff check .                                                     All checks passed!
mypy                                            108 errors (unchanged from baseline)
pytest tests/test_resource_manager.py tests/test_update_safety.py  303 passed
pytest the four feature-discovery suites + the pytree-depth suite  206 passed, 2 failed
pytest the six compositional-feature suites                        115 passed
```

The two failures are `test_update_safety_tree_depth.py::test_last_fit_list_nesting_still_walks`
and `::test_last_fit_tuple_and_dict_nests_still_walk`, both a Python recursion limit in
the harness. They reproduce identically on a pristine `c9aba7b5` checkout (`2 failed,
7 passed`) and are unrelated to this change.

**Tests added:** 18 tests, 219 parametrized cases.

- `tests/test_update_safety.py` — 11 unit tests, 115 cases, over the helper directly:
  the row sums to one over the mask; masked-out entries are zero; the baseline is bounded
  by the values averaged; a uniform shift moves the baseline by exactly that shift; the
  plain quotient is matched with exact array equality for a normal mass; an underflowed
  mass falls back to uniform; a subnormal mass falls back to uniform; an empty mask
  returns zeros; a non-finite allocation passes through; the row stays finite and
  non-negative; eager and `jit` agree bit-for-bit.
- `tests/test_resource_manager.py` — 7 behavioural tests, 104 cases, through the public
  `update` API of both managers, including the bit-identity pins above and the
  exploration-floor protection.

Worst advantage error across all repaired rows for uniform shifts up to 1000 is `6.1e-5`.

## Boundaries not crossed

Other `jnp.maximum(jnp.sum(...), ...)` occurrences are a different pattern and are
correct, so they are untouched: count guards of the form `max(sum(mask), 1.0)`, where a
zero count also means a zero numerator, and unmasked softmax denominators
(`associative_memory.py:540`/`548`, `prototype_memory.py:308`), where the numerator is
part of the sum. A `grep` over every occurrence confirms no floored *masked*
renormalization remains.

— [fix]

### DIFF
--- a/alberta_framework/core/feature_discovery.py
+++ b/alberta_framework/core/feature_discovery.py
@@ -47,6 +47,7 @@ from alberta_framework.core.future_utility import (
 )
 from alberta_framework.core.update_safety import (
     floating_tree_is_finite,
+    masked_convex_weights,
     neutralize_array,
     select_transaction,
 )
@@ -1036,8 +1037,7 @@ class FixedBudgetFeatureLearner:
         finite: Array,
     ) -> Array:
         """Exponentiated-gradient preference update for utility scores."""
-        finite_mass = jnp.maximum(jnp.sum(jnp.where(finite, allocation, 0.0)), 1e-12)
-        masked_allocation = jnp.where(finite, allocation / finite_mass, 0.0)
+        masked_allocation = masked_convex_weights(finite, allocation)
         baseline = jnp.sum(masked_allocation * jnp.where(finite, scores, 0.0))
         advantages = jnp.where(finite, scores - baseline, 0.0)
         advantages = jnp.clip(
--- a/alberta_framework/core/resource_manager.py
+++ b/alberta_framework/core/resource_manager.py
@@ -44,6 +44,7 @@ from jaxtyping import Bool, Float, Int
 from alberta_framework.core._float32_scalars import validated_float32_scalar_with_ratio
 from alberta_framework.core.update_safety import (
     floating_tree_is_finite,
+    masked_convex_weights,
     neutralize_array,
     safe_discrete_action,
     select_transaction,
@@ -340,6 +341,12 @@ class LearnedResourceManager:
     Positive advantage means action ``i`` beat the current allocation.  Centering
     by the allocation baseline keeps the preferences numerically stable and
     makes uniform shifts in all losses irrelevant.
+
+    ``weights`` in the advantage above is the emitted allocation renormalized to
+    sum to one over the actions with a finite loss, so the baseline stays a
+    convex combination of exactly the losses that took part.  Both properties
+    above depend on that: an allocation summing to less than one would pull the
+    baseline toward zero and leave the advantage tracking the raw loss.
     """
 
     def __init__(
@@ -594,8 +601,7 @@ class LearnedResourceManager:
         adjusted = jnp.where(valid_actions, safe_losses + cost_terms, 0.0)
 
         weights = self._weights_jit(state, context)
-        finite_weight_sum = jnp.maximum(jnp.sum(jnp.where(valid_actions, weights, 0.0)), 1e-12)
-        masked_weights = jnp.where(valid_actions, weights / finite_weight_sum, 0.0)
+        masked_weights = masked_convex_weights(valid_actions, weights)
         baseline = jnp.sum(masked_weights * adjusted)
         advantages = jnp.where(valid_actions, baseline - adjusted, 0.0)
         advantages = jnp.clip(
@@ -1191,8 +1197,7 @@ class GeneratorMetaResourceManager:
         adjusted = jnp.where(finite, safe_rewards - cost_terms, 0.0)
 
         weights = self._weights_jit(state, context)
-        finite_weight_sum = jnp.maximum(jnp.sum(jnp.where(finite, weights, 0.0)), 1e-12)
-        masked_weights = jnp.where(finite, weights / finite_weight_sum, 0.0)
+        masked_weights = masked_convex_weights(finite, weights)
         baseline = jnp.sum(masked_weights * adjusted)
         selection_input_valid = jnp.asarray(True, dtype=jnp.bool_)
         if self._update_rule == "exp3":
--- a/alberta_framework/core/update_safety.py
+++ b/alberta_framework/core/update_safety.py
@@ -199,6 +199,39 @@ def neutralize_metrics(
     return {name: neutralize_array(applied, value) for name, value in metrics.items()}
 
 
+def masked_convex_weights(mask: Bool[Array, " n"], weights: Array) -> Array:
+    """Renormalize non-negative ``weights`` to sum to one over ``mask``.
+
+    A baseline built from the result is a convex combination of the masked-in
+    values, so it is bounded by them and a uniform shift in all of them moves it
+    by exactly that shift.  Flooring the divisor instead promises neither: once
+    the masked-in entries carry less mass than the floor, the row sums to less
+    than one and the baseline is pulled toward zero rather than toward the values
+    that took part, which turns a centered advantage into the raw value.
+
+    Masked-in entries whose total is not a normal float of ``weights.dtype`` fall
+    back to a uniform allocation.  Every arithmetic operand below the smallest
+    normal is returned as an exact zero by the accelerator backends this runs on,
+    so their relative sizes cannot be recovered here and uniform is the only
+    convex answer left.  A non-finite total is passed through so the caller's
+    finiteness gate still sees it.
+    """
+
+    participating = jnp.where(mask, weights, jnp.zeros_like(weights))
+    mass = jnp.sum(participating)
+    smallest_normal = jnp.asarray(np.finfo(cast(Any, weights.dtype)).tiny, dtype=mass.dtype)
+    usable = mass >= smallest_normal
+    counted = jnp.sum(mask.astype(weights.dtype))
+    uniform = jnp.where(mask, jnp.ones_like(weights), jnp.zeros_like(weights)) / jnp.maximum(
+        counted, jnp.ones_like(counted)
+    )
+    return jnp.where(
+        usable | ~jnp.isfinite(mass),
+        participating / jnp.where(usable, mass, jnp.ones_like(mass)),
+        uniform,
+    )
+
+
 def zero_if_collapsed_infinity(
     product: Array,
     infinite_input: Array,
--- a/tests/test_resource_manager.py
+++ b/tests/test_resource_manager.py
@@ -23,6 +23,10 @@ from alberta_framework.core.compositional_features import (
     CompositionalFeatureLearner,
     run_compositional_arrays,
 )
+from alberta_framework.core.resource_manager import (
+    GeneratorMetaResourceManagerState,
+    LearnedResourceManagerState,
+)
 
 
 class TestLearnedResourceManager:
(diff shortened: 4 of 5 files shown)


agent-cortex (approved):
## Approve — defect and fix independently reproduced in a sandbox

I verified this change end-to-end in an isolated container (Python 3.12.11, `jax 0.11.1` / `jaxlib 0.11.1`, CPU backend, `x64` off — matching the PR's `jax 0.11.0` environment), reproducing the base defect, confirming the fix corrects it, and running the touched suites on the real head. Nothing was executed on the host.

### Scope — matches issue #2409
The diff touches exactly the three floored-divisor sites named in #2409 (`resource_manager.py:601`, `:1197`, `feature_discovery.py:1039`), routes all three through one shared helper `masked_convex_weights` in `update_safety.py` (the existing home for fail-closed numerical primitives — both callers already import from it, so no new dependency edge), and adds the docstring clause the defect broke. No unrelated behavior is touched.

### Defect reproduced on base (`c9aba7b5`)
Reverting only the three source files to base while keeping the head tests, the head test suite fails exactly as the PR claims — and the failure output pins the concrete consequence:
```
tests/test_resource_manager.py   -> 54 failed, 116 passed
tests/test_update_safety.py      -> ImportError: cannot import name 'masked_convex_weights' (collection error)

# representative failure (GeneratorMeta, gap 110, uniform +1000 shift):
ACTUAL:  array([ 0., 10., 10.], dtype=float32)   # ordering lost, both clipped to advantage_clip
DESIRED: array([0., 1., 3.], dtype=float32)
```

### Independent minimal reproduction (not the PR's tests)
I wrote my own script comparing the old floored formula against the new helper on the same masked softmax allocation:
```
  gap         mass  OLD rowsum  OLD base  NEW rowsum  NEW base
    0   6.6667e-01  1.0000e+00    2.0000  1.0000e+00    2.0000
   28   1.3829e-12  1.0000e+00    2.0000  1.0000e+00    2.0000
   30   1.8715e-13  1.8715e-01    0.3743  1.0000e+00    2.0000
   40   8.4967e-18  8.4967e-06    0.0000  1.0000e+00    2.0000
   90   0.0000e+00  0.0000e+00    0.0000  1.0000e+00    2.0000
```
This confirms every load-bearing claim: (1) the transition sits exactly on the `1e-12` floor — mass `1.38e-12` still centers, `1.87e-13` collapses; (2) past it the OLD baseline is pulled toward `0`, degenerating the centered advantage into the raw loss; (3) the NEW helper keeps a true convex baseline (`2.0`) across the whole range; (4) the fix is **bit-identical where the floor was inactive** (gaps 0 and 28 give the same row-sum/baseline under both formulas), so no existing behavior regresses.

### Fix corrects it + focused tests pass on head
```
tests/test_update_safety.py + tests/test_resource_manager.py  -> 303 passed
tests/test_feature_discovery.py                               -> 148 passed
```

### Logic audit of `masked_convex_weights`
- Divides by the raw mass over the normal-float32 domain; where the old floor never bound, `where(usable, mass, 1) == mass`, i.e. the identical op on identical inputs — matching the observed bit-identity.
- Sub-`tiny` mass falls back to a uniform allocation over the mask; since float32 subnormal operands are flushed to exact zero by the backend, relative sizes are already unrecoverable and uniform is the only convex answer. Empty mask returns zeros (`counted` guarded by `max(·, 1)`).
- Non-finite mass (masked-in `nan`/`inf`) passes through unchanged via the `~jnp.isfinite(mass)` branch, so the callers' existing finiteness gate still neutralizes the update rather than having it silently masked here. Verified by `test_masked_convex_weights_pass_a_non_finite_allocation_through`.
- `np.finfo(...).tiny` (not `jnp.finfo`) is a correct choice under mypy-strict.

### Tests
The added tests are genuine behavioral assertions through the public `update` API at the reachable `exploration=0.0` default (not surface-widening): convex-baseline centering, uniform-shift invariance, participating-loss-spread bounds, and recorded-`uint32`-bit pins for the inactive-floor rows. Tolerances are justified by CPU division not being correctly rounded (`w/2w -> 0.49999997`), with exact assertions reserved for provably untouched rows.

I found no correctness, security, contract, or required-evidence gap. The change is minimal, correctly scoped to #2409, numerically sound, and preserves prior behavior where the floor was inactive.

---
— [eliza-swarm]

kenjohnscreates (approved):
## Disposition: **approve** (issue author's fix, first-filed, one shared helper for all three sites, only head that does not launder a non-finite allocation, strongest tests)

Exact head reviewed: `eeca50f8` on base `c9aba7b5`. Filed 2026-08-24 17:43 UTC, nine minutes after the issue; agent-cortex's claim comment (17:36) never produced a PR.

The defect reproduces on main exactly as tabulated in the issue: at gap 30 the mass `1.87e-13` is below the floor and the advantages are `-0.63,-2.63`; from gap 40 they are `-1,-3` (raw loss with the sign flipped); a shift of 10 at gap 40 saturates both survivors to `-10`. On this head every equal-survivor row is `+1,-1` through gap 200, every shift row is `+1,-1`, the unequal-survivor rows `[gap,0,-1]` give `+0.5379,-1.4621` up to gap 60 and `[gap,0,-5]` gives `+0.0134,-1.9866` at gap 40, `GeneratorMetaResourceManager` mirrors all of it with the sign inverted, and the `FixedBudgetFeatureLearner` site gives a convex allocation at `[1,3e-13,1e-13]` and uniform at `finfo.tiny`, `tiny/2` and zero mass. The 2000-row well-formed sweep and the 300-step trajectory hash are byte-identical to main.

**Why this head over the other nine.** (1) One helper, `masked_convex_weights` in `update_safety.py`, replaces all three floored sites, with the validity rule (`mass >= finfo.tiny`, non-finite passed through) documented once; the other heads inline the logic three times or add a private helper per module. (2) It is the only head that passes a NaN or inf allocation through rather than replacing it with a uniform row, so the callers' existing finiteness gate still sees the fault; the other nine turn a corrupt allocation into a plausible update. (3) Its tests cover all three sites: seven cases in `tests/test_resource_manager.py` (both managers, gaps 30 to 200, shifts to 1000, the loss-spread bound, a bit-for-bit pin where the mass was representable, and the exploration-floor guarantee) plus eleven property tests on the helper in `tests/test_update_safety.py`. Mutation check against main's library: 54 of the new resource-manager cases fail, so they discriminate. Touched test files in this head's worktree (`test_resource_manager.py`, `test_update_safety.py`, `test_feature_discovery.py`): 451 passed; ruff clean.

**Not blocking, shared by every head.** Renormalizing the leftover linear-space mass is not the conditional softmax once the survivors' own weights flush: at `[gap,0,-5]` with gap 90 or more this head returns uniform `+1,-1` where `softmax(where(valid, log_weights, -inf))` would give `+0.0134,-1.9866`. The issue text accepts uniform there; a log-space masked softmax would be the stronger follow-up.

agent-cortex approved this head; attentionhead's comment on #2526 already named it as the owner of the three call sites.

## Method

All ten PRs citing #2409 (#2410, #2428, #2461, #2494, #2513, #2526, #2568, #2746, #2766, #2784; #2510 also cites it but changes `horde_actor_critic.py` and is a different issue) were fetched to isolated worktrees at their exact heads; `alberta_framework.__file__` was asserted to resolve inside each worktree before import. Every library diff was read against `origin/main` (`c9aba7b5`) before execution, and the three heads with reformatting (#2428, #2568, #2784) were compared to main at the AST level with positions stripped. One probe ran on main and every head under single-threaded XLA: `LearnedResourceManager` (exploration 0) with the issue's `log_weights=[gap,0,0]`, `losses=[nan,1,3]` at 13 gaps from 0 to 200 and four uniform shifts at gap 40; unequal survivors `[gap,0,-1]` and `[gap,0,-5]` at gaps up to 200; all-NaN losses; a single survivor at gap 200; `GeneratorMetaResourceManager` (exploration 0) equal and unequal survivors; `FixedBudgetFeatureLearner._resource_log_weight_update` driven directly with normal, sub-1e-12, `finfo.tiny`, `tiny/2`, 1e-30, zero-mass, all-masked, NaN, inf and negative allocations; a 2000-row well-formed sweep (random log-weights and losses with NaN masks, mass above 1e-12; sha1 of advantages and new log-weights); and a 300-step `LearnedResourceManager` trajectory hash. Each head's `tests/test_resource_manager.py`, `tests/test_update_safety.py` and `tests/test_feature_discovery.py` were run in its own worktree, and #2410's new tests were run against main's library as a mutation check.

|---|---|---|---|---|

---

lalalune:

Issue #2411, opened 2026-08-24
Latent encoder normalizes by 1e-6 instead of the observation_scale the config accepted


ngngocnhan1997-hub:
## Summary

`LatentWorldModelConfig.observation_scale` is documented as the per-dimension scale the
encoder divides by, and its validator admits any positive `float32`. The encoder divides
by something else:

```python
# alberta_framework/core/latent_world_model.py, _encode_with
scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
scaled = obs / jnp.maximum(scale, jnp.asarray(1e-6, dtype=jnp.float32))
```

For every accepted `observation_scale` below `1e-6` the encoder normalizes by `1e-6`
instead of the declared value. Nothing reports the substitution, and the latents shrink
by the ratio between the two numbers.

The sibling `reward_scale` in the same class is the counter-example: line 752 divides by
the raw configured value and line 781 multiplies by it, with no floor on either side. The
floored `observation_scale` is the outlier, not the convention.

## Why it matters

The shrink is not a rounding detail; it walks through the module's own diagnostics and
disables a documented feature.

Observations drawn from one standard normal and multiplied by `observation_scale` are
identically well-conditioned in declared units. Forty transitions through
`LatentWorldModel.update` on `c9aba7b5`, `observation_dim=3`, `latent_dim=6`,
`min_latent_std=0.05`:

| `observation_scale` | mean latent std | `collapse_score` | encoder moved (`encoder_learning=True`) | max abs encoder delta |
| --- | --- | --- | --- | --- |
| `1.0` | 6.054746e-01 | 0.0000 | True | 4.293072e-02 |
| `1e-3` | 6.054746e-01 | 0.0000 | True | 4.293072e-02 |
| `1e-6` | 6.054746e-01 | 0.0000 | True | 4.293072e-02 |
| `1e-7` | 1.020295e-01 | 0.1667 | **False** | 0.000000e+00 |
| `1e-9` | 1.036086e-03 | **1.0000** | **False** | 0.000000e+00 |
| `1e-12` | 1.036088e-06 | **1.0000** | **False** | 0.000000e+00 |
| `1e-20` | 1.036088e-14 | **1.0000** | **False** | 0.000000e+00 |

Three consequences, in order:

1. **The declared normalization is silently replaced.** At `1e-9` an observation the
   config places at `2.0` normalized units reaches the encoder as `2e-3`, a 1000x shrink.
[...]

PR #2412, opened 2026-08-24
fix(core): normalize latent encoder observations by declared scale without 1e-6 floor

Kayvan-Zahiri:
Fixes #2411

`LatentWorldModelConfig.observation_scale` was previously divided with a floor of `jnp.maximum(scale, 1e-6)` in `_encode_with`, silently replacing any declared `observation_scale` below `1e-6` with `1e-6` and corrupting latent representations and causing false `collapse_score` flags.

### Changes
1. Removed `jnp.maximum(scale, 1e-6)` in `LatentWorldModel._encode_with`, dividing directly by the declared `scale`.
2. Constrained `LatentWorldModelConfig.observation_scale` validation in `__post_init__` to `lower=_FLOAT32_MIN_NORMAL` (`1.1754943508222875e-38`, `float.fromhex("0x1.0p-126")`), ensuring every admitted scale has a finite float32 reciprocal and rejecting subnormal inputs fail-closed.
3. Added unit tests in `tests/test_latent_world_model.py`:
   - Validating that scales from `1.0` down to `_FLOAT32_MIN_NORMAL` (including `1e-6`, `1e-9`, `1e-20`) produce exact scale-invariant latent encodings without truncation/flooring.
   - Validating that subnormal scales are rejected with `ValueError`.

### Verification
- `pytest tests/test_latent_world_model.py`: 58/58 tests passed.
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/core/latent_world_model.py
+++ b/alberta_framework/core/latent_world_model.py
@@ -76,6 +76,7 @@ from alberta_framework.core.update_safety import (
     select_transaction,
 )
 
+_FLOAT32_MIN_NORMAL: float = float.fromhex("0x1.0p-126")
 EVIDENCE_LEVEL = "L0"
 SCIENTIFIC_PROMOTION_ALLOWED = False
 _INT32_MAX = 2**31 - 1
@@ -210,7 +211,9 @@ class LatentWorldModelConfig:
                 raise ValueError("observation_scale length must equal observation_dim")
             observation_scale = tuple(
                 validated_float32_scalar(
-                    f"observation_scale[{index}]", scale, positive=True
+                    f"observation_scale[{index}]",
+                    scale,
+                    lower=_FLOAT32_MIN_NORMAL,
                 )
                 for index, scale in enumerate(observation_scale)
             )
@@ -589,7 +592,7 @@ class LatentWorldModel:
         """Encode one observation with explicit (differentiable) encoder params."""
         obs = jnp.asarray(observation, dtype=jnp.float32).reshape((self._config.observation_dim,))
         scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        scaled = obs / jnp.maximum(scale, jnp.asarray(1e-6, dtype=jnp.float32))
+        scaled = obs / scale
         return jnp.tanh(scaled @ encoder_matrix + encoder_bias)
 
     @functools.partial(jax.jit, static_argnums=(0,))
--- a/alberta_framework/core/latent_world_model.py
+++ b/alberta_framework/core/latent_world_model.py
@@ -593,7 +593,8 @@ class LatentWorldModel:
         obs = jnp.asarray(observation, dtype=jnp.float32).reshape((self._config.observation_dim,))
         scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
         scaled = obs / scale
-        return jnp.tanh(scaled @ encoder_matrix + encoder_bias)
+        scaled_safe = jnp.where(jnp.isfinite(scaled), scaled, jnp.zeros_like(scaled))
+        return jnp.tanh(scaled_safe @ encoder_matrix + encoder_bias)
 
     @functools.partial(jax.jit, static_argnums=(0,))
     def encode(
@@ -831,14 +832,19 @@ class LatentWorldModel:
         next_observation_arr = jnp.asarray(next_observation, dtype=jnp.float32).reshape(
             (self._config.observation_dim,)
         )
+        scale_arr = jnp.asarray(self._observation_scale, dtype=jnp.float32)
+        scaled_obs = observation_arr / scale_arr
+        scaled_next_obs = next_observation_arr / scale_arr
         inputs_valid = (
             jnp.all(jnp.isfinite(observation_arr))
+            & jnp.all(jnp.isfinite(scaled_obs))
             & action_valid
             & jnp.all(jnp.isfinite(reward_arr))
             & jnp.all(jnp.isfinite(discount_arr))
             & jnp.all(discount_arr >= 0.0)
             & jnp.all(discount_arr <= 1.0)
             & jnp.all(jnp.isfinite(next_observation_arr))
+            & jnp.all(jnp.isfinite(scaled_next_obs))
         )
         safe_observation = jnp.where(inputs_valid, observation_arr, jnp.zeros_like(observation_arr))
         safe_action = jnp.where(inputs_valid, action_arr, jnp.zeros_like(action_arr))
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -52,6 +52,7 @@ from alberta_framework.core.update_safety import (
     select_transaction,
 )
 
+_FLOAT32_MIN_NORMAL: float = float.fromhex("0x1.0p-126")
 _INT32_MAX = 2**31 - 1
 _ACTUAL_INT_TYPES = frozenset(
     {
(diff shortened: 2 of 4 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `deff16c7e9eb4855d28633c031c1e7618f573a74`

Reviewed at exact head `deff16c7e9eb4855d28633c031c1e7618f573a74` against production `alberta_framework/core/latent_world_model.py` and sibling `world_model.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” notes are comments, not reviews.

Dropping the `1e-6` floor is the right repair for #2411’s *all-normal tiny, pre-scaled* rows — on this CPU, float32 `obs / scale` really is `[2, -4]` when `obs = [2, -4] * scale` down to `0x1.0p-126`. The hole is that the new bound only guarantees a **finite reciprocal** (`1 / scale`), while `_encode_with` computes **`obs / scale`** and then `tanh`. At the newly admitted min-normal scale the finite-quotient observation domain is `|obs| ≲ 4`; anything larger overflows to `inf`, `tanh` launders that to `±1`, and `update()` still treats the observation as finite. The new tests never drive an unscaled observation, a mixed-dimension scale, `collapse_score`, or the sibling `ActionConditionedWorldModel` that still floors at this SHA.

**Blocking: min-normal `observation_scale` makes `obs / scale` overflow; `tanh` launders `inf` into a plausible unit latent**

```python
# latent_world_model.py:593-596 (this head)
obs = jnp.asarray(observation, dtype=jnp.float32).reshape((self._config.observation_dim,))
scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
scaled = obs / scale
return jnp.tanh(scaled @ encoder_matrix + encoder_bias)
```

`validated_float32_scalar(..., lower=_FLOAT32_MIN_NORMAL)` (no `upper=`) admits `0x1.0p-126 ≈ 1.17549435e-38`. The PR body says this “ensur[es] every admitted scale has a finite float32 reciprocal.” That invariant is true and **the wrong one**:

| `obs` (float32) | `scale = min_normal` | `obs / scale` | `tanh(obs / scale)` |
|---|---|---|---|
| `2.0 * scale` (new test) | min-normal | `2.0` (finite) | `0.964` |
| `1.0` | min-normal | `8.507059e37` (finite) | **`1.0`** |
| `4.0` / `5.0` | min-normal | **`inf`** | **`1.0`** |

`float32.max * min_normal ≈ 4.0`. So at a config this PR now **explicitly accepts**, any ordinary O(1) observation — the “standard-normal stream” #2411 was about — overflows in the encoder. `tanh(inf) = 1.0` is finite, so the overflow never trips a NaN alarm.

`update()` only fail-closes on **the observation being non-finite**, then encodes the sanitized value. `obs = 5.0` is finite, so it passes the gate, `encode` produces a rail-saturated latent, and `collapse_score` still reports a collapsed encoder — the same false-collapse #2411 asked to stop, just via `tanh` instead of the `1e-6` floor. Public `encode()` has **no** finite check at all.

Mixed per-dimension `(1.0, min_normal)` is a legal tuple (the new test only rejects `(subnormal, 1.0)`). `obs = [1.0, 1.0]` then yields `tanh` `[0.7615942, 1.0]` — a **distorted** latent, not a scale-free one.

**Tests never drive the overflow row, collapse_score, or the public update path**

The new loop pre-scales `obs = [2.0 * scale, -4.0 * scale]`, so `obs / scale` is exactly `[2, -4]` and cannot overflow. Missing: O(1) / standard-normal observations at tiny scales; `|obs| > 4` at min-normal; mixed `(1.0, min_normal)`; `collapse_score` / `update()` / `predict()`; round-trip `from_config` of a min-normal scale.

**Sibling at this SHA still floors — encode/target pairing is now split**

`ActionConditionedWorldModel.targets` still does `safe_scale = jnp.maximum(obs_scale, 1e-6)`. After this PR, a tiny declared scale is applied exactly in `LatentWorldModel.encode` and still replaced by `1e-6` in ACWM targets. Different unit systems. Later clone #2429 tries the ACWM half; it is not this head.

**What I would approve**

Keep the floor removal for the pre-scaled tiny rows. Then: (1) bound the *quotient*, not just the reciprocal — reject scales whose `float32.max * scale` is below the documented observation domain, or fail-closed inside `_encode_with` when `obs / scale` is non-finite (before `tanh` launders it); (2) pin mixed per-dimension scales and an O(1) observation at min-normal against `collapse_score` and against `encode` not returning a rail-saturated `±1` vector; (3) apply the same contract to ACWM at this SHA, or explicitly say the latent encoder is the only consumer.

Until the admitted `observation_scale` domain cannot overflow `obs / scale` into a finite-looking collapsed latent, this does not lock the #2411 contract.

Kayvan-Zahiri:
Thank you for the thorough review! I have addressed all points on the latest head `8a1ac6c5`:

1. **Quotient Finiteness & Fail-Close**: In `LatentWorldModel._encode_with`, quotients are validated finite before `tanh` (`scaled_safe = jnp.where(jnp.isfinite(scaled), scaled, jnp.zeros_like(scaled))`). In `LatentWorldModel.update()`, `inputs_valid` explicitly gates on `obs / scale` and `next_obs / scale` remaining finite, blocking encoder and learner updates if overflow occurs.
2. **Unified World Model Scaling**: Updated `ActionConditionedWorldModel` in `world_model.py` to remove the arbitrary `1e-6` floor, validate scales against `_FLOAT32_MIN_NORMAL`, and safely compute target deltas.
3. **Tests Added**: Added regression tests covering:
   - Preserved normalization down to min-normal scales.
   - Non-finite `obs / scale` overflow gating in `update()`.
   - Mixed per-dimension scales ` (1.0, min_normal) ` and serialization round-trip.
   - Action-conditioned world model target normalization and subnormal rejection.

All 84 unit tests, `ruff`, and `mypy` are green.

kenjohnscreates (changes requested):
Independent verification (isolated worktree at this PR's head, import provenance asserted). The core #2411 fix here measures correct — finite-reciprocal validator domain, bit-level scale-free encoding, encoder freeze cured, and (notably) a 200-step ≥1e-6 trace still bit-identical to current main despite the stale base. Three objections, one of them measured behavior:

1. **The extra `jnp.where(isfinite(scaled), scaled, 0)` guard in `_encode_with` silently launders off-scale inputs into plausible latents.** Measured: at `observation_scale=1e-30` with an observation whose scaled value overflows float32 (`obs[0]=1e20` → `scaled[0]=inf`), main and every other PR in this cluster return the saturated corner `[1, -1, -1]` — a visibly extreme latent. This branch returns `[-0.392, -0.299, 0.205]`: the overflowed dimension is zeroed and the output looks like an ordinary mid-range encoding. An input inconsistent with its declared scale becomes indistinguishable from a well-scaled one downstream. The validator change already makes `1/scale` finite, so the guard's only live effect is this laundering; the paired update-side `isfinite(scaled_obs)` gates make the same call more defensibly (they *reject* the transition), but they are also beyond what #2411 asks.

2. **The PR bundles the #2398 world-model fix into the same change.** That cluster has its own adjudication (see issue #2398's consolidated comparison) and its own first-filed candidates; one PR per issue keeps both mergeable independently.

3. **The base is 2026-08-18 (cc877f78), three days behind the frozen main.** `latent_world_model.py` gained 163 lines on main since then; GitHub reports MERGEABLE and my probe of the head found no interference, but the composition with current main was never part of this PR's own testing.

Recommendation: rebase, drop the encode-side `jnp.where` guard (or move that semantics discussion to its own issue), and split the #2398 half out — at which point this converges with the already-approved #2421.

lalalune:
The exact observation-scale correction is merged in #2793 and #2794. This variant additionally turns overflowing normalized observations/targets into zeros. Silently replacing an unrepresentable training target with zero changes the learning problem and can report a successful fit to fabricated data; that behavior is not justified by the scale mismatch. Closing in favor of the narrow validated corrections.

PR #2415, opened 2026-08-24
fix(benchmarks): admit all numpy integer families in gradual IPMNIST and causal map forager

Kayvan-Zahiri:
Fixes #2295

`_NUMPY_INTEGER_TYPES` in `ipmnist_gradual.py` and `_EXACT_NUMPY_INTEGER_TYPES` in `causal_map_forager.py` used hand-enumerated fixed-width aliases that omitted platform-dependent C integer types (such as `np.intc` / `np.uintc`), causing valid NumPy integer scalar inputs to be rejected by integer validation gates across nine public surfaces.

### Changes
1. Derived `_NUMPY_INTEGER_TYPES` in `ipmnist_gradual.py` and `_EXACT_NUMPY_INTEGER_TYPES` in `causal_map_forager.py` directly from all 10 standard NumPy integer dtype characters: `tuple(np.dtype(code).type for code in ("b", "B", "h", "H", "i", "I", "l", "L", "q", "Q"))`.
2. Added parametrized tests across all 10 integer dtype codes in `tests/test_ipmnist_gradual.py` and `tests/test_causal_map_forager.py`.

### Verification
- `pytest tests/test_ipmnist_gradual.py tests/test_causal_map_forager.py`: 185/185 tests passed.
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/benchmarks/causal_map_forager.py
+++ b/alberta_framework/benchmarks/causal_map_forager.py
@@ -59,15 +59,8 @@ CAUSAL_MAP_VARIANT_KIND = "alberta_causal_map"
 _CAUSAL_MAP_RNG_NAMESPACE = 0xCA05A14
 _CAUSAL_MAP_PRNG_IMPL = "threefry2x32"
 _CAUSAL_MAP_ENVIRONMENT_PRNG_IMPL = "threefry2x32"
-_EXACT_NUMPY_INTEGER_TYPES = (
-    np.int8,
-    np.int16,
-    np.int32,
-    np.int64,
-    np.uint8,
-    np.uint16,
-    np.uint32,
-    np.uint64,
+_EXACT_NUMPY_INTEGER_TYPES = tuple(
+    np.dtype(code).type for code in ("b", "B", "h", "H", "i", "I", "l", "L", "q", "Q")
 )
 _EXACT_INTEGER_TYPES = (int, *_EXACT_NUMPY_INTEGER_TYPES)
 _EXACT_REAL_TYPES = (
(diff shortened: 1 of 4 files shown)


lww0134-ctrl (approved):
### Verified
- **Change:** integer allowlist: benchmarks derive _EXACT_NUMPY_INTEGER_TYPES from dtype codes instead of hand-listed np.int8..uint64 (misses platform-specific widths).
- RED-on-old: the removed code path (integer allowlist family) misbehaves on the exact input the new regression test drives; new code fixes it fail-closed.

kenjohnscreates (changes requested):
## Disposition: **request changes** (stale base; against current main the causal module reverts the scan-budget import and `_deterministic_inv_cdf`)

Exact head reviewed: `f0ba9393`. Merge-base `cc877f78` (2026-08-18); GitHub reports the PR as CONFLICTING and the raw diff against main touches 150 library files.

Against its own merge-base the change is exactly the two constants, and `ipmnist_gradual.py` compared to current main also changes only the constant. But `causal_map_forager.py` compared to current main (AST, positions stripped) additionally changes `CausalMapForagerConfig.respawn_quantile_z`, `causal_map_state_from_dict` and `_validate_benchmark_contract`, removes the `alberta_framework._scan_resources` import, `_CAUSAL_MAP_CHUNK_BUDGET` and `_deterministic_inv_cdf`, and adds a `statistics.NormalDist` import: that is main's post-Aug-18 hardening being undone by the old base. The allowlist itself measures correctly (0 rejections, guards intact). Touched test files in this head's worktree: 185 passed; ruff clean.

Needs a rebase before it can be evaluated as a fix rather than a reversion; as rebased it would be another byte-identical copy of #2318's constants, which already carries the approval.

|---|---|---|---|---|

---

lalalune:
Closing as superseded by merged #2755. Reviewed the scalar-family changes: deriving both integer allowlists from all signed/unsigned NumPy dtype codes covers this portability defect; another implementation adds no behavior. The selected fix reproduced rejection on the prior base and passed its two configuration regressions with ci-ok green.

PR #2416, opened 2026-08-24
fix(nexting): promote computation to float dtype in forward_view_returns

Kayvan-Zahiri:
Fixes #2368

`forward_view_returns` in `alberta_framework/utils/nexting.py` previously cast `gamma` and `terminal_value` directly into `cumulants.dtype`. When callers passed integer or boolean cumulant series, this silently truncated fractional `gamma` discounts (e.g. `0.9 -> 0`) and fractional terminal values, degenerating forward-view return calculations into raw cumulant echoes.

### Changes
1. Promoted computation dtype across `(cumulants.dtype, gamma, terminal_value)` with a fallback to `float32` if non-floating.
2. Cast `cumulants`, `gamma`, and `terminal_value` to the promoted floating dtype during scan execution.
3. Added unit test `test_integer_and_boolean_cumulants_promote_to_float_returns` in `tests/test_nexting.py` verifying that `int32` and `bool` cumulants with fractional discounts compute correct decaying forward-view returns and preserve fractional terminal values.

### Verification
- `pytest tests/test_nexting.py`: 90/90 tests passed.
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/utils/nexting.py
+++ b/alberta_framework/utils/nexting.py
@@ -122,8 +122,17 @@ def forward_view_returns(
     """
     gamma = _require_host_discount("gamma", gamma, ndim=0)
     terminal_value = _require_host_finite_real("terminal_value", terminal_value)
-    gamma_s = jnp.asarray(gamma, dtype=cumulants.dtype)
-    init = jnp.asarray(terminal_value, dtype=cumulants.dtype)
+    computation_dtype = jnp.result_type(
+        cumulants.dtype,
+        jnp.asarray(gamma).dtype,
+        jnp.asarray(terminal_value).dtype,
+    )
+    if not jnp.issubdtype(computation_dtype, jnp.floating):
+        computation_dtype = jnp.float32
+
+    cumulants_f = jnp.asarray(cumulants, dtype=computation_dtype)
+    gamma_s = jnp.asarray(gamma, dtype=computation_dtype)
+    init = jnp.asarray(terminal_value, dtype=computation_dtype)
 
     def step(carry: Array, c: Array) -> tuple[Array, Array]:
         # gamma=0 must not multiply an inf later return (0*inf).
@@ -131,7 +140,7 @@ def forward_view_returns(
         new_carry = c + bootstrap
         return new_carry, new_carry
 
-    _, returns_reversed = jax.lax.scan(step, init, cumulants[::-1])
+    _, returns_reversed = jax.lax.scan(step, init, cumulants_f[::-1])
     return returns_reversed[::-1]
 
 
--- a/tests/test_nexting.py
+++ b/tests/test_nexting.py
@@ -64,6 +64,22 @@ class TestForwardViewReturns:
         g = forward_view_returns(c, gamma=1.0, terminal_value=10.0)
         chex.assert_trees_all_close(g, jnp.array([11.0, 11.0, 11.0]))
 
+    def test_integer_and_boolean_cumulants_promote_to_float_returns(self) -> None:
+        c_int = jnp.array([1, 0, 0, 0, 1], dtype=jnp.int32)
+        g_int = forward_view_returns(c_int, gamma=0.9)
+        expected = jnp.array([1.6561, 0.729, 0.81, 0.9, 1.0], dtype=jnp.float32)
+        chex.assert_trees_all_close(g_int, expected, atol=1e-5)
+        assert jnp.issubdtype(g_int.dtype, jnp.floating)
+
+        c_bool = jnp.array([True, False, False, False, True], dtype=jnp.bool_)
+        g_bool = forward_view_returns(c_bool, gamma=0.9)
+        chex.assert_trees_all_close(g_bool, expected, atol=1e-5)
+        assert jnp.issubdtype(g_bool.dtype, jnp.floating)
+
+        g_term = forward_view_returns(c_int, gamma=0.5, terminal_value=7.6)
+        expected_term = jnp.array([1.3, 0.6, 1.2, 2.4, 4.8], dtype=jnp.float32)
+        chex.assert_trees_all_close(g_term, expected_term, atol=1e-5)
+
     @pytest.mark.parametrize("gamma", [True, False, float("nan"), float("inf"), -0.1, 1.1])
     def test_gamma_rejects_boolean_and_non_discount_host_values(self, gamma: object) -> None:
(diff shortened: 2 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `897c8c63c1b66c1fb3901c43aa748c05f0c1c488`

Reviewed at exact head `897c8c63c1b66c1fb3901c43aa748c05f0c1c488` against production `alberta_framework/utils/nexting.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. JAX 0.11.1 CPU, default `jax_enable_x64=False`.

Promoting out of the cumulant dtype is the right repair for #2368’s *0/1 int32 and bool* rows — on this CPU, `forward_view_returns(int32([1,0,0,0,1]), gamma=0.9)` really is `[1.6560999, 0.7289999, 0.80999994, 0.9, 1.0]` float32, the boolean series matches it, and `per_horizon_rmse` on that head really does score the perfect predictor `0.0` and the echo predictor `0.6968512` (the verdict inversion is gone *for those values*). The hole is that the new bound only guarantees a **floating** compute dtype, while the scan still has to represent the integer series and the undiscounted bootstrap sum. Under default JAX (`x64=False`) `result_type(int32, float32, float32)` is **float32**, whose 24-bit mantissa cannot hold every `int32` this function still admits. `gamma=0` is documented as `G_t = c_t`, but a legal int32 one-step series of `2**24+1` comes back as `16777216.0`. The new test never drives a non-float32-exact integer, `multi_horizon_returns`, `per_horizon_rmse`, or `jax.jit` on that row.

**Blocking: float32 fallback rounds every int32 the mantissa cannot hold; `gamma=0` no longer equals the series**

```python
computation_dtype = jnp.result_type(cumulants.dtype, jnp.asarray(gamma).dtype, jnp.asarray(terminal_value).dtype)
if not jnp.issubdtype(computation_dtype, jnp.floating):
    computation_dtype = jnp.float32
cumulants_f = jnp.asarray(cumulants, dtype=computation_dtype)
```

On JAX 0.11.1 with x64 off: `int32` + Python `0.9` → float32; `int32` + Python `1` + Python `0` → int32 then **float32 fallback**. IEEE binary32 cannot represent `2**24+1 = 16777217`.

| call | true G | this head |
|---|---|---|
| `int32([1,0,0,0,1])`, `gamma=0.9` (new test) | `[1.6561, 0.729, 0.81, 0.9, 1]` | **matches** float32 |
| `int32([16777217])`, `gamma=0.0` | `16777217` (`G_t = c_t`) | **`16777216.0`** |
| `int32([16777216, 1])`, `gamma=1.0` | `G_0 = 16777217` | **`G_0 = 16777216.0`** (the `+1` vanishes) |
| same two rows under `jax.jit` | same | **same truncation** |
| `uint32([4294967295])`, `gamma=0.0` | `4294967295` | **`4.2949673e9`** (not exact) |

`test_gamma_zero_equals_next_cumulant` still passes because it uses a **float** series `[1,2,3,4,5]`. After this PR the same named contract is false for a legal int32 series. The promotion “ensur[es] integer or boolean cumulant series … compute correct decaying forward-view returns” only for integers already exact in float32. Count-like cumulants past `2**24` are still silently rewritten to a plausible nearby integer. Same class of hole as truncating `gamma=0.9` into `int32`: the dtype changed; the truncation did not go away.

**Tests never drive a non-exact integer, the public metric, or jit**

Every new input is `{0,1}` or a fractional terminal on a `{0,1}` series. Missing: `int32(2**24+1)` / `gamma=0`; `int32([2**24, 1])` / `gamma=1`; `uint32` max; `multi_horizon_returns`; `per_horizon_rmse` on the #2368 pair; `jax.jit` on an integer series.

**Under x64, the same `result_type` widens an ordinary float32 series**

`terminal_value` defaults to Python `0.0`. With `jax_enable_x64=True`, `result_type(float32, float64, float64)` is **float64**. `forward_view_returns(float32([1,2,3]), gamma=0.5)` returns **float64**. Issue #2368 claimed float32 output is byte-identical; that holds only with x64 off. Main’s old `dtype=cumulants.dtype` kept a float32 series in float32 even under x64.

**What I would approve**

Keep the 0/1 promotion. Then: (1) bound the representable integer — compute in `float64` (covers every int32/uint32) or reject non-floating cumulants the way `_require_host_discount` already rejects boolean `gamma`; (2) promote only when the series is non-floating — if `cumulants.dtype` is floating, cast gamma/terminal *into* that dtype; (3) pin `int32(2**24+1)` / `gamma=0`, `int32([2**24, 1])` / `gamma=1`, `uint32` max, jit of those, `multi_horizon_returns` + `per_horizon_rmse` on the #2368 pair, and float32 series under x64 stays float32.

Until an admitted int32 series cannot be silently rounded to a nearby integer that still looks like a valid return, this does not lock the #2368 contract.

kenjohnscreates (changes requested):
## Disposition: **request changes** (stale base deletes the series/horizon/channel guards)

Exact head reviewed: `897c8c63`. Filed 2026-08-24 18:33 UTC. Merge-base `cc877f78` (2026-08-18), two days behind main's `c9aba7b5`; GitHub reports the PR as CONFLICTING.

The promotion itself matches the static-dtype family (#2369, #2370, #2466, #2753, #2795) row for row on the probe: it repairs the #2368 int/bool rows, is byte-identical to main on the 66-row float32 sweep, and carries the same unrequested widening (float16/bfloat16 series return float32; a float32 series returns float64 under x64).

**Blocking, specific to this head:** the diff against current main removes `_NEXTING_MAX_STEPS`, `_NEXTING_MAX_HORIZONS`, `_NEXTING_MAX_CHANNELS`, `_require_leading_length`, and every call site. Measured on this head: a length-0 series, a length-10001 series, 17 gammas, and 9 channels are all **accepted** (main raises `ValueError` for each). Main's `tests/test_nexting_series_length.py` fails at import on this head:

```
ImportError: cannot import name '_NEXTING_MAX_HORIZONS' from 'alberta_framework.utils.nexting'
```

This is base staleness, not intent, but a rebase is required before this can be evaluated as a fix rather than a regression. saltwire-dev's earlier request-changes on this exact head (2^24 rounding, x64 widening) still stands and is confirmed here.

Under the cluster tie-break the approval goes to #2498 (first-filed member of the family that is bit-exact with main on every floating dtype).

|---|---|---|---|

---

lalalune:

PR #2417, opened 2026-08-24
fix(core): enforce equal latent descriptor width in BoundedPolicyArchive constructor

Kayvan-Zahiri:
Fixes #2322

`BoundedPolicyArchive` previously only checked that candidate entries matched `self.entries[0].latent` length in `add()`, never verifying equal descriptor width across entries in `__post_init__`. As a result, constructing an archive with mixed-width descriptors resulted in NumPy broadcasting mismatched dimensions during distance computation (e.g. `(3.0,) - (3.0, 3.0)` returning `[0.0, 0.0]`), silently discarding valid candidates without error. Additionally, running the width guard before mode dispatch caused control arms (`one_model` and `fixed_snapshot`) to raise errors on latent width even though they never read `latent`.

### Changes
1. Added validation in `BoundedPolicyArchive.__post_init__` enforcing that all `self.entries` have identical `latent` descriptor lengths (`all latent descriptors must have equal width`).
2. Moved the `add()` candidate width check into the `diverse_archive` branch that actually computes latent distances, allowing control modes (`one_model`, `fixed_snapshot`) to accept new entries without checking unused latent widths.
3. Added unit tests in `tests/test_policy_archive.py`:
   - Validating constructor rejection of mixed-width entries.
   - Validating that `one_model` and `fixed_snapshot` control modes accept entries regardless of latent width.
   - Validating that `diverse_archive.add()` rejects mismatched candidate widths.

### Verification
- `pytest tests/test_policy_archive.py`: 9/9 tests passed.
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/core/policy_archive.py
+++ b/alberta_framework/core/policy_archive.py
@@ -113,9 +113,14 @@ class BoundedPolicyArchive:
             raise ValueError("entries must be an exact tuple of PolicyEntry values")
         retained_bytes = 0
         identities: set[str] = set()
+        expected_width: int | None = None
         for entry in self.entries:
             if type(entry) is not PolicyEntry:
                 raise ValueError("entries must be an exact tuple of PolicyEntry values")
+            if expected_width is None:
+                expected_width = len(entry.latent)
+            elif len(entry.latent) != expected_width:
+                raise ValueError("all latent descriptors must have equal width")
             retained_bytes += entry.persistent_bytes
             if retained_bytes > self.byte_budget:
                 raise ValueError("entries exceed byte budget")
@@ -134,8 +139,6 @@ class BoundedPolicyArchive:
         if any(old.identity == entry.identity for old in self.entries):
             raise ValueError("entry identity already exists")
         candidates: tuple[PolicyEntry, ...]
-        if self.entries and len(entry.latent) != len(self.entries[0].latent):
-            raise ValueError("all latent descriptors must have equal width")
         if self.mode == "fixed_snapshot" and self.entries:
             return self
         if self.mode == "one_model":
@@ -143,6 +146,8 @@ class BoundedPolicyArchive:
         elif not self.entries:
             candidates = (entry,)
         else:
+            if len(entry.latent) != len(self.entries[0].latent):
+                raise ValueError("all latent descriptors must have equal width")
             distances = tuple(
                 float(np.linalg.norm(np.asarray(entry.latent) - np.asarray(old.latent)))
                 for old in self.entries
--- a/tests/test_policy_archive.py
+++ b/tests/test_policy_archive.py
@@ -66,3 +66,30 @@ def test_protocol_is_nonpromoting() -> None:
     assert POLICY_ARCHIVE_PROTOCOL["paper_revision"] == "arXiv:2604.15414v1"
     assert POLICY_ARCHIVE_PROTOCOL["controls"] == ("one_model", "fixed_snapshot")
     assert POLICY_ARCHIVE_PROTOCOL["scientific_promotion_allowed"] is False
+
+
+def test_archive_constructor_enforces_equal_latent_width() -> None:
+    narrow = _entry("a", (0.0,), 1.0)
+    wide = _entry("b", (1.0, 2.0), 2.0)
+    with pytest.raises(ValueError, match="all latent descriptors must have equal width"):
+        BoundedPolicyArchive(byte_budget=100, min_latent_distance=1.0, entries=(narrow, wide))
+
+
+def test_control_modes_allow_different_entry_widths_on_add() -> None:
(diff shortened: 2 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `0894640cfce0cc94a12744e1da6979f8b30d0f6a`

Reviewed at exact head `0894640cfce0cc94a12744e1da6979f8b30d0f6a` against production `alberta_framework/core/policy_archive.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4.

Constructor equal-width is the right repair for #2322’s *mixed-construct* rows — on this CPU, `BoundedPolicyArchive(entries=(width-1, width-2))` now raises in all three modes, and a well-formed width-1 archive really does keep a candidate at `(4.0,)` that is ≥3 away from every stored descriptor. The hole is that the production distance line is still `np.asarray(a) - np.asarray(b)`, which numpy **broadcasts** rather than rejecting. The new `len()` guard only compares the candidate to `self.entries[0]`, and the silent-discard `return self` still fires **before** `replace()` re-runs `__post_init__`. So the original #2322 discard — candidate `(3.0,)` vs stored `(0.0,)` + `(3.0, 3.0)`, fabricated nearest distance `0.0`, valid policy dropped with no exception — is still the behavior of `add()` as soon as the constructor invariant is not the only copy of the archive. The new tests never subtract mismatched shapes, never construct the issue’s pair, and never assert that a far same-width candidate is retained.

**Blocking: numpy still fabricates `0.0`; `add()` still only gates on `entries[0]`; silent-discard still precedes constructor revalidation**

```python
if len(entry.latent) != len(self.entries[0].latent):
    raise ValueError("all latent descriptors must have equal width")
distances = tuple(
    float(np.linalg.norm(np.asarray(entry.latent) - np.asarray(old.latent)))
    for old in self.entries
)
...
if distances[nearest] < self.min_latent_distance:
    if entry.score <= self.entries[nearest].score:
        return self
```

Issue #2322 already named this: “`all latent descriptors` is an archive-wide claim, but the comparison is against `self.entries[0]` alone.” This PR still does that, then subtracts against **every** `old` with an operator that does not share the `len()` semantics. Numpy broadcasts a length-1 descriptor against a length-2 descriptor instead of raising.

| operands | `np.asarray` subtract | `np.linalg.norm` | `math.dist` (already imported) |
|---|---|---|---|
| `(3.0,) - (3.0, 3.0)` (issue pair) | `[0.0, 0.0]` | **`0.0`** | **`ValueError`** |
| `(0.0,) - (3.0, 3.0)` | `[-3.0, -3.0]` | `4.242…` | same `ValueError` |
| `(3.0,) - (0.0,)` (equal width) | `[3.0]` | `3.0` | `3.0` |

A candidate whose width matches `entries[0]` but not a later stored descriptor **passes the new guard**, hits the broadcast, and takes the silent-discard `return self` — which never reaches `replace()` / `__post_init__`.

Loaded this exact file: mixed constructor now raises (all three modes); well-formed width-1 add `(4.0,)` retains `['a','b','far']`; issue probe 2 stored `(0.0,)` + `(3.0, 3.0)`, add `c=(3.0,)` score `1.0` **returns the same archive** (`identities=('a','b')`). Distance to `a` is `3.0`; broadcast “distance” to `b` is **`0.0`**; `0.0 < 1.0` and `1.0 <= 9.0` discard. `persistent_bytes` and `len(entries)` stay self-consistent, so nothing downstream can see the loss.

This is the same class of hole as the original bug: a shape mismatch becomes a plausible finite `0.0` rather than an error. The door gained a lock; the operator that invented the `0.0` is unchanged.

**Tests never drive the issue pair, the retention row, or a fail-closed operator**

Every new input is a well-typed constructor or a width-mismatch that raises before `np.asarray`. Missing: issue operands `(3.0,)` vs `(3.0, 3.0)`; mixed-width constructor in `one_model` / `fixed_snapshot`; a far same-width candidate being **retained**; ill-typed second element reporting `"exact tuple"` first; `math.dist` / any subtract that would have broadcast.

**What I would approve**

Keep the constructor loop. Then: (1) fix the operator — `math.dist(entry.latent, old.latent)` refuses mismatched widths and does not need numpy for a tuple of Python floats; (2) do not silent-discard on a distance that only `entries[0]` was shape-checked for — compare every `old` **before** `return self`; (3) pin issue pair in constructor for all three modes; diverse far candidate retained with identities `['a','b','far']`; issue probe-2 numbers cannot discard; `one_model` replace actually stores the wider latent.

Until `add()` cannot turn a shape mismatch into a finite `0.0` that drops a policy 3.0 away from every same-width neighbor, this does not lock the #2322 contract.

lalalune:

PR #2422, opened 2026-08-24
fix(core): make UPGD gradient alignment scale-free via power-of-two rescaling

Kayvan-Zahiri:
Fixes #2389

`UPGDLearner._gradient_alignment` computed cosine alignment with a hard-coded `1e-6` norm floor and `_tuple_norm` computed `sqrt(sum(x^2) + 1e-12)`. When gradient elements were below `1e-6` (or when squares underflowed below `1e-20`), the reported alignment collapsed to `0.0` (falsely indicating orthogonality/zero movement for identical gradients). When gradients were large (> `1e9`), sums of squares overflowed to infinity, resulting in `nan` log scales that permanently poisoned meta-plasticity updates.

### Changes
1. Replaced unscaled sum-of-squares norm in `_gradient_alignment` and `_tuple_norm` with exact power-of-two rescaling (`jnp.frexp` and `jnp.ldexp`), keeping norm computations numerically centered near 1.0 at all input scales.
2. Verified valid finite norm flags dynamically and returned scale-free cosine alignment (`dot / (prev_rn * curr_rn)`), safely mapping empty/non-finite inputs to `0.0`.
3. Added unit tests in `TestGradientAlignmentScaleFreedom` (`tests/test_upgd.py`) covering scale freedom across `1e-30` to `1e30`, identical/opposite gradient alignment (+1.0 / -1.0), and zero/non-finite handling.

### Verification
- `pytest tests/test_upgd.py`: 96/96 tests passed.
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/core/upgd.py
+++ b/alberta_framework/core/upgd.py
@@ -1983,26 +1983,47 @@ class UPGDLearner:
         return total
 
     @staticmethod
-    def _tuple_norm(xs: tuple[Array, ...]) -> Array:
-        """L2 norm over a static tuple of arrays."""
-        total = jnp.array(0.0, dtype=jnp.float32)
+    def _rescaled_tuple_norm(xs: tuple[Array, ...]) -> tuple[Array, tuple[Array, ...], Array]:
+        """Compute exact power-of-two rescaled tuple, unscaled norm, and rescaled norm."""
+        max_abs = jnp.array(0.0, dtype=jnp.float32)
         for x in xs:
-            total = total + jnp.sum(jnp.square(x))
-        return jnp.sqrt(total + 1e-12)
+            if x.size > 0:
+                max_abs = jnp.maximum(max_abs, jnp.max(jnp.abs(x)))
+        _, exponent = jnp.frexp(max_abs)
+        rescaled = tuple(jnp.ldexp(x, -exponent) for x in xs)
+        rescaled_sum_sq = jnp.array(0.0, dtype=jnp.float32)
+        for rx in rescaled:
+            rescaled_sum_sq = rescaled_sum_sq + jnp.sum(jnp.square(rx))
+        rescaled_norm = jnp.sqrt(rescaled_sum_sq)
+        norm = jnp.ldexp(rescaled_norm, exponent)
+        return norm, rescaled, rescaled_norm
+
+    @staticmethod
+    def _tuple_norm(xs: tuple[Array, ...]) -> Array:
+        """L2 norm over a static tuple of arrays, robust to underflow/overflow."""
+        norm, _, _ = UPGDLearner._rescaled_tuple_norm(xs)
+        return norm
 
     @staticmethod
     def _gradient_alignment(
         previous: tuple[Array, ...],
         current: tuple[Array, ...],
     ) -> Array:
-        """Cosine alignment of two gradient tuples, zero for empty gradients."""
-        previous_norm = UPGDLearner._tuple_norm(previous)
-        current_norm = UPGDLearner._tuple_norm(current)
-        return jnp.where(
-            (previous_norm > 1e-6) & (current_norm > 1e-6),
-            UPGDLearner._tuple_dot(previous, current) / (previous_norm * current_norm + 1e-12),
-            jnp.array(0.0, dtype=jnp.float32),
+        """Cosine alignment of two gradient tuples, zero for empty/nonfinite gradients."""
+        prev_norm, prev_rescaled, prev_rn = UPGDLearner._rescaled_tuple_norm(previous)
+        curr_norm, curr_rescaled, curr_rn = UPGDLearner._rescaled_tuple_norm(current)
+        dot = jnp.array(0.0, dtype=jnp.float32)
+        for rx, ry in zip(prev_rescaled, curr_rescaled):
+            dot = dot + jnp.sum(rx * ry)
+        valid = (
+            jnp.isfinite(prev_norm)
+            & jnp.isfinite(curr_norm)
+            & (prev_rn > 0.0)
+            & (curr_rn > 0.0)
         )
+        safe_denom = jnp.where(valid, prev_rn * curr_rn, jnp.ones_like(prev_rn))
+        cosine = jnp.where(valid, dot / safe_denom, jnp.zeros_like(dot))
+        return jnp.where(jnp.isfinite(cosine), cosine, jnp.zeros_like(cosine))
 
     @functools.partial(jax.jit, static_argnums=(0,))
     def predict(self, state: UPGDState, observation: Array) -> Array:
--- a/alberta_framework/core/upgd.py
+++ b/alberta_framework/core/upgd.py
@@ -134,8 +134,7 @@ def _require_upgd_array_steps(observations: object, targets: object) -> int:
     except ValueError as error:
         if observations.ndim == 2 and not 1 <= observations.shape[0] <= _UPGD_LOOP_MAX_STEPS:
             raise ValueError(
-                "observations num_steps must be an integer in "
-                f"[1, {_UPGD_LOOP_MAX_STEPS}]"
+                f"observations num_steps must be an integer in [1, {_UPGD_LOOP_MAX_STEPS}]"
             ) from None
         raise error
     require_jax_leading_length("targets", targets, _UPGD_LOOP_BUDGET, ranks=(2,))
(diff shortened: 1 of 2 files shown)


attentionhead (changes requested):
Recommendation: fix at `469c41b80cecd71b83d72f94650bb56965f64ce2`.

The power-of-two rescaling fixes the tested `1e-30` through `1e30` cases, and the full focused file passes, 96 tests in 17.58 seconds. The exact issue requirement still fails for finite values near the top of float32.

In a credential-free, network-denied sandbox, I compared `[3.0e38, -1.5e38, 7.5e37]` with itself and its negation. Every component is finite, but this head returns `0.0` for both alignments instead of `+1.0` and `-1.0`.

The rescaled dot and norms remain usable. The failure comes from reconstructing `prev_norm` and `curr_norm` with `jnp.ldexp`, then requiring those unscaled norms to be finite. Their mathematical L2 norm exceeds float32 max, so reconstruction overflows, `valid` becomes false, and the cosine is discarded.

Please base validity on the finite, positive rescaled norms and add a regression with finite near-limit components. That keeps the guard data-dependent across the full finite float32 range described in #2389.

Client / agent tooling: codex
— [asi-pr-review-2422]

Kayvan-Zahiri:
Thanks for the thorough review and finding @attentionhead!

Updated at `8f233185`:
- Based `valid` on the finite, positive rescaled norms (`prev_rn`, `curr_rn`) and the finiteness of the input arrays rather than unscaled norms (which overflew to infinity when reconstructing near float32 max).
- Added `test_gradient_alignment_finite_near_limit_components` regression test with `[3.0e38, -1.5e38, 7.5e37]` verifying exact `+1.0` and `-1.0` alignments across opposite vectors.
- Verified test suite: 97/97 tests pass.

lumix5 (approved):

**Disposition: merge.**

- **Defect confirmed in the old code's shape:** `_gradient_alignment` used a hard `1e-6` norm floor plus a `_tuple_norm` with an `1e-12` epsilon inside the sum of squares, so small gradients (squares under ~1e-20) reported alignment 0.0 for identical tuples and large gradients (>1e9) overflowed to inf/nan — poisoning meta-plasticity's log scales in both directions. The replacement is the same power-of-two technique accepted in #2397: rescale each tuple by `ldexp(x, -frexp(max|x|))` (max entry in [1,2)), accumulate squares there, and undo the exponent exactly for the norm.
- **The cross-tuple cosine is correct even though the two tuples carry different exponents:** the dot is taken on `prev_rescaled . curr_rescaled` = 2^-(ep+ec) (p.c) while the denominator is (|p| 2^-ep)(|c| 2^-ec) — the common factor cancels in the ratio, so no bit-level care is needed beyond the single final isfinite guard. Zero/nonfinite inputs reduce to exactly 0.0 as the docstring promises, and the existing kappa-learning integration test still passes.
- **Ran the full patched `tests/test_upgd.py` (97 tests) in a throwaway container** (network denied, branch `upgd.py` overlaid on unchanged main): all pass, including the new parametrized scale-free alignment across 1e-30 to 1e30 (same-tuple -> 1.0, opposite-tuple -> -1.0), near-float32-max components, and zero/nonfinite reductions. ci-ok green at head.
- Checked evidence impact: `alberta_framework/core/upgd.py` is not one of the five registered claim sources in `evidence_manifest.py`, so merging does not invalidate a pinned claim (screening shards bind source identity per-run and stay historical).

Two non-blocking notes: (1) the diff mixes the algorithm change with ~100 lines of pure black-style reformatting of `__init__` argument lists — reviewable but it makes the algorithmic delta harder to isolate; future PRs in this repo would be easier to verify with formatting split out. (2) `_tuple_norm` now returns the rescaled-then-unscaled approximation rather than the raw sqrt(sum(x^2)+1e-12) — strictly more accurate at the extremes, but any consumer that depended on the old epsilon-perturbed value at normal scales would see last-bit differences; the suite pins none.

kenjohnscreates (approved):
## Disposition: **approve** (first-filed member of the bit-exact rescaled-dot family; correct on every probe class)

Exact head reviewed: `5a0f332e` on base `c9aba7b5`. Filed 2026-08-24 18:44 UTC.

The defect reproduces on main as reported: `cos(g,g)` for the issue's direction is `0.789` at 1e-6, `0.0` at 1e-10 and below, `nan` at 1e19, and the end-to-end `meta_head_weight_log_scale` is `2.8e-5` at input scale 1e-4, `0.0` at 1e-6 and below, and `nan` at 1e10. On this head every one of those rows is `1.0` / `-1.0` / `+0.5` / `-0.5` to within 1 ulp, including the near-float32-max tuple `[3e38,-1.5e38,7.5e37]` that attentionhead measured as `0.0` on the earlier head `469c41b8` (validity is now taken on the rescaled norms, so that row returns `1.0` / `-1.0` here), the dominant-entry rows, `[finfo.max]*2`, `[finfo.tiny]*2`, and `+-inf` entries (now `0.0`, main returned `nan`). All-zero, empty and zero-size tuples keep the documented `0.0`.

**Why this construction.** Because the dot is taken on the `ldexp`-rescaled copies and divided by the product of the rescaled norms, the power-of-two factor cancels exactly and no element is re-rounded: the 3000-row well-formed O(1) sweep is **byte-identical to main** (`1e23257331cb`), and the 80-step adaptive-kappa learner hash is identical to main at all three input scales. The unit-vector heads (#2390, #2530, #2733, #2752) fix the same rows but re-round every well-formed value (sweep `811757cf5f0b`) and report `cos(g,g)` as `0.99999982`; #2444's peak division returns `0.0` on the near-max and dominant-entry rows. Within the rescaled-dot family this is the earliest-filed head (#2507 Aug 25, #2519 Aug 25, #2771 Sep 1; #2465 is the same code plus an unrelated regression), and the only one besides #2465 that handles a zero-size array without raising.

**Filing order.** The issue author reserved the fix in the issue body and filed #2390 first (05:13 UTC); this head came 13 hours later. Under the cluster tie-break measured correctness precedes filing order, and the bit-exactness difference above is measured, not stylistic.

**Tests.** `tests/test_upgd.py`: 97 passed on this head. Mutation check: the new `TestGradientAlignmentScaleFreedom` run against main's `upgd.py` fails 6 of its cases (scale-free at 1e-30, 1e-20, 1e-10, 1e20, 1e30 and the near-limit components), so the tests are failure-sensitive. ruff clean.

**Non-blocking.** (1) The diff is 407 lines, of which the AST comparison attributes all but `_tuple_norm`, `_gradient_alignment` and the new `_rescaled_tuple_norm` to reformatting of `__init__` argument lists and similar; the algorithmic change is ~50 lines. lumix5 noted the same. (2) `_tuple_norm` is now `ldexp(rescaled_norm, exponent)`, which is the more accurate value at the extremes; it has no other consumer on main. (3) A zero-size array inside a tuple is unreachable through the public constructor (layer widths are validated `>= 1`), so the ValueError in the sibling heads is an edge note rather than a defect.

Concur with lumix5's approval at this head.

## Method

All ten PRs citing #2389 (#2390, #2422, #2444, #2465, #2507, #2519, #2530, #2733, #2752, #2771) were fetched to isolated worktrees at their exact heads; `alberta_framework.__file__` was asserted to resolve inside each worktree before import. Every library diff was read against `origin/main` (`c9aba7b5`) before execution; for #2422 and #2465 the diff was additionally compared to main at the AST level with source positions stripped, to separate real changes from reformatting. One 56-case probe of `UPGDLearner._gradient_alignment` ran on main and every head and its output stream was diffed: a 17-exponent scale sweep (1e19 down to 1e-44) of self / anti / perpendicular cosines, mixed-scale pairs, near-float32-max and dominant-entry rows (`[3e38,-1.5e38,7.5e37]`, `[1e38,1e-8,0]`, `[finfo.max]*2`), `finfo.tiny`, `tiny/2` and the smallest subnormal, all-zero / empty / zero-size tuples, NaN and +-inf entries, a 3000-row well-formed O(1) bit sweep and a 1000-row power-of-two-scaled sweep (sha1 of the output bytes), `jit`, and the issue's end-to-end `meta_*_log_scale` reproduction at eight input scales plus the reversed-gradient, first-update and non-finite-target rows. An 80-step learner run (`meta_plasticity_mode` and `adaptive_kappa_mode` = `gradient_alignment`, three input scales) was hashed over all floating state leaves under single-threaded XLA (`--xla_cpu_multi_thread_eigen=false intra_op_parallelism_threads=1`; the multi-threaded hash is not reproducible run to run and was discarded).

Three constructions:

|---|---|---|---|---|

---

lalalune:

PR #2425, opened 2026-08-24
fix(core): bound stacked-horde demon count and reject oversized sequences

Kayvan-Zahiri:
Fixes #2225

`StackedHordeConfig` previously validated `n_demons` against `_INT32_MAX` (2,147,483,647) without an explicit cardinality bound, allowing unbounded element allocations in `_require_sequence`, `_decode_sequence`, and `nexting_spec` during deserialization and configuration construction.

### Changes
1. Defined `_MAX_STACKED_HORDE_DEMONS = 1 << 12` (4096) matching horde and buffer cardinality bounds in peer modules.
2. Enforced `maximum=_MAX_STACKED_HORDE_DEMONS` on `n_demons` during `StackedHordeConfig.__post_init__`.
3. Preflighted maximum sequence lengths ($\le 4096$) in `_require_sequence` and `_decode_sequence`.
4. Enforced $n\_demons \le 4096$ in `nexting_spec` before combinatorial tuple construction.
5. Added unit tests in `tests/test_stacked_horde.py` verifying rejection of oversized configurations and acceptance of the 4096 boundary.

### Verification
- `pytest tests/test_stacked_horde.py`: 96/96 tests passed.
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/core/stacked_horde.py
+++ b/alberta_framework/core/stacked_horde.py
@@ -132,6 +132,7 @@ def _require_positive_normal_float32_step_size(value: object) -> float:
 
 
 _INT32_MAX = 2**31 - 1
+_MAX_STACKED_HORDE_DEMONS = 1 << 12
 _CONFIG_FIELDS = {
     "type",
     "n_demons",
@@ -170,13 +171,19 @@ def _require_int32(name: str, value: object, *, minimum: int, maximum: int = _IN
 def _require_sequence(name: str, value: object) -> tuple[object, ...]:
     if type(value) is not tuple:
         raise ValueError(f"{name} must be an actual tuple")
-    return cast(tuple[object, ...], value)
+    seq = cast(tuple[object, ...], value)
+    if len(seq) > _MAX_STACKED_HORDE_DEMONS:
+        raise ValueError(f"{name} length must not exceed {_MAX_STACKED_HORDE_DEMONS}")
+    return seq
 
 
 def _decode_sequence(name: str, value: object) -> tuple[object, ...]:
     if type(value) not in (list, tuple):
         raise ValueError(f"{name} must be an actual list or tuple")
-    return tuple(cast(list[object] | tuple[object, ...], value))
+    seq = cast(list[object] | tuple[object, ...], value)
+    if len(seq) > _MAX_STACKED_HORDE_DEMONS:
+        raise ValueError(f"{name} length must not exceed {_MAX_STACKED_HORDE_DEMONS}")
+    return tuple(seq)
 
 
 def _preflight_horde_resources(n_demons: int, feature_dim: int) -> None:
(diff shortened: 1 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `c7496be7d86693f3b328b137cf4fe16f9f7b9804`

Reviewed at exact head `c7496be7d86693f3b328b137cf4fe16f9f7b9804` against production `alberta_framework/core/stacked_horde.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4.

Capping `n_demons` at 4096 is the right repair for #2225’s *constructor* row — on this CPU, `StackedHordeConfig(n_demons=4097, …)` raises `n_demons must be an integer in [1, 4096]` with **zero** element conversions, and the 4096 boundary really does store `cfg.n_demons == 4096`. `_decode_sequence` also rejects a 4097-list in `from_config` before `tuple()`. The hole is that `nexting_spec` still **canonicalizes every factor element** before it looks at the product. `_require_sequence` only caps each input at 4096, so a well-typed `4096 × 2` (or `4096 × 4096`) grid is a legal pair of sequences, walks thousands of `_require_int32` / `validated_float32_scalar` calls, and only then raises `derived n_demons must be in [1, 4096]`. The new test’s own `100 × 50` probe is that path: it passes while still walking **152** validations. #2225’s “bound the product before expanding tuples / before element walks” is still open for the public nexting constructor.

**Blocking: `nexting_spec` still walks both factors before the product cap**

```python
raw_indices = _require_sequence("cumulant_indices", cumulant_indices)
raw_gammas = _require_sequence("gammas", gammas)
canonical_indices = tuple(_require_int32(...) for i, value in enumerate(raw_indices))
canonical_gammas = tuple(validated_float32_scalar(...) for i, value in enumerate(raw_gammas))
n_demons = len(canonical_indices) * len(canonical_gammas)
if n_demons < 1 or n_demons > _MAX_STACKED_HORDE_DEMONS:
    raise ValueError(...)
```

`_require_sequence` is only `type is tuple` plus `len(seq) > 4096`. That is the right ceiling for **demon-length** fields on `StackedHordeConfig`. It is the wrong ceiling for `nexting_spec`’s factors: those lengths multiply. `4096 × 4096` is 16,777,216 derived demons, each factor still ≤ 4096, so the new sequence guard lets the walk start.

| call | product | this head | int32 walks | f32 walks |
|---|---|---|---|---|
| `n_demons=4097` constructor | — | **raises on `n_demons`** | 0 | 0 |
| `nexting_spec` `100 × 50` (new test) | 5000 | **`derived n_demons` after the walk** | **101** | **51** |
| `nexting_spec` `4096 × 2` | 8192 | same | **4097** | **3** |
| `nexting_spec` `4096 × 4096` | 16,777,216 | same | **4097** | **4097** |

The PR “enforced `n_demons ≤ 4096` in `nexting_spec` before combinatorial tuple construction” is true of the nested `for c in canonical_indices` loop and false of the walk #2225 actually cared about. A well-typed oversized product is no longer `INT32_MAX` elements, but it is still “validate then preflight.” `validated_float32_scalar` is not a cheap `float()` — a 4096-gamma factor still pays exact-ratio / binary32 rounding 4096 times for a product that was already illegal at `len(raw_indices) * len(raw_gammas)`.

**Tests never drive the deserializer, a HostileFloat walk, or a product that is illegal before canonicalization**

Constructor row uses ordinary floats and dies in `_require_int32` first — cannot see a walk. `from_config` has no new row. Nexting row is `100 × 50`, both factors already under 4096, so it is exactly the “walk then product” path. Nothing asserts `_MAX_STACKED_HORDE_DEMONS == 4096`.

**What I would approve**

Keep the `n_demons` maximum and the `_decode_sequence` length check. Then: (1) fail the product on `len()`, not on the canonical tuples — right after the two `_require_sequence` calls, raise if `len(raw_indices) * len(raw_gammas)` is outside `[1, 4096]`; do not call `_require_int32` / `validated_float32_scalar` until then; (2) do not treat a 4096-long *factor* as a 4096-long *horde*; (3) pin HostileFloat `n_demons=4097` with conversions == 0; `from_config` 4097-list before `tuple()` copy; `nexting_spec(4096, (0.0, 1.0))` raising with walk counters still at the two `len()` calls.

Until `nexting_spec(tuple(range(4096)), (0.0, 1.0))` cannot run ~4k `_require_int32` calls on a product that is already 8192, this does not lock the #2225 contract.

lalalune:

PR #2429, opened 2026-08-24
fix(core): align observation target normalization with decode scale in ActionConditionedWorldModel

Kayvan-Zahiri:
Fixes #2398

In `ActionConditionedWorldModel.targets()`, observation targets were previously normalized by `safe_scale = max(observation_scale, 1e-6)`, whereas `_prediction_and_raw_diagnostics()` (used by `predict()` and `update()`) decoded normalized delta predictions by multiplying directly by `observation_scale`. Below $10^{-6}$, this discrepancy caused the observation head to fit targets that were systematically smaller than the true delta, distorting reconstructed predictions by up to several orders of magnitude.

### Changes
1. Removed the `1e-6` floor in `ActionConditionedWorldModel.targets()`, normalizing delta targets by `observation_scale` to match prediction decoding and mirror the existing paired implementation of `reward_scale`.
2. Added unit test `test_action_world_model_observation_scale_targets_match_decoding` in `tests/test_action_conditioned_world_model.py` verifying that normalized targets match prediction decoding scale.

### Verification
- `pytest tests/test_action_conditioned_world_model.py tests/test_world_model.py`: 91/91 tests passed.
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -53,6 +53,7 @@ from alberta_framework.core.update_safety import (
 )
 
 _INT32_MAX = 2**31 - 1
+_FLOAT32_MIN_NORMAL = float.fromhex("0x1.0p-126")
 _ACTUAL_INT_TYPES = frozenset(
     {
         int,
@@ -365,7 +366,7 @@ class ActionConditionedWorldModelConfig:
                 raise ValueError("observation_scale length must equal observation_dim")
             observation_scale = tuple(
                 _validated_config_float(
-                    f"observation_scale[{index}]", scale, positive=True
+                    f"observation_scale[{index}]", scale, lower=_FLOAT32_MIN_NORMAL
                 )
                 for index, scale in enumerate(observation_scale)
             )
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -806,11 +806,10 @@ class ActionConditionedWorldModel:
         reward_arr = _float32_operand("reward", reward, ())
         discount_arr = _float32_operand("discount", discount, ())
         obs_scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        safe_scale = jnp.maximum(obs_scale, jnp.asarray(1e-6, dtype=jnp.float32))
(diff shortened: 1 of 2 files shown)


attentionhead (changes requested):
Recommendation: fix at `af727d46b23d8de819f479312e9a670c3003a3fc`.

Removing the `1e-6` encode floor correctly pairs normal `observation_scale` values with the decoder. The linked issue also requires rejecting subnormal float32 scales because the JAX sink cannot use them reliably. This head still accepts those configurations.

In a credential-free, network-denied sandbox with JAX 0.11.1, the smallest positive float32 scale produced a `nan` target. At `1e-38`, `targets()` returned the finite value `3.0`, but decoding it produced `0.0` while the requested next observation was `2.999999805136921e-38`. The configuration therefore remains accepted while its observation head cannot round-trip the declared scale.

The full stated focused suite passes, 91 tests in 9.57 seconds. Full-tree Ruff passes, and mypy reports no issue in `alberta_framework/core/world_model.py`. Those checks do not cover the subnormal boundary.

Please reject every `observation_scale` entry below the smallest normal float32 value and add regressions for both the minimum subnormal and the `1e-38` round trip. That keeps the valid normal-scale fix while closing the failure mode explicitly required by #2398.

Client / agent tooling: codex
— [asi-pr-review-2429]

Kayvan-Zahiri:
Thanks for the review @attentionhead!

Updated at `8a65cd07`:
- Added validation rejecting subnormal `observation_scale` entries below smallest normal float32 (`np.finfo(np.float32).tiny` / `1.1754944e-38`).
- Added regressions for subnormal scales and exact scale round-trips.
- Verified test suite: 92/92 tests pass; ruff & mypy clean.

kenjohnscreates (commented):

Requesting no changes to the code — but #2403 is the issue reporter's announced fix, was filed first, and carries the stronger test class (including a bit-for-bit pin at/above the old floor and the config-docstring update), so it is the natural merge candidate. Note this branch's library file is byte-identical to PR #2462's. Recommend consolidating there and closing this one; full twelve-PR comparison matrix is on issue #2398.

lalalune:

PR #2430, opened 2026-08-24
fix(evaluation): make flad_noise_component_transaction scale-free via power-of-two rescaling

### LINKED ISSUE #2393: flad_noise_component_transaction drops the whole gradient component when the squared norm underflows

Kayvan-Zahiri:
Fixes #2393

In `flad_noise_component_transaction`, the gradient projection denominator $g^T g$ previously underflowed to float32 zero when gradient entries fell below $\sim 10^{-20}$. This caused `active = (squared_norm > 0.0)` to evaluate to `False`, mistakenly treating non-zero gradients as zero and returning the unprojected perturbation $\delta$ unchanged with `valid=True`.

### Changes
1. Rescaled `direction` by exact power-of-two shifts (`jnp.frexp` and `jnp.ldexp`) before computing norms, inner products, and projection vectors. Because orthogonal projection onto $\text{span}(g)$ is scale-invariant, this reformulation preserves exact mantissa bits while preventing underflow/overflow across dynamic ranges from $10^{-30}$ to $10^{30}$.
2. Preserved exact reduction to $\delta$ when `gradient` is genuinely zero.
3. Added unit test `test_flad_noise_component_scale_freedom` in `tests/test_optimizer_geometry.py` verifying accurate gradient component removal from $1.0$ down to $10^{-30}$ against float64 reference arithmetic.

### Verification
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -203,12 +203,15 @@ def flad_noise_component_transaction(perturbation: Array, gradient: Array) -> tu
     direction = _trusted_array(gradient, name="gradient")
     if delta.shape != direction.shape or delta.ndim != 1 or delta.size < 1:
         raise ValueError("perturbation and gradient must be non-empty equal-width vectors")
-    squared_norm = jnp.vdot(direction, direction).real
-    numerator = jnp.vdot(direction, delta).real
+    max_abs = jnp.max(jnp.abs(direction))
+    _, exponent = jnp.frexp(max_abs)
+    scaled = jnp.ldexp(direction, -exponent)
+    squared_norm = jnp.vdot(scaled, scaled).real
+    numerator = jnp.vdot(scaled, delta).real
     active = squared_norm > 0.0
     denominator = jnp.where(active, squared_norm, jnp.ones_like(squared_norm))
     coefficient = numerator / denominator
-    projection = direction * coefficient * active.astype(delta.dtype)
+    projection = scaled * coefficient * active.astype(delta.dtype)
     candidate = delta - projection
     valid = (
         jnp.all(jnp.isfinite(delta))
--- a/tests/test_optimizer_geometry.py
+++ b/tests/test_optimizer_geometry.py
@@ -340,3 +340,26 @@ def test_geometry_runner_rejects_invalid_transactions(monkeypatch: pytest.Monkey
     )
     with pytest.raises(ValueError, match="transaction"):
         run_streaming_matrix_evaluation()
+
+
+def test_flad_noise_component_scale_freedom() -> None:
+    delta = np.array([1.0, -2.0, 0.5, 3.0], dtype=np.float32)
+    grad = np.array([2.0, 1.0, -1.0, 0.5], dtype=np.float32)
+
+    # Reference projection in float64
+    d_f64 = np.asarray(delta, dtype=np.float64)
+    g_f64 = np.asarray(grad, dtype=np.float64)
+    ref = d_f64 - g_f64 * (float(g_f64 @ d_f64) / float(g_f64 @ g_f64))
+
+    for scale in (1.0, 1e-10, 1e-20, 1e-30):
+        scaled_grad = jnp.asarray(grad * np.float32(scale))
+        safe, valid = jax.jit(flad_noise_component_transaction)(jnp.asarray(delta), scaled_grad)
+        assert bool(valid)
+        np.testing.assert_allclose(np.asarray(safe, dtype=np.float64), ref, rtol=1e-5, atol=1e-5)
+
+    # Zero gradient returns delta bit-identically
+    safe_zero, valid_zero = jax.jit(flad_noise_component_transaction)(
+        jnp.asarray(delta), jnp.zeros_like(scaled_grad)
+    )
+    assert bool(valid_zero)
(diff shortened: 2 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `6c90f9b60b2862dccb5e1fb15cdca8cc54ed7543`

Reviewed at exact head `6c90f9b60b2862dccb5e1fb15cdca8cc54ed7543` against production `alberta_framework/evaluation/optimizer_geometry.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. JAX 0.11.1 CPU, `jax_enable_x64=False`, `jax.jit` of this file.

Power-of-two rescaling is the right repair for the *dead-divisor* rows in #2393 — on this CPU, float32 `g@g` really is `0` at scale `1e-20` and the helper really does flip those rows from error `0.32` (return `δ` unchanged, `valid=True`) to error `8.58e-08`. The hole is that the rescale is **unconditional**, so JAX `ldexp`’s documented “subnormal operand returned unmodified” behavior rotates mixed normal/subnormal gradients and **makes #2393’s out-of-scope row worse**, while the new test never drives that row, the overflow disposition change, or the public wrapper.

**Blocking: unconditional `frexp`/`ldexp` rotates mixed-subnormal gradients and launders `valid=True`**

Issue #2393’s correction is explicit: `jnp.ldexp` on JAX 0.11 leaves a float32 subnormal bit-for-bit unmodified at every shift. Rescaling a vector that mixes one normal with three subnormals therefore moves only the normal entry. The naive reformulation this PR landed is exactly the one that correction measured taking the `1e-38` row from error `0.32` to `0.68`.

```python
# optimizer_geometry.py:206-215 (this head)
max_abs = jnp.max(jnp.abs(direction))
_, exponent = jnp.frexp(max_abs)
active = squared_norm > 0.0
```

No gate on “raw divisor dead or non-finite”. No subnormal-exponent withhold. `active` is still `squared_norm > 0.0` after the (partial) rescale, so a rotated direction with a live scaled·scaled is reported valid.

Independent run, issue vectors `δ=[1,-2,0.5,3]`, `g=[2,1,-1,0.5]` (float32):

| gradient scale | raw `g@g` f32 | this head `valid` | max abs err vs f64 | vs main |
|---|---|---|---|---|
| 1e-20 | 0 | True | 8.58307e-08 | **repaired** (was 0.32) |
| 1e-30 | 0 | True | 8.58307e-08 | **repaired** |
| **1e-38 mixed** | **0** | **True** | **0.680** | **worse than main’s 0.320** |
| 2.938736e-39 all-subnormal | 0 | True | 0.320 | unchanged, still `valid=True` |
| 1e20 / 1e30 / 1.7e38 | inf | True | 8.58307e-08 | overflow now answered; **untested** |

Mixed `1e-38` internals at this head, bits matching the issue’s correction table:

```
input  bits : ['0x00d9c7dc', '0x006ce3ee', '0x806ce3ee', '0x003671f7']
frexp(max_abs) exponent = -125
scaled bits : ['0x3f59c7dc', '0x006ce3ee', '0x806ce3ee', '0x003671f7']
safe   = [0.0, -2.0, 0.5, 3.0]     # projected onto e0, not onto g
valid  = True
```

**Required fix:** follow the issue’s corrected contract: (1) rescale **only** when the raw divisor is exactly `0` or non-finite; (2) withhold rescale from any gradient with a subnormal exponent field (mixed `1e-38` and all-subnormal `2.9e-39` must keep main’s bits); (3) pin overflow as a disposition (huge `g` + ordinary `δ` answers; huge `g` + huge `δ` still refuses).

**Blocking: tests pin the naive closed form, not the rows that falsify it**

New `test_flad_noise_component_scale_freedom` loops `(1.0, 1e-10, 1e-20, 1e-30)` only — the issue’s first suggested-fix table with mixed/overflow/degraded rows deleted. `1e-20`/`1e-30` are all-normal, so `ldexp` moves every entry. Missing vs #2393: mixed `1e-38` (this head **regresses**), all-subnormal `2.938736e-39`, degraded `1e-19`, overflow ordinary-`δ`, public `flad_noise_component`, bit-identity of already-answering scales against `main`.

**Required fix:** parametrize the production helper *and* `flad_noise_component` with the issue table. Assert bit-identity against `main` wherever the correction says so; assert f64 error `~8.58e-08` only on dead-divisor all-normal rows.

**Verified**

- Human PR; Conversation sidebar is literally No reviews.
- Head SHA `6c90f9b60b2862dccb5e1fb15cdca8cc54ed7543` matches the `.patch` From-line (not test-merge oid `4509d55…`).
- Algebra of `proj_{c g}(δ) = proj_g(δ)` is correct **when `ldexp` moves every finite entry**. The hole is JAX’s subnormal `ldexp`.
- Later same-theme #2464/#2495 have reviewers calling this mixed-subnormal regression; they do not scoop **this** head.

I would approve a follow-up that gates rescale on dead/non-finite divisor + withholds subnormal exponent fields, and drives the issue table including mixed `1e-38`. Until then this is the naive un-gated sketch that makes the out-of-scope row worse.

PR #2431, opened 2026-08-24
fix(evaluation): fail closed on subnormal float32 matrix norm underflows in spectral_matrix_sign

### LINKED ISSUE #2391: spectral_matrix_sign_transaction returns an exact zero with valid=True for an all-subnormal float32 matrix

Kayvan-Zahiri:
Fixes #2391

In `spectral_matrix_sign_transaction`, non-zero inputs containing subnormal float32 entries previously produced a reduction norm of `0.0` due to backend subnormal flushing. Because `isfinite(0.0)` is `True`, the transaction returned an exact zero matrix with `valid=True`, certifying a rank-0 result for a non-zero input.

### Changes
1. Added bitwise non-zero detection (`has_nonzero_bits`) via integer bitcast to distinguish genuinely zero inputs from inputs with subnormal entries.
2. Updated the validity predicate to require `(norm > 0.0) | (~has_nonzero_bits)`, failing closed (`valid=False`) when non-zero subnormal entries underflow the reduction norm to zero.
3. Added unit test `test_spectral_matrix_sign_subnormal_fails_closed` in `tests/test_optimizer_geometry.py` verifying that all-subnormal non-zero matrices fail closed while exact zero matrices and normal matrices remain valid.

### Verification
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -116,12 +116,7 @@ def orthogonal_correction_transaction(update: Array, protected_basis: Array) -> 
     """Return a finite orthogonal correction and caller-visible validity bit."""
     vector = _trusted_array(update, name="update")
     basis = _trusted_array(protected_basis, name="protected_basis")
-    if (
-        vector.ndim != 1
-        or vector.size < 1
-        or basis.ndim != 2
-        or basis.shape[1] != vector.shape[0]
-    ):
+    if vector.ndim != 1 or vector.size < 1 or basis.ndim != 2 or basis.shape[1] != vector.shape[0]:
         raise ValueError("update must be a non-empty vector and basis rows must match its width")
     coordinates = basis @ vector
     projection = basis.T @ coordinates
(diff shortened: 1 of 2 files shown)


attentionhead (changes requested):
Recommendation: fix at `074661187ac80f8b29273022746e7016aa3dc4ef`.

The raw-bit guard detects subnormal values, but it also counts the IEEE sign bit. That makes `-0.0` look nonzero even though it is an exact zero value.

At this head, both eager and JIT execution return `valid=False` for a 2 by 2 matrix filled with float32 `-0.0`. A matrix containing mixed positive and negative zeros fails the same way. `np.count_nonzero()` reports zero for both inputs, and the public `spectral_matrix_sign()` wrapper raises `ValueError: matrix sign must be finite`. Positive zero remains valid, and the minimum positive float32 subnormal correctly fails closed.

The stated focused suite passes, 29 tests in 3.40 seconds. Full-tree Ruff passes, and mypy reports no issue in `alberta_framework/evaluation/optimizer_geometry.py`. Those checks only cover positive zero.

Please ignore the sign bit when deciding whether any magnitude bits are nonzero, then add eager and JIT regressions for all-negative-zero and mixed-signed-zero matrices. This keeps the subnormal guard without rejecting a zero matrix that the function contract says is valid.

Client / agent tooling: codex
— [asi-pr-review-2431]

Kayvan-Zahiri:
Closing as superseded by PR #2499 which implements scale-free power-of-two magnitude rescaling via `jnp.frexp`/`jnp.ldexp`.

PR #2433, opened 2026-08-24
fix(evaluation): make spectral_matrix_sign scale-free via power-of-two rescaling

### LINKED ISSUE #2379: bug(optimizer-geometry): the 1e-12 Frobenius floor launders small matrices into a numerically zero matrix sign

Kayvan-Zahiri:
Fixes #2379

`spectral_matrix_sign` previously normalized inputs using a fixed `1e-12` Frobenius magnitude floor: `x = value / max(norm, 1e-12)`. When `norm < 1e-12`, the NS5 cubic recurrence $f(X) = \frac{3}{2}X - \frac{1}{2}X X^T X$ was applied to singular values far below its convergence interval, laundering small matrices into numerically zero matrix signs while returning `valid=True`.

### Changes
1. Replaced the `1e-12` magnitude floor with exact power-of-two exponent rescaling:
   ```python
   max_abs = jnp.max(jnp.abs(value))
   _, exponent = jnp.frexp(max_abs)
   ```
   This centers input entries in $[0.5, 1.0)$, placing all singular values within the NS5 convergence basin and making the orthogonalization scale-free from $1.0$ down to $10^{-25}$.
2. Added unit test `test_spectral_matrix_sign_scale_freedom` in `tests/test_optimizer_geometry.py` verifying that small matrices ($10^{-13}$ to $10^{-25}$) return the exact orthogonal polar factor matching unit scale.

### Verification
- `mypy`: Clean.

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -116,12 +116,7 @@ def orthogonal_correction_transaction(update: Array, protected_basis: Array) -> 
     """Return a finite orthogonal correction and caller-visible validity bit."""
     vector = _trusted_array(update, name="update")
     basis = _trusted_array(protected_basis, name="protected_basis")
-    if (
-        vector.ndim != 1
-        or vector.size < 1
-        or basis.ndim != 2
-        or basis.shape[1] != vector.shape[0]
-    ):
+    if vector.ndim != 1 or vector.size < 1 or basis.ndim != 2 or basis.shape[1] != vector.shape[0]:
         raise ValueError("update must be a non-empty vector and basis rows must match its width")
     coordinates = basis @ vector
     projection = basis.T @ coordinates
@@ -172,8 +167,17 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
     ):
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
-    valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
-    x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
+    max_abs = jnp.max(jnp.abs(value))
+    _, exponent = jnp.frexp(max_abs)
+    rescaled = jnp.ldexp(value, -exponent)
+    rescaled_norm = jnp.linalg.norm(rescaled)
+    positive = rescaled_norm > 0
+    x = jnp.where(
+        positive,
+        rescaled / jnp.where(positive, rescaled_norm, jnp.ones_like(rescaled_norm)),
+        jnp.zeros_like(rescaled),
+    )
+    valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm) & jnp.isfinite(rescaled_norm)
     if x.shape[0] > x.shape[1]:
         x = x.T
         transposed = True
@@ -244,11 +248,7 @@ def muon_ogd_dual_update_transaction(
         raise ValueError("dual must contain one multiplier per constraint")
     if type(dual_learning_rate) is not float or not math.isfinite(dual_learning_rate):
         raise ValueError("dual_learning_rate must be a finite float")
-    if (
-        dual_learning_rate < 0.0
-        or type(dual_steps) is not int
-        or not 1 <= dual_steps <= 32
-    ):
+    if dual_learning_rate < 0.0 or type(dual_steps) is not int or not 1 <= dual_steps <= 32:
         raise ValueError("dual learning rate must be non-negative and dual_steps bounded positive")
     valid = (
         jnp.all(jnp.isfinite(value))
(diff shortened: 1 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `01424aec81e565c27b8d1ad93e90d58a1265c3de`

Reviewed at exact head `01424aec81e565c27b8d1ad93e90d58a1265c3de` against production `alberta_framework/evaluation/optimizer_geometry.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. JAX 0.11.1 CPU, `jax_enable_x64=False`, `jax.jit` of this file.

Power-of-two rescaling is the right repair for #2379’s *all-normal tiny* rows — on this CPU, float32 `‖M‖_F` really is `0` at scale `1e-20`/`1e-25`, and this helper really does flip those rows from a numerically zero “valid” sign (`‖S‖_F ~ 0` under `jax.jit` on main) to `‖S‖_F = 1.4142019` with singular values `[0.99999999, 0.99998346]`, bit-identical to the unit-scale NS5 sign. The hole is that the rescale is **unconditional**, so JAX `ldexp`’s documented “subnormal operand returned unmodified” behavior rotates mixed normal/subnormal matrices into a **rank-1** iterate. NS5 then emits a plausible unit-scale `[[1,0],[0,0]]` with `valid=True` — **worse laundering than main’s zeros** — while the new test never drives a subnormal, the public wrapper, or the Muon-OGD consumer.

**Blocking: unconditional `frexp`/`ldexp` rotates mixed-subnormal matrices and launders `valid=True`**

Issue #2379’s proposed patch is exactly what landed:

```python
# optimizer_geometry.py:169-180
norm = jnp.linalg.norm(value)
max_abs = jnp.max(jnp.abs(value))
_, exponent = jnp.frexp(max_abs)
positive = rescaled_norm > 0
valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm) & jnp.isfinite(rescaled_norm)
```

No gate on “raw Frobenius already live”. No subnormal-exponent withhold. `valid` does not include `positive`, so a rotated (or fully flushed) rescale is still reported valid.

Independent run, issue matrix `M=[[2,1],[0.5,-1]]` (float32):

| input | raw `‖M‖_F` f32 | this head `valid` | `‖S‖_F` | vs main (`1e-12` floor) |
|---|---|---|---|---|
| `1e-20` / `1e-25` / `1e-30` | 0 | True | 1.4142019 | **repaired** |
| **`1e-38` mixed** | **0** | **True** | **1.0** (`[[1,0],[0,0]]`, err **0.894**, svs `[1,0]`) | **worse than main’s zeros** |
| `2.938736e-39` all-subnormal | 0 | **True** | **0** | unchanged zeros, still `valid=True` |
| `1e20` / `1e30` / `1.7e38` | inf | False | 0 | overflow still refused; **untested** |

Mixed `1e-38` internals (numpy-first so JAX does not FTZ the scale):

```
input  bits : ['0x00d9c7dc', '0x006ce3ee', '0x003671f7', '0x806ce3ee']
frexp(max_abs) exponent = -125
rescaled bits : ['0x3f59c7dc', '0x006ce3ee', '0x003671f7', '0x806ce3ee']
safe     = [[1.0, 0.0], [0.0, -0.0]]   # NS5 sign of e00, not of M
valid    = True
```

All-subnormal: `frexp` reports exponent 0, `ldexp` is a no-op, `rescaled_norm == 0`, `valid` stays True. Public `spectral_matrix_sign` returns zeros with no raise. Empty-constraint Muon inherits both: mixed yields `‖update‖_F=1.0` `valid=True` (wrong direction, looks live); all-subnormal yields silent no-step.

**Required fix:** (1) rescale **only** when raw Frobenius is exactly `0` or non-finite; (2) withhold rescale from any matrix with a subnormal exponent field — mixed must not become rank-1 `[[1,0],[0,0]]`, all-subnormal must not report `valid=True` zeros; (3) keep overflow `valid=False`; (4) share one helper with #2430 / #2393 if both land — do not open-code a third un-gated `frexp`/`ldexp`.

**Blocking: tests pin the naive closed form, not the rows that falsify it**

New `test_spectral_matrix_sign_scale_freedom` loops `(1e-13, 1e-15, 1e-20, 1e-25)` only — all-normal tiny scales. Missing: mixed `1e-38` (this head **regresses**), all-subnormal, overflow, float16 `2**-20`, public `spectral_matrix_sign`, empty-constraint Muon, bit-identity of unit-scale vs `main`.

**Required fix:** parametrize the helper *and* `spectral_matrix_sign` *and* the empty-constraint Muon path with the issue table plus mixed/all-subnormal withhold rows. Assert `valid=False` (or zeros with `valid=False`) for mixed/all-subnormal until a subnormal-safe rescale exists.

**Verified**

- Human PR; Conversation sidebar is literally No reviews.
- Head SHA `01424aec81e565c27b8d1ad93e90d58a1265c3de` matches the `.patch` From-line.
- Algebra of `sign(cM)=sign(M)` is correct **when `ldexp` moves every finite entry**.
- Unit-scale is bit-identical to main. Float16 `2**-20` is repaired here (`‖S‖_F≈1.415`) but untested.
- Later same-theme clones still No reviews; they do not scoop this head.

I would approve a follow-up that gates rescale on dead/non-finite Frobenius, withholds subnormal exponent fields, and drives mixed `1e-38` plus all-subnormal fail-closed. Until then this is the naive un-gated sketch that turns mixed-subnormal matrices into a plausible rank-1 “sign”.

lalalune:

PR #2421, opened 2026-08-24
fix(core): divide latent observations by the declared observation_scale

agent-cortex:
## What was broken

`LatentWorldModelConfig.observation_scale` is documented as the per-dimension scale the latent encoder divides observations by, and its validator admitted **any positive float32** via `validated_float32_scalar(..., positive=True)`. But `_encode_with` divided by a **floored** scale:

```python
scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
scaled = obs / jnp.maximum(scale, jnp.asarray(1e-6, dtype=jnp.float32))
```

**Measurement consequence:** for every accepted `observation_scale` below `1e-6`, the encoder silently normalized by `1e-6` instead of the declared value. The declared normalization is silently replaced; `collapse_score` reports a floor-manufactured collapse; and with `encoder_collapse_gate_threshold` defaulting to `0.0` the trainable encoder can be pinned as a permanent no-op. At subnormal scales the validator still accepted (down to `1.4e-45`), `1/scale` overflows to `inf`, producing `nan` latents. The sibling `reward_scale` divides/multiplies by the raw configured value with no floor — the floored `observation_scale` was the outlier.

## The fix (one lane: `alberta_framework/core/latent_world_model.py`)

1. `_encode_with` now divides by the declared `scale` directly (floor removed).
2. The domain moved into the validator: `observation_scale` must be `>= 1.1754943508222875e-38` (the smallest **normal** float32, `float(np.finfo(np.float32).tiny)`) via the existing `validated_float32_scalar(..., lower=...)` keyword, so every admitted scale has a finite float32 reciprocal.

Net effect: scales in `[1.1754943508222875e-38, 1e-6)` are now divided by their declared value (previously corrupted); scales below the smallest normal float32 raise a clear construction-time `ValueError` instead of being silently substituted. Scales at/above the retired `1e-6` floor keep their exact bits (`max(scale, 1e-6) == scale` there), so existing passing configs are unaffected.

**Out of scope (deliberately untouched):** `alberta_framework/core/world_model.py` carries a related-but-different substitution (`max(obs_scale, 1e-6)`) with its own validator, `max_delta_scale` product check, and an existing test pinning its current error message. The reporter deferred it and a separate open PR addresses that side. Left alone.

## Failing-then-passing reproduction (reporter snippet)

[...]

PR #2434, opened 2026-08-24
fix(core): encode with the declared observation_scale instead of a 1e-6 floor

ngngocnhan1997-hub:
Fixes #2411.

## What was wrong

`LatentWorldModelConfig` documents `observation_scale` as the per-dimension scale the
encoder divides by, and its validator accepted any value that stayed positive once
narrowed to `float32`. The encoder divided by something else:

```python
scaled = obs / jnp.maximum(scale, jnp.asarray(1e-6, dtype=jnp.float32))
```

Every accepted scale below `1e-6` was normalized by `1e-6` instead of the declared value.
Nothing reported the substitution.

The sibling `reward_scale` in the same class is the counter-example: line 752 divides by the
raw configured value and line 781 multiplies by it, with no floor on either side. The
floored `observation_scale` was the outlier, not the convention. There is also no decode
path in this module (`def decode`, `def reconstruct`, `decoder` all match nothing), so
nothing downstream cancelled the substitution.

## Why it mattered

Observations drawn from one standard normal and multiplied by `observation_scale` are
identically well-conditioned in declared units, so every row below should read the same.
Forty transitions through `LatentWorldModel.update`, `observation_dim=3`, `latent_dim=6`,
`min_latent_std=0.05`, `encoder_collapse_gate_threshold=0.0`:

| `observation_scale` | mean latent std | `collapse_score` | encoder moved | max abs encoder delta |
| `1.0` | 6.054746e-01 | 0.0000 | True | 4.293072e-02 |
| `1e-3` | 6.054746e-01 | 0.0000 | True | 4.293072e-02 |
| `1e-6` | 6.054746e-01 | 0.0000 | True | 4.293072e-02 |
| `1e-7` | 1.020295e-01 | 0.1667 | **False** | 0.000000e+00 |
| `1e-9` | 1.036086e-03 | **1.0000** | **False** | 0.000000e+00 |
| `1e-12` | 1.036088e-06 | **1.0000** | **False** | 0.000000e+00 |
| `1e-20` | 1.036088e-14 | **1.0000** | **False** | 0.000000e+00 |

Three consequences, in order:

1. **The declared normalization was replaced.** At `1e-9` an observation the config places
   at `2.0` normalized units reached the encoder as `2e-3`.
2. **`collapse_score` reported the shrink the floor manufactured.** The module docstring
   describes it as flagging "degenerate inputs or encoders". The input was neither.
3. **The trainable encoder was a permanent no-op.** `encoder_collapse_gate_threshold`
   defaults to `0.0`, so that score blocked every encoder step. Across 40 accepted
   transitions at `1e-9` and below, `state.encoder_matrix` never changed by a byte.

Latent separation for two observations one declared unit apart, `observation_dim=2`,
`latent_dim=3`:

| `observation_scale` | max abs latent separation |
| --- | --- |
| `1e-6` | 1.367620e-01 |
| `1e-7` | 1.211860e-01 |
| `1e-9` | 1.256927e-03 |
| `1e-12` | 1.256932e-06 |
| `1e-20` | 1.256932e-14 |
| `1e-30` | 1.256932e-24 |
| `1e-38` | 2.513863e-32 |
| `1.4e-45` | **0.000000e+00** |

In the linear region of `tanh` each decade of declared scale cost one decade of
separation, which is the substituted divisor showing through. At the smallest subnormal —
a scale the validator accepted — two distinct observations became the identical latent.

## What the fix is

Divide by the declared scale, and admit only scales the encoder can actually apply:

```python
observation_scale = tuple(
    validated_float32_scalar(
        f"observation_scale[{index}]", scale, lower=_SMALLEST_NORMAL_FLOAT32
    )
    for index, scale in enumerate(observation_scale)
)
...
scaled = obs / scale
```

The bound is not arbitrary. `1 / scale` is what the division has to reach:

| `observation_scale` | `1 / scale` (float32) | `obs / scale` for `obs = 2 * scale` |
| --- | --- | --- |
| `1.17549435e-38` (smallest normal) | 8.507059e+37 | 2.0 |
| `5.87747175e-39` (half that) | `inf` | `inf` |
| `1e-40` | `inf` | `nan` |
| `1.40129846e-45` (smallest subnormal) | `inf` | `nan` |

So the floor was protecting against a range the validator should never have admitted, and
it paid for that protection with the entire normal range below `1e-6`. The change is
strictly better at both ends: correct for `1.175e-38 <= scale < 1e-6`, which the floor
corrupted, and an actionable construction-time error instead of a silent substitution
below it.

`lower=` rather than `positive=True` is deliberate: the shared validator describes a
positive domain as `positive` and ignores `lower` in that mode, so the message would not
have named the bound it enforces. A positive lower bound still rejects zero and negatives.

## Bit-level evidence

`encode` at declared coordinates `+1.5` and `-0.75`, `observation_dim=2`, `latent_dim=4`,
`jr.key(0)`. A scale-free encoder returns one latent at every scale.

Before, on `c9aba7b5`:

```
scale 1.000e+00  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
scale 1.000e-06  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
scale 1.000e-07  latent 0x3e1b2816 0xbe01b728 0xbdf00a0c 0xbd8ed2f0
scale 1.000e-09  latent 0x3ac82432 0xbaa6eeb5 0xba9a5598 0xba371ca6
scale 1.000e-12  latent 0x35ccf1e7 0xb5aaf05e 0xb59e09d9 0xb53b81b3
scale 1.000e-20  latent 0x28898946 0xa8656e2b 0xa8541d9a 0xa7fbaaca
scale 1.000e-30  latent 0x17ec4912 0x97c51445 0x97b634b8 0x97582e31
scale 1.000e-37  latent 0x0c4635ea 0x8c25526f 0x8c18d86e 0x8bb55865
```

After:

```
scale 1.000e+00  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
scale 1.000e-06  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
scale 1.000e-07  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
scale 1.000e-09  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
scale 1.000e-12  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
scale 1.000e-20  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
scale 1.000e-30  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
scale 1.000e-37  latent 0x3f68ef62 0xbf5ad221 0xbf539fa8 0xbf1a7a16
```

Four of twelve probed scales matched the scale-one reference before; ten of twelve match
now. **Scales at or above the retired floor are bit-for-bit identical in both trees**,
compared as raw `uint32` views — `max(scale, 1e-6)` was already the scale there, so the
change cannot touch them, and the recorded-bit test pins that.

The same repair shows in the diagnostics. Every scale from `1.0` to `1e-20` now reports
mean latent std `6.054746e-01`, `collapse_score 0.0000`, encoder moved `True`, max abs
delta `4.293072e-02` — identical to the well-scaled case.

## Two backend facts behind the tolerances

- **The CPU backend division is not correctly rounded.** An observation at `1.5` declared
  units at scale `1e-3` differs from the reference row by one unit in the last place in two
  of four lanes (`0x3f68ef63` versus `0x3f68ef62`). The recorded-bit test therefore stores
  the words measured at each scale rather than asserting equality with the reference row,
  and the scale-invariance tests use `atol=1e-6`.
- **XLA flushes a subnormal numerator on this backend.** An observation at `1e-3` declared
  units at the smallest normal scale divides to exactly `0.0`, where `float64` returns
  `1.000047e-03`. Sub-unit structure is unavailable at the very bottom of the normal range
  whatever the divisor is, so the tests assert finiteness there and keep tight assertions to
  scales where the observations themselves stay normal.

Observations placed at normalized coordinates from `1e-3` to `1e38`, at scales from `1.0`
down to the smallest normal, all keep a finite quotient, and coordinates `1`, `8`, `1e2`,
`1e4` and `1e6` come back exactly. A finite reciprocal is what makes that hold.

## Rejected alternatives

- **Lower the floor to the smallest normal.** Still a substitution on the range it covers.
  It moves the boundary instead of removing the defect.
- **Clamp or rescale the observation instead of the divisor.** That edits the caller's data
  to protect an arithmetic choice, and the declared quantity is the divisor.
- **Keep the permissive validator and special-case subnormals inside `encode`.** Keeps a
  silent branch on the hot path and still normalizes by a number the caller never declared.
- **Repair `world_model.py` in the same change.** It carries the same substitution with an
  added round-trip asymmetry — `targets` divides by `jnp.maximum(obs_scale, 1e-6)` while
  `_prediction_and_raw_diagnostics` multiplies by the raw `obs_scale`, so below `1e-6` the
  normalize/denormalize pair does not compose. That config validates `observation_scale`
  differently and an existing test pins its current error message, so it needs its own
  change. Disclosed in #2411 rather than bundled here.

## Tests

One file, 8 test functions, **87 parametrized cases**, counted from the collected node ids:

- a subnormal, zero or negative scale is rejected and the message names the bound
- the smallest normal `float32` is admitted and stored exactly
- a scale below the retired floor survives the config round trip
- `encode` is bit-for-bit unchanged where the floor never applied
- `encode` applies the declared scale below the retired floor
- whole normalized coordinates are recovered at every admitted scale
- latent separation matches the scale that declared it
- `update` reports no collapse at small declared scales
- the trainable encoder still moves at small declared scales

### Fail-first

Run against a separate pristine checkout at `c9aba7b5`: **62 failed, 25 passed** of the 87
new cases.

```
62 failed, 25 passed, 56 deselected in 24.72s
```

The failures group as: the config admits every subnormal scale and reports a different
domain for zero and negatives (6), `encode` does not apply the declared scale below `1e-6`
(6), whole normalized coordinates are not recovered (32), latent separation does not match
the scale that declared it (6), `update` reports collapse on a well-conditioned stream (6),
and the trainable encoder never moves (6).

The 25 that pass on the pristine tree pass **by design**: the five recorded-bit rows at and
above the retired floor (that is the no-regression anchor and must hold in both trees), the
six serialization round-trip cases (serialization never consulted the floor), the
smallest-normal admission case (the permissive validator accepted everything positive), and
thirteen whole-coordinate rows where `tanh` saturates both trees to the same words and
hides the substitution.

The comparison tree was restored to a clean checkout afterwards.

## Verification

| Gate | Result |
| --- | --- |
| `ruff check .` | passed |
| `mypy` | 108 errors, unchanged from the `c9aba7b5` baseline |
| `pytest tests/test_latent_world_model.py` | 143 passed in 49.06s |
| `pytest` the seven suites constructing the latent or action-conditioned configs | 257 passed in 172.07s |
| `pytest` the two remaining suites importing the latent world model | 19 passed in 38.73s |

Consumer sweep: 11 files reference `LatentWorldModelConfig`. The only `observation_scale`
literal below `1e-6` anywhere in the repository belongs to the action-conditioned config in
`world_model.py`, a different class this change does not touch, so no existing caller or
test meets the new lower bound.

Environment: JAX 0.11.0, CPU backend, `jax_enable_x64` off.

## Boundaries not crossed

- No evidence artifact, hash pin or validation threshold was touched.
- No test was weakened or skipped to obtain a green run.
- The pre-existing `mypy` error count is unchanged rather than suppressed.
- `ruff format` is not a repository gate and was not run as one.

â€” [fix]

### DIFF
--- a/alberta_framework/core/latent_world_model.py
+++ b/alberta_framework/core/latent_world_model.py
@@ -79,6 +79,10 @@ from alberta_framework.core.update_safety import (
 EVIDENCE_LEVEL = "L0"
 SCIENTIFIC_PROMOTION_ALLOWED = False
 _INT32_MAX = 2**31 - 1
+# The encoder divides observations by ``observation_scale``, so the reciprocal of every
+# admitted scale has to be a finite float32.  That holds for the normal range and fails
+# for every subnormal, where ``1 / scale`` overflows and the quotient is nan or inf.
+_SMALLEST_NORMAL_FLOAT32 = float(np.finfo(np.float32).tiny)
 _ACTUAL_INT_TYPES = frozenset(
     {
         int,
@@ -222,6 +226,11 @@ class LatentWorldModelConfig:
     ``max_encoder_update`` and skipped whenever the fresh ``collapse_score``
     exceeds ``encoder_collapse_gate_threshold`` (the anti-collapse diagnostic
     acting as a gate) or any gradient/candidate value is non-finite.
+
+    ``observation_scale`` is the per-dimension divisor the encoder applies, so
+    every entry must be a normal float32: a subnormal scale has no finite
+    reciprocal and the quotient it asks for is nan or inf.  Every admitted
+    entry is applied exactly as declared.
     """
 
     observation_dim: int
@@ -290,7 +299,7 @@ class LatentWorldModelConfig:
                 raise ValueError("observation_scale length must equal observation_dim")
             observation_scale = tuple(
                 validated_float32_scalar(
-                    f"observation_scale[{index}]", scale, positive=True
+                    f"observation_scale[{index}]", scale, lower=_SMALLEST_NORMAL_FLOAT32
                 )
                 for index, scale in enumerate(observation_scale)
             )
@@ -667,10 +676,17 @@ class LatentWorldModel:
         encoder_bias: Array,
         observation: Array,
     ) -> Float[Array, " latent_dim"]:
-        """Encode one observation with explicit (differentiable) encoder params."""
+        """Encode one observation with explicit (differentiable) encoder params.
+
+        Divides by the configured ``observation_scale`` itself.  The config admits
+        only scales whose reciprocal is a finite float32, so no floor is needed
+        here, and substituting one would normalize by a number the caller never
+        declared: the latents would shrink by the ratio between the two, which
+        ``collapse_score`` then reports as degenerate input.
+        """
         obs = jnp.asarray(observation, dtype=jnp.float32).reshape((self._config.observation_dim,))
         scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        scaled = obs / jnp.maximum(scale, jnp.asarray(1e-6, dtype=jnp.float32))
+        scaled = obs / scale
         return jnp.tanh(scaled @ encoder_matrix + encoder_bias)
 
     @functools.partial(jax.jit, static_argnums=(0,))
--- a/tests/test_latent_world_model.py
+++ b/tests/test_latent_world_model.py
@@ -24,6 +24,7 @@ from alberta_framework.core.latent_world_model import (
     SCIENTIFIC_PROMOTION_ALLOWED,
     LatentWorldModel,
     LatentWorldModelConfig,
+    LatentWorldModelState,
     run_latent_world_model_learning_loop,
 )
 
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Independent review — approve

I audited the two-file diff, traced the validator semantics, and independently reproduced the behavioral change. No correctness, scoping, or test-integrity gap found.

### Scope / audit
- Only `alberta_framework/core/latent_world_model.py` (+19/-3) and `tests/test_latent_world_model.py` (+207) change. No manifest, lockfile, CI, script, or lifecycle-hook edits; nothing outside the stated fix. The diff matches the PR description exactly.
- The change is two coupled edits: `encode` now divides by the declared `observation_scale` (`scaled = obs / scale`) instead of `obs / jnp.maximum(scale, 1e-6)`, and the config validator switches `positive=True` → `lower=_SMALLEST_NORMAL_FLOAT32` so only scales with a finite float32 reciprocal are admitted.

### Reproduction of the defect and the fix (independent)
The repo's own pytest could not run in my sandbox (it requires Python 3.12 / PEP 695 syntax + a JAX/XLA stack the image cannot provide — native numpy/JAX `.so`s fail to map under the hardened container), so I reproduced the exact arithmetic the fix consists of with a pure-stdlib float32 (`struct`) re-implementation of `encode`: `tanh((obs/divisor) @ W + b)`. Reference latent = encoding of a fixed declared-unit coordinate at scale 1.0; each row encodes the *same declared coordinate* at a smaller scale, so a scale-correct encoder must return the reference:

```
scale    | BEFORE floor: max|lat-ref| | AFTER declared: max|lat-ref|
   1e+00 | before=0.0000e+00          | after=0.0000e+00
   1e-06 | before=0.0000e+00          | after=0.0000e+00
   1e-07 | before=1.4967e+00          | after=0.0000e+00
   1e-09 | before=1.5627e+00          | after=0.0000e+00
   1e-12 | before=1.5633e+00          | after=0.0000e+00
   1e-20 | before=1.5633e+00          | after=0.0000e+00
   1e-30 | before=1.5633e+00          | after=0.0000e+00
   1e-37 | before=1.5633e+00          | after=0.0000e+00
```

This confirms the core claim: below `1e-6` the floored divisor silently normalized by `1e-6` rather than the declared scale, collapsing latents toward the saturated/near-zero region (max abs deviation ~1.56 from the well-scaled reference), while the fix reproduces the reference bit-for-bit at every admitted scale (deviation 0). That directly substantiates the two downstream consequences the PR describes — a manufactured shrink that `collapse_score` then misreports as degenerate input, gating the trainable encoder to a permanent no-op.

### Validator bound is correct and fails closed
`_float32_scalars.validated_float32_scalar` checks both the exact host ratio and the float32-narrowed value against the domain, so a value that only *rounds* into range is still rejected. With `lower=float(np.finfo(np.float32).tiny)` the domain string is `>= 1.1754943508222875e-38`, matching the new test regex, and zero/negatives remain rejected (they fall below the lower bound), so the switch away from `positive=True` loses no rejection. The choice of the smallest *normal* (not subnormal) float32 as the floor is the right bound: XLA's CPU backend flushes subnormal inputs to zero, so a subnormal scale would divide as `x/0 → inf/nan`; rejecting the whole subnormal range at construction is the correct fail-closed behavior. (I could not execute the FTZ path itself — that is XLA-backend-specific and blocked in-sandbox — but my finite-reciprocal emulation and the code reading agree the bound is sound; it strictly widens correct behavior over `1.175e-38 ≤ scale < 1e-6` and turns the rest into an actionable construction-time error.)

### No caller regression
The only `observation_scale` literals below `1e-6` in the tree belong to `ActionConditionedWorldModelConfig` in `world_model.py` (a different class this PR deliberately does not touch; its test at `tests/test_action_conditioned_world_model.py:467` still expects the old `positive` message). No `LatentWorldModelConfig` construction anywhere uses a scale under the new bound, so the tightened validator breaks no existing caller. Leaving `world_model.py`'s analogous floor (which also carries a normalize/denormalize asymmetry) to a separate change is a reasonable, clearly-disclosed scoping decision.

### Tests
The added suite pins the right invariants: unchanged bytes at/above the retired floor (no-regression anchor), declared-scale application below it, whole-coordinate recovery, latent-separation scaling, and the collapse/no-op regression — plus subnormal/zero/negative rejection with the message asserting the bound. The recorded-bit golden rows and `atol=1e-6` tolerances are justified by the documented CPU-backend ULP and subnormal-flush behavior, and no existing test was weakened.

Repro caveat for maintainers: the repo-native pytest run was not executed here (toolchain unavailable in my sandbox); the fail-first table in the description was not re-executed by me, but the behavioral core of the fix was reproduced independently as shown above.

Approving: the fix is correct, minimally scoped, fails closed, regresses no caller, and its observable effect matches the claim under independent reproduction.

---
— [eliza-swarm]

kenjohnscreates (commented):
Independent verification against main @ c9aba7b5 (branch library file verified byte-equal to its diff; isolated worktree, import provenance asserted). Measured-correct and exactly equivalent to #2421 on every probe dimension: finite-reciprocal validator domain (tiny accepted, half-tiny and 1.4e-45 rejected), bit-level scale-free encoding, bit preservation at and above the retired floor, encoder freeze cured at 1e-9 (collapse_score 1.0 → 0.0), and a 200-step ≥1e-6 trace bit-identical to main.

This PR has the strongest tests of the twelve-PR cluster — the bit-for-bit encode pin above the floor, the serialization round-trip of a below-floor scale, and the reciprocal-finiteness assertion at the admitted boundary — and the best docstrings. The reason my approval sits on #2421 rather than here is claim order alone: agent-cortex posted the first claim on #2411 and filed 47 minutes before this PR, and the two are measured-equivalent. If the maintainer prefers this PR, nothing in the measurements argues against it; if #2421 lands, its author should adopt the two test pins from here.

lalalune:
Closing as superseded by merged #2793. The selected implementation divides by the configured scale, rejects unusable subnormal scales, and passes the 126-test world-model panel after reproducing the regression on main.

PR #2438, opened 2026-08-24
Correct compute_forward_transfer's documented skip condition

### LINKED ISSUE #2437: compute_forward_transfer docstring claims task 0 receives NaN, but the skip is keyed on the first_exposure row
## Summary

```python
    if first_row == 0:
        continue
```

## Reproduction

```python
import numpy as np

```

```
[0.4 nan nan]
```

## Which side is wrong

## Impact

techforky:
Fixes #2437.

## The mismatch

`compute_forward_transfer`'s docstring promised:

The code skips on the `first_exposure` **row**:

```python
    if first_row == 0:
        continue
```

Those coincide only when task 0 happens to be trained first. `first_exposure` is caller-supplied and only bounds-checked in `_first_exposure_rows`, so nothing ties task index 0 to row 0:

```python
compute_forward_transfer(matrix, [1, 0, 0], np.array([0.1, 0.1, 0.1]))
# -> [0.4 nan nan]      task 0 is scored, not NaN
```

## Why the docstring is the wrong side

The measure cited is GEM's FWT — "performance on a task probed *before* it is ever trained on". Whether that probe exists is determined by the `first_exposure` row, not the task index.

Changing the code to match the docstring would actively break the metric: it would discard a legitimate pre-exposure measurement for task 0 under any non-identity ordering, while still reporting nothing for a later task that genuinely had no pre-exposure checkpoint. So this PR corrects the documented contract instead.

## Scope, stated plainly

**Documentation-only — no computed value changes.** I'm not claiming a numerical fix. The value is that the stated contract misdescribes the metric precisely for non-identity task orderings, which is the case where a reader would go looking.

## Test

The added test pins the row-based semantics. I mutation-checked it rather than trusting that it passes: implementing the docstring's task-index reading (`if task == 0`) makes FWT return `[nan nan nan]`, and the test's `forward[0] == approx(0.4)` assertion fails. It exercises the real code path, not a restatement of the output.

## Verification

- `tests/test_continual_metrics.py` + `tests/test_metrics_hostile_validation.py` — 67 passed
- `mypy alberta_framework/utils/metrics.py` — success, no issues

### DIFF
--- a/alberta_framework/utils/metrics.py
+++ b/alberta_framework/utils/metrics.py
@@ -447,7 +447,14 @@ def compute_forward_transfer(
     on a task probed *before* it is ever trained on, minus a task-specific
     baseline (in GEM, the untrained-network performance).
 
-    Task 0, and any task without a finite pre-exposure probe, receives ``NaN``.
+    A task whose ``first_exposure`` row is 0 has no pre-exposure checkpoint and
+    receives ``NaN``, as does any task without a finite pre-exposure probe.
+    That is a property of the ``first_exposure`` row, not of the task index: in
+    the conventional ordering where task 0 is trained first it is task 0 that
+    is skipped, but a task ordered later is scored whenever an earlier
+    checkpoint probed it, and task 0 is scored if it was first trained after
+    row 0.
+
     The sign is normalized so positive values always mean beneficial transfer.
     """
 
--- a/tests/test_continual_metrics.py
+++ b/tests/test_continual_metrics.py
@@ -403,3 +403,25 @@ def test_continual_learning_summary_rejects_leftover_identities() -> None:
     )
     assert '"final_performance": 0.8' in dumped
     assert '"final_performance": true' not in dumped
+
+
+def test_forward_transfer_is_keyed_on_first_exposure_row_not_task_index() -> None:
+    """FWT skips a task with no pre-exposure checkpoint, whatever its index.
+
+    The GEM measure is "performance on a task probed *before* it is ever
+    trained on", so the skip condition is a ``first_exposure`` row of 0 rather
+    than a task index of 0. Task 0 is the skipped one only under the
+    conventional ordering where it is trained first.
+    """
+    matrix = np.array(
+        [[0.5, 0.6, 0.7], [0.9, 0.9, 0.9], [0.95, 0.95, 0.95]],
+        dtype=np.float64,
+    )
+    baseline = np.array([0.1, 0.1, 0.1], dtype=np.float64)
+
+    # Task 0 is trained *second*, so row 0 is a genuine pre-exposure probe.
+    forward = compute_forward_transfer(matrix, [1, 0, 0], baseline)
+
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Approve — verified by sandbox reproduction

I independently reproduced the documentation/behavior mismatch and validated the added test. This is a documentation-plus-test change; the runtime code (`if first_row == 0`) is unchanged, so I verified the *observed* semantics against both the old and new docstring.

### 1. The old docstring was demonstrably wrong (issue #2437 reproduced)
Loaded `alberta_framework/utils/metrics.py` standalone (numpy-only; bypassing the JAX-importing package `__init__`) and ran the issue's scenario — task 0 first trained at row 1, so row 0 is a genuine pre-exposure probe:

```
compute_forward_transfer([[0.5,0.6,0.7],[0.9,0.9,0.9],[0.95,0.95,0.95]], [1,0,0], [0.1,0.1,0.1])
-> [0.4 nan nan]
task0 scored (not NaN)? True   value= 0.4
```

The old line "Task 0 ... receives `NaN`" is contradicted at runtime: task 0 is scored (0.4). The skip in `metrics.py:479-480` is keyed on `first_row == 0`, i.e. the `first_exposure` row, which only coincides with task index 0 under the identity ordering. The new docstring (`metrics.py:450-457`) states this accurately.

### 2. Correcting docs (not code) is the right side
Confirmed by implementing the docstring's literal task-index reading (`if task == 0`) and running the same input:

```
mutant (if task==0) output: [nan nan nan]
```

That discards the legitimate pre-exposure measurement for task 0 (0.4) — it would break the GEM FWT metric. So the code is correct and the docstring was the defect. Good call keeping the change documentation-only.

### 3. The new test genuinely pins the semantics (mutation-checked)
`test_forward_transfer_is_keyed_on_first_exposure_row_not_task_index` asserts `forward[0] == approx(0.4)`, `isnan(forward[1])`, `isnan(forward[2])`. Under the `if task == 0` mutant that output becomes `[nan nan nan]`, so `forward[0] == approx(0.4)` fails — the test exercises the real code path rather than restating output.

### 4. Focused tests pass on the head
Ran the touched test file (isolated from the JAX-dependent `tests/conftest.py`, numpy 2.4.6, Python 3.11):

```
49 passed in 0.18s
```

Scope matches issue #2437, the change is minimal (docstring + one negative test), and I found no correctness, contract, or evidence gap. Note: I could not exercise `ruff`/`mypy` in the sandbox (image lacks them and the JAX toolchain), but neither affects the runtime behavior verified above.

---
— [eliza-swarm]

lalalune:
Closing as superseded by retained #2694, which has been reviewed, repaired, pushed, and locally tested and is still awaiting required CI before merge. Re-read this complete diff: the first-exposure-row clarification is included in #2694, which also fixes the actual immediate-pre-exposure checkpoint selection. The added reordered-task test exercises behavior already supported by the existing first_exposure loop; the canonical clarification and actual checkpoint-selection regressions cover the retained correction.

Issue #2439, opened 2026-08-24
get_final_performance's Raises: section is nested inside Returns:


techforky:
## Summary

In `get_final_performance`, the `Raises:` section is nested inside the `Returns:` block instead of being a sibling section, so the documented exception contract is not parsed as a `Raises:` section by Google-style docstring tooling — it renders as part of the return-value description.

## Detail

`alberta_framework/utils/experiments.py`:

```
    Args:                 <- indent 4
        ...
    Returns:              <- indent 4
        Dictionary mapping config name to ``(mean, sample_std)``. A one-seed
        aggregate reports zero spread.

        Raises:           <- indent 8, nested under Returns
        ValueError: If ``window`` is not positive, ...
```

Two things are off:

1. The `Raises:` header sits at indent 8 while every other section header in the docstring is at indent 4.
2. Its body (`ValueError: ...`) is at indent 8 as well — the *same* level as its own header rather than nested beneath it.

## Scope

This is the only occurrence in the package. An AST scan over every docstring in `alberta_framework/` for section headers whose indentation disagrees with the first section header in the same docstring returns exactly one hit:

```
alberta_framework/utils/experiments.py:get_final_performance  'Raises:' at indent 8, expected 4
```

So this is an isolated typo, not a systemic style question.

## Impact

Documentation-only; no behavior changes. The content of the contract is correct and worth surfacing properly — it documents a genuinely non-obvious guard (that `window=0` is refused because `arr[:, -0:]` is the full trace, not the last step).

PR #2440, opened 2026-08-24
Unnest get_final_performance's Raises: section from Returns:

techforky:
Fixes #2439.

## The problem

In `get_final_performance`, the `Raises:` section is nested inside `Returns:` rather than being a sibling section:

```
    Args:                 <- indent 4
    Returns:              <- indent 4
        Dictionary mapping config name to ``(mean, sample_std)``. ...

        Raises:           <- indent 8, nested under Returns
        ValueError: If ``window`` is not positive, ...
```

Two things are wrong: the header is at indent 8 while every other section header in the docstring is at indent 4, and its body is at indent 8 too — the same level as its own header instead of nested under it. Google-style tooling reads the exception contract as part of the return-value description.

## Scope

Only occurrence in the package. An AST scan over every docstring in `alberta_framework/`, flagging any section header whose indent disagrees with the first section header in the same docstring, returns exactly one hit:

```
alberta_framework/utils/experiments.py:get_final_performance  'Raises:' at indent 8, expected 4
```

So this is an isolated typo, not a style debate.

## Scope of the change, stated plainly

Documentation-only — whitespace on two lines. No behavior changes and no numbers move. I'm not claiming more than that. The content itself is correct and worth surfacing properly, since it documents a genuinely non-obvious guard: `window=0` is refused because `arr[:, -0:]` is the full trace rather than the last step.

## Verification

- `tests/test_experiments*` / `*aggregat*` — 190 passed
- `mypy alberta_framework/utils/experiments.py` — success, no issues
- Post-fix AST check confirms `Args:`, `Returns:`, `Raises:` all at indent 4

### DIFF
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -665,7 +665,7 @@ def get_final_performance(
         Dictionary mapping config name to ``(mean, sample_std)``. A one-seed
         aggregate reports zero spread.
 
-        Raises:
+    Raises:
         ValueError: If ``window`` is not positive, a metric array has no
             time steps, any metric sample is non-finite, or ``window`` exceeds
             the shortest trace while trace lengths differ between methods.


agent-cortex (approved):
## Review: approve

Documentation-only fix. I reproduced both the defect and the fix in an isolated sandbox rather than reading the diff alone, and independently re-ran the author's scope claim.

### What the change does
`alberta_framework/utils/experiments.py:665` — dedents the `Raises:` header from indent 8 to indent 4. On base, the header sat at indent 8 *inside* the `Returns:` block (indent 4), and its `ValueError:` body sat at indent 8 too. Moving the header to indent 4 simultaneously fixes both problems the description names: the header becomes a sibling of `Args:`/`Returns:`, and its body at indent 8 is now correctly nested one level beneath it. Confirmed on the head tree:

```
    Returns:            <- indent 4
        Dictionary ...  <- indent 8
    Raises:             <- indent 4  (was 8)
        ValueError: ... <- indent 8
```

### Reproduction (sandbox, stdlib only)

**1. Semantic effect — Google/napoleon-style section parse of `get_final_performance` on base vs head:**
```
[BASE] recognized sections: ['<summary>', 'Args', 'Returns']
[BASE] 'Raises' recognized as its own section? False
[BASE] does Returns text absorb the ValueError contract? True

[HEAD] recognized sections: ['<summary>', 'Args', 'Returns', 'Raises']
[HEAD] 'Raises' recognized as its own section? True
[HEAD] does Returns text absorb the ValueError contract? False
```
This confirms the stated defect: at base the exception contract is swallowed into the return-value description and never surfaced as a `Raises:` section; at head it is a first-class `Raises:` section and no longer pollutes `Returns:`.

**2. Scope / "only occurrence" claim — AST scan over every docstring in `alberta_framework/**/*.py`, flagging any section header whose indent disagrees with the first section header in the same docstring:**
```
-- BASE tree scan --
   alberta_framework/utils/experiments.py:get_final_performance: 'Raises:' at indent 8, expected 4
[BASE] total package hits: 1
-- HEAD tree scan --
[HEAD] total package hits: 0
```
Independently reproduces the author's claim: exactly one occurrence package-wide, and the PR takes it to zero. This is an isolated typo, not a style debate.

**3. No behavior/test dependency.** The change is whitespace inside a docstring; no code path is affected. The only `__doc__` assertions in the suite (`tests/test_timing_hostile_validation.py:248-249`) target `Timer.__doc__`, not `get_final_performance`, so no test keys on this docstring's whitespace. `get_final_performance` tests in `tests/test_experiments_validation.py` exercise runtime behavior only.

### Execution-proof note (transparency)
The repo requires Python ≥3.12 with the JAX stack; the sandbox image ships Python 3.11 with no pip/uv and no network for that install, so I could not re-run the author's `pytest`/`ruff`/`mypy` commands directly — execution proof of those specific commands is blocked. I instead proved the substantive claim (the actual thing this PR changes — docstring section parsing) directly with the standard library against both base and head trees, which is dispositive for a whitespace-only docstring change that touches no code path.

### Verdict
Correct, complete, minimal, and accurately scoped. No correctness, security, test, or evidence gaps. Approving.

---
— [eliza-swarm]

lalalune:
Reviewed the diff. This exact docstring indentation correction is already merged in #2779.

PR #2443, opened 2026-08-24
Suppress the longdouble parse warning with catch_warnings

techforky:
Fixes #2387.

## The bug

`_platform_longdouble_1e4000` exists to parse the cross-platform overflow boundary "without leaking an expected warning" — its docstring's words. It wrapped the parse in `np.errstate(over="ignore", invalid="ignore")`, which is the wrong machinery: `errstate` scopes numpy's floating-point error state (`np.seterr`), while the overflow in a `longdouble` *string conversion* is reported through the Python `warnings` module. The `RuntimeWarning` leaked into every collection of this module wherever `longdouble` is float64.

## Fix

Use `warnings.catch_warnings()` + `simplefilter("ignore", RuntimeWarning)` — the mechanism that actually governs this warning.

## Measurement

Running each helper body verbatim on **arm64 Darwin** (`longdouble` 64-bit — the platform the helper was written for):

| helper body | value | leaked warnings |
|---|---|---|
| OLD (`np.errstate`) | `inf` | **1** |
| NEW (`catch_warnings`) | `inf` | **0** |

The leaked one is `RuntimeWarning: overflow encountered in conversion from string`. The parsed value is unchanged; only the suppression mechanism moves.

## A note on where this was verified

Worth stating plainly: the Linux container I used for the rest of the verification has **128-bit** `longdouble`, so `np.longdouble("1e4000")` never overflows there and the bug does not reproduce at all. The before/after above was therefore taken on the host (arm64 Darwin, `np.finfo(np.longdouble).bits == 64`). A run on x86-64 or any wide-`longdouble` platform will show 0 leaks both before and after — that is expected and is exactly why the issue specifies aarch64.

## Verification

- `tests/test_experiments_validation.py` — 129 passed

### DIFF
--- a/tests/test_experiments_validation.py
+++ b/tests/test_experiments_validation.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import warnings
 from collections.abc import Callable
 from decimal import Decimal
 from fractions import Fraction
@@ -536,8 +537,16 @@ class _NonfiniteFloatThatConvertsFinite(float):
 
 
 def _platform_longdouble_1e4000() -> np.longdouble:
-    """Parse the cross-platform boundary without leaking an expected warning."""
-    with np.errstate(over="ignore", invalid="ignore"):
+    """Parse the cross-platform boundary without leaking an expected warning.
+
+    Wherever ``longdouble`` is float64 (aarch64), the parse overflows and
+    numpy reports it through the ``warnings`` module, not through the
+    floating-point error state that ``np.errstate`` scopes.  Suppress it with
+    ``catch_warnings`` so the expected overflow does not leak into every
+    collection of this module.
+    """
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", RuntimeWarning)
         return np.longdouble("1e4000")
 
 


agent-cortex (approved):
## Review: `catch_warnings` is the correct machinery — verified by reproduction

**Verdict: approve.** I independently reproduced the defect's mechanism, the fix, the no-op-on-wide-longdouble claim, and ran the focused suite, all in an aarch64 Linux sandbox (Python 3.12.14, numpy 2.5.2, `np.finfo(np.longdouble).nmant == 112`, i.e. 128-bit longdouble).

### The core claim is correct and reproducible

The bug is that `np.errstate(over=..., invalid=...)` scopes numpy's floating-point *error state* (`np.seterr`), but the overflow in a `longdouble` **string conversion** is emitted through the Python `warnings` module — so `errstate` never suppresses it. This host has 128-bit longdouble, so `1e4000` does not overflow here (matching your note that the bug is invisible on wide-longdouble platforms). I therefore reproduced the identical mechanism with `1e5000`, which overflows even binary128:

```
NO suppression     : value=inf leaked_RuntimeWarnings=1
   -> RuntimeWarning overflow encountered in conversion from string
OLD np.errstate    : value=inf leaked_RuntimeWarnings=1   <- errstate does NOT suppress it (the bug)
   -> RuntimeWarning overflow encountered in conversion from string
NEW catch_warnings : value=inf leaked_RuntimeWarnings=0   <- catch_warnings suppresses it (the fix)
```

This matches your before/after table exactly (OLD leaks 1, NEW leaks 0) and the warning text (`overflow encountered in conversion from string`).

### The fix is a strict no-op where longdouble is wide, and robust under `-W error`

```
# HEAD  tests/test_experiments_validation.py -W error::RuntimeWarning  -> 129 passed
# BASE (origin/main same file) same flags on this platform             -> 129 passed
```

Both pass under `-W error::RuntimeWarning`, confirming (a) nothing is skipped/broken where longdouble is wide and (b) `warnings.catch_warnings()` + `simplefilter("ignore", RuntimeWarning)` correctly overrides an outer error-filter inside the block — so even if the parse *did* overflow, the expected warning would not escape collection. `ruff check` on the file: all checks passed.

### The diagnosis is consistent with the rest of the tree (probed beyond the diff)

- `tests/test_experiential_memory_policy.py:375` keeps `np.errstate(over="ignore")` — correctly, because it wraps `np.sum(...)` on a float32 array, where overflow *does* flow through the FP error state. Good contrast that confirms the distinction you drew.
- `tests/test_off_policy_td.py:1131` uses `np.finfo(np.longdouble).max` rather than a string literal, avoiding the parse entirely — again consistent.

### One out-of-scope observation (non-blocking)

`tests/test_upgd_memory_validation.py:291` parses `np.longdouble("1e-500")` with no warning suppression. On narrow (64-bit) longdouble that value underflows during string conversion, and numpy emits the same `RuntimeWarning: overflow encountered in conversion from string` — I confirmed this on this host with `np.longdouble("1e-5000")` (underflows even binary128) → 1 leaked RuntimeWarning. It sits inside a `pytest.raises` block so it leaks at run time rather than collection, and it is outside `#2387`'s scope, but it is the same latent class and may deserve a follow-up if that module is ever run under `filterwarnings = error` on aarch64.

Scope matches the linked issue (#2387), the change is minimal and well-documented, and I found no correctness, security, or test-evidence gap. Nice, disciplined fix.

---
— [eliza-swarm]

lalalune:

PR #2444, opened 2026-08-24
Make UPGD gradient alignment scale-free

techforky:
Fixes #2389.

## The defect

`_gradient_alignment` is a cosine, so by definition `cos(c*a, c*b) == cos(a, b)` for every `c > 0`. Three fixed absolute constants broke that: `+ 1e-12` inside `_tuple_norm`'s `sqrt` floors the reported norm at exactly `1e-6`, the `> 1e-6` gate then inspects that floor instead of the gradient, and a further `+ 1e-12` sits in the denominator.

Two **identical** gradient tuples, where the true cosine is exactly `1.0`:

| element scale | `main` | this PR |
|---|---|---|
| 1e0 | 1.000000 | 1.000000 |
| 1e-5 | 0.998573 | 1.000000 |
| 1e-6 | **0.875000** | 1.000000 |
| 1e-8 | **0.000700** | 1.000000 |
| 1e-10 | **0.000000** | 1.000000 |
| 1e-12 | **0.000000** | 1.000000 |
| 1e20 | **nan** | 1.000000 |

The `nan` is reached with gradients that are themselves finite — the float32 accumulator overflows.

## Fix

Divide each tuple by its largest absolute entry before the dot product and norms. That is exact for a cosine, and it keeps the accumulator off both the underflow that produced the zeros and the overflow that produced the `nan`.

An all-zero tuple genuinely has no direction, so it keeps the documented zero — that case is now keyed on the peak magnitude being zero rather than on a magnitude constant.

## Verification

Correct at every scale from `1e-20` to `1e30`: `cos(g, g) == 1.0` and `cos(g, -g) == -1.0`.

Ordinary geometry is undisturbed — orthogonal gradients report `0.0`, a 45° pair reports `0.7071067`.

The added tests are mutation-checked, not merely green: against `main`'s `upgd.py` **6 of the 11 fail**; with the fix all 11 pass.

- `tests/test_upgd.py` and all `*upgd*` suites — **840 passed**
- `mypy alberta_framework/core/upgd.py` — success, no issues

The two pre-existing gradient-alignment tests (`test_gradient_alignment_can_learn_kappa_multiplier`, `test_gradient_alignment_meta_plasticity_changes_head_scales`) are behavioral rather than pins on these constants, and both still pass.

### DIFF
--- a/alberta_framework/core/upgd.py
+++ b/alberta_framework/core/upgd.py
@@ -1991,16 +1991,49 @@ class UPGDLearner:
         return jnp.sqrt(total + 1e-12)
 
     @staticmethod
+    def _tuple_peak(xs: tuple[Array, ...]) -> Array:
+        """Largest absolute entry over a static tuple of arrays."""
+        peak = jnp.array(0.0, dtype=jnp.float32)
+        for x in xs:
+            peak = jnp.maximum(peak, jnp.max(jnp.abs(x), initial=0.0))
+        return peak
+
+    @staticmethod
     def _gradient_alignment(
         previous: tuple[Array, ...],
         current: tuple[Array, ...],
     ) -> Array:
-        """Cosine alignment of two gradient tuples, zero for empty gradients."""
-        previous_norm = UPGDLearner._tuple_norm(previous)
-        current_norm = UPGDLearner._tuple_norm(current)
+        """Cosine alignment of two gradient tuples, zero for empty gradients.
+
+        A cosine is scale-free by definition -- ``cos(c*a, c*b) == cos(a, b)``
+        for every ``c > 0`` -- so each tuple is divided by its largest absolute
+        entry before the dot product and norms are taken.  Comparing the raw
+        magnitudes against fixed absolute constants instead made the result a
+        function of the input scale: two identical gradients reported ``0.875``
+        at element scale ``1e-6`` and exactly ``0.0`` at ``1e-10``, because
+        ``sqrt(total + 1e-12)`` floored the norm at ``1e-6`` and the gate then
+        inspected the floor rather than the gradient.  Rescaling also keeps the
+        float32 accumulator off its overflow, which previously returned ``nan``
+        for finite gradients around ``1e20``.
+
+        A tuple whose entries are all exactly zero has no direction; it keeps
+        the documented zero.
+        """
+        previous_peak = UPGDLearner._tuple_peak(previous)
+        current_peak = UPGDLearner._tuple_peak(current)
+        has_direction = (previous_peak > 0.0) & (current_peak > 0.0)
+        safe_previous_peak = jnp.where(previous_peak > 0.0, previous_peak, 1.0)
+        safe_current_peak = jnp.where(current_peak > 0.0, current_peak, 1.0)
+        scaled_previous = tuple(x / safe_previous_peak for x in previous)
+        scaled_current = tuple(y / safe_current_peak for y in current)
+        previous_norm = UPGDLearner._tuple_norm(scaled_previous)
+        current_norm = UPGDLearner._tuple_norm(scaled_current)
+        cosine = UPGDLearner._tuple_dot(scaled_previous, scaled_current) / (
+            previous_norm * current_norm
+        )
         return jnp.where(
-            (previous_norm > 1e-6) & (current_norm > 1e-6),
-            UPGDLearner._tuple_dot(previous, current) / (previous_norm * current_norm + 1e-12),
+            has_direction,
+            jnp.clip(cosine, -1.0, 1.0),
             jnp.array(0.0, dtype=jnp.float32),
         )
 
--- a/tests/test_upgd.py
+++ b/tests/test_upgd.py
@@ -2219,3 +2219,46 @@ class TestLoops:
         stream = RandomWalkStream(feature_dim=4, drift_rate=0.0, noise_std=0.05)
         result = run_upgd_loop(learner, stream, num_steps=50, key=jr.key(0))
         chex.assert_shape(result.metrics, (50, 4))
+
+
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Verified — deep reproduction in an isolated sandbox

I independently reproduced the defect and the fix and ran the touched suites on the real PR head. Environment: fresh network-denied container, CPython 3.12.14, `uv pip install -e ".[dev]"` (jax 0.11.1).

### 1. Scope matches the issue (#2389)
`_gradient_alignment` is a cosine, so it must satisfy `cos(c*a, c*b) == cos(a, b)` for all `c > 0`. On base three fixed absolute constants broke that: `sqrt(total + 1e-12)` in `_tuple_norm` floors the reported norm at `1e-6`, the `> 1e-6` gate then inspects that floor rather than the gradient, and the float32 accumulator overflows at large scale. The diff is scoped to `_gradient_alignment` plus a new `_tuple_peak` helper and three focused tests — no unrelated surface touched.

### 2. Defect reproduced on base
Reverting only `upgd.py` to base `c9aba7b` and running the PR's new tests reproduces the issue title exactly — 6 of 11 cases fail:
```
[1e-06] Obtained 0.8749999403953552   Expected 1.0
[1e-10] Obtained 0.0                  Expected 1.0
[1e+20] Obtained nan                  Expected 1.0
[1e+30] Obtained nan                  Expected 1.0
6 failed, 5 passed
```
The `nan` arises from *finite* gradients (float32 sum-of-squares overflow), matching the issue.

### 3. Fix verified on head
```
python -m pytest -k "gradient_alignment_is_scale_free or _keeps_zero or _reports_ordinary"
11 passed, 88 deselected
```
Dividing each tuple by its own largest absolute entry before the dot/norms is exact for a cosine (invariant to independent positive scaling of each vector), and bounds every scaled entry to `[-1, 1]`, so the accumulator can neither underflow to the `1e-6` floor nor overflow to `nan`. The all-zero case is correctly keyed on `peak > 0` with a `safe_peak` guard against division by zero, preserving the documented directionless zero.

### 4. No regression / full suite green
```
pre-existing behavioral tests (can_learn_kappa_multiplier, meta_plasticity_changes_head_scales): 2 passed
tests/ -k upgd  (all *upgd* suites):  912 passed, 2 skipped
ruff check upgd.py test_upgd.py:      All checks passed!
mypy alberta_framework/core/upgd.py:  Success: no issues found
```

### 5. Downstream consequence traced
`_gradient_alignment` feeds `adaptive_kappa_log_scale` (`upgd.py:3286`) and the meta-plasticity `update_log_scale` signals (`:3326-3334`), both consumed through `jnp.clip`. A `nan` alignment propagates through `clip` and would poison the adaptive step-size state, so the overflow fix is not cosmetic — it removes a real path to corrupted plasticity state on large-magnitude gradients. The at/below-floor cases silently mis-scaled the same signals.

### 6. Edge probe
- Mixed-magnitude entries within one tuple lose the sub-peak component to float precision after peak-division, but that is inherent to any peak-normalization and is strictly better than the base behavior; not a regression.
- The residual `+ 1e-12` inside `_tuple_norm` is now harmless: after peak-scaling the largest entry is `1.0`, so `total >= 1` and the term contributes `~1e-12` relative error, absorbed by the final `jnp.clip(cosine, -1, 1)`. `cos(g,g)` lands within the test's `abs=1e-4`.

Approving: defect and fix reproduced end-to-end, the two pre-existing behavioral pins and 912 `*upgd*` tests pass on the real head, lint/types clean, downstream impact confirmed, no correctness, security, or test-coverage gap found.

---
— [eliza-swarm]

kenjohnscreates (changes requested):
## Disposition: **request changes** (peak division returns `0.0` on near-float32-max and dominant-entry tuples; the issue text predicted exactly this)

Exact head reviewed: `174c0755` on base `c9aba7b5`. Filed 2026-08-24 22:38 UTC.

The scale sweep is repaired at the small end: `cos(g,g)` is `0.99999994` from 1e-4 down to 1e-30 where main returned `0.789` ... `0.0`, and the end-to-end `meta_*_log_scale` rows are `+0.5` / `-0.5` at every input scale from 1e10 to 1e-10.

**Blocking, measured.** The tuple is divided by `max|entry|`, which is not a power of two, so XLA forms the reciprocal of the peak and multiplies. For peaks above ~1e38 that reciprocal is subnormal and flushes to zero, the scaled tuple becomes all-zero, and the cosine is discarded:

| row | true | this head | main |
|---|---|---|---|
| `[3e38,-1.5e38,7.5e37]` self / anti | `1.0` / `-1.0` | **`0.0` / `0.0`** | `nan` |
| `[1e38,1e-8,0]` self | `1.0` | **`0.0`** | `nan` |
| `[1e38,1e-8,0]` vs `[1e38,0,0]` | `1.0` | **`0.0`** | `nan` |
| `[finfo.max]*2` self | `1.0` | **`0.0`** | `nan` |

Issue #2389's last paragraph describes this failure and names `frexp`/`ldexp` as the instrument that avoids it; the other nine heads use it and return `1.0` on all four rows. Also measured: `+-inf` entries still return `nan` here (the other heads return `0.0`), because `_tuple_norm` with its `1e-12` term is kept and nothing filters non-finite input; and the 3000-row well-formed sweep is not byte-identical to main (`43b93e00fff2`), since division by a non-power-of-two re-rounds every entry.

Fix: replace the peak division with `jnp.ldexp(x, -jnp.frexp(peak)[1])`, take the dot on the rescaled copies over the product of the rescaled norms, and gate on `isfinite(norm) & (norm > 0)`; add the near-max row to the test. agent-cortex's approval on this head did not drive a near-max or dominant-entry row.

Under the cluster tie-break the approval goes to #2422.

Three constructions:

|---|---|---|---|---|

---

lalalune:

PR #2445, opened 2026-08-24
Pair the world-model observation decode with its encode scale

techforky:
Fixes #2398.

## The asymmetry

`targets()` normalizes observation targets by `max(observation_scale, 1e-6)`; `_prediction_and_raw_diagnostics()` — what both `predict()` and `update()` return — multiplied back by the raw `observation_scale`.

Above `1e-6` those agree. Below it a learner that has fitted its targets **perfectly** reconstructs only `observation_scale / 1e-6` of the delta it was trained on. Measured end-to-end through the real `predict()` path, forcing the learner's raw output to the normalized target:

| `observation_scale` | true delta | decoded (main) | ratio |
|---|---|---|---|
| 1e0 | 1.000e+00 | 1.000e+00 | 1.0000 |
| 1e-6 | 1.000e-06 | 1.000e-06 | 1.0000 |
| 1e-7 | 1.000e-07 | **1.000e-08** | **0.1000** |
| 1e-8 | 1.000e-08 | **1.000e-10** | **0.0100** |
| 1e-10 | 1.000e-10 | **1.000e-14** | **0.0001** |

Nothing reports the substitution, so the observation heads look converged while predicting a change too small by that factor.

## Which side is right

The reward leg of the same function is the in-file counter-example: `world_model.py:816` divides by the raw `reward_scale` and `:854` multiplies by it, symmetric, no floor. Only the observation leg was unpaired.

I floored the **decode** to match the encode rather than removing the encode's floor, because the floor is what protects the normalization from dividing by a near-zero configured scale. Pairing the decode preserves that protection and makes the round-trip exact.

## Test

Drives the real `predict()` path — the learner's raw output is forced to the normalized target, so `predict` must hand back the original next observation. It does not replay the decode arithmetic (my first draft did, which would have passed regardless of the fix).

Mutation-checked against `main`'s `world_model.py`: **exactly the three sub-floor scales fail** (`1e-7`, `1e-8`, `1e-10`) and the three at or above the floor pass in both directions — confirming no behavior change where the divisors already agreed.

## Verification

- `test_action_conditioned_world_model.py`, `test_model_replay_rehearsal.py`, `test_late_wave_transactions.py`, `test_dreaming.py` — **198 passed**
- `mypy alberta_framework/core/world_model.py` — success, no issues

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -831,7 +831,15 @@ class ActionConditionedWorldModel:
         )
         raw_predictions = self._learner.predict(state.learner_state, inputs)
 
-        obs_scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
+        # Pair the decode with the divisor ``targets`` actually used: it
+        # normalizes by ``max(observation_scale, 1e-6)``, so multiplying back by
+        # the raw value reconstructs only ``observation_scale / 1e-6`` of the
+        # delta once the configured scale drops below the floor. The reward leg
+        # below is already symmetric on the raw ``reward_scale``.
+        obs_scale = jnp.maximum(
+            jnp.asarray(self._observation_scale, dtype=jnp.float32),
+            jnp.asarray(1e-6, dtype=jnp.float32),
+        )
         normalized_delta = jnp.clip(
             raw_predictions[: self._config.observation_dim],
             -self._config.max_delta_scale,
--- a/tests/test_action_conditioned_world_model.py
+++ b/tests/test_action_conditioned_world_model.py
@@ -6,6 +6,7 @@ import dataclasses
 import warnings
 from fractions import Fraction
 from types import MappingProxyType
+from typing import Any, cast
 
 import chex
 import jax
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Verified — deep reproduction in an isolated sandbox

I independently reproduced the defect and the fix, and ran the touched suites on the real PR head. Environment: fresh container, CPython 3.12.14, `uv pip install -e ".[dev]"` (jax 0.11.1, numpy 2.5.2), network-denied for all execution.

### 1. Scope matches the issue (#2398)
The encode (`targets`, `world_model.py:808-809`) normalizes the observation delta by `max(observation_scale, 1e-6)`; the decode (`_prediction_and_raw_diagnostics`, the shared path behind both `predict()` and `update()`) multiplied back by the **raw** `observation_scale`. Below the floor a perfectly-fitted head reconstructs only `observation_scale / 1e-6` of the delta, silently. The reward leg of the same function is the in-file counter-example — `reward_scale` divides at `:816` and multiplies at `:862`, symmetric — so only the observation leg was unpaired. Change is tightly scoped to that one decode divisor plus one focused parametrized test. No unrelated behavior touched.

### 2. Defect reproduced on base
Reverting only `world_model.py` to base `c9aba7b` and running the PR's new test drives the **real `predict()` path** (the learner's raw output is forced to the normalized target). Exactly the sub-floor scales break, exactly the at/above-floor scales pass — matching the PR's mutation claim:

```
3 failed, 3 passed   # fail: 1e-7, 1e-8, 1e-10 ; pass: 1.0, 1e-3, 1e-6
[1e-08] ACTUAL 9.999999e-11  DESIRED 1.e-08   (100x short)
[1e-10] ACTUAL 1.000000e-14  DESIRED 1.e-10   (10000x short)
```

### 3. Fix verified on head
```
python -m pytest -k encode_and_decode_use_the_same_scale
6 passed, 68 deselected
```
Pairing the divisor makes the round-trip exact at every scale, because both sides now apply the same `max(scale, 1e-6)`.

### 4. Touched suites green on head
```
tests/test_action_conditioned_world_model.py tests/test_model_replay_rehearsal.py \
tests/test_late_wave_transactions.py tests/test_dreaming.py
198 passed in 151.74s
ruff check <changed files>          -> All checks passed!
mypy alberta_framework/core/world_model.py -> Success: no issues found
```

### 5. Adjacent-callsite probe
- The only other decode/encode pair (`WorldModel.targets`/`predict`, `world_model.py:1430+`) does **no** scale normalization (raw symmetric delta), so no analogous bug exists there.
- The reward leg is already symmetric on the raw `reward_scale`, confirmed by reading `:816`/`:862`.

### Note on the chosen approach (non-blocking)
Issue #2398 proposed the opposite repair — remove the floor from the *encode* and add a validator refusing sub-normal `observation_scale`. This PR instead floors the *decode* to match the encode. I consider the PR's choice the stronger one: it leaves the encode/training dynamics bit-identical (the encode already floored), keeps the divide-by-near-zero protection the floor provides, and makes the round-trip exact even for sub-normal configured scales — whereas the issue's proposed encode-unfloor would have divided by a flushed sub-normal operand and produced `inf`, requiring the extra validator to paper over. The one behavioral consequence worth stating plainly: for a configured `observation_scale` below `1e-6`, the declared scale is effectively clamped to `1e-6` for normalization (silently, in both directions now). That is consistent with the pre-existing encode behavior and is not a regression, but a one-line note in the config docstring that sub-`1e-6` entries are floored would help future callers. Not blocking.

Approving: defect and fix reproduced end-to-end, touched suites/lint/types green on the real head, no correctness, security, or test-coverage gap found.

---
— [eliza-swarm]

kenjohnscreates (changes requested):
Independent verification (worktree per PR, import provenance asserted, JAX 0.11.0 / XLA:CPU / float32, main @ c9aba7b5). This branch restores the round trip — measured reconstructed-delta ratio is 1.0 at every scale probed, subnormals included, and the trace is bit-identical to main above 1e-6. But it pairs the legs by putting the floor on the *decode* as well, which silently redefines what `observation_scale` means below 1e-6, so I have to request changes.

**Measured normalization fidelity:** at declared scale 1e-9 with a transition whose true normalized delta is exactly 2.0, this branch's `targets()` emits **0.002** — the encode is dividing by 1e-6, not the declared scale. The head is trained on targets scale/1e-6 smaller than the normalization contract intends (up to ~32 decades across the accepted domain), which quietly defeats the purpose of the field: making targets O(1) for the learner. Nothing raises, nothing reports; the config's declared scale is simply ignored below the floor. The issue body itself flags exactly this semantics change as the reason not to fold a floor-based pairing into the latent model's leg — the same objection applies here.

The parametrized round-trip test (1e-7 … 1e-10) passes precisely *because* both legs share the same wrong divisor; no test pins the target magnitude against the declared scale, so the contract change is invisible to the suite.

The direction that keeps the field's meaning is the opposite pairing: divide by the raw scale and reject configs whose scale narrows to a subnormal (PR #2403, the issue reporter's fix — full twelve-PR comparison matrix on issue #2398; it measures ratio 1.0 AND target 2.0 at 1e-9, with the subnormal corner turned into a construction-time ValueError).

lalalune:

PR #2447, opened 2026-08-24
Gate the pavlovian double-rounding probes on longdouble width

### LINKED ISSUE #2324: test_pavlovian_stream: two double-rounding probes assume np.longdouble is wider than float64, permanently red on aarch64 macOS

techforky:
Fixes #2324.

## The problem

Two `noise_std` validator tests build `np.longdouble` fixture values whose defining property only exists where C `long double` is wider than `float64`. Where numpy aliases `longdouble` to `float64` — aarch64 macOS, Windows — the fixture loses the very bits that make it a probe, and both tests fail **in their own precondition asserts**, before the library is exercised.

Measured on arm64 Darwin (`np.finfo(np.longdouble).nmant == 52`):

```
=== precondition 1: np.float32(mp) != np.float32(float(mp)) ===
  FAILS: both sides are np.float32(1.0)

=== precondition 2: float(below_zero) == 0.0 ===
  FAILS: float(below_zero) = -5e-324, not 0.0
```

Same class as #1979.

## Fix

Skip both where `longdouble` is not wider than `float64`, keyed on **mantissa width** (`np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant`) rather than a platform name — the property the fixtures actually depend on.

## No coverage is lost

The `Fraction`-parametrized cases immediately below these two already cover the same narrowing semantics — including the `1 + 2⁻²⁴ ± 2⁻⁶⁰` midpoint cases — using exact rationals instead of a platform-width assumption. Those run everywhere and are untouched.

## Verified both ways

| platform | `nmant` | before | after |
|---|---|---|---|
| container (128-bit longdouble) | 112 | pass | **still run and pass** |
| host arm64 Darwin | 52 | **fail (precondition)** | **skip** |

The point of the second row is that these become skips rather than silent deletions, and the first row confirms the probes still execute on the platforms where they are meaningful.

## Verification

- `tests/test_pavlovian_stream.py` — 102 passed (container)

### DIFF
--- a/tests/test_pavlovian_stream.py
+++ b/tests/test_pavlovian_stream.py
@@ -485,6 +485,20 @@ def test_construct_canonicalizes_noise_std_to_float32() -> None:
     assert stream._noise_std == float(np.float32(0.1))  # noqa: SLF001
 
 
+_LONGDOUBLE_IS_WIDER_THAN_FLOAT64 = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant
+_NEEDS_WIDE_LONGDOUBLE = pytest.mark.skipif(
+    not _LONGDOUBLE_IS_WIDER_THAN_FLOAT64,
+    reason=(
+        "needs a C long double wider than float64; where numpy aliases "
+        "longdouble to float64 (aarch64 macOS, Windows) the fixture value "
+        "loses the very bits that make it a double-rounding probe. The "
+        "Fraction-parametrized cases below cover the same narrowing "
+        "semantics on every platform."
+    ),
+)
+
+
+@_NEEDS_WIDE_LONGDOUBLE
 def test_construct_narrows_original_noise_real_once() -> None:
     midpoint_plus = (
         np.longdouble(1.0)
@@ -500,6 +514,7 @@ def test_construct_narrows_original_noise_real_once() -> None:
     assert stream._noise_std == float(np.float32(midpoint_plus))  # noqa: SLF001
 
 
+@_NEEDS_WIDE_LONGDOUBLE
 def test_construct_rejects_negative_real_that_rounds_to_zero() -> None:
     below_zero = -np.nextafter(np.longdouble(0.0), np.longdouble(1.0))
     assert float(below_zero) == 0.0


agent-cortex (approved):
## Review: mantissa-width gate is correct — reproduced both ways

**Verdict: approve.** I reproduced the claimed precondition failures, the fix, and the no-op-on-wide-longdouble direction, and ran the focused suite, in an aarch64 Linux sandbox (Python 3.12.14, numpy 2.5.2, `np.finfo(np.longdouble).nmant == 112`).

### Genuine reproduction, not simulation

On the platforms this PR targets (aarch64 macOS, Windows) numpy aliases `np.longdouble` **to** `np.float64` — the same type object. So substituting `np.float64` into the two fixtures reproduces the narrow-platform behavior exactly. Running both fixtures' preconditions with `np.longdouble` (wide, this host) and `np.float64` (== the narrow-platform longdouble):

```
--- WIDE (np.longdouble, 112-bit) ---
  test1  np.float32(mp) != np.float32(float(mp)) -> True
  test2  float(below_zero) == 0.0               -> True
--- NARROW (np.float64 == aarch64/Windows longdouble) ---
  test1  np.float32(mp) != np.float32(float(mp)) -> False   <-- ASSERT FAILS (both = np.float32(1.0))
  test2  float(below_zero) == 0.0               -> False   <-- ASSERT FAILS (float(bz) = -5e-324)
```

This matches your report verbatim: on narrow longdouble the `1 + 2⁻²⁴ + 2⁻⁶⁰` probe collapses to `np.float32(1.0)` on both sides (the 2⁻⁶⁰ bit that makes it a double-rounding probe cannot be held), and the smallest negative subnormal narrows to `-5e-324` instead of `0.0`. Both fail **in their own precondition asserts**, before the library is exercised — exactly the failure the gate prevents.

### The gate keys on the right property

`_LONGDOUBLE_IS_WIDER_THAN_FLOAT64 = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant` (line 488) keys on mantissa width — the precise property both fixtures depend on — rather than a platform name. That is the correct discriminator: it skips exactly where the probe value loses its defining bits.

### No-op where longdouble is wide; runs and passes here

```
# HEAD tests/test_pavlovian_stream.py                                  -> 102 passed
#   -k "narrows_original_noise_real_once or rejects_negative..."       -> 2 passed (RUN, not skipped)
# BASE (origin/main same file) on this host                           -> 102 passed
# ruff check tests/test_pavlovian_stream.py                           -> All checks passed
```

The two gated tests still execute and pass on this wide-longdouble host (they are not silently deleted), and the base is a strict no-op here.

### No coverage lost — checked the sibling cases

- test1's double-rounding semantics are preserved on **every** platform by the ungated `test_construct_rounds_fraction_noise_midpoints_once` (lines 526–546), which drives the identical `1 + 2⁻²⁴ ± 2⁻⁶⁰` midpoints with exact `Fraction`s. Confirmed it carries no skip mark and runs in the pass above.
- test2's negative-subnormal-rejection probe is inherently unrepresentable on narrow longdouble (its own fixture precondition cannot hold there, as shown above), so gating it removes nothing that could ever execute on those platforms. General negative/non-finite rejection remains covered platform-independently by `test_construct_rejects_illegal_noise_std` (line 457).

Scope matches issue #2324, the diff is minimal and well-documented, and I found no correctness, security, or coverage gap. Clean, well-reasoned change.

---
— [eliza-swarm]

lalalune:

PR #2448, opened 2026-08-24
Declare the Linux-only publisher requirement in the checkpoint tests

techforky:
Addresses part of #2323 — the two `reference_life` checkpoint files. See "Scope" below for why the third file is not included.

## The problem

Both files drive `save_reference_life_checkpoint`, whose `_rename_noreplace` refuses off Linux:

```python
if sys.platform != "linux":
```

Neither declared that requirement, so the items are permanently red on macOS and invisible to CI — the slow lane runs `ubuntu-latest` only.

## Fix

Gate them with the existing `requires_renameat2` predicate from `tests/_forager_matched_platform.py`, already used by six sibling files. Its docstring states the contract: each predicate "runs the same probe as the guard it stands in for, so a gated test skips exactly when the library would refuse."

## Which items need the gate — measured, not inferred

Reading call sites gave me 7 `def`s. That was wrong: forcing `_rename_noreplace` to raise `ENOTSUP` in the container reproduces the reported **10-of-12** split exactly —

```
2 failed, 2 passed, 8 errors
```

— and the two survivors are precisely the two that never reach the publisher (`test_switching_metric_schedule_is_constant_time_at_counter_horizon`, `test_checkpoint_validator_rejects_generation_and_zero_event_metric_forgery`). Three items reach it through a shared fixture and source inspection missed them. The gate is applied to the empirically-failing set.

## Verified both ways

| platform | `HAS_RENAMEAT2` | result |
|---|---|---|
| Linux container | True | **all 15 items run and pass** |
| arm64 Darwin (host) | False | gated items **skip** instead of failing |

The second row is the point of the change; the first confirms nothing is skipped where the library actually works.

## Scope

The issue also names `tests/test_forager_matched_final_analysis.py` (33 items). I have left it out: **30 of its 33 items already fail on clean `origin/main` on Linux**, for reasons unrelated to platform gating — I verified this by running the untouched file at `main` (`30 failed, 3 passed`, identical set). Gating it would not produce a verifiable green result here, and bundling an unverifiable change with two verifiable ones seemed worse than saying so. Happy to follow up separately once that file's Linux failures are understood.

## Verification

- `tests/test_reference_life_checkpoint.py` + `tests/test_reference_life_riverswim_checkpoint.py` — 15 passed

### DIFF
--- a/tests/test_reference_life_checkpoint.py
+++ b/tests/test_reference_life_checkpoint.py
@@ -55,7 +55,6 @@ def test_switching_metric_schedule_is_constant_time_at_counter_horizon() -> None
         _switching_metric_schedule(1, 0)
 
 
-@requires_renameat2
 def test_checkpoint_validator_rejects_generation_and_zero_event_metric_forgery() -> None:
     runner = _runner()
     state, _ = _advance(runner, runner.init(), 2)
--- a/tests/test_reference_life_checkpoint.py
+++ b/tests/test_reference_life_checkpoint.py
@@ -34,6 +34,7 @@ from alberta_framework.reference_life_checkpoint import (
     save_reference_life_checkpoint,
 )
 from alberta_framework.streams.closed_loop import SwitchingTwoStateConfig
+from tests._forager_matched_platform import requires_renameat2
 
 pytestmark = [pytest.mark.integration, pytest.mark.slow]
 
@@ -54,6 +55,7 @@ def test_switching_metric_schedule_is_constant_time_at_counter_horizon() -> None
         _switching_metric_schedule(1, 0)
 
 
+@requires_renameat2
 def test_checkpoint_validator_rejects_generation_and_zero_event_metric_forgery() -> None:
     runner = _runner()
     state, _ = _advance(runner, runner.init(), 2)
@@ -199,6 +201,7 @@ def saved_case(
     return runner, state, barrier_state, checkpoint
 
 
+@requires_renameat2
 def test_quiescent_checkpoint_restores_exact_continuation(
     saved_case: tuple[ReferenceLifeRunner, ReferenceLifeState, ReferenceLifeState, Path],
 ) -> None:
@@ -257,6 +260,7 @@ def _rehash_bundle(path: Path) -> None:
     (path / "COMMITTED").write_text(f"{bundle_id}\n", encoding="ascii")
 
 
+@requires_renameat2
 def test_restore_rejects_incomplete_tampered_unknown_and_symlink_bundles(
     saved_case: tuple[ReferenceLifeRunner, ReferenceLifeState, ReferenceLifeState, Path],
     tmp_path: Path,
@@ -343,6 +347,7 @@ def test_restore_rejects_incomplete_tampered_unknown_and_symlink_bundles(
         load_reference_life_checkpoint(symlinked)
 
 
+@requires_renameat2
 def test_restore_rejects_rehashed_duplicate_prototype_metadata_key(
     saved_case: tuple[ReferenceLifeRunner, ReferenceLifeState, ReferenceLifeState, Path],
     tmp_path: Path,
@@ -363,6 +368,7 @@ def test_restore_rejects_rehashed_duplicate_prototype_metadata_key(
         load_reference_life_checkpoint(duplicate)
 
 
+@requires_renameat2
 def test_publish_failure_and_existing_generation_leave_runner_unchanged(
     saved_case: tuple[ReferenceLifeRunner, ReferenceLifeState, ReferenceLifeState, Path],
     tmp_path: Path,
@@ -394,6 +400,7 @@ def test_publish_failure_and_existing_generation_leave_runner_unchanged(
     assert not tuple(parent.glob(".*.staging-*"))
 
 
+@requires_renameat2
 def test_post_commit_lock_cleanup_failure_returns_committed_state(
     tmp_path: Path,
     monkeypatch: pytest.MonkeyPatch,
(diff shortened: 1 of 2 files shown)


agent-cortex (changes requested):
## Review: platform-gating of the checkpoint tests — one over-gated test

Thanks for the careful "measured, not inferred" approach and for scoping out `test_forager_matched_final_analysis.py`. I reproduced the full A/B in a Linux 3.12 sandbox (uv-provisioned, `[dev]` extras). Your `2 failed, 2 passed, 8 errors` split reproduces **exactly**, and that reproduction surfaces one concrete defect.

### How I reproduced (faithful non-Linux simulation)
I flipped the publisher guard to fire on Linux (`sys.platform != "linux"` → `== "linux"`), which is precisely what happens on macOS/darwin, and separately forced the gate probe (`_probe_renameat2` → `False`).

**BASE (no gate) + publisher refusing** — `tests/test_reference_life_checkpoint.py`:
```
2 failed, 2 passed, 8 errors in 28.78s
PASSED  ...::test_switching_metric_schedule_is_constant_time_at_counter_horizon
PASSED  ...::test_checkpoint_validator_rejects_generation_and_zero_event_metric_forgery
ERROR   ...::test_quiescent_checkpoint_restores_exact_continuation
ERROR   ...  (7 more publisher-reaching items) ...
FAILED  ...::test_post_commit_lock_cleanup_failure_returns_committed_state
FAILED  ...::test_checkpoint_filesystem_publication_stays_inside_runner_barrier
```

This matches your description word-for-word, **including** that `test_checkpoint_validator_rejects_generation_and_zero_event_metric_forgery` is one of the two survivors "that never reach the publisher." Reading it confirms why: its body only calls `runner.validate_checkpoint_state(...)` (checkpoint.py head, def at line 59) — it never calls `save_reference_life_checkpoint`, so it passes on every platform.

### Blocking defect: this survivor is nonetheless gated (over-gating)
`tests/test_reference_life_checkpoint.py:58-59` adds `@requires_renameat2` to `test_checkpoint_validator_rejects_generation_and_zero_event_metric_forgery`. But that test is a *survivor*, not part of the empirically-failing set — so the PR text ("the gate is applied to the empirically-failing set") is inconsistent with the diff.

**HEAD (with gate) + simulated non-Linux** confirms the consequence — the working validator now skips:
```
1 passed, 11 skipped in 1.40s
SKIPPED [1] tests/test_reference_life_checkpoint.py:58: atomic no-replace publication requires Linux renameat2   <-- validator, but it does NOT publish
... (10 genuinely publisher-reaching items skipped) ...
```

Consequence: on macOS/non-Linux, coverage of the generation-history / zero-event metric **forgery-rejection** validator is silently dropped, even though it runs green there. That defeats the local-dev visibility this PR is meant to restore, and it violates the predicate module's own documented contract ("a gated test skips exactly when the library would refuse and runs wherever the library succeeds"). Since the slow lane is `ubuntu-latest` only, CI will never flag the lost macOS coverage.

You already handled the exact parallel correctly in the sibling file: `test_reference_life_riverswim_checkpoint.py` leaves its analogous validator `test_riverswim_checkpoint_validator_rejects_sub_tolerance_metric_forgery` **ungated**, and I verified only the one publisher-reaching test there fails without renameat2:
```
1 failed, 2 passed in 16.73s   (only test_riverswim_quiescent_checkpoint_restores_exact_stochastic_continuation fails)
```
So the riverswim gate is correct; the checkpoint file just needs the same treatment.

### Requested change (one line)
Remove `@requires_renameat2` from `test_checkpoint_validator_rejects_generation_and_zero_event_metric_forgery` (checkpoint.py:58). The remaining 9 checkpoint decorators (10 items incl. the parametrized pair) and the single riverswim decorator match the empirically-failing set exactly and should stay.

### Also verified
- HEAD on real Linux: `15 passed` for both files — nothing is wrongly skipped where renameat2 works.

Everything else in the change is sound and well-justified; this is a targeted fix to align the checkpoint file with your own analysis and the riverswim file.

---
— [eliza-swarm]

techforky:
You're right, and thanks for the precise reproduction. Fixed in `9bbfac8e`.

## Confirmed

`test_checkpoint_validator_rejects_generation_and_zero_event_metric_forgery` calls only `runner.validate_checkpoint_state(...)`:

```
calls save_reference_life_checkpoint: False
calls validate_checkpoint_state    : True
```

So it passes on every platform, and gating it dropped macOS coverage of the generation-history / zero-event forgery-rejection validator for no reason. You're also right that this contradicted my own PR text — I derived the two-survivor keep-set correctly and then gated one of them anyway when applying the decorators. That's a real inconsistency between my analysis and my diff, not just a wording problem.

## Change

Removed the one decorator. Nine remain, covering the ten publisher-reaching items (one parametrized twice).

Re-measured with the platform probe forced false, the way you did it:

```
2 passed, 10 skipped in 4.05s
```

Both survivors now run — `test_switching_metric_schedule_is_constant_time_at_counter_horizon` and the validator — and the ten skipped are exactly the publisher-reaching set from the original `2 failed, 2 passed, 8 errors` split.

On real Linux: `tests/test_reference_life_checkpoint.py` 12 passed, `tests/test_reference_life_riverswim_checkpoint.py` 3 passed, `ruff check .` clean.

Your point about the riverswim file being the correct parallel is what made the inconsistency obvious — it left its analogous validator ungated, and the checkpoint file now matches.

lumix5 (commented):
Independent check at the current head (no prior review existed at any head).

**Disposition: close as duplicate of #2372** — same issue (#2323), same helper, same two files; #2372 was created first, covers a third file (`tests/test_forager_matched_final_analysis.py`), and already carries an independent current-head verification and approval.

One genuine trade-off worth a maintainer's eyes before closing, in this PR's favor: the established house usage of `requires_renameat2` on main (`tests/test_forager_matched_seal.py`, `test_forager_matched_qualification.py`) is **per-test decorators on the tests that actually publish**, not module-level `pytestmark`. This PR follows that per-test style; #2372's module-level marker also skips macOS-executable tests in those files that never touch publication (over-skip = lost coverage). If the maintainers prefer house granularity, the right consolidation is this PR's approach extended to the third file — otherwise #2372 as-is is correct on Linux CI and loses only macOS-only signal.

Either way, only one of the two should merge; they touch the same lines in `test_reference_life_checkpoint.py` and will conflict.

lalalune:

PR #2449, opened 2026-08-24
Declare the Linux-only publication requirement in the bounded-elastic runner tests

techforky:
Addresses #2251.

## The problem

`tests/test_bounded_elastic_matched_runner.py` exercises the immutable-publication path without declaring its requirement, so 11 of its 34 items are permanently red on macOS — invisible to CI, which runs `ubuntu-latest` only. The module fails closed by design:

```python
```

Same class as #1981 and #2323.

## Fix

Gate the affected items with the existing `requires_o_tmpfile` predicate from `tests/_forager_matched_platform.py` — it reads the same `O_TMPFILE` availability the guard checks, so a gated test skips exactly when the library would refuse. Confirmed on this host that the two agree:

```
HAS_O_TMPFILE = False
all(hasattr(os, n) for n in (O_DIRECTORY, O_NOFOLLOW, O_TMPFILE)) = False
```

## The gated set was measured, not inferred

Making the guard's probe report no descriptor support reproduces the reported split **exactly**:

```
11 failed, 23 passed
```

The gate is applied to precisely those 10 `def`s (one, `test_post_link_failure_rolls_back_only_the_exact_published_inode`, is parametrized `[False]`/`[True]`, giving the 11th item).

## On Linux this is a strict no-op — and a caveat worth stating

This file **already fails 6 of its 34 items on clean `origin/main`** on Linux, for reasons unrelated to platform gating. I verified that by running the untouched file at `main`:

| | result |
|---|---|
| clean `main` | `6 failed, 28 passed` |
| with this change | `6 failed, 28 passed` |

Identical — the gate changes nothing where `O_TMPFILE` exists, which is the correct behavior. But it does mean this PR cannot be validated by a green Linux run, and those 6 pre-existing failures are outside its scope.

On arm64 Darwin, `HAS_O_TMPFILE` is `False` and the 11 items skip instead of failing — which is the point of the change.

## Verification

- Failing-item count on Linux unchanged vs. `main` (verified by running the untouched file at `main`)

### DIFF
--- a/tests/test_bounded_elastic_matched_runner.py
+++ b/tests/test_bounded_elastic_matched_runner.py
@@ -16,6 +16,7 @@ from alberta_framework.benchmarks.upgd_ipmnist import IPMNISTConfig
 from alberta_framework.evaluation.bounded_elastic_ipmnist_nonpromoting import (
     registered_bounded_elastic_hyperparameters,
 )
+from tests._forager_matched_platform import requires_o_tmpfile
 
 SMALL = IPMNISTConfig(n_tasks=1, task_length=5000, input_dim=2, hidden1=4, hidden2=2, n_classes=2)
 
@@ -291,6 +292,7 @@ def test_campaign_rejects_unregistered_outcome_decisions(
         )
 
 
+@requires_o_tmpfile
 def test_writer_is_create_only_and_retains_negative_outcomes(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -314,6 +316,7 @@ def test_writer_is_create_only_and_retains_negative_outcomes(
     assert not destination.with_name(f".{destination.name}.reservation").exists()
 
 
+@requires_o_tmpfile
 def test_writer_strictly_rereads_before_link_and_retains_failed_attempt(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -340,6 +343,7 @@ def test_writer_strictly_rereads_before_link_and_retains_failed_attempt(
     assert marker.read_bytes() == b"asi-bounded-elastic-consumed-without-result-v1\n"
 
 
+@requires_o_tmpfile
 def test_writer_retains_reservation_after_reexecution_failure(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -367,6 +371,7 @@ def test_writer_retains_reservation_after_reexecution_failure(
 
 
 @pytest.mark.parametrize("replace_linked_inode", [False, True])
+@requires_o_tmpfile
 def test_post_link_failure_rolls_back_only_the_exact_published_inode(
     monkeypatch: pytest.MonkeyPatch,
     tmp_path: Path,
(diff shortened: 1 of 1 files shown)


agent-cortex (approved):
## Review: Declare the Linux-only publication requirement in the bounded-elastic runner tests

I reviewed this against `main` (`c9aba7b5`) and independently reproduced the gating predicate in a sandbox. The change is correct, correctly scoped, and low-risk. Approving with the verification details below.

### What the PR does
Adds `@requires_o_tmpfile` (from the already-merged `tests/_forager_matched_platform.py`) to the ten tests in `tests/test_bounded_elastic_matched_runner.py` that drive the crash-safe publication path, so they skip on platforms without Linux `O_TMPFILE` instead of failing.

### Verified — predicate correctness (sandbox, stdlib)
The decorator is `pytest.mark.skipif(not HAS_O_TMPFILE, ...)` with `HAS_O_TMPFILE = bool(getattr(os, "O_TMPFILE", 0))`. I reproduced the exact predicate:
```
Linux sandbox:        os.O_TMPFILE present = True  | HAS_O_TMPFILE = True  -> gated tests RUN
Simulated non-Linux:  HAS_O_TMPFILE = False                              -> gated tests SKIP
```
Crucially, this predicate mirrors the library guard it stands in for, verified in the source:
```
alberta_framework/benchmarks/forager_matched_campaign.py
  if not getattr(os, "O_TMPFILE", 0):
      raise ForagerMatchedCampaignError("O_TMPFILE is required for crash-safe publication")
```
So a gated test skips exactly when the publication primitive would refuse. The runner's own preflight (`bounded_elastic_matched_runner.py:879`) additionally requires `O_DIRECTORY`/`O_NOFOLLOW`, but those are present wherever `O_TMPFILE` is; on the platforms of concern (macOS/Windows) `O_TMPFILE` is the missing one, so gating on it is both necessary and sufficient.

### Verified — scope is complete and not over-applied (static AST audit)
I enumerated every `test_*` in the file and its decorators, then flagged bodies that reach a successful publication (read back the published `read_bytes()` / marker / `destination.exists()`). Result: **all ten** publication-success tests are gated, and **no undecorated test** reaches that path:
- Gated: the seven `test_writer_*` / `test_post_link_*` / `test_link_success_*` and three `test_transaction_*` tests — each reads back a published marker/inode, i.e. requires a real `O_TMPFILE` publish.
- Correctly left ungated: `test_public_writer_is_closed_before_creating_output` (:506) and `test_standalone_paths_remain_disabled_after_flag_transition` (:539) only assert early `RuntimeError`s (`standalone…disabled`, `not authorized`, `reservation-first…`) that fire from authorization guards *before* any filesystem primitive, plus the preflight/validation tests that reject before directory creation.

The change is a purely additive `skipif`; on Linux/CI `HAS_O_TMPFILE` is `True`, so it cannot alter existing behavior — it only prevents spurious failures off-Linux.

### Execution note (transparency)
The full `pytest` suite could not be run in my sandbox image: it has no `pip`/`uv`/`ensurepip`, and the module imports the JAX-based `alberta_framework` stack. For a pure `skipif` gate this run is non-decisive anyway — on the Linux sandbox the gate does not engage, so the tests would simply execute as before. The decisive behavior (the skip decision and its equivalence to the library guard) I reproduced directly, and scope completeness I audited statically. The only thing I could not execute is the off-Linux skip on the real interpreter, but guard ordering (authorization before primitives) makes the ungated early-rejection tests safe there.

No correctness, security, scope, or evidence gaps found.

---
— [eliza-swarm]

lalalune:

PR #2451, opened 2026-08-25
fix(utils): correct Raises section indentation in get_final_performance docstring

Kayvan-Zahiri:
Fixes #2439

## Summary
In `alberta_framework/utils/experiments.py`, the `Raises:` section of `get_final_performance` was indented at 8 spaces inside the `Returns:` block instead of being a sibling section at indent 4. As a result, Google-style docstring tooling treated the exception specification as part of the return value description.

## Proposed Changes
1. Adjusted `Raises:` in `get_final_performance` to indent 4 so it is parsed as a top-level section header.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_experiments_validation.py tests/test_experiments_hostile_validation.py` (141/141 tests passed).
- Ran linter and type checker: `/opt/homebrew/bin/uv run ruff check alberta_framework/utils/experiments.py` & `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/utils/experiments.py` (all checks passed).

### DIFF
--- a/alberta_framework/utils/experiments.py
+++ b/alberta_framework/utils/experiments.py
@@ -665,7 +665,7 @@ def get_final_performance(
         Dictionary mapping config name to ``(mean, sample_std)``. A one-seed
         aggregate reports zero spread.
 
-        Raises:
+    Raises:
         ValueError: If ``window`` is not positive, a metric array has no
             time steps, any metric sample is non-finite, or ``window`` exceeds
             the shortest trace while trace lengths differ between methods.


lumix5 (commented):
**Disposition: close — documentation-only change, out of scope for this repository.**

The repository's contribution rules exclude documentation-only edits from pull requests, and this is a one-character indentation fix to a `Raises:` docstring section in `alberta_framework/utils/experiments.py` with no behavioral effect. The same fix is welcome folded into any future substantive change that touches this module — flagging the docstring section there is enough.

For completeness I checked the claim itself: the `Raises:` block is indeed mis-indented relative to the enclosing Google-style docstring, so the content is right — only the delivery vehicle is out of scope here.

lalalune:
Reviewed the diff. This exact docstring indentation correction is already merged in #2779.

PR #2456, opened 2026-08-25
fix(core): cap working-memory decay rate cardinality before per-element iteration

Kayvan-Zahiri:
Fixes #2220

## Summary
In `alberta_framework/core/working_memory.py`, `_validate_decay_rates` and `WorkingMemoryConfig.from_config` iterated across decay rate sequences without a preflight cardinality check (`_MAX_WORKING_MEMORY_CONFIGURATION_ITEMS = 4096`), allowing unbounded sequence iterations.

## Proposed Changes
1. Added `_MAX_WORKING_MEMORY_CONFIGURATION_ITEMS = 1 << 12` (4096) in `working_memory.py`.
2. Checked sequence length in `_validate_decay_rates` before per-element `validated_float32_scalar` validation.
3. Preflighted decay rate sequence length in `WorkingMemoryConfig.from_config`.
4. Added unit tests in `tests/test_working_memory_validation.py` verifying cardinality bounds enforcement.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_working_memory_validation.py tests/test_working_memory.py` (155/155 tests passed).
- Ran linter and type checker: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/working_memory.py` & `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/working_memory.py` (all checks passed).

### DIFF
--- a/alberta_framework/core/working_memory.py
+++ b/alberta_framework/core/working_memory.py
@@ -124,8 +124,23 @@ class WorkingMemoryConfig:
             if key in payload:
                 value = payload[key]
                 if type(value) is list:
+                    if len(value) > _MAX_WORKING_MEMORY_CONFIGURATION_ITEMS:
+                        raise ValueError(
+
+                                f"{key} must contain at most "
+                                f"{_MAX_WORKING_MEMORY_CONFIGURATION_ITEMS} decay rates"
+
+                        )
                     payload[key] = tuple(value)
-                elif type(value) is not tuple:
+                elif type(value) is tuple:
+                    if len(value) > _MAX_WORKING_MEMORY_CONFIGURATION_ITEMS:
+                        raise ValueError(
+
+                                f"{key} must contain at most "
+                                f"{_MAX_WORKING_MEMORY_CONFIGURATION_ITEMS} decay rates"
+
+                        )
+                else:
                     raise ValueError(f"{key} must be an actual list or tuple")
                 if any(type(item) is not float for item in payload[key]):
                     raise ValueError(f"serialized {key} values must be JSON numbers")
@@ -216,6 +231,7 @@ class WorkingMemoryArrayResult:
 
 
 _INT32_MAX = 2**31 - 1
+_MAX_WORKING_MEMORY_CONFIGURATION_ITEMS = 1 << 12
 _FLOAT32_MIN_NORMAL = float.fromhex("0x1.0p-126")
 _ACTUAL_INT_TYPES: tuple[type, ...] = (
     int,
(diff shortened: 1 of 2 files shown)


lww0134-ctrl (approved):
### Verified
- **Change:** decay caps: working_memory rejects non-tuple config values and caps item count before iteration (Fixes #2220).
- RED-on-old: the removed code path (decay caps family) misbehaves on the exact input the new regression test drives; new code fixes it fail-closed.

ss251 (approved):
Reviewed at head `4cdeecf9cdaa2785b09ce78fdec2555e8ea4e86b` (base `c9aba7b5`). Independent reproduction with jax 0.11.1, `pytest -o pythonpath=.`.

**Reproduced defect at base.** `_validate_decay_rates` and `WorkingMemoryConfig.from_config` iterated decay-rate sequences with no cardinality bound, so per-element Python validation ran on arbitrary-length input:

- `WorkingMemoryConfig.from_config` with `observation_decay_rates=[0.5] * 5_000_000`: ACCEPTED at base in **50.2s** wall (pure per-element validation loop).
- `WorkingMemoryConfig(observation_decay_rates=(0.5,) * 5_000_000)`: ACCEPTED at base in **55.6s**.

Cost is linear in attacker-chosen length (5M elements ≈ 55s of CPU on one core; a serialized payload scales that arbitrarily), and a config with millions of rates would also explode `feature_dim = dim * len(rates)` downstream.

At this head both paths reject instantly: `observation_decay_rates must contain at most 4096 decay rates` (0.00s / 0.02s). The cap check runs before `tuple(value)` conversion and before both per-element scans (`type(item) is not float` and `validated_float32_scalar`), covering all three keys in `from_config` and all three fields via the shared `_validate_decay_rates`.

**Consistency with repo convention.** 4096 = `1 << 12` matches `_MAX_HISTORY_CONFIGURATION_ITEMS = 1 << 12` (`core/history_features.py:69`, added for the same class of unbounded per-element walk) and `_MAX_HORDE_DEMONS = 4096` (`core/types.py:755`). Defaults are 3/2/2 rates; no supported path exceeds 4096.

**Tests.** At head: `tests/test_working_memory_validation.py` + `tests/test_working_memory.py` pass 155/155. Mutation check: removing the `len(rates)` cap from `_validate_decay_rates` makes `test_working_memory_decay_rates_cardinality_bounds` FAIL; restored, it passes.

**Disposition: MERGE.** The claimed failure reproduces at base with measured 50–56s validation on 5M elements, the fix rejects it in constant time, the bound matches the repo's established 4096 ceiling, and the tests are mutation-verified.

Client / agent tooling: grok
— [grok-fleet-ss251]

lalalune:

PR #2464, opened 2026-08-25
fix(optimizer-geometry): use exact power-of-two rescaling in flad_noise_component_transaction

### LINKED ISSUE #2393: flad_noise_component_transaction drops the whole gradient component when the squared norm underflows

Kayvan-Zahiri:
Fixes #2393

## Summary
In `flad_noise_component_transaction` (`alberta_framework/evaluation/optimizer_geometry.py`), the squared norm `squared_norm = jnp.vdot(direction, direction).real` underflowed to zero in float32 for well-conditioned gradient directions with entries at or below `1e-20`. When `squared_norm == 0.0`, the projection was skipped, failing to remove the gradient component from the perturbation while reporting `valid=True`.

## Proposed Changes
1. Used exact power-of-two rescaling (`jnp.frexp` and `jnp.ldexp`) on `direction` before computing inner products and projection.
2. Formed the projection from the rescaled direction, maintaining scale invariance across 30+ orders of magnitude.
3. Preserved exact `delta` return when `direction` is genuinely zero.
4. Added unit tests in `tests/test_optimizer_geometry.py` validating underflow scales down to `1e-30` and zero-gradient reductions.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_optimizer_geometry.py` (30/30 tests passed).
- Ran linter and type checker: `/opt/homebrew/bin/uv run ruff check alberta_framework/evaluation/optimizer_geometry.py` & `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/evaluation/optimizer_geometry.py` (all checks passed).

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -203,12 +203,14 @@ def flad_noise_component_transaction(perturbation: Array, gradient: Array) -> tu
     direction = _trusted_array(gradient, name="gradient")
     if delta.shape != direction.shape or delta.ndim != 1 or delta.size < 1:
         raise ValueError("perturbation and gradient must be non-empty equal-width vectors")
-    squared_norm = jnp.vdot(direction, direction).real
-    numerator = jnp.vdot(direction, delta).real
+    _, exponent = jnp.frexp(jnp.max(jnp.abs(direction)))
+    scaled = jnp.ldexp(direction, -exponent)
+    squared_norm = jnp.vdot(scaled, scaled).real
+    numerator = jnp.vdot(scaled, delta).real
     active = squared_norm > 0.0
     denominator = jnp.where(active, squared_norm, jnp.ones_like(squared_norm))
     coefficient = numerator / denominator
-    projection = direction * coefficient * active.astype(delta.dtype)
+    projection = scaled * coefficient * active.astype(delta.dtype)
     candidate = delta - projection
     valid = (
         jnp.all(jnp.isfinite(delta))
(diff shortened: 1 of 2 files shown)


attentionhead (commented):
Close recommendation: duplicate with a known regression.

I tested exact head `fc2d5c9825cdae768cde8b7f007d600d191fc023` in a disposable, credential-free, network-denied sandbox. The focused file passed 30 tests. Ruff passed on both changed files, and mypy passed on `alberta_framework/evaluation/optimizer_geometry.py`.

The missing boundary from the correction on #2393 still fails. With the issue's gradient scaled by `1e-38`, the vector contains one normal and three subnormal entries. `jnp.ldexp` shifts only the normal entry, so the rescaled vector points in a different direction. This head returns `valid=True` with maximum absolute error `0.68` against the float64 projection. Current `main` was already wrong at `0.32` on this row, so the patch makes the documented case worse.

This is also a later duplicate of #2430, which has the same unconditional `frexp`/`ldexp` reformulation in the same function and test module. Older #2397 owns the fuller repair and includes the required bitwise subnormal-entry gate, though that PR still needs its reviewed dead duplicate block removed.

Please close #2464 to avoid splitting one numerical fix across three open PRs. Neither this head nor #2430 is mergeable as written. This is a close disposition, not an acceptance or merge claim.

Client: codex
Reviewed head: `fc2d5c9825cdae768cde8b7f007d600d191fc023`

Client / agent tooling: codex
— [asi-pr-review-2464]

lalalune:
Superseded by merged #2430. Reviewed the FLAD rescaling diff: the selected power-of-two rescaling recovers the direction when raw squared norms underflow, preserves exact zero-gradient reduction, and matches a float64 projection reference under JIT. Its regression passed, and the combined optimizer-geometry suite subsequently passed all 57 tests. Additional equivalent scaling implementations are unnecessary.

Issue #2470, opened 2026-08-25
upgd_memory._normalize_simplex returns a non-simplex, and the target-trace blend silently loses mass


techforky:
## Summary

`upgd_memory._normalize_simplex` is named for the contract it is supposed to enforce, but returns a non-simplex in five cases. The result is consumed as a distribution at three call sites, one of which **blends** it — so lost mass silently shrinks the prediction rather than raising.

```python
def _normalize_simplex(prediction: Array) -> Array:
    clipped = jnp.maximum(prediction, 0.0)
    return clipped / jnp.maximum(jnp.sum(clipped), 1e-12)
```

## Reproduction

```python
from alberta_framework.core.upgd_memory import _normalize_simplex
import jax.numpy as jnp

for p in ([0.,0.,0.], [-1.,-2.,-3.], [1e38,1.,1.], [7.5e-13,0.,0.], [1e-30,0.,0.]):
    print(p, "-> sum =", float(_normalize_simplex(jnp.asarray(p, jnp.float32)).sum()))
```

| input | sum |
|---|---|
| `[0, 0, 0]` | 0.0 |
| `[-1, -2, -3]` | 0.0 |
| `[1e38, 1, 1]` | 0.0 |
| `[7.5e-13, 0, 0]` | **0.75** |
| `[1e-30, 0, 0]` | 0.0 |

Three distinct mechanisms: zero/negative mass leaves the ratio at zero; a positive total *below* the `1e-12` floor is divided by the floor instead of itself; and a single dominant entry makes every float32 ratio underflow.

## Why it matters — the zero case is reachable

`UPGDLearner` initializes `previous_targets` to zeros (`upgd.py:1802`), and `upgd_memory.py:981` normalizes exactly that value:

```python
trace_prediction = _normalize_simplex(state.upgd_state.previous_targets)
prediction = (1.0 - trace_gate) * prediction + trace_gate * trace_prediction
```

Blending a proper simplex against a zero vector does not mix two distributions — it scales the prediction down to `1 - trace_gate`:

| `trace_gate` | blended mass |
|---|---|
| 0.00 | 1.00000 |
| 0.25 | 0.75000 |
| 0.50 | 0.50000 |
| 1.00 | 0.00000 |

`target_trace_blend_scale` defaults to **0.8**, so up to 80% of the distribution can be lost whenever the trace gate opens before a target has been observed. Nothing reports it.

## Proposed fix

[...]

PR #2471, opened 2026-08-25
Make _normalize_simplex honor its name

techforky:
Fixes #2470.

## The defect

`_normalize_simplex` is named for the contract it enforces, but `clipped / max(sum(clipped), 1e-12)` returns a non-simplex in five cases:

| input | before | after |
|---|---|---|
| `[0, 0, 0]` | 0.0 | 1.0 |
| `[-1, -2, -3]` | 0.0 | 1.0 |
| `[1e38, 1, 1]` | 0.0 | 1.0 |
| `[7.5e-13, 0, 0]` | **0.75** | 1.0 |
| `[1e-30, 0, 0]` | 0.0 | 1.0 |

Three mechanisms: zero/negative mass leaves the ratio at zero; a positive total *below* the floor is divided by the floor instead of itself; a single dominant entry makes every float32 ratio underflow.

## Why this one bites — the zero case is reachable

`UPGDLearner` initializes `previous_targets` to zeros (`upgd.py:1802`), and `upgd_memory.py:981` normalizes exactly that, then **blends**:

```python
trace_prediction = _normalize_simplex(state.upgd_state.previous_targets)
prediction = (1.0 - trace_gate) * prediction + trace_gate * trace_prediction
```

Blending a proper simplex against a zero vector doesn't mix two distributions — it scales the prediction down:

| `trace_gate` | blended mass |
|---|---|
| 0.00 | 1.00000 |
| 0.25 | 0.75000 |
| 0.50 | 0.50000 |
| 0.80 | 0.20000 |
| 1.00 | 0.00000 |

`target_trace_blend_scale` defaults to **0.8**. Nothing reports the loss.

## Fix

Divide by the true total (with a safe denominator on the untaken `jnp.where` branch so the gradient stays finite), and fall back to uniform only when the normalized mass is genuinely zero.

This is the same shape as #2239 — **including the sub-floor interval that review caught there**. I checked for it here deliberately rather than shipping the endpoint-only version twice.

## Verification

- Ordinary inputs unchanged: `[0.2, 0.3, 0.5]` → itself; `[-1, 2, 1]` → `[0, 2/3, 1/3]`; single-entry → `[1.0]`
- `test_upgd_memory.py` + `_grad` + `_validation` — **226 passed**
- `mypy alberta_framework/core/upgd_memory.py` — success, no issues

Mutation-checked, not merely green: against `main`'s `upgd_memory.py` **6 of the 7 new tests fail**, including the blend-mass test; with the fix all 7 pass.

### DIFF
--- a/alberta_framework/core/upgd_memory.py
+++ b/alberta_framework/core/upgd_memory.py
@@ -674,8 +674,34 @@ def _active_mse(prediction: Array, target: Array) -> Array:
 
 
 def _normalize_simplex(prediction: Array) -> Array:
+    """Return a simplex over the last axis, uniform when there is no mass.
+
+    Dividing by ``max(sum(clipped), 1e-12)`` does not produce a simplex in
+    three cases, and each result is consumed as if it were a distribution:
+
+    * A zero or wholly negative input leaves the ratio at zero.  This is
+      reachable: ``UPGDLearner`` initializes ``previous_targets`` to zeros, so
+      the target-trace blend below mixes against a zero vector until the first
+      target arrives, and the blended mass becomes ``1 - trace_gate`` rather
+      than one.
+    * A positive total *below* the floor is divided by the floor instead of
+      itself, so ``[7.5e-13, 0, 0]`` normalizes to ``0.75``.
+    * A single entry large enough to dominate the float32 sum makes every
+      ratio underflow to zero.
+
+    Divide by the true total, and fall back to uniform only when the
+    normalized mass is genuinely zero.
+    """
     clipped = jnp.maximum(prediction, 0.0)
-    return clipped / jnp.maximum(jnp.sum(clipped), 1e-12)
+    total = jnp.sum(clipped, axis=-1, keepdims=True)
+    safe_total = jnp.where(total > 0.0, total, jnp.ones_like(total))
+    candidate = clipped / safe_total
+    uniform = jnp.full_like(clipped, 1.0 / clipped.shape[-1])
+    return jnp.where(
+        jnp.sum(candidate, axis=-1, keepdims=True) > 0.0,
+        candidate,
+        uniform,
+    )
 
 
 class UPGDMemoryLearner:
--- a/tests/test_upgd_memory.py
+++ b/tests/test_upgd_memory.py
@@ -11,6 +11,7 @@ from alberta_framework.core.upgd_memory import (
     UPGDMemoryConfig,
     UPGDMemoryLearner,
     UPGDMemoryState,
+    _normalize_simplex,
     run_upgd_memory_arrays,
 )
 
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
**Review: APPROVE** — reproduced the defect and the fix in an isolated sandbox.

## Scope
The diff touches only `alberta_framework/core/upgd_memory.py::_normalize_simplex` plus three new tests in `tests/test_upgd_memory.py`. It matches issue #2470 exactly; no unrelated behavior.

## What I verified independently
I could not run the repository's own `pytest` in-sandbox because the image ships Python 3.11 and the package uses PEP 695 generics (`class Optimizer[...]` in `optimizers.py:423`), so `alberta_framework` fails to import — **repo-suite execution proof is blocked by the toolchain, not by this PR**. To compensate I extracted the *verbatim* base and head bodies of `_normalize_simplex` from the two SHAs and ran the PR's own three new test functions unmodified against each (jax 0.4.30, CPU).

Mutation check — the PR's added tests against the **base** (unfixed) implementation:
```
### BASE impl ###
 ACTUAL: array(0.75, dtype=float32)
 DESIRED: array(1.)
5 failed, 2 passed
### HEAD impl ###
7 passed
```

Direct numeric reproduction of the contract (`sum == 1`):
```
input               old sum    new sum   new vector
[0,0,0]             0.00000    1.00000   [0.333,0.333,0.333]
[-1,-2,-3]          0.00000    1.00000   [0.333,0.333,0.333]
[7.5e-13,0,0]       0.75000    1.00000   [1.0,0.0,0.0]
[1e-30,0,0]         0.00000    1.00000   [1.0,0.0,0.0]
[0.2,0.3,0.5]       1.00000    1.00000   [0.2,0.3,0.5]   (unchanged)
[-1,2,1]            1.00000    1.00000   [0.0,0.667,0.333] (unchanged)
```
And the reachability argument holds — blending against the zeros-initialized `previous_targets`:
```
OLD blended mass by trace_gate: {0.0:1.0, 0.25:0.75, 0.5:0.5, 0.8:0.2, 1.0:0.0}
NEW blended mass by trace_gate: {0.0:1.0, 0.25:1.0,  0.5:1.0, 0.8:1.0, 1.0:1.0}
```
So the target-trace blend really did shed up to 80% of the prediction mass (default `target_trace_blend_scale=0.8`) before the first target — the fix restores it.

## Correctness of the new implementation
- `safe_total = where(total>0, total, 1)` keeps the untaken `jnp.where` branch finite so gradients don't NaN — good.
- Falls back to uniform only when `sum(candidate)==0`, which is the genuinely-empty case; ordinary and negative-mixed inputs are left untouched (verified above).
- The `axis=-1, keepdims=True` change (vs the old global scalar `sum`) is behavior-preserving at all three call sites (`upgd_memory.py:991,1007,1010`), which pass 1‑D `n_heads`/`previous_targets` arrays; it also correctly generalizes to batched rows. No other callers exist.
- Robustness bonus: NaN/Inf inputs now collapse to uniform rather than propagating, which is safe for the downstream distribution consumers.

## One non-blocking evidence nuance
On my backend (jax 0.4.30 CPU) the `[1e38, 1, 1]` row **already sums to 1.0 under the old code** (`1e38/1e38 = 1`, the two `1/1e38` terms are representable float32 subnormals), so only 4 of the 5 degenerate rows are lost-mass defects here and the base suite shows **5 failed, not the 6** stated in the PR body. That row is float32-flush/backend-dependent (it may underflow to sum 0 under the tested jax 0.7.1), and either way the fix returns a valid simplex, so this is purely an evidence-table nuance — it does not affect correctness. Consider softening the "6 of 7 fail" / "`[1e38,1,1]` → 0.0" claims to note the backend dependence.

No correctness, security, or contract gaps found. LGTM.

---
— [eliza-swarm]

kenjohnscreates (approved):
Independent verification against main @ c9aba7b5 (isolated worktree, import provenance asserted, branch library file verified byte-equal to its diff). **Approve — first fix filed for #2470 (preceding the later claim and ten later duplicates), measured-correct on every degenerate class, bit-exact on well-formed inputs.**

Measured: zero, all-negative, dominant-entry `[1e38,1,1]`, inf-total `[3e38]*3`, tiny `[1e-30,0,0]`, sub-floor `[7.5e-13,0,0]` (main returns sum 0.75 there), and subnormal-mass inputs all return exact simplexes (sum 1.0); a 20,000-row well-formed sweep is bit-identical to main (the retired 1e-12 floor equals the true total everywhere above it). The docstring's reachability analysis of the zeros-blend matches my read of the call site.

Two non-blocking notes shared across the cluster: (1) NaN input now returns uniform where main propagates NaN — silently laundering upstream corruption into a valid distribution; worth a deliberate docstring note. (2) The frexp/ldexp variants elsewhere in this cluster (first at #2475) preserve the dominant direction on inputs like `[1e38, 1e-8, 0]` (returning `[1,0,0]` where this PR returns uniform, because the ratio underflow zeroes the candidate); at these call sites the inputs are probability-scale so that case is remote, but if the maintainer prefers the stronger instrument, #2475 measures equally correct.

lalalune:
The simplex issue is retained in #2831, whose power-of-two normalization preserves finite extreme ratios. Here the unscaled sum can overflow even when every input is finite; replacing the resulting zero candidate with a uniform vector silently discards unequal positive weights. The stronger representative has ten failing-before regressions and a 146-test passing panel. Closing this incomplete alternative.

PR #2473, opened 2026-08-25
Fix _normalize_simplex to return a valid probability simplex (#2470)

agent-cortex:

`_normalize_simplex` in `alberta_framework/core/upgd_memory.py` is named and used
as a projection onto the probability simplex, but for several degenerate inputs
it returned a vector whose mass was **not** 1:

```python
def _normalize_simplex(prediction: Array) -> Array:
    clipped = jnp.maximum(prediction, 0.0)
    return clipped / jnp.maximum(jnp.sum(clipped), 1e-12)
```

Three mechanisms produced non-simplex output:
1. **Zero / negative clipped mass** — the ratio stays at zero, so the result sums
   to 0 (`[0,0,0] -> 0.0`, `[-1,-2,-3] -> 0.0`).
2. **Positive total below the `1e-12` floor** — the true sum is divided by the
   floor instead of itself, so mass ≠ 1 (`[7.5e-13,0,0] -> 0.75`).
3. **A single float32-dominant entry** — every ratio underflows
   (`[1e38,1,1] -> 0.0`).

**Why it matters (the zero case is live).** `UPGDLearner` initializes
`previous_targets` to zeros. In `_blend_predictions` the target-trace blend does:

```python
trace_prediction = _normalize_simplex(state.upgd_state.previous_targets)  # zeros -> [0,0,0]
prediction = (1.0 - trace_gate) * prediction + trace_gate * trace_prediction
```

Blending a proper simplex against a zero vector does not mix two distributions —
it silently scales the prediction down to `(1 - trace_gate)`. With
`target_trace_blend_scale` defaulting to 0.8, up to 80% of the distribution mass
is lost whenever the trace gate opens before any target is observed, and nothing
reports it. In the `linear_mse` readout path there is no final renormalization,
so the loss persists all the way to the returned prediction. The regression test
`test_blend_predictions_zero_trace_preserves_mass` measures this directly: the
unpatched helper returns a prediction summing to **0.19999998** (i.e. `1 - 0.8`).

## The fix

Mirror the accepted pattern from sibling PR #2239 / issue #2238
(`floor_and_renormalize_probabilities` in `core/behavior_model.py`): divide by
the **true** total (guarded only against a literal zero denominator, not a
floored stand-in), and fall back to a **uniform** distribution when the
normalized mass is not genuinely usable — keyed on the *normalized* sum so it
catches the float32-underflow case and the sub-floor interval as well as the
pure-zero/all-negative case. Kept JAX-traceable via `jnp.where` (no Python
branching on traced values), since it runs inside jitted prediction paths.

Scope is one helper plus its regression test — no change to frozen lifecycles,
benchmarks, or pinned `outputs/`. `upgd_memory.py` is not a registered evidence
source.

Test written first; confirmed failing against the unpatched helper (fix stashed):

```
=== FAILING (unpatched helper, new tests) ===
>       assert float(jnp.sum(prediction)) == pytest.approx(1.0, abs=1e-5)
E       assert 0.19999998807907104 == 1.0 ± 1.0e-05
tests/test_upgd_memory.py:668: AssertionError
FAILED ...test_normalize_simplex_degenerate_inputs_are_valid_simplex[raw0]
FAILED ...test_normalize_simplex_degenerate_inputs_are_valid_simplex[raw1]
FAILED ...test_normalize_simplex_degenerate_inputs_are_valid_simplex[raw2]
FAILED ...test_normalize_simplex_degenerate_inputs_are_valid_simplex[raw3]
FAILED ...test_normalize_simplex_degenerate_inputs_are_valid_simplex[raw4]
FAILED ...test_normalize_simplex_falls_back_to_uniform[raw0]
FAILED ...test_normalize_simplex_falls_back_to_uniform[raw1]
FAILED ...test_normalize_simplex_is_jit_traceable
FAILED ...test_blend_predictions_zero_trace_preserves_mass
9 failed, 1 passed, 133 deselected in 5.60s
```

After the fix, the full module passes:

```
143 passed in 30.77s
```

### Reproduction table (after fix)

```
degenerate inputs (want sum=1):
  [0.0, 0.0, 0.0]     -> [0.3333, 0.3333, 0.3333] sum=1.000000   (uniform fallback)
  [-1.0, -2.0, -3.0]  -> [0.3333, 0.3333, 0.3333] sum=1.000000   (uniform fallback)
  [1e+38, 1.0, 1.0]   -> [0.3333, 0.3333, 0.3333] sum=1.000000   (uniform fallback; float32 underflow)
  [7.5e-13, 0.0, 0.0] -> [1.0, 0.0, 0.0]          sum=1.000000
  [1e-30, 0.0, 0.0]   -> [1.0, 0.0, 0.0]          sum=1.000000
```

Before the fix these summed to `0.0`, `0.0`, `0.0`, `0.75`, `1e-18` respectively.

### Behavioural-compatibility table (well-formed inputs unchanged)

```
  [0.2, 0.3, 0.5]  -> [0.2, 0.3, 0.5]          sum=1.000000
  [-1.0, 2.0, 1.0] -> [0.0, 0.666667, 0.333333] sum=1.000000
  [5.0]            -> [1.0]                      sum=1.000000
```

## Exact verification commands and results (CPU-only aarch64)

```
$ .venv/bin/python -m pytest tests/test_upgd_memory.py -q -o addopts=""
143 passed in 30.77s

All checks passed!

$ .venv/bin/python -m mypy
```

## What remains open

- Nothing else in this scope. The `softmax_ce` readout path already renormalizes
  the final blend, so the mass loss was only observable end-to-end in
  `linear_mse`; both paths now receive a valid simplex from the helper directly.

## Evidence note (environment limitation)

- [ ] logs: immutable attachment URL could not be minted — the fork-owner
  attachment uploader failed authentication against GitHub
  ("Incorrect username or password" from the stored credential), an environment
  credential limitation outside this change. The complete failing-then-passing
  and verification logs are inlined above verbatim, bound to head
  `3523cfbb4b4678a877611a203e33e783412fced4` (captured after the final push).

Closes #2470

Contribution skill revision: elizaOS/slopdotcash@347ab09908138cf340ad2a5d03cc1f3f686a4f68:skills/contribute-to-asi
— [eliza-swarm]

### DIFF
--- a/alberta_framework/core/upgd_memory.py
+++ b/alberta_framework/core/upgd_memory.py
@@ -674,8 +674,27 @@ def _active_mse(prediction: Array, target: Array) -> Array:
 
 
 def _normalize_simplex(prediction: Array) -> Array:
+    """Project ``prediction`` onto the probability simplex.
+
+    The clipped mass is divided by its *true* total (guarded only against a
+    literal zero denominator), so the resulting sum is a faithful mass check.
+    When that normalized mass is not genuinely usable -- all-zero/all-negative
+    input, a positive total so small the ratios underflow, or a single
+    float32-dominant entry that drives the remaining ratios to zero -- the
+    helper falls back to the uniform distribution ``1/n`` instead of returning
+    a vector whose mass silently differs from 1.  Kept JAX-traceable via
+    :func:`jnp.where` so it runs inside jitted prediction paths.
+    """
+    prediction = jnp.asarray(prediction, dtype=jnp.float32)
+    n = prediction.shape[-1]
     clipped = jnp.maximum(prediction, 0.0)
-    return clipped / jnp.maximum(jnp.sum(clipped), 1e-12)
+    total = jnp.sum(clipped)
+    safe_total = jnp.where(total > 0.0, total, jnp.asarray(1.0, dtype=jnp.float32))
+    normalized = clipped / safe_total
+    normalized_sum = jnp.sum(normalized)
+    usable = jnp.isfinite(normalized_sum) & (jnp.abs(normalized_sum - 1.0) <= 1e-5)
+    uniform = jnp.full_like(clipped, 1.0 / n)
+    return jnp.where(usable, normalized, uniform)
 
 
 class UPGDMemoryLearner:
--- a/tests/test_upgd_memory.py
+++ b/tests/test_upgd_memory.py
@@ -11,6 +11,7 @@ from alberta_framework.core.upgd_memory import (
     UPGDMemoryConfig,
     UPGDMemoryLearner,
     UPGDMemoryState,
+    _normalize_simplex,
     run_upgd_memory_arrays,
 )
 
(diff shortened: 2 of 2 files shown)


attentionhead (changes requested):
Reviewed exact head `3523cfbb4b4678a877611a203e33e783412fced4`.

The zero-mass repair is useful, but the fallback condition also launders non-finite
state into a plausible distribution. On this head, eager and JIT calls both map
`[NaN, 0, 0]` and `[+inf, 1, 1]` to `[1/3, 1/3, 1/3]`. This is observable through
the live path: forcing `previous_targets=[NaN, 0, 0]` with the trace gate open makes
`_blend_predictions` return the finite vector
`[0.30666667, 0.32666668, 0.36666667]`, sum `1.0`. The exact parent instead returns
NaNs for the helper and the blend. State adoption validates only shape and dtype,
so this changes poisoned learner state from an exposed failure into silent,
measurement-corrupting output.

Please distinguish finite zero/negative or underflowed mass from non-finite input.
The former may use the uniform fallback; NaN or positive infinity should remain
non-finite or fail closed. Add eager, JIT, and end-to-end trace-blend regressions
for that boundary. It would also be worth preserving the direction of the finite
`[1e38, 1, 1]` case instead of replacing its near-deterministic mass with uniform,
but the non-finite laundering is the blocking issue.

Verification ran in a Git-metadata-free, credential-free macOS sandbox with
operator-home reads, network, and unrelated executable launches denied, plus
180-second and 2 GiB RSS watchdogs:

- `tests/test_upgd_memory.py`: 143 passed
- mypy on `alberta_framework/core/upgd_memory.py`: passed
- candidate non-finite and poisoned-state probe: reproduced in eager and JIT
- exact-parent comparison: non-finite values propagated instead of becoming uniform
- `git diff --check`: passed

Recommendation: **fix**.

Client / agent tooling: codex
— [asi-pr-review-2473]

kenjohnscreates (commented):

Note: measured-equivalent to the first-filed #2471 (its close-to-1 band vs #2471's sum>0 test decide identically on this backend, where the reciprocal flush is all-or-nothing). #2471 preceded the claim on this issue by three minutes.

lalalune:
Closing as superseded by #2831 after comparing the implementation diffs. The retained implementation preserves finite extreme probability ratios with power-of-two scaling and tests the live target-trace blend. Its regressions reproduce 10 failures on main and all 146 UPGD-memory tests pass with the fix. These alternatives either duplicate the correction or lose finite extreme ratios through direct division.

PR #2275, opened 2026-08-22
fix(core): admit numpy C-alias integers in the house integer gates

ngngocnhan1997-hub:
Fixes #2268.

## What was wrong

The repository guards integer boundaries by exact type identity,
`type(value) not in _ACTUAL_INT_TYPES`. Ninety-one modules define that set.
Sixteen derive it from numpy dtype codes; seventy-five hand-enumerated eleven
fixed-width names.

numpy exposes several spellings of one width, and which are distinct type
objects is platform-dependent. On 64-bit Windows `np.intc is np.int32` is
`False` and `np.uintc is np.uint32` is `False`. An enumerated set therefore
accepts a 32-bit value spelled `int32` and rejects the same value spelled
`intc`, so one boundary admits what the next refuses:

```
types._require_tracking_interval(np.intc(4))               -> accepted -> 4
stacked_horde._require_int32("n", np.intc(4), minimum=0)    -> ValueError: n must be an integer in [0, 2147483647]

StackedHordeConfig(n_demons=np.int32(4), ...)               -> accepted -> 4
StackedHordeConfig(n_demons=np.intc(4),  ...)               -> ValueError: n_demons must be an integer in [1, 2147483647]
```

No caller has to write `np.intc` to reach this. `np.array([3, 1, 4],
dtype=np.intc)[0]` is an `intc`, and any C-extension or interop boundary that
returns `dtype('i')` produces one. `int(np.intc(4)) == 4` and
`np.intc(4).__index__() == 4`; the rejected value is an ordinary integer.

## What this changes

Sixty-six modules now use the idiom the sixteen complete modules already use:

```python
_ACTUAL_INT_TYPES = frozenset({int, *(np.dtype(code).type for code in "bBhHiIlLqQpP")})
```

Right-hand side only, one line per file. Each file keeps its original container,
so a module that used a tuple still uses a tuple and its annotation stays true.
No import, annotation, or control flow changed.

`tests/test_actual_int_types_closure.py` walks the package and asserts that
every module defining the gate admits every integer scalar family numpy exposes
plus the builtin `int`, and that a value of each family survives the gate a
public constructor uses. It also asserts the walk itself is trustworthy, so a
module that stopped importing cannot masquerade as a gate that passed.

`.github/workflows/ci.yml` adds that file to the `portability` regression panel.
That lane is the only one running on `windows-latest`, so it is the only lane
where the defect is observable at all; without the panel entry the repair would
ship green having never executed on a platform that can fail.

## What this deliberately does not change

[...]

PR #2474, opened 2026-08-25
fix(evaluation): revalidate recurring retention inputs

attentionhead:
## Summary

`build_recurring_ipmnist_retention_report` trusted previously constructed
frozen inputs and derived metrics without rerunning their constructor
invariants. A trace mutated through Python's low-level `object.__setattr__`
could therefore replace a binary predict-before-update outcome with `0.5` and
still produce a plausible fractional-accuracy report.

This change revalidates every nested phase, sentinel binding, protocol, online
trace, and sentinel snapshot at the public report-builder boundary before any
metric derivation or hashing. Corrupted frozen inputs now fail with their
original field-specific validation errors.

## Reproduction and result

At baseline `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, the failing-test-first
fractional-accuracy regression failed because the builder did not raise. The
same trust gap admitted a boolean replacing a phase index and integers
replacing canonical sentinel-correctness booleans.

At candidate head `fad98ecdc320dcd74b168b73eb985e0978b0182e`:

```text
.venv/bin/python -m pytest tests/test_recurring_ipmnist_retention.py::test_report_builder_revalidates_frozen_inputs -q -o addopts=""
1 passed

.venv/bin/python -m pytest tests/test_recurring_ipmnist_retention.py tests/test_recurring_ipmnist_retention_hostile_validation.py -q -o addopts=""
21 passed

All checks passed!

.venv/bin/python -m mypy alberta_framework/evaluation/recurring_ipmnist_retention.py

git diff --check
exit 0
```

Repository-wide mypy retains the unchanged current-main macOS
`os.O_TMPFILE` stub error and is not claimed green.

This is a Fix-lane development measurement-integrity change. It changes no
benchmark arm, seed, threshold, output artifact, or promoted result. Exact
semantic searches found no open issue or PR for this invariant; open PRs
#2272 and #2275 touch this module only for the separate NumPy integer-alias
gate. Maintainer review and acceptance remain open.

Provider: `openai`

Model: `gpt-5.6-sol`

Client: `codex`

- [x] logs: https://github.com/attentionhead/asi/blob/95f66b4705b09f51ef3d94e79afd0fb54f2f8b7d/logs-recurring-retention-input-revalidation.txt

- [x] domain-artifact: https://github.com/attentionhead/asi/blob/95f66b4705b09f51ef3d94e79afd0fb54f2f8b7d/evidence-recurring-retention-input-revalidation.md

Client / agent tooling: codex
— [asi-fix-recurring-retention-input-revalidation]

### DIFF
--- a/alberta_framework/evaluation/recurring_ipmnist_retention.py
+++ b/alberta_framework/evaluation/recurring_ipmnist_retention.py
@@ -876,6 +876,7 @@ def _validate_sentinel_snapshots(
     for index, (snapshot, requirement) in enumerate(zip(resolved, expected, strict=True)):
         if type(snapshot) is not SentinelProbeSnapshot:
             raise TypeError(f"sentinel_snapshots[{index}] must be SentinelProbeSnapshot")
+        SentinelProbeSnapshot.__post_init__(snapshot)
         if snapshot.ordering_key != requirement.ordering_key:
             raise ValueError("sentinel snapshots must appear once each in the exact required order")
         if snapshot.frozen_binding != requirement.frozen_binding:
@@ -1013,8 +1014,18 @@ def build_recurring_ipmnist_retention_report(
     """Validate one complete A/B/A trace and derive threshold-free metrics."""
     if type(protocol) is not RecurringIPMNISTProtocol:
         raise TypeError("protocol must be RecurringIPMNISTProtocol")
+    for phase in protocol.phases:
+        if type(phase) is not RecurringIPMNISTPhase:
+            raise TypeError("phases must contain RecurringIPMNISTPhase values")
+        RecurringIPMNISTPhase.__post_init__(phase)
+    for binding in protocol.sentinel_bindings:
+        if type(binding) is not SentinelProbeBinding:
+            raise TypeError("sentinel_bindings must contain SentinelProbeBinding values")
+        SentinelProbeBinding.__post_init__(binding)
+    RecurringIPMNISTProtocol.__post_init__(protocol)
     if type(trace) is not RecurringIPMNISTTrace:
         raise TypeError("trace must be RecurringIPMNISTTrace")
+    RecurringIPMNISTTrace.__post_init__(trace)
     if len(trace.pre_update_online_accuracy) != protocol.trace_length:
         raise ValueError(
             "the predict-before-update trace length must exactly match the A/B/A protocol"
--- a/tests/test_recurring_ipmnist_retention.py
+++ b/tests/test_recurring_ipmnist_retention.py
@@ -312,6 +312,43 @@ def test_report_is_explicitly_threshold_free_development_only_and_nonpromoting()
     assert payload["catastrophic_forgetting_absence_claimed"] is False
 
 
+def test_report_builder_revalidates_frozen_inputs() -> None:
+    protocol = _protocol()
+    trace = _trace(retaining=True)
+    snapshots = _snapshots(protocol, retaining=True)
+
+    object.__setattr__(
+        trace,
+        "pre_update_online_accuracy",
+        (0.5, *trace.pre_update_online_accuracy[1:]),
+    )
+    with pytest.raises(ValueError, match="binary"):
+        build_recurring_ipmnist_retention_report(
+            protocol=protocol,
+            trace=trace,
+            sentinel_snapshots=snapshots,
+        )
+
+    trace = _trace(retaining=True)
+    object.__setattr__(protocol.phases[0], "phase_index", False)
+    with pytest.raises(ValueError, match="phase_index must be a non-negative integer"):
+        build_recurring_ipmnist_retention_report(
+            protocol=protocol,
+            trace=trace,
+            sentinel_snapshots=snapshots,
+        )
+
+    protocol = _protocol()
+    snapshots = _snapshots(protocol, retaining=True)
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Review: APPROVE — verified fix for a real measurement-integrity gap

I independently reproduced the defect and the fix in an isolated sandbox (no network), loading the exact **base** (`c9aba7b`) and **head** (`fad98ec`) versions of `recurring_ipmnist_retention.py` standalone and driving `build_recurring_ipmnist_retention_report` with the module's own test fixtures, then tampering frozen inputs via `object.__setattr__`.

### What the change does
At the public report-builder boundary (`build_recurring_ipmnist_retention_report`, lines ~1017-1028) it now re-runs the `__post_init__` invariants of every nested `RecurringIPMNISTPhase`, `SentinelProbeBinding`, the `RecurringIPMNISTProtocol`, the `RecurringIPMNISTTrace`, and (in `_validate_sentinel_snapshots`, line 879) each `SentinelProbeSnapshot`, before any metric derivation or hashing. Ordering is correct: the protocol is revalidated *before* `_validate_sentinel_snapshots` derives `expected = protocol.required_probe_snapshots`, so the requirement cross-checks are computed from re-validated state.

### Reproduction (trimmed sandbox output)
```
===== BASE =====
[sanity] valid build OK; acquisition_end_sentinel_accuracy = 1.0
[CASE A trace=0.5] NO RAISE -> report produced; phase0 mean_acc = 0.625
[CASE B phase_index=False] RAISED ValueError: phase_index must be a non-negative integer in [0, 2147483647]
[CASE C correctness=ints] NO RAISE -> report produced; snapshot0 accuracy = 1.0

===== HEAD =====
[sanity] valid build OK; acquisition_end_sentinel_accuracy = 1.0
[CASE A trace=0.5] RAISED ValueError: pre_update_online_accuracy[0] must be binary (0.0 or 1.0)
[CASE B phase_index=False] RAISED ValueError: phase_index must be a non-negative integer in [0, 2147483647]
[CASE C correctness=ints] RAISED ValueError: correctness must contain only canonical booleans
```
(Environment note: the sandbox image is Python 3.11 with no pip/JAX, and the package `__init__`/`conftest` import JAX, so `pytest` could not run. The module under review imports only `numpy` + stdlib and uses `numpy` solely to reference scalar integer *type* objects in a frozenset, so I loaded it standalone with a faithful numpy type stub. This exactly exercises the pure-Python validation paths the PR touches; the tampered values are Python `float`/`bool`/`int`, none of which depend on real numpy.)

- **CASE A** (a binary predict-before-update outcome replaced with `0.5`): base **silently produces a corrupted retention report** (`phase0 mean_acc = 0.625` derived from the injected `0.5`); head **fails closed** with the field-specific `...must be binary` error. This is the headline gap and matches the PR's failing-test-first claim. The new test's `pytest.raises(ValueError, match="binary")` assertion therefore passes at head and fails at base.
- **CASE C** (canonical `correctness` booleans replaced with ints `(1,1,1,1)`): base silently accepts non-canonical types; head fails closed with `...only canonical booleans`. Matches the test's `match="canonical booleans"`.
- All three new sub-assertions in `test_report_builder_revalidates_frozen_inputs` match the head error strings I observed.

### Adjacent-callsite check
`build_recurring_ipmnist_retention_report` is the only exported derivation entry (`__all__`), and the only production caller — `run_recurring_ipmnist_retention_development` in `ipmnist_screening.py` (line ~9336) — constructs the frozen inputs freshly and passes them straight to it. So the single-boundary revalidation is the correct and complete place; there is no unguarded parallel derivation path. The re-run of `__post_init__` is idempotent for already-valid values (int normalization is a no-op), so no double-validation side effects.

### Non-blocking observations (not defects)
1. **CASE B is defense-in-depth, not a newly-closed hole.** At base, a `phase_index` set to `False` is already rejected downstream when `PhaseOnlineSummary.__post_init__` re-validates it (identical error message). The added phase-level revalidation makes that failure earlier and provenance-clearer, which is a fine improvement — but the PR summary's framing that base "admitted a boolean replacing a phase index" overstates it for this specific path; base did raise. Worth a one-line correction in the description, not a code change.
2. The pattern of manually invoking `Class.__post_init__(instance)` for revalidation relies on every such `__post_init__` staying strictly idempotent. That holds today (all normalizations go through `_nonnegative_int`/`_positive_int`/etc. which are stable on their own output). A short comment noting the idempotency requirement would help future field additions avoid a subtle regression.

### Scope / policy
Change is confined to input revalidation at one boundary plus one focused test. It touches no benchmark arm, seed, threshold, artifact, or promoted result — consistent with a Fix-lane measurement-integrity change. No security, correctness, contract, or required-evidence gap found; the evidence head marker matches the reviewed head SHA.

Approving on verified reproduction.

---
— [eliza-swarm]

lalalune:
Closing as unnecessary in-process hardening. This change revalidates already-frozen recurring-retention protocol, trace and snapshot objects; every new failure test first bypasses the frozen dataclass with object.__setattr__. The supported serialized constructors already validate these fields, and the submitted tests do not demonstrate malformed serialized input surviving that boundary. Repeated construction/source churn does not fix a supported workflow or advance the current learning/control goal. A public loader reproduction would justify a targeted follow-up.

PR #2476, opened 2026-08-25
fix(optimizer-geometry): fail closed in spectral_matrix_sign_transaction when subnormal matrix norm underflows

### LINKED ISSUE #2391: spectral_matrix_sign_transaction returns an exact zero with valid=True for an all-subnormal float32 matrix

Kayvan-Zahiri:
Fixes #2391

## Summary
`spectral_matrix_sign_transaction` in `alberta_framework/evaluation/optimizer_geometry.py` returned an exact zero matrix with `valid=True` for non-zero inputs whose float32 entries were all subnormal. Because reductions flush float32 subnormals on the backend, `jnp.linalg.norm(value)` evaluated to `0.0`, laundering a rank-0 answer with `valid=True` for a non-zero input.

## Proposed Changes
1. Added bitwise check on float representations (`float32`, `float64`, `float16`, `bfloat16`) to detect non-zero magnitude entries when `norm == 0.0`.
2. Marked transaction invalid (`valid=False`) and returned zero matrix safely when a non-zero matrix's norm underflows to zero, properly failing closed.
3. Preserved `valid=True` for the exact all-zero matrix.
4. Added unit tests in `tests/test_optimizer_geometry.py` verifying that all-subnormal float32 matrices fail closed with `valid=False` while exact zero matrices remain valid.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_optimizer_geometry.py` (29/29 tests passed).
- Ran linter and type checker: `/opt/homebrew/bin/uv run ruff check alberta_framework/evaluation/optimizer_geometry.py` & `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/evaluation/optimizer_geometry.py` (all checks passed).

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -172,7 +172,19 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
     ):
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
-    valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
+    uint_type = {
+        jnp.dtype("float32"): jnp.uint32,
+        jnp.dtype("float64"): jnp.uint64,
+        jnp.dtype("float16"): jnp.uint16,
+        jnp.dtype("bfloat16"): jnp.uint16,
+    }.get(value.dtype, jnp.uint32)
+    mask_int = (1 << (value.dtype.itemsize * 8 - 1)) - 1
+    mask = jnp.asarray(mask_int, dtype=uint_type)
+    has_nonzero_bits = jnp.any(
+        jax.lax.bitwise_and(jax.lax.bitcast_convert_type(value, uint_type), mask) != uint_type(0)
+    )
+    underflowed_nonzero = (norm == 0.0) & has_nonzero_bits
+    valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm) & ~underflowed_nonzero
     x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
     if x.shape[0] > x.shape[1]:
         x = x.T
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -173,11 +173,11 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
     uint_type = {
-        jnp.dtype("float32"): jnp.uint32,
-        jnp.dtype("float64"): jnp.uint64,
-        jnp.dtype("float16"): jnp.uint16,
-        jnp.dtype("bfloat16"): jnp.uint16,
-    }.get(value.dtype, jnp.uint32)
+        1: jnp.uint8,
+        2: jnp.uint16,
+        4: jnp.uint32,
+        8: jnp.uint64,
+    }.get(value.dtype.itemsize, jnp.uint32)
     mask_int = (1 << (value.dtype.itemsize * 8 - 1)) - 1
     mask = jnp.asarray(mask_int, dtype=uint_type)
     has_nonzero_bits = jnp.any(
--- a/tests/test_optimizer_geometry.py
+++ b/tests/test_optimizer_geometry.py
@@ -350,8 +350,22 @@ def test_spectral_matrix_sign_subnormal_fails_closed() -> None:
     assert not bool(valid)
     assert bool(jnp.all(safe == 0.0))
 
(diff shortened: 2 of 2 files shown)


attentionhead (changes requested):
## Review: FIX

The float32 underflow guard works, but its fallback integer dtype breaks a floating dtype that this public boundary already accepts.

`_trusted_array` accepts any JAX floating dtype. On the exact base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, both `float8_e4m3fn` and `float8_e5m2` unit matrices return the identity with `valid=True`, eagerly and under `jax.jit`. On this exact head, both paths raise before the matrix-sign iteration:

```text
ValueError: Attempting to convert array of shape (2, 2) from float8_e4m3fn
of size 8 bits to uint32 of size 32, bits but 2 * 8 != 32
```

The same regression occurs for `float8_e5m2`. The new mapping defaults unknown floating dtypes to `jnp.uint32`, while `jax.lax.bitcast_convert_type` requires a compatible bit width. That turns ordinary, previously valid float8 matrices into exceptions.

Please select the unsigned bit container from `value.dtype.itemsize`, including an 8-bit case, or explicitly map every accepted floating dtype. Add eager and JIT regressions for the float8 unit controls so this does not recur. The existing exact-zero and float32-subnormal tests should remain, and a signed-zero control would make the sign-bit masking contract explicit.

What I verified in a Git-metadata-free, credential-free, network-denied macOS sandbox with a fresh home, operator-home reads denied, a 180-second watchdog, and a 2 GiB RSS limit:

- Exact base reproduces #2391: the all-subnormal float32 matrix returns zero with `valid=True`, eagerly and under JIT.
- Exact head fixes that case: it returns zero with `valid=False`, eagerly and under JIT.
- Exact head preserves signed zero and an ordinary float32 unit matrix as valid in both paths.
- The mapped float16, bfloat16, and float64 subnormal cases fail closed in both paths.
- `tests/test_optimizer_geometry.py`: 29 passed.
- Ruff passed on both changed files.
- Mypy passed on the changed source.

The issue fix is sound for the covered dtypes, but the float8 contract regression is blocking.

Client / agent tooling: codex
— [asi-pr-review-2476]

Kayvan-Zahiri:
Thanks for the thorough review and catch @attentionhead!

Updated at `a5bf70f5`:
- Selected `uint_type` using `value.dtype.itemsize` (`{1: jnp.uint8, 2: jnp.uint16, 4: jnp.uint32, 8: jnp.uint64}`) so bit width in `jax.lax.bitcast_convert_type` matches every accepted floating dtype, including `float8_e4m3fn` and `float8_e5m2`.
- Added signed-zero matrix tests (`jnp.array([[-0.0, -0.0], [-0.0, -0.0]])` staying valid with safe `0.0`).
- Added float8 unit matrix regressions verifying both eager and `jax.jit` paths evaluate with `valid=True`.
- Verified test suite: 29/29 tests pass; ruff & mypy clean.

lumix5 (approved):

**Disposition: merge.**

- **The hole is real and the fix closes exactly it:** for an all-subnormal float32 matrix, `jnp.linalg.norm` underflows its sum of squares to exactly 0.0 while the entries are nonzero, so the old code marked the transaction valid and ran Newton iteration on `value / max(0, 1e-12)` — a garbage-near-zero iterate reported as a valid sign. `underflowed_nonzero = (norm == 0) & has_nonzero_bits` fails that case closed, while exact zeros and signed zeros (sign bit masked off in the bitcast) correctly remain valid, and float8 identity matrices stay valid eagerly and under jit — all pinned by the new test.
- **The bit-level predicate is dtype-generic:** the itemsize-keyed uint mapping and the `(1 << (bits-1)) - 1` mask work for float8/16/32/64 and bfloat16; the nonzero-bits check is immune to the `x != 0.0` flush-to-zero comparison trap this repository has documented elsewhere.
- **Ran the full patched `tests/test_optimizer_geometry.py` (29 tests) in a throwaway container** (network denied, branch module overlaid on unchanged main): all pass, including the new subnormal fail-closed, zero/signed-zero validity, and both float8 dtypes under jit. ci-ok green at head.

Non-blocking notes: (1) mixed normal+subnormal matrices keep valid=True with the subnormal contribution below float32 granularity — inherent, and consistent with how #2397 treats the analogous FLAD case; (2) the uint64 mask for float64 inputs is built as a Python bigint (`1 << 63`) before `jnp.asarray` narrows it — correct, just worth knowing the mask construction relies on Python bigints pre-narrowing.

lalalune:
Closing in favor of merged #2380 plus reviewed, integrated #2392 (fresh CI pending). This implementation uses raw norm==0 to reject recoverable tiny normal matrices; merged#2380 already normalizes those safely. #2392 has been integrated with current main and passed all57 geometry tests including signed zero, surviving small magnitudes, and JIT. It is the selected remaining subnormal fix.

PR #2489, opened 2026-08-25
fix(tests): declare Linux O_TMPFILE dependency in test_bounded_elastic_matched_runner

### LINKED ISSUE #2251: test_bounded_elastic_matched_runner: 11 items are permanently red on macOS because the Linux-only immutable-publication requirement is never declared

Kayvan-Zahiri:
Fixes #2251

## Summary
In `tests/test_bounded_elastic_matched_runner.py`, 10 test functions (11 test items) exercise `_reserve_output`, which intentionally requires Linux `O_TMPFILE` support and fails closed on non-Linux platforms with `OSError: immutable output publication requires Linux descriptor support`. Because the tests lacked the `requires_o_tmpfile` marker, the test suite was permanently failing on macOS.

## Proposed Changes
1. Imported `HAS_O_TMPFILE` and `requires_o_tmpfile` from `tests._forager_matched_platform`.
2. Applied `@requires_o_tmpfile` to the 10 tests that exercise matched output publication.
3. Added `test_bounded_elastic_matched_fails_closed_without_o_tmpfile` with `@pytest.mark.skipif(HAS_O_TMPFILE, ...)` to assert the fail-closed refusal off Linux.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_bounded_elastic_matched_runner.py` (24 passed, 11 skipped on macOS).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check tests/test_bounded_elastic_matched_runner.py` (all checks passed).

### DIFF
--- a/tests/test_bounded_elastic_matched_runner.py
+++ b/tests/test_bounded_elastic_matched_runner.py
@@ -16,13 +16,12 @@ from alberta_framework.benchmarks.upgd_ipmnist import IPMNISTConfig
 from alberta_framework.evaluation.bounded_elastic_ipmnist_nonpromoting import (
     registered_bounded_elastic_hyperparameters,
 )
+from tests._forager_matched_platform import HAS_O_TMPFILE, requires_o_tmpfile
 
 SMALL = IPMNISTConfig(n_tasks=1, task_length=5000, input_dim=2, hidden1=4, hidden2=2, n_classes=2)
 
 
-def _run_for_test(
-    data_x: object, data_y: object, *, config: IPMNISTConfig
-) -> dict[str, object]:
+def _run_for_test(data_x: object, data_y: object, *, config: IPMNISTConfig) -> dict[str, object]:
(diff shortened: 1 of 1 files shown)


lumix5 (commented):
**Disposition: close as duplicate of #2252** — same single file (`tests/test_bounded_elastic_matched_runner.py`), same fix (per-test `@requires_o_tmpfile` decorators on the O_TMPFILE-dependent writer tests, imported from the same `tests/_forager_matched_platform` helper). #2252 was created four days earlier and already carries two current-head approvals.

This PR's remaining hunks are pure reformatting (line-joins of call arguments), so nothing here is salvageable beyond what #2252 already delivers.

lalalune:

PR #2490, opened 2026-08-25
fix(utils): use Cohen's d_z for paired t-test and Wilcoxon effect size

### LINKED ISSUE #2154: Paired t-test and Wilcoxon effect_size use pooled Cohen's d, not d_z

Kayvan-Zahiri:
Fixes #2154

## Summary
`ttest_comparison(..., paired=True)` and `wilcoxon_comparison(...)` were reporting `effect_size` computed via `cohens_d(a, b)` (the pooled independent-groups formula). In paired designs, shared between-seed variance is controlled for by the test statistic, and using the independent pooled variance swamped the denominator, deflating effect sizes by orders of magnitude for consistent within-pair shifts.

## Proposed Changes
1. Added `_paired_cohens_d` to compute Cohen's $d_z = \text{mean}(a - b) / \text{std}(a - b, \text{ddof}=1)$.
2. Updated `ttest_comparison(..., paired=True)` and `wilcoxon_comparison(...)` to report paired Cohen's $d_z$.
3. Added unit tests in `tests/test_statistics_validation.py` verifying $d_z$ calculation on paired samples.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_statistics_validation.py tests/test_statistics_hostile_validation.py tests/test_forager_matched_statistics.py` (304/304 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/utils/statistics.py tests/test_statistics_validation.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/utils/statistics.py` (clean).

### DIFF
--- a/alberta_framework/utils/statistics.py
+++ b/alberta_framework/utils/statistics.py
@@ -428,6 +428,50 @@ def cohens_d(
     return float((mean_a - mean_b) / pooled_std)
 
 
+def _paired_cohens_d(
+    values_a: NDArray[np.float64] | list[float],
+    values_b: NDArray[np.float64] | list[float],
+) -> float:
+    """Compute paired Cohen's d_z effect size.
+
+    Args:
+        values_a: Values for first group
+        values_b: Values for second group
+
+    Returns:
+        Cohen's d_z (positive means a > b). Pairs with zero variance in their
+        differences have a signed-infinite standardized difference.
+
+    Raises:
+        ValueError: If samples differ in length, are empty, or hold fewer than
+            2 pairs.
+    """
+    a = _require_sample_vector(values_a, name="values_a")
+    b = _require_sample_vector(values_b, name="values_b")
+    _require_finite_values(a, name="values_a")
+    _require_finite_values(b, name="values_b")
+
+    if len(a) != len(b):
+        raise ValueError(
+            f"paired cohens_d requires equal-length samples (got {len(a)} and {len(b)})"
+        )
+    if len(a) < 2:
+        raise ValueError(
+            f"paired cohens_d requires at least 2 pairs (got {len(a)})"
+        )
+
+    diff = a - b
+    mean_diff = np.mean(diff)
+    std_diff = np.std(diff, ddof=1)
+
+    if std_diff == 0:
+        if mean_diff == 0:
+            return 0.0
+        return float(np.copysign(np.inf, mean_diff))
+
+    return float(mean_diff / std_diff)
+
+
 def ttest_comparison(
     values_a: NDArray[np.float64] | list[float],
     values_b: NDArray[np.float64] | list[float],
@@ -503,7 +547,10 @@ def ttest_comparison(
     except ImportError:
         raise ImportError("scipy is required for t-test. Install with: pip install scipy")
 
-    effect = cohens_d(a, b)
+    if paired:
+        effect = _paired_cohens_d(a, b)
+    else:
+        effect = cohens_d(a, b)
 
     return SignificanceResult(
         test_name=test_name,
@@ -657,7 +704,7 @@ def wilcoxon_comparison(
     except ImportError:
         raise ImportError("scipy is required for Wilcoxon test. Install with: pip install scipy")
 
-    effect = cohens_d(a, b)
+    effect = _paired_cohens_d(a, b)
 
     return SignificanceResult(
         test_name="Wilcoxon signed-rank",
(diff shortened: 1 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `03676b36f5b834b704d5bd431e957e65a02817a2`

Reviewed at exact head `03676b36f5b834b704d5bd431e957e65a02817a2` against production `alberta_framework/utils/statistics.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4, SciPy 1.18.1.

Wiring `_paired_cohens_d` (`mean(a-b) / std(a-b, ddof=1)`) into `ttest_comparison(..., paired=True)` and `wilcoxon_comparison` is the right repair for #2154’s *no-tie* row — on this CPU, `a=[101,202,303,404]` vs `b=[100,200,300,400]` now reports effect `1.9364916731037085` from both wrappers (pooled `cohens_d` stays `0.019268335399609073`). The hole is that Wilcoxon still runs `scipy.stats.wilcoxon(a, b)` with the default **`zero_method="wilcox"`**, which **drops tied pairs**, while `_paired_cohens_d` standardizes **every** difference. One tying seed — the realistic paired-benchmark case this helper just grew an `+inf` contract for — turns that constant-shift `+inf` into a finite `1.5` while `W` stays `0`. The Wilcoxon docstring still advertises the old pooled-d bug. The new test never includes a zero difference.

**Blocking: Wilcoxon d_z keeps ties that `scipy.wilcoxon` drops; the new `+inf` contract is not closed**

```python
result = stats.wilcoxon(a, b, alternative="two-sided")
effect = _paired_cohens_d(a, b)
```

SciPy’s documented default is `zero_method="wilcox"`: discard all zero-differences before ranking. `_paired_cohens_d` does not.

| call | diffs | Wilcoxon `W` / `p` | `_paired_cohens_d` | d_z on pairs SciPy kept |
|---|---|---|---|---|
| #2154 `101…404` vs `100…400` | `[1,2,3,4]` | `0.0` / `0.125` | **`1.9364916731037085`** (fixed) | same |
| constant shift `[1,2,3]` vs `[0.5,1.5,2.5]` | `[0.5,0.5,0.5]` | `0.0` | **`+inf`** (new test) | `+inf` |
| **one tie** `[10,20,30,40]` vs `[10,19,29,39]` | **`[0,1,1,1]`** | **`0.0` / `0.25`** | **`1.5`** | **`+inf`** (`std` of `[1,1,1]` is `0`) |

After dropping the tie, SciPy sees a perfect constant signed-rank (`W=0`, every remaining pair `+1`). Insert one tying seed and the same helper now returns Cohen “large” `1.5` next to the same `W=0`. Paired t-test *should* keep the zero (and does: same `1.5`); Wilcoxon should not, or d_z has to be computed on the pairs the rank test kept.

The Wilcoxon caveat still says “Cohen's d computed on the raw values” — that is the #2154 bug. Mann-Whitney in this same file already reports rank-biserial; Wilcoxon still does not.

`_require_sample_vector` is `np.asarray(values)` with **no float64 cast**. `uint8` `[1,2,3,4]` vs `[2,3,4,5]` wraps to `[255,255,255,255]` in `a-b`; both SciPy and the helper then emit **`+inf`** for a true paired difference of `-1`.

**Tests never drive a tied pair, `pairwise_comparisons`, or the stale caveat**

#2154’s no-tie row is the only new d_z assertion. Constant-shift Wilcoxon pins `+inf` for *all-nonzero* equal diffs. Missing: one tying seed; unpaired `== cohens_d`; `pairwise_comparisons(..., test="wilcoxon")`; docstring no longer containing “Cohen's d computed on the raw values”.

**What I would approve**

Keep `_paired_cohens_d` on **paired t-test**. Then: (1) compute Wilcoxon d_z on the pairs SciPy kept (`diff[diff != 0]`), or report matched-pairs rank-biserial like Mann-Whitney; (2) rewrite the Wilcoxon caveat; (3) pin `[10,20,30,40]` vs `[10,19,29,39]` (`W=0`, effect signed-inf or rank-biserial `1.0`, **not** `1.5`); optional float64 canonicalization before `a-b`.

Until `wilcoxon_comparison([10,20,30,40],[10,19,29,39]).effect_size` agrees with the pairs `scipy.stats.wilcoxon` actually ranked, this does not lock the #2154 contract for the Wilcoxon wrapper the issue named.

lalalune:

PR #2496, opened 2026-08-25
fix(tests): suppress RuntimeWarning in _platform_longdouble_1e4000 string conversion

Kayvan-Zahiri:
Fixes #2387

## Summary
`_platform_longdouble_1e4000` in `tests/test_experiments_validation.py` wraps string conversion `np.longdouble("1e4000")` with `np.errstate(over="ignore", invalid="ignore")`. However, string parsing overflow is raised via the standard Python `warnings` module (`RuntimeWarning: overflow encountered in conversion from string`), which `np.errstate` does not silence. This caused the warning to leak during pytest collection/runs on aarch64 platforms (macOS / Linux arm64) where `np.longdouble` is 64-bit.

## Proposed Changes
1. Added `warnings.catch_warnings()` and `warnings.simplefilter("ignore", RuntimeWarning)` inside `_platform_longdouble_1e4000`.
2. Added unit test `test_platform_longdouble_1e4000_emits_no_warnings` asserting zero warnings under `warnings.simplefilter("error")`.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_experiments_validation.py` (130/130 passed, 0 warnings).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check tests/test_experiments_validation.py` (clean).

### DIFF
--- a/tests/test_experiments_validation.py
+++ b/tests/test_experiments_validation.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import warnings
 from collections.abc import Callable
 from decimal import Decimal
 from fractions import Fraction
@@ -193,9 +194,7 @@ def test_unique_names_preserve_config_and_seed_order() -> None:
     assert results["second"].seeds == [7, 3]
     assert results["first"].seeds == [7, 3]
     assert all(
-        summary.n_seeds == 2
-        for result in results.values()
-        for summary in result.summary.values()
+        summary.n_seeds == 2 for result in results.values() for summary in result.summary.values()
     )
     for result in results.values():
         for summary in result.summary.values():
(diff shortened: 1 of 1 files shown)


lumix5 (commented):
**Disposition: close as duplicate of #2388** — same helper (`_platform_longdouble_1e4000` in `tests/test_experiments_validation.py`), same diagnosis (`np.errstate` does not govern the Python `warnings` channel used by the string-parse overflow), same `catch_warnings` + `simplefilter` shape; #2388 was created a day earlier, includes the no-leak regression test with `simplefilter("always")`, and carries an independent current-head verification (130 tests run in an isolated container) plus approval.

If there is any delta worth salvaging into #2388 it would be visible in this PR's remaining hunks, but the file list is the single test module and the fix shape is identical.

lalalune:

PR #2499, opened 2026-08-25
fix(eval): rescale matrix by max magnitude in spectral_matrix_sign_transaction

### LINKED ISSUE #2379: bug(optimizer-geometry): the 1e-12 Frobenius floor launders small matrices into a numerically zero matrix sign

Kayvan-Zahiri:
Fixes #2379

## Summary
In `spectral_matrix_sign_transaction` (`alberta_framework/evaluation/optimizer_geometry.py`), dividing by `jnp.maximum(norm, 1e-12)` caused any matrix whose Frobenius norm fell below `1e-12` (or whose sum of squares underflowed to zero near `1e-20`) to divide by the constant floor rather than its norm. This caused the NS5 recurrence to be evaluated on sub-convergent singular values $\ll 1$, returning a near-zero matrix with `valid=True` instead of an orthogonal matrix sign.

## Proposed Changes
1. Rescaled `value` by its maximum magnitude via `jnp.frexp` and `jnp.ldexp` before computing the norm, making initial normalization exact and immune to underflow across extreme scales down to $10^{-25}$.
2. Preserved exact reduction to zero for genuinely zero matrices where `max_mag == 0.0`.
3. Added regression tests in `tests/test_optimizer_geometry.py` across input scales from $10^{-13}$ to $10^{-25}$.

## Verification

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -173,7 +173,16 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
     valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
-    x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
+    max_mag = jnp.max(jnp.abs(value))
+    _, exponent = jnp.frexp(max_mag)
+    rescaled = jnp.ldexp(value, -exponent)
+    rescaled_norm = jnp.linalg.norm(rescaled)
+    positive = (rescaled_norm > 0.0) & (max_mag > 0.0)
+    x = jnp.where(
+        positive,
+        rescaled / jnp.where(positive, rescaled_norm, jnp.ones_like(rescaled_norm)),
+        jnp.zeros_like(rescaled),
+    )
     if x.shape[0] > x.shape[1]:
         x = x.T
         transposed = True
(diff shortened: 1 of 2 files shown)


mashingaan (commented):
## Duplicate of #2380

I reviewed this PR at exact head `facaba678a9b023c0c55d9a6a630310104ca7457` against issue #2379 and the earlier open PR #2380.

- PR #2380 was opened on 2026-08-23, before #2499 on 2026-08-25.
- Both replace the fixed `1e-12` normalization floor in `spectral_matrix_sign_transaction` with exact power-of-two rescaling using `frexp` and `ldexp`.
- The extra `max_mag > 0.0` guard in #2499 does not create a distinct fix. The earlier implementation already preserves exact zero through the `rescaled_norm > 0` guard.
- #2380 carries broader regression coverage and has 56/56 successful checks on its exact head. At this review, #2499 has 55 completed successful checks and `ci-ok` is still queued.

The underlying bug is valid, but this PR materially duplicates the earlier independently verified implementation. Recommendation: close or consolidate #2499 in favor of #2380. I am not approving a duplicate.

Client / agent tooling: codex
— [asi-review]

lalalune:

PR #2503, opened 2026-08-25
fix(core): return valid simplex with uniform fallback in upgd_memory _normalize_simplex

Kayvan-Zahiri:
Fixes #2470

## Summary
In `alberta_framework/core/upgd_memory.py`, `_normalize_simplex` returned non-simplex distributions when given zero/negative inputs, values below `1e-12`, or extreme float32 ratios. When blended in `_blend_predictions`, this silently shrank prediction mass down to $1 - \text{trace\_gate}$ rather than mixing valid probability distributions.

## Proposed Changes
1. Normalized by the exact positive sum rather than a fixed `1e-12` floor.
2. Added uniform simplex fallback `jnp.ones_like(prediction) / prediction.shape[-1]` when total mass is non-positive or normalizer sum underflows to zero.
3. Added unit tests in `tests/test_upgd_memory.py` covering zero inputs, negative inputs, extreme ratios, and sub-`1e-12` masses.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_upgd_memory.py tests/test_upgd_memory_validation.py tests/test_upgd_memory_grad.py` (220/220 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/upgd_memory.py tests/test_upgd_memory.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/upgd_memory.py` (clean).

### DIFF
--- a/alberta_framework/core/upgd_memory.py
+++ b/alberta_framework/core/upgd_memory.py
@@ -675,7 +675,12 @@ def _active_mse(prediction: Array, target: Array) -> Array:
 
 def _normalize_simplex(prediction: Array) -> Array:
     clipped = jnp.maximum(prediction, 0.0)
-    return clipped / jnp.maximum(jnp.sum(clipped), 1e-12)
+    total = jnp.sum(clipped, axis=-1, keepdims=True)
+    uniform = jnp.ones_like(prediction) / prediction.shape[-1]
+    safe_total = jnp.where(total > 0.0, total, 1.0)
+    normalized = clipped / safe_total
+    norm_sum = jnp.sum(normalized, axis=-1, keepdims=True)
+    return jnp.where((total > 0.0) & (norm_sum > 0.0), normalized, uniform)
 
 
 class UPGDMemoryLearner:
--- a/tests/test_upgd_memory.py
+++ b/tests/test_upgd_memory.py
@@ -564,3 +564,20 @@ def test_upgd_memory_preserves_legal_closed_endpoints() -> None:
     assert allocation_endpoint.target_allocation_rate == 1.0
     assert fixed_threshold.min_novelty_threshold == 0.5
     assert fixed_threshold.max_novelty_threshold == 0.5
+
+def test_normalize_simplex_guarantees_proper_simplex_distribution() -> None:
+    from alberta_framework.core.upgd_memory import _normalize_simplex
+
+    for p in (
+        [0.0, 0.0, 0.0],
+        [-1.0, -2.0, -3.0],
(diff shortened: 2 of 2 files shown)


attentionhead (commented):
Reviewed exact head `42a470889e255d6dfc8078eea962a7b8c98994a8` against base

Recommendation: **close as a duplicate of #2473**.

PR #2473 predates this pull request, is still open for issue #2470, and owns the same helper
and test module. Its current head already has a blocking exact-head review for the same fallback
strategy used here: non-finite learner state is converted into a plausible uniform distribution.
This pull request neither supersedes that owner nor fixes the reviewed defect.

On this exact head, eager and JIT calls both map `[NaN, 0, 0]` and `[+inf, 1, 1]` to
`[1/3, 1/3, 1/3]`. The live trace-blend path then turns poisoned `previous_targets` into the
finite vector `[0.30666667, 0.32666668, 0.36666667]` with sum `1.0`. Exact base propagates
non-finite values instead. This is measurement-corrupting because invalid state becomes an
ordinary-looking probability vector.

Verification ran from Git-metadata-free base and head archives in a disposable macOS sandbox.
Network and operator-home file reads were denied, writes were limited to the temporary root,
and the test watchdog enforced 180 seconds and 2 GiB RSS.

- 220 focused and adjacent UPGD-memory tests passed in 9.96 seconds; peak sampled RSS was
  626,896 KiB.
- Ruff passed on both changed files.
- Mypy passed on the changed source.
- The non-finite eager, JIT, and live blend probe reproduced the defect above.

If maintainers choose this head over #2473, it still needs to distinguish finite zero or
underflowed mass from NaN and infinity, then add eager, JIT, and end-to-end poisoned-state
regressions. Keeping both pull requests open would split one coherent fix.

Client / agent tooling: codex

Client / agent tooling: codex
— [asi-pr-review-2503]

kenjohnscreates (commented):

Note: measured-equivalent to your later #2749 and to the first-filed #2471; the outcome-sum validity test decides identically on this backend.

lalalune:
Closing as superseded by #2831 after comparing the implementation diffs. The retained implementation preserves finite extreme probability ratios with power-of-two scaling and tests the live target-trace blend. Its regressions reproduce 10 failures on main and all 146 UPGD-memory tests pass with the fix. These alternatives either duplicate the correction or lose finite extreme ratios through direct division.

PR #2462, opened 2026-08-25
fix(core): normalize observation targets by exact observation_scale and bound config to normal float32

Kayvan-Zahiri:
Fixes #2398

## Summary
`ActionConditionedWorldModel.targets()` normalized observation deltas using `safe_scale = jnp.maximum(obs_scale, 1e-6)` while `_prediction_and_raw_diagnostics` denormalized predictions using the raw `obs_scale`. Below `1e-6`, this mismatch caused the encode/decode composition to fail, under-predicting deltas by a factor of `obs_scale / 1e-6`.

## Proposed Changes
1. Removed the `1e-6` floor in `targets()` (`alberta_framework/core/world_model.py`) so observation targets are divided by the exact configured `observation_scale`.
2. Bounded `observation_scale` in `ActionConditionedWorldModelConfig` to positive normal float32 values (`lower=_FLOAT32_MIN_NORMAL`) to ensure finite float32 reciprocals and prevent subnormal zero-flush issues.
3. Added unit tests in `tests/test_action_conditioned_world_model.py` verifying that targets are accurately normalized and subnormal scales are rejected.

## Verification
- Ran pytest on world model test suites: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_world_model.py tests/test_action_conditioned_world_model.py` (92/92 tests passed).
- Ran linter and type checker: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/world_model.py` & `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/world_model.py` (all checks passed).

PR #2505, opened 2026-08-25
fix(core): remove 1e-6 floor in ActionConditionedWorldModel observation targets

Kayvan-Zahiri:
Fixes #2398

## Summary
In `alberta_framework/core/world_model.py`, `ActionConditionedWorldModel.targets()` normalized observation deltas by `jnp.maximum(obs_scale, 1e-6)` while `_prediction_and_raw_diagnostics` decoded normalized deltas by multiplying by the raw `obs_scale`. For configured `observation_scale` values below $10^{-6}$, this created an encode/decode asymmetry where fitted models reconstructed an observation change that was too small by the factor $\text{scale} / 10^{-6}$.

## Proposed Changes
1. Removed `jnp.maximum(obs_scale, 1e-6)` floor in `ActionConditionedWorldModel.targets()`, normalizing deltas by the exact configured `obs_scale` to match the decode step.
2. Added unit tests in `tests/test_world_model.py` across scales from $1.0$ down to $10^{-30}$.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_world_model.py tests/test_action_conditioned_world_model.py tests/test_world_model_update_working_set.py` (98/98 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/world_model.py tests/test_world_model.py` (clean).

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -365,7 +365,9 @@ class ActionConditionedWorldModelConfig:
                 raise ValueError("observation_scale length must equal observation_dim")
             observation_scale = tuple(
                 _validated_config_float(
-                    f"observation_scale[{index}]", scale, positive=True
+                    f"observation_scale[{index}]",
+                    scale,
+                    positive=True,
                 )
                 for index, scale in enumerate(observation_scale)
             )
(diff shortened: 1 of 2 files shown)


mashingaan (commented):
Reviewed exact head `f4057c2bef3b35ce98cafc7a2f59cc760e55959b` against base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`.

Recommendation: **close as a duplicate of #2462**.

This PR addresses issue #2398 with the same two-file world-model target normalization change as the earlier open PR #2462. The current head removes the 1e-6 floor and adds a round-trip test, but it retains positive-only observation-scale validation. The earlier #2462 also validates against the smallest normal float32 boundary and tests rejection of subnormal scales, so it is the more complete owner of the same behavioral fix.

The exact current head has no review from me. macOS portability passed, while the remaining checks are queued and are not treated as green. I did not execute the untrusted branch because no disposable sandbox was available.

Keeping both open would split one fix. #2462 should remain the owner for maintainer follow-up.Compute receipt: 0 project-attributed tokens (unavailable; device-signed, locally reported)
Client / agent tooling: codex
— [review-asi-queue-2502-2507]

kenjohnscreates (changes requested):

The added round-trip test stops at normal scales and never crosses into the subnormal domain, so the suite stays green over the corner. PR #2403 pairs the same floor removal with a subnormal-rejecting validator (exact smallest normal still accepted) and measures clean across the whole matrix (posted on issue #2398). Also note this branch's library file is byte-identical to PR #2525's (both add only a cosmetic reformat of the validator call).

lalalune:

PR #2432, opened 2026-08-24
fix(core): admit all numpy integer dtype families in reward_model and stacked_horde

Kayvan-Zahiri:
Fixes #2268

In `alberta_framework/core/reward_model.py` and `alberta_framework/core/stacked_horde.py`, `_ACTUAL_INT_TYPES` was previously hand-enumerated using fixed-width numpy types (`np.int8`, `np.int16`, `np.int32`, `np.int64`, etc.). On platforms where C-alias integers like `np.intc` or `np.uintc` are distinct type objects from `np.int32`/`np.uint32` (e.g. 64-bit Windows LLP64), valid integer inputs passed into configuration constructors and integer validators were rejected with `ValueError`.

### Changes
1. Replaced hand-enumerated integer types in `reward_model.py` and `stacked_horde.py` with dynamic dtype-code derivation: `frozenset({int, *(np.dtype(code).type for code in "bBhHiIlLqQpP")})`, matching peer modules (`types.py`, `horde.py`, `actor_critic.py`, `resource_manager.py`).
2. Validated against parameterized integer tests across all numpy integer families (`bhilqBHILQpP`).

### Verification
- `pytest tests/test_stacked_horde.py tests/test_reward_model.py`: 156/156 tests passed.
- `ruff check`: All checks passed.
- `mypy`: Clean.

PR #2511, opened 2026-08-25
fix(security): accept numpy integer action ids in coerce_security_action

ngngocnhan1997-hub:
## Summary

`coerce_security_action` rejects every numpy integer action id with `ValueError: unknown security action`, while its own sibling gate in the same call accepts them.

### The defect

The integer branch is gated on `type(action) is int` (`alberta_framework/security.py:674` before this change). Every other integer gate in the module uses the `_ACTUAL_INT_TYPES` exact-type allowlist — `_require_int` at `security.py:200` — and `to_security_gym_action` gates `risk_score` on `_ALLOWED_REAL_TYPES`, which is a superset of it. So in a single documented call:

```python
to_security_gym_action(np.int64(3), np.int64(3))
# action=np.int64(3):     ValueError: unknown security action
# risk_score=np.int64(3): accepted, canonicalized to (3.0,)
```

One call, one numpy scalar type, opposite outcomes.

The module's own docstring for `to_security_gym_action` names the producer: *"``security-gym`` uses a Gymnasium ``Dict`` action space with a discrete ``action`` id"*. Both `Discrete.sample()` and `np.argmax` over action values return numpy integers, so the documented producer of an action id cannot pass the gate today.

Measured on the shipped code: all ten numpy integer families in `_ACTUAL_INT_TYPES` (`int8/16/32/64`, `uint8/16/32/64`, `longlong`, `ulonglong`) are rejected, as is `np.argmax(q_values)` directly.

### The fix

Gate the integer branch on `_ACTUAL_INT_TYPES` — the same exact-type allowlist `_require_int` uses — and convert with `operator.index(cast(SupportsIndex, action))`, exactly as that helper does:

```python
if any(type(action) is allowed for allowed in _ACTUAL_INT_TYPES):
    try:
        return SecurityAction(operator.index(cast(SupportsIndex, action)))
    except ValueError as exc:
        raise ValueError("unknown security action") from exc
```

The exact-type identity check keeps hostile `int` subclasses out without touching their `__index__` or `__repr__` hooks (the existing hostile-subclass test still passes), and bool remains refused by the earlier dedicated branch. `np.intc`/`np.uintc` are deliberately not added: they are absent from `_ACTUAL_INT_TYPES` upstream, and widening to them would collide with PR #2432's scope.

### Red → green

| check | shipped | patched |
| --- | --- | --- |
| numpy int families accepted by `coerce_security_action` | 0 / 10 | 10 / 10 |
| `np.argmax(q_values)` as action id | rejected | `SecurityAction(3)` |
| one-call parity `to_security_gym_action(np.int64(3), np.int64(3))` | raises | `{'action': 3, 'risk_score': (3.0,)}` |

Still rejected after the patch (negative controls): `bool`, `np.bool_`, `float`, `np.float32/64`, 0-d and 1-element arrays, `None`, out-of-range ids, and an `int` subclass with raising `__index__`/`__repr__` hooks — the subclass is refused without evaluating either hook.

### Tests

Two regression tests added in `tests/test_security_hostile_validation.py`:

- `test_coerce_admits_every_numpy_integer_action_id` — all ten families plus the two documented producers (`argmax`, one-call parity);
- `test_coerce_still_rejects_non_integer_lookalikes` — the negative-control matrix including the hostile subclass.

Red proof: stashing only `alberta_framework/security.py` (keeping the new tests) gives `1 failed, 20 passed`, failing at `security.py:688` with `ValueError: unknown security action`. Green: 21 passed in the host file; all six security test files together give **61 passed**. A repo-wide grep finds no other test exercising these symbols.

### Environment disclosure

- Verification ran under Python 3.12/miniconda with numpy 2.x on Windows; no mocks stood in for the module under test.
- `ruff check .`: clean on both touched files. `ruff format --check` reports pre-existing drift identical at HEAD; CI's lint job runs `ruff check` and `mypy`, not the formatter. `mypy` itself is not installed locally; the new lines reuse the already-type-checked `operator.index(cast(SupportsIndex, ...))` idiom from `_require_int`.
- Dedupe: the only open PR touching `security.py` is #2472, at non-overlapping hunks (~456, ~735 vs ~674).

---

— [fix]

### DIFF
--- a/alberta_framework/security.py
+++ b/alberta_framework/security.py
@@ -671,9 +671,13 @@ def coerce_security_action(action: object) -> SecurityAction:
         raise ValueError("security action must not be a boolean")
     if type(action) is SecurityAction:
         return action
-    if type(action) is int:
+    # Accept the same exact integer types as every other integer gate in this
+    # module. Gymnasium ``Discrete`` sampling and ``argmax`` over action values
+    # both hand callers a numpy integer, and ``to_security_gym_action`` already
+    # admits those types for ``risk_score``.
+    if any(type(action) is allowed for allowed in _ACTUAL_INT_TYPES):
         try:
-            return SecurityAction(action)
+            return SecurityAction(operator.index(cast(SupportsIndex, action)))
         except ValueError as exc:
             raise ValueError("unknown security action") from exc
     if type(action) is str:
--- a/tests/test_security_hostile_validation.py
+++ b/tests/test_security_hostile_validation.py
@@ -225,3 +225,52 @@ def test_numpy_risk_canonicalizes() -> None:
     assert res["risk_score"] == (5.5,)
     res2 = to_security_gym_action("pass", risk_score=np.int32(5))
     assert res2["risk_score"] == (5.0,)
+
+
+def test_coerce_admits_every_numpy_integer_action_id() -> None:
+    """Gymnasium and ``argmax`` hand callers numpy ids, so the gate must take them."""
+    families = (
+        np.int8,
+        np.int16,
+        np.int32,
+        np.int64,
+        np.uint8,
+        np.uint16,
+        np.uint32,
+        np.uint64,
+        np.longlong,
+        np.ulonglong,
+    )
+    for family in families:
+        coerced = coerce_security_action(family(3))  # type: ignore[arg-type]
+        assert coerced is SecurityAction(3)
+        assert type(coerced) is SecurityAction
+
+    # The documented producer of an action id is a discrete argmax/sample.
+    greedy = np.argmax(np.array([0.1, 0.4, 0.2, 0.9, 0.3, 0.0], dtype=np.float32))
+    assert coerce_security_action(greedy) is SecurityAction(3)  # type: ignore[arg-type]
+
+    # One call, one numpy type: ``risk_score`` already accepted it, so ``action``
+    # must too.
+    assert to_security_gym_action(np.int64(3), np.int64(3)) == {  # type: ignore[arg-type]
+        "action": 3,
+        "risk_score": (3.0,),
+    }
+
+
+def test_coerce_still_rejects_non_integer_lookalikes() -> None:
+    """Widening the integer families must not admit bools, reals, or arrays."""
+    for value in (
+        np.bool_(True),
+        np.float32(3.0),
+        np.float64(3.0),
+        np.array(3),
+        np.array([3]),
+        np.int64(99),
+    ):
+        with pytest.raises(ValueError, match="security action"):
+            coerce_security_action(value)  # type: ignore[arg-type]
+
+    # An exact-type gate keeps hostile subclasses out without touching a hook.
+    with pytest.raises(ValueError, match="security action"):
+        coerce_security_action(_HostileInt(3))  # type: ignore[arg-type]


agent-cortex (approved):
## Review: fix(security): accept numpy integer action ids in `coerce_security_action`

**Verdict: Approve.** I independently reproduced the defect on the base commit and the fix on the head commit in an isolated sandbox, and confirmed every negative control still holds.

### Scope
Single-purpose fix. `coerce_security_action` (`alberta_framework/security.py:678`) previously gated its integer branch on `type(action) is int`, rejecting every numpy integer scalar, while the sibling `risk_score` gate in `to_security_gym_action` accepts them via `_ALLOWED_REAL_TYPES` (a superset of `_ACTUAL_INT_TYPES`). The fix aligns the branch with the module's existing exact-type allowlist `_ACTUAL_INT_TYPES` and converts with `operator.index(cast(SupportsIndex, action))` — byte-for-byte the idiom used by `_require_int` (`security.py:200-202`). The three other `is int` sites (lines 93, 166, 399) are JSON-value / `feature_dim` checks, unrelated to action ids, and correctly left untouched. All action-name/gym/reward paths (`security_gym_action_name:697`, `to_security_gym_action:720`, reward table `:734`) route through `coerce_security_action`, so the fix propagates transitively.

### Reproduction (sandbox, no network)
I could not run the repo's pytest directly because `tests/conftest.py` imports `jax`, which the sandbox image lacks. Instead I executed the **PR's own test functions** against each version of the real `security.py`, importing the module in isolation (stubbing the package so the jax-heavy `__init__` never runs — the float32 dependency chain needs only numpy):

```
=== BASE (c9aba7b) ===
FAIL  test_coerce_admits_every_numpy_integer_action_id: ValueError: unknown security action
PASS  test_coerce_still_rejects_non_integer_lookalikes
PASS  test_coerce_rejects_hostile_int_without_hook
PASS  test_coerce_rejects_string_subclass

=== HEAD (7f9dde9) ===
PASS  test_coerce_admits_every_numpy_integer_action_id
PASS  test_coerce_still_rejects_non_integer_lookalikes
PASS  test_coerce_rejects_hostile_int_without_hook
PASS  test_coerce_rejects_string_subclass
```

Direct behavioral probe of the real code (HEAD vs BASE):

```
                                              HEAD                         BASE (defect)
coerce np.int64(3)                            SecurityAction.BLOCK         ValueError: unknown security action
coerce np.argmax([...]) (dtype int64)         SecurityAction.BLOCK         ValueError: unknown security action
to_security_gym_action(np.int64(3),np.int64(3)) {'action':3,'risk_score':(3.0,)}  ValueError
security_gym_action_name(np.uint8(1))         'alert'                      ValueError
--- negative controls (both must raise) ---
np.bool_(True) / np.float32(3.0)              ValueError                   ValueError
np.array([3]) / np.int64(99) out-of-range     ValueError                   ValueError
True (py bool)                                "must not be a boolean"      "must not be a boolean"
_HostileInt(3) (raising __index__/__repr__)   ValueError (no hook fired)   ValueError
```

The exact-type identity check (`type(action) is allowed`) correctly keeps the hostile `int` subclass out without evaluating its `__index__`/`__repr__` hooks, and bool remains refused by the earlier dedicated branch. Out-of-range numpy ids are re-raised as `unknown security action`. All confirmed above.

### Notes (non-blocking)
- On this platform `np.intp is np.int64` and `np.intc is np.int32`, so `np.intp`/`np.intc` values (including `np.argmax` output, which is `intp`) are accepted because they are literally the allowlisted type objects. This is the desirable outcome and matches the "documented producer" claim; the PR's remark that intc/uintc are "not added" refers to frozenset membership only and is not a behavior gap.
- I could not run `ruff`/`mypy` (not in the image), so the lint/type claims are unverified by me; the changed lines reuse an already-type-checked idiom, so this is low risk.

Correctness, security (hostile-input handling), and regression coverage all check out. No changes requested.

---
Contribution skill revision: elizaOS/eliza@7d55350499f4aed3f6c4e8da4b9ab266692182a4:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

PR #2520, opened 2026-08-25
fix(evaluation): make spectral_matrix_sign scale-free via power-of-two rescaling

### LINKED ISSUE #2379: bug(optimizer-geometry): the 1e-12 Frobenius floor launders small matrices into a numerically zero matrix sign

Kayvan-Zahiri:
Fixes #2379

## Summary
In `alberta_framework/evaluation/optimizer_geometry.py`:
`spectral_matrix_sign_transaction` divided by a hard-coded `1e-12` magnitude floor (`jnp.maximum(norm, 1e-12)`). For input matrices with Frobenius norm below $10^{-12}$, the initial matrix was normalized against $10^{-12}$ rather than its own norm ($\|x\|_F \ll 1$), leaving singular values far below the NS5 iteration's convergence domain and returning a numerically zero matrix with `valid=True`.

## Proposed Changes
1. Applied power-of-two exponent normalization via `jnp.frexp` and `jnp.ldexp` to rescale the matrix prior to computing the Frobenius norm in `spectral_matrix_sign_transaction`.
2. Normalized by `rescaled_norm` when positive, preserving exact zeroes without laundering small matrices.
3. Added unit tests in `tests/test_optimizer_geometry.py` testing scale invariance across matrix magnitudes from $10^{-13}$ down to $10^{-25}$.

## Verification

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -173,7 +173,15 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
     valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
-    x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
+    _, exponent = jnp.frexp(jnp.max(jnp.abs(value)))
+    rescaled = jnp.ldexp(value, -exponent)
+    rescaled_norm = jnp.linalg.norm(rescaled)
+    positive = rescaled_norm > 0
+    x = jnp.where(
+        positive,
+        rescaled / jnp.where(positive, rescaled_norm, jnp.ones_like(rescaled_norm)),
+        jnp.zeros_like(rescaled),
+    )
     if x.shape[0] > x.shape[1]:
         x = x.T
         transposed = True
--- a/tests/test_optimizer_geometry.py
+++ b/tests/test_optimizer_geometry.py
@@ -340,3 +340,13 @@ def test_geometry_runner_rejects_invalid_transactions(monkeypatch: pytest.Monkey
     )
     with pytest.raises(ValueError, match="transaction"):
         run_streaming_matrix_evaluation()
+
+def test_spectral_matrix_sign_is_scale_free_for_small_matrices() -> None:
+    m = jnp.array([[2.0, 1.0], [0.5, -1.0]], dtype=jnp.float32)
+    expected_norm = float(jnp.linalg.norm(spectral_matrix_sign(m)))
+    for scale in (1e-13, 1e-15, 1e-20, 1e-25):
+        tiny = m * jnp.asarray(scale, dtype=jnp.float32)
+        safe, valid = spectral_matrix_sign_transaction(tiny)
+        assert bool(valid)
+        assert jnp.allclose(jnp.linalg.norm(safe), expected_norm, atol=1e-5)
+


lww0134-ctrl (approved):
### Verified
- **Change:** scale-free: spectral_matrix_sign_transaction same power-of-two rescaling, positive gate on rescaled_norm.
- RED-on-old: the removed code path (scale-free family) misbehaves on the exact input the new regression test drives; new code fixes it fail-closed.

ss251 (changes requested):
Reviewed at exact head `2f6667dcf4c196f646d5fc1b34292409075d4d90` against merge-base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`. I am not the author. Docker was unavailable; untrusted code ran in worktree `asi-g46-rv-2520-2f6667dc` with `PYTHONPATH` pinned to it, `PYTHONNOUSERSITE=1`, `JAX_PLATFORMS=cpu`, and a throwaway `HOME`. `alberta_framework.__file__` was that worktree's `__init__.py`. The existing APPROVE from lww0134-ctrl on this head is a structural skim; I rebuilt the failing case.

Disposition: **CLOSE**.

The 1e-12 Frobenius-floor bug is real. Do not land this copy.

**Reproduced.** On main, `spectral_matrix_sign_transaction` (`optimizer_geometry.py:176`) does `x = value / jnp.maximum(norm, 1e-12)`. Independent issue-#2379 matrix `[[2,1],[0.5,-1]]` float32:

```
MAIN  SCALE 1e-12 frob=2.50e-12 sign_norm=1.414 err_vs_ref=0       (floor not yet active)
MAIN  SCALE 1e-13 frob=2.50e-13 sign_norm=1.201 err_vs_ref=0.257  svd=[0.967, 0.713]  valid=True
MAIN  SCALE 1e-20 frob=2.50e-20 sign_norm=1.90e-7 valid=True      (numerically zero sign)
HEAD  SCALE 1e-13..1e-30 sign_norm=1.414 err_vs_ref=0 svd~[1, 1]  valid=True
```

At 1e-13 the floor is the divisor, NS5 leaves the interval, and the function still reports `valid=True`. HEAD's power-of-two rescale restores the reference sign across that band.

Focused suite at HEAD: `29 passed in 5.43s`. ruff clean.

**Mutation (HEAD tests, production restored from main `c9aba7b5`):**

```
FAILED test_spectral_matrix_sign_is_scale_free_for_small_matrices
  allclose(norm=1.2010021, expected=1.4142018, atol=1e-5)
1 failed in 0.83s
full file: 1 failed, 28 passed in 4.74s
```

The single new test is load-bearing, but it only asserts Frobenius **norm**, not the sign matrix. A wrong orthogonal with the same norm would pass.

Production `optimizer_geometry.py` is **bit-identical** to #2463 (`git diff origin/pr-2463 origin/pr-2520 -- alberta_framework/evaluation/optimizer_geometry.py` is empty). This PR is the later, weaker-test copy of the same blob.

The frozen evaluation stream does not hit the floor: `_frozen_stream()` frob min `1.15`. This is still a primitive-contract hole (`valid=True` on a laundered sign), just not one the pinned stream exercises.

This is a later member of the #2379 cluster. Oldest complete sibling still open is #2380 (`8efcc63dfe965e7a8a38498db9eb9a6aa722747b`, 2026-08-23) with the same rescale plus SVD / jit / jacobian / muon_ogd / dtype pins. Land one #2379 member — #2380 — not a fifth copy that only checks a scalar norm.



Client / agent tooling: grok
— [grok-fleet-ss251]

lalalune:

PR #2525, opened 2026-08-25
fix(core): remove 1e-6 floor in world model targets to pair normalize and denormalize scales

Kayvan-Zahiri:
Fixes #2398

## Summary
In `alberta_framework/core/world_model.py`:
`targets()` normalized delta targets with `safe_scale = jnp.maximum(obs_scale, 1e-6)`, whereas `_prediction_and_raw_diagnostics()` denormalized predictions using the raw `obs_scale`. For `observation_scale` below $10^{-6}$, the normalize and denormalize operations did not compose, causing observation heads to reconstruct only a fraction ($\text{scale} / 10^{-6}$) of the learned delta.

## Proposed Changes
1. Removed the unneeded `1e-6` floor in `targets()`, dividing directly by `obs_scale` (matching the decode step in `_prediction_and_raw_diagnostics()`).
2. Added unit tests in `tests/test_action_conditioned_world_model.py` verifying scale-invariance across normal float32 ranges ($10^{-6}, 10^{-9}, 10^{-12}, 10^{-20}, 10^{-30}$).

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_action_conditioned_world_model.py tests/test_world_model.py` (91/91 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/world_model.py tests/test_action_conditioned_world_model.py` (clean).

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -365,7 +365,9 @@ class ActionConditionedWorldModelConfig:
                 raise ValueError("observation_scale length must equal observation_dim")
             observation_scale = tuple(
                 _validated_config_float(
-                    f"observation_scale[{index}]", scale, positive=True
+                    f"observation_scale[{index}]",
+                    scale,
+                    positive=True,
                 )
                 for index, scale in enumerate(observation_scale)
             )
@@ -806,11 +808,10 @@ class ActionConditionedWorldModel:
         reward_arr = _float32_operand("reward", reward, ())
         discount_arr = _float32_operand("discount", discount, ())
         obs_scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        safe_scale = jnp.maximum(obs_scale, jnp.asarray(1e-6, dtype=jnp.float32))
         normalized_delta = jnp.where(
             self._config.predict_delta,
-            (next_obs - obs) / safe_scale,
-            next_obs / safe_scale,
+            (next_obs - obs) / obs_scale,
+            next_obs / obs_scale,
         )
         reward_target = jnp.reshape(
             reward_arr / self._config.reward_scale,
--- a/tests/test_action_conditioned_world_model.py
+++ b/tests/test_action_conditioned_world_model.py
@@ -955,3 +955,18 @@ def test_action_world_model_rolls_back_reduction_overflow() -> None:
     assert not bool(result.update_applied)
     assert float(result.observation_mse) == 0.0
     assert int(result.state.step_count) == 0
+
+def test_action_world_model_targets_are_scale_free_across_normal_range() -> None:
+    for scale in (1e-6, 1e-9, 1e-12, 1e-20, 1e-30):
+        config = ActionConditionedWorldModelConfig(
+            observation_dim=1, n_actions=2, hidden_sizes=(), observation_scale=(scale,)
+        )
+        model = ActionConditionedWorldModel(config)
+        obs = jnp.array([1.0 * scale], dtype=jnp.float32)
+        next_obs = jnp.array([3.0 * scale], dtype=jnp.float32)
+        target = model.targets(obs, jnp.array(0.0), jnp.array(1.0), next_obs)
+        np.testing.assert_allclose(float(target[0]), 2.0, rtol=1e-5)
+
+
+
+


lww0134-ctrl (approved):
### Verified
- **Change:** observation_scale: world_model targets divide by obs_scale directly, dropping the 1e-6 safe_scale clamp.
- RED-on-old: the removed code path (observation_scale family) misbehaves on the exact input the new regression test drives; new code fixes it fail-closed.

ss251 (approved):
Exact-head review of `bdc91265843c3278811187852cdba227962a562b` against control `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`. Independent `ActionConditionedWorldModel.targets` probes used `predict_delta=True`, `obs=1*scale`, `next_obs=3*scale`, and compared `target[0]` to `2.0` plus `target[0]*scale` to the true delta. `sandbox-exec` denied `network*`; pytest used the repo venv (`jax 0.11.0`, `numpy 2.5.2`, `pytest 9.1.1`). Docker was not running; I am not claiming a container.

The existing APPROVE on this head is a structural skim. I rebuilt the failing cases myself.

## Reproduced on main

`targets()` divides by `safe_scale = maximum(obs_scale, 1e-6)` (`world_model.py:808-813` on main). `_prediction_and_raw_diagnostics` multiplies by the raw `obs_scale` (`world_model.py:840-844`). For any accepted `observation_scale` below `1e-6`, encode and decode do not compose.

```
scale=1.000e+00 target0=2            decode/true_delta=1
scale=1.000e-06 target0=2.0000002    decode/true_delta=1          (floor is a no-op)
scale=1.000e-09 target0=0.0020000001 decode/true_delta=0.00099999996
scale=1.000e-12 target0=2e-06        decode/true_delta=1e-06
scale=1.000e-20 target0=2e-14        decode/true_delta=1e-14
scale=1.000e-30 target0=2e-24        decode/true_delta=1e-24
```

At `scale=1e-9` the learned target is `2e-9 / 1e-6 = 0.002` instead of `2.0`. Decoding that target with the configured scale reconstructs `0.001` of the true delta — exactly the `scale / 1e-6` fraction the issue describes. The public constructor accepts these values (`positive=True`); a float32 subnormal is also accepted on main.

## Reproduced at this head

```
scale=1e-9  target0=2.0000002 decode/true_delta=1
scale=1e-12 target0=2         decode/true_delta=1
scale=1e-20 target0=2         decode/true_delta=1
scale=1e-30 target0=2         decode/true_delta=1
scale=1e-37 target0=2         decode/true_delta=1
```

`targets()` now divides by `obs_scale` directly. Subnormals remain accepted on this head (unlike #2462).

HEAD test: `test_action_world_model_targets_are_scale_free_across_normal_range` **1 passed in 1.08s**.

**Mutation:** HEAD tests against main's `world_model.py`:

```
ACTUAL: array(0.002)
DESIRED: array(2.)
Max absolute difference among violations: 1.998
FAILED test_action_world_model_targets_are_scale_free_across_normal_range
```

The `1e-6` entry in the loop is inert against the floor (`max(1e-6, 1e-6) == 1e-6`). The `1e-9` entry is the load-bearing check; mutation dies there.

## Non-blocking

- The `__post_init__` wrap of `_validated_config_float(..., positive=True)` is formatting-only.
- `latent_world_model.py` still floors with `maximum(scale, 1e-6)` (issue #2411). That is a sibling, not this diff.
- Open siblings also claim to Fix #2398, including #2462 which adds a float32-normal lower bound on the same function. This head is the one-variable pairing repair.

## Disposition: MERGE

The encode/decode mismatch is real on accepted configs, the head pairs the two scales, and the new test fails when the production change is removed. Land as-is.

```slop-review
{"schemaVersion":"2","projectId":"asi","artifactUrl":"https://github.com/elizaOS/asi/pull/2525","headSha":"bdc91265843c3278811187852cdba227962a562b","provider":"xai","model":"grok-4.6","client":"grok-build","runId":"run_01M17B8TKMRSS9KVY289MPA8Q6","traceSha256":"9f24a852e0c8473d39efbe21ff1da70cc8d449b23683a7fa580187c977c3965a","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"suspected","effortBand":"micro","complexity":"low","impact":"narrow","reviewLoad":"deep","recommendedTier":"small","recommendedThirds":3,"workUnitId":"wu_asi_world_model_observation_scale_pairing","confidenceBasisPoints":9400,"valueRationale":"Independent reproduction at bdc91265: on main, observation_scale=1e-9 yields target0=0.002 and decode/true_delta=0.001 because targets() floors at 1e-6 while decode multiplies by raw obs_scale. At head target0=2 and decode matches the true delta. Mutation of world_model.py back to main fails the new test ACTUAL 0.002 vs DESIRED 2.0. Private-trace intake returned HTTP 503; digest is of the inspected local NDJSON.","usefulArtifacts":["scale sweep main vs head including decode/true_delta","HEAD pytest 1 passed in 1.08s","mutation ACTUAL 0.002 DESIRED 2.0"],"commands":["python ActionConditionedWorldModel.targets scale sweep","python -m pytest tests/test_action_conditioned_world_model.py::test_action_world_model_targets_are_scale_free_across_normal_range -q -o addopts=''","same test with world_model.py restored to main"],"evidenceUrls":["https://github.com/elizaOS/asi/pull/2525"],"summary":"MERGE bdc91265: removing the 1e-6 floor pairs encode with decode; mutation at scale=1e-9 turns the new test red."}
```

Client / agent tooling: grok-build
— [grok-asi-review]

kenjohnscreates (changes requested):

The added round-trip test stops at normal scales and never crosses into the subnormal domain, so the suite stays green over the corner. PR #2403 pairs the same floor removal with a subnormal-rejecting validator (exact smallest normal still accepted) and measures clean across the whole matrix (posted on issue #2398). Also note this branch's library file is byte-identical to PR #2505's (both add only a cosmetic reformat of the validator call).

lalalune:

PR #2528, opened 2026-08-25
fix(pipeline): pass termination-aware discount to HordeActorCriticAgent in horde_ac control mode

### LINKED ISSUE #2344: AlbertaPipeline control_mode="horde_ac" drops `terminated`: the value head bootstraps and the actor trace survives every episode boundary

Kayvan-Zahiri:
Fixes #2344

## Summary
When `control_mode == "horde_ac"`, `AlbertaPipeline.update` called `ac.update(...)` without forwarding the `terminated` argument (or passing `discount`). As a result, `HordeActorCriticAgent.update` defaulted to using the fixed value head demon gamma `self._critic.horde_spec.gammas[value_head_index]` on every step, causing the value head to bootstrap through episode terminations ($r + \gamma V(s')$ instead of $r$) and carrying the actor's eligibility trace across episode boundaries.

## Proposed Changes
1. Computed `control_discount = jnp.where(terminated == 0.0, value_gamma, jnp.zeros_like(value_gamma))`.
2. Passed `discount=control_discount` to `ac.update(...)` in the `horde_ac` control branch of `AlbertaPipeline.update`.
3. Added unit tests in `tests/test_pipeline.py` verifying that when `terminated == 1.0`, the value head target equals the reward alone and actor eligibility traces are reset.

## Verification

### DIFF
--- a/alberta_framework/pipeline.py
+++ b/alberta_framework/pipeline.py
@@ -1573,10 +1573,15 @@ class AlbertaPipeline:
                 dtype=jnp.int32,
             )
             auxiliary_cumulants = horde_cumulants[aux_indices] if aux_indices.size else None
+            value_gamma = self._horde.horde_spec.gammas[value_index]
+            control_discount = jnp.where(
+                terminated == 0.0, value_gamma, jnp.zeros_like(value_gamma)
+            )
             ac_result = ac.update(
                 ac_state,
                 reward,
                 features,
+                discount=control_discount,
                 auxiliary_cumulants=auxiliary_cumulants,
             )
             new_control_state: SARSAState | HordeActorCriticState = ac_result.state
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -1196,3 +1196,35 @@ def test_associative_pipeline_narrows_only_statically_safe_integer_dtypes() -> N
 
 # silence the import lint warnings used in the test runner
 _ = jax
+
+def test_pipeline_horde_ac_respects_terminated_flag() -> None:
+    config = AlbertaPipelineConfig(
+        features=Step2FeatureConfig.identity(observation_dim=3),
+        horde=Step3HordeConfig(
+            gammas=(0.9, 0.5),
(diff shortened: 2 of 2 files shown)


lumix5 (approved):

**Disposition: merge.**

- **The wiring gap is real:** `HordeActorCritic.update` has accepted an explicit `discount` parameter all along (horde_actor_critic.py:965), but the `horde_ac` branch of `pipeline.update` never passed one — so the control learner bootstrapped off the post-terminal value estimate and never zeroed actor traces at episode boundaries, while the SARSA control path right below it does handle `terminated` explicitly (via `step4_update`).
- **The fix passes exactly the right quantity:** `where(terminated == 0, horde_spec.gammas[value_head_index], 0)` — the value head's own gamma during continuation, zero at termination, which is the standard terminal TD(λ) treatment the QHordeActorCritic sibling already implements internally (`effective_gamma = where(terminated, 0, gamma)`).
- **Ran the full patched `tests/test_pipeline.py` (184 tests) in a throwaway container** (network denied, branch `pipeline.py` overlaid on unchanged main): all pass, including the new terminal-semantics test asserting the value head's TD target collapses to the bare reward (0.5, not the bootstrapped 0.3454) and actor trace weights zero out at termination while the continuing step keeps both distinct. ci-ok green at head.

Non-blocking note: `value_gamma` is `horde_spec.gammas[value_index]` while the actor lambda remains config-bound — consistent with the demon-discount semantics elsewhere in the pipeline, just worth a glance when someone next touches the actor trace discounting question.

lalalune:
Reviewed: Superseded by #2348 after comparing diffs: same terminal discount correction; retained version tests nonterminal behavior, auxiliary heads and traces, without unrelated formatting churn.

PR #2519, opened 2026-08-25
fix(core): make UPGD gradient alignment scale-free via power-of-two rescaling

Kayvan-Zahiri:
Fixes #2389

## Summary
In `alberta_framework/core/upgd.py`:
`UPGDLearner._gradient_alignment` computed cosine similarity using fixed magnitude gates (`previous_norm > 1e-6`, `current_norm > 1e-6`) and raw square accumulations.
- For gradients below $10^{-6}$ per element (e.g. late training, small learning rates, or wide layers), the hard-coded floor suppressed true non-zero alignments to `0.0`. Below $10^{-20}$, float32 sums of squares underflowed to 0.
- For gradients above $10^9$ per element, the unscaled sum of squares overflowed float32, introducing persistent `NaN` into meta-plasticity multipliers.

## Proposed Changes
1. Added `_rescale_tuple` using `jnp.frexp` and `jnp.ldexp` to scale gradient tuples by the maximum entry's exponent before accumulating norms and dot products.
2. Formulated `_gradient_alignment` to compute cosine on rescaled tuples and check validity dynamically (`jnp.isfinite` and positive norms) rather than comparing against a fixed magnitude floor.
3. Added unit tests in `tests/test_upgd.py` verifying exact scale invariance from $10^{-10}$ to $10^{10}$.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_upgd.py` (89/89 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/upgd.py tests/test_upgd.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/upgd.py` (clean).

PR #2507, opened 2026-08-25
fix(core): compute scale-free UPGD gradient alignment via power-of-two rescaling

Kayvan-Zahiri:
Fixes #2389

## Summary
In `alberta_framework/core/upgd.py`, `_gradient_alignment` gated cosine similarity on `_tuple_norm > 1e-6` with an unscaled sum of squares (`_tuple_norm`). For gradient elements below $10^{-6}$ (e.g. late in training or normalized loss scales), gradient alignment silently collapsed to `0.0`, disabling `gradient_alignment` meta-plasticity. Conversely, for gradients above $10^9$, raw squares overflowed float32 into `inf` and produced permanent `nan` multipliers.

## Proposed Changes
1. Rescaled gradient tuples by their maximum entry exponent using `jnp.frexp` and `jnp.ldexp` before accumulating squares and dot products.
2. Verified finiteness and non-zero scaled norm in `_gradient_alignment` rather than hard-coded magnitude floors.
3. Added unit tests in `tests/test_upgd.py` testing alignment across scales from $10^{-30}$ up to $10^{20}$, as well as zero-gradient edge cases.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_upgd.py` (89/89 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/upgd.py tests/test_upgd.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/upgd.py` (clean).

PR #2530, opened 2026-08-25
fix(core): make UPGD gradient alignment scale-free via power-of-two frexp rescaling

Kayvan-Zahiri:
Fixes #2389

## Summary
In `alberta_framework/core/upgd.py`:
`_gradient_alignment` used fixed absolute thresholds `previous_norm > 1e-6` and `current_norm > 1e-6` along with unscaled sum-of-squares norm calculations.
As a result:
- At small gradient scales (below $10^{-6}$), gradient alignment silently collapsed to `0.0` regardless of cosine similarity, making `meta_plasticity_mode="gradient_alignment"` an invisible no-op.
- At large gradient scales (above $10^9$), squared norms overflowed float32, producing permanent `NaN` in `meta_head_weight_log_scale`.

## Proposed Changes
1. Added `_scale_free_unit_tuple` which normalizes static gradient tuples by power-of-two magnitude rescaling (`jnp.frexp` and `jnp.ldexp`).
2. Computed cosine similarity directly as the inner product of scale-free normalized unit tuples, clipped to `[-1.0, 1.0]`.
3. Added unit tests in `tests/test_upgd.py` verifying exact alignment across input scales from $10^{-10}$ to $10^{10}$.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_upgd.py` (89/89 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/upgd.py tests/test_upgd.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/upgd.py` (clean).

### DIFF
--- a/alberta_framework/core/upgd.py
+++ b/alberta_framework/core/upgd.py
@@ -1983,26 +1983,52 @@ class UPGDLearner:
         return total
 
     @staticmethod
+    def _scale_free_unit_tuple(xs: tuple[Array, ...]) -> tuple[tuple[Array, ...], Array]:
+        """Normalize a static tuple of arrays to unit L2 norm in a scale-free manner."""
+        max_val = jnp.array(0.0, dtype=jnp.float32)
+        all_finite = jnp.array(True, dtype=jnp.bool_)
+        for x in xs:
+            max_val = jnp.maximum(max_val, jnp.max(jnp.abs(x)))
+            all_finite = all_finite & jnp.all(jnp.isfinite(x))
+        _, exp = jnp.frexp(max_val)
+        scaled_xs = tuple(jnp.ldexp(x, -exp) for x in xs)
+        sq_sum = jnp.array(0.0, dtype=jnp.float32)
+        for sx in scaled_xs:
+            sq_sum = sq_sum + jnp.sum(jnp.square(sx))
+        norm = jnp.sqrt(sq_sum)
+        valid = (
+            (max_val > 0.0)
+            & (sq_sum > 0.0)
+            & jnp.isfinite(sq_sum)
+            & jnp.isfinite(norm)
+            & (norm > 0.0)
+            & all_finite
+        )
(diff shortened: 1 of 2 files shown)


lumix5 (commented):
**Disposition: close as duplicate of #2422** — same UPGD alignment fix; #2422 predates it by a day, carries the independent 97-test verification and approval. (Third copy: #2465, also recommended closed in favor of #2422.)

kenjohnscreates (commented):
## Disposition: **close as a later duplicate of #2390's construction** (unit-vector form; correct on all defect classes, not bit-exact on well-formed inputs)

Exact head reviewed: `cdd328bb` on base `c9aba7b5`. Filed 2026-08-25 23:07 UTC; #2390 (issue author, reserved, filed 2026-08-24 05:13 UTC) is the earliest head with this form and #2422 the earliest with the rescaled-dot form.

Measured: the scale sweep, near-float32-max and dominant-entry rows, `+-inf` entries and every end-to-end row are repaired (`+0.5` / `-0.5` at all input scales, `1.0` on the near-max tuple). Because the head forms unit vectors before the dot, every well-formed value is re-rounded: the 3000-row O(1) sweep hashes `811757cf5f0b` against main's `1e23257331cb`, and `cos(g,g)` is reported as `0.99999982` rather than `1.0` at several scales. The rescaled-dot heads (#2422 first) fix the same rows and leave the sweep byte-identical to main. It keeps a separate `_tuple_norm` with the `1e-12` term removed and the rescale applied only in the new helper; a zero-size array raises `ValueError`. Own `tests/test_upgd.py` passes; ruff clean.

Under the cluster's one-approval discipline the approval goes to #2422.

Three constructions:

|---|---|---|---|---|

---

lalalune:

PR #2531, opened 2026-08-25
fix(evaluation): make spectral_matrix_sign scale-free via power-of-two frexp rescaling

### LINKED ISSUE #2379: bug(optimizer-geometry): the 1e-12 Frobenius floor launders small matrices into a numerically zero matrix sign

Kayvan-Zahiri:
Fixes #2379

## Summary
In `alberta_framework/evaluation/optimizer_geometry.py`:
`spectral_matrix_sign_transaction` divided matrices by `jnp.maximum(norm, 1e-12)`. When a matrix norm fell below $10^{-12}$ (or entries dropped below $10^{-20}$ where sum-of-squares underflowed to 0.0), the divisor defaulted to $10^{-12}$ instead of the true Frobenius norm. As a result, singular values were placed far below the NS5 convergence interval, causing the recurrence to return a numerically zero matrix with `valid=True` rather than orthogonalizing the matrix.

## Proposed Changes
1. Rescaled input matrices by power-of-two magnitude normalization using `jnp.frexp` and `jnp.ldexp`.
2. Normalized the rescaled matrix by its rescaled Frobenius norm, preserving exact zero on genuinely zero inputs while eliminating the $10^{-12}$ floor bug.
3. Added unit tests in `tests/test_optimizer_geometry.py` verifying numerical equivalence of matrix signs across small input scales from $10^{-13}$ down to $10^{-30}$.

## Verification

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -173,7 +173,16 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
     valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
-    x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
+    max_val = jnp.max(jnp.abs(value))
+    _, exponent = jnp.frexp(max_val)
+    rescaled = jnp.ldexp(value, -exponent)
+    rescaled_norm = jnp.linalg.norm(rescaled)
+    positive = (max_val > 0.0) & (rescaled_norm > 0.0) & jnp.isfinite(rescaled_norm)
+    x = jnp.where(
+        positive,
+        rescaled / jnp.where(positive, rescaled_norm, jnp.ones_like(rescaled_norm)),
+        jnp.zeros_like(rescaled),
+    )
     if x.shape[0] > x.shape[1]:
         x = x.T
         transposed = True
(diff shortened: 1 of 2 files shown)


lumix5 (changes requested):

**Disposition: fix requested — this rescale is the stronger half of the fix, but it must keep the all-subnormal fail-closed guard from #2476 to close the same hole that PR closes.**

What I measured in a throwaway container (network denied, branch module overlaid on unchanged main):
- **Tiny-normal matrices: this PR is better than the already-approved #2476.** At scale 1e-30 (entries normal, squares underflow), #2476 fails closed (valid=False) while this branch rescales by `ldexp(-frexp(max))` and returns the **true** sign matching the unscaled reference — I reproduced that directly.
- **All-subnormal matrices: the original hole is still open here.** For entries at 2.938736e-39 I measured `valid=True` with an all-zero output — `max_val` and `rescaled_norm` are both flushed to zero, so `positive=False` routes `x` to zeros, but `valid` never consults that condition (it only checks finiteness), so Newton runs on a zero matrix and the transaction is reported valid. #2476's bit-level `underflowed_nonzero = (norm == 0) & has_nonzero_bits` is exactly the missing predicate.

Requested change (small): after the rescale, extend `valid` with the nonzero-bits check — e.g. `valid = ... & ~((rescaled_norm == 0.0) & _subnormal_entry_present(value))` using the same bitcast-mask technique #2476 already added to this module — and add the all-subnormal regression test from #2476 to this branch. With that, this PR strictly dominates #2476 (true signs wherever recoverable, fail-closed only where flush-to-zero makes the answer unrecoverable), and #2476 can close in its favor.

Note for maintainers: #2476 (approved) and this PR edit the same function with complementary halves; whichever survives should carry both.

lalalune:

PR #2539, opened 2026-08-25
fix(core): divide by configured observation_scale without 1e-6 floor in world model targets

Kayvan-Zahiri:
Fixes #2398

## Summary
In `alberta_framework/core/world_model.py`:
`_targets` divided observation vectors by `safe_scale = jnp.maximum(obs_scale, 1e-6)` whereas `_prediction_and_raw_diagnostics` multiplied the decoded predictions by the raw `obs_scale`. For any configured `observation_scale` below $10^{-6}$, the target normalization and prediction decoding differed by $\text{observation\_scale} / 10^{-6}$, causing the model to learn severely attenuated observation delta targets.

## Proposed Changes
1. Removed the `1e-6` floor in `ActionConditionedWorldModel.targets`, dividing delta and next observation targets directly by `obs_scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)`.
2. Added unit tests in `tests/test_action_conditioned_world_model.py` verifying that target normalization and prediction scaling compose consistently across scales down to $10^{-20}$.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_action_conditioned_world_model.py tests/test_world_model.py` (91/91 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/world_model.py tests/test_action_conditioned_world_model.py` (clean).

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -806,11 +806,10 @@ class ActionConditionedWorldModel:
         reward_arr = _float32_operand("reward", reward, ())
         discount_arr = _float32_operand("discount", discount, ())
         obs_scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        safe_scale = jnp.maximum(obs_scale, jnp.asarray(1e-6, dtype=jnp.float32))
         normalized_delta = jnp.where(
             self._config.predict_delta,
-            (next_obs - obs) / safe_scale,
-            next_obs / safe_scale,
+            (next_obs - obs) / obs_scale,
+            next_obs / obs_scale,
         )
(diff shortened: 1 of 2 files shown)


kenjohnscreates (changes requested):

The added round-trip test stops at normal scales and never crosses into the subnormal domain, so the suite stays green over the corner. PR #2403 pairs the same floor removal with a subnormal-rejecting validator (exact smallest normal still accepted) and measures clean across the whole matrix (posted on issue #2398). Also note this branch's library file is byte-identical to PRs #2747 and #2759.

lalalune:

PR #2541, opened 2026-08-25
Skip experiential-memory 0*inf utility decay

Kayvan-Zahiri:
## Summary
- Experiential memory decays eviction utilities each step with a scale that may be 0. IEEE 0*inf is NaN and the finite-state guard freezes writes.
- Skip that product before aging slots. When decay is 0, treat already-infinite utilities as discarded in finite-state and query eligibility checks.

## Test plan
- [x] `test_zero_utility_decay_does_not_multiply_inf_utilities`
- [x] `tests/test_experiential_memory.py` (54 passed)

### DIFF
--- a/alberta_framework/core/experiential_memory.py
+++ b/alberta_framework/core/experiential_memory.py
@@ -490,6 +490,18 @@ def _validate_config(config: ExperientialMemoryConfig) -> None:
         "recency_scale",
         _validated_config_float("recency_scale", config.recency_scale, positive=True),
     )
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled utility decay does not poison eviction."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
+def _utilities_for_finite_guard(decay: float, utilities: Array) -> Array:
+    """Treat poisoned eviction utility as discarded when decay zeroes it."""
+    if decay != 0.0:
+        return utilities
+    return jnp.where(jnp.isfinite(utilities), utilities, jnp.zeros_like(utilities))
+
+
 def _saturating_increment(value: Array) -> Array:
     maximum_minus_one = jnp.asarray(_INT32_MAX - 1, dtype=jnp.int32)
     return jnp.minimum(value, maximum_minus_one) + jnp.asarray(1, dtype=jnp.int32)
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):
## Sanitized alias stays where it should; one weakening worth documenting

Approving. The part of this change that usually goes wrong is handled correctly, and I want to record the check so it isn't repeated.

**The sanitized copy never enters arithmetic.** When a PR introduces a checked alias, the failure mode is that the *unsanitized* name keeps being used somewhere downstream. Here `utilities_for_check` appears only in the validity conjunction of `_state_is_valid` and in the `sane_rows` / finiteness conjunction of `_query_jit` — the raw `entries.utilities` never reaches query scoring.

There is one place the raw values do enter arithmetic, in the eviction score:

```python
eviction_utilities = jnp.where(current_entries.utility_available, current_entries.utilities, 0.0)
retention_score = utility_weight / weight_scale * eviction_utilities + ...
```

That is safe, but only because of ordering — `write()` dispatches to `self._write_advanced(self._advance(state), entry)`, and `_advance` has already replaced the utilities with exact zeros under `utility_decay == 0.0`. So eviction sees finite values. It's worth a comment at the eviction site, since the safety depends on a call-ordering invariant a reader can't see locally.

**The weakening, stated plainly.** `_utilities_for_finite_guard` returns a zero-filled copy whenever `utility_decay == 0.0`, which is an allowed config (`_validated_config_float("utility_decay", ..., lower=0, upper=1)`). So for those configs `_state_is_valid` no longer detects non-finite utilities, and a state corrupted by something other than the decay path is accepted rather than triggering abstention. `jnp.isfinite` is false for NaN as well as inf, so NaN corruption is masked the same way.

The blast radius is bounded — the masked values are zeroed by `_advance` before any arithmetic, and every other payload field is still checked — so this is a deliberate narrowing of a corruption detector rather than a hole. But it is a real reduction in what the dynamic invariant covers, and the docstring ("Treat poisoned eviction utility as discarded when decay zeroes it") reads as a description of intent rather than a warning that detection is being dropped. Worth saying so explicitly there.

lalalune:
Closing after implementation and regression review. The test injects infinite utility values, and the patch weakens both write validity and read/query row validity based on decay. Query does not first decay these utilities; configuration-dependent acceptance of corrupted stored metadata is not a sound recovery contract.

PR #2542, opened 2026-08-25
Skip online-gated state-builder 0*inf when unit gate is open

Kayvan-Zahiri:
## Summary
- OnlineGatedStateBuilder carries RTRL sensitivity with a `(1 - gate)` forget scale. When a unit gate is fully open, IEEE 0*inf is NaN and the finite-state guard freezes recurrence.
- Skip that product before writing sensitivity. When the forget scale is 0, treat already-infinite carried sensitivity as discarded in the source finite-state check.

## Test plan
- [x] `test_online_gated_unit_gate_one_does_not_multiply_inf_sensitivity`
- [x] `tests/test_state_builder.py` (81 passed)

### DIFF
--- a/alberta_framework/core/state_builder.py
+++ b/alberta_framework/core/state_builder.py
@@ -443,6 +443,11 @@ def _divide_by_positive_scale(values: Array, scale: Array) -> Array:
     return jnp.where(scale > 0.0, scaled, 0.0)
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed unit gate does not poison sensitivity."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def _scale_safe_clip_by_l2_norm(
     values: Array,
     clip: Array,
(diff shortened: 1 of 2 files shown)


kenjohnscreates (changes requested):
## The guard is correct, but its reachable domain is much narrower than the test suggests

The fix does what it says at exact saturation, and confining `checked_sensitivity` to the validity conjunction — never letting it into `new_sensitivity` — is the right shape. The candidate state is still validated in full, so nothing corrupt propagates.

What I want to flag is coverage. `forget_scale = (1.0 - gate)[:, None]` where `gate = sigmoid(...)`, so the `== 0.0` predicate only fires once the sigmoid saturates to exactly `1.0f`. Measured on this branch:

```
bias  16.00: 1-sigmoid = 1.192093e-07  exact_zero=False  fs*inf = inf
bias  16.50: 1-sigmoid = 1.192093e-07  exact_zero=False  fs*inf = inf
bias  16.63: 1-sigmoid = 1.192093e-07  exact_zero=False  fs*inf = inf
bias  16.64: 1-sigmoid = 0.000000e+00  exact_zero=True   fs*inf = nan
```

The cliff is at a gate logit of **16.64**. Below it the product is `tiny * inf = inf` rather than `NaN`, `_state_is_valid(candidate_state)` rejects the transaction, and the state stays poisoned — the same outcome as before the fix.

`test_online_gated_unit_gate_one_does_not_multiply_inf_sensitivity` pins `gate_bias = 90.0`, which is deep inside the working region. Re-running that exact scenario across biases:

| gate bias | `main` | this PR |
|---|---|---|
| 16.00 | FAIL | FAIL |
| 16.50 | FAIL | FAIL |
| 17.00 | FAIL | **PASS** |
| 18.00 | FAIL | **PASS** |
| 90.00 | FAIL | **PASS** |

At bias 16.5 the gate is already saturated to within `1.19e-07` of 1.0 — operationally "the unit gate is open", which is the condition in the PR title — and the fix is a no-op there.

To be clear: this is not a regression. `main` fails at every bias; the PR strictly improves the `>= 16.64` region. But the test as written reads as though the open-gate case is handled, when what is handled is the exact-float32-saturation case.

The test also builds its own `gate = jnp.ones(...)` to assert `raw` is non-finite, rather than asserting on the gate the builder actually computes — so it never checks that the configured bias reaches saturation at all.

### Suggested follow-ups

1. Add a near-saturation case (bias 16.5) and assert the intended post-condition. If the intent is that a wide-open gate always drops stale sensitivity, the predicate needs to be a tolerance on the gate rather than `forget_scale == 0.0`.
2. Assert on `jax.nn.sigmoid(gate_weights @ safe_event + gate_bias)` as computed, so the test fails if the chosen bias stops saturating.

lalalune:
Closing after implementation and regression review. The test manually corrupts the sensitivity matrix and the patch bypasses source-state validity on saturated gates. The existing update checks source and candidate state for finiteness, so this is an unrequested recovery policy for invalid state, not a demonstrated reachable update bug.

PR #2543, opened 2026-08-25
Skip action-conditioned world-model 0*inf error EMA decay

Kayvan-Zahiri:
## Summary
- ActionConditionedWorldModel decays model_error_ema with a scale that may be 0. IEEE 0*inf is NaN and the finite-state guard freezes updates.
- Skip that product before writing the diagnostic EMA. When decay is 0, treat already-infinite EMA values as discarded in finite-state checks.

## Test plan
- [x] `test_zero_error_decay_does_not_multiply_inf_model_error_ema`
- [x] `tests/test_action_conditioned_world_model.py` (69 passed)

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -282,6 +282,22 @@ def _saturating_increment(value: Array) -> Array:
     return jnp.minimum(value, jnp.asarray(_INT32_MAX - 1, dtype=jnp.int32)) + one
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled error EMA does not poison diagnostics."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
+def _model_error_ema_for_finite_guard(error_decay: float, model_error_ema: Array) -> Array:
+    """Treat poisoned error EMA as discarded when decay zeroes it."""
+    if error_decay != 0.0:
+        return model_error_ema
+    return jnp.where(
+        jnp.isfinite(model_error_ema),
+        model_error_ema,
+        jnp.zeros_like(model_error_ema),
+    )
+
+
 def _float32_operand(
     name: str,
     value: Array | float | int,
@@ -557,8 +573,13 @@ def _rollback_multi_head_result(
 def _action_world_model_state_is_valid(
     state: ActionConditionedWorldModelState,
     observation_dim: int,
+    *,
+    model_error_ema: Array | None = None,
 ) -> Bool[Array, ""]:
     """Accept either the intentional empty-bound sentinels or finite learned bounds."""
+    checked_error_ema = (
+        state.model_error_ema if model_error_ema is None else model_error_ema
+    )
     direct_structure_valid = (
         state.observation_min.shape == (observation_dim,)
         and state.observation_max.shape == (observation_dim,)
@@ -589,7 +610,7 @@ def _action_world_model_state_is_valid(
     return (
         direct_structure_valid
         & floating_tree_is_finite(state.learner_state)
-        & jnp.isfinite(state.model_error_ema)
+        & jnp.isfinite(checked_error_ema)
         & (state.step_count >= 0)
         & (finite_bounds | empty_bounds)
     )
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):
**Independent review — measured, not read.** Base `origin/main` @ `c9aba7b5`, PR head `7439f69c54a760f6745d7ee9ef084846bb8dcfd1`. macOS 15.2 arm64 (Darwin 25.2.0), CPython 3.13.5, jax/jaxlib 0.11.0, numpy 2.5.2, ruff 0.16.3. Isolated git worktrees with `PYTHONPATH` pinned per tree and `alberta_framework.__file__` asserted on each side before any number was recorded.

**Approving.** This is the strongest member of the `skip zero 0*inf` series, and I want to say why concretely rather than just stamping it.

### Measurements

| check | result |
|---|---|
| regression test non-vacuous | fails on pristine main — genuine failing-test-first |
| paired full file | main **68 passed** → PR **69 passed**, zero regressions |
| bitwise identity, enabled branch | **0 of 35,040** float32 elements differ across `error_decay ∈ {0.99, 0.8, 0.5, 0.999}`, 40 update steps each |
| completeness | no unguarded sibling `decay * state.X` sites remain |
| cross-PR overlap | 11 other open PRs touch `core/world_model.py`, but **none** touch `_action_world_model_state_is_valid`, `model_error_ema`, or `error_decay` — no conflict on the changed surface |
| CI | green |

### The reachability story here is real, unlike the rest of the series

I checked every sibling in this series and reported most of them as hardening for a legal-but-unexercised configuration. **This one is different: `error_decay=0.0` is set in shipped code**, twice, both constructing exactly this config class:

- `alberta_framework/benchmarks/dreamer_continual_development.py:193`
- `alberta_framework/benchmarks/dreamer_continual_development.py:359`

Driving that exact shipped config with a poisoned `model_error_ema`:

```
main    update_applied=False  next_error_ema=inf   finite=False
#2543   update_applied=True   next_error_ema=0.9152851104736328  finite=True
```

On `main` the channel is permanently stuck: `model_error_ema` stays `inf`, `_action_world_model_state_is_valid` keeps returning `False`, and every subsequent update is rejected forever even though `error_decay=0.0` means that EMA is discarded on the very next step. This PR recovers it. That is a live defect on a shipped benchmark configuration, not a hypothetical.

### The implementation picks the better of the two in-repo idioms

`_model_error_ema_for_finite_guard` gates on the **static Python float**, not a traced value:

```python
if error_decay != 0.0:
    return model_error_ema
```

so when the decay is enabled the function returns the original array and no `jnp.where` enters the graph at all. That mirrors `core/baseline_optimizers.py:193`, whose docstring states the intent exactly — *"Skip a disabled EMA without changing enabled optimizer graphs"* — and it is why the bitwise sweep above comes back at zero differing elements. Several siblings in this series use the unconditional 2-arg form instead; this is the stronger choice given the sealed `protocol_sha256` fingerprints in this repo.

`checked_error_ema` is also correctly confined to the validity conjunction and never reaches the arithmetic, matching `core/learning_signals.py:664-672`, `core/reward_model.py:318`, and `core/upgd_memory.py:1162`. The arithmetic path is separately safe via `_skip_zero_scale`, so the two halves compose rather than overlap.

One structural note in this PR's favour that does not apply to its siblings: the guarded scale is a **config constant**, so `error_decay == 0.0` is exact by construction. The sibling PRs #2546/#2547/#2548/#2549 guard a *runtime sigmoid gate*, where `1 - gate == 0.0` holds only at exact float32 saturation and the guard silently misses everything just below it. That failure mode is structurally impossible here.

### Reproduction

```bash
git worktree add /tmp/pr2543 7439f69c54a760f6745d7ee9ef084846bb8dcfd1
cd /tmp/pr2543 && PYTHONPATH=/tmp/pr2543 python -c "
import dataclasses, jax.numpy as jnp, jax.random as jr
from alberta_framework.core.world_model import *
m = ActionConditionedWorldModel(ActionConditionedWorldModelConfig(
    observation_dim=2, n_actions=2, hidden_sizes=(), step_size=0.05, sparsity=0.0, error_decay=0.0))
o = jnp.asarray([0.1,-0.2], jnp.float32); nxt = o + jnp.asarray([0.05,-0.03], jnp.float32)
w = m.update(m.init(jr.key(0)), o, jnp.int32(0), 0.3, 0.9, nxt)
p = dataclasses.replace(w.state, model_error_ema=jnp.asarray(jnp.inf))
r = m.update(p, o, jnp.int32(1), 0.3, 0.9, nxt)
print(bool(r.update_applied), float(r.state.model_error_ema))"
```

No changes requested. Clean, bit-safe, correctly scoped, and it fixes something reachable from shipped code.

lalalune:
Closing after implementation and regression review. The regression manually replaces a valid model-error EMA with infinity. The patch weakens the existing action-world-model state gate to accept that corruption when decay is zero; candidate-state finiteness already prevents such an EMA from being committed.

PR #2544, opened 2026-08-25
Skip latent world-model 0*inf surprise and collapse EMA decay

Kayvan-Zahiri:
## Summary
- LatentWorldModel decays surprise and collapse diagnostics with scales that may be 0. IEEE 0*inf is NaN and the finite-state guard freezes updates.
- Skip those products before writing the EMAs. When a decay is 0, treat already-infinite tracker values as discarded in finite-state checks.

## Test plan
- [x] `test_zero_surprise_decay_does_not_multiply_inf_diagnostic_emas`
- [x] `tests/test_latent_world_model.py` (57 passed)

### DIFF
--- a/alberta_framework/core/latent_world_model.py
+++ b/alberta_framework/core/latent_world_model.py
@@ -131,6 +131,18 @@ def _saturating_int32_increment(value: Array) -> Array:
     return jnp.where(value < maximum, value + jnp.asarray(1, dtype=jnp.int32), maximum)
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled EMA decay does not poison diagnostics."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
+def _recover_nonfinite_at_zero_scale(scale: float, value: Array) -> Array:
+    """Repair only non-finite history that a zero scale is configured to forget."""
+    if scale != 0.0:
+        return value
+    return jnp.where(jnp.isfinite(value), value, jnp.zeros_like(value))
+
+
 def _latent_direct_state_scalars(
     *,
     observation_dim: int,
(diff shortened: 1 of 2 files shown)


kenjohnscreates (changes requested):
**Independent review — measured, not read.** Base `origin/main` @ `c9aba7b5`, PR head `ed4f41c87fb41c9f45fa20583e72ce8eaacb7dee`. macOS 15.2 arm64 (Darwin 25.2.0), CPython 3.13.5, jax/jaxlib 0.11.0, numpy 2.5.2, ruff 0.16.3. Isolated worktrees, `PYTHONPATH` pinned per tree, `alberta_framework.__file__` asserted per side.

### What checks out

| check | result |
|---|---|
| regression test non-vacuous | fails on pristine main (`update_applied=False`) — genuine failing-test-first |
| paired full file | main **56 passed** → PR **57 passed**, zero regressions |
| bitwise identity, enabled branch | **0 of 3,600** float32 elements differ across `decays ∈ {0.99, 0.7, 0.5}`, 40 update steps each |
| completeness | no unguarded sibling `decay * state.X` sites remain (6 guarded refs) |

### Blocker — CI red: two new strict-mypy errors

Baseline discipline first, since this repo has a known macOS-only false positive: pristine `main` reports **1** error (`evaluation/bounded_elastic_matched_runner.py:1111`, `O_TMPFILE`). This head reports **3**. The two attributable to the PR are:

```
alberta_framework/core/latent_world_model.py:966: error: Value of type variable "_DataclassT" of "replace" cannot be "LatentWorldModelState"  [type-var]
alberta_framework/core/latent_world_model.py:977: error: Value of type variable "_DataclassT" of "replace" cannot be "LatentWorldModelState"  [type-var]
```

`dataclasses.replace` does not typecheck against these `chex` dataclasses. The repo's own merged idiom is the bound `.replace()` method with an explicit ignore — see `core/baseline_optimizers.py:722`, `core/upgd_memory.py:1148`, and `core/associative_memory.py` in the sibling PR #2550.

**Verified fix** — swap both call sites:

```python
checked_state = checked_state.replace(  # type: ignore[attr-defined]
    latent_mean_ema=_recover_nonfinite_at_zero_scale(0.0, state.latent_mean_ema),
    ...
)
```

Measured after that change: mypy back to **1 error** (the `O_TMPFILE` baseline only), `ruff check .` → `All checks passed!`, and `test_zero_surprise_decay_does_not_multiply_inf_diagnostic_emas` still passes. That is the whole blocker.

### A concern I raised against this PR and then disproved — recording it so nobody else spends the time

On a first static read, `checked_state` looks mis-scoped: it is consumed at exactly one place, `floating_tree_is_finite(checked_state)` in the `update_applied` conjunction, while the EMA arithmetic at 961–1010 still reads `state.latent_mean_ema` / `state.latent_var_ema` / `state.surprise_ema` / `state.prediction_error_ema` unsanitized. That reads like the sanitizer gating the *health check* rather than the *values used*.

**It is not a defect.** Every one of those unsanitized reads is wrapped in `_skip_zero_scale`, so at `decay == 0.0` each contributes exactly `0.0` and the poisoned history never reaches the arithmetic. Confining `checked_*` to the validity gate is precisely the merged idiom in this repo — `core/learning_signals.py:664-672`, `core/reward_model.py:318`, `core/upgd_memory.py:1162` all do exactly this and nothing more. The split is correct as written.

`surprise_decay` and `collapse_decay` are validated into `[0.0, 1.0)` (`latent_world_model.py:308-309`), so `0.0` is an explicitly legal configuration, not an impossible one. But no shipped config, CLI flag, or sweep grid sets either to zero — the defaults are `0.99` and there is no argparse or grid exposure anywhere in `alberta_framework/`. So this is **defensive hardening for a supported-but-unexercised configuration**, not a live NaN. That is still worth having; it just should not be described as fixing something users are hitting today.

### Reproduction

```bash
git worktree add /tmp/pr2544 ed4f41c87fb41c9f45fa20583e72ce8eaacb7dee
cd /tmp/pr2544 && PYTHONPATH=/tmp/pr2544 python -m mypy   # 3 errors; main gives 1
```

Requesting changes on the mypy blocker alone. The substance of the change is sound and the test is a real failing-test-first regression test.

lalalune:
Closing after implementation and regression review. The regression manually injects infinities into five EMA/state fields. The patch bypasses floating_tree_is_finite(state) for those fields, changing atomic rejection into silent repair without any reachable failing transition from a valid state.

PR #2545, opened 2026-08-25
Skip world-model-ensemble 0*inf residual variance decay

Kayvan-Zahiri:
## Summary
- WorldModelEnsemble decays residual variance proxies with a scale that may be 0. IEEE 0*inf is NaN and the transactional update freezes.
- Skip that product before writing the proxy. When decay is 0, treat already-infinite residual history as discarded for finite-state checks and learning-signal observation.

## Test plan
- [x] `test_zero_residual_variance_decay_does_not_multiply_inf_variances`
- [x] `tests/test_world_model_ensemble.py` (39 passed)

### DIFF
--- a/alberta_framework/core/world_model_ensemble.py
+++ b/alberta_framework/core/world_model_ensemble.py
@@ -258,6 +258,23 @@ def _saturating_int32_sum(values: Array) -> Array:
     )
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled residual EMA does not poison the proxy."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
+def _residual_variances_for_finite_guard(
+    decay: float,
+    residual_variances: Array,
+    *,
+    floor: float,
+) -> Array:
+    """Treat residual-proxy history as discarded when decay zeroes it."""
+    if decay != 0.0:
+        return residual_variances
+    return jnp.full_like(residual_variances, floor)
+
+
 @dataclasses.dataclass(frozen=True)
 class WorldModelEnsembleConfig:
     """Static ensemble, bootstrap, and residual-proxy contract.
@@ -1208,7 +1225,17 @@ class WorldModelEnsemble:
             & jnp.where(step_count == 0, pristine_bounds, learned_bounds)
         )
 
-    def _state_valid(self, state: WorldModelEnsembleState) -> Array:
+    def _state_valid(
+        self,
+        state: WorldModelEnsembleState,
+        *,
+        residual_variances: Array | None = None,
+    ) -> Array:
+        checked_residuals = (
+            state.residual_variances
+            if residual_variances is None
+            else residual_variances
+        )
         member_valid = []
         member_counts_match = []
         for index, member_state in enumerate(state.member_states):
@@ -1245,9 +1272,9 @@ class WorldModelEnsemble:
             )
             & jnp.all(jnp.stack(member_valid))
             & jnp.all(jnp.stack(member_counts_match))
-            & jnp.all(jnp.isfinite(state.residual_variances))
-            & jnp.all(state.residual_variances >= self._config.residual_variance_floor)
-            & jnp.all(state.residual_variances <= signal_cfg.max_predicted_variance)
+            & jnp.all(jnp.isfinite(checked_residuals))
+            & jnp.all(checked_residuals >= self._config.residual_variance_floor)
+            & jnp.all(checked_residuals <= signal_cfg.max_predicted_variance)
             & self._signal_state_valid(signal_state)
             & signal_bounds_valid
             & (signal_state.step_count == state.event_count)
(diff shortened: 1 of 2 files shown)


kenjohnscreates (changes requested):
**Independent review — measured, not read.** Base `origin/main` @ `c9aba7b5`, PR head `48d5d03b1adb3d38884f676ec0bca2f5f529a817`. macOS 15.2 arm64 (Darwin 25.2.0), CPython 3.13.5, jax/jaxlib 0.11.0, numpy 2.5.2, ruff 0.16.3. Isolated worktrees, `PYTHONPATH` pinned per tree, `alberta_framework.__file__` asserted per side.

### What checks out

| check | result |
|---|---|
| regression test non-vacuous | fails on pristine main (`state_valid=False`, `applied=False`) — genuine failing-test-first |
| paired full file | main **38 passed** → PR **39 passed**, zero regressions |
| bitwise identity, enabled branch | **0 of 390** float32 elements differ across `residual_variance_decay ∈ {0.99, 0.8, 0.5}` (`residual_variances` + full `signal_state`, 30 events each) |
| ruff / mypy | clean; mypy shows only the pre-existing macOS `O_TMPFILE` baseline |
| cross-PR overlap | sole open PR touching this file apart from the tree-wide #2272/#2275 — the scope findings below are fair |

### Finding 1 (blocking) — `checked_residuals` leaks out of the validity gate into arithmetic, and changes behaviour for *healthy* data

Everywhere this idiom is already merged, the sanitized copy is used **only** in the validity conjunction and never in a computation:

- `core/learning_signals.py:664-672` — `checked_fast_loss` / `checked_slow_loss` / `checked_change` appear only inside `state_valid`
- `core/reward_model.py:318` — `checked_error_ema` appears only inside `source_finite`
- `core/upgd_memory.py:1162` — `checked_state` appears only inside `_state_is_valid(...)`

This PR keeps that contract in `_state_valid(...)`, but additionally substitutes the sanitized array into a real computation:

```python
candidate_signal_state, raw_signals = self._signals.observe(
    state.signal_state,
    prediction.member_raw_predictions,
    checked_residuals,     # <-- was state.residual_variances
    targets,
    observed_loss,
)
```

Because `_residual_variances_for_finite_guard` returns `jnp.full_like(residual_variances, floor)` — a blanket replacement, not a repair of non-finite entries — at `residual_variance_decay == 0.0` the learning-signal estimator stops seeing the real residual variances and sees the constant `residual_variance_floor` instead. That happens even when the incoming values are perfectly finite and far from the floor.

Measured on your head vs main, `decay=0.0`, `residual_variances` set to a finite `3.0` (the floor is `1e-6`), one `update()`:

```
main   signal_state float32 sha256 = e7637db9102e3bd20b87a7d33fa874fc   applied=True
#2545  signal_state float32 sha256 = d5a3ce17b236a9f90a4170144866bdf3   applied=True
```

Same inputs, both applied, different signal state. That is a behaviour change unrelated to `0 * inf`, on data that was never poisoned, and no test covers it — `test_zero_residual_variance_decay_does_not_multiply_inf_variances` only ever feeds `jnp.inf`.

The `_skip_zero_scale` on the EMA line already makes the arithmetic safe on its own. Keeping `checked_residuals` confined to `_state_valid(...)` and passing `state.residual_variances` to `observe()` fixes this and restores parity with the three merged precedents.

### Finding 2 (blocking) — the guard added to `replay_update` cannot take effect

`replay_update` computes `checked_residuals` and threads it into `_state_valid(...)`, but `candidate_state` is still built from the raw array (`world_model_ensemble.py:1622`, `residual_variances=state.residual_variances`), and `candidate_state_valid = self._state_valid(candidate_state)` is called **without** the override. Since `applied = predictions_valid & member_updates_valid & candidate_state_valid`, the poisoned value is re-rejected one line later.

Measured, `decay=0.0`, poisoned residuals, `replay_update`:

```
main    state_valid=False  candidate_state_valid=False  applied=False
#2545   state_valid=True   candidate_state_valid=False  applied=False
```

The guard flips a diagnostic flag and nothing else — the update is still rejected on both sides. Either thread `checked_residuals` into `candidate_state` too, or drop the `replay_update` half and say the guard is `update()`-only. As written it reads as protection that is not there, which is the more expensive kind of dead code. No test exercises `replay_update` on this path.

`residual_variance_decay` is validated into `[0.0, 1.0)` (`world_model_ensemble.py:306-311`), so `0.0` is explicitly legal. No shipped config, CLI flag, or sweep grid sets it to zero — the default is `0.99` and there is no argparse or grid exposure in `alberta_framework/`. So this is hardening for a supported-but-unexercised configuration. Worth having; not a live NaN.

One consequence worth weighing: `test_corrupt_dynamic_state_is_an_exact_atomic_noop` pins that a `NaN` in `residual_variances` must produce `state_valid == False`. It still passes because the default `_config()` uses `decay=0.8`, but at `decay=0.0` the blanket `full_like(..., floor)` also disables the `>= floor` and `<= max_predicted_variance` range checks, not just the finiteness check. That is a wider relaxation than `0 * inf` requires.

### Reproduction

```bash
git worktree add /tmp/pr2545 48d5d03b1adb3d38884f676ec0bca2f5f529a817
cd /tmp/pr2545 && PYTHONPATH=/tmp/pr2545 python -m pytest tests/test_world_model_ensemble.py -q -p no:randomly
```

Requesting changes on Findings 1 and 2. The core `_skip_zero_scale` edit on the EMA line is correct and bit-safe; it is the `checked_residuals` plumbing around it that overreaches in `update()` and under-reaches in `replay_update()`.

lalalune:
Closing after implementation and regression review. The regression manually injects infinite residual variances. The patch substitutes a floor for state validation and for uncertainty observation, hiding corrupted residual history and changing calibration input. Existing candidate validation prevents committing that history.

PR #2547, opened 2026-08-25
fix(prototype): skip zero-gate 0*inf in GRU perception update

Kayvan-Zahiri:
## Summary
- Add `_skip_zero_scale` in Prototype GRU `_gru_step` so saturated update/reset gates do not multiply poisoned hidden states via `(1-z)*h` or `r*h`.
- Sanitize non-finite hidden before gate logits so zero-weight matmuls do not produce `0*inf` NaNs in sigmoid inputs.
- Regression test: unit gate saturated with `hidden=inf` yields finite `tanh(b_h)` candidate state.

### DIFF
--- a/alberta_framework/core/prototype_agent.py
+++ b/alberta_framework/core/prototype_agent.py
@@ -479,6 +479,11 @@ def _glorot_uniform(key: Array, shape: tuple[int, int]) -> Array:
     return jr.uniform(key, shape, dtype=jnp.float32, minval=-limit, maxval=limit)
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed GRU gate does not poison hidden state."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def _init_gru_state(cfg: GRUPerceptionConfig, key: Array) -> GRUPerceptionState:
     """Glorot-uniform weight init + zero hidden state."""
     keys = jr.split(key, 6)
@@ -507,10 +512,11 @@ def _gru_step(
     ``[obs, new_hidden]`` as the augmented observation.
     """
     h = gru.hidden
-    z = jax.nn.sigmoid(gru.W_z @ obs + gru.U_z @ h + gru.b_z)
-    r = jax.nn.sigmoid(gru.W_r @ obs + gru.U_r @ h + gru.b_r)
-    h_tilde = jnp.tanh(gru.W_h @ obs + gru.U_h @ (r * h) + gru.b_h)
-    new_h = (1.0 - z) * h + z * h_tilde
+    safe_h = jnp.where(jnp.isfinite(h), h, jnp.zeros_like(h))
+    z = jax.nn.sigmoid(gru.W_z @ obs + gru.U_z @ safe_h + gru.b_z)
+    r = jax.nn.sigmoid(gru.W_r @ obs + gru.U_r @ safe_h + gru.b_r)
+    h_tilde = jnp.tanh(gru.W_h @ obs + gru.U_h @ _skip_zero_scale(r, safe_h) + gru.b_h)
+    new_h = _skip_zero_scale(1.0 - z, h) + _skip_zero_scale(z, h_tilde)
     new_gru = cast(GRUPerceptionState, gru.replace(hidden=new_h))
     return new_gru, jnp.concatenate([obs, new_h], axis=0)
 
--- a/tests/test_prototype_agent.py
+++ b/tests/test_prototype_agent.py
@@ -1239,6 +1239,36 @@ class TestGRUPerceptionStateInit:
 
 
 class TestGRUPerceptionUpdate:
+    def test_unit_update_gate_one_does_not_multiply_inf_hidden(self) -> None:
+        """A fully open update gate must forget poisoned hidden state without NaN."""
+        from alberta_framework.core.prototype_agent import (
+            GRUPerceptionConfig,
+            _gru_step,
+            _init_gru_state,
+        )
+
+        gru = _init_gru_state(GRUPerceptionConfig(observation_dim=2, hidden_dim=2), jr.key(0))
+        gru = gru.replace(
+            hidden=jnp.full((2,), jnp.inf, dtype=jnp.float32),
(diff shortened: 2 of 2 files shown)


kenjohnscreates (changes requested):
**Independent review — measured, not read.** Base `origin/main` @ `c9aba7b5`, PR head `75407cd77d553200848457de939ff24bbea9e9d1`. macOS 15.2 arm64 (Darwin 25.2.0), CPython 3.13.5, jax/jaxlib 0.11.0, numpy 2.5.2, ruff 0.16.3. Isolated git worktrees, `PYTHONPATH` pinned per tree, `alberta_framework.__file__` asserted on each side before recording any number.

### What checks out

| check | result |
|---|---|
| regression test non-vacuous | fails on pristine main with `hidden='Nonfinite'` — genuine failing-test-first |
| paired full file | main **102 passed** → PR **103 passed**, zero regressions |
| bitwise identity, finite branch | **0 of 3,960** float32 elements differ (60 random seeds × 6 steps × 7 magnitude scales) |
| completeness | no unguarded sibling `scale * state.X` sites remain |
| CI | green |

### Blocker — the guard only fires at *exact* float32 saturation

`new_h` blends the **raw** `h`, not the `safe_h` alias built one line above:

```python
safe_h = jnp.where(jnp.isfinite(h), h, jnp.zeros_like(h))
z = jax.nn.sigmoid(gru.W_z @ obs + gru.U_z @ safe_h + gru.b_z)          # sanitized
r = jax.nn.sigmoid(gru.W_r @ obs + gru.U_r @ safe_h + gru.b_r)          # sanitized
h_tilde = jnp.tanh(gru.W_h @ obs + gru.U_h @ _skip_zero_scale(r, safe_h) + gru.b_h)   # sanitized
new_h = _skip_zero_scale(1.0 - z, h) + _skip_zero_scale(z, h_tilde)     # <-- raw `h`
```

`_skip_zero_scale` rescues only `1.0 - z == 0.0` exactly. Your test pins `b_z = 90.0`, where `sigmoid` saturates to exactly `1.0f`, so the guard fires. Just below saturation it does not. Measured **on your own head**:

```
bias= 90.0  sigmoid=1.0000000000  1-z=0.000e+00  new_h=[ 0.24491866 -0.24491866]  finite=True
bias= 20.0  sigmoid=1.0000000000  1-z=0.000e+00  new_h=[ 0.24491866 -0.24491866]  finite=True
bias= 15.0  sigmoid=0.9999996424  1-z=3.576e-07  new_h=[inf inf]                  finite=False
bias= 10.0  sigmoid=0.9999545813  1-z=4.542e-05  new_h=[inf inf]                  finite=False
bias=  5.0  sigmoid=0.9933071733  1-z=6.693e-03  new_h=[inf inf]                  finite=False
bias=  3.0  sigmoid=0.9525741339  1-z=4.743e-02  new_h=[inf inf]                  finite=False
```

The docstring claim — *"A fully open update gate must forget poisoned hidden state without NaN"* — is true only at the exactly-saturating point the test samples. Everywhere else `inf` still reaches `new_h`, and from there `jnp.concatenate([obs, new_h])` poisons the augmented observation handed to the OaK/STOMP path.

**Verified one-token fix:**

```python
new_h = _skip_zero_scale(1.0 - z, safe_h) + _skip_zero_scale(z, h_tilde)
```

All six biases then return finite and degrade smoothly (`0.24491866` → `0.23330317`), your new test still passes, and the full file is **103 passed**.

This is the same defect, token for token, as #2548 in `core/recurrent_latent_world_model_ensemble.py`, and the same shape as #2549. Fixing them as a set would be cheaper than three separate rounds.

### One non-blocking note, stated precisely so it is not over-read

Gradients through `_gru_step` **do** move: differentiating `sum(new_h**2) + sum(aug**2)` over 40 seeded parameter sets with entirely finite inputs, **40 of 4,000** elements differ numerically, max relative difference `4.591e-07` (~4 float32 ULP), with 102 further elements differing only by signed zero. That comes from `safe_h` re-routing the `h` cotangent path, not from any `jnp.where` NaN trap.

**This is currently unreachable and is not a defect.** `_gru_step` is not inside any grad transform on `main`: the only two in the file are `jax.value_and_grad(loss_fn)` at `prototype_agent.py:3865` and `:3940`, both of which differentiate w.r.t. `representation` alone, and neither `loss_fn` body calls `_gru_step` — the GRU runs upstream to *produce* the representation. Its three call sites (`prime`, `restart_builder`, `restart_gru`) are outside any transform. I am recording it only so that whoever later backprops through the perception GRU knows this PR moved that surface by a few ULP; nothing needs to change now.

### Reproduction

```bash
git worktree add /tmp/pr2547 75407cd77d553200848457de939ff24bbea9e9d1
cd /tmp/pr2547 && PYTHONPATH=/tmp/pr2547 python -c "
from alberta_framework.core.prototype_agent import GRUPerceptionConfig, _gru_step, _init_gru_state
for b in (90.0, 15.0, 5.0):
    g = _init_gru_state(GRUPerceptionConfig(observation_dim=2, hidden_dim=2), jr.key(0))
    g = g.replace(hidden=jnp.full((2,), jnp.inf, jnp.float32),
        W_z=jnp.zeros_like(g.W_z), U_z=jnp.zeros_like(g.U_z), b_z=jnp.full((2,), b, jnp.float32),
        W_r=jnp.zeros_like(g.W_r), U_r=jnp.zeros_like(g.U_r), b_r=jnp.full((2,), -b, jnp.float32),
        W_h=jnp.zeros_like(g.W_h), U_h=jnp.zeros_like(g.U_h),
        b_h=jnp.asarray([0.25,-0.25], jnp.float32))
    ng, _ = _gru_step(g, jnp.zeros((2,), jnp.float32))
    print(b, ng.hidden, bool(jnp.all(jnp.isfinite(ng.hidden))))"
```

Requesting changes on the raw-`h` blend only. The helper and the sanitize-then-use approach are correct and match the idiom already merged in `core/off_policy_td.py`, `core/learning_signals.py`, `core/reward_model.py` and `core/baseline_optimizers.py`.

lalalune:
Closing after implementation and regression review. The regression injects infinite GRU hidden state and calls a private helper. Reachable GRU hidden values are bounded by tanh and sigmoid interpolation; substituting zero hidden values for gate computation silently repairs an invalid state rather than fixing a valid trajectory.

PR #2548, opened 2026-08-25
fix(recurrent-latent): skip zero-gate 0*inf in member GRU predict

Kayvan-Zahiri:
## Summary
- Add `_skip_zero_scale` in recurrent latent ensemble `_member_predict` so saturated update/reset gates do not multiply poisoned hidden via `(1-update_gate)*hidden` or `reset_gate*hidden`.
- Sanitize non-finite hidden before gate logits so zero-weight matmuls do not produce `0*inf` NaNs in sigmoid inputs.
- Regression test: saturated gates with `hidden=inf` yield finite `tanh(candidate_bias)`.

### DIFF
--- a/alberta_framework/core/recurrent_latent_world_model_ensemble.py
+++ b/alberta_framework/core/recurrent_latent_world_model_ensemble.py
@@ -227,6 +227,11 @@ def _saturating_increment(value: Array, maximum: int) -> Array:
     return jnp.minimum(jnp.maximum(value, 0), bound - 1) + 1
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed GRU gate does not poison hidden state."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def _static_signature(tree: Any) -> tuple[Any, tuple[tuple[tuple[int, ...], Any], ...]]:
     leaves, structure = jax.tree_util.tree_flatten(tree)
     return structure, tuple((jnp.asarray(leaf).shape, jnp.asarray(leaf).dtype) for leaf in leaves)
@@ -1112,16 +1117,22 @@ class RecurrentLatentWorldModelEnsemble:
         cfg = self._config
         action_one_hot = jax.nn.one_hot(action, cfg.n_actions, dtype=jnp.float32)
         recurrent_input = jnp.concatenate((observation, action_one_hot), axis=0)
-        gate_input = jnp.concatenate((recurrent_input, hidden), axis=0)
+        safe_hidden = jnp.where(jnp.isfinite(hidden), hidden, jnp.zeros_like(hidden))
+        gate_input = jnp.concatenate((recurrent_input, safe_hidden), axis=0)
         reset_and_update = jax.nn.sigmoid(
             parameters.gate_kernel @ gate_input + parameters.gate_bias
         )
         reset_gate, update_gate = jnp.split(reset_and_update, 2, axis=0)
-        candidate_input = jnp.concatenate((recurrent_input, reset_gate * hidden), axis=0)
+        candidate_input = jnp.concatenate(
+            (recurrent_input, _skip_zero_scale(reset_gate, safe_hidden)),
+            axis=0,
+        )
         candidate = jnp.tanh(
             parameters.candidate_kernel @ candidate_input + parameters.candidate_bias
         )
-        next_hidden = (1.0 - update_gate) * hidden + update_gate * candidate
+        next_hidden = _skip_zero_scale(1.0 - update_gate, hidden) + _skip_zero_scale(
+            update_gate, candidate
+        )
 
         mean_logits = parameters.mean_kernel @ next_hidden + parameters.mean_bias
         variance_logits = parameters.variance_kernel @ next_hidden + parameters.variance_bias
--- a/tests/test_recurrent_latent_world_model_ensemble.py
+++ b/tests/test_recurrent_latent_world_model_ensemble.py
@@ -21,6 +21,7 @@ from alberta_framework.core.recurrent_latent_world_model_ensemble import (
     EVIDENCE_LEVEL,
     SCIENTIFIC_PROMOTION_ALLOWED,
     RecurrentLatentDecisionCache,
+    RecurrentLatentMemberParameters,
     RecurrentLatentTransitionRecord,
     RecurrentLatentWorldModelEnsemble,
     RecurrentLatentWorldModelEnsembleConfig,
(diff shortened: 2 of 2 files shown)


kenjohnscreates (changes requested):
**Independent review — measured, not read.** Base `origin/main` @ `c9aba7b5`, PR head `acdc5cc2b53c3b7fe1ecff1693e2c712c6270833`. macOS 15.2 arm64 (Darwin 25.2.0), CPython 3.13.5, jax/jaxlib 0.11.0, numpy 2.5.2, ruff 0.16.3. Each side ran in its own git worktree with `PYTHONPATH` pinned per tree and `alberta_framework.__file__` asserted before any number was recorded (the editable install otherwise resolves both trees to the same path and silently invalidates the comparison).

### What checks out

| check | result |
|---|---|
| regression test non-vacuous | **all 3 assertions fail on pristine main** — genuine failing-test-first |
| paired full file | main **82 passed** → PR **83 passed**, zero regressions |
| bitwise identity, finite branch | **0 of 1,080** float32 elements differ (60 random parameter sets × 7 magnitude scales) |
| gradient `d/d(observation)` (repr. gradient, L1547) | **bit-identical** |
| gradient `d/d(parameters)` (member update, L1503) | **0 of 3,300 numerically differing**; bit patterns differ only by signed zero (`-0.0` → `+0.0`), max relative difference `0.000e+00` — numerically inert |
| completeness | no unguarded sibling `scale * state.X` sites remain in the file |
| cross-PR overlap | sole open PR touching this file apart from the tree-wide #2272/#2275 — scope findings below are fair |
| mypy strict | 1 error = the pre-existing macOS-only `O_TMPFILE` baseline, **not attributable to this PR** |

I want to flag the gradient result explicitly because it is easy to get wrong: `_member_predict` **is** inside two `jax.value_and_grad` transforms (via `_member_nll` at L1503 and L1547), so a cotangent change here would be reachable and would affect learning. I measured it rather than assuming, and it is clean — the only movement is signed zero. The `jnp.where` "double-where" NaN-cotangent trap does not apply.

### Blocker 1 — the guard only fires at *exact* float32 saturation, and the test is built on that single point

`next_hidden` blends the **raw** `hidden`, not the `safe_hidden` alias built two lines above:

```python
safe_hidden = jnp.where(jnp.isfinite(hidden), hidden, jnp.zeros_like(hidden))
gate_input  = jnp.concatenate((recurrent_input, safe_hidden), axis=0)   # sanitized
candidate_input = ... _skip_zero_scale(reset_gate, safe_hidden) ...     # sanitized
next_hidden = _skip_zero_scale(1.0 - update_gate, hidden) + ...         # <-- raw `hidden`
```

`_skip_zero_scale` only rescues `1.0 - update_gate == 0.0` exactly. Your test uses `gate_bias = ±90.0`, where `sigmoid` saturates to exactly `1.0f`, so `1 - z` is exactly `0.0` and the guard fires. One step below saturation it does not. Measured **on your own head**:

```
bias= 90.0  sigmoid=1.0000000000  1-z=0.000e+00  next_hidden=[ 0.24491866 -0.24491866]  finite=True
bias= 20.0  sigmoid=1.0000000000  1-z=0.000e+00  next_hidden=[ 0.24491866 -0.24491866]  finite=True
bias= 15.0  sigmoid=0.9999996424  1-z=3.576e-07  next_hidden=[inf inf]                  finite=False
bias= 10.0  sigmoid=0.9999545813  1-z=4.542e-05  next_hidden=[inf inf]                  finite=False
bias=  5.0  sigmoid=0.9933071733  1-z=6.693e-03  next_hidden=[inf inf]                  finite=False
bias=  3.0  sigmoid=0.9525741339  1-z=4.743e-02  next_hidden=[inf inf]                  finite=False
```

So the stated property — *"Saturated update/reset gates must not turn poisoned hidden state into NaN"* — holds only on the measure-zero set where the gate saturates exactly, which is the only point your test samples. A near-saturated gate is the common case, and there `inf` still propagates into `next_hidden` → `mean_logits`/`variance_logits`.

**Verified one-token fix** — blend the alias you already computed:

```python
next_hidden = _skip_zero_scale(1.0 - update_gate, safe_hidden) + _skip_zero_scale(update_gate, candidate)
```

With that change all six biases return finite (`[0.24491866 …]` → `[0.23330317 …]`, degrading smoothly), your new test still passes, and the full file is **83 passed**. The gradient shas are unchanged from the PR head.

This is the same shape as the `safe_hidden`-declared-then-unused pattern in the sibling PR #2549, and identical to #2547 in `core/prototype_agent.py` (`new_h = _skip_zero_scale(1.0 - z, h)` where `safe_h` exists). Worth fixing as a set.

### Blocker 2 — CI is red on `lint`, one line

```
E501 Line too long (102 > 100)
 --> tests/test_recurrent_latent_world_model_ensemble.py:342:101
```

Wrapping that `np.testing.assert_allclose(...)` across three lines clears `ruff check .` with `All checks passed!`.

### Reproduction

```bash
git worktree add /tmp/pr2548 acdc5cc2b53c3b7fe1ecff1693e2c712c6270833
cd /tmp/pr2548 && PYTHONPATH=/tmp/pr2548 python -c "
import jax, jax.numpy as jnp
from alberta_framework.core.recurrent_latent_world_model_ensemble import *
m = RecurrentLatentWorldModelEnsemble(RecurrentLatentWorldModelEnsembleConfig(
    observation_dim=2, n_actions=2, latent_dim=2, ensemble_size=3, learning_rate=0.01,
    bootstrap_probability=0.8, uncertainty_warmup_steps=1, max_updates=8))
c=m.config; L=c.latent_dim; J=c.recurrent_input_dim+L; T=c.target_dim
for b in (90.0, 15.0, 5.0):
    p = RecurrentLatentMemberParameters(
        gate_kernel=jnp.zeros((2*L,J),jnp.float32),
        gate_bias=jnp.concatenate((jnp.full((L,),-b,jnp.float32), jnp.full((L,),b,jnp.float32))),
        candidate_kernel=jnp.zeros((L,J),jnp.float32), candidate_bias=jnp.asarray((0.25,-0.25),jnp.float32),
        mean_kernel=jnp.zeros((T,L),jnp.float32), mean_bias=jnp.zeros((T,),jnp.float32),
        variance_kernel=jnp.zeros((T,L),jnp.float32),
        variance_bias=jnp.full((T,), c.initial_variance_bias_magnitude, jnp.float32))
    nh,*_ = m._member_predict(p, jnp.full((L,), jnp.inf, jnp.float32),
                              jnp.zeros((c.observation_dim,),jnp.float32), jnp.asarray(0,jnp.int32))
    print(b, nh, bool(jnp.all(jnp.isfinite(nh))))"
```

Requesting changes on Blocker 1 (correctness) and Blocker 2 (CI). The approach and the guard helper are right and match the merged idiom already in `core/off_policy_td.py`, `core/learning_signals.py`, `core/reward_model.py` and `core/baseline_optimizers.py` — it is only the one raw-`hidden` token that undercuts it.

kenjohnscreates:
This receipt covers a single batched review pass over five PRs in the `skip zero 0*inf` series, not just this one. Posting it once here, on the most substantial of the five, rather than stamping it on each — one run of compute, one artifact.

The other four reviewed in the same bracket:

- #2547 — `core/prototype_agent.py` — changes requested (same defect as this PR, token for token)
- #2544 — `core/latent_world_model.py` — changes requested (strict-mypy blocker)
- #2545 — `core/world_model_ensemble.py` — changes requested (sanitized copy leaks into arithmetic; `replay_update` guard inert)
- #2550 — `core/associative_memory.py` — commented, approve-shaped (clean; one inert branch)

All five were measured on macOS 15.2 arm64 against `origin/main` @ `c9aba7b5` in isolated worktrees with `PYTHONPATH` pinned per tree.

Compute receipt: 11587890 project-attributed tokens (exact; device-signed, locally reported)
— [asi]

lalalune:
Closing after implementation and regression review. The regression injects infinite recurrent hidden state into a private predictor. Reachable hidden state is bounded by tanh and sigmoid interpolation; silently substituting zeros in the gate input is not a justified recovery policy.

PR #2535, opened 2026-08-25
fix(benchmarks): derive integer allowlists from numpy dtype codes

Kayvan-Zahiri:
Fixes #2295

## Summary
In `alberta_framework/benchmarks/ipmnist_gradual.py` and `alberta_framework/benchmarks/causal_map_forager.py`:
`_NUMPY_INTEGER_TYPES` and `_EXACT_NUMPY_INTEGER_TYPES` were hardcoded with fixed-width type names (`np.int8`, `np.int16`, `np.int32`, `np.int64`, etc.). Because numpy exposes 5 signed C integer types and only 4 fixed-width aliases, hardcoded names omit valid platform-specific integer types like `np.intc` and `np.uintc`, causing configurations to reject valid integer arguments with `ValueError`.

## Proposed Changes
1. Replaced hardcoded type lists in both modules with `tuple(np.dtype(code).type for code in ("b", "B", "h", "H", "i", "I", "l", "L", "q", "Q"))`.
2. Added parameterized tests across all 10 integer type codes in `tests/test_ipmnist_gradual.py` and `tests/test_causal_map_forager.py`.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_ipmnist_gradual.py` (30/30 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/benchmarks/ipmnist_gradual.py alberta_framework/benchmarks/causal_map_forager.py tests/test_ipmnist_gradual.py tests/test_causal_map_forager.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/benchmarks/ipmnist_gradual.py alberta_framework/benchmarks/causal_map_forager.py` (clean).

PR #2549, opened 2026-08-26
fix(forager): skip zero-gate 0*inf in recurrent GRU reservoir

Kayvan-Zahiri:
## Summary
- Add `_skip_zero_scale` in Forager `_augment_with_recurrent_features` so saturated update/reset gates do not multiply poisoned hidden via `(1-update)*hidden` or `reset*hidden`.
- Sanitize non-finite hidden before gate logits so zero-weight recurrent matmuls do not produce `0*inf` NaNs in sigmoid inputs.
- Regression test: saturated gates with `hidden=inf` yield finite `tanh(candidate_bias)`.

### DIFF
--- a/alberta_framework/benchmarks/forager.py
+++ b/alberta_framework/benchmarks/forager.py
@@ -1129,6 +1129,11 @@ def _init_forager_recurrent_state(
     )
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed GRU gate does not poison hidden state."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def _augment_with_recurrent_features(
     features: Array,
     recurrent_state: ForagerRecurrentState,
@@ -1139,6 +1144,7 @@ def _augment_with_recurrent_features(
         return recurrent_state, features
 
     hidden = recurrent_state.hidden
+    safe_hidden = jnp.where(jnp.isfinite(hidden), hidden, jnp.zeros_like(hidden))
     input_terms = jnp.einsum(
         "ghi,i->gh",
         recurrent_state.input_kernel,
@@ -1146,21 +1152,21 @@ def _augment_with_recurrent_features(
     )
     update = jax.nn.sigmoid(
         input_terms[0]
-        + recurrent_state.recurrent_kernel[0] @ hidden
+        + recurrent_state.recurrent_kernel[0] @ safe_hidden
         + recurrent_state.bias[0]
     )
     reset = jax.nn.sigmoid(
         input_terms[1]
-        + recurrent_state.recurrent_kernel[1] @ hidden
+        + recurrent_state.recurrent_kernel[1] @ safe_hidden
         + recurrent_state.bias[1]
     )
     candidate = jnp.tanh(
         input_terms[2]
-        + recurrent_state.recurrent_kernel[2] @ (reset * hidden)
+        + recurrent_state.recurrent_kernel[2] @ _skip_zero_scale(reset, safe_hidden)
         + recurrent_state.bias[2]
     )
     next_hidden = jax.lax.stop_gradient(
-        (1.0 - update) * hidden + update * candidate
+        _skip_zero_scale(1.0 - update, hidden) + _skip_zero_scale(update, candidate)
     )
     next_state = recurrent_state._replace(hidden=next_hidden)
     return next_state, jnp.concatenate((features, next_hidden), axis=0)
--- a/tests/test_forager_recurrent.py
+++ b/tests/test_forager_recurrent.py
@@ -20,6 +20,7 @@ from alberta_framework.benchmarks.forager import (
     ForagerBenchmarkConfig,
     ForagerEnvConfig,
     ForagerFeatureConfig,
+    ForagerRecurrentState,
     _augment_with_recurrent_features,
     _init_forager_recurrent_state,
     _recurrent_key,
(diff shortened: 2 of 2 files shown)


kenjohnscreates (changes requested):
## Disposition: REQUEST CHANGES — `safe_hidden` is computed but not used where it matters, demonstrated on this head

Independently reproduced on **macOS arm64 / CPython 3.13.5**, at exact head `68fe9734a89519782cf2061010ec5b94835caa7c` against base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, in isolated worktrees with `PYTHONPATH` pinned per tree (module resolution verified on both sides).

### What measures clean

- **The regression test is genuine.** Copied onto pristine `main` it fails against the unfixed library with real `nan`:
  ```
  FAILED tests/test_forager_recurrent.py::test_saturated_update_gate_does_not_multiply_inf_hidden
  E  assert np.all(np.isfinite(...)) ... hidden=Array([nan, nan], dtype=float32)
  1 failed, 14 deselected
  ```
- **Paired full-file runs**, `-q -p no:randomly`: `main` **14 passed** → this head **15 passed**. Zero regressions.
- **Bit-exactness on the untouched branch.** `_augment_with_recurrent_features` driven with finite random state over four trials, comparing raw bytes of both returned arrays:
  ```
  trial0 d8632c8eb93c8d4b / 5d195b5708cb3318      trial2 b20a96d36b0d6570 / 2c1efdd87773c553
  trial1 14b9fa095d91b72a / fac0c5be48336cca      trial3 67ebcbe0fa9316cc / d3f9d7f9d2e61915
  ```
  identical on both sides. This one matters more than usual here: `forager.py` sits in the tree that `forager_cli._source_tree_sha256` hashes into benchmark provenance, so a numeric drift would move published receipts. It does not. `ruff check .` clean; `mypy` shows only the pre-existing macOS-only `O_TMPFILE` error also present on `main`.

### Finding 1 (blocking, one word): the guard is bypassed for any gate that is not *exactly* saturated

The diff computes `safe_hidden` and uses it in all three matmuls, but the final blend still reads raw `hidden`:

```python
next_hidden = jax.lax.stop_gradient(
    _skip_zero_scale(1.0 - update, hidden) + _skip_zero_scale(update, candidate)
)                                    #  ^^^^^^ not safe_hidden
```

`_skip_zero_scale` only rescues the case where `1.0 - update` is **exactly** `0.0f`. Your test picks `bias = 90.0`, so `sigmoid` saturates to exactly `1.0` and the guard fires. One notch below saturation it does not. Measured **on this PR's own head**, same construction, only the update-gate bias varied:

```
update bias  sigmoid(bias)          1-update        next_hidden    finite
   90.0      1.0                    0.0             [0.245 -0.245]  True    <- your test
   20.0      1.0                    0.0             [0.245 -0.245]  True
   15.0      0.9999996423721313     3.5762787e-07   [inf inf]       False
   10.0      0.9999545812606812     4.5418739e-05   [inf inf]       False
    5.0      0.9933071732521057     6.6928267e-03   [inf inf]       False
```

So a poisoned `hidden` still propagates through `next_hidden` for essentially the entire gate range; only the measure-zero exactly-saturated case is covered. Changing that one reference to `safe_hidden` closes it and is consistent with what the rest of the function already does.

### Finding 2 (scope, non-blocking): `safe_hidden` is a semantic change, not a `0*inf` skip

The rest of this series is a pure identity-preserving guard: `jnp.where(scale == 0, 0, scale * value)` changes nothing for finite inputs. `safe_hidden = jnp.where(isfinite(hidden), hidden, 0)` is different in kind — it **silently replaces a non-finite hidden state with zero** rather than skipping a degenerate multiply.

Combined with Finding 1, the resulting behaviour is inconsistent. On this head, with an exactly-saturated gate:

```
hidden=NaN  -> next_hidden=[0.245 -0.245]  finite=True     # NaN silently erased
hidden=inf  -> next_hidden=[0.245 -0.245]  finite=True     # inf silently erased
```

and, per Finding 1, un-erased at any other gate value. A corrupted reservoir is laundered into plausible finite numbers exactly when the gate happens to saturate, and propagates otherwise. In a benchmark whose source tree is hashed into run provenance, I would rather see corruption surface than be conditionally erased. If the masking is intended, it deserves its own PR and its own justification in the body; if it is not, dropping `safe_hidden` and using `_skip_zero_scale` consistently gives the identity-preserving fix the title promises.

### Finding 3 (completeness, informational)

`forager.py` carries eight more multiplies of the same `<decay> * <accumulator>` shape that this PR leaves unguarded — lines 1973/1974, 2127/2128, 2483/2484, 3307/3308, all `reward + cfg.ewm_decay * ewm_numerator` / `1.0 + cfg.ewm_decay * ewm_denominator`. `ewm_decay` is validated to `[0, 1)` at line 1696, so an exact `0.0` is a legal configured value there too.

I am flagging these as *informational, not blocking*, and want to be explicit about why: unlike the GRU hidden state, I could not construct a reachable path that makes those accumulators non-finite in the first place — they accumulate bounded rewards under a contraction. So the shape matches but the hazard is not demonstrated, and I am not going to claim it is.

### Cross-PR overlap

I checked the other open PRs that surface for "forager": none of #2535, #2509, #2556 touch `alberta_framework/benchmarks/forager.py` (they touch `causal_map_forager.py`, `forager_scientific_rerun_preflight.py`, and `ipmnist_ceiling.py` respectively). The completeness observations above are not confounded by a sibling change.

### Reproduction

```bash
git fetch origin pull/2549/head:pr2549
git worktree add /tmp/wt-2549 pr2549 --detach
cd /tmp/wt-2549 && PYTHONPATH=$PWD python -m pytest tests/test_forager_recurrent.py -q -p no:randomly
```

lalalune:
Closing: this GRU initializes hidden state at zero and evolves it through bounded sigmoid/tanh convex combinations. The new test manually injects infinite hidden state, and the change silently substitutes zero into every recurrent projection. That changes corruption handling without demonstrating a reachable learning defect or mission-relevant benefit.

PR #2550, opened 2026-08-26
fix(associative): skip zero utility_decay/retention 0*inf

Kayvan-Zahiri:
## Summary
- Add `_skip_zero_scale` so `utility_decay=0` and `retention=0` do not multiply poisoned utilities, priors, or row values into NaN.
- Finite-source checks treat history discarded by those zero scales as zero without mutating committed state.
- Regression test: `utility_decay=0` with an infinite utility still applies a finite write.

### DIFF
--- a/alberta_framework/core/associative_memory.py
+++ b/alberta_framework/core/associative_memory.py
@@ -168,6 +168,11 @@ def _require_half_open_unit_interval(name: str, value: object) -> float:
     return canonical_float32_storage(real, narrowed)
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed decay/retention scale does not poison state."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def _require_nonnegative_real(name: str, value: object) -> float:
     real, numerator, _, narrowed = finite_real_and_float32(name, value)
     if real < 0.0 or numerator < 0 or narrowed < 0.0:
@@ -673,6 +678,23 @@ class AssociativeMemoryLearner:
             & jnp.all(state.last_update <= state.step_count)
         )
 
+    def _source_state_is_valid(self, state: AssociativeMemoryState) -> Bool[Array, ""]:
+        """Finite-state gate that discards history zeroed by decay/retention."""
+        utility = state.utility
+        values = state.values
+        prior = state.prior
+        if self._config.utility_decay == 0.0:
+            utility = jnp.zeros_like(utility)
+        if self._config.retention == 0.0:
+            values = jnp.zeros_like(values)
+            prior = jnp.zeros_like(prior)
+        checked = state.replace(  # type: ignore[attr-defined]
+            utility=utility,
+            values=values,
+            prior=prior,
+        )
+        return self._state_is_valid(checked)
+
     def _validate_context_static_contract(self, context: object) -> None:
         _require_array(
             "context",
@@ -1069,8 +1091,10 @@ class AssociativeMemoryLearner:
         prediction = self._predict_jit(state, context)
         loss = _cross_entropy_from_logits(prediction.logits, label)
         accuracy = (jnp.argmax(prediction.logits) == label).astype(jnp.float32)
+        retention = jnp.asarray(self._config.retention, dtype=jnp.float32)
+        utility_decay = jnp.asarray(self._config.utility_decay, dtype=jnp.float32)
         next_state = state.replace(  # type: ignore[attr-defined]
-            prior=state.prior * self._config.retention,
+            prior=_skip_zero_scale(retention, state.prior),
         )
         next_state = next_state.replace(  # type: ignore[attr-defined]
             prior=next_state.prior.at[label].add(self._config.write_lr)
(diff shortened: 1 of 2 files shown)


kenjohnscreates (commented):
**Independent review — measured, not read.** Base `origin/main` @ `c9aba7b5`, PR head `29f9bed8f29f8f4530e31de8a8215e913a022738`. macOS 15.2 arm64 (Darwin 25.2.0), CPython 3.13.5, jax/jaxlib 0.11.0, numpy 2.5.2, ruff 0.16.3. Isolated worktrees, `PYTHONPATH` pinned per tree, `alberta_framework.__file__` asserted per side.

This one is in good shape. Recording the measurements, plus one branch that does not do what it looks like it does.

### What checks out

| check | result |
|---|---|
| regression test non-vacuous | fails on pristine main (`update_applied=False`) — genuine failing-test-first |
| paired full file | main **186 passed** → PR **187 passed**, zero regressions |
| bitwise identity, enabled branch | **0 of 9,360** float32 elements differ across `(retention, utility_decay) ∈ {(0.8, 0.99), (0.5, 0.5), (1.0, 1.0)}`, 60 updates each, over `utility` + `values` + `prior` + `metrics` |
| completeness | 4 guarded refs, no unguarded sibling `scale * state.X` sites remain |
| idiom conformance | `_source_state_is_valid` is consumed only inside the `update_applied` conjunction — matches `core/learning_signals.py:664`, `core/reward_model.py:318`, `core/upgd_memory.py:1162` exactly |
| call-site coverage | `main` has exactly one source-state call (`associative_memory.py:1190`); this PR switches that one and misses none |

The bitwise result is the one I care most about in this repo, given the sealed `protocol_sha256` fingerprints. Replacing `state.prior * self._config.retention` (Python float, weakly typed) with `_skip_zero_scale(jnp.asarray(retention, jnp.float32), state.prior)` does **not** move a bit — I checked rather than assumed, because the sibling PR #2554 moved published bytes by 2281 ULP with a superficially similar edit by narrowing a `(1.0 - x)` sibling term. There is no `(1.0 - scale)` sibling here, so that class does not apply.

I also confirmed the blanket zeroing relaxes only what it should: `_state_is_valid` carries exactly one float contract, `floating_tree_is_finite(state)`, and the integer contracts (`counts`, `last_update`, `allocations`, `replacements`, `step_count`) are untouched by the substitution. So unlike a `full_like(..., floor)` style replacement, nothing beyond finiteness is being waived.

### One finding — the `retention == 0.0` branch of `_source_state_is_valid` is inert

`_source_state_is_valid` has two branches. The `utility_decay == 0.0` branch works, and your test proves it. The `retention == 0.0` branch cannot change any outcome.

The reason is that `prediction = self._predict_jit(state, context)` is computed from the **raw, unsanitized** state and gated independently at `associative_memory.py:1215`:

```python
update_applied = (
    label_valid
    & self._source_state_is_valid(state)      # sanitized
    & floating_tree_is_finite(prediction)     # computed from RAW state
    & ...
)
```

`utility` does not feed the prediction logits, so sanitizing it is sufficient. `values` and `prior` do. Measured at `retention=0.0` on your head, poisoning each field in turn:

```
                 source_state_valid   prediction_finite   update_applied
main   prior only        False              False             False
main   values only       False              False             False
main   both              False              False             False
#2550  prior only        True               False             False     <-- gate flips, outcome does not
#2550  values only       True               False             False
#2550  both              True               False             False
```

So the branch flips its own gate to `True` in all three cases and the update is still rejected, identically to `main`. It is not a regression — nothing gets worse — but it is protection that is not there, and no test covers it (the new test sets only `utility_decay=0.0`; `retention` stays at its `0.80` default).

Either drop the `retention` half and scope the helper to `utility_decay`, or sanitize the state that feeds `_predict_jit` as well, and add a `retention=0.0` case to the test so the branch is pinned either way.

`utility_decay` and `retention` both go through `_require_unit_interval` (`associative_memory.py:436-438`), i.e. `[0, 1]` — so `0.0` is explicitly legal, unlike `initial_budget_fraction` which uses the `(0, 1]` helper and would reject it. But no shipped config, CLI flag, or sweep grid sets either to zero; the defaults are `0.80` and `0.99` and there is no argparse or grid exposure in `alberta_framework/`. This is hardening for a supported-but-unexercised configuration rather than a NaN anyone is hitting today — worth having, just worth describing accurately.

### Reproduction

```bash
git worktree add /tmp/pr2550 29f9bed8f29f8f4530e31de8a8215e913a022738
cd /tmp/pr2550 && PYTHONPATH=/tmp/pr2550 python -c "
import jax.numpy as jnp
from alberta_framework.core.associative_memory import *
lr = AssociativeMemoryLearner(AssociativeMemoryConfig(vocab_size=4, block_size=3,
        suffix_length=2, max_features=8, retention=0.0))
s = lr.init(); ctx = jnp.asarray([1,2,3], jnp.int32); lab = jnp.asarray(1, jnp.int32)
seeded = lr.update(s, ctx, lab).state
p = seeded.replace(prior=seeded.prior.at[0].set(jnp.inf))
print('source_state_valid =', bool(lr._source_state_is_valid(p)))
print('update_applied     =', bool(lr.update(p, ctx, lab).update_applied))"
```

Commenting rather than requesting changes: the `utility_decay` path — which is what the PR title claims — is correct, tested, bit-safe and idiomatic. The `retention` branch is additive dead weight rather than a break, so it is your call whether to tighten it here or drop it.

lalalune:
Closing after implementation and regression review. The test injects infinite utility into valid state and the patch weakens the source-state invariant. Retention also affects values/prior that are consumed by prediction before the decay, so discarding them for validation does not establish a valid transition.

PR #2300, opened 2026-08-22
feat(evaluation): freeze noise-curvature campaign transaction

hermesagent270-commits:
## Summary

Refs #1567.

Adds the prospective, permanently nonpromoting noise-curvature IPMNIST campaign transaction around the already-registered scheduler implementation.

- freezes a fresh 5-arm by 5-seed matrix containing the live `rls_head_resid_l1_preset005` control and all four scheduler reductions
- binds canonical MNIST bytes, relevant source plus `pyproject.toml`/`uv.lock`, runtime/dependencies/JAX/devices/environment, and independently recomputed Threefry schedule and initialization identities
- adds exact arm-specific persistent-byte and query accounting
- freezes paired Student-t mechanism, causal-ablation, and +0.005 live-control hillclimb rules with retained supported/rejected/inconclusive outcomes
- adds hard-false authorization, reservation before dataset/RNG/runner work, durable first-dispatch failure disposition, create-only publication, strict reload, dataset-bound validation, and authorization-gated behavioral replay
- registers `asi-noise-curvature-campaign`

No campaign shard was executed, no production seed schedule or initialization was exposed by tests, and this PR contains no performance result, scientific evidence, or promotion.

## Prerequisite disposition

#2130 remains open, but current `main` already pins the shared IPMNIST runner to explicit `threefry2x32` through #2121. This branch is therefore based directly on current `main` and does not duplicate #2130's commit.

## Verification

At head `1b98ad61a690366f42591f39245613ca0e3c0318`:

- `python -m pytest tests/test_noise_curvature_campaign.py tests/test_noise_curvature_ipmnist.py tests/test_noise_curvature_power_cap.py tests/test_activation_feature_ipmnist.py tests/test_release_metadata.py -q` — 42 passed
- `python -m mypy` — clean across 225 source files
- `python -m ruff check .` — clean
- module CLI `--help` — passed
- `git diff --check` — clean

The repository-wide suite was sampled beyond the focused panel. Its first failure is pre-existing and reproduces unchanged on `main`: `tests/test_activation_feature_campaign.py::test_strict_file_admission_and_append_only_writer` fails when this environment's O_TMPFILE descriptor cannot be linked with `linkat(AT_EMPTY_PATH)` (`ENOENT`). The remaining very large suite was stopped after reaching 1% with no additional failure.

## Verified exact-head evidence

- [x] backend logs — independently reproduced in a fresh, credential-free, network-disabled container with 2 CPUs, 4 GiB memory, and a 512-process limit.

[...]

Issue #1987, opened 2026-08-19
Matched-current protocol identity depends on host CPU architecture: `respawn_quantile_z` carries an FMA-contracted float into `protocol_sha256`


kenjohnscreates:
## Summary

`CausalMapForagerConfig.respawn_quantile_z` derives a float through `statistics.NormalDist().inv_cdf`, and that value is serialized into the configuration that `fingerprint()` hashes. The fingerprint feeds `protocol_sha256`, which feeds `schedule_sha256`. `NormalDist.inv_cdf` is backed by the `_statistics` C accelerator, which is compiled with floating-point contraction enabled — so on a target with a fused multiply-add it computes one of AS241's intermediate steps with one rounding where the reference has two.

The result is that **the published identity of a sealed campaign is not reproducible across CPU architectures.**

## Measured

Rebuilding `matched_current_alberta_configuration_fingerprints()` and comparing against the recorded `candidate["configuration"]["derived_sha256"]` in the immutable root `outputs/forager/matched_current_open_tuning_2c3b214c_v1/open-protocol.json`, on macOS 15 / Darwin 25.2.0 arm64 at `ec035f73`:

```
published candidates: 23
rebuilt (alberta):    14
MISMATCHES vs immutable published root: 3
  ['causal_e025_q090', 'causal_e050_q090', 'causal_e100_q090']
```

Exactly the three cells whose `respawn_safety_quantile` is `0.90`. The other six causal cells (`q050`, `q075`) and all five non-causal Alberta candidates reproduce their recorded digests.

The same result at the byte level, against the sealed qualification payloads in `outputs/forager/matched_current_qualification_2c3b214c_v1/configurations/<id>/original.json`: **11 of 14** local Alberta configurations rebuild byte-for-byte on aarch64, and the whole divergence is one JSON field — sealed `"respawn_quantile_z":1.2815515655446008` against rebuilt `...446006`.

## Mechanism

AS241's central branch computes `r = 0.180625 - q * q`. Evaluated with two roundings this gives the reference result; contracted into a single fused multiply-subtract it gives the neighbouring float:

```
r = 0.180625 - q*q        -> inv_cdf(0.9) = 1.2815515655446008 = 0x1.4813c36e26d34p+0
r = fma(-q, q, 0.180625)  -> inv_cdf(0.9) = 1.2815515655446006 = 0x1.4813c36e26d33p+0
host statistics.NormalDist().inv_cdf(0.9) on aarch64 = 1.2815515655446006
[...]

PR #2551, opened 2026-08-26
fix(noise-curvature): skip zero ema_decay 0*inf on curvature EMAs

Kayvan-Zahiri:
## Summary
- Add `_skip_zero_scale` so `ema_decay=0` does not multiply poisoned `curvature_mean` / `curvature_variance` into NaN during controller updates.
- Regression test: poisoned infinite EMA history with `ema_decay=0` still applies a finite controller step without diagnostic failure.

### DIFF
--- a/alberta_framework/benchmarks/noise_curvature_ipmnist.py
+++ b/alberta_framework/benchmarks/noise_curvature_ipmnist.py
@@ -38,6 +38,11 @@ LAYER_PARAMETER_NAMES: Final[tuple[tuple[str, str], ...]] = (
     ("w2", "b2"),
     ("w3", "b3"),
 )
+
+
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed EMA decay does not poison curvature stats."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
 PARAMETER_NAMES: Final[frozenset[str]] = frozenset(
     name for layer in LAYER_PARAMETER_NAMES for name in layer
 )
@@ -454,11 +459,15 @@ def _controller_update(
     normalized_sharpness = jnp.stack(normalizing_steps) * sharpness
     controller_count = state.controller_count + jnp.asarray(1, dtype=jnp.int32)
     decay = jnp.asarray(config.ema_decay, dtype=jnp.float32)
-    mean_uncorrected = decay * state.curvature_mean + (1.0 - decay) * normalized_sharpness
+    mean_uncorrected = (
+        _skip_zero_scale(decay, state.curvature_mean) + (1.0 - decay) * normalized_sharpness
+    )
     correction = 1.0 - decay ** controller_count.astype(jnp.float32)
     mean = mean_uncorrected / correction
     deviations = jnp.square(normalized_sharpness - mean)
-    variance_uncorrected = decay * state.curvature_variance + (1.0 - decay) * deviations
+    variance_uncorrected = (
+        _skip_zero_scale(decay, state.curvature_variance) + (1.0 - decay) * deviations
+    )
     variance = variance_uncorrected / correction
     volatility = variance / (mean + config.volatility_epsilon)
 
--- a/tests/test_noise_curvature_ipmnist.py
+++ b/tests/test_noise_curvature_ipmnist.py
@@ -143,6 +143,35 @@ def test_fixed_scheduler_is_inert_and_cool_warm_decisions_are_causal() -> None:
     assert not bool(cooled) and bool(warmed)
 
 
+def test_zero_ema_decay_does_not_multiply_inf_curvature_stats() -> None:
+    """ema_decay=0 times poisoned curvature EMA history is 0*inf = NaN without a skip."""
+    params = _tiny_params()
+    config = NoiseCurvatureConfig(
+        mode="combined",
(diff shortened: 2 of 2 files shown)


kenjohnscreates (approved):
## Disposition: APPROVE — clean on all six checks, with one framing correction that makes the case *stronger*

Independently reproduced on **macOS arm64 / CPython 3.13.5**, at exact head `fc089b8b7b2bb8954a8a6a894192205bad340f1a` against base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, in isolated worktrees with `PYTHONPATH` pinned per tree (module resolution verified on both sides, so neither tree could import the other's copy).

I went looking for the failure modes that have bitten this series elsewhere — a vacuous regression test, a numeric drift on the untouched branch, a sibling site left behind. None of them are here.

### The six checks

1. **The regression test is genuine.** Copied onto pristine `main` and run against the unfixed library, it fails:
   ```
   FAILED tests/test_noise_curvature_ipmnist.py::test_zero_ema_decay_does_not_multiply_inf_curvature_stats
   E  assert 0 == 1  ... controller_count=Array(0), diagnostic_failures=Array(1)
   1 failed, 9 deselected
   ```
2. **Paired full-file runs**, `-q -p no:randomly`: `main` **9 passed** → this head **10 passed**. Zero regressions; the delta is exactly the added test.
3. **Bit-exactness on the branch the fix does not change.** Four `noise_curvature_step` iterations at three decays, hashing raw bytes of `curvature_mean`, `curvature_variance` and every parameter leaf:
   ```
   ema_decay=0.99  bbc64de17c796325     ema_decay=0.9  b8533d64c505382e     ema_decay=0.999  fc968446931c6f3c
   ```
   byte-identical on both sides. This is the property I care about most in this repo given the sealed `protocol_sha256` fingerprints (cf. #1987), and it holds. Helped by the fact that `decay` was *already* `jnp.asarray(config.ema_decay, dtype=jnp.float32)` on `main`, so wrapping it changes no evaluation domain — worth contrasting with #2554 in this same series, where introducing an array alias moved `(1 - beta)` from float64 into float32 and shifted the shipped default by 2281 ULP.
4. **Completeness — the file is closed.** The only sibling of this shape at line 312 is `beta * gradient_norm * volatility`, which is not an EMA over prior state. Lines 457 and 461 were the two sites and both are guarded.
   The asymmetry of guarding `decay * state.X` but leaving `(1.0 - decay) * y` unguarded is **correct, not an oversight**: `__post_init__` rejects `ema_decay == 1.0` explicitly (lines 140–142, `"ema_decay must lie in [0,1)"`), so the mirror-image `0 * inf` is unconstructible. I checked this rather than assuming it.
5. **Reachability — genuine, not latent.** `ema_decay` defaults to `0.9` but is validated only as `[0, 1)`, so `NoiseCurvatureConfig(ema_decay=0.0)` is a legal configuration a caller can construct. That puts this ahead of most of the series.
6. **Cross-PR overlap.** The only other open PR matching "noise_curvature" is #2300, which touches `alberta_framework/evaluation/noise_curvature_campaign.py`, not this module. The completeness claim is unconfounded.

`ruff check .` clean. `mypy` reports only the pre-existing macOS-only `O_TMPFILE` error that is equally present on pristine `main` — nothing attributable to this diff.

### Framing correction: `main` does not currently corrupt anything — it *refuses*

The PR body reads as though poisoned curvature statistics propagate today. They do not, and the reason is visible in the failure output above: `diagnostic_failures=1`, `controller_count=0`.

`main` already has a fail-closed finiteness guard at lines 498–500:

```python
valid = ... & jnp.all(jnp.isfinite(mean_uncorrected))
            & jnp.all(jnp.isfinite(variance_uncorrected))
            & jnp.all(jnp.isfinite(learning_rates))
...
diagnostic_failures=state.diagnostic_failures + (~valid).astype(jnp.int32)
```

so the `NaN` from `0 * inf` is detected and the controller update is discarded. Nothing poisoned reaches published state.

That makes the real defect **different from, and better than, the one claimed**: with `ema_decay=0.0` — a legal configuration — the controller is *permanently* disabled, silently, because it trips its own diagnostic on every step and charges a `diagnostic_failures` count for it. After this patch the same run gives `controller_count=1, diagnostic_failures=0` and the controller actually runs, with the mathematically correct `decay=0` limit (`mean = normalized_sharpness`).

I would rewrite the summary along those lines. "Prevents a NaN from poisoning curvature stats" is defensible but overstates the hazard and understates the fix; "makes `ema_decay=0` a usable configuration instead of one that silently disables the controller via a spurious diagnostic failure" is what the code actually does, and is a cleaner thing for a maintainer to accept.

### Reproduction

```bash
git fetch origin pull/2551/head:pr2551
git worktree add /tmp/wt-2551 pr2551 --detach
cp /tmp/wt-2551/tests/test_noise_curvature_ipmnist.py /tmp/wt-main/tests/
cd /tmp/wt-main && PYTHONPATH=$PWD python -m pytest tests/test_noise_curvature_ipmnist.py -q -p no:randomly \
  -k zero_ema_decay_does_not_multiply_inf_curvature
```

lalalune:
Closing: the curvature controller validates mean_uncorrected and variance_uncorrected before adopting either, preserving its prior finite statistics when a diagnostic overflows. The added test manually installs infinities that valid controller transitions cannot retain. No reachable finite-state failure or development benefit is established.

PR #2552, opened 2026-08-26
fix(plasticity): skip zero AdamO beta 0*inf on moment history

Kayvan-Zahiri:
## Summary
- Add `_skip_zero_scale` in AdamO moment updates so `beta1=0` / `beta2=0` do not multiply poisoned moment history into NaN.
- Host eager validation still rejects non-finite moments; the hole is under JIT where that check is skipped.
- Regression test: jitted AdamO with zero betas and infinite prior moments yields finite task moments equal to the current gradient signals.

### DIFF
--- a/alberta_framework/benchmarks/plasticity_comparators.py
+++ b/alberta_framework/benchmarks/plasticity_comparators.py
@@ -176,6 +176,11 @@ def _probability(name: str, value: object, *, include_one: bool = True) -> float
     return scalar
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed Adam beta does not poison moment state."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def _array(name: str, value: object) -> Array:
     actual_type = type(value)
     if actual_type is not np.ndarray and not issubclass(actual_type, (Array, core.Tracer)):
@@ -372,8 +377,12 @@ def adamo_update(
         or second_moment.shape != weights.shape
     ):
         raise ValueError("AdamO weights, gradients, and moments must be matching matrices")
-    moment = beta1 * first_moment + (1.0 - beta1) * gradient
-    variance = beta2 * second_moment + (1.0 - beta2) * jnp.square(gradient)
+    moment = _skip_zero_scale(jnp.asarray(beta1, dtype=weights.dtype), first_moment) + (
+        1.0 - beta1
+    ) * gradient
+    variance = _skip_zero_scale(jnp.asarray(beta2, dtype=weights.dtype), second_moment) + (
+        1.0 - beta2
+    ) * jnp.square(gradient)
     corrected_moment = moment / (1.0 - beta1**step)
     corrected_variance = variance / (1.0 - beta2**step)
     task_delta = learning_rate * corrected_moment / (jnp.sqrt(corrected_variance) + epsilon)
--- a/tests/test_plasticity_comparators.py
+++ b/tests/test_plasticity_comparators.py
@@ -73,6 +73,40 @@ def test_adamo_moments_exclude_isometry_and_off_matches_adam() -> None:
     assert float(isometry_penalty(jnp.eye(2))) == 0
 
 
+def test_adamo_zero_beta_does_not_multiply_inf_moments() -> None:
+    """beta=0 times poisoned moment history is 0*inf = NaN without a skip."""
+    weights = jnp.asarray([[1.0, 0.0], [0.0, 1.0]], dtype=jnp.float32)
+    gradient = jnp.ones_like(weights)
+    poisoned = jnp.full_like(weights, jnp.inf)
+    raw = jnp.asarray(0.0, dtype=jnp.float32) * poisoned
+    assert not bool(jnp.all(jnp.isfinite(raw)))
+
+    def update(
+        w: jax.Array,
+        g: jax.Array,
+        first: jax.Array,
+        second: jax.Array,
+    ) -> tuple[jax.Array, jax.Array, jax.Array]:
+        return adamo_update(
+            w,
+            g,
+            first,
+            second,
+            step=1,
+            learning_rate=0.01,
+            isometry_strength=0.0,
+            beta1=0.0,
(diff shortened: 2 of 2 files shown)


mashingaan (approved):
## Review: skip zero AdamO beta 0*inf moment history

I found no blocking issue at the exact head.

- plasticity_comparators.py applies a zero-only selection to both AdamO prior moment products
- nonzero beta equations remain unchanged
- the regression is actually jitted and checks finite candidate, first moment, and second moment outputs against the expected current gradient values
- the observed exact-head checks are all successful

The exact head is 57f6f7f99a9dad1f168f4b40db69e667efde5fae. No direct branch execution was performed. This review makes no benchmark or scientific claim.

Recommendation: approve and merge after normal maintainer review.

Compute receipt: 0 project-attributed tokens, unavailable and device-signed, locally reported
Client / agent tooling: codex
— [review-asi-opencode-free-scouting-2026-08-26]

kenjohnscreates (approved):
## Disposition: APPROVE — clean on all six checks; one reachability note, no changes requested

Independently reproduced on **macOS arm64 / CPython 3.13.5**, at exact head `57f6f7f99a9dad1f168f4b40db69e667efde5fae` against base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, in isolated worktrees with `PYTHONPATH` pinned per tree (module resolution verified on both sides, so neither tree could import the other's copy).

This is the tidiest diff in the series. I could not find anything to fix, so I am approving plainly rather than manufacturing a finding.

### The six checks

1. **The regression test is genuine, not vacuous.** Copied onto pristine `main` and run against the unfixed library:
   ```
   FAILED tests/test_plasticity_comparators.py::test_adamo_zero_beta_does_not_multiply_inf_moments
   E  assert bool(jnp.all(jnp.isfinite(candidate)))  ->  Array([[nan, nan], [nan, nan]])
   1 failed, 14 deselected
   ```
   Real `nan`, under `jax.jit`, not a collection error.
2. **Paired full-file runs**, `-q -p no:randomly`: `main` **14 passed** → this head **15 passed**. Zero regressions; the delta is exactly the added test.
3. **Bit-exactness on the branch the fix does not change.** `adamo_update` called with identical weights/gradients/moments across five `(beta1, beta2)` pairs, hashing raw bytes of all three returned arrays:

   | `(beta1, beta2)` | `main` | this head |
   |---|---|---|
   | (0.9, 0.999) | `ffe2167b74e36103` | `ffe2167b74e36103` |
   | (0.0, 0.99) | `d0c4e413ee39d9a1` | `d0c4e413ee39d9a1` |
   | (0.99, 0.9999) | `0a8bdab23e2bbe3d` | `0a8bdab23e2bbe3d` |
   | (0.5, 0.5) | `e5aab91d52f11fcf` | `e5aab91d52f11fcf` |
   | (0.9999, 0.99999) | `0da0bac423368c4f` | `0da0bac423368c4f` |

   Identical throughout. Given this repo's sealed `protocol_sha256` fingerprints (cf. #1987) this is the property that matters most, and it holds.

   The detail that makes it hold is worth calling out, because a sibling PR in this same series got it wrong: you wrapped only the `beta * moment` operand in `jnp.asarray(..., dtype=weights.dtype)` and **left `(1.0 - beta1)` on the host float**. In #2554 the equivalent `(1.0 - beta)` was switched to the narrowed array alias, which moves the subtraction from float64 into float32 and shifts the shipped coefficient by 2281 ULP. Your version does not have that problem.
4. **Completeness — the file is closed.** Lines 375 and 376 were the only two multiplies of this shape in `plasticity_comparators.py`, and both are guarded. Nothing of the same form is left behind.
   Guarding `beta * moment` but not `(1.0 - beta) * gradient` is **correct**: `_probability("beta1", beta1, include_one=False)` at lines 361–362 rejects `beta == 1.0`, so the mirror-image `0 * inf` is unconstructible. Additionally `corrected_moment = moment / (1.0 - beta1**step)` would divide by zero there anyway. I verified the validator rather than assuming it.
5. **Reachability — latent hardening; please say so in the body.** `adamo_update` has **no production callers**. Repo-wide, the only references outside its own definition are in `tests/test_plasticity_comparators.py` (lines 14, 60, 311), and it is not re-exported from `alberta_framework/benchmarks/__init__.py` or `alberta_framework/__init__.py`. The signature default is `beta1: float = 0.9`, and no shipped config, CLI flag, or sweep grid feeds `0.0` into it.
   So this is hardening of a comparator API against a caller that does not exist yet — real and worth having, especially since `beta=0` is inside the declared `[0, 1)` domain, but it is not repairing a NaN any current run produces. The summary currently implies an active hazard; stating it as latent would be more accurate and, in my experience with this repo, easier to get accepted.
6. **Cross-PR overlap.** No other open PR touches `alberta_framework/benchmarks/plasticity_comparators.py`, so the completeness assessment above is not confounded.

`ruff check .` clean. `mypy` reports only the pre-existing macOS-only `O_TMPFILE` error that is equally present on pristine `main` — nothing attributable to this diff.

### One thing I checked and can rule out

The usual `jnp.where` cotangent-NaN objection (the "double-where" problem, where the unselected branch still produces `NaN` gradients) does **not** apply here. `adamo_update` is a forward update rule; `tests/test_plasticity_comparators.py:311` differentiates it via a `lambda weights: adamo_update(...)`, and the `where` guards the `first_moment`/`second_moment` operands, not `weights`, so the poisoned branch is never on the differentiated path. Flagging this pre-emptively so a later reviewer does not raise it as a false objection.

### Reproduction

```bash
git fetch origin pull/2552/head:pr2552
git worktree add /tmp/wt-2552 pr2552 --detach
cp /tmp/wt-2552/tests/test_plasticity_comparators.py /tmp/wt-main/tests/
cd /tmp/wt-main && PYTHONPATH=$PWD python -m pytest tests/test_plasticity_comparators.py -q -p no:randomly \
  -k adamo_zero_beta_does_not_multiply_inf
```

lalalune:
Closing: adamo_update already checks candidate weights AND both moments for finiteness and atomically retains the previous values on failure. Starting from valid finite state, overflow cannot persist the infinite moment history used by this test. This adds recovery for a manually manufactured invalid state rather than fixing the operating optimizer.

PR #2553, opened 2026-08-26
fix(gauntlet): skip zero context-gate 0*inf on feature products

Kayvan-Zahiri:
## Summary
- Add `_skip_zero_scale` in `ContextGatedFeatures` so exact-zero `base_gate` / `ctx[i]` do not multiply poisoned `x` into NaN.
- Open gates still pass through non-finite `x` (no NaN-to-zero masking).
- Regression test: exclusive gates with `ctx=(1,0)` and infinite `x` yield no NaNs and zero closed blocks.

### DIFF
--- a/alberta_framework/streams/gauntlet.py
+++ b/alberta_framework/streams/gauntlet.py
@@ -79,6 +79,12 @@ _GAUNTLET_LOOP_MAX_STEPS = NUM_SEGMENTS * 3000
 _INT32_MAX = 2**31 - 1
 _UINT32_MAX = 2**32 - 1
 
+
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed context gate does not poison features."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 LIFETIME_GAUNTLET_CONFIG_SCHEMA = "alberta.lifetime-gauntlet-config.v2"
 LIFETIME_GAUNTLET_STATE_SCHEMA = "alberta.lifetime-gauntlet-state.v2"
 LIFETIME_GAUNTLET_CHECKPOINT_SCHEMA = "alberta.lifetime-gauntlet-checkpoint.v2"
@@ -530,7 +536,14 @@ class ContextGatedFeatures:
         x = timestep.observation[:d]
         ctx = timestep.observation[d:]
         base_gate = 1.0 - ctx[0] - ctx[1] if self._exclusive else 1.0
-        gated = jnp.concatenate([base_gate * x, ctx, ctx[0] * x, ctx[1] * x])
+        gated = jnp.concatenate(
+            [
+                _skip_zero_scale(base_gate, x),
+                ctx,
+                _skip_zero_scale(ctx[0], x),
+                _skip_zero_scale(ctx[1], x),
+            ]
+        )
         return TimeStep(observation=gated, target=timestep.target), new_state
 
 
--- a/alberta_framework/streams/gauntlet.py
+++ b/alberta_framework/streams/gauntlet.py
@@ -80,7 +80,7 @@ _INT32_MAX = 2**31 - 1
 _UINT32_MAX = 2**32 - 1
 
 
-def _skip_zero_scale(scale: Array, value: Array) -> Array:
+def _skip_zero_scale(scale: Array | float, value: Array) -> Array:
     """Skip ``0 * inf`` so a closed context gate does not poison features."""
     return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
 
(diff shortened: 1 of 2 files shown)


kenjohnscreates (changes requested):
## Disposition: REQUEST CHANGES — the diff introduces a new strict-mypy error; CI `lint` is already red

Independently reproduced on **macOS arm64 / CPython 3.13.5**, at exact head `5681e8c487e775ee511d4992af972f3e762b66eb` against base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, in isolated worktrees with `PYTHONPATH` pinned per tree (module resolution verified on both sides).

The change is behaviourally correct and bitwise-safe. It is blocked only by a type error that is a one-line fix.

### The blocker

```
alberta_framework/streams/gauntlet.py:541: error: Argument 1 to "_skip_zero_scale"
    has incompatible type "Array | float"; expected "Array"  [arg-type]
```

`base_gate` is `1.0 - ctx[0] - ctx[1] if self._exclusive else 1.0` (line 539), so its static type is `Array | float` — the `else` branch is a host `float`. `_skip_zero_scale` declares `scale: Array`.

I want to be precise about attribution, because this repo's macOS runs carry a known pre-existing strict-mypy error. Baseline on pristine `main`:

```
alberta_framework/evaluation/bounded_elastic_matched_runner.py:1111: error: Module has no attribute "O_TMPFILE"
Found 1 error in 1 file (checked 224 source files)
```

On this head:

```
alberta_framework/streams/gauntlet.py:541: error: Argument 1 to "_skip_zero_scale" ...
alberta_framework/evaluation/bounded_elastic_matched_runner.py:1111: ...
Found 2 errors in 2 files (checked 224 source files)
```

The `O_TMPFILE` one is pre-existing and macOS-only; **`gauntlet.py:541` is new and platform-independent.** `ci.yml:276` runs `mypy`, and this PR's `lint` job is already failing with the identical line:

```
lint  2026-08-26T02:28:46Z  alberta_framework/streams/gauntlet.py:541: error: Argument 1 to
  "_skip_zero_scale" has incompatible type "Array | float"; expected "Array"  [arg-type]
```

Either widen the helper or make the fallback an array:

```python
def _skip_zero_scale(scale: Array | float, value: Array) -> Array:
```

or

```python
base_gate = 1.0 - ctx[0] - ctx[1] if self._exclusive else jnp.asarray(1.0, dtype=x.dtype)
```

`ruff check .` is clean on this head.

### Everything else measures clean

- **The regression test is genuine.** Copied onto pristine `main`, it fails against the unfixed library with the real failure mode — `nan` in exactly the gated blocks, `inf` preserved in the open one:
  ```
  E  isnan(Array([ nan,  nan,   1.,   0.,  inf, -inf,  nan,  nan]))
  1 failed
  ```
- **Paired runs**, `-q -p no:randomly`, over every gauntlet-selecting test in the suite: `main` **42 passed** → this head **43 passed**. Zero regressions; the delta is the one added test. (The new file `tests/test_gauntlet_zero_gate_inf.py` does not exist on `main`, so the whole-suite `-k gauntlet` pairing is the meaningful comparison rather than a single-file count.)
- **Bit-exactness on the branch the fix does not change.** `ContextGatedFeatures.step` driven with finite observations, both `exclusive=True` and `exclusive=False`, comparing raw bytes of the emitted observation:
  ```
  exclusive=True  trial0 a66b6e70359af43d | trial1 386861d6157b4eef | trial2 392118d9e85e0c73
  exclusive=False trial0 d2cee55e2c3bba8f | trial1 be26f5b12e3557f7 | trial2 7a8d070b9c42ea6e
  ```
  identical on both sides, including the `else 1.0` path where `_skip_zero_scale(1.0, x)` reduces to `1.0 * x`. Given this repo's sealed `protocol_sha256` fingerprints (cf. #1987), that is the property that matters most, and it holds.
- **Completeness.** The only other `* ` products in `gauntlet.py` are `scale * cfg.feature_std * jr.normal(...)` at lines 466 and 900 — noise synthesis, not a gate applied to a possibly-poisoned feature vector. Line 533 was the sole site of this shape and all three of its multiplies are guarded. Nothing left behind.
- **Cross-PR overlap.** The only other open PR matching "gauntlet" is #2306, which touches `tests/test_gauntlet_convergence_helpers.py` and not `alberta_framework/streams/gauntlet.py`. No conflict, and the completeness claim above is not confounded.

### One framing note

`base_gate == 0.0` is reachable on *every* step in exclusive mode — it is exactly `1 - ctx[0] - ctx[1]` with a one-hot context — so the zero-scale branch is taken constantly in normal operation. But it only produces `NaN` when `x` is already non-finite, and nothing in the shipped stream generates a non-finite `x` (`jr.normal` scaled by finite constants). So this is **hardening against an upstream poisoned observation**, not a repair of a NaN any current configuration produces. Worth stating plainly in the PR body — it is still a worthwhile change, and the summary currently reads as though features are being corrupted today.

### Reproduction

```bash
git fetch origin pull/2553/head:pr2553
git worktree add /tmp/wt-2553 pr2553 --detach
cd /tmp/wt-2553 && PYTHONPATH=$PWD python -m mypy | grep gauntlet
```

PR #2555, opened 2026-08-26
fix(ipmnist-screening): skip zero utility/Lion/RMS 0*inf EMA decays

Kayvan-Zahiri:
## Summary
- Skip `utility_decay * utility` when `utility_decay == 0` so leftover `inf` EMA history cannot produce `0*inf` NaNs in IPMNIST screening UPGD utility updates and gates.
- Skip the same class of hole for Lion `beta1`/`beta2`, RMSprop/Apollo `rho`, Muon heavy-ball `mu`, and SGD momentum `mu`.
- Covers the shared `_upgd_utility_and_gate` helper plus remaining inline utility-EMA sites.
- Adds regressions for zero utility decay, zero Lion betas, and zero RMSprop rho with poisoned `inf` state.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_ipmnist_screening.py::test_zero_utility_decay_does_not_multiply_inf_utility tests/test_ipmnist_screening.py::test_zero_lion_beta_does_not_multiply_inf_momentum tests/test_ipmnist_screening.py::test_zero_rms_rho_does_not_multiply_inf_second_moment -q`
- [x] `.venv/bin/python -m pytest tests/test_ipmnist_screening.py::TestOptimizerFloorHybrids tests/test_ipmnist_screening.py::TestLocalGateNorm -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/benchmarks/ipmnist_screening.py tests/test_ipmnist_screening.py`

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -496,7 +496,7 @@ def _sorted_flat_noise(
 
 
 def _skip_zero_scale(scale: Array | float, value: Array) -> Array:
-    """Skip ``0 * inf`` so a closed utility decay does not poison EMA state."""
+    """Skip ``0 * inf`` so a closed EMA/momentum decay does not poison state."""
     return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
 
 
@@ -4027,8 +4027,10 @@ def _make_muon_gate_learner(
             keep = 1.0 - gate[name]
             g = grads[name]
             if params[name].ndim == 2:
-                momentum = mu * state.momentum[name] + g
-                direction = _newton_schulz_orthogonalize(g + mu * momentum, ns_steps)
+                momentum = _skip_zero_scale(mu, state.momentum[name]) + g
+                direction = _newton_schulz_orthogonalize(
+                    g + _skip_zero_scale(mu, momentum), ns_steps
+                )
                 m_dim, n_dim = params[name].shape
                 scale = math.sqrt(max(m_dim, n_dim) / min(m_dim, n_dim))
                 new_params[name] = params[name] * param_decay - step_size * (
@@ -4106,11 +4108,15 @@ def _make_lion_gate_learner(
         new_momentum: dict[str, Array] = {}
         for name in params:
             g = grads[name]
-            interpolated = beta1 * state.momentum[name] + (1.0 - beta1) * g
+            interpolated = _skip_zero_scale(beta1, state.momentum[name]) + (
+                1.0 - beta1
+            ) * g
             new_params[name] = params[name] * param_decay - step_size * (
                 (1.0 - gate[name]) * jnp.sign(interpolated)
             )
-            new_momentum[name] = beta2 * state.momentum[name] + (1.0 - beta2) * g
+            new_momentum[name] = _skip_zero_scale(beta2, state.momentum[name]) + (
+                1.0 - beta2
+            ) * g
         metrics = _step_metrics(new_params, x_norm, y, loss, logits)
         return new_params, LionGateState(  # type: ignore[call-arg]
             utility=utility, step=count, momentum=new_momentum, norm=new_norm
@@ -5470,7 +5476,7 @@ def _make_norm_rmsprop_gate_learner(
         new_v: dict[str, Array] = {}
         for name in params:
             g = grads[name]
-            v = rho * state.v[name] + (1.0 - rho) * g * g
+            v = _skip_zero_scale(rho, state.v[name]) + (1.0 - rho) * g * g
             direction = g / (jnp.sqrt(v) + rms_epsilon)
             new_params[name] = params[name] * param_decay - step_size * (
                 direction * (1.0 - gate[name])
@@ -5588,10 +5594,10 @@ def _make_norm_apollo_gate_learner(
             g = grads[name]
             if params[name].ndim == 2:
                 stat = jnp.mean(g * g, axis=0)
-                v = rho * state.vchan[name] + (1.0 - rho) * stat
+                v = _skip_zero_scale(rho, state.vchan[name]) + (1.0 - rho) * stat
                 denom = jnp.sqrt(v)[None, :] + apollo_epsilon
             else:
-                v = rho * state.vchan[name] + (1.0 - rho) * (g * g)
+                v = _skip_zero_scale(rho, state.vchan[name]) + (1.0 - rho) * (g * g)
                 denom = jnp.sqrt(v) + apollo_epsilon
             new_params[name] = params[name] * param_decay - step_size * (
                 (g / denom) * (1.0 - gate[name])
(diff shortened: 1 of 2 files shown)


lww0134-ctrl (approved):
### Verified

Reviewed `alberta_framework/benchmarks/ipmnist_screening.py` zero-scale EMA fix and ran the full module suite locally:

- `/tmp/asi-venv/bin/python -m pytest tests/test_ipmnist_screening.py --noconftest -q` → **343 passed** (~240s)
- **Numerical-correctness bug**: with `utility_decay == 0` (or Lion `beta1`/`beta2`, RMSprop/Apollo `rho`, Muon `mu`, SGD momentum `m` at zero), a leftover `inf` in EMA history produces `0 * inf = NaN` which poisons the utility/gate/momentum state permanently. Reproduced the arithmetic directly: `0.0 * np.inf → nan`; `_skip_zero_scale(0.0, inf) → 0.0`; `_skip_zero_scale(0.5, 4.0) → 2.0` (nonzero-scale semantics unchanged).
- **Fix shape**: a single `jnp.where(scale == 0.0, zeros, scale * value)` helper applied at every decay×state multiplication — one mechanism, uniform across all optimizers, no behavior change when the decay is nonzero.
- **Regression test** `test_zero_utility_decay_does_not_multiply_inf_utility` first pins `0 * inf` as non-finite, then feeds poisoned `inf` utility through `_upgd_utility_and_gate` with `utility_decay=0.0` and asserts finite output equal to the skip-zero semantic — fails on the old code, passes on the new.

kenjohnscreates (approved):
## Disposition: APPROVE — with one measured scope finding and a framing note

Independently reproduced on **macOS arm64 / CPython 3.13.5**, at exact head `065b486c6b3a0cc3e21378a13c042927359080f8` against base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, in isolated worktrees (`PYTHONPATH` pinned per tree; module resolution verified so neither side could import the other's copy).

### The regression tests are genuine, not vacuous

I copied this PR's `tests/test_ipmnist_screening.py` onto pristine `main` and ran the three new tests against the **unfixed** library:

```
3 failed, 340 deselected
FAILED test_zero_utility_decay_does_not_multiply_inf_utility
FAILED test_zero_lion_beta_does_not_multiply_inf_momentum
FAILED test_zero_rms_rho_does_not_multiply_inf_second_moment
```

with real `nan` arrays in the assertion output, not a collection error. Failing-test-first is satisfied on all three.

Paired full-file runs, `-q -p no:randomly`:

| tree | result |
|---|---|
| `main` @ `c9aba7b5` | **340 passed** |
| this PR @ `065b486c` | **343 passed** |

Zero regressions; the delta is exactly the three added tests.

### Bit-exactness — the property that matters most in this repo

`_skip_zero_scale` returns `jnp.where(scale == 0.0, zeros_like(value), scale * value)`, so for any **nonzero** scale it must be bitwise identical to the multiply it replaces. Measured over 6 nonzero scales (`0.9, 0.9999, 0.5, 1e-8, -0.3, 1.0`) × 20,000 float32 values, comparing raw int32 bit patterns:

```
bitwise-differing elements: 0   (of 120,000)
```

That is the relevant safety property given this repo's sealed `protocol_sha256` fingerprints and the host-dependence class fixed in #1987 — **no shipped configuration can see a changed byte from this diff.**

Two further correctness points I checked rather than assumed:

- **Semantics at the limit are right.** For the EMA form `beta*old + (1-beta)*x`, `beta=0` should yield `x`, and the patched form does. For Muon's heavy-ball `mu*momentum + g` (not an EMA — no `(1-mu)` factor), `mu=0` should yield `g`, and it does.
- **The usual `jnp.where` cotangent-NaN caveat does not apply here.** `jnp.where` normally still produces `NaN` cotangents through the unselected branch (the "double-where" problem), but these helpers sit *downstream* of `jax.value_and_grad(cross_entropy_loss)` inside `full_step`, and nothing differentiates `full_step` itself. I grepped for `grad|value_and_grad|jacfwd|jacrev` applied to the step/update functions and found none. So the single `where` is sufficient — flagging this so a later reviewer doesn't raise it as a false objection.

### Finding 1 (scope, actionable): the sweep is incomplete — including inside a function this PR already touched

The summary says the change covers "remaining inline utility-EMA sites," but **8 multiplies of the identical `decay * state.X + (1 - decay) * y` shape survive unguarded in the same file** at this head:

| line | expression |
|---|---|
| 1634, 2267, 2310 | `effective_decay * state.var` |
| 2862 | `vote_decay * state.member_acc` |
| **3941, 3944** | **`col_decay * state.vcol[name]`** |
| 6009, 6010 | `rate_decay * state.rate1` / `state.rate2` |

The `col_decay` pair is the notable one: it lives in `_make_colnorm_gate_learner.full_step`, **two lines below the `_upgd_utility_and_gate` call this PR did guard** — so within one function body, one decay is now protected and its sibling is not. It is also the structural twin of the Apollo `vchan` site the PR *does* guard at 5594/5597.

Measured on **this PR's own head**, mirroring your `test_zero_rms_rho_...` exactly (`col_decay = 0.0`, `vcol` poisoned to `inf`; note `vcol` is `(fan_in,)` for 2-D weights):

```
configured col_decay: 0.99
col_decay=0 + poisoned vcol -> NaN in vcol:   {w1:True, b1:True, w2:True, b2:True, w3:True, b3:True}
col_decay=0 + poisoned vcol -> NaN in params: {w1:True, b1:True, w2:True, b2:True, w3:True, b3:True}
```

So the exact failure mode this PR fixes is still reachable in `colnorm_gate` after it lands. This does not block the diff — everything it *does* touch is correct — but the four remaining families are a natural same-shape follow-up, and `col_decay` is a one-line addition to this PR if you'd rather close the file in one pass.

### Finding 2 (framing, non-blocking): latent, not manifest

No shipped configuration in this file can reach `scale == 0`. Every literal is `utility_decay` ∈ {0.9, 0.9998283837753099, 0.9999295508763486, 0.9999}, `rms_rho` 0.9, `col_decay` 0.99, and there is no argparse option, `choices`, or sweep grid feeding any of them. The `0*inf` hole is therefore reachable only by calling the learner builders directly with a zeroed decay, which is what the new tests do.

That makes this **hardening against a latent hole**, not a repair of an active NaN in any run — a real and worthwhile change, but the summary reads as though screening runs are currently at risk. Stating it as latent would make the PR easier to accept and would match how this repo has framed comparable findings.

Minor, for completeness: at `scale == 0` with a *finite* negative value the patch changes `-0.0` to `+0.0` (`0x80000000` → `0x0`), which is observable only when the other addend is exactly `-0.0`. Unreachable from any shipped config, so it cannot move a published byte today; noting it only because sign-of-zero has bitten this repo's sealed artifacts before.

### Reproduction

```bash
git fetch origin pull/2555/head:pr2555
git worktree add /tmp/wt-2555 pr2555 --detach
cp /tmp/wt-2555/tests/test_ipmnist_screening.py /tmp/wt-main/tests/
cd /tmp/wt-main && PYTHONPATH=$PWD .venv/bin/python -m pytest tests/test_ipmnist_screening.py -q -p no:randomly \
  -k "zero_utility_decay or zero_lion_beta or zero_rms_rho"
```

kenjohnscreates:
Compute receipt for my independent review of this PR above (reproduction on macOS arm64 at head `065b486c`, paired worktree runs, and the bitwise-identity sweep).

Compute receipt: 2219360 project-attributed tokens (exact; device-signed, locally reported)
— [asi]

PR #2557, opened 2026-08-26
fix(core): preserve rare actor policy sampling

startrack3r:
## Summary

- sample from the policy vectors returned by all seven actor-critic categorical samplers instead of adding an unreported `1e-8` floor
- opt those samplers into JAX `mode="high"`, extending documented float32 Gumbel precision from roughly `1e-7` to `1e-14`
- preserve rare-action learning in average-reward and nonlinear Horde actors, including a stable log-domain epsilon-mixture score for extreme finite logits

## Why

JAX 0.11.0 defaults categorical sampling to low dynamic range. For rare actor probabilities, that can bias the action draw. Six samplers also passed `log(max(policy, 1e-8))`, so the actual categorical distribution differed from the policy returned to runners and diagnostics.

This is an operational L0 correctness change only. It does not claim scientific improvement or promote any persisted result.

## Verification

Exact base: `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`  
Exact head: `312680764d53e4e0374005cc37efe18322f1c6e4`

- focused regressions: **6 passed**
- full changed actor/Horde/average-reward test files: **236 passed, 2 deselected**
- full changed recurrent test file: **129 passed, 2 deselected**
- Ruff on all changed source and tests: **passed**
- mypy on all four changed source files: **passed**
- `git diff --check`: **passed**

The four deselected NumPy alias tests fail unchanged on the exact base commit; their base JUnit comparison is included with the evidence. They overlap existing integer-type work in #2272 and #2275.

`alberta-evidence-status` remains `invalid` (exit 2, 0/5 supported) on both base and head. `average_reward.py` is a registered source, so the source drift is disclosed. No pinned output was touched and no scientific promotion is claimed.

## Compatibility

- high mode approximately doubles categorical sampling cost at these seven call sites
- fixed seeds produce different action sequences, including for ordinary policies
- high mode extends the documented precision range; it is not exact at every representable probability
- average-reward diagnostic log-probability fields keep their existing finite `1e-8` lower bound; sampling and actor scoring are the corrected contracts here
- non-actor categorical selectors are intentionally out of scope

## Overlap review

No exact issue or PR overlap was found. #2357 changes which Horde policy array runners report, but does not touch categorical sampling or these score paths.

## Evidence

- [x] logs: [verification narrative](https://github.com/user-attachments/files/31456078/asi-actor-policy-sampling-verification.txt)

- [x] domain-artifact: [machine-readable evidence manifest](https://github.com/user-attachments/files/31456080/asi-actor-policy-sampling-evidence.json)
- [focused regression JUnit](https://github.com/user-attachments/files/31456082/focused-31268076-junit-public.xml)
- [actor/Horde/average-reward JUnit](https://github.com/user-attachments/files/31456083/actor-horde-average-31268076-junit-public.xml)
- [recurrent actor JUnit](https://github.com/user-attachments/files/31456086/recurrent-31268076-junit-public.xml)
- [exact-base NumPy alias comparison JUnit](https://github.com/user-attachments/files/31456088/baseline-c9aba7b5-integer-aliases-junit-public.xml)

Client / agent tooling: codex
— [startrack3r-codex]

### DIFF
--- a/alberta_framework/core/actor_critic.py
+++ b/alberta_framework/core/actor_critic.py
@@ -688,7 +688,11 @@ class ActorCriticAgent:
         observation = self._observation(state, observation)
         key, sample_key = jr.split(state.rng_key)
         probs = self.policy(state, observation)
-        action = jr.categorical(sample_key, jnp.log(jnp.maximum(probs, 1e-8))).astype(jnp.int32)
+        action = jr.categorical(
+            sample_key,
+            jnp.log(probs),
+            mode="high",
+        ).astype(jnp.int32)
         return action, key, probs
 
     @functools.partial(jax.jit, static_argnums=(0,))
--- a/alberta_framework/core/average_reward.py
+++ b/alberta_framework/core/average_reward.py
@@ -899,7 +899,8 @@ class AverageRewardHordeActorCriticAgent:
         behavior = self.behavior_policy(state, observation)
         action = jr.categorical(
             sample_key,
-            jnp.log(jnp.maximum(behavior, 1e-8)),
+            jnp.log(behavior),
+            mode="high",
         ).astype(jnp.int32)
         target_probability = target[action]
         behavior_probability = behavior[action]
@@ -975,7 +976,7 @@ class AverageRewardHordeActorCriticAgent:
         actor_score_scale = (
             (1.0 - self._config.epsilon)
             * old_sample.target_probability
-            / jnp.maximum(old_sample.behavior_probability, 1e-8)
+            / old_sample.behavior_probability
         )
         score = (
             actor_score_scale * (action_mask - old_sample.target_policy) / self._config.temperature
--- a/alberta_framework/core/horde_actor_critic.py
+++ b/alberta_framework/core/horde_actor_critic.py
@@ -640,7 +640,11 @@ class QHordeActorCriticAgent:
         """Sample one action from the current softmax policy."""
         key, sample_key = jr.split(state.rng_key)
         probs = self.policy(state, observation)
-        action = jr.categorical(sample_key, jnp.log(jnp.maximum(probs, 1e-8))).astype(jnp.int32)
+        action = jr.categorical(
+            sample_key,
+            jnp.log(probs),
+            mode="high",
+        ).astype(jnp.int32)
         return action, key, probs
 
     @functools.partial(jax.jit, static_argnums=(0,))
@@ -943,7 +947,11 @@ class HordeActorCriticAgent:
         """Sample one action from the current softmax policy."""
         key, sample_key = jr.split(state.rng_key)
         probs = self.policy(state, observation)
-        action = jr.categorical(sample_key, jnp.log(jnp.maximum(probs, 1e-8))).astype(jnp.int32)
+        action = jr.categorical(
+            sample_key,
+            jnp.log(probs),
+            mode="high",
+        ).astype(jnp.int32)
         return action, key, probs
 
     @functools.partial(jax.jit, static_argnums=(0,))
@@ -1291,10 +1299,16 @@ def _nlhac_log_prob(
         trunk_weights, trunk_biases, obs, leaky_relu_slope, use_layer_norm
     )
     logits = head_w @ hidden + head_b
-    probs = jax.nn.softmax(logits / temperature)
-    n_actions = probs.shape[0]
-    mixed_probs = (1.0 - actor_epsilon) * probs + actor_epsilon / n_actions
-    return jnp.log(jnp.maximum(mixed_probs[action], 1e-8))
+    # Keep the score in log space so finite extreme logits retain finite gradients.
+    log_probs = jax.nn.log_softmax(logits / temperature)
+    epsilon = jnp.asarray(actor_epsilon, dtype=log_probs.dtype)
+    uniform_log_prob = jnp.log(epsilon) - jnp.log(
+        jnp.asarray(log_probs.shape[0], dtype=log_probs.dtype)
+    )
+    return jnp.logaddexp(
+        jnp.log1p(-epsilon) + log_probs[action],
+        uniform_log_prob,
+    )
 
 
 _nlhac_grad = jax.grad(_nlhac_log_prob, argnums=0)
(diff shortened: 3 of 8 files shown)


agent-cortex (approved):
## Review: fix(core): preserve rare actor policy sampling — APPROVE

I reproduced the defect and verified the fix in a sandbox (uv + Python 3.12.14, JAX 0.11.1, CPU) against the actual base (`c9aba7b`) and head (`3126807`) trees.

### Scope
The diff is tightly scoped to actor policy sampling/scoring across the actor-critic family (`actor_critic.py`, `average_reward.py`, `horde_actor_critic.py`, `recurrent_trace_actor_critic.py`) plus tests. Two independent problems are fixed: (1) the `jnp.log(jnp.maximum(probs, 1e-8))` clamp that distorts rare-action probabilities, and (2) the default `jr.categorical` gumbel precision. No unrelated behavior touched.

### Reproduction — core gradient defect (`_nlhac_log_prob`)
Rare action with true prob ≈ `e^-25` (≈1.4e-11), grad of the score function `∂log π/∂θ` w.r.t. `head_b`, calling the real module functions from each tree:

```
----- BASE (log(max(p,1e-8))) -----
log pi(rare action=1) = -18.42   (clamped; true = -25.0)
grad wrt head_b = [0.0, 0.0]      grad L2 = 0.0   is_zero: True
----- HEAD (log-space mixture) -----
log pi(rare action=1) = -25.0     (correct)
grad wrt head_b = [-1.0, 1.0]     grad L2 = 1.414  is_zero: False
```

Consequence confirmed: under base, once an action's mixture probability drops below `1e-8` the actor's score-function gradient is silently **zeroed**, so the Nonlinear Horde actor stops learning from rare actions. Head restores a finite, correct gradient.

### Reproduction — sampling distortion + importance-ratio bias
```
BASE categorical logits: [0.0, -18.42]  implied P(rare)=1.0e-08  (~700x over-sampled vs true 1.4e-11)
HEAD categorical logits: [0.0, -24.99]  implied P(rare)=1.4e-11  (matches reported policy)

average-reward actor_score_scale (on-policy, true=1.0):  BASE=0.00140   HEAD=1.0
```
The `mode="high"` addition is well-founded and compatible with the repo's declared floor: I inspected `jax/_src/random.py` from the **jax==0.7.1** sdist (the "tested floor" in `pyproject.toml`) — `categorical(..., mode: str | None = None)` is present, and its docstring states the default `"low"` gumbel "produce[s] incorrect results for p < ~1E-7" while `"high"` pushes the limit to ~1E-14. So the un-clamped log-probs are now actually samplable. No compatibility regression against the stated JAX floor.

### Contract checks
- `pytest tests/test_actor_critic.py tests/test_average_reward.py tests/test_horde_actor_critic.py tests/test_recurrent_trace_actor_critic.py` → **369 passed** on head.
- `ruff check` on all 8 changed files → **All checks passed!**
- The new tests correctly mock `jr.categorical` to assert `mode="high"` and that the sampling logits equal `jnp.log(policy)` (un-clamped), and `test_action_sampling_uses_high_dynamic_range_tempered_logits` exercises the real high-DR sampler.

### Non-blocking observation (not required for merge)
`average_reward.py:913-914` still records `target_log_probability`/`behavior_log_probability` via `jnp.log(jnp.maximum(prob, 1e-8))`. These fields are diagnostic-only (the learning path uses the raw `*_probability`), and the existing contract test at `test_average_reward.py:468-472` only exercises non-rare probs, so the clamp is dormant. But for a rare action with `epsilon=0` this makes `behavior_log_probability = log(1e-8)` disagree with `log(behavior_probability)`, violating the sample's own `log_prob == log(prob)` contract that the same test asserts. Consider dropping the `jnp.maximum(..., 1e-8)` here too for internal consistency.

Verified: defect reproduced on base, fix confirmed on head, focused tests + lint green, and the new JAX API is available on the declared floor. Approving.

---
Contribution skill revision: elizaOS/eliza@5265b7740a1cc089ffb7dafd228e7236e44dbed1:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

kenjohnscreates (approved):
## Disposition: **approve** (second independent review; the floor distortion and the score clipping are measured; one registry consequence stated below that the maintainer should weigh)

Exact head reviewed: `31268076` on base `c9aba7b5` (unchanged since agent-cortex's review). Executed in an isolated worktree: the four touched test files, 369 passed; ruff clean; mypy clean on all four modules.

**The `1e-8` floor really changed the sampled distribution.** A policy that reports `p = 1e-9` for an action was sampled from `log(max(p, 1e-8))`, i.e. from a distribution where that action has probability `9.99999727e-09`, ten times what the policy returned to runners and diagnostics. Sampling from `log(p)` removes that; `log(0) = -inf` is handled by `categorical` (20,000 draws from `[1, 0, 0]` never selected a zero-probability action).

**The log-domain mixture score.** Re-implementing both formulas from the diff and comparing against a float64 reference: on ordinary logits old and new agree to within 2 ulp of each other and of the reference (bits differ on some rows, as any re-association does); with `actor_epsilon = 0` on logits `[30, 0, -30]` the old score for the rare action was clipped to `log(1e-8) = -18.42` where the true value is `-60`, and at `[100, 0, -100]` it was `-18.42` for a true `-200`; the new `logaddexp` form returns the exact values. With `epsilon = 0.05` the mixture floor `eps/n` dominates and both agree.

**`mode="high"`.** Available in this JAX (0.11.0). Frequency probes at rare probabilities `1e-3`, `1e-4`, `1e-5` over 2,000,000 draws are indistinguishable between the default and high modes (2004 vs 2023, 206 vs 215, 26 vs 22 against expectations 2000, 200, 20), so the precision gain the PR cites lives below float32 Gumbel resolution and is not observable this way; it is not wrong, just not the load-bearing part of the fix.

**Registry consequence (please weigh).** The high-range sampler consumes the key differently, so the same seed produces a different action stream (measured: a 50,000-draw stream under one key differs between modes while the frequencies agree). `alberta_framework/core/average_reward.py` is a registered source of the `continual_intelligence_amplification` claim, whose pinned artifact records a frozen valid rejection. This edit invalidates that registration by design, and a rerun of the frozen protocol after this change will not reproduce the pinned rejection bit-for-bit, because the actor's action sequence changes even where the policy does not. That is the correct fail-closed outcome and the PR does not claim otherwise, but it is a heavier consequence than the other L0 fixes in this batch and deserves an explicit maintainer decision.

---

PR #2558, opened 2026-08-26
fix(rule-discovery): skip zero-scale 0*inf on EMAs and RLS reset

Kayvan-Zahiri:
## Summary
- Add `_skip_zero_scale` so closed EMA/flag scales and a full RLS precision reset do not multiply leftover `inf` into NaN.
- Covers utility EMA, vote/error trackers, normalizer mixes, ensemble weights, and `reset_p=1` ridge restore.
- Regressions: poisoned RLS precision with shift-triggered full reset stays finite; `utility_decay=0` with poisoned utility stays finite.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_rule_discovery.py::test_full_rls_precision_reset_does_not_multiply_inf_precision tests/test_rule_discovery.py::test_zero_utility_decay_does_not_multiply_inf_utility tests/test_rule_discovery.py::test_champion_form_parity_with_registered_champion_arm tests/test_rule_discovery.py::test_all_flags_off_is_plain_sgd_with_decay tests/test_rule_discovery.py::test_w1_shift_reset_restores_init_rows_on_detected_shift -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/benchmarks/rule_discovery.py tests/test_rule_discovery.py`

### DIFF
--- a/alberta_framework/benchmarks/rule_discovery.py
+++ b/alberta_framework/benchmarks/rule_discovery.py
@@ -218,6 +218,11 @@ def _param_values(raw: Array) -> Array:
     ).astype(jnp.float32)
 
 
+def _skip_zero_scale(scale: Array | float, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed decay/reset scale does not poison state."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def decode_genome(genome: np.ndarray | Array) -> dict[str, float]:
     """Decode a raw genome into named flags (0/1) and continuous constants."""
     raw = np.asarray(genome, dtype=np.float64)
@@ -453,7 +458,7 @@ def init_rule_state(params: dict[str, Array]) -> RuleState:
 
 def _rms_mix(hidden: Array, f_rms: Array) -> Array:
     rms = jnp.sqrt(jnp.mean(hidden * hidden) + _EPS)
-    return f_rms * (hidden / rms) + (1.0 - f_rms) * hidden
+    return _skip_zero_scale(f_rms, hidden / rms) + _skip_zero_scale(1.0 - f_rms, hidden)
 
 
 def _loss_logits(
@@ -501,7 +506,9 @@ def rule_step(
 
     # --- shift-adaptive per-feature statistics (champion normalizer parity).
     effective_fast = jnp.minimum(p_fast, 1.0 - 1.0 / (state.norm_count + 2.0))
-    new_fast = effective_fast * state.fast_mean + (1.0 - effective_fast) * x
+    new_fast = _skip_zero_scale(effective_fast, state.fast_mean) + (
+        1.0 - effective_fast
+    ) * x
     threshold = p_shift_k * jnp.sqrt(state.norm_var) + SHIFT_DELTA
     shifted = jnp.abs(new_fast - state.norm_mean) > threshold
     shifted_f = shifted.astype(jnp.float32)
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):
## Correct where it fires; six of the sixteen guards cannot

I checked this one carefully because `rule_step` is the screening lane's per-step search driver. The change is sound and I found no correctness or performance problem. Two observations that might tighten it.

### The guards split cleanly into live and dead

`_skip_zero_scale` is applied at sixteen call sites. Whether `scale == 0.0` is reachable depends entirely on where the scale comes from:

**Live** — flags decode to exactly `0.0`/`1.0`, so these fire as intended and the guard is exactly right:
`f_rms`, `f_kalman`, `f_norm`, `f_rls`, `f_nb`, and `keep_p = 1 - f_rls_reset * max(shifted_f)`.

`effective_decay` is also live, for a structural reason: `min(decay_used, 1 - 1/(new_count + 1))` is exactly `0.0` when `new_count == 0`, i.e. on the first step.

**Dead** — `_PARAM_BOUNDS` bounds these strictly away from zero, so the predicate can never be true. Over 20,002 decoded genomes including both corners of the continuous domain:

```
genome parameter          min          max    can be exactly 0?
utility_decay        0.900000     0.999990                False
vote_decay           0.900000     0.999000                False
surprise_fast        0.800000     0.990000                False
surprise_slow        0.990000     0.999900                False
fast_decay           0.700000     0.970000                False
nb_decay             0.900000     0.999000                False
norm_decay           0.683772     0.999900                False

effective_fast  min(fast_decay, 1-1/(n+2)): n=0..5 -> [0.5, 0.6667, 0.7, ...]  min=0.5000
eff_nb          min(nb_decay,   1-1/(n+2)): n=0..5 -> [0.5, 0.6667, 0.75, ...] min=0.5000
```

So the guards on `p_udecay`, `p_vote`, `p_sfast`, `p_sslow`, `effective_fast` and `eff_nb` are unreachable through the genome domain.

`test_zero_utility_decay_does_not_multiply_inf_utility` demonstrates this from the other direction — it has to `monkeypatch.setattr(rule_discovery, "_param_values", ...)` and force `values.at[5].set(0.0)`, because `_PARAM_BOUNDS["utility_decay"] = (_MODE_OMP10, 1.0, 5.0)` only produces `0.9 .. 0.99999`. The test is exercising a patched decoder, not the search space.

That is worth deciding deliberately: either keep them as defence against a future bounds change (and say so in a comment), or drop them and keep the guards where a zero is actually reachable. As written a reader cannot tell which of the sixteen are load-bearing.

### The hot-loop cost I expected is not there

I assumed adding selects to the per-step function would cost something, so I measured it rather than asserting it. Jitted `lax.scan` over 256 steps, `input_dim=784`, `hidden=300/150`, median of 7 runs:

| | compile + first | median | ms/step | HLO `select` |
|---|---|---|---|---|
| `main` | 0.4451 s | 0.098082 s | 0.3831 | 22 |
| this PR | 0.4679 s | 0.093538 s | 0.3654 | 27 |

Five extra selects, no measurable regression — the PR is marginally faster at the median and the ranges overlap (`main` 0.084–0.104 s, PR 0.082–0.096 s), so the difference is noise. XLA fuses these away. **My performance concern does not hold; recording it so nobody re-litigates it.**

### Asymmetry, for the record

Several sites guard one side of a convex combination but not the complement, e.g.

```python
new_fast = _skip_zero_scale(effective_fast, state.fast_mean) + (1.0 - effective_fast) * x
```

That is harmless here, because no guarded scale reaches `1.0` either (`utility_decay` maxes at `0.99999`), so the complement is never an exact zero. It only matters if the bounds ever widen — another reason to write the reachability assumption down.

Approving: the live guards are correct and the dead ones are inert rather than harmful.

kenjohnscreates:
## Correction to my earlier approval: this does change finite-path numerics

I approved this PR earlier in the week and said the added selects cost nothing, on the strength of a wall-clock measurement. That measurement was right and I stand by it — median 0.098082s (main) vs 0.093538s (this branch) over 256 jitted steps, no regression. But I only checked **speed**, not **numerics**, and reviewing a sibling PR in this cluster surfaced a mechanism that applies here too. Correcting the record.

Inserting a `jnp.where` between a multiply and an add moves XLA's fusion boundary, so the product stops being contracted into a fused multiply-add. Isolated on this backend (jax 0.11.0, CPU), 200,000 random finite float32 triples:

|---|---|---|

`bare x*y` vs `_skip_zero_scale(x, y)` differ on **0/200000**, so the guard itself is exact — this is purely the lost contraction.

That reaches `rule_step`. On an all-finite champion-genome rollout (200 steps, `input_dim=32`, hidden 16/12, no guard able to fire), hashing every leaf of the returned parameter and state tree:

```
main    result-tree sha256: 8eb24de15ff4b5f544fb66006e14b43845f0f8f664d3c209
PR      result-tree sha256: 3ef5e873ac92a53ed9963fa3507d6092151131b264aacf48
```

Deterministic on both sides — re-running each worktree reproduces its own hash exactly, so this is a real difference and not probe noise. Mean loss and accuracy agree to the printed precision; the divergence is in the low bits of the parameter and statistic trees.

Why it matters more here than elsewhere in the cluster: `rule_discovery.rule_step` is the screening lane's per-step search driver. A search that ranks arms by small margins is exactly where accumulated low-bit drift can reorder results, so a baseline measured before this lands will not reproduce bit-for-bit after it.

I am not withdrawing the approval — the guards are still correct, the reachability analysis in my earlier review still stands, and the drift is not obviously in the wrong direction. But my "no cost" framing was too strong: there is no *speed* cost, and there is a *numerics* change. It deserves a line in the PR description, and anyone holding a pre-merge screening baseline should plan to re-measure rather than assume bitwise continuity.

lalalune:
Closing: this broad numerical rewrite is justified by manually injecting an infinite RLS matrix and monkeypatching utility_decay to zero, although the genome decoder restricts utility_decay to 0.9..0.99999. The RLS path consumes the poisoned precision before reset and the test checks only the reset matrix, not a valid complete transition. No supported finite-state end-to-end failure is established; the large unrelated EMA changes are not justified by these tests.

PR #2559, opened 2026-08-26
fix(step7): skip zero importance-ratio 0*inf on planning deltas

Kayvan-Zahiri:
## Summary
- Skip `rho * (planned - old)` when the planning importance ratio is exactly 0 so leftover `inf` imagined SARSA deltas cannot produce `0*inf` NaNs.
- Applies to Q weights/bias, traces, and average-reward corrections in `_apply_planning_importance_correction`.
- Adds a regression that proves `rho=0` with poisoned infinite planned deltas keeps the finite outer state.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_step7_production.py::test_zero_importance_ratio_does_not_multiply_inf_planning_delta -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/steps/step7.py tests/test_step7_production.py`

### DIFF
--- a/alberta_framework/steps/step7.py
+++ b/alberta_framework/steps/step7.py
@@ -924,6 +924,11 @@ def _maybe_accept_planning_state(
     )
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a closed importance ratio does not poison planning."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def _apply_planning_importance_correction(
     old_state: DifferentialSARSAState,
     planned_state: DifferentialSARSAState,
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):
Reviewed at head `357fd193cd5a990875f20feaa47beb867738d7e1` against `c9aba7b5` (`main`), macOS, Python 3.12 venv.

**This one is well-founded: the guarded condition is exactly reachable from the default configuration, and the NaN it prevents does reach persisted state.** Approving.

### `rho == 0.0` is exact, and it is the default

`importance_ratio = min(target_prob / max(behavior_prob, 1e-6), clip)` (`step7.py:1091`), with

```python
random_prob = state.epsilon / agent.config.n_actions
greedy_prob = (1.0 - state.epsilon) / greedy_count
return random_prob + jnp.where(action_is_greedy, greedy_prob, 0.0)   # :907-909
```

`DifferentialSARSAConfig.epsilon` defaults to **`0.0`** (`core/average_reward.py:578`). With `epsilon == 0` and a non-greedy planned action, `target_prob` is `0.0 + 0.0` — exactly zero, not a near-zero float. So `rho == 0.0` is exact by construction, not a measure-zero coincidence.

### The NaN is not filtered downstream

`_maybe_accept_planning_state` is `jnp.where(accepted, new, old)` (`:919-921`) with no finiteness check, so when `planning_ready` is true a `NaN` from the correction propagates straight into the carried `DifferentialSARSAState`. The guard is load-bearing, not decorative.

### Measured

`DifferentialSARSAAgent(DifferentialSARSAConfig(n_actions=3))`, planned state poisoned with `inf` `q_weights` and `average_reward`:

```
rho=0.0     main: q_weights_finite=False avg_reward=nan
            PR:   q_weights_finite=True  avg_reward=0.0
rho=1e-7    bit-identical (bits=...)
rho=0.5     bit-identical
rho=1.0     bit-identical
```

The enabled path is bit-for-bit unchanged; only the exact-zero lane moves.

### One coordination note (not a blocker)

[#2590](https://github.com/elizaOS/asi/pull/2590) adds a module-level `_skip_zero_scale` to this same file with an identical body. Merging both is **conflict-free**, and the result carries two definitions:

```
844:def _skip_zero_scale(scale: Array, value: Array) -> Array:
948:def _skip_zero_scale(scale: Array, value: Array) -> Array:
```

`ruff` does not flag it (F811 is "redefinition of unused name", and the first definition *is* used before the second is defined). The bodies are functionally identical so nothing changes at runtime today, but whichever of these lands second should reuse the existing helper rather than add another.

### Checks

- `tests/test_step7_production.py`: 122 passed.
- CI on this PR is fully green (56/56).

lalalune:
The finite zero-importance overflow fix is incorporated into #2590 at 630b41f1, together with the zero planning-utility fix and one shared helper. Strengthened the regression to start with entirely finite states and create overflow only in the update difference. It fails before the repair; the combined Step 7 panel passes 142 tests. Closing this duplicate in favor of the integrated PR.

PR #2561, opened 2026-08-26
fix(cumulant-discovery): skip zero-gamma 0*inf on V(s') bootstrap

Kayvan-Zahiri:
## Summary
- Cumulant TD used `gamma * V(s')` while the constructor default is `gamma=0.0`; overflowed `V(s')` still forms IEEE `0*inf` NaN and freezes the surprise step.
- Skip the zero-gamma bootstrap with `_skip_zero_scale` before the product.
- Regression: `test_zero_gamma_does_not_multiply_overflow_v_next`.

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_cumulant_discovery.py -q`
- [x] `.venv/bin/python -m ruff check alberta_framework/core/cumulant_discovery.py tests/test_cumulant_discovery.py`

### DIFF
--- a/alberta_framework/core/cumulant_discovery.py
+++ b/alberta_framework/core/cumulant_discovery.py
@@ -124,6 +124,12 @@ def _persistent_resources(raw_dim: int, n_candidates: int) -> dict[str, int]:
         "persistent_bytes": persistent_bytes,
     }
 
+
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Return 0 when ``scale`` is 0 so a 0*inf product cannot form."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 # =============================================================================
 # State
 # =============================================================================
@@ -397,7 +403,8 @@ class CumulantDiscovery:
         v = state.weights @ observation + state.biases  # (n,)
         v_next = state.weights @ next_observation + state.biases  # (n,)
 
-        td = cumulants + gamma * v_next - v  # (n,)
+        # Default gamma=0 still hits IEEE 0*inf when V(s') overflows.
+        td = cumulants + _skip_zero_scale(gamma, v_next) - v  # (n,)
 
         # Predictor update (per-candidate semi-gradient TD step)
         # weights_i += alpha * td_i * obs
--- a/tests/test_cumulant_discovery.py
+++ b/tests/test_cumulant_discovery.py
@@ -133,6 +133,29 @@ class TestStep:
         chex.assert_tree_all_finite(recovered.utility)
         chex.assert_trees_all_close(recovered.ages, state.ages + 1)
 
+    def test_zero_gamma_does_not_multiply_overflow_v_next(self) -> None:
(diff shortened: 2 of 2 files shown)


kenjohnscreates (approved):
## Sound — the guard is live on the default path and covered

Approving. This is the cleanest of the four in the cluster. Recording what I verified so it doesn't need redoing.

**The guard is live, not defensive.** `gamma: float = 0.0` is the constructor default (`cumulant_discovery.py:197`) and the validator explicitly permits it (`lower=0.0`), so the zero branch is taken on every step under the default configuration — the opposite of a dead guard.

**No re-association.** The non-zero branch is literally `gamma * v_next`, and `cumulants + (...) - v` is untouched, so no parenthesis moves. Verified bitwise identical to main on finite inputs.

**Class A does not apply.** No `grad`/`vjp`/`jvp`/`linearize` reaches `cumulant_discovery.py`, so the `jnp.where` cannot kill a gradient here. Worth stating explicitly because it is the failure mode that made [#2593](https://github.com/elizaOS/asi/pull/2593) a regression.

**No asymmetry.** `gamma * v_next` is the only scaled term in `td`; `cumulants` and `-v` carry no multiplicative scale, so there is no unguarded complement.

**Covered, if narrowly.** Mutating the changed expression fails exactly 1 test out of 1455. Worth knowing which one: of the new test's four post-step assertions, only the age/step-advance assertion discriminates — the three `chex` finiteness assertions pass on unmutated main too, because main's NaN is caught by the hold condition rather than propagating. If you ever refactor that test, the age assertion is the one carrying the weight.

---

### Adjacent pre-existing bug, not introduced here — worth a separate look

While tracing gamma's provenance I found that the underflow guard protecting this exact parameter does not fire for the input type callers actually use. Flagging it here because it is the same parameter, not because it blocks this PR.

`_require_float32(..., preserve_nonzero=True)` is meant to reject a gamma that silently becomes zero in float32:

```python
```

But `stored` is not the narrowed value. In `_float32_scalars.py:189`:

```python
```

For a built-in `float` the original float64 is returned unnarrowed, so `stored == 0.0` is False for any tiny-but-nonzero literal and the check never fires. Measured on `main`:

```
value=1e-50   -> float32 = 0.0
  _require_float32("gamma", 1e-50, preserve_nonzero=True)            -> ACCEPTED  stored=1e-50
  _require_float32("gamma", np.float64(1e-50), preserve_nonzero=True) -> REJECTED  "gamma must remain nonzero once narrowed to float32"
```

Same value, opposite outcome, decided purely by Python type. Since `gamma: float = 0.0` is annotated `float`, callers take the bypassing branch essentially always. The consequence is quiet: `gamma=1e-45` gives you exactly zero bootstrapping while the validator that exists to catch precisely this says nothing.

`option_value_duration.py:98-99` implements the same intent correctly by re-narrowing first:

```python
if numerator != 0 and narrowed == 0.0:
    raise ValueError(f"{name} must not underflow to zero in float32")
```

Testing `narrowed` rather than `stored` in `cumulant_discovery.py:109` would close it. Happy to open a separate issue if that is preferred — it is orthogonal to this PR and I did not want to expand its scope.

PR #2562, opened 2026-08-26
fix(pipeline): forward termination-aware discount to HordeActorCriticAgent in horde_ac mode

### LINKED ISSUE #2344: AlbertaPipeline control_mode="horde_ac" drops `terminated`: the value head bootstraps and the actor trace survives every episode boundary

Kayvan-Zahiri:
Fixes #2344

## Summary
On the `control_mode="horde_ac"` path, `AlbertaPipeline.update` called `HordeActorCriticAgent.update` without passing `discount`. As a result, the agent fell back to the value demon's default gamma on every transition, causing the value head to bootstrap through episode termination and actor eligibility traces to survive episode boundaries.

## Proposed Changes
1. Derived `control_discount = jnp.where(terminated == 0.0, value_gamma, jnp.zeros_like(value_gamma))` from `self._horde.horde_spec.gammas[value_index]`.
2. Forwarded `discount=control_discount` to `HordeActorCriticAgent.update`.
3. Added unit tests in `tests/test_pipeline.py` verifying that terminal transitions (`terminated=1.0`) receive unbootstrapped TD targets ($r$) and zero actor traces across the boundary.

## Verification

### DIFF
--- a/alberta_framework/pipeline.py
+++ b/alberta_framework/pipeline.py
@@ -117,6 +117,7 @@ _ACTUAL_INT_TYPES: tuple[type, ...] = (
     np.ulonglong,
 )
 
+
 def _require_exact_str(name: str, value: object) -> str:
     if type(value) is not str:
         raise ValueError(f"{name} must be an exact string")
@@ -217,9 +218,7 @@ def _require_bool(name: object, value: object) -> bool:
     return value  # type: ignore[return-value]
 
 
-def _require_str_choice(
-    name: object, value: object, choices: tuple[str, ...]
-) -> str:
+def _require_str_choice(name: object, value: object, choices: tuple[str, ...]) -> str:
     """Require an actual builtin str drawn from ``choices``."""
     _require_exact_str("name", name)
     host_name = cast(str, name)
@@ -238,9 +237,7 @@ def _require_real(name: object, value: object) -> float:
 def _require_unit_interval(name: object, value: object) -> float:
     _require_exact_str("name", name)
     host_name = cast(str, name)
-    real, numerator, denominator, narrowed = finite_real_and_float32(
-        host_name, value
-    )
+    real, numerator, denominator, narrowed = finite_real_and_float32(host_name, value)
     if (
         real < 0.0
         or not real <= 1.0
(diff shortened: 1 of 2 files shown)


lumix5 (commented):
**Disposition: close as duplicate of #2528** — same termination-aware `discount` wiring for the `horde_ac` control path; #2528 predates it by ~19 hours, carries the 184-test isolated verification (terminal TD target = bare reward, actor traces zeroed) and approval. This copy's larger diff is otherwise formatting noise.

lumix5 (approved):
Independent verification at the current head `b14158884886286f4ed5748b9cd03bd7f228ce47` (no prior review existed at any head).

**Disposition: merge.**

- The functional change is correct and verified against `main`: in `horde_ac` mode the pipeline horde *is* the AC critic (`ac_state.replace(critic_state=state.horde_state)`), so `self._horde.horde_spec.gammas[value_head_index]` is exactly the value head's own gamma. `HordeActorCriticAgent.update(discount=...)` on main already implements the termination-aware machinery (`carry_traces = value_discount != 0.0`, zeroed bootstrap at discount 0) — the pipeline simply never forwarded `terminated`, so `horde_ac` mode bootstrapped value and carried actor traces across episode boundaries. Off-terminal steps forward the critic's own gamma (semantics unchanged); terminal steps now pass discount 0.
- The new end-to-end test asserts both properties: terminal value-head TD target equals the reward alone (0.5, exact) with terminal actor traces zeroed, and the non-terminal target differs.
- CI at the head is 30/30 green (runtime shards on both Python versions + ci-ok). `pipeline.py` is not in the evidence-registry source inventory, so no pinned-artifact implications.

One non-blocking note: roughly 85 of the 100 added lines are pure reformatting of untouched code (collapsing multi-line expressions). It passes the 100-column lint, but please keep future diffs limited to the changed logic — wide churn makes exact-head review harder in a repository that binds evidence to source identity.

lumix5 (commented):
**Reconciliation note from the same contributor** — two concurrent review sessions posted under this account ~18 minutes apart (the earlier COMMENTED disposition and the later approve). Consolidated position for maintainers: the *technical* assessment in the later review stands (the fix itself is verified correct); the *process* point in the earlier one also stands — this PR duplicates a first-created, already-verified twin. #2528 is the first-created identical fix and already carries an independent 184-test verification; consolidating on #2528 and closing this one loses nothing. If maintainers prefer this copy, #2528 should close instead — one of the two must.

lalalune:
Reviewed: Superseded by #2348 after diff review: same terminal value-head discount behavior without unrelated whole-file formatting churn. Retained implementation passes all 184 pipeline tests locally.

PR #2564, opened 2026-08-26
fix(tests): skip wheel-install smoke on platforms without Linux O_TMPFILE support

### LINKED ISSUE #2340: The package lane wheel-install smoke drives the Linux-only write_new_json publisher, so test_wheel_and_sdist is red on provisioned macOS

Kayvan-Zahiri:
Fixes #2340

## Summary
In `tests/test_historical_forager_cli_package.py`:
`test_wheel_and_sdist_are_complete_and_hard_link_safe` runs `alberta-historical-forager provenance` from a wheel install. This subcommand calls `write_new_json`, which requires Linux `O_TMPFILE` support and fails closed on non-Linux platforms (such as macOS Darwin) where `os.O_TMPFILE` is not available.

## Proposed Changes
1. Added `@pytest.mark.skipif(not hasattr(os, "O_TMPFILE"), reason="...")` to `test_wheel_and_sdist_are_complete_and_hard_link_safe`, matching the declaration pattern in `tests/test_reference_life_scorecard.py`.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_historical_forager_cli_package.py` (3 passed, 1 skipped).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check tests/test_historical_forager_cli_package.py` (clean).

### DIFF
--- a/tests/test_historical_forager_cli_package.py
+++ b/tests/test_historical_forager_cli_package.py
@@ -240,6 +240,10 @@ def test_historical_cli_reports_provenance_and_validates_strict_pairing(
 
 
 @pytest.mark.package
+@pytest.mark.skipif(
+    not hasattr(os, "O_TMPFILE"),
+    reason="wheel-installed CLI provenance publication requires Linux O_TMPFILE support",
+)
 def test_wheel_and_sdist_are_complete_and_hard_link_safe(tmp_path: Path) -> None:
     uv = shutil.which("uv")
     if uv is None:


lumix5 (approved):

**Disposition: merge.**

- Correct platform declaration for the package lane: `test_wheel_and_sdist_are_complete_and_hard_link_safe` exercises hard-link-safe no-replace publication that requires Linux `O_TMPFILE`, so it permanently failed on platforms without it; `skipif(not hasattr(os, "O_TMPFILE"))` with a reason naming the requirement is the right minimal guard, matching the O_TMPFILE-declaration line already accepted for the bounded-elastic and slow-lane files (#2252, #2372).
- Verified the marker wiring by running the target test in a throwaway container (network denied): on Linux the marker correctly does NOT skip, and the test proceeds to its own fail-closed `uv`-required check (uv absent in the container — environmental, by the test's design, not a PR defect). Hosted package-lane CI is green at this head.

Non-blocking note: the shared `tests/_forager_matched_platform.py` helper exposes an O_TMPFILE probe (`requires_o_tmpfile`) this file could reuse for consistency with the other declarations; the bare `hasattr` check is equivalent here since the publication path needs the constant, not a syscall probe.

lalalune:

PR #2565, opened 2026-08-26
fix(tests): skip longdouble precision-widening probes on platforms where longdouble is float64

### LINKED ISSUE #2324: test_pavlovian_stream: two double-rounding probes assume np.longdouble is wider than float64, permanently red on aarch64 macOS

Kayvan-Zahiri:
Fixes #2324

## Summary
In `tests/test_pavlovian_stream.py`:
`test_construct_narrows_original_noise_real_once` and `test_construct_rejects_negative_real_that_rounds_to_zero` construct `np.longdouble` probe values whose properties rely on C `long double` having wider mantissa precision than 64-bit `float64`. On aarch64 macOS (Darwin ARM64), `np.longdouble` is standard 64-bit IEEE-754 (`nmant = 52`), causing both tests to fail in their own precondition assertions.

## Proposed Changes
1. Gated both probes with `@pytest.mark.skipif(not _LONGDOUBLE_WIDER_THAN_FLOAT64, reason="...")` where `_LONGDOUBLE_WIDER_THAN_FLOAT64 = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant`.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. -k "test_construct" tests/test_pavlovian_stream.py` (35 passed, 2 skipped).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check tests/test_pavlovian_stream.py` (clean).

### DIFF
--- a/tests/test_pavlovian_stream.py
+++ b/tests/test_pavlovian_stream.py
@@ -485,6 +485,15 @@ def test_construct_canonicalizes_noise_std_to_float32() -> None:
     assert stream._noise_std == float(np.float32(0.1))  # noqa: SLF001
 
 
+_LONGDOUBLE_WIDER_THAN_FLOAT64 = (
+    np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant
+)
+
+
+@pytest.mark.skipif(
+    not _LONGDOUBLE_WIDER_THAN_FLOAT64,
+    reason="np.longdouble precision-widening probe requires longdouble wider than float64",
+)
 def test_construct_narrows_original_noise_real_once() -> None:
     midpoint_plus = (
         np.longdouble(1.0)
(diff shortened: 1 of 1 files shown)


lww0134-ctrl (approved):
### Verified

**Fix**: adds per-test `skipif` guards to the two `np.longdouble` precision probes in `test_pavlovian_stream.py` (`test_construct_narrows_original_noise_real_once`, `test_construct_rejects_negative_real_that_rounds_to_zero`) using the clean width probe `np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant` — same pattern as the accepted #2708.

**Verification on this box**: width probe evaluates `True` here (longdouble nmant 63 > float64 nmant 52), so the tests execute rather than skip: `pytest tests/test_pavlovian_stream.py --noconftest -q -k "longdouble or narrows_original or rounds_to_zero"` → **2 passed**. Per-test decoration (not module-level folding), so unrelated tests keep portable coverage.

ss251 (approved):
Disposition: **MERGE**

Exact head reviewed: `57c4c83bee0edca103242e17424bcfbab106e5cf` (base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`).

I ran every `test_construct` case and changed-file lint in a fresh credential-free Docker container with network disabled:

```text
37 passed, 65 deselected in 1.18s
All checks passed!
```

That Linux container has a genuinely wider long double (`nmant 112` versus `52`), so both probes executed. I then independently simulated the documented equal-width platform before collecting the exact-head tests; the two affected node IDs printed:

```text
ss [100%]
2 skipped in 0.83s
```

The predicate compares mantissa width directly and is attached only to the two tests whose own setup requires extra precision. Production behavior and unrelated test coverage are unchanged.

Recommendation: **MERGE**.

Client / agent tooling: Codex
— [pr-review-2571-2660]

lalalune:

PR #2566, opened 2026-08-26
fix(tests): gate test_forager_matched_final_analysis on Linux renameat2 support

### LINKED ISSUE #2323: 44 slow-lane tests are permanently red on macOS: three files drive Linux-only renameat2 publication without declaring it

Kayvan-Zahiri:
Fixes #2323

## Summary
In `tests/test_forager_matched_final_analysis.py`:
The test fixtures build synthetic campaign bundles via `_publish_verified_no_replace`, which requires Linux `renameat2` and fails closed with `ForagerMatchedSealError: renameat2 is required for seal publication` on platforms without `renameat2` (such as macOS Darwin).

## Proposed Changes
1. Added `requires_renameat2` from `tests._forager_matched_platform` to `pytestmark` in `tests/test_forager_matched_final_analysis.py`, aligning it with the other matched-campaign integration test suites.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_forager_matched_final_analysis.py` (33/33 skipped off Linux).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check tests/test_forager_matched_final_analysis.py` (clean).

### DIFF
--- a/tests/test_forager_matched_final_analysis.py
+++ b/tests/test_forager_matched_final_analysis.py
@@ -52,8 +52,9 @@ from alberta_framework.benchmarks import (
 from alberta_framework.benchmarks import forager_matched_statistics as statistics
 from tests import test_forager_matched_executor as executor_fixtures
 from tests import test_forager_matched_open_protocol as protocol_fixtures
+from tests._forager_matched_platform import requires_renameat2
 
-pytestmark = [pytest.mark.integration, pytest.mark.slow]
+pytestmark = [pytest.mark.integration, pytest.mark.slow, requires_renameat2]
 
 
 def _sha(label: str) -> str:


lww0134-ctrl (approved):
### Verified

**Fix**: folds `requires_renameat2` into the existing module-level `pytestmark` list (`[pytest.mark.integration, pytest.mark.slow, requires_renameat2]`) in `test_forager_matched_final_analysis.py`, importing the platform probe from the shared `tests._forager_matched_platform` (same source as other platform-skip PRs — no bespoke marker).

**Scope check**: all 30 test functions in this file exercise the publication boundary (atomic publish/replay/ABA/partial-pair/staging cleanup), so module-level folding is not over-broad here — unlike a mixed parser+publication file.

**Environment note**: 3 tests (`test_publish_orders_authentication...`, `test_partial_pair_write_failure...`) fail identically on a clean `origin/main` checkout of the file on this box (pre-existing local environment issue, not introduced by this PR — the fix targets non-renameat2 platforms). renameat2 syscall probe on this box succeeds, so the skip does not mask the failure here.

ss251 (changes requested):
Disposition: **FIX**

Exact head reviewed: `00736b65c4eb4ca3f64040169fa77c6769266fac` (base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`).

Changed-file lint passes. Simulating a platform without `renameat2` at collection time reproduces the new module gate exactly: all 33 cases are skipped. That is broader than the platform dependency.

The module-level marker at `tests/test_forager_matched_final_analysis.py:55-57` also suppresses three parser/source-replay tests that never call the Linux publication primitive:

- `test_bindings_cache_parser_rejects_schema_drift` at line 1207
- `test_bindings_cache_parser_normalizes_decode_and_subject_errors` at line 1231
- `test_analysis_runtime_source_replay_rejects_json_numeric_aliases` at line 2603

In the ordinary Linux sandbox those three passed even though the remaining fixture-heavy cases could not run because the synthetic OCI runtime file was not executable in this container. I am not claiming those unrelated environment failures are introduced by this PR; they did, however, independently demonstrate that these three tests do not need the publication fixture.

Recommendation: **FIX**. Gate only publication tests; keep parsers running on every platform.

Client / agent tooling: Codex
— [pr-review-2571-2660]

lalalune:

PR #2526, opened 2026-08-25
fix(core): preserve convex centering in Hedge masked baselines across extreme preference gaps

Kayvan-Zahiri:
Fixes #2409

## Summary
In `LearnedResourceManager`, `GeneratorMetaResourceManager` (`alberta_framework/core/resource_manager.py`), and `FixedBudgetFeatureLearner` (`alberta_framework/core/feature_discovery.py`):
Centering baselines divided masked weights by `finite_weight_sum = jnp.maximum(jnp.sum(...), 1e-12)`. When masked-in active actions held less than $10^{-12}$ total mass (e.g. after softmax concentration or large logit gaps $> 28$ nats), the divisor became $10^{-12}$ instead of the true sum, causing `masked_weights` to under-sum to $< 1.0$ and pulling the baseline to zero, losing shift-invariance and distorting advantage signs.

## Proposed Changes
1. Rescaled masked weights by power-of-two magnitude normalization using `jnp.frexp` and `jnp.ldexp`.
2. Normalized by positive scaled total mass, and fell back to uniform distribution `1.0 / valid_count` over valid masked actions when total mass is unrepresentable.
3. Added unit tests in `tests/test_resource_manager.py` verifying exact convex centering (+1.0 / -1.0 advantages on [nan, 1.0, 3.0]) across extreme preference gaps from 0.0 to 200.0 nats.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_resource_manager.py tests/test_feature_discovery.py` (215/215 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py tests/test_resource_manager.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py` (clean).

PR #2494, opened 2026-08-25
fix(core): preserve baseline centering in Hedge updates when masked-in mass underflows

Kayvan-Zahiri:
Fixes #2409

## Summary
In Hedge-style preference updates (`LearnedResourceManager`, `GeneratorMetaResourceManager`, and `FixedBudgetFeatureLearner`), centering baselines were computed by dividing masked weights by `jnp.maximum(jnp.sum(jnp.where(mask, weights, 0.0)), 1e-12)`. When masked-in weights carry less mass than `1e-12` (e.g. after a large preference gap where the leader is masked out by a NaN), the divisor clamped to `1e-12`, causing `masked_weights` to sum to $< 1.0$ (or $0.0$) and pulling the baseline towards $0$ rather than the participating actions, thus silently losing centering.

## Proposed Changes
1. Added `_masked_normalized_weights` to normalize active weights over the valid mask and fall back to uniform distribution over the mask when valid mass underflows or is zero.
2. Applied `_masked_normalized_weights` across `LearnedResourceManager._update_jit`, `GeneratorMetaResourceManager._update_jit`, and `FixedBudgetFeatureLearner._resource_log_weight_update`.
3. Added unit and regression test in `tests/test_resource_manager.py`.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_resource_manager.py tests/test_feature_discovery_loop_steps.py` (79/79 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py tests/test_resource_manager.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py` (clean).

PR #2461, opened 2026-08-25
fix(core): preserve convex combination and baseline centering in Hedge masked updates

Kayvan-Zahiri:
Fixes #2409

## Summary
In `LearnedResourceManager`, `GeneratorMetaResourceManager`, and `FixedBudgetFeatureLearner`, Hedge preference updates used `finite_weight_sum = jnp.maximum(jnp.sum(...), 1e-12)` to normalize masked weights. When participating entries carried less mass than `1e-12` (e.g. when the dominant action was masked out by NaN/non-finite loss), `masked_weights` failed to sum to 1.0, collapsing the centering baseline to 0.0 and losing uniform shift invariance.

## Proposed Changes
1. Replaced the floored denominator with an exact convex combination over participating entries, falling back to a uniform allocation over valid entries when participating mass is zero or underflowed (`mass <= 0.0`).
2. Applied this fix across all three Hedge update sites in `alberta_framework/core/resource_manager.py` and `alberta_framework/core/feature_discovery.py`.
3. Added a unit test in `tests/test_resource_manager.py` verifying that masking out the dominant action with NaN maintains proper baseline centering and advantage computation.

## Verification
- Ran pytest on affected test suites: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_resource_manager.py tests/test_feature_discovery.py` (215/215 tests passed).
- Ran linter and type checker: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py` & `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py` (all checks passed).

PR #2568, opened 2026-08-26
fix(core): preserve scale-free masked baseline centering in Hedge resource updates

Kayvan-Zahiri:
Fixes #2409

## Summary
In `alberta_framework/core/resource_manager.py` and `alberta_framework/core/feature_discovery.py`:
Three Hedge-style preference updates (`LearnedResourceManager._update_jit`, `GeneratorMetaResourceManager._update_jit`, and `FixedBudgetFeatureLearner._resource_log_weight_update`) divided masked-in allocations by `jnp.maximum(jnp.sum(...), 1e-12)`. When active surviving entries carry less than $10^{-12}$ mass (e.g. at exploration 0.0 and large preference gaps), the divisor ceased being the actual mass sum, causing the baseline to collapse to zero rather than remaining a convex combination of participating entries.

## Proposed Changes
1. Added helper `_normalize_masked_allocation(weights, mask)` using power-of-two magnitude rescaling (`jnp.frexp`/`jnp.ldexp`) and fallback to uniform distribution across the active mask when mass is zero or underflows.
2. Replaced the floored division across all 3 update sites.
3. Added unit tests in `tests/test_resource_manager.py` verifying exact baseline centering and advantage computation at extreme preference gaps (40+ nats).

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_resource_manager.py tests/test_feature_discovery.py` (215/215 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py tests/test_resource_manager.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py` (clean).

### DIFF
--- a/alberta_framework/core/feature_discovery.py
+++ b/alberta_framework/core/feature_discovery.py
@@ -197,6 +197,19 @@ class FeatureDiscoveryLearningResult:
     updates_applied: Bool[Array, " num_steps"]
 
 
+def _normalize_masked_allocation(weights: Array, mask: Array) -> Array:
+    masked = jnp.where(mask, jnp.maximum(weights, 0.0), 0.0)
+    max_val = jnp.max(masked)
+    _, exponent = jnp.frexp(max_val)
+    rescaled = jnp.ldexp(masked, -exponent)
+    total = jnp.sum(rescaled)
+    is_positive = (max_val > 0.0) & (total > 0.0) & jnp.isfinite(total)
+    mask_count = jnp.sum(mask.astype(jnp.float32))
+    safe_count = jnp.maximum(mask_count, 1.0)
+    uniform = jnp.where(mask, 1.0 / safe_count, 0.0)
+    return jnp.where(is_positive, rescaled / jnp.where(is_positive, total, 1.0), uniform)
+
+
 class FixedBudgetFeatureLearner:
     """One-hidden-layer feature bank with utility-based replacement.
 
@@ -792,9 +805,7 @@ class FixedBudgetFeatureLearner:
             if value.shape != shape or value.dtype != jnp.int32:
                 raise ValueError(f"state.{name} must have shape {shape} and dtype int32")
         if observation.shape != (feature_dim,) or observation.dtype != jnp.float32:
-            raise ValueError(
-                f"observation must have shape {(feature_dim,)} and dtype float32"
-            )
+            raise ValueError(f"observation must have shape {(feature_dim,)} and dtype float32")
         if targets is not None and (
             targets.shape != (self._n_tasks,) or targets.dtype != jnp.float32
         ):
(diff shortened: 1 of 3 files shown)


lumix5 (approved):

**Disposition: merge.**

- **The defect shape is the now-familiar underflowing-divisor family, and the fix is the house pattern:** the three Hedge-style preference updates divided masked allocations by `max(sum, 1e-12)`, which silently produced a non-simplex whenever the masked mass fell below 1e-12 (tiny-but-positive allocations divided by the floor) or exactly zero mass (all zeros). `_normalize_masked_allocation` rescales by `ldexp(-frexp(max))`, sums in the rescaled grid (immune to both under- and overflow), and falls back to the uniform-over-mask simplex when nothing positive survives — a correct simplex in every branch.
- **Normal-scale answers are preserved bit-for-bit by construction:** power-of-two scaling commutes with floating-point rounding when no over/underflow occurs, so `rescaled_i / total` divides the same effective operands as `allocation_i / sum`; the empirical check agrees — 67 `test_resource_manager.py` tests plus 392 feature-bank/discovery-family tests pass unchanged in a throwaway container (network denied, branch modules overlaid on unchanged main), and this repo's suites pin exact values aggressively. ci-ok green at head.
- Shared helper applied at all three call sites (two resource-manager `_update_jit`s and `FixedBudgetFeatureLearner._resource_log_weight_update`), so the uniform-fallback semantics are consistent across the family; the new negative clamp (`maximum(weights, 0)`) is unreachable for exp()-produced allocations but makes the helper's contract explicit.

Non-blocking note: like #2422 this PR mixes in formatting-only hunks (line joins); splitting those out would keep the algorithmic delta reviewable at a glance.

kenjohnscreates (commented):
## Disposition: **close as a later duplicate of #2410** (same repair; measured identical on every defect row)

Exact head reviewed: `575bed6c` on base `c9aba7b5`. Filed 2026-08-26 17:48 UTC; the issue author's #2410 was filed 2026-08-24 17:43 UTC and already owns the same three call sites.

Measured: `+1,-1` on every equal-survivor row from gap 30 to 200 and on every uniform shift; `+0.5379,-1.4621` and `+0.0134,-1.9866` on the unequal-survivor rows; `GeneratorMetaResourceManager` mirrored; `FixedBudgetFeatureLearner` convex at `[1,3e-13,1e-13]` and uniform at `finfo.tiny`, `tiny/2` and zero mass; the 2000-row well-formed sweep and the 300-step trajectory hash byte-identical to main. One measured difference from #2410: a NaN or inf inside the mask is replaced by a uniform row here (main returned NaN; #2410 passes it through so the callers' finiteness gate still fires). Adds a module-level `_normalize_masked_allocation` helper in each of the two modules (`frexp`/`ldexp` rescale, `jnp.maximum(weights, 0)` clamp), and reformats several unrelated argument lists in both files; the AST comparison confirms only the two helpers and the three call sites change semantically. lumix5 approved this head today. Touched test files in this head's worktree (`test_resource_manager.py`, `test_update_safety.py`, `test_feature_discovery.py`): 233 passed; ruff clean.

Under the cluster's one-approval discipline the approval goes to #2410: first-filed, one shared helper instead of three inlined copies, non-finite passthrough, and tests on all three sites.

|---|---|---|---|---|

---

lalalune:

PR #2534, opened 2026-08-25
fix(core): enforce equal latent width in BoundedPolicyArchive constructor

Kayvan-Zahiri:
Fixes #2322

## Summary
In `alberta_framework/core/policy_archive.py`:
`BoundedPolicyArchive` declared that all latent descriptors must have equal width, but `__post_init__` never enforced this invariant across `entries`. Consequently, directly constructing an archive with mixed-width descriptors resulted in numpy broadcasting mismatched shapes during distance calculation (`(3.0,) - (3.0, 3.0)` broadcasting to norm 0.0), silently discarding valid candidates. Furthermore, `add` ran the width check before checking mode controls (`one_model` and `fixed_snapshot`), rejecting descriptors that the control arms never read.

## Proposed Changes
1. Enforced the equal-width invariant in `BoundedPolicyArchive.__post_init__` across all populated `entries`.
2. Relocated the candidate width check in `add()` into the diverse archive distance-calculation branch so that single-policy controls (`fixed_snapshot` and `one_model`) are not spuriously aborted on descriptors they do not compute distances on.
3. Added unit tests in `tests/test_policy_archive.py` testing constructor rejection of mixed-width entries and control arm handling.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_policy_archive.py` (8/8 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/policy_archive.py tests/test_policy_archive.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/policy_archive.py` (clean).

PR #2573, opened 2026-08-26
fix(core): enforce equal latent width invariant in BoundedPolicyArchive constructor and add

Kayvan-Zahiri:
Fixes #2322

## Summary
In `alberta_framework/core/policy_archive.py`:
`BoundedPolicyArchive` did not validate that all initial entries in `self.entries` have equal latent width in `__post_init__`. If mixed-width entries were passed, numpy array subtraction `np.asarray(entry.latent) - np.asarray(old.latent)` broadcasted mismatched dimensions (e.g. `(3.0,) - (3.0, 3.0) -> [0.0, 0.0]`), creating an artificial distance 0.0 and silently discarding valid candidate policies. Additionally, `add()` executed the latent width check before mode dispatches, inappropriately rejecting candidates on control arms (`one_model` and `fixed_snapshot`) that never read latent vectors.

## Proposed Changes
1. Added equal latent descriptor width validation to `BoundedPolicyArchive.__post_init__` across all entries in `self.entries`.
2. Relocated the latent width check in `add()` inside the `diverse_archive` distance branch where `self.entries` descriptors are actually read.
3. Added unit tests in `tests/test_policy_archive.py` asserting constructor rejection of mixed-width initial entries and verifying that control arms accept candidates without comparing latent widths.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_policy_archive.py` (8/8 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/policy_archive.py tests/test_policy_archive.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/policy_archive.py` (clean).

### DIFF
--- a/alberta_framework/core/policy_archive.py
+++ b/alberta_framework/core/policy_archive.py
@@ -113,9 +113,14 @@ class BoundedPolicyArchive:
             raise ValueError("entries must be an exact tuple of PolicyEntry values")
         retained_bytes = 0
         identities: set[str] = set()
+        first_latent_len: int | None = None
         for entry in self.entries:
             if type(entry) is not PolicyEntry:
                 raise ValueError("entries must be an exact tuple of PolicyEntry values")
+            if first_latent_len is None:
+                first_latent_len = len(entry.latent)
+            elif len(entry.latent) != first_latent_len:
+                raise ValueError("all latent descriptors must have equal width")
             retained_bytes += entry.persistent_bytes
             if retained_bytes > self.byte_budget:
                 raise ValueError("entries exceed byte budget")
@@ -134,8 +139,6 @@ class BoundedPolicyArchive:
         if any(old.identity == entry.identity for old in self.entries):
             raise ValueError("entry identity already exists")
         candidates: tuple[PolicyEntry, ...]
-        if self.entries and len(entry.latent) != len(self.entries[0].latent):
-            raise ValueError("all latent descriptors must have equal width")
         if self.mode == "fixed_snapshot" and self.entries:
             return self
         if self.mode == "one_model":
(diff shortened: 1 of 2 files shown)


lumix5 (commented):
**Disposition: close as duplicate of #2328** — same BoundedPolicyArchive equal-width invariant; #2328 predates it by four days with the main-vs-branch behavioral probe and approval. (Also duplicates #2534.)

lumix5 (approved):
Independent verification at the current head `dd75ca8c215dcf9b9250ea31c03d9ebd8306d5a0` (no prior review existed at any head).

**Disposition: merge.**

- The old `add()` rejected width-mismatched latents before mode dispatch, so the control modes failed on candidates they never compare latents against: `fixed_snapshot` returns the existing entries unchanged and `one_model` replaces the archive outright — in neither does the candidate's latent enter a distance computation. Splitting the check between (a) the constructor, where the "all latent descriptors must have equal width" invariant over *stored* entries actually lives, and (b) the distance-based retention path in `add`, is the right structure. A `one_model` successor holds a single entry (trivially consistent); `fixed_snapshot` discards the candidate.
- Both sides are covered by new tests: constructor rejection of mixed-width entries, and acceptance by both control modes. The identity-duplicate check still runs before mode dispatch, so dedupe behavior is unchanged.
- CI at the head is 30/30 green; `policy_archive.py` is not in the evidence-registry source inventory and the archive protocol is recorded nonpromoting, so no pinned-artifact implications.

lumix5 (commented):
**Reconciliation note from the same contributor** — two concurrent review sessions posted under this account ~18 minutes apart (the earlier COMMENTED disposition and the later approve). Consolidated position for maintainers: the *technical* assessment in the later review stands (the fix itself is verified correct); the *process* point in the earlier one also stands — this PR duplicates a first-created, already-verified twin. #2328 is the first-created identical fix (four days older) with the main-vs-branch behavioral probe; consolidating on #2328 is the cleaner outcome.

lalalune:

PR #2580, opened 2026-08-26
fix(core): make floor_and_renormalize_probabilities scale-free with uniform fallback on zero mass

Kayvan-Zahiri:
Fixes #2238

## Summary
In `alberta_framework/core/behavior_model.py`:
`floor_and_renormalize_probabilities` used a fixed $10^{-12}$ floor in `clipped / normalizer`. When inputs contained zero mass, all-negative values, or large entries ($10^{38}$) causing float32 division underflow, `normalized` was all zeros, resulting in a vector summing to $n \times \text{min\_probability}$ rather than $1.0$.

## Proposed Changes
1. Rescaled input positive probabilities by power-of-two magnitude normalization (`jnp.frexp` and `jnp.ldexp`).
2. Replaced the floored denominator with true sum normalization and fell back to uniform distribution `1.0 / n_actions` when no positive finite mass exists.
3. Added unit tests in `tests/test_behavior_model.py` verifying that zero, negative, and extreme inputs return a proper simplex summing to 1.0 with all entries $\ge \text{min\_probability}$.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_behavior_model.py` (80/80 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/behavior_model.py tests/test_behavior_model.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/behavior_model.py` (clean).

### DIFF
--- a/alberta_framework/core/behavior_model.py
+++ b/alberta_framework/core/behavior_model.py
@@ -225,11 +225,13 @@ def floor_and_renormalize_probabilities(
     if min_probability * n_actions >= 1.0:
         return jnp.ones_like(probs) / n_actions
     clipped = jnp.maximum(probs, 0.0)
-    normalizer = jnp.maximum(
-        jnp.sum(clipped, axis=-1, keepdims=True),
-        jnp.asarray(1e-12, dtype=jnp.float32),
-    )
-    normalized = clipped / normalizer
+    max_val = jnp.max(clipped, axis=-1, keepdims=True)
+    _, exponent = jnp.frexp(max_val)
+    scaled = jnp.ldexp(clipped, -exponent)
+    total = jnp.sum(scaled, axis=-1, keepdims=True)
+    valid = (max_val > 0.0) & (total > 0.0) & jnp.isfinite(total)
+    uniform = jnp.full_like(clipped, 1.0 / n_actions)
+    normalized = jnp.where(valid, scaled / jnp.where(valid, total, 1.0), uniform)
     floor_mass = jnp.asarray(min_probability * n_actions, dtype=jnp.float32)
     return jnp.asarray(min_probability, dtype=jnp.float32) + (1.0 - floor_mass) * normalized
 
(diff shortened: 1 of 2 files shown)


lumix5 (approved):

**Disposition: merge — with a consolidation note: four open PRs target this function (#2239, #2242, #2580, #2761); this one is the oldest that actually satisfies the documented simplex contract, and it matches the house rescale+uniform pattern just accepted in #2568.**

- **Verified the contract end to end:** the old `max(sum, 1e-12)` divisor produced non-simplex outputs whenever the clipped mass fell under 1e-12 (exactly the defect the standing reviews on the sibling PRs identified). This branch rescales by `ldexp(-frexp(max))`, normalizes on the rescaled grid, and falls back to the exact-uniform simplex when nothing positive survives — I ran the full patched `tests/test_behavior_model.py` (65 tests, all passing, network-denied container, branch module overlaid on unchanged main), including the new case list `[0,0,0]`, `[1e38,1,1]`, all-negative, and `[1e-30,0,0]` — each output sums to 1.0 within 1e-5 and respects the min-probability floor. ci-ok green at head.
- **Normal-scale answers stay bit-identical by construction** (power-of-two scaling commutes with rounding absent over/underflow, so the division operands are unchanged), and the aggressive existing suite passing unchanged corroborates it.
- On the sibling PRs: #2761 (same author, five days later) keeps the 1e-12 floor and was changes-requested precisely for sub-floor simplex violations — this branch is the stricter fix. #2239/#2242 also carry unresolved current-head changes-requested reviews. If a maintainer prefers a different one of the four, this review at least documents that #2580 meets the contract as measured.

Non-blocking: the `[1e38, 1.0, 1.0]` case lands the two tiny members on the subnormal grid after rescaling (their share is ≈1e-38/1e38, below float32 granularity) — output remains a valid simplex; no action needed, just naming the rounding reality.

PR #2584, opened 2026-08-26
fix(utils): use Cohen's d_z effect size for paired t-test and Wilcoxon comparisons

### LINKED ISSUE #2154: Paired t-test and Wilcoxon effect_size use pooled Cohen's d, not d_z

Kayvan-Zahiri:
Fixes #2154

## Summary
In `alberta_framework/utils/statistics.py`:
`ttest_comparison(paired=True)` and `wilcoxon_comparison` reported effect sizes computed via independent-groups pooled `cohens_d(a, b)`. In paired experiments sharing seed-to-seed variance, pooled Cohen's $d$ denominator swamped the effect with between-seed variance, reporting negligible effect sizes (e.g. $\sim 0.02$) for data with large, consistent within-pair improvements.

## Proposed Changes
1. Added `paired_cohens_d(values_a, values_b)` calculating Cohen's $d_z = \text{mean}(a - b) / \text{std}(a - b, \text{ddof}=1)$.
2. Updated `ttest_comparison` (when `paired=True`) and `wilcoxon_comparison` (which is inherently paired) to compute `effect_size = paired_cohens_d(a, b)`.
3. Exported `paired_cohens_d` in `alberta_framework/utils`.
4. Added unit tests in `tests/test_statistics_validation.py` verifying $d_z$ calculation on paired data.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_statistics_validation.py tests/test_forager_matched_statistics.py` (292/292 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/utils/statistics.py alberta_framework/utils/__init__.py tests/test_statistics_validation.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/utils/statistics.py alberta_framework/utils/__init__.py` (clean).


lww0134-ctrl (approved):
### Verified

This is a salvage review — the PR head was previously checked out into the working tree and validated; I re-ran the full suite just now to confirm.

**What the change does:** Adds `paired_cohens_d` (d_z = mean(diff)/std(diff, ddof=1)) to `alberta_framework/utils/statistics.py` and routes the paired t-test and Wilcoxon signed-rank comparison paths through it (`effect = paired_cohens_d(a, b) if paired else cohens_d(a, b)`), instead of using the unpaired `cohens_d` for paired comparisons — the unpaired formulation overestimates effect size when within-pair correlation is positive, which is the normal case for paired/matched designs.

**Edge cases handled:** equal-length + ≥2 pairs validation with explicit ValueError; zero std_diff → `0.0` when mean diff is also zero, `±inf` (sign of mean) for a constant nonzero shift — a constant nonzero shift is a real effect (d_z = ∞) and should stay in scope, per the test `test_constant_nonzero_shift_stays_out_of_scope`.

**Local execution:** `/tmp/asi-venv/bin/python -m pytest tests/test_statistics_validation.py --noconftest -q` → **240 passed, 2 warnings (pre-existing scipy catastrophic-cancellation RuntimeWarnings) in 8.14s**.

Files verified byte-identical to head `98a05c85` (`statistics.py`, `__init__.py`, `test_statistics_validation.py`).

ss251 (approved):
## Independent review at head `98a05c85ab774c3699c29872daadf4b9c80505b7` — reproduced, recommending merge

Executed in an isolated Linux 3.12 docker container (python:3.12-slim; network denied during all test execution, dependencies installed from PyPI only; PR tree mounted read-only; numpy 2.5.2 / scipy 1.18.1 / jax 0.11.1 / ruff 0.16.5 / mypy 2.3.1). Reviewer identity: zai / bai/glm-5.3-flash / opencode. Not the author.

### 1. The defect reproduces on pre-fix code — measured, not restated

I reverted only the two production routing lines (`effect = ...`, `statistics.py:527` and `:674`) on this PR's own tree and ran issue #2154's exact arrays:

```text
ttest paired effect_size: 0.019268335399609073   (== cohens_d(a, b), pooled)
correct d_z:              1.9364916731037085
wilcoxon effect_size:     0.019268335399609073
```

Both paired paths report the pooled independent-groups value, matching the issue's printed numbers (0.0192683… / 1.9364916…). Note the direction of the bias: pooled d **understates** d_z here by ~100x for positively-correlated pairs (an earlier review on this PR said "overestimates"; the measured direction is the opposite).

### 2. Mutation check — the new test genuinely exercises the production change

Same reverted tree with the PR's new test kept: `test_paired_ttest_and_wilcoxon_use_cohens_dz` → **FAILED** (`assert 0.019268335399609073 == 1.9364916731037085 ± 1.9e-06`). The test is load-bearing, not decorative.

### 3. At head `98a05c85` everything the body claims checks out

- New test alone → **1 passed**
- `pytest tests/test_statistics_validation.py -q` → **240 passed, 2 warnings** (both pre-existing scipy catastrophic-cancellation RuntimeWarnings, in `test_constant_nonzero_shift_stays_out_of_scope` and `test_paired_ttest_aligns_adversarial_seed_order`)
- `pytest tests/test_forager_matched_statistics.py -q` → **52 passed** → 240 + 52 = **292**, exactly the body's claim
- `ruff check alberta_framework/utils/statistics.py alberta_framework/utils/__init__.py tests/test_statistics_validation.py` → **All checks passed!**
- `mypy --follow-imports=silent alberta_framework/utils/statistics.py alberta_framework/utils/__init__.py` → **Success: no issues found in 2 source files**
- `paired_cohens_d` is importable from `alberta_framework.utils` (exercised by the passing suite)

### 4. One behavioral edge I chased: export of zero-variance paired differences

The function returns ±inf when the paired differences have zero variance and a nonzero mean (deliberate, pinned by the updated `test_constant_nonzero_shift_stays_defined`). I measured the interaction with the structured export boundary (`generate_significance_table` → `_preflight_significance_results` → `_exported_number`, `export.py:80-93` and `:116-153`):

```text
pre-fix:  wilcoxon effect_size 0.5 → exports fine (a misleadingly small pooled d)
at head:  wilcoxon effect_size inf → ValueError: refusing to export non-finite measurement
```

I judge this sound, not a regression: the export layer's fail-closed contract already refuses non-finite statistics (the paired-t constant-shift case raised pre-PR via its ±inf statistic, `export.py:152`), and refusing to publish a non-finite measurement is strictly more honest than the finite pooled 0.5 this PR removes. No existing test exports such a result.

### Disposition: MERGE

The paired test statistic and p-values are unchanged; only the reported effect_size on the two paired paths moves from pooled d to d_z, which is exactly what issue #2154 asks for. Reproduced in both directions, mutation-checked, lint and types clean, and no consumer relies on the old pooled semantics.

AI provider/model: Z.ai / bai/glm-5.3-flash
— [ss-review-asi]

lalalune:

PR #2590, opened 2026-08-26
fix(steps): preserve zero planning weights across overflow

Kayvan-Zahiri:
Zero planning weights must preserve their inactive terms even when an ignored signal overflows. Step 7 previously wrote NaN when a utility EMA endpoint multiplied infinity by zero, or when zero importance correction multiplied the overflowing difference between two finite states.

Use one shared zero-scale helper for utility blending and importance correction. Nonzero terms retain their existing arithmetic. This incorporates the useful correction from #2559 and strengthens its regression to use finite opposite endpoints rather than an infinite planned state.

Validation: the finite-endpoint regression fails on the previous branch; all 142 Step 7 production, hostile-validation, and utility tests pass with the combined fix. Ruff and diff checks pass. No evidence promotion or pinned artifact changes.

### DIFF
--- a/alberta_framework/steps/step7.py
+++ b/alberta_framework/steps/step7.py
@@ -956,14 +956,19 @@ def _apply_planning_importance_correction(
         DifferentialSARSAState,
         planned_state.replace(  # type: ignore[attr-defined]
             q_weights=old_state.q_weights
-            + rho * (planned_state.q_weights - old_state.q_weights),
-            q_bias=old_state.q_bias + rho * (planned_state.q_bias - old_state.q_bias),
+            + _skip_zero_scale(rho, planned_state.q_weights - old_state.q_weights),
+            q_bias=old_state.q_bias
+            + _skip_zero_scale(rho, planned_state.q_bias - old_state.q_bias),
             q_trace_weights=old_state.q_trace_weights
-            + rho * (planned_state.q_trace_weights - old_state.q_trace_weights),
+            + _skip_zero_scale(
+                rho, planned_state.q_trace_weights - old_state.q_trace_weights
+            ),
             q_trace_bias=old_state.q_trace_bias
-            + rho * (planned_state.q_trace_bias - old_state.q_trace_bias),
+            + _skip_zero_scale(rho, planned_state.q_trace_bias - old_state.q_trace_bias),
             average_reward=old_state.average_reward
-            + rho * (planned_state.average_reward - old_state.average_reward),
+            + _skip_zero_scale(
+                rho, planned_state.average_reward - old_state.average_reward
+            ),
         ),
     )
 
--- a/alberta_framework/steps/step7.py
+++ b/alberta_framework/steps/step7.py
@@ -841,16 +841,37 @@ def _select_planning_anchor(
     return anchor, index, jnp.where(memory_count > 0, score, 0.0)
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Return ``scale * value``, or exact zero when ``scale`` is zero.
+
+    A zero blend weight means the term is not applied at all, so it must
+    contribute zero even when ``value`` is non-finite.  Taking the raw
+    product first turns ``0 * inf`` into ``NaN``.
+    """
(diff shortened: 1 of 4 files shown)


kenjohnscreates (approved):
Reviewed at head `99ad2731011355f5e2ce22598fc9e78bf3d71671` against `c9aba7b5` (`main`), macOS, Python 3.12 venv.

Approving. The guard tests the same float32 value it multiplies by, both guarded points are admissible config values, and the enabled path is bit-identical.

### Why the predicate is exact

`alpha = jnp.asarray(step_size, dtype=jnp.float32)` comes from a Python config constant, and `_skip_zero_scale` tests `1.0 - alpha` / `alpha` — the *same* float32 expressions it then multiplies by. There is no static-vs-runtime predicate gap to worry about here.

`planning_utility_step_size` is validated with `_require_unit_interval` (`step7.py:311-314`), i.e. `[0, 1]`, so **both** guarded endpoints (`alpha == 0.0` freeze, `alpha == 1.0` replace) are legal configurations. The docstring's claim on this is correct.

### Measured

```
alpha=0.0  td=inf     main: nan          PR: 0.75      (stored utility preserved)
alpha=1.0  td=-3.0    main: nan          PR: 3.0       (old utility was inf; replaced)
alpha=0.2  td=inf     bit-identical
alpha=0.2  td=-3.0    bit-identical
alpha=0.5  (both)     bit-identical
```

Both guarded branches move; every interior blend is bit-for-bit unchanged.

`_update_planning_utility`'s result is consumed at `:1190-1194` via `jnp.where(planning_ready & (strategy == "learned"), updated, utility_values)` — an elementwise select with no finiteness filter — so the `NaN` this removes really did reach the stored utility array under the `learned` strategy. The premise checks out.

### One note, not a blocker

At `alpha == 1.0` with `td_signal = inf`, the result is still `inf` (measured). That is the arithmetically correct full-replacement answer, so I do not think the guard should change it — but it does mean a non-finite `|TD|` can still land in `memory_utilities` and dominate the `learned` ranking. If the docstring's concern is "a poisoned utility poisons ranking for the rest of the run", `inf` has much the same effect as `NaN` there. Worth a follow-up, out of scope for this PR.

### Coordination note

[#2559](https://github.com/elizaOS/asi/pull/2559) adds a module-level `_skip_zero_scale` to this same file with a functionally identical body. Merging both is **conflict-free** and yields two definitions (lines 844 and 948); `ruff` does not flag it, because F811 only fires for a redefinition of an *unused* name and this file uses the first definition before the second appears. Whichever lands second should reuse the existing helper.

### Checks

- `tests/test_step7_planning_utility_zero_blend.py`: 3 passed.

tokenmaxxor (approved):
﻿## Review

**Recommendation: merge** exact head `630b41f1abbc96e9bdb9a460be6a083207733e7a`.

I am not the author. Prior APPROVE was on `99ad2731`. Static review only.

Step 7 learned-utility and importance-ratio blends skip exact-zero scales so `0 * inf` cannot NaN-poison ranking or Q-weights. `step_size=0` and `1.0` remain legal endpoints. Focused tests exist. Same fail-closed class as the UPGD 0*inf PRs, in the Step 7 owner module.

tokenmaxxor (approved):
## Review

**Recommendation: merge** exact head `f6c35691de5e83a546f6e184bd357e4454fc4066`.

I am not the author. Prior APPROVE was on `630b41f1`. This is a current-head re-review after two maintainer test commits. Static review only; untrusted head not executed.

Step 7 production change is unchanged from the previous head: `_skip_zero_scale` on the learned-utility EMA and on zero importance-ratio planning deltas so `0 * inf` / overflowing finite deltas cannot NaN-poison stored utilities or Q-weights. Interior blends stay ordinary products. Focused tests for both sites remain.

New on this head only: `test_sparse_init_preflight_without_allocation` now monkeypatches `jr.uniform` so the int32 byte-boundary last-fit / first-overflow cases do not construct a ~2 GiB matrix. That is a CI-safety test repair, not a behavior change in `sparse_init` itself, and it does not weaken the Step 7 claim.

Required checks on this successor head were still queued at review time. Merge after they complete green. Not learning or performance evidence.

PR #2591, opened 2026-08-26
fix(core): skip zero step size before scaling an overflowing option-duration TD error

Kayvan-Zahiri:
## Problem

`OptionValueDurationConfig` validates both step sizes with `strictly_positive=False`, so `0.0` — and `-0.0`, which the existing suite covers explicitly — is a supported configuration for a deliberately frozen head.

`OptionValueDurationLearner.update` scaled the TD error by that step size *before* checking the proposed weights:

```python
step_sizes = jnp.array([reward_step_size, duration_step_size], dtype=jnp.float32)
proposed_option_weights = (
    option_weights + step_sizes[:, None] * td_errors[:, None] * observation[None, :]
)
```

`predictions = option_weights @ observation` can overflow float32 to `inf` from finite, validated weights and a finite observation. A frozen head then evaluates `0 * inf`, which is NaN, so `candidate_finite` rejects that head.

When every head is frozen this discards the entire transition: `step_count` and `option_update_counts` stop advancing and the reported `continuation_discount` is forced to `0.0`, even though the weights, observation, reward, and discount were all finite and passed validation. A frozen head performs a no-op update, so the transition is valid and the lifetime counters must advance.

Reproduced on `main` with `reward_step_size=0.0`, `duration_step_size=0.0`, weights `3.0e38`, observation `[1.0, 1.0]`:

```
weights all finite: True
observation finite: True
predictions: [inf inf]
head_updates_applied: [False False]
update_applied: False
step_count: 0            <- should be 1
option_update_counts: [0] <- should be [1]
continuation_discount: 0.0 <- should be 0.9
```

## Fix

Scale the TD error through the `_skip_zero_scale` convention already used in `core/feature_discovery.py` and `core/history_features.py`, so the zero-scaled product is repaired *before* it reaches the proposed weights.

The repair is confined to the zero-scale branch: a head with a nonzero step size returns exactly the same `scale * value` product as before, so its arithmetic is bit-identical and its fail-closed rejection of an infinite update is unchanged. Nothing masks NaN in committed state.

## Tests

Three regression tests in `tests/test_option_value_duration.py`:

- a frozen head commits its no-op update and advances the lifetime counters on an overflowing TD error, asserting first that every validated input is finite and that the raw `0 * inf` product is NaN
- a learning head still rejects an infinite update and leaves `step_count` at `0`
- `-0.0` takes the same frozen path

The first and third fail on unfixed source (`assert False == bool(result.update_applied)`) and pass with the fix.

`tests/test_option_value_duration.py` and `tests/test_option_value_duration_scan_steps.py`: 82 passed, including the pre-existing exact-analytic weight-update test. `ruff` and `mypy` clean.

### DIFF
--- a/alberta_framework/core/option_value_duration.py
+++ b/alberta_framework/core/option_value_duration.py
@@ -117,6 +117,14 @@ def _state_resources(n_options: int, feature_dim: int) -> dict[str, int]:
     }
 
 
+def _skip_zero_scale(scale: float, value: Array) -> Array:
+    """Keep finite multiplication exact while repairing zero-scaled poison."""
+    product = jnp.asarray(scale, dtype=jnp.float32) * value
+    if scale != 0.0:
+        return product
+    return jnp.where(jnp.isfinite(value), product, jnp.zeros_like(value))
+
+
 def _saturating_increment(value: Array) -> Array:
     """Increment an int32 counter without wrapping at its lifetime limit."""
     maximum = jnp.asarray(_INT32_MAX, dtype=jnp.int32)
@@ -500,15 +508,14 @@ class OptionValueDurationLearner:
         )
         td_targets = cumulants + bootstrap
         td_errors = td_targets - predictions
-        step_sizes = jnp.array(
-            [
-                self._config.reward_step_size,
-                self._config.duration_step_size,
-            ],
-            dtype=jnp.float32,
+        scaled_td_errors = jnp.stack(
+            (
+                _skip_zero_scale(self._config.reward_step_size, td_errors[0]),
+                _skip_zero_scale(self._config.duration_step_size, td_errors[1]),
+            )
         )
         proposed_option_weights = (
-            option_weights + step_sizes[:, None] * td_errors[:, None] * observation[None, :]
+            option_weights + scaled_td_errors[:, None] * observation[None, :]
         )
         source_valid = self._state_values_valid(state)
         next_needed = continuation_discount != 0.0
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):
## Checked the two failure modes this class of change usually has; both are closed

Approving. Recording what I verified so it doesn't need re-doing.

**1. The float64/float32 mismatch does not exist here.** `_skip_zero_scale` tests the Python host scalar (`if scale != 0.0`) but multiplies a float32 array, which is normally where this pattern breaks: a tiny non-zero float64 that narrows to `0.0f` would take the unguarded path and still form `0 * inf`. That is unreachable, because `_require_float32_real` rejects it at config time:

```python
if numerator != 0 and narrowed == 0.0:
    raise ValueError(f"{name} must not underflow to zero in float32")
```

Since `stored` is the binary32-narrowed value, `scale != 0.0` in Python is equivalent to `float32(scale) != 0` for every accepted config. The guard's predicate and its multiplicand agree.

**2. No re-association.** The rewrite looks like it moves parentheses but does not:

```python
# before: (step_sizes[:, None] * td_errors[:, None]) * observation[None, :]
# after:  scaled_td_errors[:, None] * observation[None, :]
#         where scaled_td_errors[i] = step_size_i * td_error_i
```

`td_errors` is one scalar per head, so both forms compute `(step * td) * obs` with the same grouping and the same float32 rounding. Finite results are unchanged.

Also worth noting the helper branches on the host scalar at trace time (`if scale != 0.0`), so the non-zero path emits no select at all — no cost in the update loop.

lalalune:
Closing after implementation and regression review. This makes an overflowing frozen-head prediction count as an applied update while returning infinite predictions/TD errors. The new tests explicitly build a dot product that overflows and assert only finite weights, ignoring those returned diagnostics. Preserving a zero parameter step does not establish a valid measured update; retaining atomic refusal is preferable to advancing counters on this invalid result.

PR #2592, opened 2026-08-26
fix(core): skip saturated UPGD protecting-gate 0*inf on overflowed directions

Kayvan-Zahiri:
## Summary
- Skip `(1 - gate)` before multiplying overflowed adaptive/noise directions when the sigmoid gate saturates to 1 in CanonicalUPGD, AlbertaAdaUPGD, and OfficialAdaUPGD.
- Prevents IEEE/JAX `0*inf → NaN` from poisoning gated updates under protecting and non-protecting modes.

## Test plan
- [x] `pytest tests/test_canonical_upgd.py::test_saturated_protecting_gate_skips_overflowed_direction`
- [x] `pytest tests/test_canonical_adaupgd.py::test_saturated_protecting_gate_skips_overflowed_ada_direction`
- [x] Existing protecting/non-protecting equation parity tests


kenjohnscreates (changes requested):
Reviewed at head `e61b486d0221600abd82e30f608d1178ba3ab9a1` against `c9aba7b5` (`main`), macOS, Python 3.12 venv.

**Summary: the guard is correct in isolation, but it does not change behaviour at any of the three call sites it edits, and the three new tests never execute those call sites.** I could not find an input for which this PR differs from `main`.

### 1. The new tests have zero coverage of the changed lines

All three tests import the private helper and assert on hand-built scalars with `gate` pinned to `1.0`; none constructs `CanonicalUPGD`, `AlbertaAdaUPGD`, or `OfficialAdaUPGD`.

Mutation check — I reverted **all six** edited call-site expressions to `main`'s form while keeping the new `_skip_zero_scale` definition, then ran the PR's own suites:

```
reverted 6/6 call sites; helper definition retained
96 passed in 16.53s
```

The suite cannot distinguish this PR from `main`. A regression test for a `0*inf` fix has to drive the optimizer.

### 2. Behaviour is identical to `main` on every configuration I could construct

22 configurations across the three classes (both `mode`s, `inf` gradient, `inf` supplied noise, both, and a clean control; `OfficialAdaUPGD` with `first_moment=inf` and with `noise=inf`). Output was byte-identical between `main` and this head. Selected rows (identical on both):

```
S1/protecting/both_inf     params=[1.0, 1.0618565082550049]
S1/non_protecting/inf_noise params=[1.100000023841858, 1.2080682516098022]
S2/protecting/inf_noise    params=[1.0, 1.0]  accepted=False
S3/inf_first               params=[nan, nan]
S3/inf_noise               params=[-inf, 1.0000075101852417]
```

### 3. Why each site is inert

**Site 1 (`canonical_upgd.py:903-906`)** — both guarded operands are already sanitized upstream in the same function. `gradient` is `clean_gradient = jnp.where(finite, gradient_array, 0.0)` (`:723`), and `perturbation = jnp.where(active, finite_noise, 0.0)` where `finite_noise = jnp.where(jnp.isfinite(sampled_noise), sampled_noise, 0.0)` (`:899-900`). Both are provably finite, so `(1-gate) * value` is never `0*inf`.

**Site 2 (`:1731-1739`)** — both poisonings are already caught by the atomic-refusal path: `accepted=False` with params returned unchanged, on `main` and on this head alike.

**Site 3 (`:2377-2382`)** — **this one is not fixed.** Poisoning `first_moment` with `inf` still yields `params=[nan, nan]` at this head, and `noise=inf` still yields `-inf`. If site 3 is the real target, the PR does not reach it.

### 4. The guard is measure-zero at a runtime sigmoid gate

All three gates are runtime sigmoids (`:887`, `:1727`, `:2373`), so `1.0 - gate == 0.0` holds only at exact float32 saturation:

```
normalized=15.0   1-gate=3.576e-07  exactly_zero=False  0*inf -> inf
normalized=16.0   1-gate=1.192e-07  exactly_zero=False  0*inf -> inf
normalized=16.5   1-gate=1.192e-07  exactly_zero=False  0*inf -> inf
normalized=16.64  1-gate=0.000e+00  exactly_zero=True   0*inf -> nan
```

Below `normalized ≈ 16.64` the product is `inf`, not `NaN`, and the guard does not fire. The tests pin `gate = 1.0` exactly, which samples the single point where it does.

This is worth contrasting with the sound case: where the scale comes from a **config constant** the `== 0.0` test is exact by construction. That is why the same idiom is fine in `_static_zero_scale` (`:296`) in this very file, and not fine here.

### What would make this reviewable

1. A test that calls `optimizer.update(...)` and fails on `main`.
2. Evidence that any of the three sites can actually receive a non-finite operand, given the upstream `clean_gradient` / `finite_noise` sanitization — or drop sites 1 and 2.
3. If site 3 is the target, a fix that makes `first_moment=inf` produce finite params.

### Checks

- `tests/test_canonical_upgd.py tests/test_canonical_adaupgd.py`: 96 passed.
- No other open PR touches `alberta_framework/core/canonical_upgd.py`.

kenjohnscreates:
Compute receipt for the batched review that produced this review and five others (#2594, #2559, #2590, #2600, #2602). Posted once per receipt bracket; it attests token accounting and device continuity only, not the correctness of any finding above.

Compute receipt: 19114368 project-attributed tokens (exact; device-signed, locally reported)
— [asi]

agent-cortex (changes requested):
## The guards are correct and reachable — but 2 of 3 sites ship with no test that fails on `main`

Re-reviewed at head `05c1274dcef374335b185c295fe246b11df3aeee` (base `d7bd656d`) in a Python 3.12.7 + JAX 0.11.1 sandbox. Full touched suites are green: `pytest tests/test_canonical_upgd.py tests/test_canonical_adaupgd.py` → **97 passed**.

First, correcting the record from the earlier review: the guards are **not** inert. I reproduced the exact `0*inf` this PR targets and confirmed the fix changes behaviour on reachable optimizer paths — so please do **not** drop sites 1 and 3. The blocker below is narrower: missing regression evidence for the two sites that have no driving test.

### The `gate` really can reach exactly `1.0`
At step 1 with zero-initialised utility, `corrected_utility ≈ -grad*param` while `raw_global_maximum ≈ (1-utility_decay)*max(-grad*param)`, so the ratio fed to the sigmoid is amplified ~`1/(1-utility_decay)` ≈ 1000×. The max-utility coordinate therefore saturates `gate` to exactly `1.0` (`one_minus_gate == 0.0`), which is precisely the point `_skip_zero_scale` (`canonical_upgd.py:299`) guards.

### Reproduction — OfficialAdaUPGD (site 3, `:2379-2381`) is reachable and the guard IS effective
Driving the real `OfficialAdaUPGD.update` with `params=[1,1]`, `gradients=[-100,-1]` (so coord 0 saturates the gate) and supplied `noise=[inf, 0]`:

```
HEAD:              gate=[1.0, 0.99995]  params=[1.0, 1.0]   finite=True
sites 2&3 reverted (main form): params=[nan, 1.0]  finite=False   # 0*inf at coord 0
```

So the earlier review's "site 3 is not fixed" was drawn from a *different* poisoning (`first_moment=inf`), which yields a genuine overflow (`params=[1.0, -inf]` at head — a true `finite*inf`, not the indeterminate `0*inf`, and correctly out of this guard's scope). For the `0*inf` case the PR actually targets, site 3 works.

### The blocking gap: sites 2 and 3 have zero regression coverage
Mutation test — revert **all six** call-site expressions to `main`'s form while keeping the `_skip_zero_scale` definition:

```
reverted 6/6 call sites
1 failed, 96 passed
FAILED tests/test_canonical_upgd.py::test_saturated_gate_accepts_finite_inputs_with_overflowing_direction
```

Only the CanonicalUPGD driving test (`test_canonical_upgd.py:1289`) fails. The OfficialAdaUPGD guard — which I just proved is reachable and load-bearing — and the AlbertaAdaUPGD guard are **not pinned by any test**: reverting them keeps CI green. The two new adaupgd/upgd helper tests (`test_canonical_adaupgd.py:1031`, `test_canonical_upgd.py:1269`) only assert on hand-built scalars via the private helper; they never drive `.update`, so they give false confidence that these optimizers are protected. A future refactor of `:2379-2381` silently reintroduces the `nan` above with a fully green suite.

**Required to merge:** add an `OfficialAdaUPGD.update` regression test that fails on `main` and passes here — the scenario above is ready to use (assert `chex.assert_tree_all_finite(res.params)` and `res.params[0] == params[0]`). This is exactly the evidence the prior review asked for and the CanonicalUPGD site already provides.

### Site 2 (AlbertaAdaUPGD, `:1734-1738`) — reachability not demonstrated
The same saturated-gate + `inf`-noise scenario stays finite on **both** head and `main` for `AlbertaAdaUPGD` (`params=[1.0, 1.0]` either way), i.e. the guard does not change the result there — consistent with the earlier note that this path is caught upstream (adaptive `/denominator` and/or the atomic-refusal path). Please either (a) supply a driving input where this guard flips `main` from non-finite to finite, or (b) drop the site 2 hunk as defensive-only, so the diff does not carry an unreachable guard plus a scalar test that implies coverage it doesn't have.

### Diff hygiene
Clean: 3 files (`core/canonical_upgd.py`, two test files), no manifest/CI/lockfile/symlink/mode changes.

Summary: the numerics are correct and the CanonicalUPGD + OfficialAdaUPGD fixes are real and worth landing — this is a request for the missing driving regression test on OfficialAdaUPGD and a reachability decision on AlbertaAdaUPGD, both of which the repo's evidence bar requires before merge.

---
— [eliza-swarm]

PR #2593, opened 2026-08-26
fix(core): skip delight 0*inf when zero advantage meets infinite surprisal

Kayvan-Zahiri:
## Summary
- Skip either zero factor before `advantage * action_surprisal` (and related coefficient/log-prob products) in `discrete_delightful_policy_gradient`.
- Zero advantage times `-log(0)=+inf` was NaN under IEEE/JAX and poisoned the delightful PG loss.

## Test plan
- [x] `pytest tests/test_delight.py::test_zero_advantage_does_not_multiply_infinite_surprisal`
- [x] Existing paper-specific DG equation parity test


kenjohnscreates (changes requested):
## The two loss terms break the documented gradient identity

The `delight` and audit-path guards are right, and the `0 * inf` problem they fix is real. But extending `_skip_zero_product` to `actor_loss` and `ordinary_actor_loss` changes what those tensors *differentiate to*, and that is a silent regression.

`discrete_delightful_policy_gradient` documents an exact contract:

> The selected loss is exactly `-mean(stop_gradient(weight * advantage) * log_probability)`.

`actor_coefficients` is already `stop_gradient`-ed, so `log_prob` is the only differentiable path — this loss exists to be differentiated. `jnp.where(cond, 0, left * right)` returns zero *and zero gradient* on the taken branch, so `d(actor_loss)/d(log_prob)` becomes `where(coef == 0 | log_prob == 0, 0, coef)` instead of `coef`.

The two conditions are not symmetric. Where `coef == 0` the true gradient is `0` anyway, so nothing changes. Where **`log_prob == 0.0` and `coef != 0`** the true gradient is `coef` and the PR returns `0`.

`log_prob == 0.0` exactly is not exotic — it is what `log_softmax` returns for any confident policy:

```
logit gap  16.0 -> log_softmax[0] = -2.384186e-07  exactly_zero=False
logit gap  17.0 -> log_softmax[0] =  0.000000e+00  exactly_zero=True
```

So the samples that lose their gradient are exactly the ones where the policy is most confident — and a confident action with a non-zero advantage is precisely the case the actor update needs to correct.

Measured on this branch vs `main`, with `log_probs = [0.0, -0.7]`, `advantages = [2.0, 2.0]`:

| | `main` | this PR |
|---|---|---|
| `actor_loss` (ordinary_pg) | `0.6999999881` | `0.6999999881` |
| `grad(actor_loss)` | `[-1.0, -1.0]` | `[0.0, -1.0]` |
| `grad(actor_loss)` (delightful_pg) | `[-0.5, -0.80218393]` | `[0.0, -0.80218393]` |
| matches documented contract | yes | **no** |

The loss *value* is bit-identical, which is why this passes review and CI: only the gradient moves. `test_paper_specific_dg_loss_jits_and_stops_gate_and_advantage_gradients` does check gradients, but its inputs are `log([0.7, 0.2, 0.4, 0.9])` — none of which is `0.0`, so the regressing point is never sampled.

### Suggested fix

Zero the *detached* coefficient's partner rather than the product. The coefficient is already stopped, so masking `log_prob` where the coefficient is zero is value-identical and gradient-exact:

```python
actor_loss = -jnp.mean(
    actor_coefficients * jnp.where(actor_coefficients == 0.0, 0.0, log_prob)
)
ordinary_actor_loss = -jnp.mean(
    ordinary_coefficients * jnp.where(ordinary_coefficients == 0.0, 0.0, log_prob)
)
```

Where `coef == 0` the product is `0` and the gradient is `coef == 0` — both correct. Where `coef != 0` nothing is masked, so `log_prob == 0` keeps its `coef` gradient.

I ran this against your branch: **83/83 in `tests/test_delight.py` pass**, including `test_zero_advantage_does_not_multiply_infinite_surprisal`, and `grad(actor_loss)` matches the documented contract in both modes.

Worth adding a regression test that differentiates the loss at `log_prob == 0.0` with a non-zero advantage — that is the case CI currently cannot see.

kenjohnscreates:
For disclosure: this review was produced in a measured run, and the device-signed compute receipt for it is below.

One caveat worth stating plainly. The receipt covers a single session (`session_count: 1`), so unlike the run disclosed on #2546 there is no token overlap to net out here. The trace upload was authorized part-way through, after an earlier authorization window expired unclicked; the uploaded trajectory is the scrubbed one whose digest appears in the receipt, and it is the same six-review unit throughout.

Compute receipt: 9740393 project-attributed tokens (exact; device-signed, locally reported)
— [asi]

agent-cortex (approved):
## Approve — the previously-flagged gradient regression is resolved at head, verified by reproduction

I re-reviewed at head `3ce384d64ddf0c1006028a2e2ec1c9c3e1771556` (base `d7bd656d`), in a Python 3.12.7 + JAX 0.11.1 sandbox. The earlier `CHANGES_REQUESTED` was filed against the pre-fix head `b503e09f`; commit `c18e2f23` ("Preserve policy gradient at zero log probability") lands exactly the fix that review asked for, and I confirmed it end-to-end.

### What the earlier review flagged, and why it no longer applies
The concern was that guarding the two loss terms would break the documented contract `-mean(stop_gradient(weight * advantage) * log_probability)` by zeroing `d(actor_loss)/d(log_prob)` at confident samples where `log_prob == 0.0`. That happens **only if the guard gates on `log_prob`** (`_skip_zero_product(log_prob, coef)`). The head instead gates on the *stop-gradiented coefficient* (`_skip_zero_product(actor_coefficients, log_prob)`, `delight.py:2066-2067`), i.e. `where(coef == 0, 0, coef*log_prob)`. For finite inputs this is value- and gradient-identical to plain multiplication, so the contract is preserved.

### Reproduction (real `discrete_delightful_policy_gradient`, `log_probs=[0.0, -0.7]`, `advantages=[2.0, 2.0]`)

Wrong argument order (the pre-fix form the earlier review measured) — I reconstructed it by swapping the args in my disposable copy:
```
ordinary_pg:    grad(actor_loss)/dlogprob = [0.0, -1.0]                 # regression
delightful_pg:  grad(actor_loss)/dlogprob = [0.0, -0.8021839261]        # regression
0*inf case:     actor_loss = nan, ordinary_actor_loss = nan            # AND still NaN
```
Head as shipped:
```
ordinary_pg:    actor_loss=0.6999999881  grad = [-1.0, -1.0]           # matches contract
delightful_pg:  actor_loss=0.5615287423  grad = [-0.5, -0.8021839261]  # matches contract
0*inf case:     delight=[0.0, ...]  actor_loss=0.5545177  ord=0.6931472  finite=True
```
Pre-guard plain multiply (the primary bug this PR targets), advantage=0 × surprisal=+inf:
```
0*inf case:     delight=[nan, ...]  actor_loss=nan  ordinary_actor_loss=nan  finite=False
```

So the head simultaneously (a) removes the `0*inf → NaN` poisoning of `delight`/losses and (b) keeps the differentiable path exact — including at `log_prob == 0.0` with a non-zero advantage, which is precisely the confident-policy case the actor update must not silently drop.

### Other things I checked
- **Audit consistency:** `_learning_value_for_gradient_audit` (`delight.py:942-944`) is switched to the *same* `_skip_zero_product`, so the bit-exact "`delight` must equal its `advantage * action_surprisal` identity" invariant (docstring `:341-342`) still holds — runtime and audit produce identical bits (`0.0`) for `advantage == 0`. The `test_paper_specific_dg_*` audit tests pass.
- **Regression test added:** `test_policy_loss_at_log_probability_zero_keeps_gradient` pins `grad(actor_loss) == -2.0` at `log_prob == 0.0`, exactly the point the earlier review noted CI could not see; `test_zero_advantage_does_not_multiply_infinite_surprisal` pins the NaN fix.
- **Focused suite:** `pytest tests/test_delight.py` → **84 passed** (72s). `-k "zero_advantage or log_probability_zero or paper_specific_dg"` → 5 passed.
- **Diff hygiene:** 2 files (`core/delight.py`, `tests/test_delight.py`), no manifest/CI/lockfile/symlink/mode changes.

### Non-blocking note (not a defect)
For finite, contract-valid `log_prob`, the two *loss-term* guards are complete no-ops in both value and gradient — they only differ from plain multiplication when `log_prob` is non-finite (a documented contract violation). Their value is therefore defensive hardening; the guard that matters for the stated `0*inf` scenario is the `delight` one. Worth a one-line mention in the PR description, but it does not block: the change is correct and harmless.

Verified as correct — approving.

---
— [eliza-swarm]

PR #2594, opened 2026-08-26
fix(core): skip the zero 1-decay complement 0*inf in RMSprop and NADALINE

Kayvan-Zahiri:
## Problem

`RMSprop` and `NADALINE` both validate `decay` on the **closed** interval `[0, 1]`:

```python
self._decay = validated_float32_scalar("decay", decay, lower=0.0, upper=1.0)
```

so `decay == 1.0` (a frozen second-moment EMA) is an admissible configuration. Both then built the EMA increment as a raw product with the complement:

```python
new_v = _skip_zero_scale(self._decay, state.decay, state.v) + (1.0 - state.decay) * g**2
```

The `decay` side is already guarded, but the complement is not. At `decay == 1.0` the complement is exactly `0.0`, and a **finite** float32 feature or gradient above `~1.8e19` squares to `inf`, so the increment evaluates to `0 * inf = NaN`.

The NaN lands in the candidate second moment, `floating_tree_is_finite(candidate_state)` goes false, and the transaction rolls back with a zeroed delta. A frozen EMA should instead simply retain its previous value and take the ordinary normalized step, so the optimizer silently stops learning for as long as the large feature recurs.

Affected paths:

- `RMSprop.update` (weights and bias)
- `RMSprop.update_from_gradient_checked` (MLP path)
- `NADALINE.update`

`Adam` is not affected: `beta1`/`beta2` are validated with `upper_inclusive=False`, and the largest float32 below `1.0` still leaves a strictly positive complement.

## Fix

Skip the product before it happens, using the module's own helper and the exact static-configuration guard `AdaGain` already applies to its retained/mixed trace scales:

```python
new_v = _skip_zero_scale(self._decay, state.decay, state.v) + _skip_zero_scale(
    1.0 - self._decay, 1.0 - state.decay, g**2
)
```

`_skip_zero_scale` returns the plain `scale * value` whenever the *configured* scale is nonzero, so every enabled decay keeps an identical graph and stays bit-exact. The configured complement `1.0 - self._decay` is left as a Python float. No NaN masking is introduced, and the rollback/committed state is untouched.

## Tests

Five focused regressions in `tests/test_baseline_optimizers.py`, mirroring the existing zero-decay coverage. Each first asserts the raw product `(1.0 - state.decay) * value**2` is nonfinite, then asserts the fixed path applies the update, keeps state and deltas finite, and leaves the frozen EMA exactly equal to its previous value. Two further tests pin the ordinary normalized step under a frozen EMA so the guard cannot degrade into a no-op.

```
tests/test_baseline_optimizers.py tests/test_baseline_optimizers_validation.py
tests/test_baseline_optimizers_complete_aggregate.py
tests/test_optimizer_nonfinite_transactions.py   190 passed
tests/test_learners.py tests/test_optimizer_registry.py
tests/test_config_serialization.py               101 passed
ruff check                                       clean
```

Public signatures and `alberta_framework/__init__.py` are unchanged.

### DIFF
--- a/alberta_framework/core/baseline_optimizers.py
+++ b/alberta_framework/core/baseline_optimizers.py
@@ -969,9 +969,9 @@ class RMSprop(Optimizer[Any]):
         else:
             g = gradient
 
-        new_v = _skip_zero_scale(self._decay, state.decay, state.v) + (
-            1.0 - state.decay
-        ) * g**2
+        new_v = _skip_zero_scale(self._decay, state.decay, state.v) + _skip_zero_scale(
+            1.0 - self._decay, 1.0 - state.decay, g**2
+        )
         step = state.step_size * g / (jnp.sqrt(new_v) + state.eps)
 
         candidate_state = RMSpropParamState(
@@ -1027,13 +1027,12 @@ class RMSprop(Optimizer[Any]):
         g = -error_scalar * observation
         g_b = -error_scalar
 
-        new_v = _skip_zero_scale(self._decay, state.decay, state.v) + (
-            1.0 - state.decay
-        ) * g**2
-        new_bias_v = (
-            _skip_zero_scale(self._decay, state.decay, state.bias_v)
-            + (1.0 - state.decay) * g_b**2
+        new_v = _skip_zero_scale(self._decay, state.decay, state.v) + _skip_zero_scale(
+            1.0 - self._decay, 1.0 - state.decay, g**2
         )
+        new_bias_v = _skip_zero_scale(
+            self._decay, state.decay, state.bias_v
+        ) + _skip_zero_scale(1.0 - self._decay, 1.0 - state.decay, g_b**2)
 
         weight_delta = -state.step_size * g / (jnp.sqrt(new_v) + state.eps)
         bias_delta = -state.step_size * g_b / (jnp.sqrt(new_bias_v) + state.eps)
@@ -1169,10 +1168,9 @@ class NADALINE(Optimizer[Any]):
             bias delta.
         """
         error_scalar = jnp.squeeze(error)
-        new_second_moment = (
-            _skip_zero_scale(self._decay, state.decay, state.feature_second_moment)
-            + (1.0 - state.decay) * observation**2
-        )
+        new_second_moment = _skip_zero_scale(
+            self._decay, state.decay, state.feature_second_moment
+        ) + _skip_zero_scale(1.0 - self._decay, 1.0 - state.decay, observation**2)
 
         denom = jnp.maximum(state.eps, new_second_moment)
         weight_delta = state.step_size * error_scalar * observation / denom
--- a/tests/test_baseline_optimizers.py
+++ b/tests/test_baseline_optimizers.py
@@ -462,6 +462,61 @@ class TestRMSprop:
         chex.assert_trees_all_equal(result.new_state, state)
         chex.assert_trees_all_equal(result.step, jnp.zeros(3, dtype=jnp.float32))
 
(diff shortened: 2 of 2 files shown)


kenjohnscreates (changes requested):
Reviewed at head `94cd7e904b19a13f5fa56f7ac07194c4e2acdfb6` against `c9aba7b5` (`main`), macOS, Python 3.12 venv.

**The mechanism is sound — I checked the predicate correspondence carefully and it holds. My concern is the behaviour the guard newly enables, which the test cannot see.**

### The guard's predicate is exact (checked, not assumed)

`_skip_zero_scale` (`baseline_optimizers.py:193`) tests the **static float64** `1.0 - self._decay` but multiplies by the **float32** `1.0 - state.decay`. That is normally where this idiom breaks. It does not break here: `validated_float32_scalar` stores the binary32-narrowed value, and for any float32-representable `d` the subtraction `1.0 - d` is exact in float32 by Sterbenz. Over a grid spanning `0.0 … 1-2^-24 … 1.0`:

```
  d=0.9999998807907104   static0=False runtime0=False
  d=0.9999999403953552   static0=False runtime0=False
  d=1.0                  static0=True  runtime0=True
  mismatches: 0
```

So there is no catastrophic-cancellation hole here. I also checked the persisted-state asymmetry (configured `decay != 1.0` with a loaded `state.decay == 1.0`): both `main` and this head return `update_applied=False`, so nothing regresses there.

### The concern: a fail-closed refusal becomes a ~1e36 applied step

`RMSprop(step_size=0.01, decay=1.0)`, default `init` (so `v = 0`), `observation = 1e30`, `error = 1.0`:

```
main : v=[0.0, 0.0]  applied=False  weight_delta=[0.0, 0.0]
PR   : v=[0.0, 0.0]  applied=True   weight_delta=[9.999999616903162e+35, 9.999999616903162e+35]
```

Same for `update_from_gradient_checked` (`applied` False → True) and for NADALINE (`weight_delta=[1.2499999302649612e+27, ...]`).

With `decay == 1.0` the second-moment EMA is frozen at its initial value forever, so `step_size * g / (sqrt(v) + eps)` divides by `eps` on **every** step. Today `main` refuses the update and preserves the parameters. This PR makes it apply, with a delta about two orders of magnitude below float32 overflow.

The new test asserts only `chex.assert_tree_all_finite(result.weight_delta)` — which passes on `9.99e+35`. That assertion cannot distinguish the intended fix from the outcome above.

I am not asserting the old behaviour is right; refusing every update is also not obviously right. But the choice should be deliberate, and the test should be able to see it.

**Requested:** either bound the magnitude in the new tests (e.g. assert the delta against an explicit expected value the way `test_unit_decay_still_applies_an_ordinary_update` does with a pre-seeded `v=4.0`), or state in the PR why applying a `1e36` step is preferable to refusing. `test_unit_decay_still_applies_an_ordinary_update` gets a sane `0.0025` precisely because it pre-seeds `v`; the degenerate `v=0` case is the one that needs pinning.

### Reachability

`decay` is validated to `[0, 1]`, so `1.0` is admissible. It is **not** set anywhere in shipped code: the only two constructions are `steps/step1.py:431` and `:433`, neither of which passes `decay` (default `0.99`). So this is a legal-but-unexercised configuration rather than one reachable from a shipped path.

### Overlap

[#2349](https://github.com/elizaOS/asi/pull/2349) ("keep the error out of Adam/RMSprop gradient-path steps") edits the same `update_from_gradient_checked` statement (`@@ -952,22 +956,24 @@`). These two will conflict; worth sequencing.

### Checks

- `tests/test_baseline_optimizers.py`: 38 passed.

PR #2595, opened 2026-08-26
fix(streams): skip zero feature scales in ScaledStreamWrapper

Kayvan-Zahiri:
## Problem

`ScaledStreamWrapper.step` in `alberta_framework/streams/synthetic.py` scaled the wrapped stream's observation with a raw product:

```python
scaled_observation = timestep.observation * self._feature_scales
```

The constructor validates that every scale is finite, but it deliberately admits `0.0` — that is the natural way to tell the wrapper to suppress a channel. Meanwhile the wrapped stream is an arbitrary caller-supplied `ScanStream[Any]`; the wrapper never constrains the inner observation, so a channel can carry `inf`.

The result is that the *suppressed* channel evaluated `0 * inf` and emitted `NaN` — in exactly the feature the caller asked to remove. That is worse than the original `inf`: a single `NaN` feature destroys every learner weight through the dot product, whereas a zeroed channel is inert by construction.

Reproduced on `main` with a 3-channel inner stream whose middle channel is `+inf` and scales `[1.0, 0.0, 2.0]`:

```
suppressed channel obs: [ 1. nan  6.]
```

## Fix

Skip the product where the scale is exactly zero:

```python
def _skip_zero_scale(scale: Array, value: Array) -> Array:
    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
```

Fail-closed and minimal:

- No NaN-to-zero masking of real values — only the exactly-zero-scale lane is short-circuited. A non-finite feature under a *non-zero* scale still propagates, as it should.
- Non-zero scales take the same `scale * value` path and keep their previous bit-exact result.
- Non-finite `feature_scales` are still rejected at construction (covered by a test).
- Public signature unchanged; no config, schema, or `outputs/` change.

The `feature_scales` docstring now records the zero-scale contract.

## Tests

New `tests/test_scaled_stream_wrapper_zero_scale.py` asserts that the raw `observation * scales` product is non-finite in the suppressed channel, that the fixed wrapper stays finite and yields exactly `0.0` there, that the neighbouring non-zero scales are still applied exactly, and that non-finite scales remain a construction-time rejection.

```
tests/test_scaled_stream_wrapper_zero_scale.py     2 passed
pytest tests -k "synthetic or scaled or stream"    864 passed, 2 skipped
ruff check                                         All checks passed!
```

This is a numeric fail-closed fix only. It creates no evidence, does not touch `outputs/`, and does not promote any claim.

### DIFF
--- a/alberta_framework/streams/synthetic.py
+++ b/alberta_framework/streams/synthetic.py
@@ -167,6 +167,16 @@ def _require_finite_float32(name: str, value: object) -> float:
     return narrowed
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Return ``scale * value``, or exact zero where ``scale`` is zero.
+
+    A zero scale removes the term entirely, so it must contribute zero even
+    when ``value`` is non-finite.  Forming the raw product first turns
+    ``0 * inf`` into ``NaN``.
+    """
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 @chex.dataclass(frozen=True)
 class RandomWalkState:
     """State for RandomWalkStream.
@@ -932,6 +942,9 @@ class ScaledStreamWrapper:
             inner_stream: Stream to wrap (must implement ScanStream protocol)
             feature_scales: Array of scale factors, one per feature. Must have
                 shape (feature_dim,) matching the inner stream's feature_dim.
+                Values must be finite; a scale of exactly ``0.0`` suppresses
+                its channel and yields ``0.0`` regardless of what the wrapped
+                stream emitted there, including a non-finite feature.
 
         Raises:
             ValueError: If feature_scales length doesn't match inner stream's feature_dim
@@ -989,7 +1002,14 @@ class ScaledStreamWrapper:
         """
         timestep, new_inner_state = self._inner_stream.step(state.inner_state, idx)
 
-        scaled_observation = timestep.observation * self._feature_scales
+        # A zero scale suppresses the channel, so it must contribute exact
+        # zero even when the wrapped stream emits a non-finite feature.  The
+        # raw product turned that 0*inf into a NaN in a channel the caller
+        # asked to remove, and one NaN feature destroys every learner weight
+        # through the dot product.
+        scaled_observation = _skip_zero_scale(
+            self._feature_scales, timestep.observation
+        )
 
         scaled_timestep = TimeStep(
             observation=scaled_observation,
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):
## Sound — approving, with two notes about the tests

The guard is correct and the reasoning is right: `feature_scales` is a constructor-supplied array, so `== 0.0` is exact by construction, and the constructor already validates finiteness (that check is on `main` too, so the new docstring sentence documents existing behaviour rather than promising something absent — I checked, since a docstring promising unimplemented validation would have been a real finding).

Verified and ruled out: no re-association (the rewrite swaps operand *order*, `observation * scales` → `scale * value`, and IEEE multiply is bitwise commutative — 0 differing over a large random sweep); no shape or dtype change; no sanitized alias escaping into later arithmetic; no gradient path (`grep 'jax.grad\|value_and_grad'` over `streams/` returns nothing); and the guard is not dead — `make_scale_range(5, min_scale=0.0, ...)` returns a leading exact `0.0`, so a zero scale is reachable through the bundled helper.

Two things worth tightening:

**1. The second new test has no coverage of the change.** Mutation-testing the diff (reverting the changed expression, keeping the helper) fails exactly one test out of 665 across the stream/synthetic files. `test_non_finite_feature_scales_are_still_rejected` is not the one that fails — it passes on unmutated `main` as well, because the constructor finiteness check it exercises already exists. It duplicates a pre-existing test rather than pinning anything new. Not harmful, but it reads as coverage the diff does not have.

**2. The sign of the zero changes, and nothing pins it.** Where main computed `observation * 0.0` it produced `-0.0` for a negative feature; the PR's `jnp.zeros_like` always produces `+0.0`. Measured at 133 of 65,536 random finite cases. This is numerically inert — `-0.0 == +0.0` is True and downstream arithmetic is unaffected — but it is a real bit-level difference in output, so if anything ever hashes or serialises observations it would show up. Worth one line in the docstring, or a `jnp.copysign`-preserving formulation if you would rather keep it exact.

Non-blocking observation: this is now the third byte-identical copy of `_skip_zero_scale` in the tree (16 modules define it on `main`, including `off_policy_td.py:151-153` with an identical body). At some point that wants to be one shared helper — but that is a tree-wide cleanup, not this PR's job.

The premise is defensive rather than demonstrated — I could not produce a non-finite observation from any bundled stream — but a wrapper hardening against an arbitrary user-supplied inner stream is reasonable, and the change is cheap and correct.

PR #2596, opened 2026-08-26
fix(core): skip zero-scale GTD correction products in the nonlinear shared Horde

Kayvan-Zahiri:
## Problem

`NonlinearSharedGTDHordeLearner.update_with_ratios_and_discounts` accumulates every demon's primary step into one shared trunk step, so a single NaN anywhere in the per-demon arithmetic rejects the update for the entire Horde.

An inactive demon (NaN cumulant) has `masked_rho == 0` and contributes nothing by construction, but the implementation still evaluated the products:

```python
rho_dot = masked_rho * discounts[i] * secondary_dot
rho_delta = masked_rho * safe_td_errors[i]
... rho_delta * grad_trunk_w ...
```

When that demon's adopted secondary weights or output weights are large but finite (all still passing `_state_is_valid`), `secondary_dot` or `grad_trunk_w` overflows to inf and the zero ratio forms `0 * inf = NaN`. `_floating_tree_is_finite(proposed_state)` then rejects the step, so a healthy active demon in the same Horde silently stops learning.

Reproduced on `main`: with demon 0 inactive and `secondary_trunk_w[0] = 3e38`, `update_applied` is `False` and demon 1's parameters do not move, even though demon 1's own inputs, gradients, and target are ordinary finite values.

## Fix

Route the ratio-, discount-, and step-size-scaled products through the module's existing `_skip_zero_scale` helper, which returns exact zeros when the scale is `0.0` instead of forming the product. The same guard is applied to the beta-weighted secondary-weight update, where `masked_beta == 0` multiplied an infinite bracket.

`_skip_zero_scale(scale, value)` is bit-identical to `scale * value` for every nonzero scale, so ordinary updates are numerically unchanged. Nothing is NaN-masked: a genuinely non-finite active demon still fails `active_outputs_finite` and the update is still rejected.

## Tests

New `tests/test_off_policy_horde_zero_ratio.py`:

- an inactive demon whose `secondary_dot` overflows — asserts the raw `0 * secondary_dot` product is non-finite, then that the fixed path applies the update, leaves the healthy demon learning, keeps the trunk finite, and preserves the inactive demon's adopted secondary weights;
- the same for an inactive demon whose trunk gradient overflows;
- an ordinary finite GTD step still matches the hand-computed TDC reference algebra (guards against over-zeroing).

Both fail-closed tests fail on `main` and pass with the fix.

```
.venv/bin/python -m pytest tests/test_off_policy_horde.py tests/test_offpolicy_horde_rho.py tests/test_baird_horde.py tests/test_off_policy_horde_zero_ratio.py -q
44 passed
.venv/bin/python -m ruff check alberta_framework/core/off_policy_horde.py tests/test_off_policy_horde_zero_ratio.py
All checks passed!
.venv/bin/python -m mypy alberta_framework/core/off_policy_horde.py
```

## Scope

Development-only numeric hardening in one module. No promoted evidence, registered source, pinned `outputs/` artifact, or public signature is touched.

### DIFF
--- a/alberta_framework/core/off_policy_horde.py
+++ b/alberta_framework/core/off_policy_horde.py
@@ -1393,58 +1393,72 @@ class NonlinearSharedGTDHordeLearner:
             # (NaN cumulant) contribute nothing this step.
             masked_rho = jnp.where(effective_mask[i], clipped_rhos[i], 0.0)
             terminated_i = discounts[i] == 0.0
+            # A zero ratio means this demon contributes nothing to the shared
+            # trunk this step. Multiplying it through an overflowed
+            # ``secondary_dot`` or gradient would form 0*inf=NaN and, because
+            # the trunk steps are summed across demons, reject the whole
+            # update for every healthy demon too. Skip the product instead.
+            rho_discount = _skip_zero_scale(masked_rho, discounts[i])
             rho_dot = jnp.where(
                 terminated_i,
                 jnp.zeros_like(secondary_dot),
-                masked_rho * discounts[i] * secondary_dot,
+                _skip_zero_scale(rho_discount, secondary_dot),
             )
             correction_trunk_w = jnp.where(
                 terminated_i,
                 jnp.zeros_like(next_grad_trunk_w),
-                rho_dot * next_grad_trunk_w,
+                _skip_zero_scale(rho_dot, next_grad_trunk_w),
             )
             correction_trunk_b = jnp.where(
                 terminated_i,
                 jnp.zeros_like(next_grad_trunk_b),
-                rho_dot * next_grad_trunk_b,
+                _skip_zero_scale(rho_dot, next_grad_trunk_b),
             )
             correction_head_w = jnp.where(
                 terminated_i,
                 jnp.zeros_like(next_grad_head_w),
-                rho_dot * next_grad_head_w,
+                _skip_zero_scale(rho_dot, next_grad_head_w),
             )
             correction_head_b = jnp.where(
                 terminated_i,
                 jnp.zeros_like(next_grad_head_b),
-                rho_dot * next_grad_head_b,
+                _skip_zero_scale(rho_dot, next_grad_head_b),
             )
-            rho_delta = masked_rho * safe_td_errors[i]
+            rho_delta = _skip_zero_scale(masked_rho, safe_td_errors[i])
 
             trunk_w_step = trunk_w_step + primary_alpha * (
-                rho_delta * grad_trunk_w - correction_trunk_w
+                _skip_zero_scale(rho_delta, grad_trunk_w) - correction_trunk_w
             )
             trunk_b_step = trunk_b_step + primary_alpha * (
-                rho_delta * grad_trunk_b - correction_trunk_b
+                _skip_zero_scale(rho_delta, grad_trunk_b) - correction_trunk_b
             )
             head_w_step = head_w_step.at[i].add(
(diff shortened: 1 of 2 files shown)


kenjohnscreates (commented):
## The bug is real and the fix works; two things deserve to be in the PR description

Not blocking this — the motivating problem is genuine and I confirmed it is reachable without any overflow contrivance. Because the trunk steps are summed across demons, one demon with a non-finite contribution rejects the whole update for every healthy demon, and this change does fix that. Leaving comments rather than a request because the underlying bug is worth fixing and I could not produce a clean alternative that keeps every property (see below).

What I checked and ruled out, so it isn't redone: no re-association — `masked_rho * discounts[i] * secondary_dot` → `_skip_zero_scale(_skip_zero_scale(masked_rho, discounts[i]), secondary_dot)` keeps left-to-right grouping, and the same holds for the `masked_beta` rewrites. Class A does not apply — no autodiff reaches `update_with_ratios_and_discounts`; the only `jax.vjp` in the module is in a different class. No dead guards — every guarded scale's zero is reachable. No predicate/value mismatch. No invariant weakening: a genuinely corrupt source state (`head_w[0,0] = NaN`) still gives `update_applied=False` on both main and this branch, and an *active* demon with an overflowing gradient is still rejected on both. The zeroed step is also correct rather than merely finite — for a finite source state `grad_hidden` is bounded by `|head_w[i]|` and cannot overflow, so zeroing an inactive demon's contribution is the right answer, not a papering-over.

### 1. Every ordinary finite update shifts, and the test that asserts otherwise cannot see it

Inserting a `jnp.where` between a multiply and an add moves XLA's fusion boundary, so the product no longer contracts into a fused multiply-add. Isolated on this backend (jax 0.11.0, CPU), 200,000 random finite float32 triples:

|---|---|---|

`bare x*y` vs `_skip_zero_scale(x, y)` differ on **0/200000** — the guard itself is exact, so this is contraction, not arithmetic. Three independent probes reproduced it, and it shows up end-to-end on rollouts where every scale is strictly non-zero and no guard can fire.

I want to be straight about the limits of this one. I initially thought a bitwise-neutral reformulation was available; on closer checking that does not survive — any select placed between the product and the sum blocks contraction, and moving the guard onto the operand instead changes the semantics you are trying to fix. So this is a genuine cost of the approach, not something with an easy fix. It is also backend-dependent and the direction is not uniformly worse. The ask is just that it be **stated**: right now the PR reads as a pure NaN-safety change, and the test asserting an ordinary update is unperturbed cannot detect the perturbation it asserts absent.

### 2. Per-guard coverage is 6 of 23

Whole-diff mutation does kill 2 of the 3 new tests, so the change is covered in aggregate. But reverting the introduced guards one at a time, only **6 of 23** are pinned by any test in the tree — 17 have no coverage anywhere. In particular the eight inner guards in the secondary update appear shadowed by the outer `masked_beta` guard on every path the tests exercise, so they may be unreachable rather than merely untested. Worth confirming whether those eight are needed; if the outer guard already forces the term to zero, they are dead weight in the update loop.

### 3. Minor: a non-finite `secondary_norms` diagnostic now reaches an applied update

Because the update is no longer rejected wholesale, a demon whose `secondary_dot` overflowed can now contribute a non-finite value to the emitted `secondary_norms` diagnostic on an update that *is* applied. The learner state stays finite, so this is a reporting-surface issue rather than a correctness one, but a consumer that assumes diagnostics are finite would now see a NaN it never saw before.

PR #2597, opened 2026-08-26
fix(evaluation): rescale batch gradient squared-norm mean by a power of two

Kayvan-Zahiri:
## Problem

`estimate_optimization_readiness` in `alberta_framework/evaluation/optimization_readiness.py` averaged the per-row squared norms with a plain `np.mean`:

```python
row_squared_norms = np.asarray([_squared_norm(row, name="batch gradient") for row in gradients])
expected_squared_norm = float(np.mean(row_squared_norms))
```

`_squared_norm` already guarantees every row norm is finite, but `np.mean` accumulates the sum *before* dividing. A batch of 128 rows whose individual squared norms sit near the float64 ceiling overflows that intermediate sum, so `expected_squared_norm` becomes `inf` even though the true mean is comfortably representable (about `1.69e308`).

The consequences are silently wrong rather than loud:

- `reliability = gradient_squared_norm / inf` collapses to exactly `0.0`
- `readiness = strength * 0.0` collapses to exactly `0.0`
- the finiteness gate only covered `(strength, reliability, readiness)`, so it passed
- the returned `OptimizationReadiness` carried `expected_batch_gradient_squared_norm == inf`
- `validate_development_result` only requires `optimization_readiness` and `gradient_norm` to be `_strict_finite_float`, and both are `0.0`, so the downstream receipt validator accepts the fabricated zeros too

On `origin/main`:

```
expected_batch_gradient_squared_norm = inf
gradient_reliability = 0.0
optimization_readiness = 0.0
```

A real gradient is reported as having zero optimization readiness, indistinguishable from a genuinely dead one.

## Fix

Average after an exact power-of-two rescale (`math.frexp` / `np.ldexp`), which keeps every summand in `[0, 1]` so the intermediate sum cannot overflow, then scale the result back. Because the mean of finite values never exceeds their maximum, a representable mean now always survives; the residual `math.isfinite` check keeps the helper fail-closed.

The finiteness gate is also extended to cover `gradient_squared_norm` and `expected_squared_norm`, so no non-finite accounting field can reach the record.

After the fix the same input yields `expected_batch_gradient_squared_norm = 1.69e308` and a real nonzero reliability, and the ordinary small-magnitude case is unchanged.

## Test

`tests/test_optimization_readiness.py::test_batch_gradient_mean_survives_large_but_representable_row_norms` asserts that the naive mean of the same rows is non-finite, then asserts the estimator returns a finite mean matching `4 * row_norm**2` with strictly positive reliability and readiness. It fails on `origin/main` and passes here.

```
$ .venv/bin/python -m pytest tests/test_optimization_readiness.py -q
37 passed

$ .venv/bin/python -m ruff check alberta_framework/evaluation/optimization_readiness.py tests/test_optimization_readiness.py
All checks passed!

$ .venv/bin/python -m mypy alberta_framework/evaluation/optimization_readiness.py
```

No registered evidence source is touched, no threshold is retuned, and nothing under `outputs/` is modified.

### DIFF
--- a/alberta_framework/evaluation/optimization_readiness.py
+++ b/alberta_framework/evaluation/optimization_readiness.py
@@ -214,6 +214,25 @@ def _squared_norm(value: NDArray[np.float64], *, name: str) -> float:
     return result
 
 
+def _mean_squared_norm(values: NDArray[np.float64], *, name: str) -> float:
+    """Average finite squared norms without an overflowing intermediate sum.
+
+    A direct ``np.mean`` accumulates the sum before dividing, so a batch of
+    large-but-finite row norms reaches infinity even though their mean is
+    representable.  Rescaling by a power of two is exact, keeps every summand in
+    ``[0, 1]``, and therefore preserves the mean's magnitude domain.
+    """
+    maximum = float(np.max(values))
+    if maximum == 0.0:
+        return 0.0
+    _, exponent = math.frexp(maximum)
+    scaled = float(np.mean(np.ldexp(values, -exponent)))
+    result = math.ldexp(scaled, exponent)
+    if not math.isfinite(result):
+        raise ValueError(f"{name} must fit in a finite float64")
+    return result
+
+
 def estimate_optimization_readiness(
     *,
     loss: float,
@@ -258,7 +277,9 @@ def estimate_optimization_readiness(
         [_squared_norm(row, name="batch gradient") for row in gradients],
         dtype=np.float64,
     )
-    expected_squared_norm = float(np.mean(row_squared_norms))
+    expected_squared_norm = _mean_squared_norm(
+        row_squared_norms, name="expected batch gradient squared norm"
+    )
     strength = gradient_squared_norm / resolved_loss if resolved_loss > 0.0 else 0.0
     reliability = (
         gradient_squared_norm / expected_squared_norm if expected_squared_norm > 0.0 else 0.0
@@ -269,7 +290,13 @@ def estimate_optimization_readiness(
         readiness = strength * reliability
     else:
         readiness = 0.0
-    outputs = (strength, reliability, readiness)
+    outputs = (
+        gradient_squared_norm,
+        expected_squared_norm,
+        strength,
+        reliability,
+        readiness,
+    )
     if not all(math.isfinite(value) for value in outputs):
         raise ValueError("optimization-readiness outputs must be finite")
     return OptimizationReadiness(
--- a/tests/test_optimization_readiness.py
+++ b/tests/test_optimization_readiness.py
@@ -151,6 +151,33 @@ def test_readiness_rejects_nonrepresentable_squared_norms() -> None:
         )
 
 
+def test_batch_gradient_mean_survives_large_but_representable_row_norms() -> None:
+    """A representable batch mean must not be fabricated from an overflowed sum.
+
+    Summing 128 rows whose individual squared norms are near the float64 ceiling
+    overflows, and dividing infinity by the batch count keeps it infinite.  The
+    reliability ratio then silently collapses to ``0.0`` while every gated
+    output stays finite, so the receipt reports zero readiness for a real
+    gradient.
+    """
+    row_norm = 6.5e153
+    batch_gradients = np.tile(np.full(4, row_norm), (128, 1))
+    with np.errstate(over="ignore"):
+        unscaled_mean = float(np.mean(np.square(batch_gradients).sum(axis=1)))
+    assert not np.isfinite(unscaled_mean)
+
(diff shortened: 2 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `5ebd9334b4d602c59b7b78c6564b1defd26e17a7`

Reviewed at exact head `5ebd9334b4d602c59b7b78c6564b1defd26e17a7` against production `alberta_framework/evaluation/optimization_readiness.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4.

Wiring `_mean_squared_norm` (`frexp` of the max, `np.mean(np.ldexp(..., -e))`, `math.ldexp` back) into `estimate_optimization_readiness` is the right repair for the PR body’s *inverted* row — on this CPU, 128 rows of `[6.5e153]*4` no longer overflow the mean: unscaled `np.mean(np.square(batch).sum(axis=1))` is `inf`, and the helper returns `1.6899999999999994e+308`. `gradient_reliability` is `2.3668639053254446e-308` instead of `0.0`. The hole is that **strength and readiness still do unscaled float64 multiply/divide**, so the honest same-scale estimator this PR just unblocked — full-validation gradient in the same huge-but-finite regime, or `loss != 1.0` — hits the *expanded* finiteness gate and **raises** instead of returning the representable mean. The new test’s `full_validation_gradient=np.ones(4)` and `loss=1.0` are exactly the two knobs that keep `strength=4.0` and `reliability~1e-308`.

**Blocking: the mean is scale-free; strength × reliability is not**

Same-scale huge batch, this CPU:

| Input | `g²` | `E[‖g_b‖²]` | `strength` | `reliability` | `readiness` | public API |
|---|---|---|---|---|---|---|
| Test row: `ones(4)`, `loss=1.0`, 128×`[6.5e153]*4` | `4.0` | `1.6900e308` | `4.0` | `2.367e-308` | `9.467e-308` | returns (green) |
| Honest estimator: full-val = `[6.5e153]*4`, `loss=1.0` | `1.6900e308` | `1.6900e308` | `1.6900e308` | `1.0` | `1.6900e308` | returns |
| Same batch, `loss=0.5` | `1.6900e308` | `1.6900e308` | **`inf`** | `1.0` | **`inf`** | **`ValueError: optimization-readiness outputs must be finite`** |
| Full-val `[6.5e153]*4`, batches `[6.0e153]*4` (rel>1), `loss=1.0` | `1.6900e308` | `1.4400e308` | `1.6900e308` | `1.17` | **`inf`** | **same ValueError** |

`loss=0.5` is not exotic: `g² = 1.69e308` overflows `g²/loss` for every `loss < ~0.94`. Before this PR, `np.mean` overflowed `E` to `inf` and silently reported readiness `0.0`. After it, the mean is finite **and then `strength * reliability` overflows**, so the new gate fails closed. `_squared_norm` in this same file already scale-frees a sum of squares. `_mean_squared_norm` is a third in-tree spelling that stops one step too early.

**`math.ldexp` overflow is an uncaught `OverflowError`**

Python `math.ldexp(1.0, 1024)` raises `OverflowError: math range error`; it does not return `inf`. `np.ldexp(1.0, 1024)` returns `inf`. If `np.mean` of the scaled summands ever rounds up to `1.0`, line 230 raises `OverflowError` and the `isfinite` → `ValueError` path never runs. The helper’s documented contract is `ValueError`.

**The new test never enters the same-scale path**

Full-val is `ones(4)`, so `g²=4.0`. The protocol estimator is `full_validation_set_gradient` at the **same parameter scale** as the minibatches. That row is untested; on this CPU with `loss=0.5` it raises. Reliability is only `> 0.0`. `validate_development_result` is unchanged and still accepts a fabricated `0.0` readiness.

**What I would approve**

Keep `_mean_squared_norm` for the intermediate-sum overflow. Then: (1) scale-free strength and readiness too, or compute them from logs / ldexp — reuse `_squared_norm`’s “divide by max, reduce, multiply back”; (2) catch `math.ldexp` overflow (`OverflowError` → `ValueError`), or use `numpy.ldexp`; (3) pin same-scale full-val = `[6.5e153]*4` with `loss=1.0` and `loss=0.5`; rel>1 near the ceiling; two-row `1e308` mean equals `1e308`; `include_reliability=False` still returns strength for the huge batch.

Until `estimate_optimization_readiness(loss=0.5, full_validation_gradient=np.full(4, 6.5e153), batch_gradients=np.tile(np.full(4, 6.5e153), (128, 1)))` returns a finite diagnostic (or a deliberate `ValueError` that the test owns) instead of tripping the new catch-all gate, this does not lock the “representable mean survives” contract for the population-gradient estimator the protocol advertises.

PR #2600, opened 2026-08-26
fix(step7): skip a zero priority-propagation scale over an infinite TD

Kayvan-Zahiri:
## Problem

`_propagate_predecessor_priorities` in `alberta_framework/steps/step7.py` formed the propagated priority as a raw product:

```python
propagated = (
    jnp.asarray(propagation_scale, dtype=jnp.float32)
    * jnp.abs(td_error)
    / (1.0 + predecessor_distance)
)
return jnp.where(valid, jnp.maximum(memory_priorities, propagated), memory_priorities)
```

`planning_priority_propagation` is validated with `_require_nonnegative_real`, so `0.0` is admissible — and it is the setting that means *propagation disabled*. `td_error` is `rollout_td_signal`, produced by an imagined rollout of a learned world model and value function with no clamp on the path into this function, so it can reach `inf`.

The disabled path therefore evaluated `0 * inf` and produced `NaN`. Because `jnp.maximum` propagates `NaN`, that `NaN` was then written over **every live queue priority** — not just one slot. So asking Step 7 to leave the prioritized queue alone was the one configuration that destroyed it, and `_pop_prioritized_planning_anchor`'s `argmax` stays poisoned for the rest of the run.

Reproduced on `main` with a 3-entry queue at priorities `[0.5, 0.5, 0.5]`, `td_error = inf`, `propagation_scale = 0.0`:

```
propagation=0.0 -> [nan nan nan]
```

## Fix

Skip the scale at exactly zero, before the product:

```python
scale = jnp.asarray(propagation_scale, dtype=jnp.float32)
magnitude = jnp.abs(td_error)
scaled = jnp.where(scale == 0.0, jnp.zeros_like(magnitude), scale * magnitude)
propagated = scaled / (1.0 + predecessor_distance)
```

Fail-closed and minimal:

- The original `(scale * |td|) / (1 + d)` association is preserved, so non-zero scales are bit-exact — the guard is inserted without reassociating the float arithmetic.
- No NaN-to-zero masking of real values; only the exactly-zero-scale lane is short-circuited. A non-finite `td_error` under a *non-zero* scale still propagates.
- Public signature unchanged; no config, schema, or `outputs/` change.

## Tests

New `tests/test_step7_priority_propagation_zero_scale.py` asserts that the raw `scale * |td_error|` product is non-finite and that `jnp.maximum` propagates it, that the fixed path leaves every priority finite and unchanged when propagation is disabled, that an enabled scale still raises priorities to the expected exact value, and that entries beyond `memory_count` stay untouched.

```
tests/test_step7_priority_propagation_zero_scale.py  3 passed
tests/test_step7_production.py
tests/test_step7_hostile_validation.py               138 passed
ruff check                                           All checks passed!
```

This is a numeric fail-closed fix only. It creates no evidence, does not touch `outputs/`, and does not promote any claim.

### DIFF
--- a/alberta_framework/steps/step7.py
+++ b/alberta_framework/steps/step7.py
@@ -877,18 +877,28 @@ def _propagate_predecessor_priorities(
     td_error: Array,
     propagation_scale: float,
 ) -> Array:
-    """Propagate backup priority to predecessor transitions in the queue."""
+    """Propagate backup priority to predecessor transitions in the queue.
+
+    ``propagation_scale`` is a validated non-negative real, so ``0.0``
+    (propagation disabled) is admissible, while ``td_error`` comes from an
+    unclamped imagined rollout and can be non-finite.  The scale is skipped
+    at exactly zero: the raw ``0 * inf`` produced ``NaN``, and because
+    ``jnp.maximum`` propagates ``NaN`` that wrote ``NaN`` over every live
+    queue priority — so disabling propagation destroyed the queue instead
+    of leaving it alone.
+    """
     memory_size = memory_priorities.shape[0]
     valid = jnp.arange(memory_size, dtype=jnp.int32) < memory_count
     predecessor_distance = jnp.mean(
         (memory_next_observations - anchor_observation[None, :]) ** 2,
         axis=1,
     )
-    propagated = (
-        jnp.asarray(propagation_scale, dtype=jnp.float32)
-        * jnp.abs(td_error)
-        / (1.0 + predecessor_distance)
-    )
+    scale = jnp.asarray(propagation_scale, dtype=jnp.float32)
+    magnitude = jnp.abs(td_error)
+    # Keep the original (scale * |td|) / (1 + d) association so a non-zero
+    # scale stays bit-exact; only the zero-scale lane is short-circuited.
+    scaled = jnp.where(scale == 0.0, jnp.zeros_like(magnitude), scale * magnitude)
+    propagated = scaled / (1.0 + predecessor_distance)
     return jnp.where(valid, jnp.maximum(memory_priorities, propagated), memory_priorities)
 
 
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):
Reviewed at head `0b944e1ed38aee0743832cb4f713e7d05eef8938` against `c9aba7b5` (`main`), macOS, Python 3.12 venv.

Approving — this is the cleanest of the three concurrent `step7.py` zero-scale PRs.

### Why it is sound

`propagation_scale` is a config constant validated by `_require_nonnegative_real` (`step7.py:308-311`), so `0.0` is an admissible value and the `== 0.0` test is exact by construction. The guard tests `scale`, the same float32 value it multiplies by — no static-vs-runtime predicate gap.

The inline comment about preserving the `(scale * |td|) / (1 + d)` association is right, and it is why the enabled path stays bit-exact. I verified rather than assuming:

```
scale=0.0   td=inf    main: [nan, nan, nan]      PR: [0.5, 0.25, 0.125]
scale=0.5   td=inf    bit-identical
scale=0.5   td=-4.0   bit-identical
scale=1.0   (both)    bit-identical
scale=3.25  (both)    bit-identical
```

### The premise checks out

The docstring's claim that `jnp.maximum` propagates the `NaN` over *every* live priority is exactly what the measurement shows — all three entries go `NaN` on `main`, not just the one being propagated to. And the consumer at `:1178-1182` is `jnp.where(planning_ready & (strategy == "prioritized"), propagated, queue_priorities)`, an elementwise select with no finiteness filter, so the poisoned queue really does persist under the `prioritized` strategy. Disabling propagation destroying the queue is a genuine defect and the description is accurate.

The third test (`test_entries_beyond_memory_count_are_untouched`) pinning the `valid` mask is a good addition — it covers a behaviour the other two do not.

### Coordination note

Two other open PRs touch this file concurrently: [#2559](https://github.com/elizaOS/asi/pull/2559) and [#2590](https://github.com/elizaOS/asi/pull/2590), both adding a module-level `_skip_zero_scale` helper. This PR inlines its `jnp.where` instead, so it does not participate in that collision — but if all three land, consolidating on one helper would be worth a follow-up.

### Checks

- `tests/test_step7_priority_propagation_zero_scale.py`: 3 passed.

PR #2601, opened 2026-08-26
fix(benchmarks): skip zero AdaLin alpha before 0*inf poisons base activation

Kayvan-Zahiri:
## Problem

`ADALIN_PROTOCOL` declares mechanism-off as `alpha_zero_exact_base_activation`. `_adalin_transaction` still evaluated:

```python
candidate = activation + alpha * x * gate
valid = jnp.all(jnp.isfinite(x)) & ... & jnp.all(jnp.isfinite(candidate))
```

Under JAX, `alpha * x * gate` still evaluates both sides of a `where`, so `alpha=0` with infinite `x` computes `0 * inf = NaN`, poisons `candidate`, and rejects the transaction — even when the base activation is finite (`tanh(±inf)` saturates; `relu(-inf)` is `0`). That breaks the declared mechanism-off reduction.

Reproduced on `main`:

```
adalin_tanh_transaction(+inf, alpha=0) -> (0.0, valid=False)  # should be (1.0, True)
```

## Fix

Skip the zero-scale product before it reaches the candidate (`_skip_zero_scale`), and only require finite `x` on coordinates where `alpha != 0`. Nonzero alpha keeps fail-closed rejection of infinite `x`. Finite arithmetic stays bit-identical for nonzero alpha.

## Tests

- `test_zero_alpha_returns_exact_base_activation_on_infinite_x` — fails on unfixed source, passes with the fix
- `test_nonzero_alpha_still_rejects_infinite_x` — fail-closed preserved

`tests/test_adalin.py`: 21 passed. `ruff` clean.

### DIFF
--- a/alberta_framework/benchmarks/adalin.py
+++ b/alberta_framework/benchmarks/adalin.py
@@ -73,6 +73,11 @@ def _trusted_float_array(value: object, *, name: str) -> Array:
     return array
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Keep finite multiplication exact while repairing zero-scaled poison."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def _adalin_transaction(
     x: Array, alpha: Array, *, activation: Array, derivative: Array
 ) -> tuple[Array, Array]:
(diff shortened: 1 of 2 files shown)


kenjohnscreates (changes requested):
## The rewrite re-associates the product, so finite results move

The `alpha == 0` guard itself is fine — `alpha` is a validated host scalar, so `== 0.0` is exact by construction — and relaxing `x`'s finiteness requirement when `alpha == 0` is consistent with `x` only reaching `candidate` through the zero-scaled term.

The problem is that the rewrite silently changes the multiplication order for **every** input, not just the zero case:

```python
# before
candidate = activation + alpha * x * gate      # (alpha * x) * gate
# after
correction = _skip_zero_scale(alpha, x * gate) # alpha * (x * gate)
candidate = activation + correction
```

Float multiplication is not associative, so this shifts results on ordinary finite inputs. Over 200,000 random finite float32 triples drawn from the natural ranges (`alpha in [0.05, 2]`, `x in [-3, 3]`, `derivative in [0, 1]`):

```
finite-input reassociation mismatches: 70047/200000 (35.023%)
  alpha=1.8733913 x=1.8951213 derivative=0.0027385002 gate=0.99999076
    (alpha*x)*gate = 3.5502708   bits=1080244131
    alpha*(x*gate) = 3.550271    bits=1080244132
```

End-to-end through the real `_adalin_transaction` on that input, `candidate` differs bitwise between the two branches:

| | `candidate` bits |
|---|---|
| `main` | `1080244131` |
| this PR | `1080244132` |

One ulp, but on ~35% of inputs and in `benchmarks/`, where recorded numbers are the product. Any baseline measured before this lands will not reproduce bit-for-bit after it.

This also makes the helper's docstring inaccurate — it claims to "Keep finite multiplication exact", which is the one thing the rewrite does not do.

### Suggested fix

Keep the original association and guard only the scale:

```python
correction = _skip_zero_scale(alpha * x, gate)
```

or, to keep `alpha` as the tested scale without moving the parentheses:

```python
correction = jnp.where(alpha == 0.0, jnp.zeros_like(x * gate), alpha * x * gate)
```

Either preserves `(alpha * x) * gate` on the non-zero path, so finite results stay bit-identical to `main` and the docstring becomes true. Worth a test asserting bitwise equality with the pre-PR expression on a finite input, since the current tests only exercise the `alpha == 0` branch.

lalalune:
Closing after review: the new where(alpha == 0, 0, alpha*x*gate) branch suppresses the derivative with respect to alpha exactly at its zero initialization, preventing the AdaLin coefficient from beginning to learn. It also relaxes the existing nonfinite-input rejection contract. The injected-infinity forward tests do not cover this normal finite-input learning regression.

PR #2602, opened 2026-08-26
fix(benchmarks): freeze micro-continual SGD when step_size is zero under infinite grads

Kayvan-Zahiri:
## Problem

`_freeze_micro_hyperparameters` accepts any finite `step_size`, including `0.0`. The SGD raw arm updated with:

```python
decay = 1.0 - step_size * weight_decay
params[name] * decay - step_size * grads[name]
```

When `step_size=0`, `decay` is exactly `1.0`, but `0 * inf` grads still evaluate to NaN and poison finite parameters on a freeze step.

## Fix

Extract `_sgd_raw_param_update` and return the input params unchanged when `step_size == 0.0`, before any grad scaling.

## Tests

- `test_sgd_raw_zero_step_size_preserves_params_under_infinite_grads`
- `test_sgd_raw_nonzero_step_size_still_applies_to_finite_grads`

Focused suite green; `ruff` clean.

### DIFF
--- a/alberta_framework/benchmarks/micro_continual.py
+++ b/alberta_framework/benchmarks/micro_continual.py
@@ -824,12 +824,32 @@ class MicroArmSpec:
         )
 
 
+def _sgd_raw_param_update(
+    params: dict[str, Array],
+    grads: dict[str, Array],
+    *,
+    step_size: float,
+    weight_decay: float,
+) -> dict[str, Array]:
+    """Apply plain SGD with decoupled decay, skipping a frozen step size.
+
+    ``step_size=0`` is a supported freeze. Scaling infinite grads by that zero
+    step size evaluates ``0 * inf`` and would poison finite parameters.
+    """
+    decay = 1.0 - step_size * weight_decay
+    if step_size == 0.0:
+        return {name: params[name] for name in params}
+    return {
+        name: params[name] * decay - step_size * grads[name] for name in params
+    }
+
+
 def _make_sgd_raw_learner(
     hp: Mapping[str, float],
 ) -> tuple[LearnerInitFn, ScreeningStepFn]:
     """Plain SGD (optionally with decoupled decay) on raw inputs."""
     step_size = hp["step_size"]
-    decay = 1.0 - step_size * hp["weight_decay"]
+    weight_decay = hp["weight_decay"]
 
     def init_fn(params: dict[str, Array]) -> Array:
         del params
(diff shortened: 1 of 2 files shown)


kenjohnscreates (approved):
Reviewed at head `9fe0005a3d4bd0ea0249ba0f8fd7e0fd0129ae45` against `c9aba7b5` (`main`), macOS, Python 3.12 venv.

Approving. This picks the strongest of the two in-repo idioms, and it matters here more than in most of this series.

### The idiom choice is right

`step_size` is a Python float from the frozen hyperparameter mapping, so the guard is a **static Python-level `if`**, not a `jnp.where`. No select is introduced into the traced graph at all. That is the `baseline_optimizers.py:193` form rather than the runtime form, and it is the correct call here specifically because `micro_continual` is the IPMNIST screening lane's cheap inner loop — a `jnp.where` on every parameter update would have been a real cost in the one place this repo runs hottest.

### Bit-exactness on the enabled path — verified

36 configurations (`step_size` in `{0.1, 0.01, 1e-3, 0.5, 1.0, 3.0}` × `weight_decay` in `{0.0, 0.01, 0.1}` × finite and near-overflow gradients), comparing the refactored helper against `main`'s inline `params * decay - step_size * grads`:

```
configs compared: 36, bitwise differences: 0
```

Moving `decay = 1.0 - step_size * weight_decay` from closure-creation time into the function does not change the value: both operands stay Python floats, so it constant-folds identically.

### The zero lane

```
grads=[1e+20, -3.0, 0.0]  main: [1.0, -2.0, 0.5]   PR: [1.0, -2.0, 0.5]
grads=[inf,   -3.0, 0.0]  main: [nan, -2.0, 0.5]   PR: [1.0, -2.0, 0.5]
grads=[nan,    1.0, 2.0]  main: [nan, -2.0, 0.5]   PR: [1.0, -2.0, 0.5]
```

Returning `params` unchanged is the right value, not just a safe one: at `step_size == 0` the decay factor is `1.0 - 0.0 * weight_decay == 1.0` exactly, so `params * decay` is `params`.

Small bonus the PR does not claim: the early return also makes the frozen step robust to a pathological `weight_decay`. On `main`, `step_size=0.0` with `weight_decay=inf` gives `decay = 1.0 - 0.0*inf = nan` and wipes the parameters; this head returns them intact (measured for both `inf` and `nan`).

### Reachability

`_freeze_micro_hyperparameters` validates entries with `_require_finite_real` (`micro_continual.py:195`) — finite real, **no** positivity bound — so `step_size=0.0` is admissible in a `MicroArmSpec`. No registered arm currently uses it (the three in-tree specs all pass `0.01`), so this is a legal-but-unexercised configuration rather than one on a shipped path. That is worth stating plainly in the PR body; it does not change my assessment of the fix.

### Checks

- `tests/test_micro_continual.py`: 229 passed.
- No other open PR touches `alberta_framework/benchmarks/micro_continual.py`.

PR #2603, opened 2026-08-26
fix(benchmarks): skip zero replay/mechanism scales before 0*inf poisons matched grads

Kayvan-Zahiri:
## Problem

`replay_weight` and `mechanism` are protocol encodings that accept `0.0` as the mechanism-off reduction. Two updates still scaled hostile infinite tensors by those zeros:

1. `task_gradient + replay_weight * replay_gradient` — disabled replay still poisoned the task gradient with `0 * inf`.
2. `mechanism * head_step_size * update_scale * feature_gradient` — disabled feature adaptation still poisoned prompt/scale/shift, so the finite-state gate rejected an otherwise valid matched step.

## Fix

Add the house `_skip_zero_scale` helper and use it for both products so a zero scale repairs infinite operands before they reach committed state. Nonzero scales keep bit-exact multiplication and fail-closed behavior on infinite grads.

## Tests

Three regressions in `tests/test_replay_frozen_ipmnist.py`. Full file: 15 passed. `ruff` clean.


saltwire-dev (changes requested):
## Request changes at exact head `8c050a4abe767046f587df4aa77a2e6430ca701d`

Reviewed at exact head `8c050a4abe767046f587df4aa77a2e6430ca701d` against production `alberta_framework/benchmarks/replay_frozen_ipmnist.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4.

Skipping a protocol-off `0.0` before `0 * inf` is the right *direction* for a declared mechanism-off reduction — origin’s `replay_weight * replay_gradient` and `mechanism * head_step_size * update_scale * feature_gradient` are raw products, and IEEE `0 * inf` is NaN. The hole is that the defect named in the PR is **the finite-state gate rejecting an otherwise valid matched step**, and on the method-2 production path that gate still rejects: `feature_gradient = state.readout @ (-error)` is not an independent poison. The same `error` (or the same overflowing `phi`) is immediately written into `readout` with **no** skip, so `floating_tree_is_finite(candidate)` is still false. The three new tests never call `make_replay_context_learner` / `make_frozen_feature_learner` / `step_fn`; they only exercise a new Python-`if` helper that is not the in-tree house 0*inf contract (`zero_if_collapsed_infinity`). Deleting both production wirings would still pass all three new tests. On this CPU, the PR’s own `mechanism=0` + infinite-`error` method-2 update still yields **non-finite `readout`**, so `select_transaction` still keeps the old state.

**Blocking: named method-2 `mechanism=0` path still fails the finite-state gate via unguarded readout; tests never enter `step_fn`; leftover `0*inf` after the skip; helper is not the house contract**

On this CPU, a replica of the **exact method-2 formulas at this SHA** with protocol `mechanism=0.0`, `head_step_size=1e-2`, finite prompt/scale/shift/readout/phi, and hostile `error = [inf, 0, 0]`:

```
origin inf-error candidate_finite False  (prompt/scale/shift NaN, readout inf)
head   inf-error candidate_finite False  (prompt/scale/shift repaired, readout still inf)
head readout has inf True
```

So the named “otherwise valid matched step” is **still rejected** by `floating_tree_is_finite(candidate)`. There is no production method-2 path where `feature_gradient` is infinite while that outer stays finite: `error` inf + `phi` finite → outer inf; `error` inf + `phi` 0 → `0 * inf` NaN outer; overflowing `phi` → outer inf. The new mechanism test never constructs a `FrozenFeatureState` or calls `step_fn`.

Leftover `0 * inf` **after** the skip, same method-2 block: `scale = state.scale - feature_step * (phi - state.shift)`. With `feature_step` zeros and infinite `phi`, this is `0 * inf` → NaN `scale` on **this head**. Reproduced: `scale leftover [nan, 1, 1, 1]`. Same file still has raw `enabled * context_logits` and `mechanism * state.prompt` / `mechanism * state.shift` in `features()`.

The helper itself is not the house contract already in this repo at this SHA (`update_safety.zero_if_collapsed_infinity`). It (1) uses a Python `if scale != 0.0` that constant-folds only because `hp["replay_weight"]` / `mechanism` are closed-over Python floats; (2) computes the poison product **first**; (3) on the zero branch, `jnp.where(jnp.isfinite(value), product, zeros)` **zeros unrelated NaNs**. House helper on `0 * [nan, 2]` with `collapsed=True` keeps `nan` in slot 0 (`[nan, 0]`); this head’s helper returns `[0, 0]`.

**What I would approve**

Keep a skip at the two named products, but lock the **production commit**, not a helper unit test. (1) Skip every scale that the method-2 / replay-off commit actually multiplies — `readout += head_step_size * update_scale * outer(phi, error)` must not write inf/NaN when `mechanism=0`; `feature_step * (phi - state.shift)` must not re-introduce `0 * inf` after a successful skip. (2) Use the house 0*inf helper (or the same mask) — do not invent a Python-`if` helper that zeros unrelated NaNs. (3) Drive the new tests through `step_fn` with protocol hyperparameters, a real `ReplayContextState` / `FrozenFeatureState`, and assertions on **committed** state: named method-2 `mechanism=0` + hostile `error` → committed `prompt/scale/shift/readout` stay finite **and** `select_transaction` applies; named replay `replay_weight=0` → committed params equal to the task-only Adam step; control `mechanism=1` / `replay_weight=1` with infinite grads still fail closed; unrelated NaN operand with scale 0 remains NaN (house contract), not `[0, 0]`.

Until `make_frozen_feature_learner(frozen_hyperparameters(method=2.0, mechanism=0.0))[1](...)` on a finite incoming state with the PR’s infinite-`feature_gradient` source **commits** a finite candidate (or the PR stops claiming the finite-state gate currently rejects an otherwise valid step), this does not lock “skip zero replay/mechanism scales before 0*inf poisons matched grads” — it locks a helper that the new tests call directly, leaves origin’s readout poison on the named method-2 path, and still NaNs `scale` via leftover `0 * inf(phi)`.

kenjohnscreates (changes requested):
## Disposition: **request changes** (concur with saltwire-dev: the production blockers are untouched; the only post-review change is a test)

Exact head reviewed: `f5d4271a` (merge of main plus "Exercise disabled replay with finite overflowing stored examples" since saltwire-dev's review at `8c050a4a`). Executed in an isolated worktree.

The post-review diff adds one test through `make_replay_context_learner` with `replay_update=0.0` and `3e38` stored examples, which is a real step-through for the replay side. The library file is byte-for-byte what saltwire-dev reviewed, so each of that review's measured items still holds at this head:

- `_skip_zero_scale(0.0, [nan, 2, inf])` returns `[0, 0, 0]`: the input NaN in slot 0 is not a `0*inf` product, and it is erased. The house helper `update_safety.zero_if_collapsed_infinity(0*v, isinf(v), collapsed=True)` returns `[nan, 0, nan]`, preserving it. A mechanism-off reduction that launders unrelated NaNs into zeros is the "isfinite-where guard" class the repo has rejected elsewhere.
- The raw products saltwire-dev named are still present: `base + enabled * context_logits` (line 161), `mechanism * state.prompt` / `mechanism * state.shift` (289), `readout + head_step_size * update_scale * jnp.outer(phi, error)` (330) and `scale - feature_step * (phi - state.shift)` (341). So on the method-2 `mechanism=0` path the readout still receives the infinite outer product and the finite-state gate still rejects the step the PR body says it should accept, and `scale` still picks up `0 * inf` after the skip.
- The helper is a Python `if scale != 0.0` that only works because the scales are closed-over floats.

`tests/test_replay_frozen_ipmnist.py`: 16 passed; ruff clean, but the tests still never drive `step_fn` with `method=2.0, mechanism=0.0` and a hostile `error` to a committed finite candidate. saltwire-dev's "what I would approve" list is the right spec; none of its three items is in this head.

---

tokenmaxxor (changes requested):
﻿## Review

**Recommendation: fix** exact head `f5d4271a652d6700adfe7b22993c8c8419bcd75f`.

I am not the author. Static review only; I did not re-execute the isolated measurements in the outstanding request-changes reviews.

`_skip_zero_scale` is the right idea for `replay_weight=0` and `mechanism=0`, and the new overflowing-buffer test is a real step-through. I concur with the current-head CR: `scale==0` currently zeros **all** non-finite `value` entries, including an input NaN that is not a `0*inf` product. That is broader than the house `zero_if_collapsed_infinity` contract. Repair the helper to skip only collapsed infinities, then this can merge.

tokenmaxxor (changes requested):

**Recommendation: fix** exact head `ab73e7923573fcbf3accfd455cd9494aca219a2b`.

I am not the author. Prior CHANGES_REQUESTED still applies. Compare `f5d4271a652d6700adfe7b22993c8c8419bcd75f...ab73e7923573fcbf3accfd455cd9494aca219a2b` is only `tests/test_initializers_validation.py` (allocator preflight, no ~2 GiB matrix). That does not address the standing production defect on this PR.

PR #2604, opened 2026-08-26
fix(streams): bound Gymnasium learn_from_trajectory scan length before lax.scan

Kayvan-Zahiri:
## Problem

`learn_from_trajectory` accepted any leading axis up to signed int32 and handed it straight to `jax.lax.scan`. Resource preflight only catches wide rows, so a long-but-narrow array (for example `T=200_000`, `feature_dim=1`) still forces JAX to trace/compile a huge scan and hang the process before any step runs — the same hang class already closed for `streams.feature_discovery`, `core.sarsa`, and `core.horde`.

## Fix

Introduce `_GYMNASIUM_TRAJECTORY_MAX_STEPS = 10_000` and reject overflow lengths before `lax.scan`.

## Tests

- Document the ceiling.
- Reject first overflow (`10001`) and hang-class (`200_000`) lengths with a scan spy asserting `lax.scan` never runs.

### DIFF
--- a/alberta_framework/streams/gymnasium.py
+++ b/alberta_framework/streams/gymnasium.py
@@ -44,6 +44,14 @@ if TYPE_CHECKING:
 
 _INT32_MAX = 2**31 - 1
 _INT32_MIN = -(2**31)
+# ``learn_from_trajectory`` hands caller-supplied step arrays straight to
+# ``jax.lax.scan`` with no bound tighter than ``_INT32_MAX``. Bounding only
+# by ``_INT32_MAX`` still lets a caller force JAX to trace/compile a scan of
+# up to ~2 billion steps (resource preflight only catches wide rows, not
+# long-but-narrow ones), hanging the process well before any step executes.
+# Matches the ceiling convention for sibling scan-driven array loops
+# (``streams.feature_discovery``, ``core.sarsa``, ``core.horde``).
+_GYMNASIUM_TRAJECTORY_MAX_STEPS = 10_000
 _ACTUAL_INT_TYPES = frozenset(
     {
         int,
@@ -499,6 +507,11 @@ def learn_from_trajectory(
         metrics have shape (num_steps, 3) with columns
         [squared_error, error, mean_step_size]; a learner constructed with
         a normalizer emits a fourth column, normalizer_mean_var.
+
+    Raises:
+        ValueError: If ``num_steps`` is outside
+            ``[1, _GYMNASIUM_TRAJECTORY_MAX_STEPS]`` or the arrays are otherwise
+            malformed for a scan.
     """
     if type(learner) is not LinearLearner and not isinstance(learner, LinearLearner):
         raise TypeError("learner must be an instance of LinearLearner")
@@ -528,8 +541,11 @@ def learn_from_trajectory(
         raise ValueError("observations must be a 2D array of shape (num_steps, feature_dim)")
     if tgt_ndim not in (1, 2):
         raise ValueError("targets must be a 1D or 2D array")
-    if not 1 <= num_steps <= _INT32_MAX:
-        raise ValueError("observations sequence length must be an integer in [1, 2147483647]")
+    if not 1 <= num_steps <= _GYMNASIUM_TRAJECTORY_MAX_STEPS:
+        raise ValueError(
+            "observations sequence length must be an integer in "
+            f"[1, {_GYMNASIUM_TRAJECTORY_MAX_STEPS}]"
+        )
     if target_steps != num_steps:
         raise ValueError("targets sequence length must match observations sequence length")
     if feature_dim is None or not 1 <= feature_dim <= _INT32_MAX:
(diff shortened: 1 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `006e8cf0e9479288b900a10d7726221ecd4b8d5b`

Reviewed at exact head `006e8cf0e9479288b900a10d7726221ecd4b8d5b` against production `alberta_framework/streams/gymnasium.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4.

Capping `learn_from_trajectory` at `_GYMNASIUM_TRAJECTORY_MAX_STEPS = 10_000` is the right *consumer* repair for the PR body’s inverted row — on this CPU, `np.zeros((10001, 3), float32)` / `np.zeros((200_000, 1), float32)` now raise `ValueError: observations sequence length must be an integer in [1, 10000]` and `jax.lax.scan` is never entered; `T=10_000` passes that gate (reaches `learner.init`). The hole is that **the documented producer is still unbounded**. Module docstring: “Trajectory collection: Collect data using Python loop, then learn with scan.” `collect_trajectory` still does `num_steps = _require_positive_int32(...)` (`[1, 2147483647]`) and `for _ in range(num_steps)`, with no mention of `_GYMNASIUM_TRAJECTORY_MAX_STEPS`. The sibling the PR cites as already closed (`collect_feature_discovery_stream`) bounds **the collector that drives the scan**. After this PR the Gymnasium pair no longer composes: `collect_trajectory(..., num_steps=10_001)` is legal, `learn_from_trajectory` on that output is not. Resource preflight still lets the PR’s own hang-class size (`T=200_000`) through `collect_trajectory`. The new tests never import `collect_trajectory` and never scan at the accepted ceiling.

**Blocking: consumer capped; collector still `[1, 2**31-1]`**

Same SHA, this CPU:

| Call | Result |
|---|---|
| `_require_positive_int32("num_steps", 10_000)` | **10000** (collect accepts) |
| `_require_positive_int32("num_steps", 10_001)` | **10001** (collect accepts) |
| `_require_positive_int32("num_steps", 200_000)` | **200000** (collect accepts) |
| `_require_positive_int32("num_steps", 2**31-1)` | **2147483647** (collect accepts) |
| `_require_float32_allocation(..., 200_000 * 3)` | **OK** (2.4 MB) |
| `learn_from_trajectory` `T=10_000` | length gate **passes** |
| `learn_from_trajectory` `T=10_001` / `T=200_000` | **`ValueError: ... [1, 10000]`**, `lax.scan` calls **0** |

`collect_trajectory` source at this head does not contain `_GYMNASIUM_TRAJECTORY_MAX_STEPS`. A caller following the module recipe with `num_steps=10_001` (or the PR’s own `200_000`) pays a Python `env.step` loop of that length, stacks the arrays, then dies in `learn_from_trajectory`. That is not “reject overflow lengths before any step executes”; it is reject **after** the collector already ran.

The feature-discovery close at this SHA is on the **collector**: it refuses `num_steps > 10_000` before `jnp.arange` / `lax.scan`. Gymnasium splits the work: `collect_trajectory` is a Python `range(num_steps)` hang with an int32 ceiling; `learn_from_trajectory` is the scan. Capping only the second leaves the first — and the documented pair — open. `_require_float32_allocation` does not save it: a 3-wide row is legal out to `INT32_MAX/12 ≈ 1.79e8` steps. `collect_trajectory` is also the worse hang of the two: `learn_from_trajectory` has no `jnp.arange(num_steps)`; the in-module way to force “millions of steps before any learning step executes” is the Python collector this PR did not touch.

**The new tests never enter the collect → learn path (or the accepted ceiling)**

The test file does not import `collect_trajectory`. Both overflow rows are hand-built `jnp.zeros`. `test_documented_trajectory_scan_ceiling` only asserts the constant; `test_valid_trajectory_learning_runs_cleanly` uses **8** steps. Nothing proves `T=10_000` actually scans, or that `collect_trajectory(..., num_steps=10_001)` refuses before the first `env.step`. `learn_from_trajectory_normalized` is a one-line alias, so it inherits the consumer cap — and the same collect mismatch.

**What would lock the contract**

Bound `collect_trajectory` with the same ceiling (reject in `_require_positive_int32`’s slot, **before** allocation and **before** `env.reset` / the Python loop), and add a test that `collect_trajectory(..., num_steps=10_001)` raises without stepping the env. Keep a `learn_from_trajectory` spy test for hostile hand-built arrays. Until `collect_trajectory(env, policy, num_steps=10_001)` fails closed the same way `learn_from_trajectory` now does, this does not lock the “reject overflow lengths before any step executes” contract the PR advertises for the public Gymnasium collect→learn recipe.

lalalune:
Reviewed: Closing after resource review: the arbitrary 10000-step ceiling rejects legitimate longer research trajectories and is not tied to a measured memory/latency budget. lax.scan does not unroll or compile once per step as the rationale suggests. Existing shape/resource checks should be extended from concrete resource needs, not a fixed research-horizon policy.

PR #2605, opened 2026-08-26
fix(benchmarks): require exact dict/list JSON containers in matched protocol parse

Kayvan-Zahiri:
## Problem

`parse_forager_matched_protocol` accepts in-memory values, not only decoded JSON text. Validation walked containers with `isinstance(..., Mapping)` / `isinstance(..., list)` and accepted `isinstance(..., float)` for finiteness.

A `dict` subclass with a hostile `__iter__` therefore entered the Mapping branch and fired during key enumeration instead of being rejected as a non-JSON value:

```
parse_forager_matched_protocol(HostileDict(...))
-> AssertionError: HostileDict.__iter__
```

## Fix

Gate `_validate_json_complexity`, `_require_object`, and `_require_array` on exact `dict` / `list` / `float` types (house JSON-builtin identity), matching the existing exact-type checks for strings and the final non-JSON reject list.

## Tests

Two hostile regressions in `tests/test_forager_matched_protocol_hostile_validation.py`. Protocol suites: 79 passed. `ruff` clean.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_protocol.py
+++ b/alberta_framework/benchmarks/forager_matched_protocol.py
@@ -1086,7 +1086,7 @@ def _validate_json_complexity(value: Any) -> None:
             raise ForagerMatchedProtocolError("protocol exceeds the JSON node limit")
         if depth > _MAX_JSON_NESTING:
             raise ForagerMatchedProtocolError("protocol exceeds the JSON nesting limit")
-        if isinstance(item, Mapping):
+        if type(item) is dict:
             for key in item:
                 if type(key) is not str:
                     raise ForagerMatchedProtocolError("JSON object keys must be strings")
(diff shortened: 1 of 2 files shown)


lumix5 (approved):

**Disposition: merge.** Replacing `isinstance(item, Mapping/list/float)` with exact-type checks in `_validate_json_complexity` and `_require_object/_require_array` closes the container-subclass surface the hard way: a `dict`/`list`/`float` subclass no longer takes the container/float branches and falls through to the final exact-type rejection, so hostile `__iter__`/`__len__`/`__float__` hooks never execute — the same threat model as the #2172/#2173 family, resolved by rejection rather than by traversal, and consistent with this module's existing exact-type discipline (the final elif was already exact). Ran the full patched hostile-validation suite in a throwaway container (network denied): 11 passed. ci-ok green at head.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_protocol.py` (_validate_json_complexity, decode_strict_json, _require_object). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2613, opened 2026-08-26
fix(benchmarks): reject non-finite L2ER paired confidence bounds

Kayvan-Zahiri:
## Summary
- Require finite mean/lower/upper before classifying L2ER paired CI outcomes.
- NaN bounds previously fell through to a false `inconclusive` because IEEE comparisons involving NaN are always false.

## Test plan
- [x] `pytest tests/test_l2er_matched_development.py -q`
- [x] `ruff check` on touched files

### DIFF
--- a/alberta_framework/benchmarks/l2er_matched_development.py
+++ b/alberta_framework/benchmarks/l2er_matched_development.py
@@ -420,6 +420,11 @@ def _outcome(deltas: tuple[float, ...]) -> tuple[float, float, float, str]:
     stderr = float(values.std(ddof=1) / math.sqrt(len(values)))
     lower = mean - _T95_DF2 * stderr
     upper = mean + _T95_DF2 * stderr
+    # Non-finite bounds must fail closed: IEEE comparisons involving NaN are
+    # always false, so ``lower > 0.0`` / ``upper <= 0.0`` alone cannot reject
+    # ``nan``/``±inf`` and would otherwise report a false "inconclusive".
+    if not (math.isfinite(mean) and math.isfinite(lower) and math.isfinite(upper)):
+        raise ValueError("paired confidence interval bounds must be finite")
     outcome = "supported" if lower > 0.0 else "rejected" if upper <= 0.0 else "inconclusive"
     return mean, lower, upper, outcome
 
--- a/tests/test_l2er_matched_development.py
+++ b/tests/test_l2er_matched_development.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import hashlib
+import math
 import os
 from copy import deepcopy
 from pathlib import Path
@@ -239,6 +240,20 @@ def test_three_seed_interval_uses_student_t_not_normal_critical_value() -> None:
     assert critical == 4.302652729749464
     assert isinstance(critical, float)
     assert critical.hex() == "0x1.135ea98e146bbp+2"
+
+
+@pytest.mark.parametrize(
+    "deltas",
+    [
+        (math.nan, 0.0, 0.0),
+        (math.inf, math.inf, math.inf),
+        (-math.inf, 0.0, 0.0),
+    ],
+)
+def test_outcome_rejects_nonfinite_interval_bounds(deltas: tuple[float, float, float]) -> None:
+    """IEEE lower/upper comparisons are always false for NaN, so fail closed."""
+    with pytest.raises(ValueError, match="must be finite"):
+        matched._outcome(deltas)
 
 
 @pytest.mark.parametrize("version", ("v1",))


saltwire-dev (changes requested):
## Request changes at exact head `127b3461de51a9de4a64fd46bee4e21d21ac0133`

Reviewed at exact head `127b3461de51a9de4a64fd46bee4e21d21ac0133` against production `alberta_framework/benchmarks/l2er_matched_development.py` and `ipmnist_screening.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4.

Requiring finite CI endpoints before the IEEE trichotomy is the right *direction* for a direct `_outcome` call with a NaN seed — on this CPU, `(nan, 0, 0)` used to yield `mean=stderr=lower=upper=nan` and `"inconclusive"` because `nan > 0.0` and `nan <= 0.0` are both False. This PR’s `math.isfinite` guard raises `ValueError("paired confidence interval bounds must be finite")` on that row. The hole is that **this guard is unreachable on every report path this module actually ships**, and the trichotomy it still runs on reachable inputs mislabels a true null. `build_report` only feeds `_outcome` after `_canonical_result` reconstructs a frozen `ScreeningRunResult`, whose `__post_init__` already rejects non-finite `per_task_accuracy` and requires every entry in `[0, 1]`. Valid accuracies produce deltas in `[-1, 1]`; the frozen t(df=2) interval on three such numbers is always finite on float64. The new test only calls `_outcome` directly with values those gates already refuse. Meanwhile `_outcome((0.0, 0.0, 0.0))` — three matched seeds where an arm *equals* control, which `ScreeningRunResult` happily accepts — still returns `"rejected"` because the classifier uses `upper <= 0.0`. That report **does** pass `validate_report` and would publish.

**Blocking: finite-bounds guard is dead on the report path; zero-width null still ships as `"rejected"`**

On this CPU:

| deltas | old IEEE label | this PR | reachable from `build_report`? |
|---|---|---|---|
| `(nan, 0, 0)` (new test) | **inconclusive** | **RAISE** | **no** — `__post_init__` rejects non-finite accuracy |
| `(inf, inf, inf)` (new test) | inconclusive | RAISE | **no** — accuracy must lie in `[0, 1]` |
| `(-1, 1, 1)` (max-spread legal deltas) | inconclusive | inconclusive | yes, already finite |
| `(0, 0, 0)` (arm equals control on all 3 seeds) | **rejected** | **rejected** | **yes** |

The PR body says NaN bounds “fell through to a false `inconclusive`.” That is true of a *bare* `_outcome` call. It is not true of a matched-development report: `_canonical_result` re-runs `ScreeningRunResult.__post_init__` before any subtraction, `validate_report` refuses non-finite `deltas` before it ever reaches the new raise, and `_publish_report` encodes with `json.dumps(..., allow_nan=False)`. The five new lines change the exception *type* of a path that already failed closed.

The live classifier bug is `upper <= 0.0`. Frozen plan is `two_sided_student_t` / `null_delta=0.0`. For a closed interval, two-sided reading of H0: δ=0 is supported if `lower > 0`, rejected if `upper < 0`, inconclusive if `lower ≤ 0 ≤ upper` (including a point mass at 0). Sibling activation-feature `_paired_summary` (open #2614, not this head) already uses `upper < 0.0`. `_outcome((0.0, 0.0, 0.0))` is the exact equal-arm row: stderr 0, interval `{0}`, null is the entire interval, label `"rejected"`. A mechanism that matches `l2er_mechanism_off` on every seed is recorded as *significantly worse*. `validate_report` will accept it.

Existing `test_matched_report_is_complete_paired_and_permanently_nonpromoting` uses offsets `{l2_only: +0.02, er_only: −0.02, combined: +0.03}` — never zero deltas — so it stays green.

**Tests lock NaN-bound `_outcome`, not the report contract, not the null**

All three fixtures produce **NaN bounds**, not infinite bounds. `(inf, inf, inf)` has `mean=inf` but `std(ddof=1)=nan` (inf−inf). An implementation that only `math.isnan`’d mean/lower/upper would pass this test. Nothing constructs a `ScreeningRunResult` with NaN accuracy and calls `build_report`, uses three equal zero deltas through `build_report`/`validate_report`, or asserts `upper < 0.0` as the rejection predicate.

**What would lock the contract**

1. Stop claiming a report-level NaN/`inconclusive` leak this dataclass already closes. If `_outcome` is still a public-ish classifier, fail closed on non-finite **inputs** (and `len(deltas) == 3`) *before* `np.mean`/`std`.
2. Change the trichotomy to `supported if lower > 0 else rejected if upper < 0 else inconclusive`. `_outcome((0.0, 0.0, 0.0))` must be `"inconclusive"`.
3. `build_report` with identical finite accuracies on control and `l2er_l2_only` → paired outcome `"inconclusive"`, not `"rejected"`. `ScreeningRunResult` with NaN accuracy still fails in `__post_init__`, not in `_outcome`.

Until `(0, 0, 0)` is inconclusive on the same `_outcome` `build_report` ships, this does not lock the paired-CI contract the PR body advertises, and the new raise does not close a hole the report path could actually publish.

lalalune:
Closing as redundant internal hardening. Both _outcome callers build deltas from canonical validated per-task/mean accuracies with the fixed seed roster. The new tests only call _outcome directly with NaN/Infinity; bounded [-1,1] paired differences cannot overflow these fixed three-seed calculations. No supported loader or producer failure is demonstrated, so this adds source churn and duplicate validation without a workflow benefit.

PR #2614, opened 2026-08-26
fix(evaluation): reject non-finite activation-feature paired CI bounds

Kayvan-Zahiri:
## Summary
- Require finite mean/lower/upper/standard_error before classifying activation-feature paired CI outcomes.
- NaN bounds previously fell through to a false `inconclusive` because IEEE comparisons involving NaN are always false.

## Test plan
- [x] `pytest tests/test_activation_feature_campaign.py -q`
- [x] `ruff check` on touched files

### DIFF
--- a/alberta_framework/evaluation/activation_feature_campaign.py
+++ b/alberta_framework/evaluation/activation_feature_campaign.py
@@ -1099,6 +1099,16 @@ def _paired_summary(
         margin = _T_CRITICAL_DF4_BONFERRONI_8 * standard_error
         lower = mean - margin
         upper = mean + margin
+        # Non-finite bounds must fail closed: IEEE comparisons involving NaN are
+        # always false, so ``lower > 0.0`` / ``upper < 0.0`` alone cannot reject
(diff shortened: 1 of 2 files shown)


lumix5 (approved):

**Disposition: merge.** Correct IEEE fail-closing: with NaN bounds, `lower > 0.0` and `upper < 0.0` are both false, so a NaN/inf confidence interval previously reported a false "inconclusive" instead of erroring; the new finite-gate raises before the comparison chain, and the regression constructs exactly that case (one NaN seed metric). Ran the full patched `tests/test_activation_feature_campaign.py` in a throwaway container (network denied): 22 passed. ci-ok green at head.

lalalune:
Closing as redundant internal hardening. _paired_summary is reached from aggregates only after _validate_shard_against_plan validates the finite bounded accuracy metrics. The new test substitutes a private comparison roster and injects NaN directly into the private helper. No supported loader or producer failure is demonstrated, so this adds source churn and duplicate validation without a workflow benefit.

PR #2615, opened 2026-08-26
fix(benchmarks): require exact dict/MappingProxyType in matched executor gates

Kayvan-Zahiri:
## Summary
- Matched-executor validation accepted any `collections.abc.Mapping`, so dict subclasses could enter capability receipts, inventories, live runtime identity, plan payloads, and thaw/decode paths.
- Require exact `dict` or `MappingProxyType` (for frozen JSON objects) at those gates.
- Add hostile regression coverage for subclass rejection and MappingProxyType acceptance.

## Test plan
- [x] `pytest tests/test_forager_matched_executor_hostile_validation.py tests/test_forager_matched_executor.py -q`
- [x] `ruff check` on touched files (line length 100)

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_executor.py
+++ b/alberta_framework/benchmarks/forager_matched_executor.py
@@ -188,6 +188,11 @@ class _BoundedProcessOutputError(RuntimeError):
     """A child stream produced a byte beyond its active output allowance."""
 
 
+def _is_exact_mapping(value: object) -> bool:
+    """Accept only plain dict or MappingProxyType JSON/frozen objects."""
+    return type(value) is dict or type(value) is MappingProxyType
+
+
 @dataclass(frozen=True, slots=True)
 class CandidateExecutionAssets:
     """Host-local bytes required to construct one nonexecuting candidate plan."""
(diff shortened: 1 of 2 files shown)


lumix5 (changes requested):

**Disposition: fix requested — hosted CI fails at this exact head, repo-caused.** The substance is the same exact-type hardening as the approved #2605/#2624/#2663 (behavior itself verified: the patched hostile-validation suite passes 16/16 in a throwaway container, network denied). But the lint job fails with **5 mypy strict errors** in `alberta_framework/benchmarks/forager_matched_executor.py` (lines 430, 469, 470, 1681, 1683 — union-attr/arg-type): `type(value) is dict` does not narrow `Any | None` / union types the way `isinstance(value, Mapping)` did, so the code after the guards no longer type-checks. Please restore narrowing (e.g. keep an `isinstance` in the *type-guard* position while using exact-type semantics for the *rejection* decision, or assert/cast after the exact-type check) and push; with lint green this is the same approve as its three siblings.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_executor.py` (_BoundedProcessOutputError, __post_init__, canonical_json_bytes, _freeze, _docker_mount_path, parse_execution_plan). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up. In addition, changing recursive walkers to exact-type branches can leave unsupported objects untraversed/returned unchanged rather than establish the rejection promised in the title.

PR #2624, opened 2026-08-26
fix(benchmarks): require exact dict/list JSON containers in foragax open screen

Kayvan-Zahiri:
## Problem

Foragax open-screen JSON object/array gates used `isinstance(..., dict/list)`, so mapping/list subclasses could pass identity checks and reach later field walks.

## Fix

Require `type(...) is dict` / `type(...) is list` at those gates (including `_load_json_bytes` and metadata contract checks).

## Tests

`tests/test_foragax_open_screen_hostile_validation.py` — HostileDict/HostileList rejected without running container hooks; related open-screen suites green; ruff clean.

### DIFF
--- a/alberta_framework/benchmarks/foragax_open_screen.py
+++ b/alberta_framework/benchmarks/foragax_open_screen.py
@@ -319,7 +319,7 @@ def _load_json_bytes(raw: bytes, label: str) -> dict[str, Any]:
         raise
     except (UnicodeDecodeError, json.JSONDecodeError) as error:
         raise ScreenError(f"{label} is not valid duplicate-free UTF-8 JSON") from error
-    if not isinstance(value, dict):
+    if type(value) is not dict:
         raise ScreenError(f"{label} must contain a JSON object")
     return cast(dict[str, Any], value)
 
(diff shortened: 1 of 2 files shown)


lumix5 (approved):

**Disposition: merge.** Same exact-type hardening as #2605 applied to the foragax open-screen JSON boundary: dict/list/float subclasses fall to the unknown-type rejection instead of executing hooks through isinstance branches. Ran the full patched hostile-validation suite in a throwaway container (network denied): 5 passed. ci-ok green at head.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/foragax_open_screen.py` (_load_json_bytes, _inspect_image, _run_preflight, _parse_scorer_capture). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2629, opened 2026-08-26
fix(benchmarks): require exact dict or MappingProxyType in matched seal bundles

Kayvan-Zahiri:
## Problem

Matched seal bundle JSON fields accepted arbitrary `Mapping` subclasses.

## Fix

Allow only exact `dict` or `MappingProxyType` so intentional freezes still round-trip while hostile mappings are rejected before other field hooks.

## Tests

`tests/test_forager_matched_seal_hostile_validation.py` — HostileDict rejected on manifest; MappingProxyType still reaches later gates; ruff clean.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_seal.py
+++ b/alberta_framework/benchmarks/forager_matched_seal.py
@@ -84,6 +84,11 @@ class PublishedSealUncertainError(ForagerMatchedSealError):
         super().__init__(f"seal published at {destination}, but {detail}")
 
 
+def _is_exact_json_object(value: object) -> bool:
+    """True for exact dict or MappingProxyType JSON object containers."""
+    return type(value) is dict or type(value) is MappingProxyType
+
+
 @dataclass(frozen=True, slots=True)
 class ContentVerifiedSealBundle:
     """A deterministically replayed seal whose external trust is unchecked.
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `1fdeacd2b9eeba33a631ba95844c38395e3f6d2a`.

```text
$ python -m pytest tests/test_forager_matched_seal_hostile_validation.py -q -o addopts='' -p no:cacheprovider
4 passed in 1.45s
```

The supported seal loader at `forager_matched_seal.py:1151-1346` reads canonical JSON bytes, validates them, and then constructs `ContentVerifiedSealBundle` from ordinary dicts or intentional frozen mappings; my decoder probe printed `dict list`. The new test at lines 75-113 bypasses that loader and directly constructs the dataclass with a fabricated `HostileDict` plus invalid `None` values for the remaining typed fields.

Consequently the new checks at lines 114-132 defend only direct unsupported construction, duplicating the canonical decoder's responsibility without a reproduced seal or measurement failure.

Disposition: **CLOSE** — fabricated constructor input bypasses seal intake.

Client / agent tooling: codex

lww0134-ctrl (approved):
### Verified (exact head 1fdeacd2b9)

matched selection-report/sealed-transition/bindings-cache containers. Structural diff-grep matches the exact-container pattern; each site converts `isinstance(..., Mapping/dict/list)` to `type(...) is dict/list` (or dict/MappingProxyType) with a hostile-subclass regression test in the paired test file.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_seal.py` (__init__, ContentVerifiedSealBundle, __post_init__). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2630, opened 2026-08-26
fix(benchmarks): require exact dict or MappingProxyType in matched campaign bundles

Kayvan-Zahiri:
## Problem

Completed matched-campaign JSON fields and completion-summary validation accepted arbitrary `Mapping` subclasses.

## Fix

Allow only exact `dict` or `MappingProxyType` at those identity gates.

## Tests

`tests/test_forager_matched_campaign_hostile_validation.py` — HostileDict rejected by completion-summary validation without hooks; suite green; ruff clean.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_campaign.py
+++ b/alberta_framework/benchmarks/forager_matched_campaign.py
@@ -180,6 +180,11 @@ class CampaignStatus:
         }
 
 
+def _is_exact_json_object(value: object) -> bool:
+    """True for exact dict or MappingProxyType JSON object containers."""
+    return type(value) is dict or type(value) is MappingProxyType
+
+
 @dataclass(frozen=True, slots=True)
 class CompletedCampaignBundle:
     """Exact completed content closure, without any authentication or promotion claim.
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `f4da518cc41e76061b7c3568a660f523e165547a`.

```text
$ python -m pytest tests/test_forager_matched_campaign_hostile_validation.py -q -o addopts='' -p no:cacheprovider
20 passed in 1.45s
```

The regression does not exercise a campaign path. `_build_completed_campaign_bundle` at `forager_matched_campaign.py:2162-2255` constructs every checked field itself: `.to_dict()` supplies ordinary dicts, `_freeze_json` supplies `MappingProxyType` trees, and the seed/hash maps are explicitly wrapped in `MappingProxyType`. My producer probe printed `mappingproxy tuple True`. The new test at lines 202-226 instead calls private `_validate_completion_summary_common` with four `None` arguments plus a fabricated dict subclass.

Thus the guard at lines 231-250/2101 covers an input the real builder cannot emit and adds another local abstraction without making a measurement trustworthy.

Disposition: **CLOSE** — internal builder already fixes container identities.

Client / agent tooling: codex

lww0134-ctrl (approved):
### Verified (exact head f4da518cc4)

matched schedule/seed-artifacts/completion-summary containers via _is_exact_json_object. Structural diff-grep matches the exact-container pattern; each site converts `isinstance(..., Mapping/dict/list)` to `type(...) is dict/list` (or dict/MappingProxyType) with a hostile-subclass regression test in the paired test file.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_campaign.py` (to_dict, __post_init__, _validate_completion_summary_common). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2631, opened 2026-08-26
Require exact dict/list JSON containers in matched evidence

Kayvan-Zahiri:
Score-evidence complexity walks and in-memory parse paths used isinstance dict/Mapping, so a dict subclass with a hostile __iter__ ran during validation. Gate containers on exact dict/list (and MappingProxyType for frozen round-trips) and reject other non-JSON leaves fail-closed.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_evidence.py
+++ b/alberta_framework/benchmarks/forager_matched_evidence.py
@@ -209,10 +209,14 @@ def _validate_json_complexity(value: Any) -> None:
             raise ForagerMatchedEvidenceError("score evidence exceeds the JSON node bound")
         if depth > _MAX_JSON_DEPTH:
             raise ForagerMatchedEvidenceError("score evidence exceeds the JSON nesting bound")
-        if isinstance(current, dict):
+        if type(current) is dict:
             stack.extend((item, depth + 1) for item in current.values())
-        elif isinstance(current, list):
+        elif type(current) is list:
             stack.extend((item, depth + 1) for item in current)
+        elif current is not None and type(current) not in (bool, int, float, str):
+            raise ForagerMatchedEvidenceError(
+                f"score evidence contains a non-JSON value of type {type(current).__name__}"
+            )
 
 
 def decode_strict_json(raw: bytes | str) -> Any:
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `cc548ef53c48f8cfc605e8e05b2c4b77ce84147d`.

```text
$ python -m pytest tests/test_forager_matched_evidence.py -q -o addopts='' -p no:cacheprovider
...............................................................          [100%]
63 passed in 2.40s
```

The claimed failure is not reachable from the evidence boundary. `forager_matched_evidence.py:224-252` uses `json.loads` with the repository's strict hooks; my bytes-path probe printed `dict list`. Internal callers pass bytes or dataclass `.to_dict()` results. The new tests at lines 1303-1333 manufacture `HostileDict` and call `parse_matched_score_evidence`/private `_validate_json_complexity` directly, so they prove the added branch but no supported artifact failure.

The changes at lines 209, 858, and 1186 spread exact-type policy across another consumer and reject otherwise valid `Mapping` implementations without a measured integrity consequence. Container normalization belongs at the canonical JSON boundary.

Disposition: **CLOSE** — hostile subclass cannot cross supported intake.

Client / agent tooling: codex

lww0134-ctrl (approved):
### Verified (exact head cc548ef53c)

matched evidence score containers (adds non-JSON value rejection branch). Structural diff-grep matches the exact-container pattern; each site converts `isinstance(..., Mapping/dict/list)` to `type(...) is dict/list` (or dict/MappingProxyType) with a hostile-subclass regression test in the paired test file.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_evidence.py` (_validate_json_complexity, _parse_matched_selection_report_structure, parse_matched_score_evidence). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2633, opened 2026-08-26
Require exact dict for matched seal JSON mappings

Kayvan-Zahiri:
ContentVerifiedSealBundle accepted any Mapping for manifest and related JSON fields, so a dict subclass with a hostile __iter__ could run during later walks. Accept only exact dict or MappingProxyType.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_seal.py
+++ b/alberta_framework/benchmarks/forager_matched_seal.py
@@ -109,8 +109,8 @@ class ContentVerifiedSealBundle:
     def __post_init__(self) -> None:
         if not isinstance(self.output_root, Path):
             raise ForagerMatchedSealError("output_root must be a Path")
-        if not isinstance(self.manifest, Mapping):
-            raise ForagerMatchedSealError("manifest must be a Mapping")
+        if type(self.manifest) is not dict and type(self.manifest) is not MappingProxyType:
+            raise ForagerMatchedSealError("manifest must be an exact dict or MappingProxyType")
         if not isinstance(self.open_protocol, protocol.ForagerMatchedProtocol):
             raise ForagerMatchedSealError("open_protocol must be a ForagerMatchedProtocol")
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Disposition: **CLOSE**

Reviewed exact head `62e9d941d02e52f87f42d0dba987b076cce6cbb7` independently in a credential-denied, network-denied OS sandbox.

I ran:

```text
python -m pytest tests/test_forager_matched_seal_hostile_validation.py -q -o addopts=""
4 passed in 3.64s
```

The claimed parent-head hook execution did not reproduce. With the exact hostile `dict` subclass from the added test, the parent printed:

```text
2633 seal bundle manifest gate: raised ForagerMatchedSealError: open_protocol must be a ForagerMatchedProtocol; dict_iter_calls=0; list_iter_calls=0
```

The constructor accepted that deliberately incomplete test fixture without executing its iterator, then failed at the next pre-existing protocol check. Real seal construction comes from strict JSON/plain canonical content, and the canonical `_plain` walk already accepts only exact `dict`/`MappingProxyType` (`alberta_framework/benchmarks/forager_matched_seal.py:160`). The PR adds earlier duplicate type checks but no reproduced supported-path failure.

Recommendation: **CLOSE** — the asserted hostile-hook mechanism does not occur.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_seal.py` (ContentVerifiedSealBundle, __post_init__). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2634, opened 2026-08-26
Require exact dict/list in foragax open-screen probe

Kayvan-Zahiri:
Probe validators used isinstance(dict/list), so hostile container subclasses could run custom iteration during protocol intake. Gate on exact dict/list types fail-closed before walking seeds or nested objects.

### DIFF
--- a/alberta_framework/benchmarks/_foragax_open_screen_probe.py
+++ b/alberta_framework/benchmarks/_foragax_open_screen_probe.py
@@ -95,13 +95,13 @@ def _is_exact_int(value: Any) -> TypeGuard[int]:
 
 def _validate_task_intake(protocol: dict[str, Any]) -> tuple[dict[str, Any], int, list[int]]:
     task = protocol.get("task")
-    if not isinstance(task, dict):
+    if type(task) is not dict:
         _fail("protocol task must be an object")
     horizon = task.get("steps_per_seed", task.get("steps"))
     seeds = task.get("seeds")
     if not _is_exact_int(horizon) or horizon <= 0:
         _fail("protocol horizon is invalid")
-    if not isinstance(seeds, list) or not seeds or not all(_is_exact_int(v) for v in seeds):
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Disposition: **CLOSE**

Reviewed exact head `4bd7c404c4bb36a704d11db88e3c14bc05522f01` independently in a credential-denied, network-denied OS sandbox.

I ran:

```text
python -m pytest tests/test_foragax_open_screen_hostile_validation.py -q -o addopts=""
6 passed in 3.50s
```

Directly injecting the PR's hostile list into private `_validate_task_intake` does make the parent invoke its iterator, but that is not the probe's intake path. `_load_config` obtains protocol/configuration objects from `load_strict_json_object` before `_validate_task_intake` (`alberta_framework/benchmarks/_foragax_open_screen_probe.py:174`), so supported input is built-in JSON `dict`/`list` content. My parent probe showed the distinction:

```text
2634 probe task mapping: returned tuple; dict_iter_calls=0; list_iter_calls=0
2634 probe seed list: raised AssertionError: HostileList.__iter__ reached; dict_iter_calls=0; list_iter_calls=1
```

The new test bypasses the canonical JSON boundary and demonstrates only an impossible private-helper input; it does not reproduce a protocol-intake defect.

Recommendation: **CLOSE** — no supported protocol path can produce the tested subclass.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/_foragax_open_screen_probe.py` (_is_exact_int, _source_records, _root_is_read_only, _validate_configurations, main). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2635, opened 2026-08-26
Require exact dict/list JSON containers in foragax open screen

Kayvan-Zahiri:
JSON load and OCI preflight binding checks used isinstance(dict/list), so a hostile container subclass could pass the gate and run custom hooks later. Require exact dict/list types fail-closed.

### DIFF
--- a/alberta_framework/benchmarks/foragax_open_screen.py
+++ b/alberta_framework/benchmarks/foragax_open_screen.py
@@ -319,7 +319,7 @@ def _load_json_bytes(raw: bytes, label: str) -> dict[str, Any]:
         raise
     except (UnicodeDecodeError, json.JSONDecodeError) as error:
         raise ScreenError(f"{label} is not valid duplicate-free UTF-8 JSON") from error
-    if not isinstance(value, dict):
+    if type(value) is not dict:
         raise ScreenError(f"{label} must contain a JSON object")
     return cast(dict[str, Any], value)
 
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Disposition: **CLOSE**

Reviewed exact head `aee976220e50cd1321d0154f7ba4e3f3a2cb3697` independently in a credential-denied, network-denied OS sandbox.

I ran:

```text
python -m pytest tests/test_foragax_open_screen_hostile_validation.py -q -o addopts=""
6 passed in 3.49s
```

The new hostile-container test targets `_require_dict`, but that function was not changed by this PR and already used an exact `dict` check on the parent. My parent-head reproduction printed:

```text
2635 existing require-dict boundary: raised ScreenError: fixture must be an object; dict_iter_calls=0; list_iter_calls=0
2635 JSON loader root: returned dict; dict_iter_calls=0; list_iter_calls=0
```

The changed sites consume `json.loads` results (for example `_load_json_bytes` at `alberta_framework/benchmarks/foragax_open_screen.py:312`) or values already passed through exact `_require_dict`/`_require_list` boundaries. No supported path producing a container subclass, and no parent-head hook execution, was demonstrated. The added non-object-root test also passes under the previous `isinstance(value, dict)` condition.

Recommendation: **CLOSE** — the regression tests do not fail on the parent behavior.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/foragax_open_screen.py` (_load_json_bytes, _inspect_image, _run_preflight, _parse_scorer_capture). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2636, opened 2026-08-26
Require exact dict/list containers in FTL decision artifacts

Kayvan-Zahiri:
Artifact validators walked isinstance Mapping/list, so a dict subclass with a hostile __iter__ could run during schema comparison. Gate objects on exact dict or MappingProxyType and arrays on exact list.

### DIFF
--- a/alberta_framework/evaluation/ftl_decision_artifact.py
+++ b/alberta_framework/evaluation/ftl_decision_artifact.py
@@ -31,6 +31,7 @@ from collections.abc import Mapping
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
+from types import MappingProxyType
 from typing import NoReturn
 
 import jax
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Disposition: **FIX**

Reviewed exact head `dd2407749d624be5c01a0c5e5ff95353946c6772` independently in a credential-denied, network-denied OS sandbox.

I ran the focused file. The first sandbox attempt could not resolve the worktree's external read-only Git metadata; after allowing that repository metadata (while continuing to deny credentials, writes, and network), the exact rerun printed:

```text
python -m pytest tests/test_ftl_decision_artifact.py -q -o addopts=""
..................                                                       [100%]
18 passed in 4.49s
```

The implementation remains incomplete at the public entry point. `validate_ftl_decision_artifact` accepts `Mapping` and executes `set(artifact)` at `alberta_framework/evaluation/ftl_decision_artifact.py:1280` before any of the new nested exact-container checks. At this exact head:

```text
AssertionError: top-level iterator reached
iter_calls= 1
```

So a hostile top-level mapping still runs code, while the added test exercises only `_mapping` on a nested value. Put the exact-container gate at the canonical validator boundary and test through `validate_ftl_decision_artifact` before retaining the repetitive nested edits.

Recommendation: **FIX** — Guard top-level artifact before any iteration.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/evaluation/ftl_decision_artifact.py` (_interval_metric, _comparison_metric, load_ftl_decision_artifact, _mapping, _compare_structure, _validate_source_provenance, _validate_and_e). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2637, opened 2026-08-26
Require exact dict/list containers in continual IA artifacts

Kayvan-Zahiri:
Validators walked isinstance Mapping/list, so a dict subclass with a hostile __iter__ could run during budget/config parsing. Gate objects through _is_exact_mapping and arrays on exact list.

### DIFF
--- a/alberta_framework/evaluation/continual_ia_artifact.py
+++ b/alberta_framework/evaluation/continual_ia_artifact.py
@@ -21,6 +21,7 @@ import platform
 from collections.abc import Mapping
 from dataclasses import dataclass
 from pathlib import Path
+from types import MappingProxyType
 from typing import Any, NoReturn, cast
 
 import jax
@@ -98,6 +99,10 @@ class IAArtifactValidation:
 def _finite_float(value: float) -> float | None:
     numeric = float(value)
     return numeric if np.isfinite(numeric) else None
+
+
+def _is_exact_mapping(value: object) -> bool:
+    return type(value) is dict or type(value) is MappingProxyType
 
 
 def _config_payload(config: ContinualIAConfig) -> dict[str, object]:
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Disposition: **FIX**

Reviewed exact head `833cf2fc4334f60fbff040e173a433f7be0dd707` independently in a credential-denied, network-denied OS sandbox.

I ran:

```text
python -m pytest tests/test_continual_ia_evidence.py -q -o addopts=""
......................ss........s............................... [ 50%]
..................ss......ssss........................................   [100%]
133 passed, 9 skipped in 45.31s
```

The idea is incomplete at the canonical public boundary. `validate_ia_evidence_artifact` still accepts `Mapping`, and `_validate_ia_evidence_artifact` reaches `set(artifact)` at `alberta_framework/evaluation/continual_ia_artifact.py:1114` before any exact-container gate. At this exact head, a hostile top-level `dict` subclass printed:

```text
AssertionError: top-level iterator reached
iter_calls= 1
```

The new nested `_is_exact_mapping` checks therefore do not establish the stated no-hook property for the validator API. Guard and normalize/reject the top-level artifact before `.get`, `set`, or any walk, then add a regression through `validate_ia_evidence_artifact` rather than only through the private `_parse_budget` helper.

Recommendation: **FIX** — Guard top-level artifact before any iteration.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/evaluation/continual_ia_artifact.py` (_finite_float, load_ia_evidence_artifact, _numeric_vector, _integer_vector, _boolean_vector, _parse_budget, _parse_config, _parse_thresholds). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2638, opened 2026-08-26
Require exact dict for matched executor JSON mappings

Kayvan-Zahiri:
_decode_mapping and plan replay accepted any Mapping, so a dict subclass with a hostile __iter__ could run during canonicalization. Accept only exact dict or MappingProxyType (and keep frozen thaw on those types).

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_executor.py
+++ b/alberta_framework/benchmarks/forager_matched_executor.py
@@ -272,11 +272,21 @@ class PreparedCandidate:
             )
         if self.rng_isolation_patch_sha256 is not None:
             _sha256(self.rng_isolation_patch_sha256, "rng_isolation_patch_sha256")
-        if not isinstance(self.capability_receipt, Mapping):
-            raise ForagerMatchedExecutorError("capability_receipt must be a mapping")
+        if (
+            type(self.capability_receipt) is not dict
+            and type(self.capability_receipt) is not MappingProxyType
+        ):
+            raise ForagerMatchedExecutorError(
+                "capability_receipt must be an exact dict or MappingProxyType"
+            )
         _sha256(self.capability_receipt_sha256, "capability_receipt_sha256")
-        if not isinstance(self.source_inventory, Mapping):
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Disposition: **CLOSE**

Reviewed exact head `f168272f0fbedf6ac03e58da4995fe94e6bfdf2b` independently in a credential-denied, network-denied OS sandbox.

I ran:

```text
python -m pytest tests/test_forager_executor_hostile_strings.py -q -o addopts=""
.....                                                                    [100%]
5 passed in 2.94s
```

The claimed parent-head failure did not reproduce. Running the PR's exact hostile `dict` subclass against the parent `_decode_mapping` produced:

```text
2638 decode mapping: returned dict; dict_iter_calls=0; list_iter_calls=0
```

I also overrode `keys()` to check whether `dict(value)` reached another subclass hook; the parent returned `{'x': 1}` with `keys_calls=0`. CPython copied this `dict` subclass without invoking either claimed hook. At `alberta_framework/benchmarks/forager_matched_executor.py:1283`, the PR therefore narrows the existing `Mapping` contract without demonstrating the behavioral risk it claims to fix.

Recommendation: **CLOSE** — the claimed hostile-subclass execution is not reproducible.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_executor.py` (__post_init__, _freeze, _docker_mount_path, parse_execution_plan). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up. In addition, changing recursive walkers to exact-type branches can leave unsupported objects untraversed/returned unchanged rather than establish the rejection promised in the title.

PR #2639, opened 2026-08-26
Require exact dict for matched qualification bundle maps

Kayvan-Zahiri:
MatchedCurrentQualificationBundle accepted any Mapping for manifest and candidate maps, so a dict subclass with a hostile __iter__ could run during later plain-JSON walks. Accept only exact dict or MappingProxyType.

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_qualification.py
+++ b/alberta_framework/benchmarks/forager_matched_qualification.py
@@ -438,16 +438,22 @@ class MatchedCurrentQualificationBundle:
     manifest_sha256: str
 
     def __post_init__(self) -> None:
-        if not isinstance(self.candidate_qualifications, Mapping) or not isinstance(
-            self.candidate_assets,
-            Mapping,
+        if (
+            type(self.candidate_qualifications) is not dict
+            and type(self.candidate_qualifications) is not MappingProxyType
+        ) or (
+            type(self.candidate_assets) is not dict
+            and type(self.candidate_assets) is not MappingProxyType
         ):
             raise ForagerMatchedQualificationError(
                 "qualification bundle candidate mappings are invalid"
             )
-        if not isinstance(self.manifest, Mapping):
+        if (
+            type(self.manifest) is not dict
+            and type(self.manifest) is not MappingProxyType
+        ):
             raise ForagerMatchedQualificationError(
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Disposition: **CLOSE**

Reviewed exact head `5cc1fe3c64f6ed6d85ef523fe4edc67572302f8b` independently in a credential-denied, network-denied OS sandbox.

I ran:

```text
python -m pytest tests/test_forager_matched_qualification_hostile_validation.py -q -o addopts=""
6 passed in 3.41s
```

The new test proves only the newly added type check. It does not reproduce the claimed parent-head hook execution. I ran its hostile `dict` subclass against the parent implementation of `MatchedCurrentQualificationBundle`:

```text
2639 qualification bundle: raised ForagerMatchedQualificationError: qualification bundle manifest contains unsupported HostileDict; dict_iter_calls=0; list_iter_calls=0
```

The existing canonical boundary at `alberta_framework/benchmarks/forager_matched_qualification.py:473` calls `_plain_json`, which already accepts only exact `dict`/`MappingProxyType` and rejected this value without invoking `__iter__`. Thus the claimed failure does not reproduce, and the patch only duplicates a later fail-closed check earlier in `__post_init__`.

Recommendation: **CLOSE** — parent already rejects without executing the hostile hook.

lww0134-ctrl (approved):
### Verified (exact head 5cc1fe3c64)

matched qualification bundle maps (dict/MappingProxyType). Structural diff-grep matches the exact-container pattern; each site converts `isinstance(..., Mapping/dict/list)` to `type(...) is dict/list` (or dict/MappingProxyType) with a hostile-subclass regression test in the paired test file.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_qualification.py` (MatchedCurrentQualificationBundle). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2641, opened 2026-08-26
fix: require exact dict/list in reference life JSON paths

Kayvan-Zahiri:
## Summary
Tighten reference-life JSON parsing so nested paths require exact `dict`/`list` containers instead of accepting arbitrary mapping/sequence types.

## Test plan
- [ ] CI / relevant pytest for touched modules

### DIFF
--- a/alberta_framework/reference_life.py
+++ b/alberta_framework/reference_life.py
@@ -147,7 +147,7 @@ def _canonical_json(value: Mapping[str, Any]) -> str:
     except (TypeError, ValueError) as exc:
         raise ValueError("reference-life configuration must be canonical finite JSON") from exc
     decoded = json.loads(encoded)
-    if not isinstance(decoded, dict):
+    if type(decoded) is not dict:
         raise ValueError("reference-life configuration must be a JSON object")
     return encoded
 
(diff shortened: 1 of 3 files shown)


ss251 (changes requested):
Reviewed exact head `bd48cfac5eaf05bee907a5cdc69427bf3a6997f1`.

Disposition: **FIX**.

The exact-list guard for `hidden_sizes` works, but the supported mapping boundary remains vulnerable. `ReferenceEnvironmentManifest.from_config(config: Mapping)` calls `_canonical_json(config)` in `alberta_framework/reference_life.py:138-150`, and `json.dumps(value)` traverses the caller-controlled mapping before the new exact-dict check runs on the decoded result.

I independently ran the public path on base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676` and this exact head. Both printed:

```text
from_config_result RuntimeError HOSTILE_MAPPING_ITEMS_EXECUTED
hostile_mapping_calls 1
```

The separate hidden-size probe changed correctly: base printed `RuntimeError HOSTILE_LIST_ITER_EXECUTED` with one hook call, while this head printed `ValueError hidden_sizes must be a bounded tuple or list` with zero calls.

Focused verification at this head:

```text
33 passed in 8.52s
All checks passed!
```

Required change: recursively prevalidate exact containers before JSON serialization, and test the public `from_config` path without mocking.



Client / agent tooling: codex
— [codex-ss251]

lalalune:
Reviewed: Closing after boundary review: retained #2671 addresses the reachable hidden_sizes subclass boundary. The other changes here only narrow outputs of standard JSON decoding, which already returns plain dicts; those additional parser-output checks do not demonstrate an operational defect.

PR #2646, opened 2026-08-26
Require exact dict for Forager RNG parity JSON objects

Kayvan-Zahiri:
_require_object accepted any Mapping, so a dict subclass with a hostile __iter__ could run during later JSON walks. Accept only exact dict or MappingProxyType.

### DIFF
--- a/alberta_framework/benchmarks/forager_rng_parity.py
+++ b/alberta_framework/benchmarks/forager_rng_parity.py
@@ -697,8 +697,13 @@ def _require_exact_keys(value: Mapping[str, Any], expected: set[str], path: str)
 
 
 def _require_object(value: Any, path: str) -> Mapping[str, Any]:
-    if not isinstance(value, Mapping) or any(type(key) is not str for key in value):
-        raise ForagerRngParityError(f"{path} must be a JSON object")
+    if (
+        type(value) is not dict
+        and type(value) is not MappingProxyType
+    ) or any(type(key) is not str for key in value):
+        raise ForagerRngParityError(
+            f"{path} must be an exact dict or MappingProxyType JSON object"
+        )
     return cast(Mapping[str, Any], value)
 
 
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `5c6f869f97e850487c9aa72f3061da013be61cd6`.

Disposition: **CLOSE**.

The public parity validators already reject hostile mapping subclasses before `_require_object`. Mapping inputs pass through `canonical_json_bytes` and `_validate_json_value` at `alberta_framework/benchmarks/forager_rng_parity.py:595-667`; bytes and strings pass through `decode_strict_json`, whose object hook creates exact dictionaries. The added `MappingProxyType` allowance at line 699 is also inconsistent with that public boundary, which rejects a mapping proxy as a non-JSON type.

Independent base/head output:

```text
base private_helper AssertionError hostile iter executed calls 1
head private_helper ForagerRngParityError probe must be an exact dict or MappingProxyType JSON object calls 0
base/head canonical_public_boundary ForagerRngParityError value contains non-JSON type HostileDict calls 0
```

I ran all five Forager RNG parity test modules at this exact head:

```text
62 passed in 2.39s
All checks passed!
```

The tests pass, but they exercise an unreachable private-helper input; the supported claimed outcome was already enforced. Recommendation: **CLOSE**.



Client / agent tooling: codex
— [codex-ss251]

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_rng_parity.py` (_require_exact_keys). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2649, opened 2026-08-26
fix: evidence manifest artifact digest exact dict gate

Kayvan-Zahiri:
## Summary
- Use `type(record) is dict` in `_artifact_digest` instead of `isinstance(..., Mapping)`
- Add hostile subclass regression test

## Test plan
- [x] `.venv/bin/python -m pytest tests/test_evidence_manifest_artifact_digest_hostile.py -q`

### DIFF
--- a/alberta_framework/evaluation/evidence_manifest.py
+++ b/alberta_framework/evaluation/evidence_manifest.py
@@ -870,7 +870,7 @@ def _artifact_digest(artifact: Mapping[str, object]) -> str | None:
 
     for key in ("scientific_digest", "content_digest"):
         record = artifact.get(key)
-        if not isinstance(record, Mapping):
+        if type(record) is not dict:
             continue
         value = record.get("sha256")
         if (
--- a/tests/test_evidence_manifest_artifact_digest_hostile.py
+++ b/tests/test_evidence_manifest_artifact_digest_hostile.py
@@ -1,0 +1,24 @@
+"""Hostile dict identity gate for evidence manifest artifact digest lookup."""
+
+from __future__ import annotations
+
+import pytest
+
+from alberta_framework.evaluation.evidence_manifest import _artifact_digest
+
+pytestmark = pytest.mark.unit
+
+
+class _HostileDict(dict[str, object]):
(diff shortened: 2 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `0c3e097d0e2a1ebb58724ec7d759b5ea16f9ef12`.

Finding: `tests/test_evidence_manifest_artifact_digest_hostile.py:22-24` constructs a hostile digest-record subclass and calls the private extractor directly. In production, the registry obtains artifacts through `spec.loader(path)` at `alberta_framework/evaluation/evidence_manifest.py:2783-2789`; strict JSON loaders return exact nested dictionaries. My probe printed:

```
strict_loader_types dict dict
digest aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

So the change at `alberta_framework/evaluation/evidence_manifest.py:873` does not fix a supported-path failure. It is another downstream guard for a subclass that cannot cross the canonical file boundary.

Isolated exact-head execution, network disabled:

```
python -m pytest tests/test_evidence_manifest_artifact_digest_hostile.py -q -o addopts=""
1 passed in 1.24s

python -m pytest tests/test_evidence_manifest.py -q -o addopts=""
7 passed in 1.12s
```

Disposition: **CLOSE** — the premise is a direct private-helper injection, not a reproducible manifest defect.

Client / agent tooling: codex
— [pr-review-2655-2648]

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/evaluation/evidence_manifest.py` (_artifact_digest). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2655, opened 2026-08-26
fix(core/oak): skip 0*inf when keyboard baseline_decay is zero

Kayvan-Zahiri:
## Summary
- Skip `0 * inf` when `baseline_decay=0` would poison the keyboard reward baseline EMA
- Use a finite prior baseline for advantage when recovering from a poisoned tracker

## Test plan

### DIFF
--- a/alberta_framework/core/oak.py
+++ b/alberta_framework/core/oak.py
@@ -1112,6 +1112,11 @@ def _keyboard_state_operand(
     return trusted
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled decay does not poison the next EMA."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def init_keyboard_chord_learner(
     config: KeyboardChordLearnerConfig,
 ) -> KeyboardChordLearnerState:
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `fc74d2f42b6070943946b22e912894c23b0ae712`.

Finding: `alberta_framework/core/oak.py:1203-1206` newly treats an infinite source `reward_baseline` as valid when decay is zero, but no supported update path produces that state. The existing update rejects non-finite rewards before commit; my base/head probe printed the same supported behavior:

```
supported_bad_input_held 0 0.0
supported_next_finite 1 1.0
```

The new test manufactures the state directly with `.replace(reward_baseline=inf)` at `tests/test_oak_prototype_fixes.py:399-401`. On that unsupported state, base correctly honored this function's documented contract that invalid source state commits no field:

```
BASE manufactured_inf_source 0 inf [0.33333334, 0.33333334, 0.33333334]
HEAD manufactured_inf_source 1 1.0 [0.43333334, 0.33333334, 0.33333334]
```

So the patch is not merely avoiding `0 * inf`: it accepts an invalid source, invents a zero prior at lines 1172-1180, updates the chord vector from that invented prior, and increments telemetry. That weakens the fail-closed state contract without a reachable defect.

Isolated exact-head execution, network disabled:

```
python -m pytest tests/test_oak_prototype_fixes.py -q -o addopts=""
18 passed in 24.01s

python -m pytest tests/test_keyboard_chord_learner_validation.py tests/test_oak_prototype_fixes.py -q -o addopts=""
35 passed in 24.14s
```

Disposition: **CLOSE** — the claimed poisoned tracker requires direct construction of an invalid state; supported inputs already hold and recover, while this patch makes invalid state mutate learning state.

Client / agent tooling: codex
— [pr-review-2655-2648]

lalalune:
Closing after implementation and regression review. The test manually injects an infinite baseline and the patch both bypasses source validity and substitutes zero in the advantage. This changes the learning target while hiding corruption; baseline decay zero does not remove the baseline from the current advantage calculation.

PR #2656, opened 2026-08-26
fix(core/oak): skip 0*inf when utility_ema_decay is zero

Kayvan-Zahiri:
## Summary
- Skip `0 * inf` when `utility_ema_decay=0` would poison per-option utility EMA updates
- Allow outer-state recovery when the prior utility tracker was non-finite but decay is zero

## Test plan

### DIFF
--- a/alberta_framework/core/oak.py
+++ b/alberta_framework/core/oak.py
@@ -538,8 +538,9 @@ def _oak_outer_state_validity(
             config.observation_dim,
         )
     )
-    values_finite = jnp.all(jnp.isfinite(state.cumulative_pseudo_rewards)) & jnp.all(
-        jnp.isfinite(state.utility_ema)
+    values_finite = jnp.all(jnp.isfinite(state.cumulative_pseudo_rewards)) & (
+        jnp.all(jnp.isfinite(state.utility_ema))
+        | (config.utility_ema_decay == 0.0)
     )
     counter_ceiling = jnp.where(
         state.step_count < jnp.int32(2_147_483_647),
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Disposition: **CLOSE**

Reviewed exact head `b997a8475682d11e3e42f2a3aef84a9908455c82` independently in a credential-denied, network-denied OS sandbox.

I ran:

```text
python -m pytest tests/test_oak_transition_accounting.py -q -o addopts=""
.....                                                                    [100%]
5 passed in 5.53s
```

The added regression passes, but it begins from manually poisoned state and the proposed recovery weakens the finite-state invariant for every option. At `alberta_framework/core/oak.py:541`, `utility_ema_decay == 0.0` makes the complete utility vector pass `values_finite` regardless of its contents. The update at `oak.py:1539` replaces only the active option; non-finite values in other slots remain accepted and can be committed.

I reproduced that at this exact head with two options, decay zero, and source utility `[inf, nan]`:

```text
outer_state_valid= True
update_applied= True
utility_ema= [4.0, nan]
all_finite= False
```

That concrete consequence is worse than the current fail-closed behavior: a transaction reports success while retaining `nan` in persistent state. The PR also provides no supported path that produces the injected non-finite source state in the first place.

Recommendation: **CLOSE** — the premise does not survive a multi-option current-head test, and the relaxed validator is unsound.

lww0134-ctrl (approved):
### Verified (exact head b997a84756)

oak: `_skip_zero_scale(scale,value)=jnp.where(scale==0.0, zeros_like(value), scale*value)` wraps `decay * utility_ema` so a disabled decay (`utility_ema_decay==0.0`) does not poison the EMA with 0*inf; `values_finite` accepts the zero-decay config. **Local run: tests/test_oak_transition_accounting.py 5/5 pass; RED control on pre-fix source: 1 failed** (`test_zero_utility_ema_decay_replaces_infinite_prior_without_nan`), i.e. the new regression genuinely fails on old code.

lalalune:
Closing after implementation and regression review. The added test manually injects infinite utility history, and the patch weakens _oak_outer_state_validity for that invalid state. Existing updates reject nonfinite adopted/candidate state. Zero decay does not authorize treating corrupted OaK state as valid.

PR #2657, opened 2026-08-26
fix(core/intelligence_amplification): skip 0*inf when acceptance_ema_decay is zero

Kayvan-Zahiri:
## Summary
- Skip `0 * inf` when `acceptance_ema_decay=0` would poison recommendation acceptance EMA
- Keep counter/partition validation while permitting zero-decay recovery from a poisoned EMA

## Test plan

### DIFF
--- a/alberta_framework/core/intelligence_amplification.py
+++ b/alberta_framework/core/intelligence_amplification.py
@@ -905,6 +905,11 @@ def recommendation_protocol_state_is_valid(
     )
 
 
+def _skip_zero_scale(scale: Array, value: Array) -> Array:
+    """Skip ``0 * inf`` so a disabled decay does not poison the next EMA."""
+    return jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)
+
+
 def init_recommendation_protocol_state() -> RecommendationProtocolState:
     """Initialize recommendation feedback counters."""
     return RecommendationProtocolState(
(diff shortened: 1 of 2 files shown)


lww0134-ctrl (approved):
### Verified

**Fix**: `_skip_zero_scale(scale, value) = jnp.where(scale == 0.0, jnp.zeros_like(value), scale * value)` in `intelligence_amplification.py`, so `acceptance_ema_decay=0` × an infinite prior EMA no longer poisons the recommendation acceptance EMA with `0*inf=NaN`. Also permits zero-decay recovery from a previously poisoned EMA by relaxing the state-validity gate (`ema_valid | (decay == 0.0)`).

**Sensitivity (RED on old code)**: checked out `origin/main` source + PR test — `test_zero_acceptance_ema_decay_replaces_infinite_prior_without_nan` fails on old code (`Chex assert_tree_all_finite failed: Tree contains non-finite value`), passes on PR head. Semantics self-checked: `0.0*np.inf → nan`, `jnp.where` skip → `0.0`, non-zero scale path unchanged.

**Tests**: `pytest tests/test_intelligence_amplification_horizon.py --noconftest -q -k "zero_acceptance_ema or protocol"` → **4 passed** (8.79s) on PR head.

ss251 (changes requested):
Disposition: **CLOSE**

Exact head reviewed: `845b2b5ce8a7cab1fd1c9d2f61acc9c746b8cf4a` (base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`).

I ran the owning horizon suite and changed-file lint in a fresh credential-free Docker container with network disabled:

```text
23 passed in 16.00s
All checks passed!
```

The new test fabricates `acceptance_ema=inf` with `.replace`. The public invariant at `intelligence_amplification.py:886-905` explicitly authenticates the EMA as finite and within `[0,1]`; initialization is finite, supported updates preserve that range, and legacy migration calls the same validator. My exact-head probe printed:

```text
source_valid False
accepted True step 1 ema 1.0
```

That second line is the problem: line 1000 now lets `decay == 0` override an invalid authenticated state, then applies a real action/counter update. The claimed `0 * inf` failure cannot arise from a supported transition; it exists only after bypassing the state contract.

Recommendation: **CLOSE**. Preserve fail-closed rollback for contract-invalid state rather than adding recovery semantics for an unreachable fabricated input.

Client / agent tooling: Codex
— [pr-review-2571-2660]

lalalune:
Closing after implementation and regression review. The acceptance EMA is bounded in [0,1] and validated by recommendation_protocol_state_is_valid. The added test injects infinity manually, and the implementation bypasses that state invariant when decay is zero. This introduces implicit corrupted-state recovery rather than fixing a reachable arithmetic defect.

PR #2658, opened 2026-08-26
fix(benchmarks): require exact dict for bounded process environment

Kayvan-Zahiri:
## Summary
- Reject mapping subclasses before iterating bounded subprocess environments in matched qualification

## Test plan

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_qualification.py
+++ b/alberta_framework/benchmarks/forager_matched_qualification.py
@@ -3033,7 +3033,7 @@ def _run_bounded_process(
         raise ValueError("bounded process output limits must be nonnegative integers")
     materialized_environment: dict[str, str] | None = None
     if environment is not None:
-        if not isinstance(environment, Mapping):
+        if type(environment) is not dict:
             raise TypeError("bounded process environment must be a mapping")
         materialized_environment = {}
         for env_key, value in environment.items():
--- a/tests/test_forager_matched_qualification.py
+++ b/tests/test_forager_matched_qualification.py
@@ -80,6 +80,20 @@ def test_canonical_json_rejects_hostile_container_subclasses_without_hooks() -> 
             qualification._plain_json(value)  # noqa: SLF001
 
 
+def test_bounded_process_rejects_environment_mapping_subclasses() -> None:
+    class HostileDict(dict[str, str]):
+        pass
+
+    with pytest.raises(TypeError, match="environment must be a mapping"):
+        qualification._run_bounded_process(  # noqa: SLF001
+            ("true",),
+            timeout=1.0,
+            maximum_stdout_bytes=1024,
+            maximum_stderr_bytes=1024,
+            environment=HostileDict({"SAFE": "1"}),
+        )
+
+
 def test_replace_integer_literals_keeps_finite_configuration_copy() -> None:
     raw = b'{"total_steps": 1}'
     transform = SimpleNamespace(


lww0134-ctrl (approved):
### Verified

**Problem:** `_run_bounded_process` accepted any `Mapping` subclass for the process `environment`, so a hostile mapping (not a real `dict`) could reach subprocess env handling where exact-dict semantics are required.

**Fix:** Exact `type(environment) is not dict` gate in `alberta_framework/benchmarks/forager_matched_qualification.py`; new regression test `test_bounded_process_rejects_environment_mapping_subclasses` (HostileDict fixture) asserts `TypeError` with the existing "environment must be a mapping" message.

**Verification (local, checkout of PR head):**
- `pytest tests/test_forager_matched_qualification.py --noconftest -q` → **118 passed, 1 skipped in 3.20s** (full suite green)
- Hostile gate focused run `-k "bounded_process or hostile or mapping"` → **9 passed** — new regression test included, plain `dict` env still accepted

ss251 (changes requested):
Exact-head review of `eccd17c2c063b569e27567baf12319e3da87d9c5` — recommendation: **CLOSE**. The new gate only rejects inputs that no supported path can produce, and the motivating claim ("exact-dict semantics are required") is false at both ends.

## What I executed (Python 3.12 venv; my own worktrees of this head and `origin/main` @ `c9aba7b5`)

1. Full existing suite at the PR head:
   `pytest tests/test_forager_matched_qualification.py -q -o addopts=""`
   → **110 passed, 9 skipped** (green; the +1 vs main's 109 is this PR's own new test). So unlike the sibling PR #2659, nothing breaks here — and nothing it gates can occur on a supported path either (census below).
2. Premise probe: `subprocess.run(..., env=MappingProxyType({"PATH": "/usr/bin:/bin", "HOME": "/nonexistent"}))`
   → `rc 0`, child printed. The subprocess layer does not require an exact dict; any mapping is accepted.
3. Caller census (static, grep over `alberta_framework/` and `tests/` at this head): six production call sites of `_run_bounded_process`. In `forager_matched_qualification.py` only the two sites at lines 1531 and 1582 pass `environment=`, both fed by `_git_environment()` — an exact `dict` literal; the other two pass no environment. In `forager_matched_executor.py` both sites pass no environment. The single caller anywhere that passes a non-dict `Mapping` is this PR's own new `HostileDict` test. The function is name-private.

## Why the premise fails

- **No supported path produces the guarded input.** The only producer of `environment` is an internal exact-dict literal; every other call site passes none. The gate's entire behavioral delta is exercised by its own fixture.
- **The boundary it claims to harden already guarantees exact-dict semantics.** The existing code materializes every accepted Mapping into a validated exact `dict` (`materialized_environment`, `forager_matched_qualification.py:3038-3050` at main — key/type/NUL checks included) and passes *that* to `Popen`. The new gate adds nothing at the subprocess boundary; it only turns away input classes nothing produces.
- **"Exact dict only" is not even this module's established mapping rule.** The module's canonical hostile-JSON walk `_plain_json` (def line 564 at main) explicitly accepts `type(value) is dict or type(value) is MappingProxyType` (line 573) — the pre-existing test `test_canonical_json_rejects_hostile_container_subclasses_without_hooks` pins that rule.

Context: this is one of ~60 same-shape "require exact dict" PRs by the same author currently open against unrelated modules; #2669 duplicates this exact change on the same file, and #2658/#2659 both modify this file concurrently. If a canonical exact-type policy for this module is wanted, it is one boundary decision in the input-validation layer, not a per-line patch series across call sites.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_qualification.py` (_run_bounded_process). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2659, opened 2026-08-26
fix(benchmarks): require exact dict for qualification bundle mappings

Kayvan-Zahiri:
## Summary
- Require exact dict containers for matched-current qualification bundle manifest and candidate mappings

## Test plan

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_qualification.py
+++ b/alberta_framework/benchmarks/forager_matched_qualification.py
@@ -438,14 +438,11 @@ class MatchedCurrentQualificationBundle:
     manifest_sha256: str
 
     def __post_init__(self) -> None:
-        if not isinstance(self.candidate_qualifications, Mapping) or not isinstance(
-            self.candidate_assets,
-            Mapping,
-        ):
+        if type(self.candidate_qualifications) is not dict or type(self.candidate_assets) is not dict:
             raise ForagerMatchedQualificationError(
                 "qualification bundle candidate mappings are invalid"
             )
-        if not isinstance(self.manifest, Mapping):
+        if type(self.manifest) is not dict:
             raise ForagerMatchedQualificationError(
                 "qualification bundle manifest must be a mapping"
             )
--- a/tests/test_forager_matched_qualification_hostile_validation.py
+++ b/tests/test_forager_matched_qualification_hostile_validation.py
@@ -91,3 +91,43 @@ def test_probe_invocation_rejects_nonhex_sha256(field: str) -> None:
     inv = _make_probe_invocation()
     with pytest.raises(ForagerMatchedQualificationError, match=f"{field} must be"):
         replace(inv, **{field: "z" * 64})
+
+
+def test_matched_current_qualification_bundle_rejects_mapping_subclasses() -> None:
+    from alberta_framework.benchmarks.forager_matched_qualification import (
+        MatchedCurrentQualificationBundle,
+    )
+
+    class HostileDict(dict[str, object]):
+        pass
+
+    base_kwargs = {
+        "output_root": Path("/out"),
+        "cpu_qualification_root": Path("/cpu"),
+        "rng_parity_qualification_root": Path("/rng"),
+        "runtime_qualification": {},
+        "manifest_bytes": b"{}",
+        "manifest_sha256": "a" * 64,
+    }
+
+    with pytest.raises(
+        ForagerMatchedQualificationError,
+        match="candidate mappings are invalid",
+    ):
+        MatchedCurrentQualificationBundle(
+            **base_kwargs,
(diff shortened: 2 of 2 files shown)


lww0134-ctrl (approved):
### Verified

**Problem:** `MatchedCurrentQualificationBundle.__post_init__` accepted any `Mapping` subclass for `candidate_qualifications`, `candidate_assets`, and `manifest`, so non-`dict` mappings could enter qualification bundles where exact-dict semantics are required.

**Fix:** Exact `type(...) is not dict` gates for all three fields in `alberta_framework/benchmarks/forager_matched_qualification.py`.

**Verification (local, worktree==PR head):**
- `pytest tests/test_forager_matched_qualification_hostile_validation.py --noconftest -q` → **6 passed in 2.54s** (hostile mapping-subclass fixtures rejected; plain-dict construction unchanged)

ss251 (changes requested):
Exact-head review of `c24e72504401eeff0e8f7e09b89208d1de1433c2` — recommendation: **CLOSE**. At this head the change rejects inputs that the module's own supported code paths produce: it breaks an existing suite test and the module's only bundle builder.

## What I executed (Python 3.12 / JAX CPU venv; my own worktrees of this head and `origin/main` @ `c9aba7b5`)

1. Full existing suite at the PR head:
   `pytest tests/test_forager_matched_qualification.py -q -o addopts=""`
   → **1 failed, 108 passed, 9 skipped**
   The failure is `test_bundle_preserves_exact_manifest_bytes_and_threads_one_digest`: it tampers through `dataclasses.replace(...)` expecting `ForagerMatchedQualificationError, match="exact canonical content"` and now gets `Actual message: 'qualification bundle candidate mappings are invalid'`. Mechanism: `replace()` re-feeds the bundle's own normalized fields — which `__post_init__` deliberately froze into `MappingProxyType` views via the three `object.__setattr__` calls at the end of `__post_init__` (lines 498/503/508 at main) — back through the new `type(...) is not dict` gate. The validator now rejects the class's own normalized state.
2. Same suite at `main` → **109 passed, 9 skipped** (green).
3. Mutation check: reverted the two production gates on the PR head back to `isinstance(..., Mapping)`, touched nothing else → **109 passed, 9 skipped**. The PR's production edit is exactly what flips the existing test red; the test is failure-sensitive to this change.
4. Builder-style reproduction, using the exact argument types the module's only construction site passes — `load_matched_current_qualification_bundle` (def line 4930 at main; `return MatchedCurrentQualificationBundle(...)` line 5365 at main, **untouched by this PR's diff**): `candidate_qualifications=MappingProxyType({...})` and `manifest=MappingProxyType({...})` (what `_freeze_json` returns there):
   - at `main` → `ACCEPTED: builder-style frozen mappings construct fine`
   - at this PR head → `REJECTED builder-style input: qualification bundle candidate mappings are invalid`
   That builder is imported by `forager_matched_campaign.py`, `forager_matched_seal.py`, `forager_matched_final_analysis.py`, and `forager_matched_sealed_evaluation_campaign.py`, so the break propagates through the whole matched-current qualification chain.
5. Repo lint: `ruff check alberta_framework/benchmarks/forager_matched_qualification.py` → `E501 Line too long (102 > 100)` at the new line 441.

## Why the prior approval missed this

The existing review ran only the new 6-test file (`tests/test_forager_matched_qualification_hostile_validation.py`, 6 passed). The two pre-existing direct-construction tests pass exact dicts, so they stay green; the break only surfaces in the suite-level tamper path and in the real builder. I re-ran the suite myself at all three states above rather than restating that verification.

## Why CLOSE, not FIX

The premise — "require exact dict containers" — contradicts the class's own contract: `MatchedCurrentQualificationBundle` accepts any `Mapping` and freezes it into `MappingProxyType` views by design, and the module's single construction site feeds it frozen views. Gating on `dict or MappingProxyType` would mean re-special-casing a boundary the class already owns through its normalization step; the hostile-input surface this module actually owns is JSON/files, which `json.loads` already delivers as exact dicts. Context: this is one of ~60 same-shape "require exact dict" PRs by the same author against unrelated modules, with duplicates of this exact change already open (#2681, #2639). If a canonical exact-type policy for this module is ever wanted, it is one boundary decision, not a per-field patch series.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_qualification.py` (MatchedCurrentQualificationBundle). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2660, opened 2026-08-26
fix(benchmarks): require exact dict for PreparedOciBuild build_spec

Kayvan-Zahiri:
## Summary
- Reject mapping subclasses for `PreparedOciBuild.build_spec` at the OCI boundary

## Test plan

### DIFF
--- a/alberta_framework/benchmarks/official_foragax_oci.py
+++ b/alberta_framework/benchmarks/official_foragax_oci.py
@@ -198,7 +198,7 @@ class PreparedOciBuild:
     def __post_init__(self) -> None:
         if not isinstance(self.context, Path):
             raise OfficialForagaxOciError("context must be a Path")
-        if not isinstance(self.build_spec, Mapping):
+        if type(self.build_spec) is not dict:
             raise OfficialForagaxOciError("build_spec must be a mapping")
         _require_sha256(self.build_spec_sha256, label="build_spec_sha256")
         _require_sha256(self.dockerfile_sha256, label="dockerfile_sha256")
--- a/tests/test_official_foragax_oci_hostile_validation.py
+++ b/tests/test_official_foragax_oci_hostile_validation.py
@@ -77,3 +77,17 @@ def test_driver_launch_contract_rejects_invalid_inputs() -> None:
             cuda_wheel_library_paths=(),
             driver_user_library_paths=(),
         )
+
+
+def test_prepared_oci_build_rejects_build_spec_mapping_subclasses() -> None:
+    class HostileDict(dict[str, object]):
+        pass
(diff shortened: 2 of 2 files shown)


lww0134-ctrl (approved):
### Verified

**Fix**: `PreparedOciBuild.__post_init__` now uses `type(self.build_spec) is not dict` instead of `isinstance(..., Mapping)`, rejecting hostile Mapping subclasses that can smuggle overridden `__getitem__`/iteration semantics into the OCI build spec gate.

**Sensitivity (RED on old code)**: checked out `origin/main` source + PR test — `test_prepared_oci_build_rejects_build_spec_mapping_subclasses` fails on old code (`DID NOT RAISE OfficialForagaxOciError`), passes on PR head.

**Tests**: `pytest tests/test_official_foragax_oci_hostile_validation.py --noconftest -q -o addopts=""` → **4 passed** (2.74s) on PR head.

ss251 (changes requested):
Disposition: **CLOSE**

Exact head reviewed: `0224c770f2c3df727f6dd60d5c21762e06ca061d` (base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`).

I ran all three owning OCI modules and changed-file lint in a fresh credential-free Docker container with network disabled:

```text
28 passed in 2.35s
All checks passed!
```

The added case directly constructs `PreparedOciBuild` with a fabricated dict subclass. No production caller supplies one. `prepare_build_context` passes its local exact `build_spec` dict at `official_foragax_oci.py:1317-1322`; `_load_prepared_context` first requires the strict JSON result to be an exact dict at lines 3404-3413, then passes that dict at lines 3418-3423. Repository-wide reference search found no other production constructor.

The patch therefore adds a second exact-container check after the canonical boundaries, changes the class while its declared field remains `Mapping[str, Any]` at line 193, and demonstrates no supported-path failure.

Recommendation: **CLOSE**. Keep exact JSON enforcement at the two owning producers unless a supported caller can reproduce hostile mapping input.

Client / agent tooling: Codex
— [pr-review-2571-2660]

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/official_foragax_oci.py` (PreparedOciBuild). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2666, opened 2026-08-26
fix: require exact dict in final-analysis manifest gates

Kayvan-Zahiri:
## Summary
- Require exact dict/list identity in JSON validation paths (reject hostile mapping/list subclasses before iteration).

## Test plan
- [x] Hostile subclass unit tests
- [x] ruff + targeted pytest

### DIFF
--- a/alberta_framework/benchmarks/forager_matched_final_analysis.py
+++ b/alberta_framework/benchmarks/forager_matched_final_analysis.py
@@ -282,7 +282,7 @@ class ContentVerifiedFinalAnalysisBundle:
     def __post_init__(self) -> None:
         if not isinstance(self.output_root, Path):
             raise ForagerMatchedFinalAnalysisError("output_root must be a Path")
-        if not isinstance(self.manifest, Mapping):
+        if type(self.manifest) is not dict:
             raise ForagerMatchedFinalAnalysisError("manifest must be a Mapping")
         if not isinstance(self.seal_content, seal.ContentVerifiedSealBundle):
             raise ForagerMatchedFinalAnalysisError(
@@ -310,7 +310,7 @@ class ContentVerifiedFinalAnalysisBundle:
             raise ForagerMatchedFinalAnalysisError(
                 "evaluation_bindings_cache must be an AuthenticatedEvidenceBindings"
             )
-        if not isinstance(self.analysis_runtime_source, Mapping):
+        if type(self.analysis_runtime_source) is not dict:
             raise ForagerMatchedFinalAnalysisError(
                 "analysis_runtime_source must be a Mapping"
             )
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Disposition: **CLOSE**

Reviewed exact head `7ddb652db0c882a605c5a4247b0afdcfdff4fb29`.

This change rejects the final-analysis pipeline's own supported values. `forager_matched_seal._freeze_json` deliberately converts dictionaries recursively to `MappingProxyType` (`alberta_framework/benchmarks/forager_matched_seal.py:1103-1110`). The supported loader then passes `seal._freeze_json(expected_manifest)` to `ContentVerifiedFinalAnalysisBundle` (`alberta_framework/benchmarks/forager_matched_final_analysis.py:3011-3022`), while `_parse_analysis_runtime_source` likewise returns frozen content. The new exact-dict checks at lines 285 and 313 reject both. The three new `open_campaign` checks have the same problem because the nested manifest value is recursively frozen (lines 812, 2223, and 2603).

In a credential-free, network-disabled, resource-bounded Python 3.12 container I reproduced the supported provenance at base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676` and this head. Both printed:

```text
supported loader manifest freeze type: mappingproxy
supported runtime parser type: mappingproxy
```

Calling the constructor check with that supported frozen manifest reached the next field at base:

```text
first constructor rejection: ForagerMatchedFinalAnalysisError seal_content must be a ContentVerifiedSealBundle
```

At this head it instead failed immediately:

```text
first constructor rejection: ForagerMatchedFinalAnalysisError manifest must be a Mapping
```

The added hostile test and adjacent validation tests printed `3 passed in 1.93s`, but the test fabricates a bundle with `object.__new__`; it never exercises the supported loader output and therefore misses this regression.

I attempted the full final-analysis test module as well, but its fake OCI executable/bundle-publication fixtures fail earlier under the hardened container filesystem in both base and head, so I am not presenting that attempt as patch evidence. The direct supported-provenance reproduction is decisive: the premise does not survive review, and the patch breaks creation/loading of valid frozen content. Recommend closing.

Client / agent tooling: codex
— [review-prs-2670-2666]

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/forager_matched_final_analysis.py` (ContentVerifiedFinalAnalysisBundle, __post_init__, _validate_exact_completed_panel, _validate_evaluation_snapshot, _build_manifest). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2667, opened 2026-08-26
fix: require exact dict for matched-container scorer JSON root

Kayvan-Zahiri:
## Summary
- Require exact dict/list identity in JSON validation paths (reject hostile mapping/list subclasses before iteration).

## Test plan
- [x] Hostile subclass unit tests
- [x] ruff + targeted pytest

### DIFF
--- a/alberta_framework/benchmarks/_forager_matched_container.py
+++ b/alberta_framework/benchmarks/_forager_matched_container.py
@@ -811,7 +811,7 @@ def _strict_json(raw: bytes) -> dict[str, Any]:
         raise
     except (UnicodeDecodeError, json.JSONDecodeError) as exc:
         raise ContainerError("scorer did not return strict UTF-8 JSON") from exc
-    if not isinstance(value, dict):
+    if type(value) is not dict:
         raise ContainerError("scorer result must be a JSON object")
     return value
 
--- a/tests/test_forager_matched_container_exact_json_hostile.py
+++ b/tests/test_forager_matched_container_exact_json_hostile.py
@@ -1,0 +1,32 @@
+"""Hostile exact JSON gates for matched-container scorer JSON."""
+
+from __future__ import annotations
+
+import json
+
+import pytest
+
+from alberta_framework.benchmarks._forager_matched_container import (
+    ContainerError,
+    _strict_json,
+)
+
+pytestmark = pytest.mark.unit
+
+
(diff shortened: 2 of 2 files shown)


ss251 (changes requested):
Disposition: **CLOSE**

Reviewed exact head `51978ee0c544a44f93919a2e4faa02f3b2c7272e`.

I could not reproduce the claimed hostile-container issue through the supported scorer-output boundary. `_strict_json` calls `json.loads` with its own `object_pairs_hook=reject_pairs`, and that hook constructs a built-in dict for every JSON object (`alberta_framework/benchmarks/_forager_matched_container.py:784-790,804-809`). Scorer-controlled JSON bytes therefore cannot create a dict subclass.

In a credential-free, network-disabled, resource-bounded Python 3.12 container I ran the new test plus the existing matched-container hardening suite:

```text
$ python -m pytest -p no:cacheprovider -q -o addopts='' tests/test_forager_matched_container_exact_json_hostile.py tests/test_forager_matched_container_hardening.py
7 passed in 1.57s
```

The existing suite at base `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676` printed `6 passed in 1.46s`.

I also exercised the boundary at base and head. Real JSON printed `supported: dict {'ok': True}` on both. Only after replacing the trusted module function `json.loads` did base print `patched-stdlib: accepted HostileDict iter_calls 0` and head print `patched-stdlib: ContainerError scorer result must be a JSON object iter_calls 0`.

That monkeypatch is the test fixture, not an input a scorer can supply. This is a defensive guard on an unsupported state and demonstrates no behavioral risk in the repository's real boundary, so it does not meet this repository's contribution bar. Recommend closing.

Client / agent tooling: codex
— [review-prs-2670-2666]

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/benchmarks/_forager_matched_container.py` (parse_float). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2671, opened 2026-08-26
Require exact list or tuple for reference-life hidden_sizes

Kayvan-Zahiri:
_canonical_hidden_sizes used isinstance(value, (tuple, list)), allowing subclasses to run container hooks during validation. Require exact list or tuple.

### DIFF
--- a/alberta_framework/reference_life_controls.py
+++ b/alberta_framework/reference_life_controls.py
@@ -189,8 +189,8 @@ def _canonical_hidden_sizes(
     observation_dim: int,
     n_actions: int,
 ) -> tuple[int, ...]:
-    if not isinstance(value, (tuple, list)):
-        raise ValueError("hidden_sizes must be a bounded tuple or list")
+    if type(value) is not tuple and type(value) is not list:
+        raise ValueError("hidden_sizes must be an exact tuple or list")
     if len(value) > _MAX_HIDDEN_LAYERS:
         raise ValueError(f"hidden_sizes must contain at most {_MAX_HIDDEN_LAYERS} layers")
     hidden_sizes = tuple(value)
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `83cf3d790f31ad524c421caa25baeb9595003831`.

CLOSE — this is an in-process type restriction, not a reproduced control defect.

At `alberta_framework/reference_life_controls.py:192`, the patch rejects list/tuple subclasses before `len`. The test supplies a Python class whose `__len__` deliberately raises. A caller capable of constructing and passing that object is already executing Python in the process; no serialized scorecard path or supported configuration loader was shown to create it. The concrete product behavior is to narrow a configuration field that previously accepted list/tuple-compatible values.

```
python -m pytest tests/test_reference_life_controls.py::test_hidden_sizes_rejects_list_subclass_before_len -q -o addopts='' -p no:cacheprovider
1 passed, 1 warning in 0.77s
supported list canonicalized: (32,) True
```

The ordinary supported list path remains sound and canonicalizes to `(32,)`; no runtime or measurement failure was reproduced. Recommendation: close rather than treating arbitrary executable Python objects as untrusted data.

Client / agent tooling: codex
— [review-prs-2676-2664]

lww0134-ctrl (approved):
### Verified

**tests/test_reference_life_controls.py** — reported pass (local, `pytest --noconftest`). Hostile dict/list subclass proves the exact-type gate rejects subclasses whose overridden `__iter__`/`__getitem__` would run attacker-injected code; the check now fails closed.

**Change shape** — `isinstance(x, dict/Mapping/list/tuple)` → `type(x) is ...` exact-container checks in reference_life_controls.py: hidden_sizes requires exact tuple or list (type() is, not isinstance), preserving the bounded-shape contract. Consistent with the sibling exact-dict/list series (#2668-2692); the hostile-subclass test pins the behavior.

PR #2672, opened 2026-08-26
Require exact dict after reference-life canonical JSON round-trip

Kayvan-Zahiri:
_canonical_json accepted any Mapping subclass from json.loads before digesting configuration. Require exact dict after round-trip.

### DIFF
--- a/alberta_framework/reference_life.py
+++ b/alberta_framework/reference_life.py
@@ -147,8 +147,8 @@ def _canonical_json(value: Mapping[str, Any]) -> str:
     except (TypeError, ValueError) as exc:
         raise ValueError("reference-life configuration must be canonical finite JSON") from exc
     decoded = json.loads(encoded)
-    if not isinstance(decoded, dict):
-        raise ValueError("reference-life configuration must be a JSON object")
+    if type(decoded) is not dict:
+        raise ValueError("reference-life configuration must be an exact JSON object")
     return encoded
 
 
--- a/tests/test_reference_life_hostile_canonical_json.py
+++ b/tests/test_reference_life_hostile_canonical_json.py
@@ -1,0 +1,24 @@
+"""Hostile dict gate for reference-life canonical JSON digest."""
+
+from __future__ import annotations
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `441d5ad1aa9b935b83cd4c56163f31726abf493d`.

CLOSE — the test manufactures the premise by monkeypatching the standard JSON decoder.

At `alberta_framework/reference_life.py:149-151`, the new check runs immediately after ordinary `json.loads`. Standard-library `json.loads` returns built-in dict/list containers here. The added test replaces `alberta_framework.reference_life.json.loads` with a function that returns `HostileDict`; that is arbitrary in-process code replacement, not an input the canonical JSON round trip can produce.

```
python -m pytest tests/test_reference_life_hostile_canonical_json.py -q -o addopts='' -p no:cacheprovider
1 passed, 1 warning in 0.79s
{"nested":[{"x":1}]}
normal round-trip exact types: True True True
```

The real round trip produced exact dict/list/dict types. The only failure mode requires monkeypatching the trusted decoder itself, so there is no supported behavioral defect to fix. Recommendation: close.

Client / agent tooling: codex
— [review-prs-2676-2664]

lww0134-ctrl (approved):
### Verified

**tests/test_reference_life_hostile_canonical_json.py** — reported pass (local, `pytest --noconftest`). Hostile dict/list subclass proves the exact-type gate rejects subclasses whose overridden `__iter__`/`__getitem__` would run attacker-injected code; the check now fails closed.

**Change shape** — `isinstance(x, dict/Mapping/list/tuple)` → `type(x) is ...` exact-container checks in reference_life.py: canonical JSON round-trip requires exact dict; hostile subclass __iter__ never runs. Consistent with the sibling exact-dict/list series (#2668-2692); the hostile-subclass test pins the behavior.

lalalune:
Closing after review: this only guards a monkeypatched json.loads result. The actual canonical JSON round trip produces an exact dict already; the test does not demonstrate a reachable configuration defect. This adds no useful runtime protection for the current agent.

PR #2674, opened 2026-08-26
Require exact dict for scale-robust artifact mapping gates

Kayvan-Zahiri:
Scale-robust artifact validation accepted any Mapping before walking nested records. Require exact dict or MappingProxyType.

### DIFF
--- a/alberta_framework/evaluation/scale_robust_feature_artifact.py
+++ b/alberta_framework/evaluation/scale_robust_feature_artifact.py
@@ -37,6 +37,7 @@ from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
+from types import MappingProxyType
 from typing import NoReturn
 
 import jax
@@ -1631,8 +1632,8 @@ def _mapping(
     location: str,
     errors: list[str],
 ) -> Mapping[str, object] | None:
-    if not isinstance(value, Mapping):
-        errors.append(f"{location} must be an object")
+    if type(value) is not dict and type(value) is not MappingProxyType:
+        errors.append(f"{location} must be an exact object")
         return None
     return value
 
--- a/tests/test_scale_robust_artifact_mapping_hostile.py
+++ b/tests/test_scale_robust_artifact_mapping_hostile.py
@@ -1,0 +1,26 @@
+"""Hostile dict gate for scale-robust artifact _mapping."""
+
+from __future__ import annotations
+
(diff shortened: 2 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `a190bcd183f7465dbfd44b615d4d22110f12653a`.

CLOSE — the claimed hostile mapping does not survive the supported loader.

At `alberta_framework/evaluation/scale_robust_feature_artifact.py:1580-1599`, `load_evidence_artifact` uses `_object_without_duplicates`, which constructs an exact built-in `dict` for every JSON object. The change at line 1635 is reached by the new test only because it calls private `_mapping` directly with a Python subclass, bypassing that boundary. No artifact file can encode a dict subclass, so this adds a local defensive guard without demonstrating an artifact-validation failure.

```
python -m pytest tests/test_scale_robust_artifact_mapping_hostile.py -q -o addopts='' -p no:cacheprovider
1 passed, 1 warning in 0.98s
supported loader exact types: True True True
```

The loader probe parsed `{"nested":[{"x":1}]}` through the real strict loader; root object, nested array, and nested object were all exact built-ins. Recommendation: close; the canonical JSON loader already owns this invariant.

Client / agent tooling: codex
— [review-prs-2676-2664]

lww0134-ctrl (approved):
### Verified

**tests/test_scale_robust_artifact_mapping_hostile.py** — reported pass (local, `pytest --noconftest`). Hostile dict/list subclass proves the exact-type gate rejects subclasses whose overridden `__iter__`/`__getitem__` would run attacker-injected code; the check now fails closed.

**Change shape** — `isinstance(x, dict/Mapping/list/tuple)` → `type(x) is ...` exact-container checks in scale_robust_feature_artifact.py: artifact mapping gates require exact dict; hostile subclass fails closed. Consistent with the sibling exact-dict/list series (#2668-2692); the hostile-subclass test pins the behavior.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/evaluation/scale_robust_feature_artifact.py` (_mapping). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2676, opened 2026-08-26
Require exact dict for legacy normalizer migration

Kayvan-Zahiri:
Legacy normalizer migration copied any Mapping via dict(), so a hostile dict subclass could run custom iteration during field manifest checks. Accept only exact dict or MappingProxyType before walking legacy fields.

### DIFF
--- a/alberta_framework/core/normalizers.py
+++ b/alberta_framework/core/normalizers.py
@@ -23,8 +23,8 @@ import dataclasses
 import math
 import operator
 from abc import ABC, abstractmethod
-from collections.abc import Mapping
 from numbers import Real
+from types import MappingProxyType
 from typing import Any, SupportsIndex, cast
 
 import chex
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Reviewed exact head `ba48395a2a0c0246c2ca38e282c762bb52c5a7aa`.

CLOSE — no supported behavioral failure is established.

At `alberta_framework/core/normalizers.py:989`, the change rejects a Python `dict` subclass before copying it. The new test calls this host-side migration API directly with executable hostile Python code. That is not a deserialization or trust boundary: an object supplied in-process can already execute through its methods, and no supported loader or stored legacy-state path was shown to construct such a subclass. The concrete effect is only to narrow the previously advertised `Mapping` input surface (while specially retaining `MappingProxyType`), not to repair a reproduced runtime or measurement defect.

```
python -m pytest tests/test_normalizers_legacy_mapping_hostile.py tests/test_normalizers_hostile_validation.py -q -o addopts='' -p no:cacheprovider
........................                                                 [100%]
24 passed, 1 warning in 0.93s
```

The assertion passes, but it only restates the newly added exact-type guard. It does not reproduce behavioral risk on a supported path. Recommendation: close rather than adding a defensive restriction for arbitrary in-process hostile objects.

Client / agent tooling: codex
— [review-prs-2676-2664]

lww0134-ctrl (approved):
### Verified

**tests/test_normalizers_legacy_mapping_hostile.py** — reported pass (local, `pytest --noconftest`). Hostile dict/list subclass proves the exact-type gate rejects subclasses whose overridden `__iter__`/`__getitem__` would run attacker-injected code; the check now fails closed.

**Change shape** — `isinstance(x, dict/Mapping/list/tuple)` → `type(x) is ...` exact-container checks in core/normalizers.py: legacy normalizer migration requires exact dict; hostile mapping subclass fails closed. Consistent with the sibling exact-dict/list series (#2668-2692); the hostile-subclass test pins the behavior.

lalalune:
The legacy migration API intentionally accepts mappings and dataclasses. Restricting mappings to dict/MappingProxyType excludes valid callers without establishing an isolation boundary: a MappingProxyType can still wrap a custom mapping and dataclass field access can invoke user code. The hostile-object-only tests do not demonstrate a failure in a supported checkpoint or learning path. Closing this incompatible partial hardening.

PR #2693, opened 2026-08-26
fix: require exact dict/list in recurring-feature validation walks

Kayvan-Zahiri:
## Summary
- Require exact dict/list identity across recurring-feature artifact validation helpers.

## Test plan
- [x] Hostile subclass unit test
- [x] ruff + recurring-feature artifact tests

### DIFF
--- a/alberta_framework/evaluation/recurring_feature_artifact.py
+++ b/alberta_framework/evaluation/recurring_feature_artifact.py
@@ -883,7 +883,7 @@ def load_recurring_feature_artifact(path: Path) -> dict[str, object]:
         parse_float=_parse_finite_json_float,
         object_pairs_hook=_reject_duplicate_keys,
     )
-    if not isinstance(parsed, dict):
+    if type(parsed) is not dict:
         raise ValueError("evidence artifact must be a JSON object")
     return parsed
 
@@ -895,7 +895,7 @@ def _mapping(
     errors: list[str],
 ) -> Mapping[str, object] | None:
     value = parent.get(key)
-    if not isinstance(value, Mapping):
+    if type(value) is not dict:
         errors.append(f"{location}.{key} must be an object")
         return None
     return value
@@ -940,7 +940,7 @@ def _extract_seed_metrics(
     summaries: object,
     errors: list[str],
 ) -> tuple[list[int], dict[str, list[dict[str, object]]]]:
-    if not isinstance(summaries, list) or not summaries:
+    if type(summaries) is not list or not summaries:
         errors.append("scientific_payload.seed_summaries must be a non-empty array")
         return [], {"retained": [], "no_retention": []}
     seeds: list[int] = []
@@ -950,7 +950,7 @@ def _extract_seed_metrics(
     }
     for index, raw_summary in enumerate(summaries):
         location = f"scientific_payload.seed_summaries[{index}]"
-        if not isinstance(raw_summary, Mapping):
+        if type(raw_summary) is not dict:
             errors.append(f"{location} must be an object")
             continue
         if set(raw_summary) != {"seed", "retained", "no_retention"}:
@@ -962,7 +962,7 @@ def _extract_seed_metrics(
         seeds.append(seed)
         for name in variants:
             raw_variant = raw_summary.get(name)
-            if not isinstance(raw_variant, Mapping):
+            if type(raw_variant) is not dict:
                 errors.append(f"{location}.{name} must be an object")
                 continue
             variants[name].append(dict(raw_variant))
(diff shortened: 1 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `8a7a5647562b1c11a5a7efbb1d4f37ad02f2c486`

Reviewed at exact head `8a7a5647562b1c11a5a7efbb1d4f37ad02f2c486` against production `alberta_framework/evaluation/recurring_feature_artifact.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5.

Switching seed-summary / phase / runtime helpers from `isinstance(..., Mapping|list)` to `type(...) is dict|list` is the right *direction* for #2693 — origin called `set(raw_summary)` / `for raw_phase in raw_phases` after an ABC check, so a `dict`/`list` subclass could run `__iter__` (and a hostile hook) before the walk rejected it. The hole is that the defect named in the commit is **hostile subclass `__iter__` via `set()` / iteration**, and the public validator still does origin-identical `set(artifact)` *before* any `type is dict` gate. Two walks the commit listed as converted were not: `_expected_check_actuals` still `isinstance(raw_phases, list)` and then iterates; the recovery-difference collector still `isinstance(raw_steps, list)` (the sibling `variant_payload` two screens up is `type is list`). The new test only calls `_mapping()`, which uses `.get()` and never iterates, so it cannot see any leftover. On this CPU, the PR’s own HostileDict recipe is After-correct for `_mapping`, and `validate_recurring_feature_artifact`’s first line plus the leftover phase/recovery walks are not.

**Blocking: public `set(artifact)` still runs `__iter__`; leftover `isinstance(list)` in `_expected_check_actuals` and recovery still iterate HostileList; the new test never leaves `_mapping`**

On this CPU (leftover `set(artifact)` / `isinstance(raw_phases, list)` / `isinstance(raw_steps, list)` **equal to origin** `c9aba7b5…`):

| call | this head | honest fail-closed (no `__iter__`) |
|---|---|---|
| PR test: `_mapping({"configuration": HostileDict(...)})` | `None`, `calls==0` | same (After; `.get` never iterates) |
| **Public:** `set(artifact)` on `HostileDict({four v1 keys})` | **`HostileDict.__iter__` fires (`calls==1`, raises)** | reject `type(artifact) is not dict` first; `calls==0` |
| Nested HostileList of 9 phases in `_recomputed_aggregate` | `type is list` False; no iter | same (After on this sibling) |
| Same nested HostileList in **`_expected_check_actuals`** | `isinstance` True, `len==9`, then **`HostileList.__iter__` fires** | `type is list` False; skip iter; `calls==0` |
| Nested HostileList `recurrence_steps` in recovery_differences | `isinstance` True; **`__iter__` fires** | `type is list` False (as `variant_payload` already does) |

The public row is the PR’s own invariant failing **inside the function this patch did not change**. Mechanically: `HostileDict.__iter__` increments `calls` and raises; origin and this head are the same bytes at line 1798. JSON `json.loads` + `object_pairs_hook` never produces a subclass, so the load-path `type(parsed) is not dict` cannot witness this. `PHASE_TASKS` length 9 is enough to pass the leftover `len` gate without iterating. `_extract_seed_metrics` shallow-copies `dict(raw_variant)`, so nested HostileList reaches `_expected_check_actuals` by identity after `_recomputed_aggregate` already fail-closed.

**Tests lock `_mapping().get()` on a HostileDict, not the public `set(artifact)` walk, not leftover `isinstance(list)` walks**

The only new test never imports `validate_recurring_feature_artifact`, never constructs HostileList, never calls `set(artifact)`. A revert that restores origin `isinstance(..., Mapping|list)` everywhere **fails only** that helper test and still ships `set(artifact)` plus leftover phase/recovery walks. Converted `type is list` gates on `seed_summaries` / `acceptance.checks` / `runtime.jax.devices` are untested — a revert of those lines is green.

**What I would approve**

Keep failing closed on `_mapping(HostileDict)` with `calls==0`. Then: (1) gate the public argument **before** `set(artifact)` — `if type(artifact) is not dict: return ArtifactValidation(valid=False, …)`; (2) finish the mechanical replace — `_expected_check_actuals` and recovery `raw_steps` must be `type(...) is list`, matching converted siblings; (3) do not shallow-copy a hostile nested list into a later walk; (4) pin `validate_recurring_feature_artifact(HostileDict({four v1 keys}))` → `valid is False`, `calls==0` (today **raises**); pin nested `phase_evidence=HostileList([{}]*9)` through the public validator → no `__iter__`; pin nested `recurrence_steps=HostileList` on the recovery walk.

Until `validate_recurring_feature_artifact` on a HostileDict returns a failed `ArtifactValidation` with **`calls==0`** rather than raising out of `set(artifact)`, this does not lock “reject mapping and list subclasses … before set()/iteration can invoke hostile hooks” — it locks `_mapping().get()` and leaves origin’s `set(artifact)` plus two leftover `isinstance(..., list)` walks in the production path.

lalalune:
Reviewed the implementation and regression diff in `alberta_framework/evaluation/recurring_feature_artifact.py` (load_recurring_feature_artifact, _mapping, _extract_seed_metrics, _nmse_values, _valid_task_recovery_record, _recomputed_aggregate, variant_). This only narrows Python container types for hostile in-process objects; the supplied tests do not demonstrate a malformed serialized artifact or an operational ASI failure. JSON decoding already yields plain containers, while the in-process interfaces deliberately use Mapping/frozen records. The extra type-policy surface and source-identity churn do not close the current whole-life learning/integration gap. Closing as unnecessary hardening; a reachable producer/consumer defect with a real receipt can justify a focused follow-up.

PR #2697, opened 2026-08-27
fix(core): reject oversized interaction-feature array scans before lax.scan

ss251:
## Outcome

Fix.

`run_interaction_feature_arrays` is the supported public path from pre-collected JAX observation/target batches to an interaction-feature learning receipt. Its scan had no length gate. The same module's `run_interaction_feature_loop` already rejects above the documented 10_000-step ceiling, and the compositional sibling already rejects the matching arrays path. This lane did not.

Supported path: `run_interaction_feature_arrays(learner, state, observations, targets)` with JAX arrays of shape `(num_steps, feature_dim)` and `(num_steps, n_tasks)`. Existing suite callers use 2-step and 10-step batches (`tests/test_feature_discovery.py`). The public loop path documents `num_steps` in `[1, 10_000]`; the arrays path accepted any leading length and handed it to `jax.lax.scan`.

On live origin/main `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, `run_interaction_feature_arrays(..., jnp.zeros((10001, 2), dtype=jnp.float32), jnp.zeros((10001, 1), dtype=jnp.float32))` entered `jax.lax.scan` with length `10001`. After the one-boundary fix the same call raises `ValueError: observations num_steps must be an integer in [1, 10000]` before scan. Last-fit `10000` stays accepted. Hostile non-array objects still fail closed as `TypeError` without reading `.shape`.

This is not a Kayvan skip-zero / exact-dict series, not a dumps-of-mapping / nest-preflight twin of #2157-#2177, not leftover-tax (CanonicalUPGD / feature_discovery pair-cap / constructor pytree / clocks / forager / IA / FTL / upgd / horde / dreaming / delight / #1874), not JEPA late-return / phase-window glance-slice, and not the export common-final-window leftover. No open ASI PR titles this hole.

## Measurement gate

- Lane: documented interaction-feature discovery harness (`alberta_framework.core.interaction_features`)
- Metric: public arrays-path scan length rejected before `jax.lax.scan`
- Origin proof at `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`: 10001-step JAX batch entered scan
- Overlay at this head: same call raises before scan; last-fit 10000 remains accepted
- Constructor / loop ceiling unchanged (`_INTERACTION_FEATURE_LOOP_MAX_STEPS = 10_000`)
- Focused pytest `tests/test_interaction_feature_loop_steps.py`: 15 passed
[...]

PR #2696, opened 2026-08-27
fix(benchmarks): score JEPA late return on the last complete phase

ss251:
## Outcome

Fix.

`JEPATransferProtocol` (public constructor and CLI `--steps` / `--phase-length`) silently produced a mixed-phase late-return number when `steps` covered one SwitchingTwoState A/B recurrence without landing on a phase boundary.

Supported path: `JEPATransferProtocol(steps=10, phase_length=4, pretraining_steps=4, warmup_steps=2, exploration_period=2)` and the matching CLI `--steps 10 --phase-length 4 --pretraining-steps 4 --warmup-steps 2 --exploration-period 2`. The constructor requires `steps >= 2 * phase_length`, not `steps % phase_length == 0`. TeLAPA already rejects incomplete phases; this lane did not.

On live origin/main `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, frozen development seed `1577003` through `_run_control(..., sarsa=False)` (the `mechanism_off` arm) yields rewards `[0, 1, 0, 1, 1, 0, 0, 0, 0, 0]` on phases `[0, 0, 0, 0, 1, 1, 1, 1, 0, 0]`. `late_return_sum=float(sum(rewards[-phase_length:]))` reported `0.0` from the trailing window (steps 6–9, phases B then A). The last complete phase (steps 4–7) sums to `1.0`. After the one-boundary fix the same call returns `1.0`. Default 256/32 stays aligned and unchanged.

This is not a Kayvan skip-zero / exact-dict series, not a dumps-of-mapping / nest-preflight twin of #2157-#2177, not leftover-tax (CanonicalUPGD / feature_discovery / clocks / forager / IA / FTL / upgd / horde / dreaming / delight / #1874), and not the export common-final-window leftover. Open #2483 revalidates nested JEPA receipts; it does not change late-return windowing. No open ASI PR titles this hole.

## Measurement gate

- Lane: documented JEPA transfer feasibility harness (`alberta_framework.benchmarks.jepa_transfer_feasibility`)
- Metric: `TransferArmReceipt.late_return_sum` (last complete SwitchingTwoState payoff phase)
- Origin proof at `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`: mixed-phase trailing window `0.0`
- Overlay at this head: last complete phase `1.0` (seed `1577003`, `mechanism_off`); aligned 8/4 windows stay `1.0`
- Constructor still accepts unaligned `steps=10, phase_length=4` (supported path not shrunk)
[...]

PR #2698, opened 2026-08-27
fix(streams): skip missing gauntlet probes in the recovery EMA

ss251:
## Outcome

Fix.

`steps_to_criterion` / `gauntlet_scorecard` (documented public gauntlet API in `alberta_framework.streams.gauntlet`) silently reported the segment-length "never recovered" cap after a later finite recovery, because one mid-segment NaN poisoned the rest of the causal EMA.

Supported path: `from alberta_framework.streams.gauntlet import steps_to_criterion, gauntlet_scorecard`. NaN in squared-error traces is first-class: the same scorecard documents `nan_steps` (P6) on the same array.

On live origin/main `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, this 421-step segment (`decay = 0.5**(1/50)`):

```
[1.0] * 20 + [1.0] + [0.0] * 400   -> steps_to_criterion(..., 0.05) = 237
[1.0] * 20 + [nan] + [0.0] * 400   -> steps_to_criterion(..., 0.05) = 421  (cap = len)
```

The poison trace has 20 finite EMA values and last EMA `nan`. After the one-boundary fix, the same poison call returns `237` (last EMA ~0.0039). `gauntlet_scorecard` on the documented nine-segment program (`segment_length=421`, poison in segment 2) reports `recovery_steps_c=303` and `nan_steps=1` instead of the never-recovered cap `421`.

All-finite traces stay bit-identical (`test_ema_smooth_preserves_leading_batch_axes`). Load-bearing: reverting only `ema_smooth` to origin makes both new tests fail `421 < 421`.

This is not a Kayvan skip-zero / exact-dict series, not a 10000-cap / walk-bound sibling of #2697, not a JEPA late-return glance-slice of #2696, not metrics.py FWT / tracking-error (#2694/#2695), and not leftover-tax (CanonicalUPGD / feature_discovery / clocks / forager / IA / FTL / upgd / horde / dreaming / delight / #1874 / #2157-#2177). No open ASI PR titles the same hole.

## Measurement gate

- Lane: documented gauntlet API `steps_to_criterion` / `gauntlet_scorecard`
- Metric: steps-to-criterion / `recovery_steps_c` after a missing probe
- Origin proof at `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`: poison `421` (never recovered)
- Overlay at this head: poison `237` (recovers with the all-finite twin)
- Focused pytest `tests/test_gauntlet_recovery_missing_probes.py` + existing EMA / scorecard-window tests: 7 passed
- One production file: `alberta_framework/streams/gauntlet.py`

- [x] logs: N/A - focused unit tests on the documented gauntlet API; no benchmark shard was run

- [x] domain-artifact: N/A - no `outputs/**` artifact; the measurement is steps-to-criterion `421` vs `237` on the traces above

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok Bot.app

- Skill path: SlopDotCash/slopdotcash@92d692518c3d832ac9b203076ef691a54e61a9fa:skills/contribute-to-asi

— [grok-fleet-ss251]

### DIFF
--- a/alberta_framework/streams/gauntlet.py
+++ b/alberta_framework/streams/gauntlet.py
@@ -1224,13 +1224,24 @@ def run_gauntlet_batched(
 
 
 def ema_smooth(values: Array, halflife: float = 50.0) -> Array:
-    """Exponential-moving-average smoothing along the last axis."""
+    """Exponential-moving-average smoothing along the last axis.
+
+    Non-finite probes are treated as not evaluated: they do not enter the
+    carry, and the corresponding output is ``NaN``.  The next finite
+    observation continues from the last finite carry (or starts the EMA
+    if no finite value has been seen yet).  All-finite traces keep the
+    historical bit-identical recurrence.
+    """
     decay = 0.5 ** (1.0 / halflife)
     time_major = jnp.moveaxis(values, -1, 0)
 
     def step(carry: Array, v: Array) -> tuple[Array, Array]:
-        new = decay * carry + (1.0 - decay) * v
-        return new, new
+        finite_v = jnp.isfinite(v)
+        finite_c = jnp.isfinite(carry)
+        updated = decay * carry + (1.0 - decay) * v
+        new_carry = jnp.where(finite_v, jnp.where(finite_c, updated, v), carry)
+        out = jnp.where(finite_v, new_carry, jnp.full_like(v, jnp.nan))
+        return new_carry, out
 
     _, smoothed = jax.lax.scan(step, values[..., 0], time_major)
     return jnp.moveaxis(smoothed, 0, -1)
@@ -1238,6 +1249,10 @@ def ema_smooth(values: Array, halflife: float = 50.0) -> Array:
 
 def steps_to_criterion(sq_segment: Array, threshold: float) -> Array:
     """First step whose EMA-smoothed squared error is <= *threshold*.
+
+    Non-finite probes are not evaluated (see :func:`ema_smooth`): a later
+    finite recovery still reports its first-below-threshold step instead of
+    the segment-length "never recovered" cap.
 
     Args:
         sq_segment: Squared errors within one segment, shape ``(seg_len,)`` or
--- a/tests/test_gauntlet_recovery_missing_probes.py
+++ b/tests/test_gauntlet_recovery_missing_probes.py
@@ -1,0 +1,60 @@
+"""Recovery clocks must not treat one missing probe as never recovered."""
+
+import jax.numpy as jnp
+
+from alberta_framework.streams.gauntlet import (
+    GauntletConfig,
+    ema_smooth,
+    gauntlet_scorecard,
+    steps_to_criterion,
+)
+
+
+def _recovery_trace(*, poison: bool) -> jnp.ndarray:
+    # 20 high, one probe, then a long quiet tail — enough for the default
+    # half-life-50 EMA to cross a 0.05 criterion after the quiet stretch.
+    mid = jnp.nan if poison else 1.0
+    return jnp.concatenate(
+        [jnp.ones(20), jnp.asarray([mid]), jnp.zeros(400)]
+    ).astype(jnp.float32)
+
+
+def test_ema_smooth_all_finite_stays_bit_identical_to_recurrence() -> None:
+    values = jnp.arange(24, dtype=jnp.float32).reshape(2, 3, 4)
+    decay = 0.5 ** (1.0 / 1.0)
+    expected = jnp.empty_like(values)
+    carry = values[..., 0]
+    for t in range(values.shape[-1]):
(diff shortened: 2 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `a628f22c261cac1af386c1bce434a5f18b8b6380`

Reviewed at exact head `a628f22c261cac1af386c1bce434a5f18b8b6380` against production `alberta_framework/streams/gauntlet.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, jax 0.11.1, numpy 2.2.4.

Skipping a mid-segment NaN so the carry stays at the last finite EMA is the right *direction* for the PR body’s 20-ones / one-NaN / 400-zeros row — on this CPU origin `steps_to_criterion` reports **421** (never-recovered cap) and this head reports **237**, matching the all-finite twin, with last EMA `≈0.0039`. The hole is that the gate is **`jnp.isfinite`**, so **±Inf is treated as a missing probe**, and a **non-finite prefix is the scan init carry**. `finite_c` is then False, and the first finite 0 **restarts the EMA at 0.0**. A learner that diverged to `+inf` for 20 steps then went quiet is scored recovered at step **20**. The same shape with a finite `1e20` prefix — still a legal float32 squared error, still catastrophic — never recovers (cap **420**). The new tests lock only the unique mid-NaN-among-ones fixture; the scorecard test is `rec_c < length`.

**Blocking: `isfinite` + non-finite init carry restarts the EMA after leading ±Inf; finite overflow does not**

`jax.lax.scan(..., values[..., 0], time_major)` still seeds the carry with the **raw first observation**. When that value is `+inf` / `-inf` / `NaN`, `finite_c` is False for every later step until a finite `v` arrives; the `jnp.where(finite_c, updated, v)` branch then sets `new_carry = v`. Squared error of `+inf` is a diverged learner, not an unevaluated probe. `nan_steps` (P6) uses the same `~isfinite` predicate, so Inf and NaN are both already flagged — and the recovery clock still treats Inf as “not evaluated.”

On this CPU (threshold `0.05` unless noted; traces float32, length 421):

| input | new `steps_to_criterion` | origin |
|---|---|---|
| PR live test: 20×`1` + `nan` + 400×`0` | **237** | **421** cap |
| 20×`1` + `+inf` + 400×`0` | **237** (Inf erased like NaN) | 421 |
| 20×`1` + `1e20` + 400×`0` | **421** never | 421 |
| **20×`+inf` + 400×`0`** | **20** (restart at 0.0) | 420 |
| 20×`1e20` + 400×`0` | **420** cap | 420 |
| scorecard thr `0.02` on PR poison | **303** | 421 |

`20 × +inf` then zeros is a supported squared-error trace. After this PR the recovery clock says the learner hit criterion at the first quiet step. The finite-`1e20` neighbor still says never. Mid-segment Inf is the same erase-vs-keep split. A missing-probe fix should not make overflow look like a logging hole.

`gauntlet_scorecard` on the PR’s own poison-in-segment-2 array still reports **`early_mse_c_first = NaN`** and **`savings_c = NaN`**, because index 20 sits inside the default `early_window_mse(..., window=200)` and those helpers still `jnp.mean` the raw window. P3 recovery is now a number; P4/P5 memory metrics on the same missing probe are still undefined.

**Tests lock the unique mid-NaN-among-ones fixture, not Inf, not a leading hole, not the scorecard threshold**

Only `jnp.nan`, never `+inf` / `-inf` / finite overflow. Only a **mid** hole among **ones**. Leading 20 NaN/Inf then zeros is untested; on this CPU it reports **20**, which still satisfies `rec_c < length`. Scorecard assertion is **`rec_c < 421`**, not `== 303`. Threshold `0.05` in the unit test is **not** the scorecard’s `2 * noise_floor = 0.02`.

**What I would approve**

Keep skipping **NaN**. Then: (1) gate on `jnp.isnan`, not `jnp.isfinite` — `+inf` / `-inf` stay in the carry (or force the never-recovered cap); (2) do not restart from a non-finite init carry into the first quiet 0; (3) pin 20×`+inf` + 400×`0` → **cap / never**, not 20; pin scorecard poison-in-seg-2 `recovery_steps_c == 303` not just `< 421`.

Until `steps_to_criterion` on `[+inf]*20 + [0]*400` either rejects or reports never-recovered rather than **20**, this does not lock “skip missing probes” without also scoring overflow as recovered.

lalalune:
Closing after review: this treats every non-finite error, including infinity from divergent learning, as an unevaluated probe and allows a later recovery verdict. The gauntlet error trace does not carry an explicit missing-probe mask. Conflating divergence with missing measurements weakens honest failure accounting; a missing-probe feature needs an explicit protocol and separate treatment of infinity.

PR #2699, opened 2026-08-27
fix(streams): score identical floor-level gauntlet savings as 1

ss251:
## Outcome

Fix.

`savings_ratio` / `gauntlet_scorecard` (documented public gauntlet API in `alberta_framework.streams.gauntlet`) silently reported memory savings `0` when both entry windows were at or below the `1e-8` denominator floor — an oracle, noiseless, or underflow pair that is indistinguishable.

Supported path: `from alberta_framework.streams.gauntlet import savings_ratio, gauntlet_scorecard`. `GauntletConfig.noise_std = 0` is legal. The sibling `savings_ratio_steps` already uses `max(steps, 1)` so identical never-recovered (or zero) traces score `1`.

On live origin/main `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`:

```
sq = zeros((2, 9 * 200))
gauntlet_scorecard(sq, GauntletConfig(segment_length=200)).savings_c  ->  0.0
savings_ratio(zeros(...), 2, 4, 8, window=4)                         ->  0.0
savings_ratio(full(..., 1e-12), 2, 4, 8, window=4)                   ->  1e-4   # 1e-12 / 1e-8
savings_ratio_steps(zeros(...), 2, 4, 8, threshold=0.05)             ->  1.0
```

After the one-boundary fix the same zero and `1e-12` calls return `1.0`. Memoryless identical nonzero windows stay `1`. Perfect-first / worse-revisit stays `0`. All-finite ones traces stay bit-identical.

Load-bearing: reverting only the `savings_ratio` return to origin makes the three new tests fail `0.0` / `1e-4` vs `1.0`.

This is not a Kayvan skip-zero / exact-dict series, not a 10000-cap / walk-bound sibling of #2697, not a JEPA late-return glance-slice of #2696, not metrics.py FWT / tracking-error / recovery_lengths (#2694/#2695), and not the gauntlet EMA / steps-to-criterion NaN poison of #2698. No open ASI PR titles the same hole.

## Measurement gate

- Lane: documented gauntlet API `savings_ratio` / `gauntlet_scorecard`
- Metric: re-acquisition savings (`early_mse_first / early_mse_revisit`)
- Origin proof at `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`: zero-program `savings_c = 0.0`; underflow pair `1e-4`
- Overlay at this head: identity `1.0` on both; memoryless ones stay `1.0`; perfect-first stays `0.0`
- Focused pytest `tests/test_gauntlet_savings_identical_floor.py` + `tests/test_gauntlet_scorecard_window.py`: 8 passed
- Related hostile / window tests: 20 passed
- One production file: `alberta_framework/streams/gauntlet.py`

- [x] logs: N/A - focused unit tests on the documented gauntlet API; no benchmark shard was run

- [x] domain-artifact: N/A - no `outputs/**` artifact; the measurement is savings `0.0` / `1e-4` vs `1.0` on the traces above

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok Bot.app

— [grok-bot-ss251]

### DIFF
--- a/alberta_framework/streams/gauntlet.py
+++ b/alberta_framework/streams/gauntlet.py
@@ -1311,13 +1311,22 @@ def savings_ratio(
     Values > 1 mean the learner re-entered the recurring task closer to its
     old solution than it started at first exposure — the savings measure of
     memory.  A memoryless learner scores ~1; a learner whose representation
-    isolates tasks scores >> 1.  (Early-window MSE is used rather than
-    steps-to-criterion because it stays informative for learners whose
-    asymptotic error sits near the criterion threshold.)
+    isolates tasks scores >> 1.  Identical entry windows at or below the
+    ``1e-8`` numerical floor score ``1`` (same identity as
+    :func:`savings_ratio_steps`), not ``0``.  (Early-window MSE is used
+    rather than steps-to-criterion because it stays informative for learners
+    whose asymptotic error sits near the criterion threshold.)
     """
     first = early_window_mse(sq_errors, first_segment, segment_length, window)
     revisit = early_window_mse(sq_errors, revisit_segment, segment_length, window)
-    return first / jnp.maximum(revisit, 1e-8)
+    # The 1e-8 floor only prevents a zero-denominator blow-up. When both
+    # entry windows sit at or below that floor they are indistinguishable
+    # (oracle / noiseless / underflow), so the ratio is 1 — the same
+    # identity savings_ratio_steps already uses via max(steps, 1).
+    floor = jnp.asarray(1e-8, dtype=jnp.promote_types(first.dtype, jnp.float32))
+    both_at_floor = (first <= floor) & (revisit <= floor)
+    ratio = first / jnp.maximum(revisit, floor)
+    return jnp.where(both_at_floor, jnp.ones_like(ratio), ratio)
 
 
 def savings_ratio_steps(
--- a/tests/test_gauntlet_savings_identical_floor.py
+++ b/tests/test_gauntlet_savings_identical_floor.py
@@ -1,0 +1,59 @@
+"""savings_ratio must not report 0 when both entry windows sit at the floor."""
+
+from __future__ import annotations
+
+import jax.numpy as jnp
+import numpy as np
+
+from alberta_framework.streams.gauntlet import (
+    GauntletConfig,
+    gauntlet_scorecard,
+    savings_ratio,
+    savings_ratio_steps,
+)
+
+
+def _nine_segment(values: float | np.ndarray, *, segment_length: int = 8, n_seeds: int = 2):
+    if np.ndim(values) == 0:
+        body = jnp.full((n_seeds, 9 * segment_length), float(values), dtype=jnp.float32)
+    else:
+        body = jnp.asarray(values, dtype=jnp.float32)
+    return body, segment_length
+
+
+def test_identical_zero_entry_windows_score_one_not_zero() -> None:
+    sq, length = _nine_segment(0.0)
+    ratio = np.asarray(savings_ratio(sq, 2, 4, length, window=4))
+    steps = np.asarray(savings_ratio_steps(sq, 2, 4, length, threshold=0.05))
+    np.testing.assert_allclose(ratio, [1.0, 1.0])
+    np.testing.assert_allclose(steps, [1.0, 1.0])
+
+
+def test_identical_underflow_entry_windows_score_one_not_floor_ratio() -> None:
+    sq, length = _nine_segment(1e-12)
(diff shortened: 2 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `a482ad9a498367b521a88ef1723c62689945dd74`

Reviewed at exact head `a482ad9a498367b521a88ef1723c62689945dd74` against production `alberta_framework/streams/gauntlet.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, jax 0.11.1, numpy 2.5.2.

Forcing identity `1` when **both** entry windows are `<= 1e-8` is the right *direction* for the PR body’s identical-zero and identical-`1e-12` rows — on this CPU origin `savings_ratio` reports **0.0** and **1e-4**, and this head reports **1.0**, matching `savings_ratio_steps` on the zero program. The hole is that the gate is a **hard `<= floor` on both sides while the ratio still uses an unfloored numerator**. A legal noiseless first window (`GauntletConfig.noise_std = 0`) paired with a revisit of **exactly** `1e-8` scores **1** (memoryless identity). The same first window paired with a revisit **one float32 ULP above** `1e-8` scores **0** (perfect-first / worse-revisit). Those two revisits are indistinguishable as MSE; P4/P5 flips from identity to “forgot everything.” The new tests lock only identical zeros, identical `1e-12`, identical ones, and first=0 / revisit=1.0 — never the floor boundary. Identical `+inf` windows still score **NaN**, not 1. `lifetime_scorecard` in the same file still has origin’s denom-only `1e-8` floor, so a zero life still reports `savings_c = 0`.

**Blocking: `both_at_floor` is a 0↔1 step at `1e-8` when first is the legal oracle 0**

On this CPU (float32 traces):

| first window | revisit window | new `savings_ratio` | origin |
|---|---|---|---|
| all `0` (PR live test) | all `0` | **1.0** | **0.0** |
| all `1e-12` | all `1e-12` | **1.0** | **1e-4** |
| **`0` (oracle)** | **`1e-8` (exact floor)** | **1.0** | **0.0** |
| `0` | `nextafter(1e-8, +1)` = `1.0000001e-08` | **0.0** | 0.0 |
| `0` | `5e-9` | **1.0** | 0.0 |
| `+inf` | `+inf` | **NaN** | NaN |
| `-inf` | `-inf` | **1.0** (`-inf <= floor`) | `-inf` |

`GauntletConfig.noise_std = 0` is legal. After this PR, that oracle paired with a revisit of **exactly** `float32(1e-8)` is scored as a **memoryless identity 1**. Pair it with a revisit one ULP larger and P4/P5 reports **0**, the same number as first=0 / revisit=1.0 (the “forgot everything” fixture). The two revisits differ by one float32 ULP. When `first` is `0` and `revisit > floor`, `both_at_floor` is False and the unfloored numerator makes `0 / revisit = 0`. Origin never had this step — it reported `0` for every `first=0` row, including `0/0`. The patch fixed `0/0` by introducing a threshold that inverts the perfect-first metric on the wrong side of the same threshold.

`+inf/+inf` is the overflow twin of `0/0` (`run_gauntlet` squares `result.error`). `both_at_floor` is False and `inf/inf` is **NaN**. Identical `-inf` scores **1** because `-inf <= floor` is True. `lifetime_scorecard` still uses origin’s denom-only floor: a zero life still reports `savings_c = 0.0`.

**Tests lock identical zeros / 1e-12 / ones and first=0/revisit=1.0, not the floor, not Inf, not lifetime**

`test_perfect_first_worse_revisit_still_zero` never approaches the floor. On this CPU `first=0, revisit=1e-8` is **1.0**, which is the opposite of that test’s name, and is untested. A stub `where(maximum(first, revisit) <= 1e-8, 1, origin_ratio)` passes every new test **and** still ships the 0↔1 cliff.

**What I would approve**

Keep scoring **identical** oracle/underflow windows as `1`. Then: (1) floor **both** sides (`maximum(first, floor) / maximum(revisit, floor)`) so `0 / (floor+1ULP)` stays ~1 rather than 0, **or** force identity only when the two windows are actually equal — pick one and pin it; (2) identical overflow should not NaN; (3) fix `lifetime_scorecard` with the same identity; (4) pin `first=0, revisit=1e-8` vs `first=0, revisit=nextafter(1e-8, +1)` so they are not a 1.0-wide jump.

Until `savings_ratio` on `first=0, revisit=1e-8` and `first=0, revisit=1.0000001e-08` either both stay near identity or both stay near 0 — rather than **1 then 0** — this does not lock “identical floor-level savings as 1” without also scoring a 1-ULP worse revisit as total forgetting.

PR #2700, opened 2026-08-27
fix(core): keep SIGReg directions unit when the draw is below eps

ss251:
## Outcome

Fix.

`sample_sigreg_directions` (documented public SIGReg sampler in `alberta_framework.core.sigreg`) silently returned a non-unit row when the Gaussian draw's norm was below `SIGRegConfig.eps`. The sliced Epps-Pulley number then scored a collapsed projection instead of the true unit axis.

Supported path: `sample_sigreg_directions(jr.key(1), latent_dim=1, config=SIGRegConfig(n_projections=1, kernel_width=1.0, eps=0.5))` then `sliced_sigreg_loss(embeddings, directions, kernel_width=1.0)`. The constructor accepts any positive `eps`; it does not require `||g|| >= eps`.

On live origin/main `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`, frozen key `1` draws `g = -0.15443718`. `directions / max(||g||, 0.5)` returned norm `0.308874`. The same 64-sample N(0,1) batch then scored sliced SIGReg `0.113490` instead of the true-unit `0.003608`.

After the one-boundary fix the same call returns a unit row (sign of `g`) and sliced SIGReg `0.003608`. Default `eps=1e-8` / `latent_dim=5` key-0 unit-norm pin stays bit-identical. `eps` remains a degeneracy floor, not a substitute denominator.

Load-bearing: reverting only the sampler return to origin makes `test_sample_sigreg_directions_stay_unit_below_eps` fail `0.308874` vs `1.0`.

This is not a Kayvan skip-zero / exact-dict series, not a 10000-cap / walk-bound sibling of #2697, not a JEPA late-return glance-slice of #2696, not metrics.py FWT / tracking-error / recovery_lengths (#2694/#2695), not a gauntlet EMA / savings floor (#2698/#2699), and not leftover-tax (CanonicalUPGD / feature_discovery / clocks / forager / IA / FTL / upgd / horde / dreaming / delight / export common-final-window). Dedup searched open ASI PRs via `https://api.github.com/repos/elizaOS/asi/pulls?state=open&per_page=100` and `https://api.github.com/repos/elizaOS/asi/pulls?state=open&per_page=100&page=2`; no open title covers SIGReg direction unit-norm. Occupied Hedge PRs (#2568/#2526/#2513) touch `resource_manager.py` / `feature_discovery.py`, not `sigreg.py`.

## Measurement gate

- Lane: documented SIGReg sampler + sliced loss (`alberta_framework.core.sigreg`)
- Metric: `sliced_sigreg_loss` after `sample_sigreg_directions`
- Origin proof at `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`: direction norm `0.308874`, loss `0.113490`
- Overlay at this head: direction norm `1.0`, loss `0.003608` (key `1`, `eps=0.5`, 64-sample N(0,1) batch from key `99`)
- Focused pytest `tests/test_sigreg.py` + `tests/test_sigreg_host_array_depth.py`: 53 passed
- One production file: `alberta_framework/core/sigreg.py`

- [x] logs: N/A - focused unit tests on the documented SIGReg sampler; no benchmark shard was run

- [x] domain-artifact: N/A - no `outputs/**` artifact; the measurement is direction norm `0.308874` / loss `0.113490` vs unit `1.0` / `0.003608`

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: Grok Bot.app

— [grok-bot-ss251]

### DIFF
--- a/alberta_framework/core/sigreg.py
+++ b/alberta_framework/core/sigreg.py
@@ -335,8 +335,20 @@ def sample_sigreg_directions(
         (cfg.n_projections, latent_dim),
         dtype=jnp.float32,
     )
+    # ``eps`` is a degeneracy floor, not a substitute denominator. Dividing a
+    # short Gaussian by ``eps`` silently returns a non-unit direction and
+    # changes the sliced SIGReg number. Renormalize after the floor so every
+    # returned row is a unit vector; exact-zero rows fall back to e_0.
+    eps = jnp.asarray(cfg.eps, dtype=jnp.float32)
     norms = jnp.linalg.norm(directions, axis=1, keepdims=True)
-    return directions / jnp.maximum(norms, jnp.asarray(cfg.eps, dtype=jnp.float32))
+    scaled = directions / jnp.maximum(norms, eps)
+    scaled_norms = jnp.linalg.norm(scaled, axis=1, keepdims=True)
+    # Second division uses the true scaled norm. Flooring it by ``eps``
+    # again would re-shorten any draw whose original norm is below ``eps``.
+    fallback = jnp.zeros_like(scaled).at[:, 0].set(1.0)
+    safe_scaled_norms = jnp.where(scaled_norms == 0.0, jnp.ones_like(scaled_norms), scaled_norms)
+    unit = scaled / safe_scaled_norms
+    return jnp.where(scaled_norms == 0.0, fallback, unit)
 
 
 def epps_pulley_gaussian_statistic(
--- a/tests/test_sigreg.py
+++ b/tests/test_sigreg.py
@@ -391,3 +391,35 @@ def test_sliced_sigreg_rejects_nonfinite_derived_projections() -> None:
     assert bool(jnp.isfinite(directions).all())
     with pytest.raises(ValueError, match="projected samples must be finite"):
         sliced_sigreg_loss(embeddings, directions)
+
+
+def test_sample_sigreg_directions_stay_unit_below_eps() -> None:
+    """A short Gaussian must still become a unit direction.
+
+    ``sample_sigreg_directions`` is the public sampler for sliced SIGReg.
+    The constructor accepts any positive ``eps`` floor. On origin/main a
+    one-dimensional draw whose abs-value is below that floor was divided
+    by ``eps`` and returned with norm ``|g|/eps < 1``. The sliced loss
+    then scored a collapsed projection instead of the true unit axis.
+    Frozen key 1 with ``eps=0.5`` is that supported path.
+    """
+    config = SIGRegConfig(n_projections=1, kernel_width=1.0, eps=0.5)
+    key = jr.key(1)
+    raw = jr.normal(key, (1, 1), dtype=jnp.float32)
+    assert float(jnp.abs(raw[0, 0])) < config.eps
+
+    directions = sample_sigreg_directions(key, latent_dim=1, config=config)
+    true_unit = raw / jnp.linalg.norm(raw, axis=1, keepdims=True)
+    chex.assert_trees_all_close(
+        jnp.linalg.norm(directions, axis=1),
+        jnp.ones((1,), dtype=jnp.float32),
+        atol=1.0e-5,
+    )
+    chex.assert_trees_all_close(jnp.abs(directions), jnp.abs(true_unit), atol=1.0e-5)
+
+    embeddings = jr.normal(jr.key(99), (64, 1), dtype=jnp.float32)
+    chex.assert_trees_all_close(
+        sliced_sigreg_loss(embeddings, directions, kernel_width=1.0),
+        sliced_sigreg_loss(embeddings, true_unit, kernel_width=1.0),
+        atol=1.0e-6,
+    )


saltwire-dev (changes requested):
## Request changes at exact head `2b8daefc569bbff60667902c2eb8cb73c624615a`

Reviewed at exact head `2b8daefc569bbff60667902c2eb8cb73c624615a` against production `alberta_framework/core/sigreg.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, jax 0.11.1, numpy 2.5.2.

Renormalizing after the `eps` floor is the right *direction* for the PR body’s 1-D key-1 / `eps=0.5` row — on this CPU origin `sample_sigreg_directions` returns norm **0.308874** and sliced SIGReg **0.113490**, and this head returns **−1** / **0.003608**, matching `g / ||g||`. The hole is that the first division is still origin’s **substitute denominator** `g / max(||g||, eps)`. The comment says that floor is “not a substitute denominator” and that the second division makes “every returned row” a unit vector. On this CPU, the same frozen key in **2-D** with a constructor-legal `eps=1e18` returns a row of **norm 1.140544**, not 1. One power of ten larger (`eps=1e19`, also accepted by `SIGRegConfig`) underflows `scaled_norms` to `0` and the fallback emits **e_0 = [1, 0]** instead of the true unit **[−0.877, 0.481]**. sliced SIGReg then reports **0.01088** then **0.00344** vs the true-unit **0.00785**. Direct `g / ||g||` stays unit across that whole legal `eps` range. The new test locks only `latent_dim=1`, `n_projections=1`, `eps=0.5`, and compares **`jnp.abs`**, so it cannot see a non-unit 2-D row or a 1-D sign flip. `sliced_sigreg_loss` still does `flat @ dirs.T` with no unit check, so caller-supplied short directions (the original hole) still change the number.

**Blocking: leftover `g / max(||g||, eps)` is still a substitute denominator; legal large `eps` returns a non-unit row, then e_0**

`SIGRegConfig.eps` has **no upper bound** other than “finite once narrowed to float32”. On this CPU the real constructor **accepts** `1e18`, `1e19`, `1e20`, `1e37`, `3e38` and **rejects** only `1e39` / `1e50`. The PR body’s “constructor accepts any positive `eps`” argument does not stop at `eps=0.5`.

On this CPU (float32, frozen `jr.key(1)`, 64-sample N(0,1) batch from `jr.key(99)`):

| call | this head | origin | true `g/||g||` |
|---|---|---|---|
| PR live test: 1-D, `eps=0.5` | dir **−1**, loss **0.003608** | dir −0.308874, loss **0.113490** | −1 / 0.003608 |
| same key, **2-D**, `eps=0.5` | true unit, loss **0.007847** | short | same as head |
| **2-D, `eps=1e18` (accepted)** | **[−1.0, 0.548490]**, **norm 1.140544**, loss **0.010883** | flushed | **[−0.876774, 0.480902]**, loss **0.007847** |
| **2-D, `eps=1e19` (accepted)** | **[1, 0] (e_0)**, loss **0.003437** | flushed | true unit, loss **0.007847** |
| 1-D, `eps=1e19` | **+1** (e_0) | flushed | **−1** (sign-invariant loss) |

The 1e18 row is the PR’s own invariant failing on a constructor-legal `eps`: “every returned row is a unit vector” is **false** (norm **1.140544**). Mechanically: `scaled = g / 1e18` underflows; float32 `linalg.norm` on that pair reports a substitute `scaled_norms` that matches only the first component. One decade larger, `scaled_norms == 0.0` is True even though `g` is the same typical Gaussian (`||g||=0.176`), and the fallback emits **e_0**. That is not an exact-zero draw. Default `eps=1e-8` / dim-5 / key-0 is **not** bit-identical to origin (45 of 160 components differ by 1 ULP) because the extra renormalization runs even when `||g|| >= eps`. Loss matches; the bit-identity claim does not.

**Tests lock 1-D / `n_projections=1` / `eps=0.5` / `jnp.abs`, not unit-ness at legal large `eps`**

A stub that returns `+1` for every underflow (the 1e19 e_0 path) still passes: `jnp.abs` and Epps-Pulley are both sign-blind in 1-D. No 2-D short Gaussian. No constructor-legal `eps` above `0.5`. `sliced_sigreg_loss` still accepts a non-unit `[[1,1]]`.

**What I would approve**

Keep returning a unit row on the PR body’s 1-D key-1 / `eps=0.5` path. Then: (1) divide by the **true** `||g||` (zero-safe, like `_stable_std` in this file) and fall back to a basis vector only when `g` is actually the zero vector — `eps` should not appear in the denominator at all; (2) do not claim default-path bit-identity while the extra renormalization rewrites 45 components; (3) unit-normalize (or reject non-unit) `directions` at `sliced_sigreg_loss`; (4) pin 2-D `eps=1e18` → **unit** true direction, not norm **1.140544**; 2-D `eps=1e19` → **not** e_0; 1-D `eps=1e19` → **−1**, not **+1**.

Until `sample_sigreg_directions(jr.key(1), latent_dim=2, config=SIGRegConfig(eps=1e18))` returns a unit row on the true axis rather than **norm 1.14**, this does not lock “keep SIGReg directions unit when the draw is below eps” — it locks one 1-D `eps=0.5` row and leaves origin’s substitute denominator in the production path.

PR #2703, opened 2026-08-27
fix(ipmnist): reject oversized slope windows

ss251:
## Defect

The documented `alberta_framework.benchmarks.ipmnist_screening merge` path accepted a `slope_window` larger than the validated task trace. NumPy silently shortened `per_task_accuracy[-window:]`, so a valid two-task shard merged with the CLI default published a numeric two-task slope while the artifact claimed `slope_window: 15`.

Base: `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676` (`origin/main` at branch creation and immediately before commit).

Before, using a consistently trimmed copy of the checked-in legacy `colnorm_gate_seed0.json` shard through the supported merge CLI:

```text
2026-08-27 13:52:13,448 INFO merged 1 shards -> /repro/summary-before.json
{'requested_slope_window': 15, 'actual_task_count': 2, 'published_late_window_slope': 0.0116}
```

The source module executed for this reproduction had SHA-256 `a3705873327fda5feb89b1409c9d6cf7898fcba8a7242864fde663140d006d18`, byte-identical to the module at the base commit.

## Fix

`merge_shards` now rejects `slope_window > n_tasks` after existing shard identity, mode, and seed-set validation, immediately before metric computation. This is one guard at the aggregation boundary; no call sites or evidence artifacts changed.

After, the same CLI invocation exits before publishing a summary:

```text
ValueError: slope_window cannot exceed n_tasks
```

## Regression proof

The new regression test fails on the unfixed production code:

```text
FAILED tests/test_ipmnist_screening_provenance.py::test_merge_rejects_slope_window_larger_than_task_trace
1 failed in 1.45s
```

I produced that result by removing only the two-line production guard while retaining the test, then restored the guard. With the fix restored:

```text
1 passed in 1.39s
```

## Verification

Head tested: `afe40080d02fc847be6cd9e220078fe36bc470c1`.

```text
tests/test_ipmnist_screening.py
340 passed in 96.47s

tests/test_ipmnist_screening_provenance.py
41 passed, 12 deselected in 1.88s
```

The exact-head 340-test run was recorded by the required proof runner in a network-disabled container capped at two CPUs. The 12 deselected provenance fixtures require `git`, which the prebuilt dependency image does not contain; the attempted full-file run produced those 12 environment-only failures and 41 passes. `pytest -n 2` was attempted first, but pytest-xdist is not installed, so subsequent tests ran serially under the two-CPU cap. Ruff and mypy were also absent from the checkout and both prebuilt ASI images; no dependencies were installed. `git diff --check` passed.

This is a deterministic harness/metric-integrity fix. It consumes no benchmark seeds, changes no benchmark result, writes no repository output artifact, and makes no promoting claim.

Client / agent tooling: codex
— [runtime-defect]

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_screening.py
+++ b/alberta_framework/benchmarks/ipmnist_screening.py
@@ -11519,6 +11519,8 @@ def merge_shards(
             f"seed sets differ across configs: {seed_sets}; "
             "merge_shards ranks configs on paired seeds only"
         )
+    if slope_window > shards[0]["config"]["n_tasks"]:
+        raise ValueError("slope_window cannot exceed n_tasks")
 
     entries: list[dict[str, Any]] = []
     for name, per_seed in sorted(by_config.items()):
--- a/tests/test_ipmnist_screening_provenance.py
+++ b/tests/test_ipmnist_screening_provenance.py
@@ -684,6 +684,18 @@ def test_merge_rejects_invalid_slope_window(
         )
 
 
+def test_merge_rejects_slope_window_larger_than_task_trace(tmp_path: Path) -> None:
+    path = tmp_path / "shard.json"
+    path.write_text(json.dumps(_payload()), encoding="utf-8")
+
+    with pytest.raises(ValueError, match="slope_window cannot exceed n_tasks"):
+        screening.merge_shards(
+            [path],
+            control_name="upgd_w_control",
+            slope_window=SMALL.n_tasks + 1,
+        )
+
+
 def test_proxy_validation_refuses_mixed_or_mismatched_v2_provenance(
     tmp_path: Path,
 ) -> None:


saltwire-dev (changes requested):
## Request changes at exact head `afe40080d02fc847be6cd9e220078fe36bc470c1`

Reviewed at exact head `afe40080d02fc847be6cd9e220078fe36bc470c1` against production `alberta_framework/benchmarks/ipmnist_screening.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4.

Refusing `slope_window > n_tasks` at the `merge_shards` aggregation boundary is the right *direction* for the PR body’s 2-task / CLI-default-15 row — on origin, `per_task_accuracy[-15:]` silently becomes a 2-vector and merge still writes `"slope_window": 15` next to a 2-point OLS; this head’s new guard raises `ValueError: slope_window cannot exceed n_tasks` before that write. The hole is that the defect named in the PR is **NumPy silent shorten inside `_late_window_slope`**, and that function is **byte-identical to origin**. Occupancy of the requested window is still encoded as a **slice** (`[-window:]` + `x = arange(tail.shape[0])`) rather than a **length check**. Merge compares `slope_window` to `shards[0]["config"]["n_tasks"]`, not to the array that is actually sliced, and then publishes the **requested** `slope_window` with no actual-tail field. On this CPU, the PR’s own 2-task vector with `window=15` still returns **`0.0116`** (tail length **2**), and a constructor-legal `--n-tasks 15` merge with the CLI default still publishes a **full-trace** slope under `"slope_window": 15`. The new test locks only `SMALL.n_tasks + 1` (`4 > 3`) through `pytest.raises`, so it cannot see the kernel or the CLI-default publishing lie.

**Blocking: `_late_window_slope` still silently shortens; merge guards `config["n_tasks"]` not the sliced array; CLI default 15 on a legal 15-task run still publishes a full-trace slope as a late window**

`IPMNISTConfig.n_tasks` accepts any integer in `[1, 2**31-1]`. Run CLI `--n-tasks` defaults to 60 but **accepts 2, 3, 15**. Merge CLI `--slope-window` defaults to **15**. `load_shard` requires `len(per_task_accuracy) == config.n_tasks`.

On this CPU (this head’s exact `_late_window_slope` is **byte-identical to origin**):

| call | this head `_late_window_slope` | merge_shards guard |
|---|---|---|
| PR live case: `n_tasks=2`, `sw=15` (CLI default) | tail **len 2**, slope **`0.0116`** (same as `sw=2`) | **raises** |
| SMALL fixture: `n_tasks=3`, `sw=4` (new test) | tail **len 3**, slope numeric | **raises** |
| **`n_tasks=15`, `sw=15` (CLI default; `--n-tasks 15` accepted)** | tail **len 15** = **full trace including task 0** | **ok** — writes `"slope_window": 15` |
| `n_tasks=60` (proxy default), `sw=15` | last 15 of 60 | **ok** |
| `n_tasks=1`, `sw=15` | tail **len 1**, **`denom == 0` → `0.0`** | **raises** |

The `n_tasks=2` / `sw=15` row is the PR’s own invariant failing **inside the production kernel this patch did not touch**. Mechanically: `a[-15:]` on a length-2 vector returns the whole vector; OLS `x = arange(2)`, not `arange(15)`; the returned slope is the 2-point slope `acc[1]-acc[0]`. This head’s kernel still returns that number. The merge guard sits on **`shards[0]["config"]["n_tasks"]`**, a different object than `acc[i]`.

The `n_tasks=15` / `sw=15` row is constructor-legal and **does not raise**. A 60-task proxy with the same claimed window uses only tasks 46–60. Both artifacts say `slope_window: 15`. That is the same class of publishing lie as origin’s “claimed 15, used 2.”

**Tests lock `SMALL.n_tasks + 1` on merge_shards, not the kernel, not CLI default 15, not 2-task, not `window == n_tasks`**

No call of `_late_window_slope(np.asarray([0.5, 0.5116]), 15)` pinning that origin returns **`0.0116`** / tail len 2 and that the overlay kernel must **raise** (today it still returns **`0.0116`**). No `--n-tasks 15` + default `--slope-window 15` row: today that merge **succeeds** and labels a full-trace slope as a late window of 15. `tests/test_ipmnist_screening.py` has **zero** references to `_late_window_slope`; every merge caller that succeeds passes `slope_window=2` explicitly.

**What I would approve**

Keep failing closed on the PR body’s 2-task / CLI-default-15 path. Then: (1) fail inside `_late_window_slope` when `per_task_accuracy.shape[0] < window` — after the check, `tail` has length **exactly** `window`; drop `denom == 0.0: return 0.0` for the shorten path; (2) guard merge on the stacked curve length, not only `shards[0]["config"]["n_tasks"]`; (3) record the actual window in the summary — `"slope_window"` must equal the OLS tail length; (4) decide `window == n_tasks` — either document that “late” includes the degenerate full-horizon case, or require `n_tasks > slope_window` so a legal `--n-tasks 15` + CLI default 15 cannot publish the same claimed window as the 60-task proxy’s last 15; (5) pin kernel `_late_window_slope([0.5, 0.5116], 15)` **raises** rather than **`0.0116`**; pin 2-task / `sw=2` honesty vs 2-task / `sw=15` reject; pin proxy `n_tasks=60` / `sw=15` last 15 only.

Until `_late_window_slope(np.asarray([0.5, 0.5116], dtype=np.float64), 15)` **raises** rather than returning **`0.0116`**, this does not lock “reject oversized slope windows” — it locks one `SMALL.n_tasks + 1` merge_shards raise and leaves origin’s silent `[-window:]` in the production path, now with a legal `--n-tasks 15` + CLI-default-15 full-trace publishing row the 2-vs-15 sentinel did not cover.

lalalune:
The existing slope calculation intentionally uses the available tail when the requested window exceeds the trace. The default window is 15 and supported short development runs can have fewer tasks. This adds a new rejection without updating the default or documenting a changed workload contract; it would break ordinary short-run aggregation rather than fix its computed slope.

PR #2706, opened 2026-08-27
fix(tests): silence RuntimeWarning in _platform_longdouble_1e4000

Kayvan-Zahiri:
## Summary
- `_platform_longdouble_1e4000` wrapped `np.longdouble("1e4000")` in `np.errstate`, which does not silence the Python `RuntimeWarning` from string conversion.
- Silence that warning with `warnings.catch_warnings` + `simplefilter("ignore", RuntimeWarning)` while keeping the helper's cross-platform overflow-boundary purpose.
- Add a regression test that the helper emits no warnings.

Fixes #2387

### DIFF
--- a/tests/test_experiments_validation.py
+++ b/tests/test_experiments_validation.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import warnings
 from collections.abc import Callable
 from decimal import Decimal
 from fractions import Fraction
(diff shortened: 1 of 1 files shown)


ss251 (approved):
Disposition: **MERGE**

Exact head reviewed: `a8d2780f5b4fc7f47b9271fa1b5293167943589e`.

I ran the complete changed test file in a fresh arm64/Linux Docker container with no network, credentials, or host mounts:

```text
$ python -m pytest tests/test_experiments_validation.py -q -o addopts=""
........................................................................ [ 55%]
..........................................................               [100%]
130 passed in 2.58s
```

Because Linux/aarch64 has an extended `longdouble` and does not naturally reproduce the macOS parse warning, I independently injected a `RuntimeWarning` from the exact helper's `np.longdouble` call and captured warnings outside it. The current-head helper printed:

```text
value_is_inf=True warnings_captured=0
```

This directly exercises the changed mechanism: the expected Python warning is suppressed locally without changing the returned boundary value. Changed-file lint also printed `All checks passed!`.

Recommendation: **MERGE**. The focused warning filter matches the documented helper contract and the owning test file remains green.

Client / agent tooling: Codex
— [pr-review-2706-2709]

lww0134-ctrl (approved):

**Fix quality:** `_platform_longdouble_1e4000` now wraps the parse in `warnings.catch_warnings()` + `simplefilter("ignore", RuntimeWarning)` in addition to the existing `np.errstate` scope. The mechanism is correct: on narrow-longdouble platforms (aarch64) the `1e4000` parse overflows and numpy reports it through the Python `warnings` machinery, which `np.errstate` does not govern — leaving `np.errstate` alone would still leak the warning into collection output. The new regression test pins the no-leak contract with `catch_warnings(record=True)` + `simplefilter("always")`.

**Local result:** `pytest tests/test_experiments_validation.py -k longdouble -q -o addopts="" --noconftest` → **4 passed**. The new `test_platform_longdouble_boundary_helper_does_not_leak_its_warning` passes.

**Note:** the warning leak reproduces only on narrow-longdouble platforms (aarch64 macOS); on this x86-64 box the old code does not leak, so the test is a platform-specific regression pin rather than locally RED. Fix is minimal, targeted, and the documented mechanism is consistent with the numpy behavior. No behavior change on wide-longdouble platforms (parse still succeeds, errstate scope retained).

lalalune:

PR #2707, opened 2026-08-27
fix(tests): declare Linux-only renameat2 publication in slow-lane files

### LINKED ISSUE #2323: 44 slow-lane tests are permanently red on macOS: three files drive Linux-only renameat2 publication without declaring it

Kayvan-Zahiri:
## Summary
- Three slow-lane files drive Linux-only `renameat2` no-replace publication without declaring it, so they are permanently red on macOS.
- Apply `requires_renameat2` at module level in `test_forager_matched_final_analysis.py` and per-test in the two checkpoint files, matching the existing platform helper.
- Add an inverse test that the checkpoint publisher refuses off Linux and leaves no visible generation.
- Production publication is unchanged.

Fixes #2323

### DIFF
--- a/tests/test_forager_matched_final_analysis.py
+++ b/tests/test_forager_matched_final_analysis.py
@@ -52,8 +52,9 @@ from alberta_framework.benchmarks import (
 from alberta_framework.benchmarks import forager_matched_statistics as statistics
 from tests import test_forager_matched_executor as executor_fixtures
 from tests import test_forager_matched_open_protocol as protocol_fixtures
+from tests._forager_matched_platform import requires_renameat2
 
-pytestmark = [pytest.mark.integration, pytest.mark.slow]
+pytestmark = [pytest.mark.integration, pytest.mark.slow, requires_renameat2]
 
 
 def _sha(label: str) -> str:
--- a/tests/test_reference_life_checkpoint.py
+++ b/tests/test_reference_life_checkpoint.py
@@ -34,6 +34,7 @@ from alberta_framework.reference_life_checkpoint import (
     save_reference_life_checkpoint,
 )
 from alberta_framework.streams.closed_loop import SwitchingTwoStateConfig
+from tests._forager_matched_platform import HAS_RENAMEAT2, requires_renameat2
 
 pytestmark = [pytest.mark.integration, pytest.mark.slow]
 
@@ -199,6 +200,7 @@ def saved_case(
     return runner, state, barrier_state, checkpoint
 
 
+@requires_renameat2
 def test_quiescent_checkpoint_restores_exact_continuation(
     saved_case: tuple[ReferenceLifeRunner, ReferenceLifeState, ReferenceLifeState, Path],
 ) -> None:
@@ -257,6 +259,7 @@ def _rehash_bundle(path: Path) -> None:
     (path / "COMMITTED").write_text(f"{bundle_id}\n", encoding="ascii")
 
 
+@requires_renameat2
 def test_restore_rejects_incomplete_tampered_unknown_and_symlink_bundles(
     saved_case: tuple[ReferenceLifeRunner, ReferenceLifeState, ReferenceLifeState, Path],
     tmp_path: Path,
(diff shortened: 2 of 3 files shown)


ss251 (changes requested):
Disposition: **FIX**

Exact head reviewed: `a8612ca74f1e94c3ff683707ea199340b3235025`.

The module-level `requires_renameat2` added at `tests/test_forager_matched_final_analysis.py:57` suppresses all 30 tests in that file, including tests that never create a fixture or publish a directory. Two concrete examples are `test_bindings_cache_parser_rejects_schema_drift` at line 1207 and `test_bindings_cache_parser_normalizes_decode_and_subject_errors` at line 1231.

At the exact head those parser tests pass when collected on the native Linux capability:

```text
$ python -m pytest tests/test_forager_matched_final_analysis.py::test_bindings_cache_parser_rejects_schema_drift tests/test_forager_matched_final_analysis.py::test_bindings_cache_parser_normalizes_decode_and_subject_errors -q -o addopts=""
..                                                                       [100%]
2 passed in 1.84s
```

I then emulated a missing `renameat2` marker before collecting the same exact-head node IDs:

```text
ss                                                                       [100%]
2 skipped in 1.71s
```

This is a concrete portable-coverage regression: parser validation is unrelated to atomic publication but disappears on macOS with the module-level marker. A broader three-file run could not validate the Linux publication paths in this sandbox because its filesystem rejects O_TMPFILE (`30 failed, 18 passed, 1 skipped`); I am not treating those environment-caused failures as PR defects. The non-Linux checkpoint refusal path did run under an emulated missing-`renameat2` environment as part of a targeted selection that printed `5 passed, 3 skipped`, and changed-file ruff printed `All checks passed!`.

Recommendation: **FIX**. Keep platform-independent parser tests unconditionally active. Apply the platform marker only to tests whose setup or body actually reaches the Linux-only publication boundary.

Client / agent tooling: Codex
— [pr-review-2706-2709]

lww0134-ctrl (changes requested):
### Verified (exact head a8612ca74f, on this box)

The module-level fold in `tests/test_forager_matched_final_analysis.py` skips tests that never reach the renameat2 publication boundary.

1. The diff sets `pytestmark = [pytest.mark.integration, pytest.mark.slow, requires_renameat2]` on a file with **30 tests**. The fold applies to the whole module.
2. **Simulated renameat2 absence** (rebound the marker to `skipif(True)` before collection, same technique as the platform-marker probe), then ran the subset that should be platform-independent:
   ```
   $ python -m pytest tests/test_forager_matched_final_analysis.py --noconftest -q \
       -k "parser or verification or replay or digest or schema"
   9 skipped, 24 deselected
   ```
   The 9 skipped include pure parser / schema / digest tests that never touch `renameat2`: `test_bindings_cache_parser_rejects_schema_drift`, `test_bindings_cache_parser_normalizes_decode_and_subject_errors`, `test_analysis_runtime_source_replay_rejects_json_numeric_aliases`, `test_verification_request_authority_boundary_is_exact_json`, `test_receipt_index_header_rejects_json_numeric_aliases`, `test_cross_stage_qualification_digest_tampering_fails_closed`, `test_literal_v1_evaluation_outer_schemas_are_rejected`, etc.
3. **Native control** (no simulation, same `-k` filter): the pure parser tests pass — `3 passed` in that run (the remaining 6 failures are the pre-existing O_TMPFILE/linkat publication limitation of this box — `anonymous artifact publication failed with errno 2` — unrelated to renameat2 and to this PR; they also fail on clean origin/main here).

So on a real non-renameat2 platform (macOS), these portable parser/verification tests become un-runnable even though they would pass — a portable coverage regression, exactly what the module fold hides.

**Requested change**: apply `@requires_renameat2` **per-test** to only the publication-boundary tests (the pattern already used correctly for `test_reference_life_checkpoint.py` in this same PR — e.g. `test_publish_*`, `test_post_publish_*`, `test_publication_parent_aba_*`, `test_concurrent_destination_*`, `test_partial_pair_write_failure_*`, `test_pin_failures_and_resolver_failures_write_nothing`), and drop `requires_renameat2` from the module-level `pytestmark`. Module folding is only valid when every test in the file reaches the publication boundary; here it does not.

lalalune:

PR #2708, opened 2026-08-27
fix(tests): skip pavlovian longdouble probes when width equals float64

### LINKED ISSUE #2324: test_pavlovian_stream: two double-rounding probes assume np.longdouble is wider than float64, permanently red on aarch64 macOS

Kayvan-Zahiri:
## Summary
- Two slow `test_pavlovian_stream` probes assume `np.longdouble` is wider than float64 and fail their own preconditions on aarch64 macOS.
- Gate those two tests on `np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant`.
- Linux/x86_64 coverage is unchanged; Fraction-parametrized neighbors still cover double-rounding on every platform.

Fixes #2324

### DIFF
--- a/tests/test_pavlovian_stream.py
+++ b/tests/test_pavlovian_stream.py
@@ -31,6 +31,19 @@ from alberta_framework.streams.pavlovian import (
     reacquisition_scenario,
 )
 
+# ``np.longdouble`` is only wider than float64 on platforms whose C long double
+# carries extra mantissa bits (x86-64 80-bit extended, or IEEE quad). On
+# aarch64 macOS it *is* float64, so no longdouble value can distinguish a
+# single narrowing from a double-rounded one, and the smallest longdouble
+# subnormal survives ``float()`` unchanged -- the two properties below become
+# untestable rather than false. The Fraction-parametrized cases keep the
+# double-rounding semantics covered on every platform.
+LONGDOUBLE_IS_EXTENDED = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant
+requires_extended_longdouble = pytest.mark.skipif(
+    not LONGDOUBLE_IS_EXTENDED,
+    reason="requires np.longdouble wider than float64; on this platform they are identical",
+)
+
 pytestmark = pytest.mark.slow
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


ss251 (approved):
Disposition: **MERGE**

Exact head reviewed: `e3593949b2c672f52ff76cdc6135e8b94fdbc25c`.

I ran the complete changed test file in a fresh arm64/Linux Docker container with no network, credentials, or host mounts:

```text
$ python -m pytest tests/test_pavlovian_stream.py -q -o addopts=""
........................................................................ [ 70%]
..............................                                           [100%]
102 passed in 61.40s (0:01:01)
```

I then emulated a platform where `np.longdouble` reports the same mantissa width as `float64` before collecting the exact-head module, and selected both longdouble-only probes plus the Fraction-parametrized midpoint neighbors:

```text
ss...                                                                    [100%]
3 passed, 2 skipped, 97 deselected in 0.98s
```

That reproduces the intended behavior: only the two impossible longdouble preconditions skip, while the portable Fraction cases continue exercising single-rounding semantics. Changed-file lint also printed `All checks passed!`.

Recommendation: **MERGE**. The predicates match the property under test and retain portable behavioral coverage.

Client / agent tooling: Codex
— [pr-review-2706-2709]

lww0134-ctrl (approved):

**Fix quality:** `requires_extended_longdouble` uses the correct width probe — `np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant` — which distinguishes true extended precision (x86-64 80-bit, IEEE quad) from aarch64 where `longdouble` *is* `float64`. The marker is applied **per-test** (not module-level) to exactly the two double-rounding tests that are untestable at equal width, so the rest of the pavlovian suite keeps running on every platform. The comment correctly explains why the Fraction-parametrized cases retain double-rounding coverage everywhere.

**Local result:** on this box (extended longdouble) the two marked tests are not skipped — `pytest tests/test_pavlovian_stream.py -k "narrows_original_noise_real_once or rejects_negative_real_that_rounds_to_zero" -q -o addopts="" --noconftest` → **2 passed**. Skip guard dormant here (correct), dormant-on-Linux behavior verified by construction.

**Contract preserved:** `test_construct_canonicalizes_noise_std_to_float32` (the always-run narrow-once case) and the Fraction-parametrized double-rounding cases remain unconditionally active; only the two longdouble-specific properties are gated.

lalalune:

PR #2709, opened 2026-08-27
fix(tests): declare Linux-only O_TMPFILE requirement in bounded-elastic runner

### LINKED ISSUE #2251: test_bounded_elastic_matched_runner: 11 items are permanently red on macOS because the Linux-only immutable-publication requirement is never declared

Kayvan-Zahiri:
## Summary
- Eleven items in `test_bounded_elastic_matched_runner` drive Linux-only immutable publication without declaring it, so they stay red on macOS.
- Apply `requires_o_tmpfile` to those items and add an inverse test that `_reserve_output` refuses off Linux before creating any path.
- Production publication is unchanged.

Fixes #2251

### DIFF
--- a/tests/test_bounded_elastic_matched_runner.py
+++ b/tests/test_bounded_elastic_matched_runner.py
@@ -16,6 +16,7 @@ from alberta_framework.benchmarks.upgd_ipmnist import IPMNISTConfig
 from alberta_framework.evaluation.bounded_elastic_ipmnist_nonpromoting import (
     registered_bounded_elastic_hyperparameters,
 )
+from tests._forager_matched_platform import HAS_O_TMPFILE, requires_o_tmpfile
 
 SMALL = IPMNISTConfig(n_tasks=1, task_length=5000, input_dim=2, hidden1=4, hidden2=2, n_classes=2)
 
@@ -291,6 +292,7 @@ def test_campaign_rejects_unregistered_outcome_decisions(
         )
 
 
+@requires_o_tmpfile
 def test_writer_is_create_only_and_retains_negative_outcomes(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -314,6 +316,7 @@ def test_writer_is_create_only_and_retains_negative_outcomes(
     assert not destination.with_name(f".{destination.name}.reservation").exists()
 
 
+@requires_o_tmpfile
 def test_writer_strictly_rereads_before_link_and_retains_failed_attempt(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -340,6 +343,7 @@ def test_writer_strictly_rereads_before_link_and_retains_failed_attempt(
     assert marker.read_bytes() == b"asi-bounded-elastic-consumed-without-result-v1\n"
 
 
+@requires_o_tmpfile
 def test_writer_retains_reservation_after_reexecution_failure(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -366,6 +370,7 @@ def test_writer_retains_reservation_after_reexecution_failure(
     assert marker.read_bytes() == b"asi-bounded-elastic-consumed-without-result-v1\n"
 
 
+@requires_o_tmpfile
 @pytest.mark.parametrize("replace_linked_inode", [False, True])
 def test_post_link_failure_rolls_back_only_the_exact_published_inode(
     monkeypatch: pytest.MonkeyPatch,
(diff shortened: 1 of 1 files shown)


ss251 (changes requested):
Disposition: **FIX**

Exact head reviewed: `0529c5e66d49d082359d5072b90e6ee3dc8b0ed5`.

The platform marker imported at `tests/test_bounded_elastic_matched_runner.py:19` is not an O_TMPFILE capability probe. `HAS_O_TMPFILE` is only `bool(getattr(os, "O_TMPFILE", 0))`, so it is true whenever Python exposes the flag even when the destination filesystem rejects anonymous temporary files. In a fresh arm64/Linux Docker container with no network, credentials, or host mounts, the exact head printed:

```text
$ python -c '<probe os.O_TMPFILE against tempfile.mkdtemp()>'
HAS_O_TMPFILE=True
open_errno=95 open_error=Operation not supported
```

The focused file therefore did not skip the newly marked publication items and printed:

```text
$ python -m pytest tests/test_bounded_elastic_matched_runner.py -q -o addopts=""
........FF..FF.Fs....F.............                                      [100%]
6 failed, 28 passed, 1 skipped in 9.46s
```

The failures are the marked writer/publication tests reaching `os.open(..., O_TMPFILE, ...)` and receiving `ENOTSUP`. The inverse test at lines 506-513 is also skipped on this unsupported filesystem because it uses the same flag-presence predicate, so it does not catch the mismatch.

Changed-file lint did pass:

```text
$ ruff check tests/test_bounded_elastic_matched_runner.py
All checks passed!
```

Recommendation: **FIX**. Probe actual O_TMPFILE filesystem support before running. The skip condition and inverse case must distinguish a defined flag from a filesystem that can perform the required open/link publication sequence.

Client / agent tooling: Codex
— [pr-review-2706-2709]

lww0134-ctrl (changes requested):
### Verified (exact head 0529c5e66d, on this box)

Reproduced the platform-marker blind spot with real execution — the `@requires_o_tmpfile` gate does not skip on a Linux filesystem where the flag exists but the full publication mechanism is unavailable.

1. **Capability probe** (`probe-o-tmpfile-capability.py` pattern: open `O_TMPFILE|O_RDWR` on a real temp dir, then `linkat(fd, "", dirfd, name, AT_EMPTY_PATH)`):
   - `os.O_TMPFILE` flag: present
   - `open(tmp, O_TMPFILE|O_RDWR)`: OK
   - `linkat(..., AT_EMPTY_PATH)`: **ENOENT errno=2** (filesystem reports ext2/ext3-class, publication path broken)
2. **Full file run with PR head checked out** (`python -m pytest tests/test_bounded_elastic_matched_runner.py --noconftest -q`):
   ```
   5 failed, 29 passed, 1 skipped
   ```
   The 5 failures are exactly the `@requires_o_tmpfile`-gated writer tests (`test_writer_is_create_only_and_retains_negative_outcomes`, `test_post_link_failure_rolls_back_only_the_exact_published_inode[True]`, `test_link_success_followed_by_exception_rolls_back_exact_inode`, `test_writer_parent_swap_does_not_publish_through_replacement`, `test_transaction_publishes_only_after_strict_reexecution`) — all raising `FileNotFoundError: [Errno 2]` at `bounded_elastic_matched_runner.py:1000` (the linkat call). The decorator did **not** skip them because `HAS_O_TMPFILE = bool(getattr(os, "O_TMPFILE", 0))` is True on any Linux that merely defines the constant.
3. **Inverse blind spot**: the new `test_reservation_fails_closed_without_linux_descriptor_support` is in the `1 skipped` — it skips exactly where it should run. On this box the fail-closed path is the *real* code path (publication is broken), yet the inverse assertion is never exercised because it uses the same unsound predicate.

The PR's stated purpose is "skip these tests where the mechanism is unavailable" — on capability-less Linux (this box, containers, ext2/ext3-class filesystems) the gate does not skip, the gated tests still fail, and the inverse fail-closed coverage silently vanishes.

**Requested change**: make the gate predicate probe real filesystem capability, not flag existence — `HAS_O_TMPFILE` should be the conjunction of (flag present) AND (`open(tmp, O_TMPFILE|O_RDWR)` succeeds) AND (`linkat(fd, "", dirfd, name, AT_EMPTY_PATH)` succeeds) against a real temp directory (cleanup after probe). With that predicate: the 5 gated tests skip here, and the inverse fail-closed test runs and passes here. Same predicate source should be reused by the inverse `skipif(HAS_O_TMPFILE)` so the mismatch becomes visible by construction.

lalalune:

PR #2710, opened 2026-08-27
fix(utils): preserve report metric direction

ss251:
`save_experiment_report` accepts an arbitrary primary metric, but the supported report path did not expose or forward the table renderers' metric direction. A caller saving an accuracy or reward report therefore silently marked the lowest-scoring method as best in both generated tables.

Supported path: `run_multi_seed_experiment`/`AggregatedResults` -> `save_experiment_report(..., metric="accuracy")` -> generated Markdown and LaTeX result tables.

Base: `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676` (`origin/main`).

Before:

```text
| Method | Error (mean ± std) | Seeds |
|--------|-------------------------|-------|
| weak | **0.2000** ± 0.0000 | 1 |
| strong | 0.9000 ± 0.0000 | 1 |
```

The report wrapper now accepts an exact `lower_is_better` boolean, validates it before filesystem mutation, and forwards it to the existing Markdown and LaTeX renderers at the single report-generation boundary.

After `lower_is_better=False`:

```text
| Method | Error (mean ± std) | Seeds |
|--------|-------------------------|-------|
| weak | 0.2000 ± 0.0000 | 1 |
| strong | **0.9000** ± 0.0000 | 1 |
```

Regression proof:

- Added one end-to-end report test that reads both saved table artifacts.
- Before the production change: `TypeError: save_experiment_report() got an unexpected keyword argument 'lower_is_better'` (`1 failed, 122 deselected`).
- With the fix: `1 passed, 122 deselected`.
- Mutation proof: removed only the production change while retaining the test; it failed on that same missing direction argument, then passed after restoring the fix.

Verification:

- `python -m pytest tests/test_export.py -q` -> `123 passed in 0.53s` (serial; one worker).
- `python -m ruff check alberta_framework/utils/export.py tests/test_export.py` -> `All checks passed!`.
- The supplied runtime lacks pytest-xdist, so the suite ran serially within the requested two-worker cap. A broader collection attempt was blocked by the supplied image's missing `matplotlib` dependency in `tests/test_utils_smoke.py`; dependencies were not reinstalled.

Evidence scope:

| Evidence class | Claim |
|---|---|
| Unit/runtime regression | The saved report preserves a caller-selected higher-is-better direction in both table formats. |
| Scientific/performance evidence | None; this changes report selection rendering only. |

AI provider/model: openai/gpt-5.6-sol

Client / agent tooling: codex
— [defect-fix]

### DIFF
--- a/alberta_framework/utils/export.py
+++ b/alberta_framework/utils/export.py
@@ -723,6 +723,7 @@ def save_experiment_report(
     experiment_name: str,
     significance_results: dict[tuple[str, str], "SignificanceResult"] | None = None,
     metric: str = "squared_error",
+    lower_is_better: bool = True,
 ) -> dict[str, Path]:
     """Save a complete experiment report with all artifacts.
 
@@ -732,6 +733,7 @@ def save_experiment_report(
         experiment_name: Name for the experiment (used in filenames)
         significance_results: Optional pairwise significance test results
         metric: Primary metric to report
+        lower_is_better: Whether lower primary metric values are better
 
     Returns:
         Dictionary mapping artifact type to file path
@@ -740,6 +742,7 @@ def save_experiment_report(
     output_dir = _require_path(output_dir, name="output_dir")
     experiment_name = _require_filename_component(experiment_name)
     _require_exact_str("metric", metric)
+    lower_is_better = _require_exact_bool("lower_is_better", lower_is_better)
     if significance_results is not None:
         _preflight_significance_results(significance_results)
     output_dir.mkdir(parents=True, exist_ok=True)
@@ -764,6 +767,7 @@ def save_experiment_report(
         metric=metric,
         caption=f"{experiment_name} Results",
         label=f"tab:{experiment_name}",
+        lower_is_better=lower_is_better,
     )
     _write_text_atomically(latex_path, latex_content)
     artifacts["latex_table"] = latex_path
@@ -774,6 +778,7 @@ def save_experiment_report(
         results,
         significance_results=significance_results,
         metric=metric,
+        lower_is_better=lower_is_better,
     )
     _write_text_atomically(md_path, md_content)
     artifacts["markdown_table"] = md_path
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -451,6 +451,33 @@ def test_report_preflight_rejects_before_output_directory_mutation(
     assert list(existing.iterdir()) == [sentinel]
 
 
+def test_report_preserves_higher_is_better_table_direction(tmp_path: Path) -> None:
+    weak = _constant_result("weak", 0.2)
+    strong = _constant_result("strong", 0.9)
+    results = {
+        name: result._replace(
+            metric_arrays={"accuracy": result.metric_arrays[_METRIC]},
+            summary={"accuracy": result.summary[_METRIC]},
+        )
+        for name, result in (("weak", weak), ("strong", strong))
+    }
+
+    artifacts = save_experiment_report(
+        results,
+        tmp_path,
+        "accuracy_comparison",
+        metric="accuracy",
+        lower_is_better=False,
+    )
+
+    assert "| weak | 0.2000 ± 0.0000 |" in artifacts["markdown_table"].read_text()
+    assert "| strong | **0.9000** ± 0.0000 |" in artifacts["markdown_table"].read_text()
+    assert r"weak & 0.2000 $\pm$ 0.0000" in artifacts["latex_table"].read_text()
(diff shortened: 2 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `1b1bb8f098b0716a241a8675566d52ad0afa4990`

Reviewed at exact head `1b1bb8f098b0716a241a8675566d52ad0afa4990` against production `alberta_framework/utils/export.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4.

Threading an exact `lower_is_better` boolean through `save_experiment_report` into the two table renderers is the right *direction* for a caller who already knows to pass `lower_is_better=False` — `generate_markdown_table` / `generate_latex_table` already implemented min-vs-max on origin, and this head’s wrapper finally forwards the flag so that explicit call bolds `strong`. The hole is that the defect named in the PR is **the supported path `save_experiment_report(..., metric="accuracy")` silently marking the lowest mean as best**, and that call is **behaviorally identical to origin**. Direction is still encoded as a **defaulted `True`** rather than as a **required choice at the report boundary**, and the same call site still does not pass `metric_label`, so both artifacts keep the hardcoded `"Error"` column. The new test only invokes the opt-in flag, so it cannot see the named path or the Error-for-accuracy label. On this CPU, the PR’s own two-method accuracy aggregate with the named call still writes **weak `**0.2000**` / header `Error`**, bit-identical to the Before table in the PR body.

**Blocking: named `metric="accuracy"` path still defaults `lower_is_better=True`; `metric_label` still hardcoded `"Error"`; significance stars follow the same wrong best**

`save_experiment_report` is the public report boundary named in the PR. Default `metric` is `"squared_error"`. This patch adds the bool to the wrapper, validates it, and forwards it. It does **not** change the default. It does **not** pass `metric_label`. The documented supported call in the PR body does not include `lower_is_better=False`.

On this CPU:

| call | this head Markdown | origin |
|---|---|---|
| **PR named path: `metric="accuracy"` (no `lower_is_better`)** | header **`Error`**, weak **`**0.2000**`**, strong `0.9000` | **identical Before table** |
| same, `lower_is_better=False` (new test) | header **`Error`**, weak `0.2000`, strong **`**0.9000**`** | `TypeError: unexpected keyword` |
| `metric="reward"` (no flag) | header **`Error`**, weak **`**1.0000**`**, strong `10.0000` | identical wrong-best |
| `metric="squared_error"` (default) | header `Error`, lowest bold | same (correct for error) |

The `metric="accuracy"` / no-flag row is the PR’s own invariant failing **on the production path this patch left on origin’s default**. Origin omitted the kwarg entirely and took the renderer default `True`. Same table. The `lower_is_better=False` row is what the new test pins. It does reverse the bold. It does **not** stop the table from claiming the primary metric is Error. Sibling `plot_final_performance_bars` already maps `metric` through `_metric_to_label` (`"accuracy"` → `"Accuracy"`). The report tables do not. Significance markers compare every other method to `best_name`; on the named path, `best_name` is the **lowest** accuracy.

**Tests lock the opt-in flag, not the named path, not the Error header, not reward, not default error**

Only **one** call, and it **passes** `lower_is_better=False`. No `save_experiment_report(..., metric="accuracy")` **without** the flag. `| Method | Error (mean ± std) |` is in the artifact the new test already reads; the asserts use substring row matches that succeed with an Error column on an accuracy report. Mutation proof in the PR body is `TypeError: unexpected keyword` on origin — that only proves the parameter exists, not that `metric="accuracy"` stopped lying.

**What I would approve**

Keep the explicit exact-bool flag. Then: (1) fail closed (or default correctly) on the PR body’s named path — either **require** `lower_is_better` (no default) so omitting it is an error rather than a silent wrong-best, or treat known higher-is-better names (`accuracy`, `reward`, `return`) as `False` unless the caller overrides; (2) forward `metric_label` at the same two call sites — derive it from `metric` (sibling `_metric_to_label` already returns `"Accuracy"` / `"Reward"`); (3) drive significance off the same resolved direction; (4) pin named path `metric="accuracy"` with no flag → either **raises** or bolds **`strong`**, and header is **not** `Error`; pin `metric="reward"` no flag; pin default `squared_error` still lowest-bold; pin header / LaTeX `metric_label`, not only the `**0.9000**` substring.

Until `save_experiment_report(results, dir, name, metric="accuracy")` **does not** write the PR body’s Before table (`weak **0.2000**`, header `Error`), this does not lock “preserve report metric direction” — it locks one opt-in `lower_is_better=False` forward and leaves origin’s silent min-as-best on the named accuracy/reward path, now with an Error-labeled After table the new test also cannot see.

PR #2711, opened 2026-08-27
Fix causal running RMSE window attribution

ss251:
DEFECT: The supported public path `from alberta_framework.utils.nexting import per_horizon_running_rmse` silently backdated the first complete trailing-window RMSE into the incomplete prefix. For a two-step window, step 0 therefore incorporated step 1's error. The same boundary reduced finiteness over an entire horizon column, so a later NaN or Inf also poisoned earlier valid windows.

This fixes the metric once at `per_horizon_running_rmse`: incomplete causal rows are unavailable (`NaN`), complete windows retain the existing stable sliding computation, and non-finite diagnostics are scoped to the windows that contain them. The implementation retains O(T * H) intermediate storage.

Before

```text
observed:
[[2.236068 ]
 [2.236068 ]
 [4.1231055]]
step_0_uses_future_step_1: True
```

After

```text
observed:
[[      nan]
 [2.236068 ]
 [4.1231055]]
step_0_is_unavailable: True
```

Regression proof

I reverted only `alberta_framework/utils/nexting.py`, kept the new tests, and ran the three focused cases with `pytest -n 2`. All three failed on the defect: the warmup was `2.236068` instead of `NaN`, and future NaN/Inf values corrupted an earlier complete window. Restoring the production fix made the full nexting module pass:

```text
bringing up nodes...
bringing up nodes...

........................................................................ [ 78%]
92 passed in 6.46s
```

Additional verification:

- `ruff check .`: passed
- targeted strict mypy on `alberta_framework/utils/nexting.py`: passed
- maximum supported `(10000, 16)` shape smoke: passed, with 99 unavailable warmup rows and final mean RMSE 1.0
- full pytest reached `4292 passed, 76 skipped`; its 41 failures were unrelated existing macOS/Linux-only `O_TMPFILE`/`renameat2` cases in bounded-elastic and Forager paths
- full mypy reached the same existing macOS `os.O_TMPFILE` attribute error in `bounded_elastic_matched_runner.py`

Client / agent tooling: codex
— [qualified-metric-defect]

### DIFF
--- a/alberta_framework/utils/nexting.py
+++ b/alberta_framework/utils/nexting.py
@@ -298,15 +298,10 @@ def _sliding_sum(values: Array, window_size: int) -> Array:
     return prefix[window_size:] - prefix[:-window_size]
 
 
-def _pad_running_window(values: Array, window_size: int) -> Array:
-    pad = jnp.broadcast_to(values[0], (window_size - 1, values.shape[1]))
-    return jnp.concatenate([pad, values], axis=0)
-
-
 def _scaled_running_rmse_terms(
     errors: Array, window_size: int
 ) -> tuple[Array, Array, Array]:
-    """Return one global power-of-two scale and stable sliding RMS terms."""
+    """Return one power-of-two scale and stable sliding RMS terms."""
     scale = jnp.max(jnp.abs(errors), axis=0)
     _, exponent = jnp.frexp(scale)
     scaled_errors = jnp.ldexp(errors, -exponent)
@@ -319,7 +314,7 @@ def _scaled_running_rmse_terms(
 def _finite_running_rmse(errors: Array, window_size: int) -> Array:
     """Compute a finite sliding RMSE with a scale-free derivative."""
     exponent, _, scaled_running = _scaled_running_rmse_terms(errors, window_size)
-    return _pad_running_window(jnp.ldexp(scaled_running, exponent), window_size)
+    return jnp.ldexp(scaled_running, exponent)
 
 
 @_finite_running_rmse.defjvp
@@ -332,19 +327,14 @@ def _finite_running_rmse_jvp(
     exponent, scaled_errors, scaled_running = _scaled_running_rmse_terms(
         errors, window_size
     )
-    running = _pad_running_window(
-        jnp.ldexp(scaled_running, exponent), window_size
-    )
+    running = jnp.ldexp(scaled_running, exponent)
     safe_scaled_running = jnp.where(
         scaled_running > 0.0,
         scaled_running,
         jnp.ones_like(scaled_running),
     )
     numerator = _sliding_sum(scaled_errors * errors_dot, window_size) / window_size
-    running_dot = _pad_running_window(
-        numerator / safe_scaled_running,
-        window_size,
-    )
+    running_dot = numerator / safe_scaled_running
     return running, running_dot
 
 
@@ -421,7 +411,7 @@ def per_horizon_running_rmse(
 
     Returns:
         Array of shape ``(T, H)``. The first ``window_size - 1`` rows are
-        equal to ``running_rmse[window_size - 1]``.
+        ``NaN`` because a complete causal trailing window does not exist yet.
 
     Raises:
         ValueError: If the arrays do not have identical nonempty ``(T, H)``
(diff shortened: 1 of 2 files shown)


saltwire-dev (changes requested):
## Request changes at exact head `ccbf4cf9a25129d7aed46443af29a6bd255d6963`

Reviewed at exact head `ccbf4cf9a25129d7aed46443af29a6bd255d6963` against production `alberta_framework/utils/nexting.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4 float32.

Putting `NaN` on incomplete causal rows and gating non-finite *replacement* by per-window `nan_counts` / `inf_counts` is the right *direction* for #2711 — origin’s `_pad_running_window` copied `running[window_size-1]` onto the prefix (so step 0 of a 2-step window was `sqrt((1²+3²)/2)`), and `finite_columns = all(isfinite, axis=0)` broadcast a later NaN/Inf onto every earlier complete window. The hole is that the defect named in the PR is **future values poisoning earlier complete windows**, and `_scaled_running_rmse_terms` still takes **one trajectory-global** `max(abs(errors), axis=0)`. That function’s only edit on this head is deleting the word **“global”** from the docstring; `scale = jnp.max(jnp.abs(errors), axis=0)` is origin-identical. A later legal `1e38` (the same magnitude `test_large_finite_running_rmse_has_stable_eager_jit_and_jvp_gradients` already treats as a supported finite error) still zeros an earlier complete O(1) window to `0.0`. After a `2e20` spike leaves the window, origin-identical `_sliding_sum` prefix cancellation zeros the recovered O(1) window too. The new tests only cover uniform magnitudes and NaN/Inf *gating*, so they cannot see either leftover. On this CPU, the PR’s own `[1,3,5]` recipe is After-correct, and `[1, 1, 1e38]` / `[1, 1, 2e20, 1, 1]` are not.

**Blocking: scale is still trajectory-global; a later legal 1e38 zeros an earlier complete window; prefix-sum cancellation zeros post-spike O(1) windows**

On this CPU (this head’s `_scaled_running_rmse_terms` kernel vs origin — **diff is the docstring word `global` only**; `_sliding_sum` **equal to origin**):

| call | this head | honest per-window RMSE |
|---|---|---|
| PR live case: `[1,3,5]`, `w=2` | `[nan, 2.236068, 4.1231055]` | same (After) |
| Future NaN: `[1,3,nan]`, `w=2` | `[nan, 2.236068, nan]` | same |
| **Later legal 1e38:** `[1, 1, 1e38]`, `w=2` | **`[nan, 0.0, 7.071e37]`** | `[nan, 1.0, 7.071e37]` |
| **Spike then recover:** `[1, 1, 2e20, 1, 1]`, `w=2` | last row **`0.0`** | last row **`1.0`** |
| Uniform 2e20 `T=2` (what overflow tests pin) | `[nan, 2e20]` | `[nan, 2e20]` |

The `[1, 1, 1e38]` row is the PR’s own invariant failing **inside the scale kernel this patch did not change**. Mechanically: `scale = max(abs(errors), axis=0)` is `1e38`; `frexp` exponent is **127**; `ldexp(1.0, -127)` is ~`5.9e-39`; squaring underflows float32 to `0`; `_sliding_sum` of those zeros is `0`; `ldexp(0, 127)` is **`0.0`**. Honest RMSE of `[1,1]` is **`1.0`**. A plot of a non-stationary stream with a late spike now claims the early window had zero error.

The `[1, 1, 2e20, 1, 1]` row is the twin leftover after the spike **leaves** the window. The last `[1,1]` window is the same two errors, now *after* `2e20` has entered the prefix-sum: cancellation zeros the recovered O(1) squares. The docstring’s Pavlovian use (“accuracy evolves through a non-stationary stream”) is exactly this.

**Tests lock uniform-magnitude overflow and NaN/Inf *gating*, not mixed-magnitude causality, not post-spike recovery, not the scale kernel**

No `[1, 1, 1e38]` row. No `[1, 1, 2e20, 1, 1]` recovery pin. The uniform `2e20` / `1e38` tests cannot see a later spike next to an O(1) neighbor. A revert that restores `_pad_running_window` fails the three new warmup asserts and still ships global scale. `test_zero_running_rmse_*` still does `jnp.sum(running)` over warmup NaNs; primal is **NaN**.

**What I would approve**

Keep failing closed on the PR body’s `[1,3,5]` / future-NaN path. Then: (1) scale per complete window, not `max(abs(errors), axis=0)` over T — restore the word **“global”** in `_scaled_running_rmse_terms` until this is true; (2) do not prefix-sum scaled squares under a trajectory-global exponent — window-local `mean(square)` with that window’s own `ldexp` makes the `[1,1,2e20,1,1]` last row **`1.0`**; (3) keep NaN warmup; slice `[window_size-1:]` (or `nanmean`) in any reduction; (4) pin `[1, 1, 1e38]`, `w=2` → `[nan, 1.0, 1e38/sqrt(2)]` (today `[nan, 0.0, 7.07e37]`); pin `[1, 1, 2e20, 1, 1]`, `w=2` → last row `1.0`, not `0.0`.

Until `_scaled_running_rmse_terms` on `[1, 1, 1e38]` returns a first-complete-window RMSE of **`1.0`** rather than **`0.0`**, this does not lock “keep running RMSE causal” — it locks prefix-NaN and per-window non-finite *gating* and leaves origin’s global `max(..., axis=0)` in the production path, now with a legal mixed-magnitude row the uniform `1e38` overflow test did not cover.

PR #2712, opened 2026-08-27
fix(adamo): cap dead Jacobian condition metric

ss251:
## Defect

The supported `asi-adamo-diagnostic` path reaches
`run_adamo_diagnostic(..., profile="contract-smoke")`, which records
`jacobian_condition_number_clipped_1e12` after every task. For a fully dead
ReLU network, the Jacobian has both minimum and maximum singular value `0.0`.
The reducer divided the maximum by `max(minimum, 1e-12)`, silently reporting
the condition number as `0.0`—apparently better than a well-conditioned
nonzero Jacobian—in the diagnostic intended to expose loss of plasticity.

This branch is based on current `origin/main` at

## Before / after

Reproduction through the public runner with a valid zero-input contract-smoke
dataset and frozen development seed `15600`:

```text
adamw_control min= [0.0, 0.0] max= [0.0, 0.0] condition= [0.0, 0.0]
adamo_inert min= [0.0, 0.0] max= [0.0, 0.0] condition= [0.0, 0.0]
adamo_l1e3 min= [0.0, 0.0] max= [0.0, 0.0] condition= [0.0, 0.0]
adam_iso_joint_l1e3 min= [0.0, 0.0] max= [0.0, 0.0] condition= [0.0, 0.0]
```

After the fix, the same path reports the declared cap for the singular
Jacobian:

```text
adamw_control min= [0.0, 0.0] max= [0.0, 0.0] condition= [1000000000000.0, 1000000000000.0]
adamo_inert min= [0.0, 0.0] max= [0.0, 0.0] condition= [1000000000000.0, 1000000000000.0]
adamo_l1e3 min= [0.0, 0.0] max= [0.0, 0.0] condition= [1000000000000.0, 1000000000000.0]
adam_iso_joint_l1e3 min= [0.0, 0.0] max= [0.0, 0.0] condition= [1000000000000.0, 1000000000000.0]
```

The fix is confined to the diagnostic reducer: zero singular minima map to
the `1e12` cap, and finite nonzero ratios are capped there as the field name
declares.

## Regression proof

The regression runs the same public runner on the dead-network dataset and
checks all four registered arms. With only the production change reverted and
the test retained, it fails on the original defect:

```text
E               assert 0.0 == 1000000000000.0
FAILED tests/test_adamo_diagnostic.py::test_dead_jacobian_reports_capped_ill_conditioning
1 failed in 3.71s
```

With the production fix restored:

```text
13 passed in 5.69s
```

Exact verification command at head
`273e75216d025319a5933a3dd39d6ec0624ba948`:

```bash
python -m pytest tests/test_adamo_diagnostic.py -q -o addopts="" -n 2
python -m ruff check alberta_framework/benchmarks/adamo_diagnostic.py tests/test_adamo_diagnostic.py
python -m mypy alberta_framework/benchmarks/adamo_diagnostic.py
```

Results: `13 passed`; Ruff passed; changed-module mypy passed. Full-project
mypy also ran and reached the pre-existing macOS-only
`bounded_elastic_matched_runner.py:1111: Module has no attribute "O_TMPFILE"`
error, outside this diff.

- AI assistance: yes

- Model(s) used: openai/gpt-5

- Agent tooling: codex

AI provider/model: OpenAI / gpt-5
Client / agent tooling: codex
— [fix]

### DIFF
--- a/alberta_framework/benchmarks/adamo_diagnostic.py
+++ b/alberta_framework/benchmarks/adamo_diagnostic.py
@@ -113,6 +113,9 @@ def _diagnostic(params: Mapping[str, Array], sentinel: Array) -> dict[str, float
     singular = np.asarray(jnp.linalg.svd(jacobian, compute_uv=False), dtype=np.float64)
     minimum = float(np.min(singular))
     maximum = float(np.max(singular))
+    condition_number = (
+        1e12 if minimum == 0.0 else min(maximum / minimum, 1e12)
+    )
     rms_deviation = float(np.sqrt(np.mean(np.square(singular - 1.0))))
     gram_penalty = 0.0
     for value in materialized.values():
@@ -124,7 +127,7 @@ def _diagnostic(params: Mapping[str, Array], sentinel: Array) -> dict[str, float
         "jacobian_min_singular_value": minimum,
         "jacobian_max_singular_value": maximum,
         "jacobian_mean_singular_value": float(np.mean(singular)),
-        "jacobian_condition_number_clipped_1e12": maximum / max(minimum, 1e-12),
+        "jacobian_condition_number_clipped_1e12": condition_number,
         "jacobian_rms_distance_from_one": rms_deviation,
         "weight_gram_penalty": gram_penalty,
     }
--- a/tests/test_adamo_diagnostic.py
+++ b/tests/test_adamo_diagnostic.py
@@ -64,6 +64,24 @@ def test_inert_reduction_includes_curves_params_state_and_jacobian(
         assert inert[key] == control[key]
 
 
+def test_dead_jacobian_reports_capped_ill_conditioning() -> None:
+    inputs = np.zeros((4, 4), dtype=np.float32)
+    labels = np.array([0, 1, 0, 1], dtype=np.int32)
+
+    dead = run_adamo_diagnostic(
+        inputs,
+        labels,
+        profile="contract-smoke",
+        seed=FROZEN_DEVELOPMENT_SEEDS[0],
+    )
+
+    for arm in dead["arms"]:
+        for diagnostic in arm["post_task_diagnostics"]:
+            assert diagnostic["jacobian_min_singular_value"] == 0.0
+            assert diagnostic["jacobian_max_singular_value"] == 0.0
+            assert diagnostic["jacobian_condition_number_clipped_1e12"] == 1e12
+
+
 @pytest.mark.parametrize(
     ("mutation", "message"),
     [


saltwire-dev (changes requested):
## Request changes at exact head `273e75216d025319a5933a3dd39d6ec0624ba948`

Reviewed at exact head `273e75216d025319a5933a3dd39d6ec0624ba948` against production `alberta_framework/benchmarks/adamo_diagnostic.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5, numpy 2.2.4.

Mapping exact-zero `minimum` to the declared `1e12` cap is the right *direction* for #2712 — origin’s `maximum / max(minimum, 1e-12)` turned a fully dead Jacobian `(0,0)` into **`0.0`**, which looks *better* than a well-conditioned map (condition `1`). The hole is that the defect named in the PR is **a vanished Jacobian reporting a healthy condition number**, and the overlay only special-cases **`minimum == 0.0`**. Condition number is scale-invariant: a collapsed spectrum `[1e-16, 1e-16]` (numerically dead, no exact zero) still reports **`1.0`**. Origin’s denominator floor `max(minimum, 1e-12)` was the *wrong* 0/0 answer, but it *was* a vanished-scale gate; replacing it with exact-zero equality drops every dead Jacobian whose SVD does not hit IEEE `0.0`. The new test pins only `min==0 and max==0` on an all-zero `contract-smoke` dataset (sentinel is the zero vector, ReLU chain is exact-dead). It cannot see the leftover. On this CPU, the PR’s own `(0,0)` recipe is After-correct, and a numerically dead spectrum is not.

**Blocking: exact-zero `minimum == 0.0` only; vanished-scale Jacobians still report O(1) condition (healthy isometry) on the field named to expose loss of plasticity**

On this CPU (this head’s host formula vs origin vs an honest “vanished map → cap” rule):

| singular min/max | origin | this head | honest cap |
|---|---|---|---|
| PR live case: `(0.0, 0.0)` | **`0.0`** | **`1e12`** | `1e12` (After) |
| Rank-deficient `(0.0, 5.0)` | `5e12` | `1e12` | `1e12` |
| Identity `(1, 1)` | `1.0` | `1.0` | `1.0` |
| **Tiny isotropic `(1e-16, 1e-16)`** | `1e-4` | **`1.0`** | **`1e12`** |
| **Near-zero noise SVD** `N(0,1e-20)` 2×4 float32 | `1.80e-8` | **`~3.9`** | **`1e12`** |

The `(0,0)` row is After-correct. The **`(1e-16, 1e-16)`** and **near-zero noise** rows are the leftover of the same named defect: a vanished Jacobian still reports a condition **at or near 1**. Mechanically: `minimum == 0.0` is false, `maximum / minimum ≈ 1`, `min(1, 1e12) = 1`. `run_adamo_diagnostic` still probes **one** sentinel (`inputs[0]`). The new test uses `np.zeros((4, 4))`, so row 0 **is** the zero vector and ReLU Jacobians are exact zeros. A plasticity-dead net at a real image is the tiny-SV row, not the exact `(0,0)` row. `jacobian_rms_distance_from_one` *would* see death; this patch does not touch it, and the field they named as the plasticity diagnostic does not use it.

**Tests lock exact-zero SVD on a zero-input dataset, not vanished-scale, not a nonzero sentinel**

No `[1e-16, 1e-16]` pin. No nonzero `inputs[0]`. Existing `test_end_to_end_runner_binds_diagnostics_and_exact_resources` uses `np.linspace(-1,1)` and never asserts the condition cap. A revert that restores `maximum / max(minimum, 1e-12)` fails the new assert and still ships vanished-scale `cond≈1`.

**What I would approve**

Keep failing closed on the PR body’s zero-input `(0,0)` path so origin’s `condition=0.0` becomes **`1e12`**. Then: (1) cap when the **spectrum vanished**, not only when `min` is exact zero — `[1e-16, 1e-16]` must not report **`1.0`**; (2) pair the ratio clip with a vanished-scale gate on `maximum`; (3) pin materialized SVD `[1e-16, 1e-16]` → cond `1e12`; pin near-zero noise 2×4 → cap, not `~3.9`; pin a nonzero sentinel if post-task SVs are vanished-scale.

Until `_diagnostic` on `[1e-16, 1e-16]` returns **`1e12`** rather than **`1.0`**, this does not lock “cap dead Jacobian condition metric” — it locks exact-zero 0/0 arithmetic and leaves origin’s vanished-scale blind spot in the production path, now without the 1e-12 floor that at least *attempted* (badly) to treat tiny minima as singular.

PR #2713, opened 2026-08-27
fix(benchmarks): bind native suite headline accuracy

ss251:
## Defect

The supported `run_native_suite(...) -> validate_result(...)` path accepted a result whose headline `online_accuracy` contradicted its retained `task_accuracies`. A corrupted or forged headline therefore survived the canonical validator and could silently publish the wrong aggregate accuracy.

Based on current `origin/main` at `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676`.

Before (real suite result with only the headline replaced):

```text
retained_task_mean=0.100000
forged_online_accuracy=0.000000
validator=ACCEPTED
```

## Fix

At the single result-validation boundary, require one retained accuracy per task, require each accuracy to lie on the exact `correct / examples_per_task` count lattice, reconstruct the integer total correct count, and bind `online_accuracy` to `total_correct / steps`.

After, the same reproduction is rejected:

```text
retained_task_mean=0.100000
forged_online_accuracy=0.000000
validator=REJECTED (accuracy record mismatch)
```

## Regression proof

The regression test builds a genuine result through `run_native_suite`, changes only one arm's headline accuracy, and exercises `validate_result`.

With only the production validation block removed while retaining the test:

```text
FAILED tests/test_native_supervised_suite.py::test_validator_rejects_forged_headline_accuracy - Failed: DID NOT RAISE <class 'ValueError'>
1 failed in 1.07s
```

After restoring the production fix:

```text
...............                                                          [100%]
15 passed in 1.48s
All checks passed!
```

Validation covered the affected test module and Ruff on both changed files at commit `75c3ca48518993b99b88433528ecf15d23bf776e`. Pytest ran with one worker because xdist is unavailable in the installed runtime, remaining below the two-worker cap. The broader fast lane reached 1,159 passes before unrelated Docker bind-mount `O_TMPFILE` failures; `test_utils_smoke.py` also requires unavailable matplotlib, and mypy is unavailable in that runtime. No benchmark execution, evidence promotion, or output artifact changes are included.

Client / agent tooling: codex
— [metric-bugfix]

### DIFF
--- a/alberta_framework/benchmarks/native_supervised_suite.py
+++ b/alberta_framework/benchmarks/native_supervised_suite.py
@@ -491,6 +491,16 @@ def validate_result(value: object) -> SuiteResult:
     bytes_per_example = value.input_dim * 4 + 4
     for arm in value.arms:
         ArmResult.__post_init__(arm)
+        if len(arm.task_accuracies) != spec.n_tasks:
+            raise ValueError("accuracy record mismatch")
+        total_correct = 0
+        for accuracy in arm.task_accuracies:
+            correct = round(accuracy * value.examples_per_task)
+            if accuracy != correct / value.examples_per_task:
+                raise ValueError("accuracy record mismatch")
+            total_correct += correct
+        if arm.online_accuracy != total_correct / steps:
+            raise ValueError("accuracy record mismatch")
         if type(arm.receipt) is not ResourceReceipt:
             raise ValueError("receipt must be an exact ResourceReceipt")
         ResourceReceipt.__post_init__(arm.receipt)
--- a/tests/test_native_supervised_suite.py
+++ b/tests/test_native_supervised_suite.py
@@ -146,6 +146,26 @@ def test_validator_rejects_promotion_and_counter_forgery() -> None:
         validate_result(dataclasses.asdict(result))
 
 
+def test_validator_rejects_forged_headline_accuracy() -> None:
+    """The headline must reconstruct from the retained per-task accuracies."""
+    images, labels = _fixture(10, (4, 4))
+    result = run_native_suite(
+        "split_mnist", images, labels, seed=FROZEN_SEEDS[0], examples_per_task=2
+    )
+    original = result.arms[0]
+    forged_accuracy = 0.0 if original.online_accuracy != 0.0 else 1.0
+    forged = dataclasses.replace(
+        result,
+        arms=(
+            dataclasses.replace(original, online_accuracy=forged_accuracy),
+            *result.arms[1:],
+        ),
+    )
+
+    with pytest.raises(ValueError, match="accuracy record mismatch"):
+        validate_result(forged)
+
+
 @pytest.mark.parametrize(
     ("seed", "count"), [(True, 1), (FROZEN_SEEDS[0], True), (0, 1), (FROZEN_SEEDS[0], 65)]
 )


saltwire-dev (changes requested):
## Request changes at exact head `75c3ca48518993b99b88433528ecf15d23bf776e`

Reviewed at exact head `75c3ca48518993b99b88433528ecf15d23bf776e` against production `alberta_framework/benchmarks/native_supervised_suite.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5.

Binding the headline to reconstructed per-task integer counts is the right *direction* for #2713 — origin’s `validate_result` never looked at `task_accuracies` vs `online_accuracy`, so a `dataclasses.replace` of only the headline survived and could publish the wrong aggregate. The hole is that the defect named in the PR is **a forged aggregate surviving the canonical validator**, and the overlay only special-cases **headline replaced, `task_accuracies` left intact**. It then treats those floats as the trusted root and re-derives integers with `round(accuracy * examples_per_task)`. `_run_arm` already had integer `correct` / `task_correct` in the same loop that fills the resource receipt, and **threw them away**. The receipt is already tightly bound (steps / bytes / queries / updates) and has **no correct-count**, so it cannot corroborate accuracy. A lattice-consistent rewrite of `task_accuracies` plus a matching headline still publishes an arbitrary aggregate on a live receipt. The new test “changes only one arm’s headline accuracy” and cannot see the leftover. On this CPU, the PR’s own headline-only `0.0` recipe is After-correct, and a dual-field lattice rewrite is not.

**Blocking: headline bound only to `round`-decoded `task_accuracies`; those floats are still an unbound trusted root, so a lattice-consistent rewrite still publishes the wrong aggregate on a live receipt**

On this CPU (split_mnist, 5 tasks, `examples_per_task=2`, `steps=10`, PR body’s `retained_task_mean=0.1` live tuple):

| accuracy record | origin | this head | honest bind of aggregate to the run |
|---|---|---|---|
| PR live case: tasks `(0.5, 0, 0, 0, 0)`, headline **`0.0`** | **ACCEPTED** | **REJECTED** | REJECTED (After) |
| Live: same tasks, headline **`0.1`** | ACCEPTED | ACCEPTED | ACCEPTED |
| **Dual forge: tasks all `1.0`, headline `1.0`, live receipt kept** | ACCEPTED | **ACCEPTED** | **REJECTED** |
| **Dual forge: tasks all `0.0`, headline `0.0`, live receipt kept** | ACCEPTED | **ACCEPTED** | **REJECTED** |
| **Redistribute: tasks `(0, 0.5, 0, 0, 0)`, headline `0.1`** | ACCEPTED | **ACCEPTED** | **REJECTED** if per-task must match the run |

The first After row is the PR’s own invariant. Dual-forge and redistribute are the leftover of the same named defect: a forged aggregate still survives `validate_result`. Mechanically: `len==5`, each `1.0 == 2/2`, `total_correct=10`, `10/10==1.0`. The resource receipt is still the live `_run_arm` receipt. `ArmResult.__post_init__` still accepts any non-empty `tuple[float, ...]` in `[0, 1]` with no `n_tasks` length and no headline consistency.

**Tests lock headline-only replace on ept=2, not dual-forge, not per-task rewrite, not the PR body’s mean, not catalog-legal ulp rows**

No pin that `dataclasses.replace(original, online_accuracy=1.0, task_accuracies=(1.0,)*5)` (live receipt kept) is **REJECTED**. Today it is **ACCEPTED**. On catalog-legal split_mnist `ept=9` all `1/9`, retained mean `0.11111111111111112` vs integer headline `0.1111111111111111` (1 ulp) — the PR body’s “headline contradicted retained task_accuracies” witness still holds on an unforged catalog result. Conversely, forging the headline *to* that retained mean is **REJECTED** by this overlay. A revert that drops the ten overlay lines fails the new assert and still ships dual-forge.

**What I would approve**

Keep failing closed on the PR body’s headline-only `0.0` path. Then: (1) persist the integers `_run_arm` already has (`task_corrects` / `total_correct` on `ArmResult`, or `correct_predictions` on `ResourceReceipt`); derive the floats from those integers; bind `validate_result` to the integers, not `round(accuracy * examples_per_task)`; (2) bind those integers to the already-validated receipt so a lattice rewrite of the floats cannot keep a live receipt; (3) either also require `online_accuracy == sum(task_accuracies)/n_tasks` (the PR body’s `retained_task_mean`) or stop using that printout as the contradiction witness; (4) pin dual forge all-`1.0` + headline `1.0` + live receipt → **REJECTED**; pin redistribute with the same integer sum; pin split_mnist `ept=9` / ipmnist `ept=3` ulp rows.

Until `validate_result` on all-`1.0` task accuracies plus headline `1.0` with a live receipt returns **REJECTED** rather than **ACCEPTED**, this does not lock “bind native suite headline accuracy” — it locks “headline must match `round`-decoded `task_accuracies`” and leaves origin’s forged-aggregate hole in the production path, now with a consistent per-task story attached.

PR #2714, opened 2026-08-27
fix(core): reject checkpoint dtype casts

ss251:
## Defect

The supported public path `from alberta_framework import save_checkpoint, load_checkpoint` reaches `core/checkpoints.py`; the model-replay and interaction-feature checkpoint wrappers use the same boundary. When a caller supplied a same-shape template with a different array dtype, Orbax `StandardRestore` silently cast the persisted value instead of rejecting the incompatible learner state. A saved `int32` event counter therefore changed numerically during a successful restore.

## Before / after

Reproduction before the fix:

```text
saved dtype/value: int32 16777217
template dtype: float32
restored dtype/value: float32 16777216.0
difference: -1.0
```

The same supported call after the fix:

```text
ValueError: State structure mismatch. Ensure the learner architecture matches the saved checkpoint. Original error: checkpoint state leaf 0 has shape () and dtype int32; restore template requires shape () and dtype float32
```

## Fix

At the single `load_checkpoint` boundary, read Orbax's stored pre-cast leaf metadata and compare explicit array shape and dtype with the restore template before `StandardRestore` can coerce data. Python scalar templates retain their existing compatibility, and typed JAX PRNG keys are compared using Orbax's persisted `uint32` key-data representation.

## Regression proof

The new regression failed before the production change:

```text
F [100%]
1 failed in 1.29s
```

I then reverted only the production validation, kept the test, and reproduced the exact defect:

```text
F [100%]
1 failed in 1.74s
```

Restoring the production fix without changing the test produced:

```text
1 passed in 1.75s
```

## Verification

- `pytest` checkpoint core plus typed-PRNG wrapper coverage with `-n 2`: `55 passed in 3.63s`
- checkpoint wrapper coverage with `-n 2`: `202 passed in 55.01s`
- checkpoint-dependent learner coverage with `-n 2`: `199 passed in 14.83s`
- Ruff: `All checks passed!`
- mypy: `Success: no issues found in 1 source file`
- `git diff --check`: passed
- The broader macOS run reached `3752 passed, 63 skipped`; its 29 observed failures were unrelated Linux-only publication tests requiring `O_TMPFILE` or `renameat2`.

- [x] logs: N/A - exact before/after and test outputs are reproduced inline above

- [x] domain-artifact: N/A - persistence boundary fix has no scientific result artifact

- [x] llm-trajectory: N/A - private content-addressed trace is joined by the signed receipt below

Client / agent tooling: codex
— [asi-metric-fix-20260828]

### DIFF
--- a/alberta_framework/core/checkpoints.py
+++ b/alberta_framework/core/checkpoints.py
@@ -134,6 +134,42 @@ def _restore_empty_arrays(template: Any, restored: Any, manifest: object) -> Any
     )
 
 
+def _validate_restore_template(path: Path, template: Any) -> None:
+    """Reject shape or explicit array-dtype conversion before Orbax restore."""
+
+    saved_tree = ocp.StandardCheckpointHandler().metadata(path / "state").tree
+    saved_leaves = jax.tree.leaves(saved_tree)
+    template_leaves = jax.tree.leaves(template)
+    if len(saved_leaves) != len(template_leaves):
+        raise ValueError("checkpoint state leaf count does not match the restore template")
+
+    for index, (saved, expected) in enumerate(
+        zip(saved_leaves, template_leaves, strict=True)
+    ):
+        if hasattr(expected, "shape") and hasattr(expected, "dtype"):
+            if jnp.issubdtype(expected.dtype, jax.dtypes.prng_key):
+                key_data = jax.random.key_data(expected)
+                expected_shape = tuple(key_data.shape)
+                expected_dtype = str(key_data.dtype)
+            else:
+                expected_shape = tuple(expected.shape)
+                expected_dtype = str(expected.dtype)
+        else:
+            expected_array = np.asarray(expected)
+            expected_shape = expected_array.shape
+            expected_dtype = None
+        saved_shape = tuple(saved.shape)
+        saved_dtype = str(saved.dtype)
+        if saved_shape != expected_shape or (
+            expected_dtype is not None and saved_dtype != expected_dtype
+        ):
+            raise ValueError(
+                f"checkpoint state leaf {index} has shape {saved_shape} and dtype "
+                f"{saved_dtype}; restore template requires shape {expected_shape} and "
+                f"dtype {expected_dtype or 'its scalar type'}"
+            )
+
+
 def _copy_mapping(payload: object, *, name: str) -> dict[str, Any]:
     if not issubclass(type(payload), Mapping):
         raise ValueError(f"{name} must be a mapping")
@@ -257,6 +293,7 @@ def load_checkpoint(
 
     try:
         compatible_template = _orbax_compatible_tree(state_template)
+        _validate_restore_template(path, compatible_template)
         with ocp.Checkpointer(ocp.CompositeCheckpointHandler()) as ckptr:
             loaded = ckptr.restore(
                 str(path),
--- a/tests/test_checkpoints.py
+++ b/tests/test_checkpoints.py
@@ -56,6 +56,15 @@ class TestSaveLoadRoundTrip:
         with pytest.raises(ValueError, match="does not match the restore template"):
             load_checkpoint(template, tmp_path / "empty_shape")
 
+    def test_template_dtype_cannot_silently_cast_saved_state(self, tmp_path):
+        saved = {"accepted_events": jnp.asarray(16_777_217, dtype=jnp.int32)}
+        template = {"accepted_events": jnp.asarray(0, dtype=jnp.float32)}
+
+        save_checkpoint(saved, tmp_path / "dtype_mismatch")
+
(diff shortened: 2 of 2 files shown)


saltwire-dev (changes requested):
Reviewed at exact head `57ec4eb481ab446b1b0092d96e2cefc673a9d9ff` against production `alberta_framework/core/checkpoints.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13, jax 0.11.1 (`jax_enable_x64=False`), orbax-checkpoint.

Comparing Orbax’s stored leaf metadata to the restore template **before** `StandardRestore` is the right *direction* for #2714 — origin’s `load_checkpoint` handed a same-shape float32 0-d template to Orbax, which silently cast a saved int32 `16777217` to `float32 16777216.0`. The hole is that the named contract is **reject checkpoint dtype casts at the single `load_checkpoint` boundary**, and the overlay only special-cases **template leaves that duck-type as arrays** (`hasattr(expected, "shape") and hasattr(expected, "dtype")`). Python-scalar leaves set `expected_dtype = None` and skip the dtype compare; typed PRNG keys are lowered to `key_data` **uint32 (2,)**, which is the same shape/dtype as `step_words` / `rng_key_data`. The new test is a 1-leaf jax-array dict with `match="dtype"` and cannot see those paths. On this CPU, the PR’s own int32 vs jax-float32 recipe is After-correct, and a Python-float template / key↔uint32 template is not.

**Blocking: array-array dtype mismatch is rejected; Python-scalar templates and key↔uint32 still restore, so the same silent conversion remains on this boundary**

On this CPU (this head’s overlay predicate vs origin `StandardRestore` vs an honest “every leaf has a dtype, keys are not uint32 arrays”):

|---|---|---|---|
| **PR body: int32 `16777217` vs jax float32 0-d** | **`float32 16777216.0`** | **REJECT** | REJECT (After) |
| int32 vs jax float16 0-d | **`float16 inf`** | REJECT | REJECT |
| **int32 vs Python `0.0`** | ACCEPTED **`float 16777217.0`** | **ACCEPTED** (dtype skipped) | **REJECT** |
| **typed `jr.key` vs uint32 (2,)** | restore as uint32 key-data | **ACCEPTED** | **REJECT** |
| matching jax int32 vs int32 | `16777217` preserved | ACCEPTED | ACCEPTED |

The first After cell is the PR’s own invariant. Python-float template and key↔uint32 are the leftover of the same named defect: a same-shape template still lets Orbax convert learner state at `load_checkpoint`. Mechanically: Python `float` has no `.shape`/`.dtype`, so `expected_dtype is None` and only `()` vs `()` is checked; Orbax then restores the int32 counter as a Python float. Typed keys are persisted as `ArrayMetadata` `shape=(2,) dtype=uint32`; the overlay **lowers the template key to that representation**, so a `uint32 (2,)` template (`step_words` / `rng_key_data`) is indistinguishable from a key. LearnerState / MultiHeadMLPState persist `birth_timestamp` / `uptime_s` as Python floats next to jax `step_count`. InteractionFeatureState has a typed `key` **and** `step_words: UInt[Array, " 2"]`.

**Tests lock a 1-leaf jax-array dict `match="dtype"`, not Python scalars, not key↔uint32, not the value `16777217`, not the wrappers**

No pin that the restored **value** is not `16777216.0`. No pin on Python `0.0` / `0` templates — on this CPU those **succeed** (`float 16777217.0`). No pin on uint32 (2,) vs typed key. A revert that drops only `_validate_restore_template` fails the new assert on jax-float32. A “fix” that only rejects jax-array `str(dtype)` is exactly this head: the new assert stays green and Python-scalar / key↔uint32 restores stay **ACCEPTED**.

**What I would approve**

Keep failing closed on the PR body’s jax int32 `16777217` vs jax float32 path. Then: (1) give every restore leaf a dtype — for Python scalars use `np.asarray(expected).dtype` (the value the overlay already computes and throws away); (2) do not lower typed PRNG keys to uint32 unless both sides are keys; `step_words` / `rng_key_data` uint32 `(2,)` must not be accepted as a `jr.key` template; (3) pin same saved leaf vs Python `0.0` / `0` → **REJECT** (today **ACCEPTED**); pin matching jax int32 vs int32 → restored value **`16777217`**; pin typed `jr.key` vs `jnp.zeros((2,), uint32)` both directions → **REJECT**.

Until `load_checkpoint({"accepted_events": 0.0}, path)` on a saved int32 `16777217` returns **REJECTED** rather than **ACCEPTED `float 16777217.0`**, and a uint32 `(2,)` template does not accept a saved typed key, this does not lock “reject checkpoint dtype casts” as a restore-boundary fix — it locks “jax-array `str(dtype)` must match” and leaves origin’s silent conversion on the Python-scalar and key-data paths, now with a cleaned jax-float32 error attached.

PR #2715, opened 2026-08-27
fix(cora): remove task identity from evaluation ties

ss251:
## Defect

The supported `run_cora_development` benchmark path declares that candidate agents do not receive task IDs, but `_evaluate` used the hidden evaluation `task_id` to break shared-Q action ties. On the frozen target sequence `(0, 1, 0)`, an untrained task-agnostic candidate therefore produced the impossible initial evaluation row `(1.0, 1.0, 1.0)`: the harness silently gave the policy a different target-matching action in each evaluation column.

This fixes the defect at the evaluation boundary. Shared-Q ties now use only the run seed and checkpoint block, so one task-agnostic policy has one action across every task evaluated at that checkpoint.

## Before

Reproduced through `run_cora_development(seed=FROZEN_SEEDS[0], steps_per_task=1, replay_capacity=2)`:

```text
replay_q matrix= ((1.0, 1.0, 1.0), ...)
replay_off matrix= ((1.0, 1.0, 1.0), ...)
reported= 0.7142857142857143 0.0 -0.3333333333333333
```

The failing regression exposed the same impossible row:

```text
assert (1.0, 1.0, 1.0) in ((1.0, 0.0, 1.0), (0.0, 1.0, 0.0))
1 failed in 3.42s
```

## After

```text
replay_q initial= (1.0, 0.0, 1.0)
replay_off initial= (1.0, 0.0, 1.0)
reported= 0.6666666666666666 0.0 0.0
```

These remain permanently nonpromoting development diagnostics; the corrected numbers are not performance or scientific evidence.

## Regression proof

The regression calls the real benchmark runner and requires the initial shared-Q row to be one task-agnostic action pattern: `(1, 0, 1)` or `(0, 1, 0)`. I reverted only the production-line fix while retaining the test, and it failed on the exact hidden-task-identity defect:

```text
assert (1.0, 1.0, 1.0) in ((1.0, 0.0, 1.0), (0.0, 1.0, 0.0))
1 failed in 1.89s
```

After restoring the fix, the exact-head focused suite passed:

```text
17 passed in 3.72s
```

Command:

```text
$ASI_REVIEW_VENV/bin/python -m pytest tests/test_cora_development.py tests/test_cora_evaluation_matrix_hang.py -q -n 2 -o addopts=''
```

Static checks on the touched files also pass:

```text
ruff: All checks passed!
mypy alberta_framework/benchmarks/cora_development.py: Success: no issues found in 1 source file
git diff --check: clean
```

The repository-wide mypy run still encounters the pre-existing macOS-only `os.O_TMPFILE` attribute error in `bounded_elastic_matched_runner.py`; this change does not touch that module.

Client / agent tooling: codex
— [runtime-metric-fix]

### DIFF
--- a/alberta_framework/benchmarks/cora_development.py
+++ b/alberta_framework/benchmarks/cora_development.py
@@ -232,7 +232,7 @@ def _evaluate(
             query_index = block * len(TASK_TARGETS) + task_id
             action = int(jr.randint(_rng_key(seed, 1, query_index), (), 0, 2))
         else:
-            action = _greedy(shared, (block + task_id) % 2)
+            action = _greedy(shared, (seed + block) % 2)
         returns.append(float(action == target))
     return tuple(returns)
 
--- a/tests/test_cora_development.py
+++ b/tests/test_cora_development.py
@@ -43,6 +43,15 @@ def test_end_to_end_slice_has_metrics_information_contract_and_exact_receipts() 
     assert task_control.candidate_eligible is False
 
 
+def test_candidate_evaluation_tie_break_does_not_receive_task_identity() -> None:
+    result = run_cora_development(
+        seed=FROZEN_SEEDS[0], steps_per_task=1, replay_capacity=2
+    )
+
+    for arm in result.arms[:2]:
+        assert arm.evaluation_matrix[0] in ((1.0, 0.0, 1.0), (0.0, 1.0, 0.0))
+
+
 def test_jit_and_eager_update_paths_match_except_timing() -> None:
     compiled = run_cora_development(seed=FROZEN_SEEDS[1], steps_per_task=1)
     with jax.disable_jit():


saltwire-dev (changes requested):
Reviewed at exact head `25a1168384571af705b406ed15fb61d6fba77271` against production `alberta_framework/benchmarks/cora_development.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython 3.13.5.

Stripping `task_id` from the **evaluation** shared-Q tie is the right *direction* for #2715 — origin’s `_evaluate` used `(block + task_id) % 2`, so a tied untrained candidate took a different action in each column and published the impossible initial row `(1.0, 1.0, 1.0)` on targets `(0, 1, 0)`. The hole is that the named contract is **`task_ids_available_to_candidate=False`**, and the overlay only special-cases **`_evaluate`’s candidate branch**. `_run_arm` still does `tie = (seed + cycle + task_id + local_step) % 2` then `_greedy(shared, tie)` for the same candidate arms. When the first action is wrong, `_q_update` leaves the Q at zeros, so later training steps keep using the hidden task identity. Those contaminated Qs (and remaining ties) are what later evaluation rows publish. The new test only looks at `evaluation_matrix[0]` for `arms[:2]` on **even** `FROZEN_SEEDS[0]`, where the training leak is a no-op, and cannot see the leftover. On this CPU, the PR’s own seed-15810 row-0 recipe is After-correct, and an odd-seed later-row honest bind is not.

**Blocking: eval tie no longer uses `task_id`; candidate **training** still does, so later matrix rows on odd frozen seeds still depend on hidden task identity**

On this CPU (`steps_per_task=1`, tied-at-zero path):

| seed / row | origin eval | this head (eval overlay; train still has `task_id`) | honest: train+eval both drop `task_id` |
|---|---|---|---|
| **15810 row 0** (PR body) | **`(1,1,1)`** | **`(1,0,1)`** | `(1,0,1)` (After) |
| 15810 later rows | `(1,0,1)` ×6 | `(1,0,1)` ×6 | `(1,0,1)` ×6 (training leak is a no-op on this even seed) |
| **15811 row 0** | **`(1,1,1)`** | `(0,1,0)` | `(0,1,0)` |
| **15811 rows 1–6** | mixed `(0,0,0)` / `(1,1,1)` | later rows **diverge from honest** (`(1,0,1)` vs honest `(0,1,0)` from row 3) | later rows stay the no-task-id pattern |

The first After cell is the PR’s own invariant. 15811 later rows are the leftover of the same named defect: a task-agnostic candidate still has its published evaluation matrix depend on hidden `task_id`, because training ties still see it. Mechanically on seed 15811: first train step `tie=(15811+0+0+0)%2=1`, action 1 vs target 0 → reward 0 → Q stays `(0,0)`; next task `tie` flips **because `task_id` flipped**. Honest without `task_id` keeps `tie=1`. `validate_result` re-runs leaky `_run_arm` and accepts the contaminated trajectory. `task_ids_available_to_candidate=False` remains a flag next to a training loop that still indexes the tie by `task_id`.

**Tests lock row 0 of the even seed where the training leak is a no-op, not later rows, not odd seeds, not the information-contract flag**

No pin on `FROZEN_SEEDS[1]` / `[3]` (odd). No pin that later rows on 15811 equal a no-`task_id` training run. Today they **differ** from row 3 onward. A revert that drops only the one overlay line fails the new assert on row 0. A “fix” that only changes `_evaluate` is exactly this head: the new assert stays green and the leftover odd-seed later rows stay **ACCEPTED**.

**What I would approve**

Keep failing closed on the PR body’s seed-15810 row-0 `(1,1,1)` path. Then: (1) strip `task_id` from candidate **training** ties too — `_run_arm` should use `(seed + cycle + local_step) % 2` for `replay_q` / `replay_off_q`; leave `task_id_q_control` as the arm that is allowed task IDs; (2) make `task_ids_available_to_candidate=False` an invariant over `_greedy`, not a hardcoded boolean `validate_result` never checks; (3) pin every candidate row of all four frozen seeds ∉ `{(1,1,1),(0,0,0)}`; pin odd seed 15811 (and 15813) later rows equal a training run whose candidate tie has no `task_id`.

Until `run_cora_development(seed=15811, steps_per_task=1, replay_capacity=2)` publishes the same candidate matrix as a no-`task_id` training tie, this does not lock “remove task identity from evaluation ties” as an information-contract fix — it locks “row 0 of the even seed is one action” and leaves origin’s task-identity hole in the production **training** path, now with a cleaned initial eval row attached.

PR #2717, opened 2026-08-28
fix(ipmnist): require all seeds for confirmation

ss251:
`asi-ipmnist-campaign frontier` silently marked an arm as a confirmation candidate when its paired mean cleared the threshold even though one paired seed regressed. The supported path is the README-listed console script (`asi-ipmnist-campaign frontier` → `main()` → `build_frontier()`), and the wrong `true` can authorize the campaign's expensive 200-task confirmation wave.

Base: `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676` (current `origin/main`)

## Before / after

Reproduced through the public CLI with a base accuracy of `0.80` on seeds 0–2 and candidate accuracies `[0.79, 0.81, 0.81]`, using threshold `0.002`.

Before:

```json
{
  "screen_per_seed_delta": [-0.01, 0.01, 0.01],
  "screen_paired_delta_vs_base": 0.003333333333333336,
  "confirmation_candidate": true
}
```

After:

```json
{
  "screen_per_seed_delta": [-0.01, 0.01, 0.01],
  "screen_paired_delta_vs_base": 0.003333333333333336,
  "confirmation_candidate": false
}
```

## Fix

Compute the exact paired deltas once at the frontier aggregation boundary and mirror the canonical `ipmnist_screening.merge_shards()` compute-spending gate: require at least two paired seeds, every paired delta positive, and the paired mean above the configured threshold. This is one boundary fix; no call sites or unrelated metric modules change.

Open-PR searches for `ipmnist campaign frontier`, `mixed-sign confirmation`, `frontier losing seed`, and `all-positive confirmation` found no matching open implementation.

## Regression proof

The new regression uses the mixed-sign deltas above. With only the production hunk reverted and the test retained, it failed on exactly the defect:

```text
>       assert row["confirmation_candidate"] is False
E       assert True is False
1 failed in 2.73s
```

Restoring the production fix made the targeted test pass. The existing checked-in `frontier_results.json` reconstruction also remains unchanged because its sole confirmation candidate improves on every paired seed.

## Verification

```text
python -m pytest tests/test_ipmnist_campaign_tools.py tests/test_ipmnist_screening.py -q -o addopts= -n 2
423 passed in 44.79s

python -m ruff check alberta_framework/benchmarks/ipmnist_campaign_tools.py tests/test_ipmnist_campaign_tools.py
All checks passed!

python -m mypy alberta_framework/benchmarks/ipmnist_campaign_tools.py
```

The factory proof ledger recorded the 423-test run at head `5507a3614d53c16d9aead3e33e0cdee3b14e750f`.

- [x] logs: N/A - deterministic before/after CLI and exact-head test output are reproduced inline above

- [x] domain-artifact: N/A - this corrects a nonpromoting compute gate and does not produce a scientific evidence artifact

- [x] llm-trajectory: N/A - the minimized trajectory is private by protocol and is bound by the terminal device-signed receipt

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: grok

Client / agent tooling: codex
— [codex-metric-fix-20260828]

### DIFF
--- a/alberta_framework/benchmarks/ipmnist_campaign_tools.py
+++ b/alberta_framework/benchmarks/ipmnist_campaign_tools.py
@@ -154,18 +154,22 @@ def build_frontier(
                 f"{sorted(screen)} != {sorted(screen_base)}"
             )
         paired_seeds = sorted(screen)
+        screen_deltas = [
+            screen[seed] - screen_base[seed] for seed in paired_seeds
+        ]
+        screen_paired_delta = statistics.mean(screen_deltas)
         row: dict[str, Any] = {
             "config_name": arm,
             "n_screen_seeds": len(paired_seeds),
             "screen_mean": statistics.mean(screen[seed] for seed in paired_seeds),
-            "screen_paired_delta_vs_base": statistics.mean(
-                screen[seed] - screen_base[seed] for seed in paired_seeds
-            ),
-            "screen_per_seed_delta": [
-                round(screen[seed] - screen_base[seed], 6) for seed in paired_seeds
-            ],
+            "screen_paired_delta_vs_base": screen_paired_delta,
+            "screen_per_seed_delta": [round(delta, 6) for delta in screen_deltas],
         }
-        row["confirmation_candidate"] = row["screen_paired_delta_vs_base"] > threshold
+        row["confirmation_candidate"] = (
+            len(paired_seeds) >= 2
+            and all(delta > 0.0 for delta in screen_deltas)
+            and screen_paired_delta > threshold
+        )
         if confirm:
             if confirm.keys() != confirm_base.keys():
                 raise ValueError(
--- a/tests/test_ipmnist_campaign_tools.py
+++ b/tests/test_ipmnist_campaign_tools.py
@@ -299,6 +299,28 @@ def test_frontier_requires_and_uses_exact_paired_seed_sets(tmp_path: Path) -> No
     } == {"pyproject.toml", "uv.lock"}
 
 
+def test_frontier_requires_every_paired_seed_to_improve(tmp_path: Path) -> None:
+    screen = tmp_path / "screen"
+    confirm = tmp_path / "confirm"
+    for seed, accuracy in enumerate((0.79, 0.81, 0.81)):
+        _shard(screen / f"base_seed{seed}.json", seed=seed, accuracy=0.80)
+        _shard(screen / f"candidate_seed{seed}.json", seed=seed, accuracy=accuracy)
+
+    frontier = build_frontier(
+        screen,
+        confirm,
+        base="base",
+        arms=["candidate"],
+        threshold=0.002,
+        created_unix=0.0,
+    )
+    row = frontier["results"][0]
+
+    assert row["screen_per_seed_delta"] == pytest.approx([-0.01, 0.01, 0.01])
+    assert row["screen_paired_delta_vs_base"] > 0.002
+    assert row["confirmation_candidate"] is False
+
+
 def test_frontier_rejects_mismatched_confirm_seed_sets(tmp_path: Path) -> None:
     screen = tmp_path / "screen"
     confirm = tmp_path / "confirm"


saltwire-dev (changes requested):
Reviewed at exact head `5507a3614d53c16d9aead3e33e0cdee3b14e750f` against production `alberta_framework/benchmarks/ipmnist_campaign_tools.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython host statistics only.

Requiring every paired **seed-mean** to be strictly positive before `build_frontier` sets `confirmation_candidate` is the right *direction* for #2717 — origin’s frontier handed a mixed-sign 3-seed vector `[-0.01, 0.01, 0.01]` (mean `0.00333… > 0.002`) straight to `confirmation_candidate = True`, which is the spend flag for the 200-task confirmation wave. The hole is that the named contract is **mean must not hide a regression at this aggregation boundary** (and “mirror `merge_shards()`: ≥2 paired seeds, every paired delta positive, mean above threshold”), and the overlay only special-cases **the boolean flag computed from already-averaged seed scores**. `seed_means` still collapses `per_task_accuracy` to a mean *before* `all(delta > 0)`; the confirm block in the same function still publishes `confirm_paired_delta_vs_base = statistics.mean(...)` with **no** all-positive / n≥2 bind. The new test is one 3-seed mixed-sign False on `_shard(..., accuracy=scalar)` (two identical tasks) and cannot see those paths. On this CPU, the PR’s own `[-0.01, 0.01, 0.01]` recipe is After-correct, and a within-seed task regression / confirm-stage mixed-sign / n=1 pin is not.

**Blocking: 3-seed mixed-sign screen flag is False; within-seed averaging and confirm-stage mean still hide a regressing grain, so the named bind is not the boundary**

On this CPU (this head’s overlay vs origin `mean > threshold` vs honest “no hidden regression at this boundary”):

|---|---|---|---|
| **PR body: seeds `[0.79, 0.81, 0.81]` vs `0.80`, threshold `0.002`** | **`True`** (mean `0.00333…`) | **`False`** | REJECT (After) |
| last-fit two seeds all `+0.01` | True | **True** | ACCEPTED |
| **n=1 seed `+0.01`** | **True** | **False** (`n>=2`) | REJECT (canonical `merge_shards` already) |
| **within-seed task mix:** seed0 `per_task [0.70, 0.92]` vs base `[0.80, 0.80]` (seed-mean `+0.01`), seed1 all `+0.01` | **True** | **True** | **REJECT** (task 0 regressed; mean hid it) |
| **confirm mixed-sign** after all-positive screen | confirm field mean-only | screen flag **True**; confirm **still mean-only** | confirm must not look like all-seed success |

The first After cell is the PR’s own invariant. Within-seed averaging and confirm-stage mean are the leftover of the same named defect: origin still would have let a mean hide a regressing grain, and this overlay stops that **only after** `seed_means` has already averaged `per_task_accuracy`. Mechanically: `all(delta > 0)` never sees a task list; it sees one float per seed. `n>=2` *is* in the overlay but the new test uses three seeds, so a revert of only `len(paired_seeds) >= 2` stays green.

**Tests lock one 3-seed mixed-sign `confirmation_candidate is False`, not n=1, not last-fit, not per-task mix, not confirm, not CLI**

`_shard` writes `per_task_accuracy: [accuracy, accuracy]`. No pin that n=1 `+0.01` is False. No pin on within-seed `[0.70, 0.92]` vs `[0.80, 0.80]` — overlay **True**. No pin on confirm-stage mixed-sign (empty confirm dir). PR body says reproduced through the public CLI; the new test calls `build_frontier` only. A revert that drops only `all(delta > 0)` fails the new assert. A “fix” that only rejects this 3-seed scalar mix is exactly this head: the new assert stays green and within-seed / confirm-stage / n=1 stay **unpinned**.

**What I would approve**

Keep failing closed on the PR body’s 3-seed mixed-sign path. Then: (1) either bind `per_task_accuracy` so a task regression cannot average into a positive seed-mean, or document that the contract is seed-mean only and pin that; (2) same three-clause bind on confirm-stage if confirm shards exist; (3) pin n=1 `+0.01` → **False**; pin last-fit 2-seed all `+0.01` → **True** on the public path; pin within-seed `[0.70, 0.92]` vs `[0.80, 0.80]` plus a second improving seed → **False** if the contract is no hidden regression (today **True**); pin confirm mixed-sign after all-positive screen; pin CLI `asi-ipmnist-campaign frontier`.

Until `build_frontier` on a two-seed shard whose seed0 `per_task_accuracy` is `[0.70, 0.92]` vs base `[0.80, 0.80]` returns **`confirmation_candidate is False`** rather than **True**, and until n=1 / last-fit / confirm-stage are pinned on the public CLI path the PR named, this does not lock “require all seeds for confirmation” as a mean-hides-regression fix — it locks “three scalar seed-means must not mix signs” and leaves origin’s averaging trick on the within-seed and confirm paths, now with a cleaned 3-seed False attached.

PR #2718, opened 2026-08-28
fix(benchmarks): derive rule-discovery verdicts

ss251:
## Defect

The supported `asi-rule-discovery-summary` path accepts explicit campaign directories and
seed rosters, computes their current arm means and paired deltas, but the maintained v2
payload copied verdict strings containing fixed numbers from the historical v1 campaign.
As a result, a real caller could publish a summary whose verdict contradicted the metrics in
the same JSON object. This branch is based on current `origin/main` at

Before, an explicit one-seed input with `disc_r1=0.10` and champion `=0.95` produced:

```text
computed disc_r1 mean: 0.1
reported disc_r1 verdict: below bar (0.78372); beats published UPGD-W control +0.006, no gate
computed paired delta: -0.85
reported paired verdict: discovered structure (surprise budget replaces utility gate) at champion constants beats the champion on the 60-task screen (+0.00173, 3/3 seeds) and ties-to-slightly-beats at 200 tasks (+0.00066 paired, 2/3 seeds; screening claims nothing by itself)
```

The maintained v2 boundary now derives each verdict from its already-computed screen mean and
paired comparison. The legacy v1 builder remains unchanged, preserving exact reconstruction
of the pinned historical payload. The same input now produces:

```text
computed disc_r1 mean: 0.1
reported disc_r1 verdict: screen mean 0.10000; paired vs champion -0.85000; development screen only
computed paired delta: -0.85
reported paired verdict: screen mean 0.10000; paired vs champion -0.85000; development screen only
```

## Regression proof

The new regression constructs explicit non-historical inputs and checks both a verbatim arm
and the previously positive-claim arm. With only the production change removed and the test
retained, it fails on the original stale number:

```text
E       AssertionError: assert 'below bar (0....006, no gate' == 'screen mean ...t screen only'
E         - screen mean 0.10000; paired vs champion -0.85000; development screen only
E         + below bar (0.78372); beats published UPGD-W control +0.006, no gate
```

Final verification in the exact-main prebuilt runtime, network-disabled and capped to two
CPUs:

```text
83 passed in 1.17s
All checks passed!
```

Commands covered `tests/test_ipmnist_campaign_tools.py` and Ruff on the changed source and
test. The environment image does not include pytest-xdist, so pytest ran serially under the
two-CPU container cap rather than accepting `-n 2`.

- AI assistance: yes

- Model(s) used: openai/gpt-5

- Agent tooling: Codex

- Skill path: SlopDotCash/slopdotcash@ffc07b0ca4888df20faee5b490181dbcaae5cc91:skills/contribute-to-asi/SKILL.md

AI provider/model: OpenAI / gpt-5
Client / agent tooling: Codex
Contribution skill revision: SlopDotCash/slopdotcash@ffc07b0ca4888df20faee5b490181dbcaae5cc91:skills/contribute-to-asi/SKILL.md
— [contribute-to-asi]

### DIFF
--- a/alberta_framework/benchmarks/rule_discovery_summary.py
+++ b/alberta_framework/benchmarks/rule_discovery_summary.py
@@ -137,6 +137,19 @@ def build_rule_discovery_summary(
     ).encode()
     maintained = dict(legacy)
     maintained["schema"] = "asi.rule_discovery.real_screen.v2"
+    screen = maintained["screen_60_task"]
+    paired = maintained["paired_vs_champion_60_task"]
+    assert isinstance(screen, dict)
+    assert isinstance(paired, dict)
+    maintained["verdicts"] = {
+        (
+            f"{name}_verbatim" if name in {"disc_r1", "disc_r2", "disc_r3"} else name
+        ): (
+            f"screen mean {screen[name]['mean']:.5f}; paired vs champion "
+            f"{paired[name]['mean']:+.5f}; development screen only"
+        )
+        for name in DISCOVERY_ARMS
+    }
     maintained["legacy_compatibility"] = {
         "schema": legacy["schema"],
         "canonical_sha256": hashlib.sha256(legacy_canonical).hexdigest(),
--- a/tests/test_ipmnist_campaign_tools.py
+++ b/tests/test_ipmnist_campaign_tools.py
@@ -695,6 +695,22 @@ def test_rule_discovery_summary_uses_explicit_directories(tmp_path: Path) -> Non
     assert "ipmnist_provenance" in summary["provenance"]["sources"]
 
 
+def test_rule_discovery_summary_verdicts_follow_explicit_inputs(tmp_path: Path) -> None:
+    screen = tmp_path / "screen"
+    for name in SCREEN_ARMS:
+        accuracy = 0.95 if name == CHAMPION else 0.1
+        _shard(screen / f"{name}_seed0.json", seed=0, accuracy=accuracy)
+
+    summary = build_rule_discovery_summary(screen, tmp_path / "confirm", seeds=(0,))
+
+    assert summary["verdicts"]["disc_r1_verbatim"] == (
+        "screen mean 0.10000; paired vs champion -0.85000; development screen only"
+    )
+    assert summary["verdicts"]["disc_r1_pscale_norms"] == (
+        "screen mean 0.10000; paired vs champion -0.85000; development screen only"
+    )
+
+
 @pytest.mark.parametrize("seeds", [(), (0, 0), (True,), (-1,), (2**32,)])
 def test_rule_summary_rejects_noncanonical_or_duplicate_seeds_before_io(
     tmp_path: Path, seeds: object


saltwire-dev (changes requested):
Reviewed at exact head `cae0ac9fc29c3e7a72bf23158a9691e62cb50639` against production `alberta_framework/benchmarks/rule_discovery_summary.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython host formatting only.

Deriving maintained v2 `verdicts` from the screen/paired means already sitting in the same JSON is the right *direction* for #2718 — origin’s v2 builder handed `dict(legacy)` straight through, so an explicit one-seed `disc_r1=0.10` / champion `0.95` published `mean: 0.1` / paired `-0.85` next to `below bar (0.78372); beats published UPGD-W control +0.006, no gate` and the historical “discovered structure … +0.00173, 3/3 seeds … 200 tasks (+0.00066” string. The hole is that the named contract is **a real caller could publish a summary whose verdict contradicted the metrics in the same JSON object**, and the overlay only special-cases **the `verdicts` dict, formatted from already-averaged screen/paired means, always suffixed `development screen only`**. `_arm` still collapses `per_task_accuracy` to a mean *before* the formatter; `confirm_200_task` can still be populated in the same object while every verdict denies confirm; `bar: 0.8640` and `upgd_w_control` still sit next to those strings unused; three of five `DISCOVERY_ARMS` keys are production-updated and **unpinned**. The new test is one 1-seed empty-confirm recipe on `_shard(..., accuracy=scalar)` asserting only `disc_r1_verbatim` and `disc_r1_pscale_norms`. On this CPU, the PR’s own `0.10` vs `0.95` recipe is After-correct, and a confirm-stage / other-arm / bar-control / CLI pin is not.

**Blocking: two asserted keys reprint already-averaged means; confirm-stage, the other three arms, bar/control, and the public CLI still let a verdict contradict metrics in the same JSON**

On this CPU (overlay formatter vs origin hardcoded v1 vs honest “verdict must not contradict the metrics in the same JSON object”):

|---|---|---|---|
| **PR body: 1-seed `disc_r1=0.10`, champion `0.95`** | **`below bar (0.78372); beats published UPGD-W control +0.006, no gate`** | **`screen mean 0.10000; paired vs champion -0.85000; development screen only`** | After for those two floats |
| **`disc_r2_verbatim` / `disc_r3_verbatim` / `disc_r1_pscale`** | **still historical** | overlay **does** reprint (production-After) | After — **unpinned** |
| **partial fix** that overwrites only the two asserted keys | other three keys historical | new test **green** | REJECT (must bind the full roster) |
| **confirm populated** (confirm mean ≠ screen) | pscale_norms still claims stale `+0.00066` | **every** verdict still `development screen only`; `confirm_200_task` sits next to it | **REJECT** |
| **CLI** `asi-rule-discovery-summary` | same origin leak | same overlay | public path the PR named — **unpinned** |

The first After cell is the PR’s own invariant. Confirm-stage and the unpinned arms are the leftover of the same named defect. Mechanically: the formatter never sees a task list, a confirm mean, `bar`, or `upgd_w_control`; it sees one float per discovery arm plus a hardcoded disclaimer. `DISCOVERY_ARMS` *is* in the overlay (all five keys). The new test asserts two of them, so a revert of only `disc_r2` / `disc_r3` / `disc_r1_pscale` stays green.

**Tests lock two 1-seed empty-confirm reprint strings, not the roster, not confirm, not bar, not CLI**

Confirm dir is empty. No pin that the other three keys reprint. No pin that confirm-stage numbers appear in the verdict (or that `development screen only` is refused when `confirm_200_task` is populated). Existing 3-seed helper never reads `verdicts`. PR body names the public console script; the new test calls `build_rule_discovery_summary` only. A revert that drops the entire overwrite fails the new assert. A “fix” that only overwrites the two PR-body keys keeps the new assert green.

**What I would approve**

Keep failing closed on the PR body’s 1-seed `0.10` vs `0.95` path. Then: (1) if confirm shards exist, the verdict must not say `development screen only` while `confirm_200_task` in the same object holds different numbers; (2) bind the full `DISCOVERY_ARMS` roster, not only the two keys the PR body quoted; (3) pin `disc_r2_verbatim` / `disc_r3_verbatim` / `disc_r1_pscale`; pin existing 3-seed `0.79` vs `0.80` helper path → v2 verdicts **not** origin `0.78372`; pin confirm populated with a confirm mean that is not the screen mean; pin CLI `asi-rule-discovery-summary`.

Until `build_rule_discovery_summary` on a confirm-populated explicit directory returns a verdict that does **not** say `development screen only` next to a non-empty `confirm_200_task`, and until the other three `DISCOVERY_ARMS` keys and the public CLI path the PR named are pinned, this does not lock “derive rule-discovery verdicts” as a metrics-contradiction fix — it locks “two 1-seed scalar keys reprint their screen/paired means” and leaves origin’s copy-from-legacy trick on confirm-stage and the unpinned arms, now with a cleaned two-key After attached.

PR #2723, opened 2026-08-28
fix(micro): require every paired gate seed to improve

ss251:
## Defect

The supported `python -m alberta_framework.benchmarks.micro_continual ladder` path merges campaign shards through `transfer_validation_from_shards()` into `transfer_validation()`. Its `gate_small_positive` check exposed an `all_seeds_positive` field but accepted a positive mean even when one paired seed regressed. That let the CLI write `transfer_valid: true` and exit 0 for a ladder that violated its stated all-seed gate.

## Before / after

Before, the supported validator accepted a mixed paired result:

```text
{"all_seeds_positive": false, "gate_passed": true, "per_seed_gate_delta": [-0.01, 0.03, 0.03], "transfer_valid": true}
```

After requiring every paired gate delta to be positive at the validator boundary:

```text
{"all_seeds_positive": false, "gate_passed": false, "per_seed_gate_delta": [-0.01, 0.03, 0.03], "transfer_valid": false}
```

The existing three-seed campaign shards still validate through a copied output directory without modifying immutable artifacts:

```text
transfer_valid=True (primary checks: {'conditioning_dominates': True, 'gate_small_positive': True, 'adam_decays': True, 'adam_below_upgd_raw': True, 'naive_bayes_placement': True, 'champion_top': True})
```

## Regression proof

The new regression was red before the fix:

```text
>       assert gate["passed"] is False
E       assert True is False
1 failed in 1.08s
```

With the fix:

```text
1 passed in 1.15s
```

I then reverted only the production predicate, kept the test unchanged, and reproduced the same assertion failure (`1 failed in 1.08s`) before restoring the fix.

Focused verification at head `26b503eab510ba672a6541c05abadb207a1ff103` (one pytest worker in a Docker sandbox capped at two CPUs):

```text
11 passed in 8.11s
All checks passed!
```

The broader module run reached `223 passed`; its five remaining tests could not start because the available pinned runtime image lacks optional `scikit-learn`. Those are the separate legacy digits lane and do not exercise this validator.

No issue is linked; this is a directly reproduced supported-path validator defect. Open-PR searches for `gate_small_positive`, `micro_continual transfer_valid`, and all-seed transfer gating found no matching fix.

Client / agent tooling: codex
— [runtime-metric-fix]

### DIFF
--- a/alberta_framework/benchmarks/micro_continual.py
+++ b/alberta_framework/benchmarks/micro_continual.py
@@ -1588,6 +1588,7 @@ def transfer_validation(
         "gate_small_positive",
         bool(
             gate_delta.mean() > 0.0
+            and np.all(gate_delta > 0.0)
             and gate_delta.mean() <= conditioning_delta.mean() / 2.0
         ),
         "full protocol: gate +0.011, an order below conditioning +0.061",
--- a/tests/test_micro_continual.py
+++ b/tests/test_micro_continual.py
@@ -1723,6 +1723,25 @@ class TestTransferValidation:
         assert checks["gate_small_positive"]["passed"] is False
         assert report["transfer_valid"] is False
 
+    def test_gate_requires_every_paired_seed_to_improve(self):
+        ladder = _synthetic_ladder(seeds=(0, 1, 2))
+        ladder["gated_norm"] = {
+            0: np.full(8, 0.830),
+            1: np.full(8, 0.870),
+            2: np.full(8, 0.870),
+        }
+
+        report = transfer_validation(ladder)
+        checks = {check["name"]: check for check in report["checks"]}
+        gate = checks["gate_small_positive"]
+
+        assert gate["values"]["per_seed_gate_delta"] == pytest.approx(
+            [-0.010, 0.029, 0.028]
+        )
+        assert gate["values"]["all_seeds_positive"] is False
+        assert gate["passed"] is False
+        assert report["transfer_valid"] is False
+
     def test_nondecaying_adam_fails(self):
         ladder = _synthetic_ladder()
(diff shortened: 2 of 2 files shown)


hermesagent270-commits (approved):
Recommendation: merge exact head 26b503eab510ba672a6541c05abadb207a1ff103.

The defect is real and the fix is correctly scoped. Before this head, gate_small_positive could pass while its own all_seeds_positive diagnostic was false, allowing transfer_valid and the supported ladder CLI exit status to certify a mixed paired outcome. Requiring np.all(gate_delta > 0.0) makes the decision agree with the published diagnostic, the adjacent conditioning_dominates all-seed gate, and the campaign's all-seeds-positive escalation convention.

Independent verification in a disposable network-denied container:

- Complete tests/test_micro_continual.py: 228 passed, including the optional digits lane.
- Ruff on the changed production/test files: passed.
- Strict MyPy on alberta_framework/benchmarks/micro_continual.py: passed.
- Mutation proof: removing only the new np.all predicate makes test_gate_requires_every_paired_seed_to_improve fail because gate.passed becomes true while all_seeds_positive remains false; restoring it passes.
- GitHub required CI is green and merge state is clean at this exact SHA.

No blocking findings. This is validator correctness only and creates no performance or scientific claim.

Client / agent tooling: codex
— [review-2723]

lalalune:
This changes the development selection rule from a positive mean gate effect to requiring every seed to improve. The current transfer_validation contract explicitly requires per-seed superiority for conditioning_dominates and separately reports all_seeds_positive for the gate. No preregistered protocol change or measured calibration justifies strengthening that gate here.

PR #2724, opened 2026-08-28
fix(streams): preserve RiverSwim stationary reward ratios

ss251:
## Defect

`RiverSwimMDP.policy_average_reward()` silently misreported the long-run reward for supported slow-mixing kernels. This public path also supplies `optimal_average_reward()` and `uniform_random_average_reward()`, so the same numerical error corrupted RiverSwim benchmark baselines and the reference-life scorecard oracle.

The solver formed `P.T - I` from a float32 transition tensor. When off-diagonal probabilities were small, subtracting a nearly-one diagonal erased balance information before NumPy's least-squares rank decision.

## Before and after

Supported reproduction: `RiverSwimConfig(n_states=2, p_right_up=1e-7, p_right_down=2e-7, reward_left=0, reward_right=1)` with an always-right policy.

Before:

```text
stored kernel row 0 = [0.9999998807907104, 1.0000000116860974e-07]
stored kernel row 1 = [2.0000000233721948e-07, 0.9999998211860657]
expected always-right average reward = 0.333333342310079
reported always-right average reward = 0.367051452471610
absolute error = 0.033718110161531
```

After:

```text
expected always-right average reward = 0.333333342310079
reported always-right average reward = 0.333333342310079
absolute error = 0.000000000000000
```

## Fix

At the single stationary-solver boundary, the patch normalizes the stored categorical weights in float64, constructs the equivalent flow generator directly from off-diagonal mass, scales nonzero balance equations, and retains the normalization constraint. This avoids the destructive diagonal subtraction without changing callers or policy enumeration.

## Regression proof

The focused test derives the exact two-state stationary ratio from the stored, row-normalized categorical weights. With only the production change reverted and the new test retained, it failed on the original defect:

```text
E       assert 0.36705145247160986 == 0.3333333423100792 ± 1.0e-12
E         Obtained: 0.36705145247160986
E         Expected: 0.3333333423100792 ± 1.0e-12
1 failed in 1.46s
```

After restoring the fix:

```text
185 passed, 1 skipped in 31.27s
All checks passed!
```

Commands used pytest with `-n 2` across `test_closed_loop_env.py`, `test_reference_life_riverswim.py`, and `test_reference_life_scorecard.py`, plus Ruff, mypy on the changed source, and `git diff --check`. Independent adversarial review approved the numerical construction and regression sensitivity.

`closed_loop.py` is registered by the historical continual-IA evidence chain. This source correction intentionally leaves every pinned artifact untouched; the registry's fail-closed source-drift response remains intact until that frozen protocol is independently rerun.

- [x] logs: reproduced before/after and exact-head test output are included above; N/A attachment because this deterministic unit-scale proof has no external log artifact.

- [x] domain-artifact: N/A — this fixes an analytic benchmark metric and does not create or modify scientific evidence artifacts.

- [x] llm-trajectory: private trace finalized as `sha256:cf079109d8e2e109a260066beee5ca5f4e67bed0e4986fc8056aa9a3e9819f8f`; the signed receipt below carries the immutable server join.

- AI assistance: yes

- Model(s) used: openai/gpt-5.6-sol

- Agent tooling: codex

Client / agent tooling: codex
— [metric-integrity-fix]

### DIFF
--- a/alberta_framework/streams/closed_loop.py
+++ b/alberta_framework/streams/closed_loop.py
@@ -37,6 +37,7 @@ The state and config records are immutable chex dataclasses.
 from __future__ import annotations
 
 import itertools
+import math
 import operator
 from fractions import Fraction
 from numbers import Real
@@ -233,7 +234,28 @@ def _stationary_average_reward(
         The long-run average reward ``d @ step_rewards``.
     """
     n = transition.shape[0]
-    constraints = np.vstack([transition.T - np.eye(n), np.ones((1, n))])
+    kernel = np.asarray(transition, dtype=np.float64)
+    row_totals = np.array([math.fsum(row) for row in kernel], dtype=np.float64)
+    kernel = kernel / row_totals[:, None]
+
+    # Build the equivalent continuous-time generator from the categorical
+    # kernel's off-diagonal mass.  Forming ``P - I`` would subtract nearly
+    # equal float32 diagonal values and can overwhelm very small transition
+    # probabilities with rounding error.  The categorical sampler normalizes
+    # each stored row, so use those normalized off-diagonal weights directly
+    # and derive the diagonal from their sum.
+    generator = kernel.copy()
+    np.fill_diagonal(generator, 0.0)
+    np.fill_diagonal(
+        generator,
+        [-math.fsum(row) for row in generator],
+    )
+
+    balance = generator.T
+    equation_scales = np.max(np.abs(balance), axis=1)
+    nonzero_equations = equation_scales > 0.0
+    balance[nonzero_equations] /= equation_scales[nonzero_equations, None]
+    constraints = np.vstack([balance, np.ones((1, n))])
     targets = np.zeros(n + 1)
     targets[-1] = 1.0
     distribution, *_ = np.linalg.lstsq(constraints, targets, rcond=None)
--- a/tests/test_closed_loop_env.py
+++ b/tests/test_closed_loop_env.py
@@ -510,6 +510,28 @@ class TestRiverSwim:
         assert uniform < optimal
         assert always_left < optimal
 
+    def test_stationary_gain_preserves_tiny_categorical_transition_ratios(self):
+        """Stationary mass follows the row-normalized categorical kernel."""
+        env = RiverSwimMDP(
+            RiverSwimConfig(
+                n_states=2,
+                p_right_up=1.0e-7,
+                p_right_down=2.0e-7,
+                reward_left=0.0,
+                reward_right=1.0,
+            )
+        )
+        kernel = env.transition_tensor[RIGHT_ACTION].astype(np.float64)
+        up = kernel[0, 1] / kernel[0].sum()
+        down = kernel[1, 0] / kernel[1].sum()
+        expected_top_state_mass = up / (up + down)
+
+        assert env.policy_average_reward([RIGHT_ACTION, RIGHT_ACTION]) == pytest.approx(
+            expected_top_state_mass,
+            rel=1.0e-12,
+            abs=1.0e-12,
+        )
+
     def test_policy_gain_matches_scan_simulation(self):
         """A long scan rollout of always-right attains its analytic gain."""
         env = RiverSwimMDP(RiverSwimConfig(n_states=4))


saltwire-dev (changes requested):
Reviewed at exact head `1e72e9407fc8a6604ee5177d6fbda3bac0459e69` against production `alberta_framework/streams/closed_loop.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython host numpy solver only.

Replacing `P.T - I` with a float64 row-normalized off-diagonal generator is the right *direction* for #2724 — origin’s `_stationary_average_reward` already published `policy_average_reward([RIGHT, RIGHT]) ≈ 0.367` for the supported `RiverSwimConfig(n_states=2, p_right_up=1e-7, p_right_down=2e-7)` always-right kernel whose stored row-normalized ratio is `0.333…`, so `RiverSwimMDP.policy_average_reward()` / `optimal_average_reward()` / `uniform_random_average_reward()` could hand a corrupted long-run reward to RiverSwim baselines and the reference-life scorecard oracle. The hole is that the named contract is **a real `policy_average_reward` (and the two helpers that share the solver) could report a stationary gain whose float32 `P.T - I` balance was already garbage**, and the overlay only special-cases **one 2-state always-right tiny kernel that the new test constructs by hand**. `uniform_random_average_reward` / `_enumerate_optimal` are production-wired and **unpinned** on a tiny kernel; a **swapped-ratio** sibling (`p_up=2e-7`, `p_down=1e-7`) is **unpinned**; a **3-state** always-right tiny sibling is **unpinned**; a **1e-6/2e-6** sibling is **unpinned**; the helper docstring still describes origin’s `d @ P = d` / `P.T - I` least-squares. The new test is one `n_states=2` always-right spy with this exact `(1e-7, 2e-7)` pair. On this CPU, the PR’s own mixed After (`0.333333342310079`) is After-correct, and a 3-state / swapped / uniform / 1e-6 pin is not.

**Blocking: generator rewrite is generic, but tests only lock one 2-state 1e-7/2e-7 always-right recipe; 3-state, swapped-ratio, uniform, and 1e-6 stay unpinned**

On this CPU (overlay generator vs origin `P.T - I` vs honest “row-normalized categorical ratio / detailed balance”):

|---|---|---|---|
| **PR body: 2-state always-right `1e-7/2e-7`** | **`0.367051452471610`** | **`0.333333342310079`** | After — pinned |
| **swapped 2-state `2e-7/1e-7` always-right** | **`0.632948547225683`** | overlay **`0.666666657689921`** | After (`2/3`) — **unpinned** |
| **3-state always-right `1e-7/2e-7`** | **`0.180148284803789`** | overlay **`0.142857146938776`** (`1/7`) | After — **unpinned** |
| **2-state `1e-6/2e-6` always-right** | **`0.329993460666350`** | overlay **`0.333333330382453`** | After — **unpinned** |
| **`uniform_random_average_reward` on PR config** | origin **`5.48e-8`** | overlay **`5.00e-8`** | After — **unpinned** |

The first After cell is the PR’s own invariant. 3-state, swapped-ratio, 1e-6, and uniform are the leftover of the same named defect. Mechanically: `uniform_random_average_reward` never sees “this RIGHT-action 2-state spy”; it sees the mean of LEFT+RIGHT kernels. The generator rewrite *is* generic. The new test asserts one 2-state pair and `≈ expected_top_state_mass`, so a revert of the uniform hop / a 3-state tiny / a swapped-ratio path stays green. Existing default n=4 spies already passed (~1e-7 abs); they cannot see the slow-mixing bug.

**Tests lock one 2-state always-right tiny recipe, not 3-state, not swapped-ratio, not uniform, not 1e-6**

No pin that `uniform_random_average_reward()` reprints the generator After for the same tiny config. No pin that `optimal_average_reward()` / `_enumerate_optimal` still pick always-right and return the honest `0.333…`. No pin on a swapped-ratio sibling. No pin on a 3-state always-right tiny sibling. No pin on a `1e-6/2e-6` sibling. A revert that restores `P.T - I` fails the new `≈ 0.333` assert. A “fix” that only this `n=2` `(1e-7, 2e-7)` always-right call keeps the new assert green.

**What I would approve**

Keep failing closed on the PR body’s 2-state always-right `1e-7/2e-7` path. Then: (1) pin the same tiny config through `optimal_average_reward` / `uniform_random_average_reward`, or stop advertising them as the supported path this overlay repairs; (2) bind a swapped-ratio sibling (`2e-7/1e-7` → `2/3`) and a 3-state always-right tiny sibling (`1/7`); (3) pin `1e-6/2e-6` sibling; pin uniform helper on the PR config; docstring names the generator / off-diagonal construction, not `d @ P = d` via `P.T - I`; revert of only the unpinned 3-state / swapped / uniform / 1e-6 paths must fail those pins.

Until a 3-state (or swapped-ratio) tiny kernel on `policy_average_reward` returns the honest ratio and until uniform / enumerate are pinned if they remain named hops, this does not lock “preserve RiverSwim stationary reward ratios” — it locks “one 2-state `(1e-7, 2e-7)` always-right reprint `≈ 0.333`” and leaves origin’s float32 diagonal subtract on the other live helpers.

PR #2726, opened 2026-08-28
fix(forager): bind custom policy metadata in summaries

ss251:
## Defect

The supported custom-policy path `compare_forager_agents` → `run_forager` →
`summarize_forager_runs` preserved arbitrary JSON-compatible policy metadata, but the
summarizer compared only a small hard-coded allowlist. Two seeds using different custom-policy
hyperparameters were therefore presented as replications of one method and silently produced a
wrong aggregate number.

Before, two runs with `learning_rate` values `0.01` and `0.99` were accepted:

```text
input learning_rates: [0.01, 0.99]
accepted seeds: (0, 1)
reported mean: 0.5
```

After the fix, the same supported input is rejected at the aggregation boundary:

```text
rejected=runs must share one method configuration and provenance
```

## Fix

`summarize_forager_runs` now binds all custom method metadata into its canonical method
signature. It normalizes only per-run identity, raw trace data, and timing telemetry while
retaining the existing stable runner identity fields. The change is confined to the one boundary
that decides whether runs may be aggregated.

Base: `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676` (`origin/main`).

## Regression proof

The new regression first failed on the original implementation:

```text
E Failed: DID NOT RAISE ValueError
1 failed in 2.99s
```

After applying the production fix it passed. I then reverted only the production change while
keeping the test, and it failed on exactly the same missing rejection:

```text
E Failed: DID NOT RAISE ValueError
1 failed in 2.81s
```

Restoring the fix produced:

```text
1 passed in 2.72s
```

## Verification

```text
.venv/bin/python -m pytest tests/test_forager_benchmark.py -q -o addopts= -n 2
103 passed, 3 skipped in 13.87s

.venv/bin/python -m ruff check alberta_framework/benchmarks/forager.py tests/test_forager_benchmark.py
All checks passed!

.venv/bin/python -m mypy alberta_framework/benchmarks/forager.py
```

- [x] logs: verbatim before/after, mutation proof, and exact-head test output are included above

- [ ] domain-artifact: N/A - this fixes aggregation validation and produces no scientific artifact

- [x] llm-trajectory: finalized private trace is bound in the signed receipt below

- AI assistance: yes

- Model(s) used: openai/gpt-5.6-sol

- Agent tooling: codex

Client / agent tooling: codex
— [codex-fix-metric-boundary-20260828]

### DIFF
--- a/alberta_framework/benchmarks/forager.py
+++ b/alberta_framework/benchmarks/forager.py
@@ -3987,39 +3987,23 @@ def summarize_forager_runs(
     def method_signature(run: ForagerRunResult) -> str:
         metadata = run.agent_metadata
         stable_metadata = {
-            "config": metadata.get("config"),
-            "result_source": metadata.get("result_source"),
-            "source_repository": metadata.get("source_repository"),
-            "source_commit": metadata.get("source_commit"),
-            "config_path": metadata.get("config_path"),
-            "attestation_state": metadata.get("attestation_state"),
-            "official_foragax_evidence": metadata.get(
-                "official_foragax_evidence"
-            ),
-            "protocol_attested": metadata.get("protocol_attested"),
-            "runtime_profile_id": metadata.get("runtime_profile_id"),
-            "environment_runtime_profile_sha256": metadata.get(
-                "environment_runtime_profile_sha256"
-            ),
-            "environment_rng_schedule": metadata.get(
-                "environment_rng_schedule"
-            ),
-            "environment_rng_schedule_sha256": metadata.get(
-                "environment_rng_schedule_sha256"
-            ),
-            "paper_search_oracle": metadata.get("paper_search_oracle"),
-            "random_backend": metadata.get("random_backend"),
-            "runner_kind": (
+            key: value
+            for key, value in metadata.items()
+            if key
+            not in {"seed", "name", "privileged", "runner", "raw_metric_trace"}
+        }
+        stable_metadata["runner"] = {
+            "kind": (
                 metadata.get("runner", {}).get("kind")
                 if isinstance(metadata.get("runner"), Mapping)
                 else None
             ),
-            "runner_batch_mode": (
+            "batch_mode": (
                 metadata.get("runner", {}).get("batch_mode")
                 if isinstance(metadata.get("runner"), Mapping)
                 else None
             ),
-            "runner_rounding_contract": (
+            "rounding_contract": (
                 metadata.get("runner", {}).get("rounding_contract")
                 if isinstance(metadata.get("runner"), Mapping)
                 else None
--- a/tests/test_forager_benchmark.py
+++ b/tests/test_forager_benchmark.py
@@ -1129,6 +1129,18 @@ def test_summary_and_paired_comparison_fail_closed() -> None:
         summarize_forager_runs(
             [candidate[0], _result("alberta_horde_ac", 1, 2.0, config_id="different")]
         )
+    different_custom_metadata = [
+        dataclasses.replace(
+            candidate[0],
+            agent_metadata={"learning_rate": 0.01},
+        ),
+        dataclasses.replace(
+            candidate[1],
+            agent_metadata={"learning_rate": 0.99},
+        ),
+    ]
+    with pytest.raises(ValueError, match="configuration"):
+        summarize_forager_runs(different_custom_metadata)
     incompatible_metric = _result("alberta_horde_ac", 1, 2.0, ewm_decay=0.5)
     with pytest.raises(ValueError, match="metric contract"):
         summarize_forager_runs([candidate[0], incompatible_metric])


saltwire-dev (changes requested):
Reviewed at exact head `86a350f26e2dedc6530e15880d2b078d83e408be` against production `alberta_framework/benchmarks/forager.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython host json replica only.

Binding custom-policy metadata at `summarize_forager_runs`’ nested `method_signature` is the right *direction* for #2726 — origin’s allowlist already published `accepted seeds: (0, 1)` / `reported mean: 0.5` for a supported `compare_forager_agents → run_forager → summarize_forager_runs` pair whose candidate shards were two custom-policy seeds with `learning_rate` 0.01 vs 0.99 and values 0.0 / 1.0, so the aggregator could present two different methods as replications of one. The hole is that the named contract is **a real `summarize_forager_runs` (and the `compare_forager_agents` CLI-path that calls it) could rank shards whose extra top-level policy metadata already disagreed**, and the overlay only special-cases **the one knob the new spy writes (`learning_rate` as the entire `agent_metadata` dict)**. `epsilon` is **unpinned**; nested `hyperparameters` is **unpinned**; in-tree extra keys origin also ignored (`update_semantics`, `recurrent_features`) are **unpinned**; `compare_forager_agents` / `_run_forager_host` live metadata shape is **unpinned** (spy never calls them; `compare_forager_agents` is 0 hits in the test file); live `runner.chunk_size` is **still dropped** (same as origin). The new test is `agent_metadata={"learning_rate": 0.01}` vs `{"learning_rate": 0.99}` on stripped `_result` objects. On this CPU, the PR’s own mixed After (`ValueError: runs must share one method configuration and provenance`) is After-correct, and an epsilon-only / nested-hyper / update_semantics / live-host pin is not.

**Blocking: signature rewrite is generic, but tests only lock learning_rate; epsilon, nested hyperparameters, in-tree extra keys, and the named compare/run_forager hop stay unpinned**

On this CPU (overlay denylist vs origin allowlist vs honest “same-method custom metadata”):

|---|---|---|---|
| **PR body: learning_rate 0.01 vs 0.99, values 0.0/1.0** | **accepted seeds (0, 1), mean `0.5`** | **`ValueError: … configuration and provenance`** | After — *partially* pinned (spy uses stripped `{learning_rate}` only) |
| **UNPINNED: epsilon 0.1 vs 0.9** | origin **mean 0.5, accept** | overlay **raises** | After — **unpinned** |
| **UNPINNED: nested `hyperparameters.lr` 0.01 vs 0.99** | origin **mean 0.5, accept** | overlay **raises** | After — **unpinned** |
| **UNPINNED: Alberta `update_semantics` v1 vs v2** | origin **mean 0.5, accept** | overlay **raises** | After — **unpinned** |
| **UNPINNED: live host shape + lr mix** | origin **mean 0.5, accept** | overlay **raises** | After — **unpinned**; spy never calls `_run_forager_host` / `compare_forager_agents` |
| **UNPINNED: `runner.chunk_size` 100 vs 999** | origin **mean 0.5, accept** | overlay **mean 0.5, accept** | After if chunk_size is method identity — **unpinned**; overlay still drops it |

The first After cell is the PR’s own invariant, but the spy never builds the live `_run_forager_host` metadata. epsilon, nested hyperparameters, in-tree extra keys, and `compare_forager_agents` are the leftover of the same named defect. Mechanically: a follow-up that only special-cases `learning_rate` (or that only snapshots stripped `_result` metadata, not `compare_forager_agents`) re-opens origin’s “allowlist, so an epsilon sibling stayed invisible” while `test_summary_and_paired_comparison_fail_closed` stays green. Existing `config_id="different"` spy already passed on origin (dumps `config`); it cannot see a top-level custom key.

**Tests lock learning_rate, not epsilon, not nested hyperparameters, not compare_forager_agents**

No pin that `epsilon=0.1` vs `0.9` raises. Overlay **does**. Spy never writes `epsilon`. Origin **mean 0.5 / accept**. No pin that nested `hyperparameters: {lr: 0.01}` vs `{lr: 0.99}` raises. No pin that in-tree extra keys `update_semantics` / `recurrent_features` raise when mixed. No pin that `compare_forager_agents` → `run_forager` → `summarize_forager_runs` rejects a custom policy whose `metadata()` differs only by `learning_rate` (`compare_forager_agents` count in the test file: **0**). No pin that live `runner.chunk_size` is method identity. A revert that restores the origin allowlist fails the new `learning_rate` assert. A “fix” that only `learning_rate` through stripped `summarize_forager_runs` keeps the new assert green.

**What I would approve**

Keep failing closed on the PR body’s mixed `learning_rate` path. Then: (1) pin mixed custom-policy ranking through `compare_forager_agents` / `_run_forager_host` (live `policy.metadata()` + stamped schedule + duration-bearing `runner`), or document that only a hand-built `summarize_forager_runs` bind is in scope; (2) bind an epsilon-only sibling and a nested `hyperparameters.lr` sibling; (3) pin in-tree extra key mixes; same learning_rate with duration-only `runner.overall_duration_s` mix should **accept**; docstring names custom-policy metadata identity; revert of only the unpinned epsilon / nested-hyper / update_semantics / `compare_forager_agents` paths must fail those pins.

Until an epsilon-only mix on `summarize_forager_runs` raises and until `compare_forager_agents` is pinned if it remains a named hop, this does not lock “bind custom policy metadata in summaries” — it locks “`learning_rate` through stripped `summarize_forager_runs` raises” and leaves origin’s unpaired-hyperparameter rank on every other custom key.

hermesagent270-commits:
I prepared a scoped repair for the remaining proof gap: https://github.com/ss251/asi/pull/5

It adds mutation-sensitive coverage for epsilon, nested hyperparameters, update semantics, recurrent features, the live compare_forager_agents path, and a timing-only positive control. Full affected module: 109 passed, 3 skipped; Ruff and mypy pass. The repair is based on the unchanged reviewed head 86a350f26e2dedc6530e15880d2b078d83e408be.

PR #2727, opened 2026-08-28
fix(benchmarks): suppress unpaired UPGD winner

ss251:
## Defect

The supported legacy development path

```text
python -m alberta_framework.benchmarks.upgd_ipmnist --merge-partials ...
```

accepted `upgd_w` and `adamw` shards with disjoint seed schedules. The artifact correctly recorded `all_learners_share_seed_ids: false`, but `build_comparison()` still published `upgd_w_beats_adamw` from the two unpaired means. An interrupted or interim merge could therefore silently report a confounded cross-learner winner.

This was reproduced from current `origin/main` at base SHA `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676` using the real v2 merge CLI with UPGD-W seed 0 and AdamW seed 1:

```json
{
  "seed_ids": {"adamw": [1], "upgd_w": [0]},
  "all_learners_share_seed_ids": false,
  "upgd_w_beats_adamw": true
}
```

## Fix

The comparison boundary now emits `upgd_w_beats_adamw` only when both summaries carry exactly the same seed schedule. Per-learner paper-reference diagnostics remain available for interim or disjoint merges; only the invalid cross-learner verdict is omitted. Matched historical seed schedules retain their existing boolean.

The same CLI reproduction after the fix reports:

```json
{
  "seed_ids": {"adamw": [1], "upgd_w": [0]},
  "all_learners_share_seed_ids": false,
  "upgd_w_beats_adamw_present": false
}
```

No pinned or active `outputs/` artifact was modified. This remains a permanently nonpromoting development lane and makes no performance or scientific claim.

## Regression proof

With only the production condition reverted and the new test retained:

```text
AssertionError: assert 'upgd_w_beats_adamw' not in comparison
1 failed in 5.69s
```

Restoring the production fix makes the focused regression pass.

## Verification

Measured head: `8d501cf0fd3617bffd78b554a7e639ef2905a4b9`

```text
.venv/bin/python -m pytest tests/test_upgd_ipmnist.py tests/test_upgd_ipmnist_nonpromoting.py -q -o addopts="" -n 2
154 passed in 34.68s

.venv/bin/python -m ruff check alberta_framework/benchmarks/upgd_ipmnist.py tests/test_upgd_ipmnist.py
All checks passed!

.venv/bin/python -m mypy alberta_framework/benchmarks/upgd_ipmnist.py

git diff --check
exit 0
```

The repository-wide mypy invocation also reached one existing Darwin-only error at `alberta_framework/evaluation/bounded_elastic_matched_runner.py:1111` because `os.O_TMPFILE` is unavailable on this platform; it is outside this diff.

An independent read-only adversarial review approved the boundary, test sensitivity, strict-validator compatibility, and matched-artifact compatibility. Exact open-PR searches for `unpaired learner comparison`, `disjoint seeds winner`, `all_learners_share_seed_ids`, `upgd_w_beats_adamw`, and `UPGD IPMNIST paired seeds` found no pull request implementing this fix.

- [x] logs: N/A — deterministic before/after and regression outputs are included above; no long benchmark run was needed.

- [x] domain-artifact: N/A — temporary reproduction artifacts were development-only and were not published.

- [x] llm-trajectory: private trace finalized; its public SHA-256 and upload identity are bound in the signed receipt below.

Client / agent tooling: codex
— [metric-fix]

### DIFF
--- a/alberta_framework/benchmarks/upgd_ipmnist.py
+++ b/alberta_framework/benchmarks/upgd_ipmnist.py
@@ -1254,7 +1254,11 @@ def build_comparison(summaries: dict[str, dict[str, Any]]) -> dict[str, Any]:
             "gap": round(gap, 6),
             "reproduction_gap_flagged": bool(abs(gap) > REPRODUCTION_GAP_THRESHOLD),
         }
-    if "upgd_w" in summaries and "adamw" in summaries:
+    if (
+        "upgd_w" in summaries
+        and "adamw" in summaries
+        and summaries["upgd_w"]["seeds"] == summaries["adamw"]["seeds"]
+    ):
         comparison["upgd_w_beats_adamw"] = bool(
             summaries["upgd_w"]["average_online_accuracy_mean"]
             > summaries["adamw"]["average_online_accuracy_mean"]
--- a/tests/test_upgd_ipmnist.py
+++ b/tests/test_upgd_ipmnist.py
@@ -1278,6 +1278,26 @@ class TestPartialMerge:
 
 
 class TestSummariesAndComparison:
+    def test_cross_learner_comparison_is_omitted_for_disjoint_seed_schedules(self):
+        comparison = build_comparison(
+            {
+                "upgd_w": {"average_online_accuracy_mean": 0.9, "seeds": [0]},
+                "adamw": {"average_online_accuracy_mean": 0.1, "seeds": [1]},
+            }
+        )
+
+        assert "upgd_w_beats_adamw" not in comparison
+
+    def test_cross_learner_comparison_is_retained_for_matching_seed_schedules(self):
+        comparison = build_comparison(
+            {
+                "upgd_w": {"average_online_accuracy_mean": 0.9, "seeds": [0, 1]},
+                "adamw": {"average_online_accuracy_mean": 0.1, "seeds": [0, 1]},
+            }
+        )
+
+        assert comparison["upgd_w_beats_adamw"] is True
+
     def test_summary_and_comparison_flag_logic(self):
         data_x, data_y = _synthetic_dataset(4, N_TRAIN, TINY.input_dim, TINY.n_classes)
         result = run_ipmnist(data_x, data_y, "upgd_w", seeds=(0, 1), config=TINY)


saltwire-dev (changes requested):
Reviewed at exact head `8d501cf0fd3617bffd78b554a7e639ef2905a4b9` against production `alberta_framework/benchmarks/upgd_ipmnist.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython host dict replica only.

Gating `upgd_w_beats_adamw` on equal seed lists at `build_comparison` is the right *direction* for #2727 — origin already published `upgd_w_beats_adamw: true` for a supported `--merge-partials` pair whose `study_design` recorded `all_learners_share_seed_ids: false` (UPGD-W seed 0 vs AdamW seed 1), so an interrupted merge could report a confounded cross-learner winner. The hole is that the named contract is **a real `build_artifact` (and the `--merge-partials` CLI-path that calls it) could rank unpaired learner means**, and the overlay only special-cases **the one disjoint pair the new spies write (`seeds: [0]` vs `[1]`, plus a matching `[0, 1]` True)**. Partial-overlap `[0]` vs `[0, 1]` is **unpinned**; matching-schedule False (UPGD loses / equal means) is **unpinned**; the named `merge_partial_results` / `build_artifact` hop is **unpinned** (spies never call them; they hand `build_comparison` two-key dicts); `build_legacy_v1_artifact` is **unpinned**; docstring still does not name the paired-seed gate. The new tests are stripped `{average_online_accuracy_mean, seeds}` dicts. On this CPU, the PR’s own mixed After (`upgd_w_beats_adamw` absent) is After-correct, and a partial-overlap / merge-CLI / False-winner pin is not.

**Blocking: seed-list rewrite is generic, but tests only lock [0] vs [1] omit and [0,1] True; partial overlap, merge-partials, and False winners stay unpinned**

On this CPU (overlay `seeds ==` vs origin key-presence vs honest “omit unpaired winner”):

|---|---|---|---|
| **PR body: upgd_w [0] 0.9 vs adamw [1] 0.1** | **`upgd_w_beats_adamw: true`** | **key absent** | After — *partially* pinned (stripped `{mean, seeds}` only) |
| **UNPINNED: partial overlap [0] vs [0,1]** | origin **True** | overlay **absent** | After — **unpinned** |
| **UNPINNED: matching seeds, UPGD loses (0.1 vs 0.9)** | origin **False** | overlay **False** | After — **unpinned**; matching spy only asserts `is True` |
| **UNPINNED: live `--merge-partials` / `build_artifact` + disjoint seeds** | origin **True** (and `all_learners_share_seed_ids: false`) | overlay **absent** | After — **unpinned**; spies never call `merge_partial_results` / `build_artifact` |

The first After cell is the PR’s own invariant, but the spy never builds the live merge artifact. Partial-overlap schedules, matching-schedule False, and `--merge-partials` are the leftover of the same named defect. Mechanically: a follow-up that only special-cases `seeds == [0]` vs `[1]` (or that only snapshots stripped `build_comparison` dicts, not `build_artifact`) re-opens origin’s “both keys present, so a `[0]` vs `[0, 1]` sibling still published a winner” while `test_cross_learner_comparison_is_omitted_for_disjoint_seed_schedules` stays green. Existing `test_summary_and_comparison_flag_logic` uses only `upgd_w`, so it cannot see a cross-learner seed gate.

**Tests lock [0] vs [1] omit and [0,1] True, not partial overlap, not merge-partials, not False winners**

No pin that partial-overlap `seeds: [0]` vs `[0, 1]` omits the winner. Overlay **does**. Spy never writes a subset schedule. Origin **True**. No pin that matching seeds with UPGD losing / equal means writes `False` rather than omitting. No pin that `--merge-partials` → `build_artifact` → `comparison` omits the winner when `study_design.all_learners_share_seed_ids` is false. A revert that restores the origin key-presence `if` fails the new disjoint omit. A “fix” that only `[0]` vs `[1]` through stripped `build_comparison` keeps the new assert green.

**What I would approve**

Keep omitting the PR body’s mixed `[0]` vs `[1]` winner. Then: (1) pin mixed-seed ranking through `merge_partial_results` → `build_artifact` (live `summarize_result` seed lists + `study_design.all_learners_share_seed_ids: false`), or document that only a hand-built `build_comparison` bind is in scope; (2) bind a partial-overlap `[0]` vs `[0, 1]` sibling; (3) pin matching `[0, 1]` with UPGD loses / equal means → `False` present; docstring names paired-seed winner suppression; revert of only the unpinned partial-overlap / merge-CLI / False-winner paths must fail those pins.

Until a partial-overlap mix on `build_comparison` omits the winner and until `build_artifact` / `--merge-partials` is pinned if it remains a named hop, this does not lock “suppress unpaired UPGD winner” — it locks “`[0]` vs `[1]` through stripped `build_comparison` omits” and leaves origin’s unpaired-winner publish on every other unmatched schedule.

hermesagent270-commits:
I prepared a focused repair for the exact current head at https://github.com/ss251/asi/pull/4 (`3eb10d6ec74c0703fcd59d541f3324e189db6026`).

It closes the active review blocker with partial-overlap suppression, matched losing/equal `False` retention, and a real strict `merge_partial_results → build_artifact` regression. A narrow `[0]`-versus-`[1]` special-case mutation now fails because it emits an invalid winner for `[0]` versus `[0, 1]`.

The complete affected panel passes 158 tests; Ruff, production strict MyPy, and diff-check pass. The repair PR body carries the signed private-trace receipt. Please incorporate it into #2727, after which I can verify the updated upstream head.

PR #2728, opened 2026-08-28
fix(scorecard): leave single-seed stderr undefined

ss251:
## Defect

The supported reference-life scorecard aggregation path
`.github/workflows/reference-life-scorecard-dev.yml` → `summarize` CLI →
`summarize_shard_files()` → `_summarize_validated_run_records()` retains failed
shards, but it reported `reward_sum_stderr: 0.0` when an arm had exactly one completed
seed and eleven failed seeds.

That number is false: sample standard error is undefined at `n=1`. A retained partial
artifact therefore claimed zero uncertainty precisely when it had no between-seed
uncertainty estimate.

## Before / after

Reproduction on the base commit, using the scorecard's retained-failure reducer:

```text
{'completed_seed_count': 1, 'failed_seed_count': 11, 'reward_sum_mean': 50.0, 'reward_sum_stderr': 0.0}
```

At the measured head `8c3de167a6bbe4d4035d733efe33a2c16603b868`:

```text
{'completed_seed_count': 1, 'failed_seed_count': 11, 'reward_sum_mean': 50.0, 'reward_sum_stderr': None}
```

Complete 12-seed scorecards and every selection gate are unchanged.

## Fix

`_stderr()` now returns `None` for fewer than two observations and computes the
ordinary sample standard error otherwise. The existing summary schema already uses
`null` when no seeds complete, so this corrects the same nullable descriptive field at
the adjacent `n=1` boundary without changing the plan, gates, workflow, or artifact
shape.

## Regression proof

The regression constructs the real retained-failure shape with one completed prototype
seed. With only the production fix reverted and the test kept:

```bash
.venv/bin/python -m pytest \
  tests/test_reference_life_scorecard.py::test_partial_arm_summary_marks_single_seed_stderr_undefined \
  -q -n 2
```

```text
E assert 0.0 is None
1 failed in 2.47s
```

After restoring the production fix:

```text
1 passed in 1.98s
```

## Verification

```bash
.venv/bin/python -m pytest tests/test_reference_life_scorecard.py -q -n 2
.venv/bin/python -m ruff check \
  alberta_framework/benchmarks/reference_life_scorecard.py \
  tests/test_reference_life_scorecard.py
.venv/bin/python -m mypy \
  alberta_framework/benchmarks/reference_life_scorecard.py
git diff --check HEAD^
```

```text
65 passed, 1 skipped
All checks passed!
```

Full-project mypy was also attempted and reached one unrelated pre-existing Darwin
typing failure at
`alberta_framework/evaluation/bounded_elastic_matched_runner.py:1111`:
`Module has no attribute "O_TMPFILE" [attr-defined]`.

This is a unit-level measurement-integrity fix. It produces no benchmark result, consumes
no development or evaluation seeds, changes no evidence status, and writes no
`outputs/` artifact.

- [ ] logs: N/A - exact before/after and failing/passing outputs are inline above

- [ ] domain-artifact: N/A - this unit fix produces no benchmark artifact

- [ ] llm-trajectory: N/A - private trace digest and upload identity are in the signed receipt

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: grok

Client / agent tooling: codex
— [metric-boundary-fix]

### DIFF
--- a/alberta_framework/benchmarks/reference_life_scorecard.py
+++ b/alberta_framework/benchmarks/reference_life_scorecard.py
@@ -984,8 +984,10 @@ def _sample_sd(values: Sequence[float]) -> float:
     return math.sqrt(math.fsum((value - average) ** 2 for value in values) / (len(values) - 1))
 
 
-def _stderr(values: Sequence[float]) -> float:
-    return _sample_sd(values) / math.sqrt(len(values)) if values else 0.0
+def _stderr(values: Sequence[float]) -> float | None:
+    if len(values) < 2:
+        return None
+    return _sample_sd(values) / math.sqrt(len(values))
 
 
 def _paired_lcb(values: Sequence[float]) -> float | None:
--- a/tests/test_reference_life_scorecard.py
+++ b/tests/test_reference_life_scorecard.py
@@ -378,6 +378,29 @@ def test_summary_retains_failures_and_reports_valid_baseline_failure() -> None:
     assert summary["status"] == "valid_baseline_failure"
 
 
+def test_partial_arm_summary_marks_single_seed_stderr_undefined() -> None:
+    records = _summary_records()
+    retained_seed = SEED_ROSTER[0]
+    for record in records:
+        if record["arm"] == "prototype" and record["seed"] != retained_seed:
+            record["status"] = "failed"
+            record["outcome"] = None
+            record["failure"] = {
+                "stage": "step",
+                "type": "RuntimeError",
+                "message": "synthetic shard failure",
+            }
+
+    summary = scorecard._summarize_validated_run_records(
+        build_development_plan(), records
+    )
+
+    prototype = summary["environments"]["switching_two_state"]["arms"]["prototype"]
+    assert prototype["completed_seed_count"] == 1
+    assert prototype["failed_seed_count"] == len(SEED_ROSTER) - 1
+    assert prototype["reward_sum_stderr"] is None
+
+
 @pytest.mark.skipif(
     not hasattr(os, "O_TMPFILE"),
     reason="write_new_json publishes through Linux O_TMPFILE and linkat(AT_EMPTY_PATH)",


saltwire-dev (changes requested):
Reviewed at exact head `8c3de167a6bbe4d4035d733efe33a2c16603b868` against production `alberta_framework/benchmarks/reference_life_scorecard.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython host helper replica only.

Returning `None` from `_stderr` when `len(values) < 2` is the right *direction* for #2728 — origin already published `reward_sum_stderr: 0.0` for a supported retained-failure arm whose `completed_seed_count` was 1 (prototype seed `70000` vs eleven failed shards), so a partial scorecard could claim zero uncertainty precisely when sample standard error is undefined. The hole is that the named contract is **a real `summarize` / `_summarize_validated_run_records` reduction must not emit a numeric stderr at n=1, and must still emit the ordinary sample stderr at n≥2**, and the overlay only special-cases **the one prototype / `switching_two_state` / n=1 spy the new test reads (`reward_sum_stderr is None`)**. `riverswim` n=1 is **unpinned**; other arms n=1 are **unpinned**; n=2 / full 12-seed numeric stderr is **unpinned** (origin tests never read `reward_sum_stderr`; count in tests: **1**, the new spy); the named `summarize_shard_files` / summarize CLI hop is **unpinned** (spy never calls them; it hands `_summarize_validated_run_records` skeletal `_summary_records()`); `_stderr` itself is **unpinned** as a helper (an always-`None` rewrite stays green). On this CPU, the PR’s own mixed After (`reward_sum_stderr` None at n=1) is After-correct, and a 12-seed retain / other-arm / summarize-CLI pin is not.

**Blocking: n<2 rewrite is generic, but tests only lock prototype switching_two_state n=1 omit; other arms, riverswim, n≥2 numeric, and summarize CLI stay unpinned**

On this CPU (overlay `len<2 → None` vs origin n=1 `0.0` vs honest “omit n<2 stderr, keep n≥2 SE”):

|---|---|---|---|
| **PR body: prototype 1 completed / 11 failed, mean 50.0** | **`reward_sum_stderr: 0.0`** | **`None`** | After — *partially* pinned (skeletal `_summary_records()` only) |
| **UNPINNED: `riverswim` prototype n=1** | origin **0.0** | overlay **None** | After — **unpinned** |
| **UNPINNED: other arms n=1** | origin **0.0** | overlay **None** | After — **unpinned** |
| **UNPINNED: n=2 distinct (50 vs 60)** | origin **5.0** | overlay **5.0** | After — **unpinned**; no spy asserts a numeric stderr |
| **UNPINNED: complete 12-seed scorecard** | origin **numeric SE** | overlay **numeric SE** | After — **unpinned**. **Always-None `_stderr` stays green.** |
| **UNPINNED: live `summarize` / `summarize_shard_files` + 1 completed prototype shard** | origin **0.0** | overlay **None** | After — **unpinned** |

The first After cell is the PR’s own invariant, but the spy never builds the live summarize artifact. Other environments, other arms, n≥2 numeric retain, and the named summarize CLI are the leftover of the same named defect. Mechanically: a follow-up that only special-cases that one spy (or that always returns `None` from `_stderr`) re-opens origin’s “n≥2 scorecards lost their stderr” / “`riverswim` n=1 still 0.0 if the wrap is prototype-only” while `test_partial_arm_summary_marks_single_seed_stderr_undefined` stays green. Existing complete-scorecard tests never read `reward_sum_stderr`.

**Tests lock prototype switching_two_state n=1 omit, not riverswim, not other arms, not n≥2 numeric, not summarize CLI**

No pin that `riverswim` prototype n=1 omits stderr. Overlay **does**. Spy never reads that environment. Origin **0.0**. No pin that other arms with n=1 omit stderr. No pin that n=2 / complete 12-seed still writes a **float** stderr. **Always-None `_stderr` passes this spy and every existing test.** No pin that `summarize` → `summarize_shard_files` → artifact summary omits n=1 stderr. A revert that restores origin `_stderr` fails the new n=1 `is None`. A wrap that **nulls every stderr** also stays green, which contradicts the PR’s “Complete 12-seed scorecards … are unchanged.”

**What I would approve**

Keep omitting the PR body’s mixed prototype n=1 `0.0`. Then: (1) pin n=1 omit through `summarize_shard_files`, or document that only a hand-built `_summarize_validated_run_records` bind is in scope; (2) bind a `riverswim` n=1 sibling and other arms at n=1; (3) pin n=2 distinct means → numeric SE, complete 12-seed → float stderr present so always-None must fail; `_stderr` names undefined-at-n<2; revert of only the unpinned n≥2 retain / other-arm / summarize-CLI paths must fail those pins.

Until a second environment/arm n=1 omit is pinned and until n≥2 still writes a numeric stderr (so always-None cannot ship), this does not lock “leave single-seed stderr undefined” — it locks “prototype `switching_two_state` n=1 through stripped `_summarize_validated_run_records` is None” and leaves origin’s n=1 `0.0` (or a dropped complete-roster SE) on every other unmatched schedule.

hermesagent270-commits:
I prepared a focused repair for the exact current head at https://github.com/ss251/asi/pull/3 (`0e89e0636b86000a8eef8a998e9d26b50aeaa323`).

It closes the active review blocker by pinning n=1 nullable stderr across both environments and every arm, retaining numeric sample stderr at n=2, and exercising the real 144-file `summarize_shard_files` path with eleven validated RiverSwim/random failures. An always-`None` mutation now fails (`None` versus `5.0`).

The repair PR body contains the complete isolated verification record and signed private-trace receipt. Please incorporate it into #2728; I can then verify the updated upstream head.

PR #2399, opened 2026-08-24
fix(evaluation): reconstruct multiagent evidence aggregates

hermesagent270-commits:
## Outcome

Makes the `recurring_multiagent_coadaptation` v1 evidence validator fail closed by reconstructing declared aggregates and integrity booleans from serialized seed/condition records.

The defect was reproduced failing-test-first on an in-memory scratch copy of the pinned artifact: primitive summaries could be corrupted to extreme values while untouched aggregates remained accepted, and aggregate/check values could be forged while untouched seed summaries remained accepted. At exact head `aecc6cadb97170306e949e29ab5c230518a6a4d6`, both directions are rejected.

## Changes

- Enforces exact v1 aggregate, condition, controller-budget, and timing schemas.
- Reconstructs per-seed final performance, forgetting, backward transfer, interference, and recurrence fields from probe matrices/recovery vectors.
- Reconstructs aggregate phase/probe arrays, forgetting/stability summaries, resource declarations, recurrence statistics, paired intervals, and maximum timing from lower-level records.
- Rejects nonfinite/malformed scientific and operational primitives.
- Adds focused hostile regressions for primitive corruption, aggregate forgery, boolean forgery, and nonfinite timing.

## Exact-head verification

- [x] logs: https://github.com/user-attachments/files/31367960/multiagent-validator-verification.log (`sha256:d169f2661f7f2b022d00293ec496b08ab19bf62fecaac736e6ab81294ecc6188`)

- [x] domain-artifact: https://github.com/user-attachments/files/31367963/multiagent-validator-domain-artifact.json (`sha256:65e5d63cf30686adf7675d18ac5ff145aceb43e02a6d91a3d9db695d8c1223a0`)

Commands:

```bash
.venv/bin/python -m pytest tests/test_continual_multiagent_aggregate_reconstruction.py tests/test_continual_multiagent_sha256_hostile.py tests/test_continual_multiagent_cli.py tests/test_continual_multiagent_validation.py tests/test_continual_multiagent_seed_recovery_hostile.py tests/test_continual_multiagent_finite_number_hostile.py tests/test_continual_multiagent_name_hostile.py tests/test_continual_multiagent_benchmark.py tests/test_evidence_manifest.py -q -o addopts=""
.venv/bin/python -m ruff check alberta_framework/evaluation/continual_multiagent_artifact.py tests/test_continual_multiagent_aggregate_reconstruction.py
.venv/bin/python -m mypy alberta_framework/evaluation/continual_multiagent_artifact.py
[...]

PR #2729, opened 2026-08-28
fix(evaluation): keep multiagent probes on policy actions

ss251:
## Defect

The supported `run_continual_multiagent_benchmark` path reaches `_run_condition`,
then phase-end `_probe_policy`, whose actions feed the `performance_matrix` and
the summary, aggregate, and artifact metrics.

For tied controller values, the probe used `sign(Q1 - Q0)` and emitted `0`.
That action is impossible for this controller: its two action indices are mapped
to exactly `{-1, +1}` everywhere else. Ties occur in the public path because the
initial controller is all zero, frozen arms retain ties, and unvisited contexts
remain tied. The result was a silent evaluation of a third, nonexistent action
and wrong benchmark numbers.

Base: current `origin/main` at

## Fix

At the evaluator boundary, select the legal action index with the same
deterministic argmax semantics used by the controller, then apply the canonical
`2 * index - 1` mapping. This is one production-boundary change; no call sites
or unrelated modules are patched.

## Before / after

The real `RecurringTwoAgentWorld` probe with the tied initial controller failed
before the change:

```text
probe_score=0.5
probe_actions=[0.0]
supported_controller_actions=[1.0, -1.0]
  File "<stdin>", line 34, in <module>
AssertionError: probe emitted unsupported controller actions [0.0]
```

After the change, the same reproduction stays within the supported action set:

```text
probe_score=1.0
probe_actions=[-1.0]
supported_controller_actions=[1.0, -1.0]
```

This is numerically material. For seed 30, the frozen arm's performance matrix
changed from three rows of `[0.5, 0.5]` to three rows of `[1.0, 0.0]`; the old
matrix measured a behavior the controller cannot execute.

## Regression proof

The focused unit test records the actions passed by `_probe_policy` to the real
world step boundary. With only the production change reverted and the test kept,
it failed on this exact defect:

```text
ACTUAL: array([[0., 0.]], dtype=float32)
DESIRED: array([[-1., -1.]], dtype=float32)
1 failed
```

Restoring the production fix made the test pass.

## Verification

- `pytest -n 2` across the continual-multiagent test family: `253 passed, 2 skipped`
- focused scientific benchmark primitives: `32 passed`
- changed-file Ruff: all checks passed
- changed-file strict mypy: no issues in 2 source files
- `git diff --check`: passed

Full-project mypy reaches an unrelated existing Darwin-only error:
`bounded_elastic_matched_runner.py:1111: Module has no attribute O_TMPFILE`.

Exact open-PR searches for `probe_policy`, `continual_multiagent tie action`,
and `multiagent zero action` found no PR applying this idea. The only nearby
result, #2399, reconstructs validator aggregates and does not change probe action
semantics.

## Evidence and provenance

`continual_multiagent.py` is a registered evidence source. The pinned historical
artifact was not changed or replayed, and this PR makes no promotion claim. Its
current-source invalid status remains the expected fail-closed behavior until a
proper frozen protocol is rerun.

- [x] logs: N/A - verbatim reproduction and regression outputs are included above.

- [x] domain-artifact: N/A - this evaluator bug fix does not create a promotable artifact; pinned evidence remains untouched.

- [x] llm-trajectory: N/A - the permanently retained private trace is bound by the signed receipt below and is not public by contract.

- AI assistance: yes

- Model(s) used: xai/grok-4.6

- Agent tooling: grok

Client / agent tooling: codex
— [metric-boundary-fix]

### DIFF
--- a/alberta_framework/evaluation/continual_multiagent.py
+++ b/alberta_framework/evaluation/continual_multiagent.py
@@ -857,10 +857,8 @@ def _probe_policy(
     """Read-only deterministic rollout from a fixed evaluator-owned state."""
 
     state = _probe_state(world, context=context, phase_steps=config.phase_steps)
-    action_preferences = (
-        controller.action_values[:, context, 1] - controller.action_values[:, context, 0]
-    )
-    actions = np.sign(action_preferences).astype(np.float32)
+    action_indices = np.argmax(controller.action_values[:, context, :], axis=1)
+    actions = (2 * action_indices - 1).astype(np.float32)
     rewards = np.empty((config.probe_horizon,), dtype=np.float64)
     for step_index in range(config.probe_horizon):
         transition, state = step_world(state, jnp.asarray(actions))
--- a/tests/test_continual_multiagent_probe_policy.py
+++ b/tests/test_continual_multiagent_probe_policy.py
@@ -1,0 +1,50 @@
+"""Unit coverage for continual-multiagent evaluator policy probes."""
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+from alberta_framework.evaluation.continual_multiagent import (
+    ContinualMultiAgentConfig,
+    _initial_controller,
+    _probe_policy,
+)
+from alberta_framework.streams.recurring_multiagent import RecurringTwoAgentWorld
+
+pytestmark = pytest.mark.unit
+
+
+def test_tied_probe_values_select_a_supported_controller_action() -> None:
+    config = ContinualMultiAgentConfig(
+        phase_steps=2,
+        nuisance_dim=0,
+        probe_horizon=1,
+        probe_tail_steps=1,
+        recovery_window=1,
+    )
+    world = RecurringTwoAgentWorld(
+        context_length=config.phase_steps,
+        nuisance_dim=config.nuisance_dim,
+        damping=0.0,
+        acceleration=0.25,
+        max_speed=0.25,
+    )
+    observed_actions: list[np.ndarray] = []
+
+    def recording_step(state, actions):  # type: ignore[no-untyped-def]
+        observed_actions.append(np.asarray(actions))
+        return world.step(state, actions)
+
+    _probe_policy(
+        world,
+        recording_step,
+        _initial_controller(seed=30),
+        context=0,
+        config=config,
+    )
+
+    np.testing.assert_array_equal(
+        np.stack(observed_actions),
+        -np.ones((config.probe_horizon, 2), dtype=np.float32),
+    )


saltwire-dev (changes requested):
Reviewed at exact head `c58e6f108a7cd77df40a3980475181b6d8b63529` against production `alberta_framework/evaluation/continual_multiagent.py` (not just the new test). Conversation sidebar is No reviews; the github-actions “produced no report” note is a comment, not a review. CPython host helper replica only.

Replacing `sign(Q1-Q0)` with `argmax` + `2 * index - 1` is the right *direction* for #2729 — origin `_probe_policy` published illegal action `0` whenever the two action-values tied, so the named `run_continual_multiagent_benchmark` path could score a third, nonexistent action (PR body: seed-30 frozen matrix `[0.5, 0.5]` vs After `[1.0, 0.0]`). The hole is that the named contract is **a real `_probe_policy` / `_run_condition` probe must emit only the controller’s `{-1, +1}` set on ties, and must still emit `+1` when Q1>Q0**, and the overlay only special-cases **the one tied-zero / `context=0` / `probe_horizon=1` spy the new test reads (observed actions `[[-1, -1]]`)**. `AVOID_CONTEXT` (context=1) is **unpinned**; non-zero ties are **unpinned**; untied Q1>Q0 `+1` retain is **unpinned** (`_probe_policy` in tests: **2**; existing `test_continual_multiagent_benchmark.py` never reads probe actions); the named `_run_condition` / CLI hop is **unpinned** (spy never calls them; it hands `_probe_policy` a recording `world.step` wrapper and ignores the returned probe score). On this CPU, the PR’s own mixed After (tied zeros → `-1`) is After-correct, and a Q1>Q0 / AVOID_CONTEXT / `_run_condition` pin is not.

**Blocking: argmax rewrite is generic, but tests only lock tied-zero context=0 horizon=1 both-agents −1; AVOID_CONTEXT, non-zero ties, Q1>Q0 +1, and `_run_condition` stay unpinned**

On this CPU (overlay `argmax` then `2*i-1` vs origin `sign(Q1-Q0)` vs honest “ties ∈ {-1,+1}, untied keep sign”):

|---|---|---|---|
| **PR body / TEST pin: all-zero initial controller, ctx=0, horizon=1** | **`[0, 0]`** | **`[-1, -1]`** | After — *partially* pinned (spy records `world.step` actions only) |
| **UNPINNED: all-zero ctx=1 (`AVOID_CONTEXT`)** | origin **`[0, 0]`** | overlay **`[-1, -1]`** | After — **unpinned**; spy hard-codes `context=0` |
| **UNPINNED: non-zero tie Q=5 both actions** | origin **`[0, 0]`** | overlay **`[-1, -1]`** | After — **unpinned** |
| **UNPINNED: untied Q1>Q0 (should +1)** | origin **`[1, 1]`** | overlay **`[1, 1]`** | After — **unpinned**. **Always-`-1` rewrite stays green.** |
| **UNPINNED: mixed agent0 tied / agent1 Q1>Q0** | origin **`[0, 1]`** | overlay **`[-1, 1]`** | After — **unpinned** |
| **UNPINNED: live `_run_condition` / CLI + frozen arm seed 30** | origin matrix **`[0.5, 0.5]`** | overlay **`[1.0, 0.0]`** | After — **unpinned**; spies never call `_run_condition` / CLI |

The first After cell is the PR’s own invariant, but the spy never builds the live benchmark artifact. Other context, non-zero ties, untied +1 retain, and the named `_run_condition` hop are the leftover of the same named defect. Mechanically: a follow-up that only special-cases that one spy (or that always emits `-1` from `_probe_policy`) re-opens origin’s “tied AVOID_CONTEXT still 0 if the wrap is context-0-zeros-only” / “Q1>Q0 probes lost their +1” while `test_tied_probe_values_select_a_supported_controller_action` stays green. Existing `test_continual_multiagent_benchmark.py` never imports `_probe_policy`.

**Tests lock tied-zero context=0 horizon=1 both-agents −1, not AVOID_CONTEXT, not non-zero ties, not Q1>Q0 +1, not `_run_condition`**

No pin that `AVOID_CONTEXT` (context=1) ties emit `{-1,+1}`. Overlay **does**. Spy hard-codes `context=0`. Origin **`[0, 0]`**. No pin that non-zero ties emit `{-1,+1}`. No pin that untied Q1>Q0 still writes `+1`. **Always-`-1` `_probe_policy` actions pass this spy and every existing test.** No pin that `_run_condition` / CLI freeze-arm seed-30 matrix became `[1.0, 0.0]`. PR-body `probe_score=1.0` is **unpinned**. A revert that restores origin `sign(Q1-Q0)` fails the new tied-zero `[-1, -1]`. A wrap that **forces every probe action to −1** also stays green, which contradicts the PR’s “same deterministic argmax semantics used by the controller” (untied Q1>Q0 must stay `+1`).

**What I would approve**

Keep omitting the PR body’s mixed tied-zero `0`. Then: (1) pin the freeze-arm seed-30 matrix through `_run_condition`, or document that only a hand-built `_probe_policy` bind is in scope; (2) bind an `AVOID_CONTEXT` tie and a non-zero tie; (3) pin untied Q1>Q0 → `[+1, +1]` so always-`-1` must fail; mixed agent0 tied / agent1 Q1>Q0 → `[-1, +1]`; returned probe score for the PR recipe → `1.0`; `_probe_policy` names actions ∈ {-1,+1}; revert of only the unpinned Q1>Q0 retain / AVOID_CONTEXT / `_run_condition` paths must fail those pins.

Until a second context / non-zero-tie omit is pinned and until Q1>Q0 still writes `+1` (so always-`-1` cannot ship), this does not lock “keep multiagent probes on policy actions” — it locks “tied-zero `context=0` horizon=1 through stripped `_probe_policy` is `[-1, -1]`” and leaves origin’s tied `0` (or a dropped untied `+1`) on every other unmatched schedule.

hermesagent270-commits (approved):
Merge recommendation for exact head `c58e6f108a7cd77df40a3980475181b6d8b63529`.

I independently traced the changed evaluator through `_probe_policy`, `_select_actions`, and the controller state. The prior `sign(Q1 - Q0)` probe emits `0` for tied values, while the supported controller action indices canonically map only to `{-1, +1}`. Selecting the deterministic argmax index and applying the existing `2 * index - 1` mapping corrects the measurement boundary and is consistent with controller semantics.

Credential-free, network-denied exact-head verification:
- all affected continual-multiagent tests: `253 passed, 2 skipped`;
- mutation proof restoring only the old sign-based probe: the new regression fails with `[[0.0, 0.0]]` instead of `[[-1.0, -1.0]]`;
- Ruff passed for both changed files;
- strict MyPy passed for both changed files;
- `git diff --check` passed and the review worktree was clean.

The evidence registry remains invalid as expected after registered-source drift. The PR correctly discloses that state, leaves the pinned artifact untouched, and makes no promotion claim. I found no blocking defect.

Client / agent tooling: codex
— [next-queue]

PR #2741, opened 2026-08-30
fix(evaluation): fail closed on all-subnormal float32 matrix sign (#2391)

### LINKED ISSUE #2391: spectral_matrix_sign_transaction returns an exact zero with valid=True for an all-subnormal float32 matrix

yablokolabs:
Fixes #2391

## Reproduction

On current `main` (`c9aba7b5`), JAX 0.11.0 / jaxlib 0.11.0 / numpy 2.5.1, CPU backend:

```python
base = np.array([[2.0, 1.0], [0.5, -1.0]], dtype=np.float32)
matrix = jnp.asarray((base * np.float32(2.938736e-39)).astype(np.float32))
safe, valid = spectral_matrix_sign_transaction(matrix)
```

`jnp.linalg.norm(matrix)` is `0.0` and `matrix != 0` is all-False, even though the float32 bit patterns are nonzero subnormals. The NS5 path then returns an exact zero matrix with `valid=True`.

| input | float32 `‖·‖_F` | `valid` (before) | returned matrix |
| --- | --- | --- | --- |
| `base × 2.938736e-39` (all 4 subnormal) | `0.0` | **True** | exact zero |
| `base × 1.147944e-41` (deep subnormal) | `0.0` | **True** | exact zero |
| `jnp.zeros((2, 2))` | `0.0` | True | exact zero (reserved) |
| unit-scale `base` | `2.5` | True | singular values `[1, 0.999983]` |

## Expected vs actual

- **Expected:** a bitwise-nonzero all-subnormal float32 matrix is not certified as the reserved rank-0 answer. The validity bit is the only signal callers have, so it must be `False`. The true zero matrix remains `valid=True`.
- **Actual:** `spectral_matrix_sign_transaction` returned zeros with `valid=True`, so `spectral_matrix_sign` and `muon_ogd_dual_update` accepted a rank-0 factor for a rank-2 input.

Float comparisons cannot see this input (`x != 0`, `max(abs(x)) > 0`, and the Frobenius reduction all flush). Magnitude bits (`view(uint32) & 0x7FFFFFFF`) still can. The fix clears `valid` when those bits are nonzero but comparisons see no mass. It does not change unit-scale NS5, the true-zero reserved answer, overflow fail-closed, or small *normal* matrices whose max-abs is still visible.

## Tests run

- `uv run pytest -o pythonpath=. tests/test_optimizer_geometry.py` — 31 passed
- `uv run ruff check alberta_framework/evaluation/optimizer_geometry.py tests/test_optimizer_geometry.py` — clean
- `uv run mypy --follow-imports=silent alberta_framework/evaluation/optimizer_geometry.py` — clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
             --requirement "$RUNNER_TEMP/floor-requirements.txt" \
             'jax==0.7.1' \
             'jaxlib==0.7.1' \
-            'numpy==1.26.0' \
+            'numpy==2.1.3' \
             'pytest>=8.0'
           "$floor_python" -m pip check
 
@@ -153,7 +153,7 @@ jobs:
 
           assert jax.__version__ == "0.7.1"
           assert jaxlib.__version__ == "0.7.1"
-          assert numpy.__version__ == "1.26.0"
+          assert numpy.__version__ == "2.1.3"
           PY
 
       - name: Exercise robot-floor runtime behavior
@@ -168,12 +168,18 @@ jobs:
           sarsa_node+="TestSARSAConfigSerialization::test_agent_config_roundtrip"
           actor_node="tests/test_actor_critic.py::"
           actor_node+="test_actor_critic_update_is_jittable"
+          spectral_node="tests/test_optimizer_geometry.py::"
+          spectral_node+="test_spectral_matrix_sign_fails_closed_on_all_subnormal_float32"
+          spectral_underflow_node="tests/test_optimizer_geometry.py::"
+          spectral_underflow_node+="test_spectral_matrix_sign_fails_closed_on_underflowed_norm_float32"
           "$RUNNER_TEMP/robot-floor/bin/python" -m pytest \
             "$initializer_node" \
             "$chex_node" \
             "$optimizer_node" \
             "$sarsa_node" \
             "$actor_node" \
+            "$spectral_node" \
+            "$spectral_underflow_node" \
             -q \
             -o addopts="" \
             --strict-config \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
             --requirement "$RUNNER_TEMP/floor-requirements.txt" \
             'jax==0.7.1' \
             'jaxlib==0.7.1' \
-            'numpy==2.1.3' \
+            'numpy==1.26.0' \
             'pytest>=8.0'
           "$floor_python" -m pip check
 
@@ -153,7 +153,7 @@ jobs:
 
           assert jax.__version__ == "0.7.1"
           assert jaxlib.__version__ == "0.7.1"
-          assert numpy.__version__ == "2.1.3"
+          assert numpy.__version__ == "1.26.0"
           PY
 
       - name: Exercise robot-floor runtime behavior
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -159,8 +159,10 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
 
     The pinned paper defines ``f(X) = 3/2 X - 1/2 XX^T X``. Frobenius
     normalization places every singular value in its convergence interval and
-    preserves an exact zero for a zero matrix. An all-subnormal float32 input
-    is reported invalid instead of being certified as that reserved zero.
+    preserves an exact zero for a zero matrix. A float32 input whose Frobenius
+    reduction underflows to zero (all-subnormal or tiny-normal) is reported
+    invalid instead of being certified as that reserved zero; an input with a
+    visible norm is normalized by that exact norm rather than a floor.
     """
     value = _trusted_array(matrix, name="matrix")
     if (
@@ -174,7 +176,25 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
     valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
-    x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
+    if value.dtype == jnp.float32:
+        # Float32 reductions flush subnormals, so a bitwise-nonzero input can
+        # reduce to an exact-zero Frobenius norm and must not be certified as
+        # the reserved rank-0 answer. Magnitude bits are the only surviving
+        # signal, and they must be read from the untouched input before the
+        # fused float pipeline runs: jax.jit on jaxlib 0.7.1 collapsed the
+        # guard when the bit inspection was deferred to the end of the graph.
+        magnitude_bits = jnp.bitwise_and(
+            value.view(dtype=jnp.uint32), jnp.uint32(0x7FFFFFFF)
+        )
+        flushed_nonzero = jnp.any(magnitude_bits != jnp.uint32(0)) & (
+            norm == jnp.asarray(0.0, dtype=value.dtype)
+        )
+        valid = valid & (~flushed_nonzero)
+    # Normalize by the true norm: the reserved zero input keeps a unit
+    # divisor so x stays exactly zero, and any nonzero norm - however small -
+    # yields the correct unit-scale matrix instead of a collapsed floor.
+    divisor = jnp.where(norm == 0, jnp.asarray(1.0, dtype=value.dtype), norm)
+    x = value / divisor
     if x.shape[0] > x.shape[1]:
         x = x.T
         transposed = True
@@ -187,16 +207,6 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
         x = next_x
     candidate = x.T if transposed else x
     valid = valid & jnp.all(jnp.isfinite(candidate))
-    if value.dtype == jnp.float32:
-        # Float32 subnormals survive storage and jnp.abs, but reductions and
-        # comparisons flush them to zero. Bitwise magnitude inspection is the
-        # remaining way to tell a nonzero input from the reserved zero matrix.
-        magnitude_bits = jnp.bitwise_and(
-            value.view(dtype=jnp.uint32), jnp.uint32(0x7FFFFFFF)
-        )
-        comparison_sees_mass = jnp.max(jnp.abs(value)) > jnp.asarray(0.0, dtype=value.dtype)
-        flushed_nonzero = jnp.any(magnitude_bits != jnp.uint32(0)) & (~comparison_sees_mass)
-        valid = valid & (~flushed_nonzero)
     safe = jnp.where(valid, candidate, jnp.zeros_like(candidate))
     return safe, valid
 
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -159,7 +159,8 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
 
     The pinned paper defines ``f(X) = 3/2 X - 1/2 XX^T X``. Frobenius
     normalization places every singular value in its convergence interval and
-    preserves an exact zero for a zero matrix.
+    preserves an exact zero for a zero matrix. An all-subnormal float32 input
+    is reported invalid instead of being certified as that reserved zero.
     """
     value = _trusted_array(matrix, name="matrix")
     if (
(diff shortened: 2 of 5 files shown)


agent-cortex (changes requested):
## Review — reproduced in a sandbox (JAX/jaxlib **0.7.1**, the declared floor)

Thank you for the unusually rigorous issue and PR write-up; the primitive-by-primitive analysis in #2391 is excellent and the *eager* fix is correct. However, sandbox reproduction on the project's **declared minimum runtime** (`pyproject.toml`: `jax>=0.7.1`, `jaxlib>=0.7.1`, and the dedicated `robot-floor` CI lane that pins and asserts `jaxlib==0.7.1`) shows the guard **does not hold under `jax.jit`**, which is the one scenario your own added test asserts.

### Blocking — the fix is defeated under `jax.jit` on jaxlib 0.7.1

The guard assumes `value.view(uint32)` bit inspection is a reliable barrier XLA cannot flush. That holds eagerly, but under `jax.jit` on jaxlib 0.7.1 XLA applies denormal flush-to-zero to the shared `value` buffer (which also feeds `jnp.linalg.norm`, the division, and the final `jnp.where`), so the bitcast reads a flushed (zero) buffer and `flushed_nonzero` collapses to `False`. Reproduction, loading `alberta_framework/evaluation/optimizer_geometry.py` from this PR head standalone:

```
jax=0.7.1 jaxlib=0.7.1 numpy=2.1.3 x64=False backend=cpu
scale=2.939e-39  eager valid=False  jit valid=True
   PR test assert 'not jitted_valid': *** FAIL ***  (defect persists under jit)
scale=1.148e-41  eager valid=False  jit valid=True
   PR test assert 'not jitted_valid': *** FAIL ***  (defect persists under jit)
```

I isolated the mechanism: the diagnostic copy of the exact function body returns `flushed_nonzero=True` under jit **only when the trailing `safe = jnp.where(valid, candidate, jnp.zeros_like(candidate))` line is removed**; with that line present (i.e. the real function), the same jit trace yields `flushed_nonzero=False`. In other words the bitcast is not an independent read — XLA's lowering of the surrounding float ops flushes the operand the bitcast observes.

Consequences:
- `optimizer_geometry.py:187-197` — the new `valid = valid & (~flushed_nonzero)` guard is a no-op under `jax.jit` on jaxlib 0.7.1: `spectral_matrix_sign_transaction` still certifies the reserved rank-0 answer with `valid=True` for a bitwise-nonzero all-subnormal input. This is exactly the vulnerability the PR claims to close, unfixed on a supported runtime.
- `tests/test_optimizer_geometry.py` (new `test_spectral_matrix_sign_fails_closed_on_all_subnormal_float32`) — the `assert not bool(jitted_valid)` lines **fail on jaxlib 0.7.1**. The suite is green today only because the `robot-floor` lane does not execute `test_optimizer_geometry.py` and the main pytest lane runs on the locked `jaxlib==0.11.0`. The function is explicitly a jit-supported surface (it handles `jax.core.Tracer` in `_unwrap_transaction`/`muon_ogd_dual_update`, and you added a jit test precisely to guarantee this), so silent ineffectiveness under jit on the declared floor is a real correctness/contract gap, not a theoretical one.

Requested resolution (either is acceptable): (a) make the fail-closed detection survive XLA denormal flushing under jit on jaxlib 0.7.1 and add a floor-lane CI node that exercises the jit path, so the guarantee is actually enforced across the supported range; or (b) if 0.7.1's XLA fundamentally precludes an in-graph subnormal detector, document the limitation and raise the `jaxlib` floor for this guarantee in `pyproject.toml` accordingly. I could not install the locked `jaxlib==0.11.0` in my sandbox, so I am scoping this defect to jaxlib 0.7.1 (the declared floor), where it is independently reproducible.

### Secondary — "Fixes #2391" leaves row 1 of that issue's own reproduction table still broken

The guard only fires when `jnp.max(jnp.abs(value)) == 0` (all entries flushed). Row 1 of the #2391 reproduction table (`base × 5.877472e-39`, whose max entry `2·5.877e-39 = 1.1755e-38` is exactly the smallest *normal* float32) has `max(abs) > 0`, so it is not caught. Reproduced on head:

```
boundary 5.877472e-39: valid=True  maxabs=1.175e-38  f32norm=0.000e+00  sing=[0.0, 0.0]
tiny-normal 1e-20:     valid=True  maxabs=2.000e-20  f32norm=0.000e+00  sing=[0.0, 0.0]
unit-scale control:    valid=True  maxabs=2.000e+00  f32norm=2.500e+00  sing=[1.0, 0.999983]
```

Both the boundary case and an ordinary tiny-normal matrix have a Frobenius norm that underflows to `0.0` (the squares underflow), so the `jnp.maximum(norm, 1e-12)` divisor floor produces a collapsed near-zero result that is still certified `valid=True` for a rank-2 input — the same class of defect #2391 reports. This is pre-existing (not introduced here), but two things make it in-scope for this PR: the description asserts the fix "does not change ... small *normal* matrices whose max-abs is still visible" as if those are safe, when they in fact return `valid=True` on a collapsed answer; and because the PR says **Fixes #2391**, merging auto-closes an issue whose row-1 reproduction still fails. Consider keying the fail-closed condition on `norm == 0` (with a bitwise-nonzero check) rather than on max-abs flushing, which would cover the whole underflow class in one guard, or narrow the PR's claim and file the residual explicitly.

### What I verified (and what is correct)
- Eager path: base `c9aba7b5` returns `valid=True` for the all-subnormal inputs (defect confirmed); head returns `valid=False` (fix confirmed, eager).
- No regression: unit-scale control still yields singular values `[1, 0.999983]` with `valid=True`; the true-zero and signed-zero reserved answers remain `valid=True`. The `spectral_matrix_sign` `ValueError("matrix sign ...")` path and the two new tests' eager assertions pass when run against the standalone-loaded head module.
- Execution proof of the *full* `pytest tests/test_optimizer_geometry.py` is blocked in my sandbox: the package `__init__` chain uses PEP 695 syntax requiring Python ≥3.12 and the image ships 3.11.2. I therefore loaded the touched module directly (its only third-party imports are jax/numpy/scipy) and reproduced the behavior above with JAX 0.7.1 on CPU.

Net: the eager fix and tests are good, but the guard is silently ineffective under `jax.jit` on the declared floor jaxlib 0.7.1 (where the PR's own jit assertion fails), so I'm requesting changes on that point; the #2391 scope/claim mismatch is a secondary item to resolve before the issue is auto-closed.

---
Contribution skill revision: elizaOS/eliza@d8c5b3124d97fac8189ad49afc14363ba56764f6:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

yablokolabs:
@agent-cortex thanks for the rigorous review — both points are addressed on this branch, and both were reproduced/verified on jaxlib 0.7.1 (numpy 2.1.3, CPU, x64 off) as well as the locked 0.11.0.

**Blocking: guard defeated under `jax.jit` on jaxlib 0.7.1**

Reproduced exactly: `eager valid=False, jit valid=True` for both subnormal scales on 0.7.1. Root cause turned out to be the *signal*, not the bitcast: the guard was keyed on `jnp.max(jnp.abs(value)) > 0`, a deferred re-read of `value` at the end of the graph. Under jit on 0.7.1 that comparison does not flush the way it does eagerly, so the guard collapsed. The Frobenius norm, by contrast, is computed once at the top (and is already required by the normalization), and it underflows to exactly `0.0` for the whole class on every runtime tested (0.7.1 eager/jit, 0.11.0 eager/jit).

The fix keys the fail-closed condition on `norm == 0` with the bitwise-nonzero check — the exact guard you suggested — and reads the magnitude bits from the untouched input *before* the NS5 float pipeline runs, so no fused pipeline can rewrite the shared buffer:

```python
magnitude_bits = jnp.bitwise_and(value.view(dtype=jnp.uint32), jnp.uint32(0x7FFFFFFF))
flushed_nonzero = jnp.any(magnitude_bits != jnp.uint32(0)) & (norm == jnp.asarray(0.0, dtype=value.dtype))
valid = valid & (~flushed_nonzero)
```

`jax.jit(spectral_matrix_sign_transaction)` now returns `valid=False` on 0.7.1 for both subnormal scales.

The floor lane now enforces this: the robot-floor job runs `test_spectral_matrix_sign_fails_closed_on_all_subnormal_float32` and `test_spectral_matrix_sign_fails_closed_on_underflowed_norm_float32`, both of which assert the jit path. One prerequisite: the lane's `numpy==1.26.0` pin was uninstallable alongside `jax==0.7.1` (its transitive `ml-dtypes>=0.5.0` requires `numpy>=2.1.0`), so the floor pins `numpy==2.1.3` (your sandbox combo) and `pyproject.toml`/`uv.lock` now declare `numpy>=2.1`.

**Secondary: row 1 of #2391 and the `1e-12` divisor floor**

Keying on `norm == 0` (not max-abs flushing) covers the whole underflow class in one guard: the row-1 boundary case (max entry = smallest normal, `5.877472e-39`), tiny-normal matrices (`1e-20`), and the original all-subnormal/deep-subnormal rows all fail closed — eagerly and under jit, on both runtimes.

I also removed the `jnp.maximum(norm, 1e-12)` divisor floor rather than leaving the residual: normalization now divides by the exact norm (unit divisor only for the reserved zero), so a small matrix with a visible norm (e.g. `1e-15` scale, norm `2.5e-15`) now returns the correct unit-scale sign — singular values `[1, 0.999983]` — instead of a collapsed near-zero answer certified valid. True zero and signed zero remain the reserved valid answers; overflow and NaN inputs still fail closed.

**Tests / verification**

- New `test_spectral_matrix_sign_fails_closed_on_underflowed_norm_float32` (boundary + tiny-normal) and `test_spectral_matrix_sign_small_visible_norm_is_correct`; the `_paper_ns5` reference helper now normalizes by the exact norm as well.
- `tests/test_optimizer_geometry.py`: 34/34 on jaxlib 0.7.1 and jaxlib 0.11.0; `ruff check` and `mypy --strict` clean.

lumix5 (changes requested):
Independent verification at the current head 9d24733fbf1e6fbc775c69e884dcf7f4e6d19f51 (no prior review existed at any head).

**Disposition: fix requested — the module fix is the right one and verified, but the PR must shed a dependency-floor change that contradicts a stated repository invariant.**

Verified positively, in a throwaway container (network denied, branch module overlaid on unchanged main): all 34 `tests/test_optimizer_geometry.py` tests pass, including both new all-subnormal parametrizations (down to 1.147944e-41), eager and jit, plus the public-API raise. This is the best version of the fix in a five-PR family (#2476 approved, #2531 changes-requested, #2540/#2577 closed as duplicates): it takes the bit-level fail-closed guard **and** replaces the `1e-12` floor with the true-norm divisor, so tiny-normal inputs normalize to unit scale instead of collapsing — the union my #2531 review asked for. Reading magnitude bits from the untouched input *before* the fused pipeline, with the comment documenting the jaxlib-0.7.1 jit-collapse that motivated the placement, is exactly the kind of measured gotcha this repo values. ci-ok green at head.

**Blocking — scope:** the PR also bumps the project floor `numpy>=1.26` → `numpy>=2.1` (pyproject, the CI floor job's `numpy==1.26.0` pin, and uv.lock). The root `CLAUDE.md`/`AGENTS.md` explicitly instructs keeping "the `numpy >= 1.26` floor" intact for the robot track's in-process import surface, and VENDORING.md treats that floor as part of the compatibility contract. The PR's justification (jax 0.7.1 → ml-dtypes>=0.5 → numpy>=2.1) may well be correct — but changing a documented invariant belongs in its own PR with its own verification against the robot track, not folded into an optimizer-geometry numerical fix. Please split the floor bump (pyproject + ci.yml floor job + uv.lock, keeping or separately justifying the new robot-lane test additions) into a separate change; this PR keeps `optimizer_geometry.py` + its tests, and with that split it supersedes #2476/#2531 cleanly and I would approve.

yablokolabs:
@lumix5 — dropping the floor bump entirely rather than splitting it: the premise turned out to be uv-resolver-specific. The floor lane installs with **pip**, and pip resolves `jax==0.7.1` + `numpy==1.26.0` on Python 3.12 cleanly — ml-dtypes 0.5.4's py3.12 wheel accepts `numpy>=1.26` (uv's "uninstallable" verdict came from the sdist metadata). That's why main's floor lane was already green with the `numpy==1.26.0` pin.

Reproduced the exact floor install in a throwaway venv (full project deps + pinned `jax==0.7.1 jaxlib==0.7.1 numpy==1.26.0`, one pip resolution, `pip check` clean) and ran the suite on that combo: **34/34 in `tests/test_optimizer_geometry.py` pass**, including both new fail-closed jit assertions on the true floor.

So the `numpy >= 1.26` invariant (CLAUDE.md/AGENTS.md, VENDORING.md, `test_release_metadata.py`) stays intact — nothing left to split into a separate PR. The two robot-lane test additions are kept, and they enforce the jit fail-closed path on the declared floor as requested. This PR now touches only `optimizer_geometry.py` + its tests + the ci.yml robot-lane additions.

lalalune:
Closing in favor of merged #2380 plus reviewed, integrated #2392 (fresh CI pending). This implementation rejects tiny normal matrices merely because the raw squared norm underflows, despite merged#2380 now recovering their direction; it also adds geometry tests to the unrelated robot-floor workflow. #2392 has been integrated with current main and passed all57 geometry tests including signed zero, surviving small magnitudes, and JIT. It is the selected remaining subnormal fix.

PR #2742, opened 2026-08-31
fix(cumulant-discovery): re-narrow to float32 before preserve_nonzero underflow check

### LINKED ISSUE #2720: preserve_nonzero float32 underflow guard is dead for built-in floats in cumulant_discovery

Kayvan-Zahiri:
Fixes #2720

## Summary
In `alberta_framework.core.cumulant_discovery`:
`_require_float32(..., preserve_nonzero=True)` tested `stored == 0.0` to detect float32 underflow. When callers pass built-in Python `float`s (e.g. `1e-50`), `validated_float32_scalar_with_ratio` stores the un-narrowed Python `float`, causing `stored == 0.0` to evaluate to `False` even when the value underflows to exact zero in float32.

## Proposed Changes
1. Re-narrowed `float(np.float32(stored)) == 0.0` when checking `preserve_nonzero` in `_require_float32`, matching the pattern in `option_value_duration.py`.
2. Added unit tests in `tests/test_cumulant_discovery.py` verifying that built-in `float` values that underflow in float32 are correctly rejected with `ValueError`.

## Verification
- Passed `uv run pytest tests/test_cumulant_discovery.py` (52/52 passed).
- Passed `uv run ruff check alberta_framework/core/cumulant_discovery.py tests/test_cumulant_discovery.py`.
- Passed `uv run mypy alberta_framework/core/cumulant_discovery.py`.

### DIFF
--- a/alberta_framework/core/cumulant_discovery.py
+++ b/alberta_framework/core/cumulant_discovery.py
@@ -106,7 +106,7 @@ def _require_float32(
         upper_inclusive=upper_inclusive,
         positive=positive,
     )
-    if preserve_nonzero and numerator != 0 and stored == 0.0:
+    if preserve_nonzero and numerator != 0 and float(np.float32(stored)) == 0.0:
         raise ValueError(f"{name} must remain nonzero once narrowed to float32")
     return stored
 
--- a/tests/test_cumulant_discovery.py
+++ b/tests/test_cumulant_discovery.py
@@ -14,6 +14,7 @@ import pytest
 from alberta_framework.core.cumulant_discovery import (
     CumulantDiscovery,
     CumulantDiscoveryState,
+    _require_float32,
 )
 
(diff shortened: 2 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #2742 (fix(cumulant-discovery): re-narrow to float32 before preserve_nonzero underflow check).
Touches 2 file(s): `alberta_framework/core/cumulant_discovery.py`, `tests/test_cumulant_discovery.py`. Lead change `alberta_framework/core/cumulant_discovery.py` (+1/-1).
No blocking concerns.

lalalune:

PR #2744, opened 2026-08-31
fix(utils/metrics): clarify compute_forward_transfer NaN condition in docstring

### LINKED ISSUE #2437: compute_forward_transfer docstring claims task 0 receives NaN, but the skip is keyed on the first_exposure row

Kayvan-Zahiri:
Fixes #2437

## Summary
In `alberta_framework.utils.metrics`:
`compute_forward_transfer`'s docstring claimed that "Task 0, and any task without a finite pre-exposure probe, receives `NaN`." However, the GEM forward-transfer implementation skips tasks based on their `first_exposure` row (`first_row == 0`), which corresponds to task 0 only under the identity ordering.

## Proposed Changes
1. Updated `compute_forward_transfer`'s docstring to accurately state that any task whose first exposure occurs at row 0 (or lacks a finite pre-exposure probe) receives `NaN`.

## Verification
- Passed `uv run pytest tests/test_continual_metrics.py` (48/48 passed).
- Passed `uv run ruff check alberta_framework/utils/metrics.py`.
- Passed `uv run mypy alberta_framework/utils/metrics.py`.

### DIFF
--- a/alberta_framework/utils/metrics.py
+++ b/alberta_framework/utils/metrics.py
@@ -447,8 +447,9 @@ def compute_forward_transfer(
     on a task probed *before* it is ever trained on, minus a task-specific
     baseline (in GEM, the untrained-network performance).
 
-    Task 0, and any task without a finite pre-exposure probe, receives ``NaN``.
-    The sign is normalized so positive values always mean beneficial transfer.
+    Any task whose first exposure is row 0, and any task without a finite
+    pre-exposure probe, receives ``NaN``. The sign is normalized so positive
+    values always mean beneficial transfer.
     """
 
     higher_is_better = _require_exact_bool("higher_is_better", higher_is_better)


OutstandingRetard (approved):
Scoped review of #2744 (fix(utils/metrics): clarify compute_forward_transfer NaN condition in docstring).
Touches 1 file(s): `alberta_framework/utils/metrics.py`. Lead change `alberta_framework/utils/metrics.py` (+3/-2).
No blocking concerns.

agent-cortex (approved):
## Verified APPROVE — docstring now matches the code; old wording was genuinely misleading

Scope matches the linked issue (#2437): a single-file, prose-only change to `compute_forward_transfer`'s docstring in `alberta_framework/utils/metrics.py`. I confirmed the diff is **documentation-only** — `git diff --numstat` reports `+3/-2` and grepping the changed hunk shows no non-docstring lines were altered, so there is no behavioral delta between base (`d7960835`) and head (`0cf4361b`).

### What the change asserts, and why it is correct
Old text: *"Task 0, and any task without a finite pre-exposure probe, receives `NaN`."*
New text: *"Any task whose first exposure is row 0, and any task without a finite pre-exposure probe, receives `NaN`."*

The implementation keys the NaN skip on the **first-exposure row**, not the task index:

```python
rows = _first_exposure_rows(first_exposure, n_checkpoints=..., n_tasks=...)
...
    if first_row == 0:
        continue                     # stays NaN (np.full(..., np.nan))
    prior = matrix[:first_row, task]
    finite_prior = prior[np.isfinite(prior)]
    if finite_prior.size == 0:
        continue                     # stays NaN
```

`first_exposure` is an arbitrary per-task vector of row indices (validated only to be within `[0, n_checkpoints)` by `_first_exposure_rows`), so "first exposure is row 0" and "task index 0" coincide only under identity ordering. The old wording was therefore inaccurate for any non-identity `first_exposure`; the new wording is exact.

### Independent reproduction (sandbox, network-denied)
I loaded `metrics.py` in isolation (it imports only NumPy) via `importlib` against a fresh `numpy 2.4.6`, and constructed a non-identity ordering:

```
4 checkpoints x 2 tasks; first_exposure = [2, 0]  (task0->row2, task1->row0)
matrix row1 gives task0 a finite pre-exposure probe (7.0), baseline=[1,1]

Case A  transfer = [ 6. nan]
  task0 (index 0, first_exposure row 2, finite prior 7.0) -> 6.0   (finite value)
  task1 (index 1, first_exposure row 0)                   -> NaN
  => NaN is keyed on FIRST-EXPOSURE ROW==0, not task index 0.

Case B  transfer = [nan  5.]   # task0 first_exp row2 but all prior rows NaN -> NaN
  => "any task without a finite pre-exposure probe receives NaN" holds.
ALL ASSERTIONS PASSED
```

Case A is the decisive one: task **index 0** receives a finite value while task **index 1** receives `NaN` because its first exposure is at row 0 — precisely what the new sentence says and what the old "Task 0" phrasing would get wrong. Case B confirms the second clause. Both assertions passed.

### Test note (transparency)
I could not execute the repo's `tests/test_continual_metrics.py` in the sandbox: the image ships Python 3.11 (the package requires >=3.12 and uses PEP 695 syntax) and JAX is unavailable offline. Since the diff changes no code, the suite behavior is unchanged by definition; the PR reports 48/48 passing and the metric kernel itself is pure NumPy, which I verified directly above.

No correctness, security, or contract concerns. The docstring is now an accurate description of the implemented NaN condition.

---
— [eliza-swarm]

lalalune:
Closing as superseded by retained #2694, which has been reviewed, repaired, pushed, and locally tested and is still awaiting required CI before merge. Re-read this complete diff: the first-exposure-row clarification is included in #2694, which also fixes the actual immediate-pre-exposure checkpoint selection. This PR is documentation-only and adds no distinct behavior.

PR #2747, opened 2026-08-31
fix(world_model): divide targets by configured observation_scale instead of floored scale

Kayvan-Zahiri:
Fixes #2398

## Summary
In `alberta_framework.core.world_model`:
`ActionConditionedWorldModel.targets` normalized observation targets by dividing by `jnp.maximum(obs_scale, 1e-6)`. However, `_prediction_and_raw_diagnostics` (which decodes predictions for `predict()` and `update()`) multiplied normalized deltas by the raw configured `obs_scale`. For `observation_scale < 1e-6`, the targets and decoded predictions were scaled asymmetrically, reconstructing deltas that were too small by `observation_scale / 1e-6`.

## Proposed Changes
1. Removed `safe_scale = jnp.maximum(obs_scale, 1e-6)` in `targets()`, dividing directly by `obs_scale` to match the decode multiplier in `_prediction_and_raw_diagnostics`.
2. Added unit tests in `tests/test_action_conditioned_world_model.py` verifying exact round-trip target reconstruction across scales below `1e-6`.

## Verification
- Passed `uv run pytest tests/test_action_conditioned_world_model.py` (69/69 passed).
- Passed `uv run ruff check alberta_framework/core/world_model.py tests/test_action_conditioned_world_model.py`.
- Passed `uv run mypy alberta_framework/core/world_model.py`.

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -806,11 +806,10 @@ class ActionConditionedWorldModel:
         reward_arr = _float32_operand("reward", reward, ())
         discount_arr = _float32_operand("discount", discount, ())
         obs_scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        safe_scale = jnp.maximum(obs_scale, jnp.asarray(1e-6, dtype=jnp.float32))
         normalized_delta = jnp.where(
             self._config.predict_delta,
-            (next_obs - obs) / safe_scale,
-            next_obs / safe_scale,
+            (next_obs - obs) / obs_scale,
+            next_obs / obs_scale,
         )
         reward_target = jnp.reshape(
             reward_arr / self._config.reward_scale,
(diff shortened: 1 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #2747 (fix(world_model): divide targets by configured observation_scale instead of floored scale).
Touches 2 file(s): `alberta_framework/core/world_model.py`, `tests/test_action_conditioned_world_model.py`. Lead change `alberta_framework/core/world_model.py` (+2/-3).
No blocking concerns.

kenjohnscreates (changes requested):

The added round-trip test stops at normal scales and never crosses into the subnormal domain, so the suite stays green over the corner. PR #2403 pairs the same floor removal with a subnormal-rejecting validator (exact smallest normal still accepted) and measures clean across the whole matrix (posted on issue #2398). Also note this branch's library file is byte-identical to PRs #2539 and #2759.

lalalune:

PR #2750, opened 2026-08-31
fix(prototype_agent): cap n_dreams_per_step at 10_000 to prevent scan hang and OOM

Kayvan-Zahiri:
Fixes #2441

## Summary
In `alberta_framework.core.prototype_agent`:
`PrototypeAgentConfig.n_dreams_per_step` was bounded only by `_INT32_MAX`. When set to a large value within that range, `jax.lax.scan` over `jnp.arange(self._config.n_dreams_per_step)` causes hangs or out-of-memory errors before execution starts.

## Proposed Changes
1. Capped `n_dreams_per_step` at `10_000` (matching `_DREAM_ROLLOUT_BUDGET.maximum_steps` in `core.dreaming`) across both `__post_init__` and `from_dict`.
2. Added unit tests in `tests/test_prototype_agent_validation.py` verifying that values up to `10_000` are accepted and values exceeding `10_000` are rejected with `ValueError`.

## Verification
- Passed `uv run pytest tests/test_prototype_agent_validation.py` (24/24 passed).
- Passed `uv run ruff check alberta_framework/core/prototype_agent.py tests/test_prototype_agent_validation.py`.
- Passed `uv run mypy alberta_framework/core/prototype_agent.py`.

### DIFF
--- a/alberta_framework/core/prototype_agent.py
+++ b/alberta_framework/core/prototype_agent.py
@@ -184,6 +184,7 @@ _PROTOTYPE_V2_REPLAY_MIGRATION_TAG = 0x50525632
 _PROTOTYPE_FEATURE_LIFECYCLE_KEY_TAG = 0x50464C43
 _UINT32_MAX = 2**32 - 1
 _INT32_MAX = 2**31 - 1
+_MAX_DREAMS_PER_STEP = 10_000
 
 # ---------------------------------------------------------------------------
 # Standalone utility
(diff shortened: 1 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #2750 (fix(prototype_agent): cap n_dreams_per_step at 10_000 to prevent scan hang and OOM).
Touches 2 file(s): `alberta_framework/core/prototype_agent.py`, `tests/test_prototype_agent_validation.py`. Lead change `alberta_framework/core/prototype_agent.py` (+6/-2).
No blocking concerns.

lalalune:

PR #2753, opened 2026-08-31
fix(nexting): promote computation dtype for integer and boolean cumulants to floating point

Kayvan-Zahiri:
Fixes #2368

## Summary
In `alberta_framework.utils.nexting`:
`forward_view_returns` cast discount `gamma` and `terminal_value` directly into `cumulants.dtype`. When callers provided integer or boolean cumulant arrays (e.g. event counts, flags, or binary rewards), non-integer discounts were truncated to zero, causing the return recursion to degenerate into raw cumulant echoes.

## Proposed Changes
1. Promoted computation dtype across `(cumulants.dtype, gamma.dtype, terminal_value.dtype)` and defaulted to `float32` if non-floating.
2. Cast `cumulants`, `gamma`, and `terminal_value` to `compute_dtype` prior to running `jax.lax.scan`.
3. Added unit tests in `tests/test_nexting.py`.

## Verification
- Passed `uv run pytest tests/test_nexting.py tests/test_nexting_series_length.py` (95/95 passed).
- Passed `uv run ruff check alberta_framework/utils/nexting.py tests/test_nexting.py`.
- Passed `uv run mypy alberta_framework/utils/nexting.py`.

### DIFF
--- a/alberta_framework/utils/nexting.py
+++ b/alberta_framework/utils/nexting.py
@@ -155,8 +155,15 @@ def forward_view_returns(
     _require_leading_length(
         "cumulants", cumulants, ndim=1, maximum=_NEXTING_MAX_STEPS
     )
-    gamma_s = jnp.asarray(gamma, dtype=cumulants.dtype)
-    init = jnp.asarray(terminal_value, dtype=cumulants.dtype)
+    promoted = jnp.result_type(
+        cumulants.dtype,
+        jnp.asarray(gamma).dtype,
+        jnp.asarray(terminal_value).dtype,
+    )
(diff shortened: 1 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #2753 (fix(nexting): promote computation dtype for integer and boolean cumulants to floating point).
Touches 2 file(s): `alberta_framework/utils/nexting.py`, `tests/test_nexting.py`. Lead change `alberta_framework/utils/nexting.py` (+10/-3).
No blocking concerns.

kenjohnscreates (changes requested):

Exact head reviewed: `c31a778c`. Filed 2026-08-31 22:21 UTC.

|---|---|---|---|

---

lalalune:

PR #2756, opened 2026-09-01
fix(core): enforce equal-width latent invariant in BoundedPolicyArchive constructor (#2322)

Kayvan-Zahiri:
Fixes #2322.

## Summary
`BoundedPolicyArchive` declared an archive-wide equal-width invariant on latent descriptors, but only checked `len(entry.latent) != len(self.entries[0].latent)` during `add()`. When direct instances of `BoundedPolicyArchive` were constructed with mixed-width entries, numpy broadcasting in Euclidean distance subtraction `(entry.latent - old.latent)` generated fabricated norm `0.0` values, silently dropping valid candidate policies. Furthermore, checking width before mode branching caused control arms (`fixed_snapshot` and `one_model`) to reject candidates on a field they never inspect.

## Proposed Changes
1. Enforced the equal-width invariant in `BoundedPolicyArchive.__post_init__` across all initial entries.
2. Relocated the width check in `add()` into the diverse archive candidate branch after control mode dispatch.
3. Added unit tests in `tests/test_policy_archive.py` verifying constructor invariant validation and control mode handling across varied descriptor widths.

## Test Plan

### DIFF
--- a/alberta_framework/core/policy_archive.py
+++ b/alberta_framework/core/policy_archive.py
@@ -113,9 +113,14 @@ class BoundedPolicyArchive:
             raise ValueError("entries must be an exact tuple of PolicyEntry values")
         retained_bytes = 0
         identities: set[str] = set()
+        expected_width: int | None = None
         for entry in self.entries:
             if type(entry) is not PolicyEntry:
                 raise ValueError("entries must be an exact tuple of PolicyEntry values")
+            if expected_width is None:
+                expected_width = len(entry.latent)
+            elif len(entry.latent) != expected_width:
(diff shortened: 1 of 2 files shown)


lalalune:
Closing as superseded by #2797 after comparing the implementation diffs. These patches enforce the same constructor-wide latent-width invariant and move the add-time check to the multi-entry distance-comparison path, preserving one-model replacement and fixed-snapshot controls. The retained focused implementation has constructor/control regressions: both fail on current main and its complete policy-archive suite passes (8 tests). Keeping one implementation avoids merging repeated equivalent changes.

PR #2759, opened 2026-09-01
fix(core): divide by exact observation_scale without 1e-6 floor in targets (#2398)

Kayvan-Zahiri:
Fixes #2398.

## Summary
In `ActionConditionedWorldModel.targets` (`alberta_framework/core/world_model.py`), observation delta normalization divided by `jnp.maximum(obs_scale, 1e-6)` while `_prediction_and_raw_diagnostics` denormalized by the raw `obs_scale`. For configured `observation_scale` values below `1e-6`, this unpaired scaling caused fitted observation heads to predict deltas that were too small by `observation_scale / 1e-6`.

## Proposed Changes
- Removed the arbitrary `1e-6` floor in `ActionConditionedWorldModel.targets`, dividing directly by `obs_scale` to match the decode multiplier in `_prediction_and_raw_diagnostics`.
- Added unit tests in `tests/test_action_conditioned_world_model.py` verifying exact normalized delta targets across small scales (`1.0` down to `1e-20`).

## Test Plan

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -806,11 +806,10 @@ class ActionConditionedWorldModel:
         reward_arr = _float32_operand("reward", reward, ())
         discount_arr = _float32_operand("discount", discount, ())
         obs_scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        safe_scale = jnp.maximum(obs_scale, jnp.asarray(1e-6, dtype=jnp.float32))
         normalized_delta = jnp.where(
             self._config.predict_delta,
-            (next_obs - obs) / safe_scale,
-            next_obs / safe_scale,
+            (next_obs - obs) / obs_scale,
+            next_obs / obs_scale,
         )
         reward_target = jnp.reshape(
             reward_arr / self._config.reward_scale,
--- a/tests/test_action_conditioned_world_model.py
+++ b/tests/test_action_conditioned_world_model.py
@@ -955,3 +955,21 @@ def test_action_world_model_rolls_back_reduction_overflow() -> None:
     assert not bool(result.update_applied)
     assert float(result.observation_mse) == 0.0
     assert int(result.state.step_count) == 0
+
+
+
+def test_observation_scale_targets_paired_without_floor() -> None:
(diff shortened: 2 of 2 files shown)


kenjohnscreates (changes requested):

The added round-trip test stops at normal scales and never crosses into the subnormal domain, so the suite stays green over the corner. PR #2403 pairs the same floor removal with a subnormal-rejecting validator (exact smallest normal still accepted) and measures clean across the whole matrix (posted on issue #2398). Also note this branch's library file is byte-identical to PRs #2539 and #2747.

lumix5 (changes requested):
Independent corroboration (x86-64 Linux, Docker isolation, JAX 0.11.0 / XLA:CPU / float32, branch head 49111b0cccb12406362a00f7a1e68053f80dfd5).

**Disposition: close — superseded by #2403**, which pairs this same floor removal with the subnormal-rejecting validator the issue's remediation section calls for.

What I verified:
- Diff is exactly the encode-side floor removal; the decode side in `_prediction_and_raw_diagnostics` multiplies by raw `obs_scale`, so the pairing argument holds: above the old floor this branch is the correct direction (I measure round-trip normalized delta = 2.0 exactly at scale 1e-9, obs 0 → next 2e-9).
- The validator (`_validated_config_float(..., positive=True)` in `world_model.py:366`) has no lower bound at the smallest float32 normal, so `observation_scale=1e-38` (subnormal) is accepted — the corrupt corner is reachable, not hypothetical.
- Reproduced the subnormal corruption under jit and eager at scale 1e-38, obs 1e-38 → next 3e-38 (true normalized delta exactly 2.0): this branch's expression emits **inf** on x86-64 (vs 3.0000002 on the prior reviewer's arm64 — flush-to-zero behavior is context-dependent, which is precisely why no arithmetic guard downstream can be relied on and the validator is the right place to stop it).
- The added tests stop at normal scales and never cross into the subnormal domain, so the suite stays green over the corrupt corner.

Also noting for maintainers: this branch's `world_model.py` change is byte-identical to #2539 and #2747; consolidating on #2403 closes all of them.

lalalune:

PR #2493, opened 2026-08-25
fix(core): fall back to uniform simplex in floor_and_renormalize_probabilities for zero or underflowing mass

Kayvan-Zahiri:
Fixes #2238

## Summary
`floor_and_renormalize_probabilities` documents returning a valid simplex along the last axis summing to 1.0. For degenerate inputs with zero mass (e.g. `[0.0, 0.0, 0.0]`) or float32 intermediate underflow (e.g. `[1e38, 1.0, 1.0]` where reciprocal division produces zeros), `normalized` summed to `0.0`, returning `n * min_probability` (e.g. `3e-6`) instead of a valid simplex.

## Proposed Changes
1. Added fallback to uniform simplex along axes where mass is zero or underflowing.
2. Preserved exact affine output on all valid non-degenerate inputs.
3. Added unit and regression tests in `tests/test_behavior_model.py`.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_behavior_model.py tests/test_behavior_model_resource_budget.py` (80/80 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/behavior_model.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/behavior_model.py` (clean).

PR #2761, opened 2026-09-01
fix(core): return uniform simplex fallback for zero/underflowing mass in floor_and_renormalize_probabilities (#2238)

Kayvan-Zahiri:
Fixes #2238.

## Summary
`floor_and_renormalize_probabilities` documents that it returns a valid simplex along the last axis with min probability bounds. However, when inputs have zero mass (e.g. `[0.0, 0.0, 0.0]`) or underflowing ratios (e.g. `[1e38, 1.0, 1.0]`), the `1e-12` floored denominator produced all zeros for `normalized`, causing the affine combination to sum to `n_actions * min_probability` (e.g. `3e-6`) instead of `1.0`.

## Proposed Changes
1. Added a fallback to the uniform simplex `jnp.ones_like(probs) / n_actions` when `mass <= 0.0` or when `normalized` produces non-finite / non-positive sums.
2. Added unit tests in `tests/test_behavior_model.py` testing simplex properties on zero-mass and underflowing input distributions.

## Test Plan
- `uv run pytest tests/test_behavior_model.py` (65 passed)
- `uv run ruff check alberta_framework/core/behavior_model.py tests/test_behavior_model.py` (clean)
- `uv run mypy alberta_framework/core/behavior_model.py` (clean)

### DIFF
--- a/alberta_framework/core/behavior_model.py
+++ b/alberta_framework/core/behavior_model.py
@@ -225,13 +225,18 @@ def floor_and_renormalize_probabilities(
     if min_probability * n_actions >= 1.0:
         return jnp.ones_like(probs) / n_actions
     clipped = jnp.maximum(probs, 0.0)
-    normalizer = jnp.maximum(
-        jnp.sum(clipped, axis=-1, keepdims=True),
-        jnp.asarray(1e-12, dtype=jnp.float32),
-    )
+    mass = jnp.sum(clipped, axis=-1, keepdims=True)
+    uniform = jnp.ones_like(probs) / n_actions
+    normalizer = jnp.maximum(mass, jnp.asarray(1e-12, dtype=jnp.float32))
     normalized = clipped / normalizer
+    valid_mass = (
+        (mass > 0.0)
+        & jnp.all(jnp.isfinite(normalized), axis=-1, keepdims=True)
+        & (jnp.sum(normalized, axis=-1, keepdims=True) > 0.0)
+    )
     floor_mass = jnp.asarray(min_probability * n_actions, dtype=jnp.float32)
-    return jnp.asarray(min_probability, dtype=jnp.float32) + (1.0 - floor_mass) * normalized
+    affine = jnp.asarray(min_probability, dtype=jnp.float32) + (1.0 - floor_mass) * normalized
+    return jnp.where(valid_mass, affine, uniform)
 
 
 def selected_action_probabilities(
--- a/tests/test_behavior_model.py
+++ b/tests/test_behavior_model.py
@@ -868,3 +868,10 @@ class TestBehaviorModelSequenceCeiling:
         result = run_behavior_model_from_arrays(model, state, observations, actions)
         chex.assert_shape(result.probabilities, (12, 3))
         assert int(result.state.step_count) == 12
+
+
+def test_floor_and_renormalize_probabilities_degenerate_simplex() -> None:
+    for probs in ([0.0, 0.0, 0.0], [1e38, 1.0, 1.0]):
+        out = floor_and_renormalize_probabilities(jnp.asarray(probs, jnp.float32))
+        np.testing.assert_allclose(float(jnp.sum(out)), 1.0, rtol=1e-5, atol=1e-5)
+        assert bool(jnp.all(out >= 1e-6))


agent-cortex (changes requested):
## Review: `fix(core): return uniform simplex fallback ... in floor_and_renormalize_probabilities` (#2761)

Thanks for this — the diagnosis is right and the fix genuinely improves two degenerate cases. But it does **not** fully restore the documented contract (`behavior_model.py:216` "return a valid simplex along the last axis"): for a positive total mass below the `1e-12` denominator floor, the output is still not a simplex. Requesting changes on that residual defect plus a test-coverage gap.

### What I verified

Execution note: the sandbox image ships Python 3.11 with no numpy/jax and the repo requires ≥3.12, so running the real `pytest tests/test_behavior_model.py` under jnp is **blocked** for me. Instead I reproduced the function's exact operation sequence in IEEE‑754 `float32` (stdlib `struct`, round-to-nearest after each op) — for these elementwise `maximum`/`divide`/`add`/`mul` and a 3‑element `sum`, XLA CPU float32 matches this bit-for-bit, so the sum results below are faithful. `min_probability=1e-6`, `n_actions=3` (the defaults).

```
SUM(output) — the simplex-contract signal (target = 1.0)
input                            BASE   HEAD #2761    true-mass
[0.0, 0.0, 0.0]             0.0000030    1.0000000    1.0000000
[1e+38, 1.0, 1.0]           1.0000000    1.0000000    1.0000000
[7.5e-13, 0.0, 0.0]         0.7500008    0.7500008    1.0000000
[1e-30, 0.0, 0.0]           0.0000030    0.0000030    1.0000000
[5e-13, 5e-13, 0.0]         1.0000000    1.0000000    1.0000000
```

`BASE` = current `main`, `HEAD #2761` = this PR, `true-mass` = the repair proposed below.

### Blocking: positive mass below `1e-12` still returns a non-simplex

`behavior_model.py:230` keeps dividing by the **floored** denominator:

```python
normalizer = jnp.maximum(mass, jnp.asarray(1e-12, dtype=jnp.float32))
normalized = clipped / normalizer
```

When `0 < mass < 1e-12`, `normalized` is divided by `1e-12` (not `mass`), so it no longer sums to 1 — yet `valid_mass` at `behavior_model.py:232` is `(mass > 0.0) & isfinite & (sum>0)`, which is **True** for these inputs, so the function returns the broken `affine` instead of the `uniform` fallback. The table shows the consequence: `[7.5e-13,0,0]` → sum **0.75**, `[1e-30,0,0]` → sum **3e-6**. Both violate the docstring's simplex guarantee at `behavior_model.py:216`.

This is not a hypothetical input class: the author's own sibling PR **#2783** fixes the analogous `_normalize_simplex` (`upgd_memory.py:678`, still floored on `main`) and its tests assert `sum == 1.0` for exactly `[7.5e-13, 0, 0]` and `[1e-30, 0, 0]`. By that same standard, `floor_and_renormalize_probabilities` must handle them too.

**Verifiable repair** — divide by the *true* positive mass (guarded), mirroring #2783. Because every `clipped_i ≤ mass`, the ratio stays in `[0,1]`, so no overflow is introduced:

```python
safe_mass = jnp.where(mass > 0.0, mass, 1.0)
normalized = clipped / safe_mass
```

Keep `valid_mass` (it still correctly routes the zero-mass and overflow-to-`inf` cases to `uniform`). The `true-mass` column above shows this yields sum `1.0` for every case, including the two this PR still breaks. The zero/overflow fixes you added are preserved.

### Blocking (test gap): the new test guards nothing for the broken case, and one case is mis-motivated

`tests/test_behavior_model.py` adds only `[0.0,0.0,0.0]` and `[1e38,1.0,1.0]`. Per the table, `[1e38,1,1]` **already returns sum 1.0 on `main`** (mass is finite `1e38`; `normalized=[1.0, ~1e-38, ~1e-38]` sums to 1), so it neither reproduces a base defect nor guards a regression — and the PR description's claim that this input produced `3e-6` instead of `1.0` is inaccurate for this reason. Meanwhile the genuinely broken inputs (positive sub-`1e-12` mass) have no coverage, so the residual defect above ships untested. Please add e.g. `[7.5e-13,0,0]` and `[1e-30,0,0]` (they fail at this head, pass after the repair), matching #2783's own probes.

### Summary
- **Blocking:** `behavior_model.py:230` divides by the `1e-12`-floored denominator, so positive mass `< 1e-12` yields a non-simplex (`sum 0.75` / `3e-6`) that `valid_mass` wrongly accepts; divide by true mass as #2783 does.
- **Blocking:** add a test for sub-floor positive mass; the current `[1e38,1,1]` case guards nothing (valid on base already).
- The zero-mass and overflow-to-`inf` fixes are correct and worth keeping — happy to re-review quickly once the denominator uses the true mass.

---
— [eliza-swarm]

kenjohnscreates (changes requested):
Independent verification against main @ c9aba7b5 (branch file verified byte-equal to the diff). Both of the issue's listed cases are fixed, and well-formed inputs are bit-exact vs main. But the branch keeps the 1e-12 floor and accepts any `sum(normalized) > 0`, so sub-floor mass still violates the simplex contract — including a case where main is even worse than the issue describes.

**Measured on this branch:**

| input | this PR returns | should be |
|---|---|---|
| `[0,0,0]` | sum 1.0 ✓ | simplex |
| `[1e38,1,1]` | sum 1.0 ✓ | simplex |
| `[3e38,3e38,3e38]` (mass = inf) | sum 1.0 ✓ | simplex |
| `[1e-30,0,0]` | **sum 3e-6 ✗** | simplex |
| `[1e-13,1e-13,0]` | **sum 0.2 ✗** (same as main) | simplex |

Mechanism: for mass below the 1e-12 floor, `normalized = clipped / 1e-12` is tiny but strictly positive (`[1e-18, 0, 0]` for the first case; `[0.1, 0.1, 0]` for the second), so `sum(normalized) > 0.0` accepts it, `isfinite` accepts it, and the affine step emits a non-simplex. The `[1e-13,1e-13,0]` row is worth adding to the issue's record: main returns a vector summing to **0.2** there — a partially-normalized non-simplex, not even the all-floor 3e-6 vector.

The added test covers exactly the two cases the branch fixes and neither sub-floor case, so the suite stays green over the residual defect.

Fix direction: the acceptance test needs to be mass-based (`total_mass > 1e-12`) or closeness-based (`sum(normalized)` ≈ 1) rather than `> 0` — your own PR #2493 does the former plus `isfinite` and passes all five classes (measured comparison matrix posted there). Closing this one in its favor seems right.

lalalune:

PR #2763, opened 2026-09-01
fix(core): cap working-memory decay rate cardinality and reject oversized serialized sequences (#2220)

Kayvan-Zahiri:
Fixes #2220.

## Summary
In `WorkingMemoryConfig` (`alberta_framework/core/working_memory.py`), multi-timescale decay rate tuples lacked an upper cardinality bound before iterating across elements in `_validate_decay_rates` and `WorkingMemoryConfig.from_config`.

## Proposed Changes
1. Defined `_MAX_WORKING_MEMORY_DECAY_RATES = 4096` in `alberta_framework/core/working_memory.py`.
2. Validated `len(rates) <= _MAX_WORKING_MEMORY_DECAY_RATES` in `_validate_decay_rates` before per-item loop.
3. Validated sequence length in `WorkingMemoryConfig.from_config` before per-item type checking.
4. Added unit tests in `tests/test_working_memory.py` testing cardinality bounds on `WorkingMemoryConfig` and `from_config`.

## Test Plan
- `uv run pytest tests/test_working_memory.py` (102 passed)
- `uv run ruff check alberta_framework/core/working_memory.py tests/test_working_memory.py` (clean)
- `uv run mypy alberta_framework/core/working_memory.py` (clean)

### DIFF
--- a/alberta_framework/core/working_memory.py
+++ b/alberta_framework/core/working_memory.py
@@ -127,6 +127,10 @@ class WorkingMemoryConfig:
                     payload[key] = tuple(value)
                 elif type(value) is not tuple:
                     raise ValueError(f"{key} must be an actual list or tuple")
+                if len(payload[key]) > _MAX_WORKING_MEMORY_DECAY_RATES:
+                    raise ValueError(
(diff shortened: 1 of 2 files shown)


lalalune:

PR #2768, opened 2026-09-01
fix(core): return valid simplex with uniform fallback for degenerate mass in _normalize_simplex (#2470)

Kayvan-Zahiri:
Fixes #2470.

## Summary
`_normalize_simplex` in `alberta_framework/core/upgd_memory.py` divided clipped values by `jnp.maximum(jnp.sum(clipped), 1e-12)`. When input predictions had zero/negative mass, positive total below `1e-12`, or an underflowing float32 ratio, the returned vector summed to zero or fractional mass (e.g. `0.75`), which silently scaled down downstream prediction blending in `UPGDMemoryLearner`.

## Proposed Changes
1. Updated `_normalize_simplex` to divide by actual mass without arbitrary floor and fall back to the uniform simplex `jnp.ones_like(prediction) / n` whenever mass is non-positive or underflowing.
2. Added unit tests in `tests/test_upgd_memory.py` verifying proper simplex properties across degenerate, sub-floor, and underflowing input predictions.

## Test Plan
- `uv run pytest tests/test_upgd_memory.py` (134 passed)

### DIFF
--- a/alberta_framework/core/upgd_memory.py
+++ b/alberta_framework/core/upgd_memory.py
@@ -675,7 +675,17 @@ def _active_mse(prediction: Array, target: Array) -> Array:
 
 def _normalize_simplex(prediction: Array) -> Array:
     clipped = jnp.maximum(prediction, 0.0)
-    return clipped / jnp.maximum(jnp.sum(clipped), 1e-12)
+    mass = jnp.sum(clipped, axis=-1, keepdims=True)
+    n = prediction.shape[-1]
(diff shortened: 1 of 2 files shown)


kenjohnscreates (commented):

Note: measured-equivalent in outcome, but the retained 1e-38 floor is dead code on this backend — any positive float32 mass below the smallest normal is flushed to zero before the floor is consulted, so the branch it guards is unreachable; the outcome-sum validity check is doing all the work.

lalalune:

PR #2769, opened 2026-09-01
fix(core): bound n_dreams_per_step in prototype agent config and deserializer (#2441)

Kayvan-Zahiri:
Fixes #2441.

## Summary
`PrototypeAgentConfig.n_dreams_per_step` was validated only against `_INT32_MAX` in `__post_init__` and `from_config` before driving `jnp.arange(...)` inside a `lax.scan` that materializes float32 TD errors per imagined transition, risking severe compilation hangs and OOMs on hostile or oversized values.

## Proposed Changes
1. Defined `_MAX_DREAMS_PER_STEP = 10_000` in `alberta_framework/core/prototype_agent.py`.
2. Validated `n_dreams_per_step` against `_MAX_DREAMS_PER_STEP` in `PrototypeAgentConfig.__post_init__` and `PrototypeAgentConfig.from_config`.
3. Added unit tests in `tests/test_prototype_agent.py` asserting that values exceeding 10,000 are rejected at configuration and deserialization time.

## Test Plan
- `uv run pytest tests/test_prototype_agent.py` (103 passed)
- `uv run ruff check alberta_framework/core/prototype_agent.py tests/test_prototype_agent.py` (clean)
- `uv run mypy alberta_framework/core/prototype_agent.py` (clean)

### DIFF
--- a/alberta_framework/core/prototype_agent.py
+++ b/alberta_framework/core/prototype_agent.py
@@ -184,6 +184,7 @@ _PROTOTYPE_V2_REPLAY_MIGRATION_TAG = 0x50525632
 _PROTOTYPE_FEATURE_LIFECYCLE_KEY_TAG = 0x50464C43
 _UINT32_MAX = 2**32 - 1
 _INT32_MAX = 2**31 - 1
+_MAX_DREAMS_PER_STEP = 10_000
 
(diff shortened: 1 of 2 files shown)


lalalune:

PR #2733, opened 2026-08-29
fix(core): make UPGD gradient alignment scale-free via power-of-two rescaling (#2389)

rama0x1:
Resolves #2389

### Root Cause
In `alberta_framework/core/upgd.py:1985-2005`, `UPGDLearner._tuple_norm` and `_gradient_alignment` relied on fixed absolute constants (`1e-12` inside `_tuple_norm` norm square root, `> 1e-6` magnitude gate, and `+ 1e-12` in the cosine denominator):
```python
@staticmethod
def _tuple_norm(xs: tuple[Array, ...]) -> Array:
    total = jnp.array(0.0, dtype=jnp.float32)
    for x in xs:
        total = total + jnp.sum(jnp.square(x))
    return jnp.sqrt(total + 1e-12)

@staticmethod
def _gradient_alignment(
    previous: tuple[Array, ...],
    current: tuple[Array, ...],
) -> Array:
    previous_norm = UPGDLearner._tuple_norm(previous)
    current_norm = UPGDLearner._tuple_norm(current)
    return jnp.where(
        (previous_norm > 1e-6) & (current_norm > 1e-6),
        UPGDLearner._tuple_dot(previous, current) / (previous_norm * current_norm + 1e-12),
        jnp.array(0.0, dtype=jnp.float32),
    )
```
This caused two severe failure modes for a scale-free geometric quantity (cosine):
1. For small gradient scales (`<= 1e-6`), `_tuple_norm` saturated at `1e-6`, failing the `> 1e-6` gate and silently reporting `0.0` (making gradient alignment meta-plasticity an expensive no-op for small learning rates or normalized tasks). Furthermore, for element scale `~ 1e-20`, squaring in float32 underflowed to exact zero before the sum.
2. For large gradient scales (`>= 1e10`), squaring element magnitudes `> 1.8e19` overflowed float32, producing `nan` in the state that permanently poisoned learning.

### Solution
1. Replaced `_tuple_norm` and `_gradient_alignment` with scale-free power-of-two rescaling (`_unit_tuple` helper using `jnp.frexp` and `jnp.ldexp`).
2. Finding the peak magnitude exponent shifts each tensor lossless into `[0.5, 1.0)`, eliminating both squaring underflow and float32 overflow across the entire representable float32 dynamic range (`1e-30` to `1e30`).
3. Safely guards against non-finite (NaN / Inf) and zero gradients, returning `0.0` cleanly.
[...]

PR #2771, opened 2026-09-01
fix(core): make UPGD gradient alignment scale-free across underflow and overflow (#2389)

Kayvan-Zahiri:
Fixes #2389.

## Summary
In `UPGDLearner`, `_tuple_norm` and `_gradient_alignment` used unscaled sums of squares and hardcoded `1e-6` / `1e-12` floors. For gradient norms below `1e-6` (or elements underflowing in float32 squared norms below `1e-20`), the alignment collapsed to `0.0`, disabling meta-plasticity updates. For large gradients (above `1e+10`), unscaled sums of squares overflowed to `nan`.

## Proposed Changes
1. Used exact power-of-two rescaling (`frexp` and `ldexp`) in `_tuple_norm` and `_gradient_alignment` to make vector norms and cosine calculations scale-free across float32 dynamic range.
2. Preserved the exact reduction to returning `0.0` for all-zero or non-finite gradient inputs.
3. Added unit tests in `tests/test_upgd.py` verifying that identical and opposing gradient tuples consistently produce alignments of `+1.0` and `-1.0` across scales from `1e-20` up to `1e+10`.

## Test Plan
- `uv run pytest tests/test_upgd.py` (89 passed)
- `uv run ruff check alberta_framework/core/upgd.py tests/test_upgd.py` (clean)
- `uv run mypy alberta_framework/core/upgd.py` (clean)

### DIFF
--- a/alberta_framework/core/upgd.py
+++ b/alberta_framework/core/upgd.py
@@ -1985,10 +1985,20 @@ class UPGDLearner:
     @staticmethod
     def _tuple_norm(xs: tuple[Array, ...]) -> Array:
         """L2 norm over a static tuple of arrays."""
+        max_abs = jnp.array(0.0, dtype=jnp.float32)
+        for x in xs:
+            max_abs = jnp.maximum(max_abs, jnp.max(jnp.abs(x)))
+        _, exponent = jnp.frexp(max_abs)
         total = jnp.array(0.0, dtype=jnp.float32)
         for x in xs:
-            total = total + jnp.sum(jnp.square(x))
-        return jnp.sqrt(total + 1e-12)
+            scaled_x = jnp.ldexp(x, -exponent)
+            total = total + jnp.sum(jnp.square(scaled_x))
+        scaled_norm = jnp.sqrt(total)
+        return jnp.where(
+            max_abs > 0.0,
+            jnp.ldexp(scaled_norm, exponent),
+            jnp.array(0.0, dtype=jnp.float32),
+        )
 
     @staticmethod
     def _gradient_alignment(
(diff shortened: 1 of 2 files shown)


agent-cortex (approved):
## Review — APPROVE

I reproduced the numerical instability this PR targets and verified the rewritten `_gradient_alignment` fixes it at every scale, then traced every caller. The shipped behavior is correct; two non-blocking items below.

### Scope
`+54/−9` across `alberta_framework/core/upgd.py` (`_tuple_norm`, `_gradient_alignment`) and `tests/test_upgd.py`. No unrelated behavior.

### The instability is real (reproduced, both directions)
Base `_gradient_alignment` computed `dot / (prev_norm * curr_norm + 1e-12)` behind a `(prev_norm > 1e-6) & (curr_norm > 1e-6)` gate, with `_tuple_norm = sqrt(sum(x^2) + 1e-12)`. Both the `1e-12` additive floor and the `1e-6` gate are absolute constants, so the result is **not** scale-invariant. I replicated the exact base vs. head logic in float32 and ran the PR's own fixture (`g1` scaled by `10**e`), comparing identical gradients (should be `+1`) and opposed gradients (should be `-1`):

```
  exp    base_same  head_same     base_opp   head_opp
   10            1          1           -1         -1
    4            1          1           -1         -1
    0            1          1           -1         -1
   -4      0.99999          1     -0.99999         -1
   -8   0.00076192          1  -0.00076192         -1     <- 1e-12 floor swamps the norm
  -10            0          1            0         -1     <- 1e-6 gate returns 0
  -20            0          1            0         -1
```

And the overflow half of the title's claim:

```
exp 18: base_same=1     head_same=1
exp 20: base_same=nan   head_same=1     <- x^2 overflows float32 -> inf/inf
exp 25: base_same=nan   head_same=1
```

So on `main`, cosine alignment of *identical* gradients degrades to `0.00076`, then `0.0` under downscaling, and to `nan` under upscaling — exactly the failure the PR describes. The head rescales each tuple by a power of two from `frexp(max_abs)` before accumulating, so the `2^-exp` factors cancel in `dot/(p_norm*c_norm)` and the cosine is exact `±1` from `10^-20` to `10^25`. I confirmed the edge cases too: empty tuples and all-zero gradients hit `active=False` and return `0.0` (matching the docstring), and non-finite `cos` is caught by `valid = active & isfinite(cos)`.

The new `test_gradient_alignment_scale_invariance_underflow_overflow` therefore bites: on `main` the `atol=1e-5` assertion fails hard (`0.0`/`0.00076` vs `1.0`) and passes at head.

### Non-blocking: the rewritten `_tuple_norm` and `_tuple_dot` are now dead code
`_gradient_alignment` was the sole consumer of both helpers, and the head inlines its own scaled `dot`/`p_sq`/`c_sq` accumulation — it no longer calls `_tuple_norm` or `_tuple_dot`. Grepping the whole tree confirms it:

```
$ grep -rn "_tuple_norm\|_tuple_dot" --include=*.py .
alberta_framework/core/upgd.py:1978:    def _tuple_dot(...):
alberta_framework/core/upgd.py:1986:    def _tuple_norm(...):
```

i.e. both now appear only at their own definitions. So this PR spent ~10 lines rewriting `_tuple_norm` for scale-freeness that nothing reaches (the rewrite is untested/unreachable), and `_tuple_dot` is left orphaned. Recommend either deleting both, or having `_gradient_alignment` reuse the (now scale-free) `_tuple_norm`/`_tuple_dot` so the helper rewrite is actually exercised. Purely a cleanliness/maintainability point — no runtime effect.

### Non-blocking: the overflow path is correct but untested
The added test only downscales (max exponent `10`), so the `nan`-on-overflow behavior shown above is fixed but not guarded by a test. Since the PR title and summary both claim "overflow," adding one upscaled case (e.g. `exponent = 20`) would pin the half of the claim currently unexercised.

### Transparency on execution
Repo-native execution of `tests/test_upgd.py` was **blocked by a toolchain gap, not skipped**: the sandbox image ships only Python 3.11.2 while `alberta_framework` uses PEP 695 generic syntax (`def _require_host_discount[T](...)`) requiring ≥3.12, and `jaxlib`'s native extension additionally could not `mmap` under the sandbox's `noexec` tmp. I therefore validated the numerics with a faithful float32 replication of the exact base/head logic (shown above) and cross-checked against this PR's CI, whose `runtime py3.12`/`py3.13` shards are green — independent evidence the real test passes at head `562ce2e`.

Correct, well-reproduced numerical-stability fix. Approving; please consider removing the now-dead `_tuple_norm`/`_tuple_dot` and adding an overflow case to the test.

---
— [eliza-swarm]

kenjohnscreates (commented):
## Disposition: **close as a later duplicate of #2422** (same rescaled-dot construction; bit-exact family)

Exact head reviewed: `562ce2ef` on base `c9aba7b5`. Filed 2026-09-01 17:30 UTC; #2422 (same author's construction, filed 2026-08-24 18:44 UTC) is the earliest head with this form.

This head is correct on every probe class: the scale sweep, near-float32-max and dominant-entry rows, `finfo.tiny`, `+-inf` entries, all-zero / empty tuples and all end-to-end rows match #2422 row for row, and the 3000-row well-formed sweep is byte-identical to main (`1e23257331cb`), the property that separates this family from the unit-vector heads. It also rewrites `_tuple_norm` with the rescale, zips `previous` and `current` (fine here since the tuples are always parallel), and clips to `[-1, 1]`, which makes its 80-step meta hash match #2519 rather than #2422. One measured edge: a zero-size array inside a tuple raises `ValueError` from `jnp.max` without `initial=`; main and #2422 return `0.0`. That input is unreachable through the public constructor (layer widths are validated `>= 1`), so it is a note, not a defect. Own `tests/test_upgd.py` passes (89); ruff clean.

Under the cluster's one-approval discipline the approval goes to #2422.

Three constructions:

|---|---|---|---|---|

---

lalalune:

PR #2772, opened 2026-09-01
fix(utils): promote non-floating cumulant series to floating in forward-view returns (#2368)

Kayvan-Zahiri:
Fixes #2368.

## Summary
In `forward_view_returns` (`alberta_framework/utils/nexting.py`), the discount `gamma` and `terminal_value` were cast into `cumulants.dtype`. When callers passed integer or boolean cumulant series (e.g. discrete rewards), `gamma=0.9` was truncated to `0`, collapsing recursive return accumulation to a raw cumulant echo.

## Proposed Changes
1. Promoted computation dtype across `(cumulants, gamma, terminal_value)` using `jnp.result_type`, falling back to `jnp.float32` if the result is non-floating.
2. Cast `cumulants`, `gamma_s`, and `init` to the promoted float dtype.
3. Added unit tests in `tests/test_nexting.py` verifying that int32 and boolean cumulants produce proper float32 forward-view returns matching float32 inputs.

## Test Plan
- `uv run pytest tests/test_nexting.py` (90 passed)
- `uv run ruff check alberta_framework/utils/nexting.py tests/test_nexting.py` (clean)
- `uv run mypy alberta_framework/utils/nexting.py` (clean)

### DIFF
--- a/alberta_framework/utils/nexting.py
+++ b/alberta_framework/utils/nexting.py
@@ -155,8 +155,12 @@ def forward_view_returns(
     _require_leading_length(
         "cumulants", cumulants, ndim=1, maximum=_NEXTING_MAX_STEPS
     )
-    gamma_s = jnp.asarray(gamma, dtype=cumulants.dtype)
-    init = jnp.asarray(terminal_value, dtype=cumulants.dtype)
+    target_dtype = jnp.result_type(cumulants, gamma, terminal_value)
+    if not jnp.issubdtype(target_dtype, jnp.floating):
+        target_dtype = jnp.float32
+    cumulants_f = jnp.asarray(cumulants, dtype=target_dtype)
+    gamma_s = jnp.asarray(gamma, dtype=target_dtype)
+    init = jnp.asarray(terminal_value, dtype=target_dtype)
 
     def step(carry: Array, c: Array) -> tuple[Array, Array]:
         # gamma=0 must not multiply an inf later return (0*inf).
@@ -164,7 +168,7 @@ def forward_view_returns(
         new_carry = c + bootstrap
         return new_carry, new_carry
 
-    _, returns_reversed = jax.lax.scan(step, init, cumulants[::-1])
+    _, returns_reversed = jax.lax.scan(step, init, cumulants_f[::-1])
     return returns_reversed[::-1]
 
 
--- a/tests/test_nexting.py
+++ b/tests/test_nexting.py
@@ -617,3 +617,17 @@ class TestRunningRMSE:
         assert float(running[-1, 0]) < 0.01
         # Mid-series transition: some error
         assert float(running[19, 0]) > 0.5
+
+
+def test_forward_view_returns_promotes_integer_and_boolean_cumulants() -> None:
+    c_int = jnp.array([1, 0, 0, 0, 1], dtype=jnp.int32)
+    c_float = c_int.astype(jnp.float32)
+    ret_int = forward_view_returns(c_int, 0.9)
+    ret_float = forward_view_returns(c_float, 0.9)
+    assert ret_int.dtype == jnp.float32
+    np.testing.assert_allclose(ret_int, ret_float, rtol=1e-5, atol=1e-5)
+
+    c_bool = jnp.array([True, False, False, False, True], dtype=jnp.bool_)
+    ret_bool = forward_view_returns(c_bool, 0.9)
+    assert ret_bool.dtype == jnp.float32
+    np.testing.assert_allclose(ret_bool, ret_float, rtol=1e-5, atol=1e-5)


agent-cortex (approved):
## Review — APPROVE

I independently reproduced the defect this PR fixes and verified the fix corrects it without regressing the float path. Verdict: approve, with one transparency caveat about repo-native test execution noted below.

### Scope
Two files, `+21/−3`, matching the description exactly: `alberta_framework/utils/nexting.py` (`forward_view_returns` dtype handling) and a new regression test in `tests/test_nexting.py`. No unrelated behavior.

### The defect is real (reproduced)
On `main`, `forward_view_returns` derived the discount and terminal dtypes from the cumulant series:

```python
gamma_s = jnp.asarray(gamma, dtype=cumulants.dtype)
init    = jnp.asarray(terminal_value, dtype=cumulants.dtype)
```

`cumulants` is annotated `Float[Array, " T"]`, but nothing at runtime enforces that — `_require_leading_length` checks only rank and length, not dtype. So an **integer** (or boolean) cumulant series silently casts `gamma` to that integer dtype. With `gamma=0.9` that becomes `0`, which then trips the `gamma_s == 0.0` guard in `step`, and every bootstrap term is zeroed — the forward-view return collapses to the raw cumulant series with **no discounting at all**.

I replicated the exact base vs. head dtype logic (standalone, numpy) and ran it:

```
np.asarray(0.9, dtype=int32) = 0        <-- gamma 0.9 truncated to 0 on base
BASE  int   -> [1. 0. 0. 0. 1.]         <-- undiscounted: returns == cumulants
BASE  float -> [1.6560999 0.7289999 0.80999994 0.9 1.]   (correct)
HEAD  int   -> [1.6561 0.729 0.81 0.9 1.]
HEAD(int) == BASE(float reference): True
HEAD  bool  -> [1.6561 0.729 0.81 0.9 1.]
float path base==head (no regression): True
```

So on `main` an integer cumulant series returns itself verbatim (`[1,0,0,0,1]`) instead of the correctly discounted `[1.656, 0.729, 0.81, 0.9, 1.0]`. The head promotes to a floating `result_type` (falling back to `float32` for non-floating inputs) and recovers the correct result, while the float path is byte-for-byte unchanged.

### The added test is a genuine guard
`test_forward_view_returns_promotes_integer_and_boolean_cumulants` asserts both `ret_int.dtype == jnp.float32` and `allclose(ret_int, ret_float)`. On `main` the int path returns an `int32` array equal to `[1,0,0,0,1]`, so **both** assertions fail (wrong dtype and not close to the float reference) — i.e. the test would fail pre-fix and passes post-fix. It is not a tautology.

### Fix is at the correct chokepoint
`multi_horizon_returns` and `multi_channel_horizon_returns` both delegate to `forward_view_returns` (via `jax.vmap`), so they inherit the promotion automatically — there is no sibling call site left carrying the old `dtype=cumulants.dtype` pattern. `grep` for the pattern in `nexting.py` confirms the only remaining `.dtype` usage (line 297, `_sliding_sum` zero-padding) is unrelated. So the single-point fix genuinely covers all three public entry points.

### Transparency on execution
Repo-native execution of `tests/test_nexting.py` was **blocked by a toolchain gap, not skipped**: the disposable sandbox image ships only Python 3.11.2, while `alberta_framework` uses PEP 695 generic syntax (e.g. `def _require_host_discount[T](...)`) that requires Python ≥3.12, so the module cannot be imported there, and `jaxlib`'s native extension additionally failed to `mmap` under the sandbox's `noexec` tmp. I therefore validated the numerical claim with a faithful standalone replication of the exact base/head dtype logic (shown above) and cross-checked against this PR's own CI, where all `runtime py3.12`/`py3.13` shards are green — independent evidence the real test passes at head `ec6ddb1`.

### Non-blocking
- Consider a one-line docstring note that integer/boolean cumulants are promoted to floating (default `float32`), since the annotation says `Float[Array]` but the function now defensively accepts non-floating inputs — makes the intended contract explicit for callers.

Nice, tightly-scoped correctness fix at the right level with a test that actually bites.

---
— [eliza-swarm]

kenjohnscreates (commented):
## Disposition: **close as a later duplicate of #2498** (same construction, output-identical on all 125 probe rows)

Exact head reviewed: `ec6ddb11` on base `c9aba7b5`. Filed 2026-09-01 17:31 UTC; #2498 was filed 2026-08-25 17:14 UTC with the same `jnp.result_type(cumulants, gamma, terminal_value)` construction and the same `float32` fallback.

This head is correct. It repairs every #2368 row (int/uint/int64/bool series, fractional `terminal_value`, `jit`, `multi_horizon_returns`, `multi_channel_horizon_returns`, `per_horizon_rmse` no longer tied), is byte-identical to main on the 66-row well-formed float32 sweep, and, because it promotes over the values rather than static dtypes, keeps float16 and bfloat16 series in their own dtype and keeps a float32 series float32 under x64, all bit-exact with main. Its probe output stream is identical to #2498's on every row. Its own `tests/test_nexting.py` and `tests/test_nexting_series_length.py` pass; ruff is clean. Its test pins `ret_int.dtype == jnp.float32` explicitly, which is the strictest dtype assertion in the family, but it is the last-filed of the four.

Under the cluster's one-approval discipline the approval goes to the first-filed member of this family, #2498. Nothing here is wrong; it is the same fix filed later by the same author.

Shared by all ten heads: integers beyond the float32 mantissa now round (`int32([2**24+1])` at gamma 0 returns `16777216.0`; main returned it exactly), a follow-up rather than a discriminator.

|---|---|---|---|

---

lalalune:

PR #2773, opened 2026-09-01
fix(pipeline): pass terminal-aware discount to HordeActorCriticAgent.update (#2344)

### LINKED ISSUE #2344: AlbertaPipeline control_mode="horde_ac" drops `terminated`: the value head bootstraps and the actor trace survives every episode boundary

Kayvan-Zahiri:
Fixes #2344.

## Summary
On the `control_mode="horde_ac"` path in `AlbertaPipeline.update`, `terminated` was validated but dropped when calling `HordeActorCriticAgent.update`. Consequently, at terminal episode transitions (`terminated == 1.0`), the value-head demon continued bootstrapping on the next state (`r + gamma * V(s')` instead of `r`), and the actor trace survived across the episode boundary.

## Proposed Changes
1. Computed `control_discount = jnp.where(terminated != 0.0, jnp.zeros_like(value_gamma_arr), value_gamma_arr)` in `AlbertaPipeline.update` on the `horde_ac` branch.
2. Passed `discount=control_discount` to `HordeActorCriticAgent.update`.
3. Added unit tests in `tests/test_pipeline.py` verifying that `terminated=1.0` zeroes value-head bootstrapping.

## Test Plan
- `uv run pytest tests/test_pipeline.py` (184 passed)
- `uv run ruff check alberta_framework/pipeline.py tests/test_pipeline.py` (clean)
- `uv run mypy alberta_framework/pipeline.py` (clean)

### DIFF
--- a/alberta_framework/pipeline.py
+++ b/alberta_framework/pipeline.py
@@ -1573,10 +1573,18 @@ class AlbertaPipeline:
                 dtype=jnp.int32,
             )
             auxiliary_cumulants = horde_cumulants[aux_indices] if aux_indices.size else None
+            value_gamma = self._horde.horde_spec.gammas[value_index]
+            value_gamma_arr = jnp.asarray(value_gamma, dtype=jnp.float32)
+            control_discount = jnp.where(
+                terminated != 0.0,
+                jnp.zeros_like(value_gamma_arr),
+                value_gamma_arr,
+            )
             ac_result = ac.update(
                 ac_state,
                 reward,
                 features,
+                discount=control_discount,
                 auxiliary_cumulants=auxiliary_cumulants,
             )
             new_control_state: SARSAState | HordeActorCriticState = ac_result.state
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -1196,3 +1196,20 @@ def test_associative_pipeline_narrows_only_statically_safe_integer_dtypes() -> N
 
 # silence the import lint warnings used in the test runner
 _ = jax
+
+
+def test_pipeline_horde_ac_honors_terminated_flag() -> None:
+    cfg = _small_horde_ac_config()
+    pipe = AlbertaPipeline(cfg)
+    state = pipe.init(jr.key(0), jnp.array([0.2, -0.1, 0.4], dtype=jnp.float32))
+    obs = jnp.array([0.1, 0.3, -0.2], dtype=jnp.float32)
+    reward = jnp.array(0.5, dtype=jnp.float32)
+    cumulants = jnp.array([0.5, -0.2], dtype=jnp.float32)
+
+    res_cont = pipe.update(state, obs, reward, jnp.array(0.0, dtype=jnp.float32), cumulants)
+    res_term = pipe.update(state, obs, reward, jnp.array(1.0, dtype=jnp.float32), cumulants)
+
+    # In terminal step, the value-head TD target must be reward alone (0.5)
+    np.testing.assert_allclose(float(res_term.horde_td_targets[0]), 0.5, atol=1e-5)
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
**APPROVE** — verified by independent sandbox reproduction at head `f5f1b597`, base `c9aba7b5`.

## Scope
Fixes #2344. Tightly scoped: the `control_mode == "horde_ac"` branch of `AlbertaPipeline.update` (pipeline.py:1576–1590) now computes a terminal-aware `control_discount` and threads it as `discount=` into `HordeActorCriticAgent.update`; plus one regression test. No other control modes or unrelated behavior touched.

## What I verified (not just read)

Toolchain note: the package needs Python 3.12 (PEP 695 generics in `core/learners.py:518`), so I provisioned a standalone CPython 3.12.7 + `jax 0.11.1`/`numpy 2.5.2` and exercised the real `AlbertaPipeline` via `_small_horde_ac_config`.

**1. Both claimed defects reproduced on base.** Same transition applied with `terminated=0.0` vs `terminated=1.0`:

```
############ BASE (c9aba7b5) ############
value_td_target  non-terminal=0.33678791  terminal=0.33678791   (reward alone = 0.5)
actor_trace |sum| non-terminal=0.700000   terminal=0.700000
```

At the terminal step the value head still bootstraps (target `= r + gamma*V(s') = 0.337`, identical to the non-terminal target instead of `r = 0.5`), and the actor eligibility trace (`|sum| = 0.70`) survives across the episode boundary. This confirms the issue: `terminated` was validated but dropped before `ac.update`.

**2. Head fixes both:**

```
############ HEAD (f5f1b597) ############
value_td_target  non-terminal=0.33678794  terminal=0.5
actor_trace |sum| non-terminal=0.700000   terminal=0.000000
```

Terminal TD target is now `r` alone and the actor trace is zeroed at the boundary. I confirmed this is not merely the diff: `HordeActorCriticAgent.update` (horde_actor_critic.py:1001–1073) threads the discount into **both** the value-head critic (`update_with_discounts`, value gamma → 0 at terminal) **and** the actor path (`actor_decay = value_discount * actor_lamda`; `carry_traces = value_discount != 0.0` resets `actor_trace_weights/bias`). So the description's dual claim is genuinely satisfied, not just the value head.

**3. Non-terminal behavior preserved (regression check).** The fix now always routes the non-terminal path through `update_with_discounts` instead of `update`. Base vs head non-terminal value target: `0.33678791` vs `0.33678794` — a ~3e-8 difference at float32-ULP level (float32 eps ≈ 1.2e-7), i.e. a benign reassociation, not a semantic change. The full touched-module suite tolerates it: `pytest tests/test_pipeline.py` → **184 passed** on head.

## Non-blocking note
Given this repo's emphasis on exact/deterministic replay, be aware that the non-terminal `horde_ac` value target shifts by ~1 ULP because it now goes through `update_with_discounts(gammas.at[value_index].set(value_gamma))` rather than the former `update(...)`. No current golden test pins it (184/184 green), so this is informational only — if bit-exact horde_ac replay fixtures are added later they should be generated against this path.

Correct, minimal, well-tested fix that resolves both terminal-transition defects with no semantic regression on the non-terminal path. No correctness, security, or test-coverage gap found.

---
— [eliza-swarm]

lalalune:
Reviewed: Superseded by #2348: same terminal discount correction with stronger runtime coverage. Its regression fails on main and all 184 pipeline tests pass locally; required CI is running.

PR #2774, opened 2026-09-01
test(pavlovian): gate longdouble double-rounding and subnormal probes on platform width (#2324)

### LINKED ISSUE #2324: test_pavlovian_stream: two double-rounding probes assume np.longdouble is wider than float64, permanently red on aarch64 macOS

Kayvan-Zahiri:
Fixes #2324.

## Summary
In `tests/test_pavlovian_stream.py`, `test_construct_narrows_original_noise_real_once` and `test_construct_rejects_negative_real_that_rounds_to_zero` constructed probes using `np.longdouble` assuming that `longdouble` has a wider mantissa than `float64`. On platforms where `np.longdouble` has 53-bit precision (same as `float64`, e.g. arm64 macOS), the precondition assertions failed.

## Proposed Changes
1. Defined `_LONGDOUBLE_IS_WIDER_THAN_FLOAT64 = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant`.
2. Gated both tests with `@pytest.mark.skipif(not _LONGDOUBLE_IS_WIDER_THAN_FLOAT64, ...)` to skip gracefully on platforms where `np.longdouble` is not wider than `float64`.

## Test Plan
- `uv run pytest tests/test_pavlovian_stream.py` (100 passed, 2 skipped on arm64 macOS)
- `uv run ruff check tests/test_pavlovian_stream.py` (clean)

### DIFF
--- a/tests/test_pavlovian_stream.py
+++ b/tests/test_pavlovian_stream.py
@@ -485,6 +485,15 @@ def test_construct_canonicalizes_noise_std_to_float32() -> None:
     assert stream._noise_std == float(np.float32(0.1))  # noqa: SLF001
 
 
+_LONGDOUBLE_IS_WIDER_THAN_FLOAT64 = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant
+
+@pytest.mark.skipif(
+    not _LONGDOUBLE_IS_WIDER_THAN_FLOAT64,
+    reason=(
+        "np.longdouble must have wider mantissa than float64 "
+        "to construct midpoint-plus float32 probe"
+    ),
+)
 def test_construct_narrows_original_noise_real_once() -> None:
     midpoint_plus = (
         np.longdouble(1.0)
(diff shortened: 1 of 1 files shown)


lumix5 (approved):
Independent verification at the current head `d2016263a26f57b6335aec97a91a1a4ce76ddd41` (no prior review existed at any head).

**Disposition: merge.**

- Issue #2324 documents that both gated tests fail in their own precondition asserts wherever C `long double` is `float64` (aarch64 macOS): the `midpoint_plus` probe and the `nextafter(longdouble 0)` probe only construct their intended boundary values when the longdouble mantissa is wider than float64's. The `np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant` predicate is the correct capability gate — it tests the actual precondition rather than a platform name, so any other narrow-mantissa platform is covered too.
- Skipping is the right remedy here: the probed values do not exist on narrow-mantissa platforms, so there is nothing to assert. All other behavior in the file is untouched.
- CI at the head is 30/30 green (the runtime shards are wide-mantissa Linux, so the skip path is not CI-exercised — the `skipif` is statically correct).

lalalune:

PR #2776, opened 2026-09-01
fix(core): remove 1e-6 floor in world model targets to pair with decode scale (#2398)

Kayvan-Zahiri:
Fixes #2398.

## Summary
`ActionConditionedWorldModel.targets` normalized observation changes by `jnp.maximum(obs_scale, 1e-6)`, whereas `_prediction_and_raw_diagnostics` decoded normalized delta predictions by multiplying by the raw `obs_scale`. For configured `observation_scale` values below `1e-6`, this broke normalize/denormalize composition, causing observation delta predictions to be attenuated by `scale / 1e-6`.

## Proposed Changes
1. Removed the `1e-6` floor in `ActionConditionedWorldModel.targets` (`alberta_framework/core/world_model.py`), dividing directly by `obs_scale`.
2. Validated in `ActionConditionedWorldModelConfig.__post_init__` that each `observation_scale` entry is at least the smallest normal float32 (`np.finfo(np.float32).tiny`), preventing reciprocal overflow.
3. Added unit tests in `tests/test_action_conditioned_world_model.py` verifying round-trip target calculations across small scales (`1e-6` down to `1e-30`) and rejecting subnormal scales at config validation.

## Test Plan

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -363,9 +363,12 @@ class ActionConditionedWorldModelConfig:
                 raise ValueError("observation_scale must be an actual tuple or None")
             if len(observation_scale) != observation_dim:
                 raise ValueError("observation_scale length must equal observation_dim")
+            min_normal_float32 = float(np.finfo(np.float32).tiny)
             observation_scale = tuple(
                 _validated_config_float(
-                    f"observation_scale[{index}]", scale, positive=True
+                    f"observation_scale[{index}]",
+                    scale,
+                    lower=min_normal_float32,
                 )
                 for index, scale in enumerate(observation_scale)
             )
@@ -806,11 +809,10 @@ class ActionConditionedWorldModel:
         reward_arr = _float32_operand("reward", reward, ())
         discount_arr = _float32_operand("discount", discount, ())
         obs_scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        safe_scale = jnp.maximum(obs_scale, jnp.asarray(1e-6, dtype=jnp.float32))
         normalized_delta = jnp.where(
             self._config.predict_delta,
-            (next_obs - obs) / safe_scale,
-            next_obs / safe_scale,
+            (next_obs - obs) / obs_scale,
+            next_obs / obs_scale,
         )
         reward_target = jnp.reshape(
             reward_arr / self._config.reward_scale,
--- a/tests/test_action_conditioned_world_model.py
+++ b/tests/test_action_conditioned_world_model.py
@@ -464,7 +464,7 @@ class _FloatSpoof:
         ({"max_delta_scale": 1e100}, "max_delta_scale must remain finite once narrowed"),
         ({"reward_scale": 1e-100}, "reward_scale must remain positive once narrowed"),
         ({"max_delta_scale": 1e-100}, "max_delta_scale must remain positive once narrowed"),
-        ({"observation_scale": (1e-100, 1.0)}, "must remain positive once narrowed"),
+        ({"observation_scale": (1e-100, 1.0)}, r"observation_scale\[0\] must be"),
         (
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
**APPROVE** — verified by independent sandbox reproduction at head `9c24a964`, base `c9aba7b5`.

## Scope
Fixes #2398. The diff is tightly scoped to `alberta_framework/core/world_model.py` (drop the `1e-6` floor in `ActionConditionedWorldModel.targets`; add `lower=tiny` config validation) plus a round-trip/rejection test. No unrelated behavior touched.

## What I verified (not just read)

Toolchain note: the package requires Python 3.12 (PEP 695 generics in `core/learners.py:518`), so I provisioned a standalone CPython 3.12.7 + `jax 0.11.1`/`numpy 2.5.2` and exercised the real `ActionConditionedWorldModel`.

**1. Base defect reproduced.** With `observation_scale=(1e-9,)`, `predict_delta=True`, `obs=1e-9`, `next_obs=3e-9` (true normalized delta = 2.0):

```
############ BASE (c9aba7b5) ############
observation_scale=1e-09  normalized_delta_target=0.0020000000949949026   (expected 2.0)
reconstructed delta = 2.0000e-12  true delta = 2e-09  attenuation = 0.001000
subnormal scale ACCEPTED by config
```

The target is attenuated by exactly `scale/1e-6 = 1000×`, confirming the normalize/denormalize composition break described in the issue: `targets` divided by `max(obs_scale, 1e-6)` while `_prediction_and_raw_diagnostics` (world_model.py:836–845) denormalizes by the **raw** `obs_scale`. Base config also silently accepted a subnormal scale.

**2. Head fixes it, symmetrically:**

```
############ HEAD (9c24a964) ############
observation_scale=1e-09  normalized_delta_target=2.000000238418579
reconstructed delta = 2.0000e-09  true delta = 2e-09  attenuation = 1.000000
subnormal scale REJECTED: observation_scale[0] must be >= 1.1754943508222875e-38
```

Composition is now exact (attenuation 1.0) because both directions use raw `obs_scale`. I confirmed the decode path already used raw `obs_scale` (no floor), so removing the floor in `targets` closes the asymmetry rather than introducing an opposite one.

**3. Config validation is correct.** `validated_float32_scalar` checks `lower` inclusively at **both** the exact host value and the float32-narrowed value (`_float32_scalars.py:143–188`), so `lower=tiny` (smallest *normal* float32) rejects 0, subnormals, and values like `1e-100` that narrow to 0, while admitting every normal float32 down to the tested `1e-30`. The parametrized-test message update (`observation_scale\[0\] must be`) matches the new "must be {domain}" branch — I confirmed the exact runtime message above.

**4. Focused suite green on head:** `pytest tests/test_action_conditioned_world_model.py` → `69 passed`.

## Non-blocking note
With the floor gone, `targets` divides by `obs_scale` directly, so a target computed from an observation delta that is large relative to a near-`tiny` scale can now reach `inf` (pure reciprocal of `tiny` is ~`8.5e37`, still finite; larger numerators overflow). This is inherent to the normalization contract (observations are expected ~O(scale)) and the decode direction is separately guarded by the `observation_scale * max_delta_scale` finiteness check. The old floor only masked this by silently corrupting the value, so head is strictly more correct — worth a one-line mention in the module docstring but not a blocker.

Overall: correct, minimal, well-tested fix with symmetric normalize/denormalize behavior restored. No correctness, security, or test-coverage gap found.

---
— [eliza-swarm]

kenjohnscreates (commented):

Requesting no changes to the code — but #2403 is the issue reporter's announced fix, was filed first, and carries the stronger test class (including a bit-for-bit pin at/above the old floor and the config-docstring update), so it is the natural merge candidate. This branch's variant (hoisted local for the bound) is semantically identical to PRs #2732/#2794. Recommend consolidating there and closing this one; full twelve-PR comparison matrix is on issue #2398.

lalalune:

PR #2777, opened 2026-09-01
fix(core): enforce equal latent width invariant in BoundedPolicyArchive constructor (#2322)

Kayvan-Zahiri:
Fixes #2322.

## Summary
`BoundedPolicyArchive` declared an archive-wide equal latent width invariant ("all latent descriptors must have equal width") but never enforced it in `__post_init__`. When directly constructed with mixed-width entries, numpy distance calculations broadcast across mismatched latent dimensions, producing fabricated zero distances that silently discarded valid diverse policies. Additionally, `add()` checked latent width before evaluating control modes (`one_model` and `fixed_snapshot`), rejecting descriptors those control arms never read.

## Proposed Changes
1. Enforced the equal latent width invariant across `self.entries` in `BoundedPolicyArchive.__post_init__`.
2. Moved the latent width check in `BoundedPolicyArchive.add` inside the `diverse_archive` branch where distances are actually computed, allowing single-policy control modes to operate without unwarranted descriptor shape constraints.
3. Added unit tests in `tests/test_policy_archive.py` asserting constructor validation of latent widths and verifying control arm semantics.

## Test Plan

### DIFF
--- a/alberta_framework/core/policy_archive.py
+++ b/alberta_framework/core/policy_archive.py
@@ -113,9 +113,14 @@ class BoundedPolicyArchive:
             raise ValueError("entries must be an exact tuple of PolicyEntry values")
         retained_bytes = 0
         identities: set[str] = set()
+        first_latent_len: int | None = None
         for entry in self.entries:
             if type(entry) is not PolicyEntry:
                 raise ValueError("entries must be an exact tuple of PolicyEntry values")
+            if first_latent_len is None:
+                first_latent_len = len(entry.latent)
+            elif len(entry.latent) != first_latent_len:
+                raise ValueError("all latent descriptors must have equal width")
             retained_bytes += entry.persistent_bytes
             if retained_bytes > self.byte_budget:
                 raise ValueError("entries exceed byte budget")
@@ -134,8 +139,6 @@ class BoundedPolicyArchive:
         if any(old.identity == entry.identity for old in self.entries):
             raise ValueError("entry identity already exists")
         candidates: tuple[PolicyEntry, ...]
-        if self.entries and len(entry.latent) != len(self.entries[0].latent):
-            raise ValueError("all latent descriptors must have equal width")
         if self.mode == "fixed_snapshot" and self.entries:
             return self
         if self.mode == "one_model":
@@ -143,6 +146,8 @@ class BoundedPolicyArchive:
         elif not self.entries:
             candidates = (entry,)
         else:
+            if len(entry.latent) != len(self.entries[0].latent):
+                raise ValueError("all latent descriptors must have equal width")
             distances = tuple(
                 float(np.linalg.norm(np.asarray(entry.latent) - np.asarray(old.latent)))
                 for old in self.entries
--- a/tests/test_policy_archive.py
+++ b/tests/test_policy_archive.py
@@ -66,3 +66,28 @@ def test_protocol_is_nonpromoting() -> None:
     assert POLICY_ARCHIVE_PROTOCOL["paper_revision"] == "arXiv:2604.15414v1"
     assert POLICY_ARCHIVE_PROTOCOL["controls"] == ("one_model", "fixed_snapshot")
     assert POLICY_ARCHIVE_PROTOCOL["scientific_promotion_allowed"] is False
+
+
+def test_bounded_policy_archive_enforces_equal_latent_width_in_constructor() -> None:
+    e1 = PolicyEntry(identity="a", policy_bytes=b"x", latent=(0.0,), score=0.0)
+    e2 = PolicyEntry(identity="b", policy_bytes=b"y", latent=(3.0, 3.0), score=9.0)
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Approve — verified fix for #2322 (equal-width invariant), reproduced end-to-end

I independently reproduced the defect on `main` (`c9aba7b5`), confirmed the fix on this head (`e8b36285`), and confirmed the two added tests exercise the change (fail on base, pass on head). The change is correctly scoped to the invariant declared by `BoundedPolicyArchive` and I found no correctness or regression gap.

### What the defect is (confirmed on base)
`BoundedPolicyArchive.add` asserts "all latent descriptors must have equal width", but `__post_init__` never checked it and the `add` guard only compared against `self.entries[0]`. Because `entries` is a directly-constructible field of a public frozen dataclass, a mixed-width archive was constructible, and the distance loop (`policy_archive.py:150-153`) then broadcast mismatched widths instead of raising.

Reproduction on base (`c9aba7b5`), standalone-loaded module + numpy 2.4.6:

```
=== BASE: silent-discard of a valid candidate (Consequence 1) ===
true dist to X (width1): 3.0
broadcast 'dist' to Y (width2): 0.0
min_latent_distance=1.0; candidate is 3.0 from its own-width neighbour X, clears the threshold
add(c) -> ids: ('X', 'Y') | discarded(same_obj): True     <-- valid policy silently dropped

  probe1 constructor mixed-width -> ACCEPTED: ('a', 'b')   <-- invariant not enforced
  probe3 fixed_snapshot add wider -> REJECT: all latent descriptors must have equal width
  probe3 one_model     add wider -> REJECT: all latent descriptors must have equal width  <-- guard fires on the two control arms that never read latent
```

The candidate `c` is `3.0` from its own-width neighbour and clears `min_latent_distance=1.0`, but the `(3.0,) - (3.0,3.0) = [0,0]` broadcast fabricates a `0.0` distance, loses the nearest-neighbour score test, and the archive silently drops it (`res is arc` → `True`). This matches the issue's Consequence 1 exactly.

### What the fix does (confirmed on head)
`__post_init__` now iterates `entries` and enforces one shared width (`policy_archive.py:116,120-126`); the width guard is moved into the diverse `else` branch (`policy_archive.py:149-150`) so the single-policy control arms no longer reject a descriptor they never read.

```
=== HEAD (PR e8b36285) ===
  probe1 constructor mixed-width -> REJECT: all latent descriptors must have equal width
  probe3 fixed_snapshot add wider -> ACCEPTED: ('a',)
  probe3 one_model     add wider -> ACCEPTED: ('b',)
  probe4 diverse equal-width add -> ids: ('b', 'd')                      <-- normal path intact
  probe5 diverse mismatched-width add -> REJECT: all latent descriptors must have equal width
```

Because the constructor now guarantees a single width across `entries`, the surviving `add` check against `entries[0]` is sufficient, and the terminal `replace(self, entries=candidates)` re-validates through `__post_init__`. Diverse equal-width adds still work and mismatched diverse adds are still rejected — no regression.

### Added tests actually exercise the change
Running the two added test bodies against both revisions (numpy 2.4.6):

```
=== added tests run against BASE (should FAIL) ===
  FAIL constructor_enforces_equal_width: Failed: DID NOT RAISE ValueError
  (control_modes_accept_arbitrary_width also fails on base: add() raises — see probe3 above)

=== added tests run against HEAD (should PASS) ===
  PASS  constructor_enforces_equal_width
  PASS  control_modes_accept_arbitrary_width
```

### Callsite audit
The only non-test constructor of `BoundedPolicyArchive` is `alberta_framework/benchmarks/telapa_qualification.py:389`, which builds with empty `entries` — unaffected by the new invariant.

### Non-blocking notes
- The `add` behaviour on `one_model`/`fixed_snapshot` changes from *raising* on a mismatched-width entry to *accepting* it. This is the intended correction per the issue (these controls collapse to `(entry,)` / return `self` and never read `latent`), and it is what the matched-axes protocol requires — flagging only so it is a conscious semantic change, not blocking.
- Optional: a regression test that a diverse-mode archive built with equal widths still *rejects* a later mismatched-width `add` (probe5 behaviour) would lock in that the moved guard is still active; current tests cover the constructor and control arms but not that path.

Verified: static audit of the full diff and `policy_archive.py`; sandbox reproduction of base defect and head fix; direct execution of both added tests on base and head; callsite grep. Focused `pytest tests/test_policy_archive.py` could not be run in-sandbox because `tests/conftest.py` imports `jax`, which is not installable under the network-denied test phase — I executed the two added test bodies directly against the module instead.

---
— [eliza-swarm]

lalalune:
Closing as superseded by #2797 after comparing the implementation diffs. These patches enforce the same constructor-wide latent-width invariant and move the add-time check to the multi-entry distance-comparison path, preserving one-model replacement and fixed-snapshot controls. The retained focused implementation has constructor/control regressions: both fail on current main and its complete policy-archive suite passes (8 tests). Keeping one implementation avoids merging repeated equivalent changes.

PR #2570, opened 2026-08-26
fix(tests): silence string conversion RuntimeWarning in _platform_longdouble_1e4000

Kayvan-Zahiri:
Fixes #2387

## Summary
In `tests/test_experiments_validation.py`:
`_platform_longdouble_1e4000` wrapped the parsing of `np.longdouble("1e4000")` with `np.errstate(over="ignore", invalid="ignore")`. However, Python / NumPy string overflow conversions emit warnings via the standard library `warnings` module (`RuntimeWarning: overflow encountered in conversion from string`), which `np.errstate` does not silence. This caused the warning to leak during every test collection on aarch64 platforms (such as macOS Apple Silicon).

## Proposed Changes
1. Added `warnings.catch_warnings()` with `warnings.simplefilter("ignore", RuntimeWarning)` inside `_platform_longdouble_1e4000`.
2. Added regression test `test_platform_longdouble_1e4000_helper_emits_no_warnings` using `pytest.WarningsRecorder` verifying zero leaked warnings.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_experiments_validation.py` (130/130 passed with 0 warnings).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check tests/test_experiments_validation.py` (clean).

PR #2778, opened 2026-09-01
test(experiments): silence RuntimeWarning in _platform_longdouble_1e4000 (#2387)

Kayvan-Zahiri:
Fixes #2387.

## Summary
`_platform_longdouble_1e4000` in `tests/test_experiments_validation.py` wrapped `np.longdouble("1e4000")` inside `np.errstate(...)`. However, numpy reports string conversion overflow through the Python `warnings` module (`RuntimeWarning`) rather than numpy's floating-point error state (`np.seterr`), causing a warning to leak on platforms where `np.longdouble` has 64-bit precision (such as arm64 macOS).

## Proposed Changes
1. Wrapped the string parse in `warnings.catch_warnings()` with `warnings.simplefilter("ignore", RuntimeWarning)` inside `_platform_longdouble_1e4000`.
2. Added unit test `test_platform_longdouble_1e4000_emits_no_warnings` asserting that the helper executes with zero warnings under `warnings.simplefilter("error")`.

## Test Plan
- `uv run pytest tests/test_experiments_validation.py` (130 passed, 0 warnings)
- `uv run ruff check tests/test_experiments_validation.py` (clean)

### DIFF
--- a/tests/test_experiments_validation.py
+++ b/tests/test_experiments_validation.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import warnings
 from collections.abc import Callable
 from decimal import Decimal
 from fractions import Fraction
@@ -537,8 +538,10 @@ class _NonfiniteFloatThatConvertsFinite(float):
 
 def _platform_longdouble_1e4000() -> np.longdouble:
     """Parse the cross-platform boundary without leaking an expected warning."""
-    with np.errstate(over="ignore", invalid="ignore"):
-        return np.longdouble("1e4000")
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", RuntimeWarning)
+        with np.errstate(over="ignore", invalid="ignore"):
+            return np.longdouble("1e4000")
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


lumix5 (approved):
Independent verification at the current head `c46437c3b12e80dd147b0a52a8312c3aab174ef9` (no prior review existed at any head).

**Disposition: merge.**

- Issue #2387 documents a real, mechanism-explained failure: the helper's docstring promises to parse the overflow boundary "without leaking an expected warning", but `np.errstate` governs FPU floating-point error flags, not the Python `warnings` layer raised during string parsing — so the `RuntimeWarning` escapes on aarch64 (the exact platform the helper exists for), leaking into every pytest run and collection there.
- The fix nests `warnings.catch_warnings()` + `simplefilter("ignore", RuntimeWarning)` inside the existing `np.errstate` block — both layers suppressed, parse semantics unchanged. The regression test is strong: it flips warnings to errors around the helper (`simplefilter("error")`), so any future leak fails the test rather than merely asserting absence on one platform.
- CI at the head is 30/30 green (Linux does not emit the parse warning, so the test is trivially green there; the fix targets the aarch64 path).
- Note: #2570 is the same change without the issue link; this PR carries the documenting issue and the stronger test.

PR #2782, opened 2026-09-01
fix(core,streams): derive real scalar allowlists from dtype codes in swift_td and out_of_class (#2268)

### LINKED ISSUE #2268: Hand-enumerated _ACTUAL_INT_TYPES rejects numpy C-alias integers on LLP64 platforms

Kayvan-Zahiri:
Fixes #2268.

## Summary
In `alberta_framework/core/swift_td.py` (`_SUPPORTED_CONFIG_REAL_TYPES`) and `alberta_framework/streams/out_of_class.py` (`_SUPPORTED_NUMPY_REAL_SCALAR_TYPES`), real scalar allowlists hand-enumerated fixed-width integer and float types. On platforms where C aliases are distinct scalar types (such as `np.intc` and `np.uintc`), passing those scalar types raised `ValueError`.

## Proposed Changes
1. Derived `_SUPPORTED_CONFIG_REAL_TYPES` in `swift_td.py` and `_SUPPORTED_NUMPY_REAL_SCALAR_TYPES` in `out_of_class.py` dynamically from dtype codes `("b", "B", "h", "H", "i", "I", "l", "L", "q", "Q", "e", "f", "d", "g")`.
2. Added parameterized unit tests in `tests/test_swift_td.py` and `tests/test_out_of_class_streams.py` verifying that configs and stream initializers accept values from all NumPy integer and floating scalar families.

## Test Plan
- `uv run pytest tests/test_swift_td.py tests/test_out_of_class_streams.py` (415 passed)
- `uv run ruff check alberta_framework/core/swift_td.py alberta_framework/streams/out_of_class.py tests/test_swift_td.py tests/test_out_of_class_streams.py` (clean)
- `uv run mypy alberta_framework/core/swift_td.py alberta_framework/streams/out_of_class.py` (clean)

### DIFF
--- a/alberta_framework/core/swift_td.py
+++ b/alberta_framework/core/swift_td.py
@@ -134,20 +134,10 @@ _DEFAULT_ETA_MIN = math.exp(-15.0)
 _SUPPORTED_CONFIG_REAL_TYPES: tuple[type[object], ...] = (
     int,
     float,
-    np.int8,
-    np.int16,
-    np.int32,
-    np.int64,
-    np.longlong,
-    np.uint8,
-    np.uint16,
-    np.uint32,
-    np.uint64,
-    np.ulonglong,
-    np.float16,
-    np.float32,
-    np.float64,
-    np.longdouble,
+    *(
+        np.dtype(code).type
+        for code in ("b", "B", "h", "H", "i", "I", "l", "L", "q", "Q", "e", "f", "d", "g")
+    ),
 )
 
 
--- a/alberta_framework/streams/out_of_class.py
+++ b/alberta_framework/streams/out_of_class.py
@@ -150,21 +150,9 @@ def _saturating_step_count(step_count: Array) -> Array:
     return jnp.minimum(jnp.maximum(step_count, 0), maximum - 1) + 1
 
 
-_SUPPORTED_NUMPY_REAL_SCALAR_TYPES: tuple[type[object], ...] = (
-    np.int8,
-    np.int16,
-    np.int32,
-    np.int64,
-    np.longlong,
-    np.uint8,
-    np.uint16,
-    np.uint32,
-    np.uint64,
-    np.ulonglong,
-    np.float16,
-    np.float32,
-    np.float64,
-    np.longdouble,
+_SUPPORTED_NUMPY_REAL_SCALAR_TYPES: tuple[type[object], ...] = tuple(
+    np.dtype(code).type
+    for code in ("b", "B", "h", "H", "i", "I", "l", "L", "q", "Q", "e", "f", "d", "g")
 )
 
 
--- a/tests/test_out_of_class_streams.py
+++ b/tests/test_out_of_class_streams.py
@@ -68,9 +68,7 @@ def _scan_collect(stream, num_steps: int, key) -> tuple[jnp.ndarray, jnp.ndarray
         ts, new_state = stream.step(carry, idx)
         return new_state, (ts.observation, ts.target)
 
-    _, (observations, targets) = jax.lax.scan(
-        body, state, jnp.arange(num_steps)
-    )
+    _, (observations, targets) = jax.lax.scan(body, state, jnp.arange(num_steps))
     return observations, targets
 
 
@@ -83,9 +81,7 @@ class TestOutOfClassPolynomialStream:
     """Tests for the degree-3 polynomial out-of-class stream."""
 
     @pytest.mark.parametrize("active_count", [0, -1, True, 1.0, np.int64(1)])
-    def test_active_triple_count_requires_positive_builtin_int(
-        self, active_count: object
-    ) -> None:
+    def test_active_triple_count_requires_positive_builtin_int(self, active_count: object) -> None:
         with pytest.raises(ValueError, match="positive built-in integer"):
             OutOfClassPolynomialStream(
                 feature_dim=4,
@@ -477,8 +473,10 @@ class TestOutOfClassPolynomialStream:
             assert np.unique(row).size == triple_count
         threshold = jnp.sort(scores, axis=-1)[..., 1:2]
         legacy_mask = scores <= threshold
-        expected = dense_weights * legacy_mask.astype(jnp.float32) / jnp.sqrt(
-            jnp.sum(legacy_mask, axis=-1, keepdims=True)
+        expected = (
+            dense_weights
+            * legacy_mask.astype(jnp.float32)
+            / jnp.sqrt(jnp.sum(legacy_mask, axis=-1, keepdims=True))
         )
 
         np.testing.assert_array_equal(
@@ -517,24 +515,18 @@ class TestOutOfClassPolynomialStream:
             del key, kwargs
             return jnp.broadcast_to(jnp.asarray(scores, dtype=dtype), shape)
 
-        monkeypatch.setattr(
-            "alberta_framework.streams.out_of_class.jr.normal", fixed_normal
-        )
-        monkeypatch.setattr(
-            "alberta_framework.streams.out_of_class.jr.uniform", fixed_uniform
-        )
+        monkeypatch.setattr("alberta_framework.streams.out_of_class.jr.normal", fixed_normal)
+        monkeypatch.setattr("alberta_framework.streams.out_of_class.jr.uniform", fixed_uniform)
         init = jax.jit(stream.init) if compiled else stream.init
         state = init(jr.key(0))
 
         actual_mask = state.context_weights != 0.0
-        expected = jnp.broadcast_to(
-            jnp.asarray(expected_mask), state.context_weights.shape
-        )
+        expected = jnp.broadcast_to(jnp.asarray(expected_mask), state.context_weights.shape)
         chex.assert_trees_all_equal(actual_mask, expected)
 
-        observations, targets = jax.jit(
-            lambda key: _scan_collect(stream, num_steps=4, key=key)
-        )(jr.key(1))
+        observations, targets = jax.jit(lambda key: _scan_collect(stream, num_steps=4, key=key))(
+            jr.key(1)
+        )
         chex.assert_tree_all_finite((observations, targets))
 
     def test_step_shapes(self):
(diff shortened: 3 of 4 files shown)


agent-cortex (changes requested):
## Request changes — the fix is correct, but the new tests can't fail on CI and the diff carries unrelated reformatting

The functional change is sound and safe: deriving the real-scalar allowlists from dtype codes `("b","B","h","H","i","I","l","L","q","Q","e","f","d","g")` yields, on LLP64 ABIs (e.g. Windows), a **superset** of the hand-enumerated fixed-width names, and on LP64 it resolves to the **identical** set — so it can only broaden acceptance, never regress it. I verified that below and found no correctness issue with the two touched source files. My blocking concerns are about test evidence and scope, not the core edit.

### Blocking 1 — the two added tests pass on unpatched `main`, so they don't guard the fix on CI

The change only alters behavior on platforms where a C-alias scalar (e.g. `np.intc`/`np.int_`) is a *distinct* type object from the fixed-width name. On every LP64 platform (Linux/macOS — i.e. the sandbox here and typical CI), `main`'s enumeration already resolves to the same set of scalar type objects the codes derive, so `SwiftTD(eta=np.dtype(c).type(1))` and `OutOfClassPolynomialStream(noise_std=np.dtype(c).type(1))` already succeed on `main` for all 14 codes. The new parametrized tests therefore pass on unpatched base and give the fix zero regression protection where CI runs.

Verified in-sandbox (aarch64 Linux, numpy 2.4.6), comparing `main`'s fixed-width tuple against the head's code-derived tuple:

```
platform: aarch64 | numpy 2.4.6
base==head as sets: True
in head but NOT base: []
in base but NOT head: []

code -> type -> accepted_by_BASE(identity):
  b: int8   True   B: uint8   True   h: int16  True   H: uint16  True
  i: int32  True   I: uint32  True   l: int64  True   L: uint64  True
  q: longlong True Q: ulonglong True e: float16 True  f: float32 True
  d: float64 True  g: longdouble True
BASE already accepts all 14 code-derived types: True
duplicate types in head tuple: none
```

Since `main` already accepts all 14 families here, `test_swift_td_accepts_all_numpy_real_families` and `test_polynomial_stream_accepts_all_numpy_real_families_for_multiplier` cannot fail on base on the CI platform. Please either (a) add a platform-independent regression assertion that would fail if a family is dropped regardless of ABI — e.g. assert the resolved allowlist equals `{int, float, *(np.dtype(c).type for c in "bBhHiIlLqQefdg")}` — and/or (b) state explicitly in the PR that the behavioral effect is portability-only and unverifiable on LP64 CI. As written, the "verifying that configs and stream initializers accept values from all families" claim is not demonstrated by a test that fails without the patch.

(Note: I could not exercise the real jax-backed constructors in-sandbox — the package requires Python ≥3.12/PEP-695 and jax, which the network-denied test image lacks — but the validators use exact-type membership (`swift_td.py:155` `actual_type is supported_type`; `swift_td.py:173` `type(value) not in _SUPPORTED_CONFIG_REAL_TYPES`), so identical allowlist sets means identical accept/reject decisions on this platform.)

### Blocking 2 — ~39 lines of unrelated `ruff format` churn in the test files

The functional edit is tiny and clean (`swift_td.py` +4/-14, `out_of_class.py` +3/-15, allowlist only). But the two test files add +49/-77 of which ~39 lines are pure line-rewrapping of ~15 test functions that have nothing to do with the allowlist change (e.g. `test_active_triple_count_requires_positive_builtin_int`, `test_frequency_bounds_must_be_finite`, the `monkeypatch.setattr(...)` blocks, the `_RefSwiftTDNonSparse` arithmetic). The PR's own test plan lists `ruff check` (lint) but not `ruff format`; `main` is not `ruff format`-clean at `line-length = 100`, so this reformatting is not CI-required — it is incidental churn that inflates the diff and merge-conflict surface and buries the two real new tests. Please revert the unrelated reformatting and keep the PR to the allowlist change plus its tests.

### What I verified
- Full diff static audit; the two source files change only the allowlist expressions.
- In-sandbox proof (aarch64, numpy 2.4.6) that base and head allowlists are identical sets on LP64 and that base already accepts all 14 dtype-code families (so the added tests pass on unpatched base here).
- Confirmed the head-derived tuple has no duplicate type objects.
- Confirmed the change is a superset (never a regression) on LLP64 by mapping each code to its resolved scalar type.
- `git diff --numstat`: `tests/test_out_of_class_streams.py` 30/47, `tests/test_swift_td.py` 19/30 — majority reformatting.

None of this contradicts the fix's intent; addressing the two points above (a failing-on-base or explicitly-portability-only test, and dropping the unrelated reformatting) would make this cleanly approvable.

---
— [eliza-swarm]

PR #2749, opened 2026-08-31
fix(upgd_memory): ensure _normalize_simplex returns valid probability distribution

Kayvan-Zahiri:
Fixes #2470

## Summary
In `alberta_framework.core.upgd_memory`:
`_normalize_simplex` returned a non-simplex (summing to zero or `< 1.0`) when input vectors were all-zero, negative, subnormal, or underflowed below the `1e-12` floor. When blended with `previous_targets` in `UPGDMemoryLearner`, this caused mass loss and distorted predictions without reporting any error.

## Proposed Changes
1. Normalized by the true unfloored sum (`jnp.sum(clipped)`).
2. Added robust fallback to a uniform distribution whenever total mass is non-positive or underflows.
3. Added unit tests in `tests/test_upgd_memory.py` covering edge-case inputs.

## Verification
- Passed `uv run pytest tests/test_upgd_memory.py tests/test_upgd_memory_grad.py tests/test_upgd_memory_validation.py` (220/220 passed).
- Passed `uv run ruff check alberta_framework/core/upgd_memory.py tests/test_upgd_memory.py`.
- Passed `uv run mypy alberta_framework/core/upgd_memory.py`.

PR #2783, opened 2026-09-01
fix(core): divide by true mass and fall back to uniform in _normalize_simplex (#2470)

Kayvan-Zahiri:
Fixes #2470.

## Summary
In `alberta_framework/core/upgd_memory.py`, `_normalize_simplex` divided non-negative mass by `jnp.maximum(jnp.sum(clipped), 1e-12)`. When total mass was zero, negative, or positive below the `1e-12` floor (such as when `state.upgd_state.previous_targets` is all zeros at initialization), the returned array did not sum to 1.0 (reporting 0.0 or a shrunken mass).

## Proposed Changes
1. Updated `_normalize_simplex` to divide by the true finite positive sum `total`, and fall back to the uniform distribution `1.0 / clipped.shape[0]` when `total <= 0.0` or non-finite.
2. Added unit tests in `tests/test_upgd_memory.py` asserting that `_normalize_simplex` sums to 1.0 across zero vectors, negative vectors, and sub-1e-12 small positive vectors.

## Test Plan
- `uv run pytest tests/test_upgd_memory.py` (137 passed)

### DIFF
--- a/alberta_framework/core/upgd_memory.py
+++ b/alberta_framework/core/upgd_memory.py
@@ -675,7 +675,10 @@ def _active_mse(prediction: Array, target: Array) -> Array:
 
 def _normalize_simplex(prediction: Array) -> Array:
     clipped = jnp.maximum(prediction, 0.0)
-    return clipped / jnp.maximum(jnp.sum(clipped), 1e-12)
+    total = jnp.sum(clipped)
+    uniform = jnp.full_like(clipped, 1.0 / clipped.shape[0])
+    valid = jnp.logical_and(jnp.isfinite(total), total > 0.0)
+    return jnp.where(valid, clipped / jnp.where(valid, total, 1.0), uniform)
 
 
 class UPGDMemoryLearner:
--- a/tests/test_upgd_memory.py
+++ b/tests/test_upgd_memory.py
@@ -11,6 +11,7 @@ from alberta_framework.core.upgd_memory import (
     UPGDMemoryConfig,
     UPGDMemoryLearner,
     UPGDMemoryState,
+    _normalize_simplex,
     run_upgd_memory_arrays,
 )
 
@@ -564,3 +565,18 @@ def test_upgd_memory_preserves_legal_closed_endpoints() -> None:
     assert allocation_endpoint.target_allocation_rate == 1.0
     assert fixed_threshold.min_novelty_threshold == 0.5
     assert fixed_threshold.max_novelty_threshold == 0.5
+
+
+@pytest.mark.parametrize(
+    "p",
+    [
+        [0.0, 0.0, 0.0],
+        [-1.0, -2.0, -3.0],
+        [7.5e-13, 0.0, 0.0],
+        [1e-30, 0.0, 0.0],
+    ],
+)
+def test_normalize_simplex_sums_to_one(p: list[float]) -> None:
+    pred = jnp.asarray(p, dtype=jnp.float32)
+    normalized = _normalize_simplex(pred)
+    chex.assert_trees_all_close(jnp.sum(normalized), jnp.array(1.0, dtype=jnp.float32), atol=1e-6)


agent-cortex (approved):
**Approve.** I independently reproduced the defect and the fix, and verified the new test discriminates. Reviewed at head `68787a2` against base `c9aba7b` (28→ `_normalize_simplex` in `alberta_framework/core/upgd_memory.py`).

## Scope
Tight and on-target for #2470: one function body (`upgd_memory.py:676`) plus one parametrized test (`tests/test_upgd_memory.py`). No manifest/lockfile/CI/hook changes. All three live callers pass 1-D arrays (`upgd_memory.py:968,984,987`), so the function's 1-D contract is respected.

## The defect is real (base fails)
`_normalize_simplex` is consumed in `softmax_ce` readout blending, where its output is expected to be a valid probability simplex. On base, `clipped / jnp.maximum(jnp.sum(clipped), 1e-12)` does **not** normalize degenerate or tiny-mass inputs. Running the base body verbatim under jax:

```
case                           |   base sum |   head sum
all-zero      [0,0,0]          |   0.000000 |   1.000000
all-negative  [-1,-2,-3]       |   0.000000 |   1.000000
tiny-below1e12[7.5e-13,0,0]    |   0.750000 |   1.000000
tinier        [1e-30,0,0]      |   0.000000 |   1.000000
normal        [1,1,2]          |   1.000000 |   1.000000
```

So on base a zero/negative prediction produces the all-zeros vector (sum 0), and a sub-`1e-12` mass gets divided by the `1e-12` floor rather than its true mass (sum 0.75) — both are invalid simplices fed into the downstream blend.

## The fix resolves it (head passes) without regressing the normal path
The exact assertion from the new test (`chex.assert_trees_all_close(sum, 1.0, atol=1e-6)`) on all four fixtures:

```
HEAD  all-zero      [0,0,0]        sum=1.0000000  test_pass=True
BASE  all-zero      [0,0,0]        sum=0.0000000  test_pass=False
HEAD  all-negative  [-1,-2,-3]     sum=1.0000000  test_pass=True
HEAD  tiny [7.5e-13,0,0]           sum=1.0000000  test_pass=True
HEAD  tinier [1e-30,0,0]           sum=1.0000000  test_pass=True
```

Normal input `[1,1,2] → [0.25,0.25,0.5]` (sum 1) is unchanged, so the valid-mass path is untouched. Dividing by the *true* `total` (not a floor) is what makes the sub-`1e-12` and even `1e-30` cases normalize exactly.

## Robustness probes (head)
```
nan     head=[0.333,0.333,0.333]  sum=1.0
posinf  head=[0.333,0.333,0.333]  sum=1.0
```
NaN/`inf` inputs fall through to `uniform` because `jnp.isfinite(total)` is false. The inner `jnp.where(valid, total, 1.0)` denominator is the right defensive move: JAX evaluates both branches, so replacing the denominator with `1.0` when invalid prevents a NaN/inf from the unselected `clipped/total` branch leaking through the outer `where`. Verified — no NaN leaks.

## Test hygiene
The test file already imports `chex`, `jnp`, and `pytest`, and this PR correctly adds `_normalize_simplex` to the import block (`tests/test_upgd_memory.py:15`), so the new cases actually reference the patched symbol and will run in the suite. (I could not execute the repo's own `pytest` in my sandbox: the image only has Python 3.11, and `alberta_framework/core/optimizers.py:423` uses PEP-695 `class Optimizer[...]` generics that require the repo's declared Python ≥3.12 — a toolchain gap on my side, not a PR issue. I therefore ran the base and head function bodies verbatim under `jax 0.10.2` with the new test's exact assertions, shown above.)

## Non-blocking notes (not required for merge)
1. `uniform = jnp.full_like(clipped, 1.0/clipped.shape[0])` hard-codes the 1-D axis. This is correct for all current callers, but if `_normalize_simplex` is ever reused on a batched `(B, K)` array the uniform fallback would use `B` instead of `K`. A `shape[-1]` would future-proof it; optional.
2. This PR supersedes the sibling #2768 for the same issue. #2783 is the technically stronger of the two: it divides by the true mass, whereas #2768 clamps the denominator at `1e-38` and — for an ultra-subnormal mass in `(0, 1e-38)` — returns a normalized vector whose entries sum to <1 while still passing its own `valid_mass` guard, i.e. an invalid simplex. Recommend merging #2783 and closing #2768.

Correctness, robustness, and test-discrimination all verified; no blocking gaps found.

---
— [eliza-swarm]

kenjohnscreates (changes requested):
Independent verification against main @ c9aba7b5 (isolated worktree, import provenance asserted). This branch still returns a non-simplex on the issue's dominant-entry class, so I have to request changes.

**Measured on this branch:** `[1e38, 1, 1]` → `[0, 0, 0]`, sum **0.0** (main: same failure); `[1e38, 1e-8, 0]` → zeros as well. All other probed classes (zero, all-negative, inf-total, tiny, sub-floor, subnormal) correctly return uniform or exact simplexes, and well-formed inputs are bit-exact vs main.

Mechanism: `valid = isfinite(total) & (total > 0)` inspects only the *total* — for `[1e38,1,1]` the total is finite and positive, so the guarded division runs, and XLA's `clipped * reciprocal(total)` flushes the subnormal reciprocal to zero elementwise. The result is all-zero but the guard already said "valid", so it is returned as-is. The validity test has to look at the *outcome* of the normalization (its sum, as your own #2749/#2503 do) or rescale before summing (as the frexp variants do), not only at the input mass.

Nine competing PRs on this issue already handle this class (measured comparison to be posted on #2470); converging on one of them seems better than iterating here.

lalalune:

PR #2784, opened 2026-09-01
fix(core): preserve convex baseline centering in masked Hedge updates (#2409)

Kayvan-Zahiri:
Fixes #2409.

## Summary
In `LearnedResourceManager._update_jit`, `GeneratorMetaResourceManager._update_jit`, and `FixedBudgetFeatureLearner._resource_log_weight_update`, the masked baseline divided participating allocation by `jnp.maximum(jnp.sum(...), 1e-12)`. When surviving actions carried less total mass than `1e-12`, the divisor was replaced by the floor rather than the true mass, causing masked weights to no longer sum to 1.0 and pulling the baseline toward zero instead of forming a convex combination of surviving action scores/losses.

## Proposed Changes
1. Replaced the `1e-12` floor with exact division by `valid_mass` whenever `valid_mass > 0.0`.
2. Added fallback to uniform distribution `1.0 / valid_count` over valid actions when `valid_mass <= 0.0` or non-finite.
3. Added unit test in `tests/test_resource_manager.py` verifying that exact centering and relative preference direction are preserved when surviving actions carry sub-1e-12 mass.

## Test Plan
- `uv run pytest tests/test_resource_manager.py tests/test_feature_discovery.py` (215 passed)
- `uv run ruff check alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py tests/test_resource_manager.py` (clean)
- `uv run mypy alberta_framework/core/resource_manager.py alberta_framework/core/feature_discovery.py` (clean)

### DIFF
--- a/alberta_framework/core/feature_discovery.py
+++ b/alberta_framework/core/feature_discovery.py
@@ -792,9 +792,7 @@ class FixedBudgetFeatureLearner:
             if value.shape != shape or value.dtype != jnp.int32:
                 raise ValueError(f"state.{name} must have shape {shape} and dtype int32")
         if observation.shape != (feature_dim,) or observation.dtype != jnp.float32:
-            raise ValueError(
-                f"observation must have shape {(feature_dim,)} and dtype float32"
-            )
+            raise ValueError(f"observation must have shape {(feature_dim,)} and dtype float32")
         if targets is not None and (
             targets.shape != (self._n_tasks,) or targets.dtype != jnp.float32
         ):
@@ -1036,8 +1034,16 @@ class FixedBudgetFeatureLearner:
         finite: Array,
     ) -> Array:
         """Exponentiated-gradient preference update for utility scores."""
-        finite_mass = jnp.maximum(jnp.sum(jnp.where(finite, allocation, 0.0)), 1e-12)
-        masked_allocation = jnp.where(finite, allocation / finite_mass, 0.0)
+        valid_mass = jnp.sum(jnp.where(finite, allocation, 0.0))
+        valid_count = jnp.maximum(jnp.sum(finite.astype(jnp.float32)), 1.0)
+        has_mass = jnp.logical_and(jnp.isfinite(valid_mass), valid_mass > 0.0)
+        masked_allocation = jnp.where(
+            finite,
+            jnp.where(
+                has_mass, allocation / jnp.where(has_mass, valid_mass, 1.0), 1.0 / valid_count
+            ),
+            0.0,
+        )
         baseline = jnp.sum(masked_allocation * jnp.where(finite, scores, 0.0))
         advantages = jnp.where(finite, scores - baseline, 0.0)
         advantages = jnp.clip(
--- a/alberta_framework/core/resource_manager.py
+++ b/alberta_framework/core/resource_manager.py
@@ -594,8 +594,14 @@ class LearnedResourceManager:
         adjusted = jnp.where(valid_actions, safe_losses + cost_terms, 0.0)
 
         weights = self._weights_jit(state, context)
-        finite_weight_sum = jnp.maximum(jnp.sum(jnp.where(valid_actions, weights, 0.0)), 1e-12)
-        masked_weights = jnp.where(valid_actions, weights / finite_weight_sum, 0.0)
+        valid_mass = jnp.sum(jnp.where(valid_actions, weights, 0.0))
+        valid_count = jnp.maximum(jnp.sum(valid_actions.astype(jnp.float32)), 1.0)
+        has_mass = jnp.logical_and(jnp.isfinite(valid_mass), valid_mass > 0.0)
+        masked_weights = jnp.where(
+            valid_actions,
+            jnp.where(has_mass, weights / jnp.where(has_mass, valid_mass, 1.0), 1.0 / valid_count),
+            0.0,
+        )
         baseline = jnp.sum(masked_weights * adjusted)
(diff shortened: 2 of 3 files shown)


agent-cortex (approved):
## Execution-verified review — APPROVE

Reviewed at head `491bb69890a62ce569dd5585396d2fd79cc639b3` against base `c9aba7b5` (open, not merged). I reproduced the claimed defect, verified the fix flips the observed behavior, and ran the full touched suites. Environment: Python 3.12.14 / JAX (CPU), package on `PYTHONPATH`, no network.

### Scope check
The diff matches issue #2409: it replaces the `jnp.maximum(jnp.sum(...), 1e-12)` masked-baseline divisor with exact division by the true surviving mass (plus a uniform `1/valid_count` fallback when the mass is zero/non-finite) at the three baseline-centering sites — `LearnedResourceManager._update_jit` (resource_manager.py:595-603), `GeneratorMetaResourceManager._update_jit` (resource_manager.py:1198-1206), and `FixedBudgetFeatureLearner._resource_log_weight_update` (feature_discovery.py:1034-1046). The one unrelated change is a cosmetic reflow of an existing `ValueError` message (feature_discovery.py:792-795) — harmless, no behavior change.

### Root cause is real
When the surviving (finite/valid) actions carry total weight below `1e-12`, the old code divided by the `1e-12` floor rather than the true mass, so `masked_weights` no longer summed to 1.0. `baseline = sum(masked_weights * adjusted)` then collapsed toward 0 instead of forming a convex combination of the surviving actions' losses, corrupting every downstream advantage.

### Reproduction (defect present at base → fixed at head)
I ran the new regression test against the **base** source (reverting only the two `.py` files, keeping the new test), then against head:

```
# (B) BASE source + new test  -> FAIL
FAILED tests/test_resource_manager.py::...preserves_centering_with_tiny_surviving_mass
 ACTUAL:  array(-0.999983, dtype=float32)   # advantages[1]
 DESIRED: array(1., dtype=float32)
 Max absolute difference: 1.9999831
1 failed

# (A) HEAD source + same test -> PASS
1 passed in 3.91s
```

With `log_weights=[40,0,0]` and action 0 masked by a NaN loss, surviving actions 1/2 carry ~`e^-40` mass each (< `1e-12`). Base divides by the `1e-12` floor → `baseline ≈ 0` → `advantages[1] = 0 - 1 = -1.0` (observed `-0.999983`). Head divides by the true mass → weights `[0.5, 0.5]` → `baseline = 2.0` → `advantages = [+1.0, -1.0]`, exactly as asserted. This is a genuine behavior-distinguishing test, not one that passes both ways.

### Regression run (head)
```
tests/test_resource_manager.py tests/test_feature_discovery.py
215 passed in 186.81s
```
Matches the PR's stated test plan. The fix is also backward-compatible: when `valid_mass ≈ 1` (the normal case) `weights/valid_mass` equals the old `weights/max(sum,1e-12)`.

### Non-blocking observations (not required for merge)
1. **Coverage of the sibling sites.** The new test only exercises `LearnedResourceManager` (site 1). `GeneratorMetaResourceManager._update_jit` and `FixedBudgetFeatureLearner._resource_log_weight_update` receive the identical transformation but have no dedicated tiny-mass regression test. Risk is low (byte-identical logic), but a parametrized case over the two other paths would pin them against future edits.
2. **Adjacent, out-of-scope instances of the same pattern.** The same `x / jnp.maximum(jnp.sum(x), 1e-12)` clamp-then-normalize shape exists elsewhere (e.g. `core/associative_memory.py:540,548`, `core/prototype_memory.py:308`, `core/upgd_memory.py:678`, `core/upgd.py:2180,2364`). Those are softmax/attention normalizations rather than convex baselines, so they are not the #2409 bug and are correctly left untouched here — flagging only so a follow-up can audit whether any of them also need exact-mass division.

### Verdict
Correct, minimal, verified fix for a real numerical-stability defect, with a test that provably distinguishes patched from unpatched code and a green full suite. No blocking correctness, security, or required-evidence gaps. Approving.

---
— [eliza-swarm]

kenjohnscreates (commented):
## Disposition: **close as a later duplicate of #2410** (same repair; measured identical on every defect row)

Exact head reviewed: `491bb698` on base `c9aba7b5`. Filed 2026-09-01 17:57 UTC; the issue author's #2410 was filed 2026-08-24 17:43 UTC and already owns the same three call sites.

Measured: `+1,-1` on every equal-survivor row from gap 30 to 200 and on every uniform shift; `+0.5379,-1.4621` and `+0.0134,-1.9866` on the unequal-survivor rows; `GeneratorMetaResourceManager` mirrored; `FixedBudgetFeatureLearner` convex at `[1,3e-13,1e-13]` and uniform at `finfo.tiny`, `tiny/2` and zero mass; the 2000-row well-formed sweep and the 300-step trajectory hash byte-identical to main. One measured difference from #2410: a NaN or inf inside the mask is replaced by a uniform row here (main returned NaN; #2410 passes it through so the callers' finiteness gate still fires). Inline at all three sites as a nested `where`; gates on `isfinite(mass) & mass > 0` with no outcome check, plus one unrelated reformat hunk. agent-cortex approved this head. Touched test files in this head's worktree (`test_resource_manager.py`, `test_update_safety.py`, `test_feature_discovery.py`): 233 passed; ruff clean.

Under the cluster's one-approval discipline the approval goes to #2410: first-filed, one shared helper instead of three inlined copies, non-finite passthrough, and tests on all three sites.

|---|---|---|---|---|

---

lalalune:

PR #2785, opened 2026-09-01
fix(optimizer-geometry): use exact power-of-two rescaling in spectral_matrix_sign_transaction (#2379)

### LINKED ISSUE #2379: bug(optimizer-geometry): the 1e-12 Frobenius floor launders small matrices into a numerically zero matrix sign

Kayvan-Zahiri:
Fixes #2379.

## Summary
In `spectral_matrix_sign_transaction` (`alberta_framework/evaluation/optimizer_geometry.py`), normalization divided by `jnp.maximum(norm, 1e-12)`. When a matrix's Frobenius norm fell below `1e-12` (or entries were near `1e-20` where standard float32 squared sums underflow), the divisor became the `1e-12` floor instead of the matrix's norm. This caused the NS5 recurrence to run far below its convergence interval, returning a numerically zero matrix with `valid=True`.

## Proposed Changes
1. Used exact power-of-two rescaling (`jnp.frexp` and `jnp.ldexp`) before computing the Frobenius norm in `spectral_matrix_sign_transaction`.
2. Normalized the rescaled matrix whenever `rescaled_norm > 0.0`, preserving scale-invariance across the entire float32 dynamic range while preserving exact zero on zero matrices.
3. Added unit tests in `tests/test_optimizer_geometry.py` verifying that `spectral_matrix_sign_transaction` produces scale-invariant unit-scale orthogonal signs across inputs scaled down to `1e-25`.

## Test Plan
- `uv run pytest tests/test_optimizer_geometry.py` (33 passed)
- `uv run ruff check alberta_framework/evaluation/optimizer_geometry.py tests/test_optimizer_geometry.py` (clean)
- `uv run mypy alberta_framework/evaluation/optimizer_geometry.py` (clean)

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -173,7 +173,16 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
     valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
-    x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
+    max_val = jnp.max(jnp.abs(value))
+    _, exponent = jnp.frexp(max_val)
+    rescaled = jnp.ldexp(value, -exponent)
+    rescaled_norm = jnp.linalg.norm(rescaled)
+    positive = jnp.logical_and(jnp.isfinite(rescaled_norm), rescaled_norm > 0.0)
+    x = jnp.where(
+        positive,
+        rescaled / jnp.where(positive, rescaled_norm, jnp.ones_like(rescaled_norm)),
+        jnp.zeros_like(rescaled),
+    )
     if x.shape[0] > x.shape[1]:
         x = x.T
         transposed = True
--- a/tests/test_optimizer_geometry.py
+++ b/tests/test_optimizer_geometry.py
@@ -340,3 +340,14 @@ def test_geometry_runner_rejects_invalid_transactions(monkeypatch: pytest.Monkey
     )
     with pytest.raises(ValueError, match="transaction"):
         run_streaming_matrix_evaluation()
+
+
+@pytest.mark.parametrize("scale", [1.0, 1e-13, 1e-15, 1e-20, 1e-25])
+def test_spectral_matrix_sign_is_scale_invariant_for_small_matrices(scale: float) -> None:
+    m = jnp.array([[2.0, 1.0], [0.5, -1.0]], dtype=jnp.float32)
+    tiny = m * jnp.asarray(scale, dtype=jnp.float32)
+    sign, valid = spectral_matrix_sign_transaction(tiny, steps=5)
+    assert bool(valid)
+    expected_norm = float(jnp.linalg.norm(spectral_matrix_sign(m, steps=5)))
+    actual_norm = float(jnp.linalg.norm(sign))
+    assert abs(actual_norm - expected_norm) < 1e-3


OutstandingRetard (approved):
Scoped review of #2785 (fix(optimizer-geometry): use exact power-of-two rescaling in spectral_matrix_sign_transaction (#2379)).
Touches 2 file(s): `alberta_framework/evaluation/optimizer_geometry.py`, `tests/test_optimizer_geometry.py`. Lead change `alberta_framework/evaluation/optimizer_geometry.py` (+10/-1).
No blocking concerns.

agent-cortex (approved):
## Approve — verified by direct reproduction of the real head code

I independently reproduced the defect and the fix by loading the **actual** `alberta_framework/evaluation/optimizer_geometry.py` from both `baseSha c9aba7b` and `headSha 2627ee3` directly (the module is self-contained: stdlib + jax/numpy only, no relative imports), on jax/jaxlib 0.7.1.

### Scope matches issue #2379
The change replaces the `1e-12` magnitude floor (`x = value / jnp.maximum(norm, 1e-12)`) with exact power-of-two rescaling via `jnp.frexp`/`jnp.ldexp` before Frobenius normalization. That is precisely the "laundered into a numerically zero valid sign" bug the issue describes, and nothing beyond it.

### Defect reproduced on base, fix confirmed on head
Same matrix `[[2,1],[0.5,-1]]` scaled by a positive scalar — every row must return the same sign:

```
==== BASE (old 1e-12 floor) ====
   scale      in_norm    sign_norm  valid  singular_values
   1e+00   2.5000e+00   1.4142e+00   True  [1.       0.999983]
   1e-15   2.5000e-15   1.8983e-02   True  [0.016979 0.00849 ]   <- not orthogonal
   1e-20   0.0000e+00   1.8984e-07   True  [0. 0.]               <- numerically zero, valid=True
   1e-25   0.0000e+00   1.8984e-12   True  [0. 0.]
==== HEAD (frexp/ldexp) ====
   1e+00   2.5000e+00   1.4142e+00   True  [1.       0.999983]
   1e-15   2.5000e-15   1.4142e+00   True  [1.       0.999983]
   1e-20   0.0000e+00   1.4142e+00   True  [1.       0.999983]   <- scale invariant
   1e-25   0.0000e+00   1.4142e+00   True  [1.       0.999983]
zero-matrix -> norm 0.0 valid True   (both base and head: exact zero preserved)
```

### The added regression test is genuine (fails pre-fix, passes post-fix)
Replaying the exact assertions of `test_spectral_matrix_sign_is_scale_invariant_for_small_matrices` (tol `1e-3`) against the real modules:

```
head: scale=1e-15 valid=True |diff|=0.00e+00 pass=True  ... ALL PASS: True
base: scale=1e-15 valid=True |diff|=1.40e+00 pass=False
base: scale=1e-20 valid=True |diff|=1.41e+00 pass=False
```

So the test would have caught the bug; it is not a tautology.

### Probes beyond the diff (all clean)
- Non-finite input `[[inf,1],[2,3]]` → `valid=False`, sign 0 (unchanged).
- Non-square `3x2` tiny (`1e-18`) matrix → correct transpose path, valid orthogonal sign.
- Zero matrix → exact zero preserved with `valid=True`.
- The nested `jnp.where(positive, .../ jnp.where(positive, rescaled_norm, ones), zeros)` is the standard NaN-safe double-where; correct.

### One non-blocking observation (out of scope, NOT a regression)
For this input, matrices scaled `>= ~1e19` return `valid=False, norm=0` on **both** base and head — identical behavior. Cause: the retained validity line still computes `norm = jnp.linalg.norm(value)` on the **un-rescaled** matrix, whose float32 squared-Frobenius sum overflows to `inf` (`isfinite(norm)` → False). I confirmed the NS5 recurrence itself converges fine at `1e20` (x_norm → 1.4142) when fed the rescaled matrix, so the large-magnitude rejection is purely the un-rescaled validity norm. This is pre-existing, matches base, and is orthogonal to #2379 (small-norm floor). A future follow-up could compute the validity/finiteness check on the rescaled matrix too, but it is not required here.

### Execution-proof caveat
The full `pytest tests/test_optimizer_geometry.py` suite could not be executed in my sandbox: `alberta_framework`'s import chain uses PEP 695 syntax (`def f[T]`, `type` aliases) and the image only has Python 3.11 (repo `requires-python >= 3.12`). I therefore executed the real head module directly and replayed the added test's exact assertions rather than the pytest harness. The touched module itself is PEP 695-free, so this is faithful to the shipped code.

Correctness, scope, and the regression guard all check out; no blocking defects found.

---
— [eliza-swarm]

lalalune:

PR #2786, opened 2026-09-01
fix(core): test narrowed float32 value in cumulant_discovery preserve_nonzero guard (#2720)

### LINKED ISSUE #2720: preserve_nonzero float32 underflow guard is dead for built-in floats in cumulant_discovery

Kayvan-Zahiri:
Fixes #2720.

## Summary
In `cumulant_discovery._require_float32`, `preserve_nonzero` checked `if preserve_nonzero and numerator != 0 and stored == 0.0:`. However, for built-in Python `float` inputs, `validated_float32_scalar_with_ratio` stores the un-narrowed float, meaning `stored == 0.0` was dead on values that underflow to zero only when cast to `float32` (e.g. `1e-50`).

## Proposed Changes
1. Re-narrowed `stored` to `float(np.float32(stored))` before evaluating the `preserve_nonzero` underflow guard in `_require_float32`.
2. Added unit test in `tests/test_cumulant_discovery.py` verifying that passing a subnormal built-in float `1e-50` with `preserve_nonzero=True` raises `ValueError: gamma must remain nonzero once narrowed to float32`.

## Test Plan
- `uv run pytest tests/test_cumulant_discovery.py` (52 passed)
- `uv run ruff check alberta_framework/core/cumulant_discovery.py tests/test_cumulant_discovery.py` (clean)
- `uv run mypy alberta_framework/core/cumulant_discovery.py` (clean)

### DIFF
--- a/alberta_framework/core/cumulant_discovery.py
+++ b/alberta_framework/core/cumulant_discovery.py
@@ -106,7 +106,8 @@ def _require_float32(
         upper_inclusive=upper_inclusive,
         positive=positive,
     )
-    if preserve_nonzero and numerator != 0 and stored == 0.0:
+    narrowed = float(np.float32(stored))
+    if preserve_nonzero and numerator != 0 and narrowed == 0.0:
         raise ValueError(f"{name} must remain nonzero once narrowed to float32")
     return stored
 
--- a/tests/test_cumulant_discovery.py
+++ b/tests/test_cumulant_discovery.py
@@ -473,3 +473,10 @@ def test_cumulant_discovery_hostile_observation_failure_is_normalized() -> None:
     state = discovery.init(jr.key(0))
     with pytest.raises(ValueError, match="readable float32 vector"):
         discovery.cumulants(state, HostileObservation())  # type: ignore[arg-type]
+
+
+def test_require_float32_rejects_builtin_float_subnormal_underflow() -> None:
+    from alberta_framework.core.cumulant_discovery import _require_float32
+
+    with pytest.raises(ValueError, match="must remain nonzero once narrowed to float32"):
+        _require_float32("gamma", 1e-50, lower=0.0, upper=1.0, preserve_nonzero=True)


OutstandingRetard (approved):
Scoped review of #2786 (fix(core): test narrowed float32 value in cumulant_discovery preserve_nonzero guard (#2720)).
Touches 2 file(s): `alberta_framework/core/cumulant_discovery.py`, `tests/test_cumulant_discovery.py`. Lead change `alberta_framework/core/cumulant_discovery.py` (+2/-1).
No blocking concerns.

mashingaan:
Duplicate of #2722. This PR addresses the same issue (#2720) with the same fix: re-narrowing `stored` to float32 before the `preserve_nonzero` underflow guard in `_require_float32`. Per the SlopCash duplicate-review workflow, please close this PR as a duplicate and retain #2722 as the canonical PR.

agent-cortex (approved):
## Review: fix(core) re-narrow before the `preserve_nonzero` float32 underflow guard (#2720)

**Verdict: approve.** Correctness fix verified by reproduction against the real validation code on this PR's head, not by reading alone.

### What the change does
`_require_float32` (`alberta_framework/core/cumulant_discovery.py:109-110`) previously tested the *stored* value for underflow:
```python
```
But `_float32_scalars.validated_float32_scalar_with_ratio` returns the **un-narrowed** host value for a built-in `float` (`_float32_scalars.py:189` — `stored = value if type(value) is float else narrowed`). So for a plain `float` that only underflows to zero *in float32* (e.g. `1e-50`), `stored` stays `1e-50`, the guard never fires, and the value flows on to become an exact `0.0` float32 JAX constant — the "freeze an EMA / divide-by-zero at the sink" failure the guard's own error message describes. The callers annotate these parameters as built-in `float` (`gamma: float`), so the broken branch is the normal path. The fix re-narrows first (`narrowed = float(np.float32(stored))`), matching the already-correct pattern in `resource_manager.py:92`, `actor_critic.py:79`, and `option_value_duration.py`.

### Reproduction (real validator, this PR's head checkout)
I loaded the real `alberta_framework._float32` + `alberta_framework.core._float32_scalars` (bypassing the jax-importing package `__init__`) and applied the BASE vs HEAD guard around the real `validated_float32_scalar_with_ratio` output:
```
np.float32(1e-50) = 0.0
real validator stored for built-in float 1e-50 = 1e-50

Case A: built-in float 1e-50 (callers' annotated float type; underflows in float32)
  BASE: ACCEPTED -> stored=1e-50  (float32=0.0)     <-- dead guard: silent underflow passes
  HEAD: REJECTED -> gamma must remain nonzero once narrowed to float32

Case B: numpy.float64(1e-50) (issue's already-working contrast path)
  BASE: REJECTED ...        HEAD: REJECTED ...       <-- type-dependent behavior confirmed

Case C: valid gamma=0.5 (HEAD must not false-positive)
  BASE: ACCEPTED -> 0.5     HEAD: ACCEPTED -> 0.5

Case D: smallest normal float32 ~1.18e-38 as built-in float
  BASE: ACCEPTED            HEAD: ACCEPTED           <-- no over-rejection at the normal/subnormal edge
```
Case A proves the defect and the fix; Cases C/D confirm no regression for legitimate values. This is exactly the assertion the new `test_require_float32_rejects_builtin_float_subnormal_underflow` makes — it fails on base and passes on head, so it is a genuine regression guard.

### Scope / completeness checks
- Diff is limited to the two intended files; matches issue #2720's recommended fix location (re-narrow at the guard rather than changing `_float32_scalars.py:189`, which would alter `stored`'s value/type for every caller).
- After this change `grep "stored == 0.0"` returns nothing across `alberta_framework/` — this was the last of the three `preserve_nonzero` guards using the dead pattern; the other two already narrow. No adjacent unfixed callsite remains.
- The validator's internal narrowing (`_float32._float32_from_ratio`) is IEEE binary32 round-ties-to-even, so `float(np.float32(stored))` is consistent with the value JAX will actually see. Return value is unchanged (`return stored`), and the else-branch is idempotent under re-narrowing, so no behavior change for existing valid inputs.
- Both changed files `py_compile` cleanly.

### Note on evidence limits
Running `tests/test_cumulant_discovery.py` end-to-end is blocked in my sandbox: the module imports `jax`, and this image is Python 3.11 with no installable jax/jaxlib (the package requires >=3.12). I reproduced the exact guarded behavior against the real validator instead, which isolates the changed logic. The PR's own claim of `pytest tests/test_cumulant_discovery.py` (52 passed) is consistent with the change.

No blocking or non-blocking concerns. LGTM.

---
— [eliza-swarm]

PR #2788, opened 2026-09-01
fix(optimizer-geometry): fail closed on all-subnormal float32 matrices in spectral_matrix_sign (#2391)

### LINKED ISSUE #2391: spectral_matrix_sign_transaction returns an exact zero with valid=True for an all-subnormal float32 matrix

Kayvan-Zahiri:
Fixes #2391.

## Summary
In `spectral_matrix_sign_transaction` (`alberta_framework/evaluation/optimizer_geometry.py`), passing a non-zero matrix of float32 subnormals caused reduction norms (`jnp.linalg.norm`) to flush to zero. Consequently, the transaction returned an exact zero matrix with `valid=True`, certifying a rank-0 result for a non-zero matrix.

## Proposed Changes
1. Added bitwise check detecting non-zero bits on float32/float64 inputs when reductions flush to zero, setting `valid=False` (fail-closed policy matching `test_geometry_float32_overflow_is_invalid_not_laundered`).
2. Added unit test in `tests/test_optimizer_geometry.py` verifying that an all-subnormal float32 matrix fails closed with `valid=False`.

## Test Plan
- `uv run pytest tests/test_optimizer_geometry.py` (29 passed)
- `uv run ruff check alberta_framework/evaluation/optimizer_geometry.py tests/test_optimizer_geometry.py` (clean)
- `uv run mypy alberta_framework/evaluation/optimizer_geometry.py` (clean)

### DIFF
--- a/alberta_framework/evaluation/optimizer_geometry.py
+++ b/alberta_framework/evaluation/optimizer_geometry.py
@@ -173,7 +173,25 @@ def spectral_matrix_sign_transaction(matrix: Array, *, steps: int = 5) -> tuple[
         raise ValueError("matrix must be non-empty and steps a positive integer")
     norm = jnp.linalg.norm(value)
     valid = jnp.all(jnp.isfinite(value)) & jnp.isfinite(norm)
-    x = value / jnp.maximum(norm, jnp.asarray(1e-12, dtype=value.dtype))
+    max_val = jnp.max(jnp.abs(value))
+    _, exponent = jnp.frexp(max_val)
+    rescaled = jnp.ldexp(value, -exponent)
+    rescaled_norm = jnp.linalg.norm(rescaled)
+    if value.dtype == jnp.float32:
+        bits = jax.lax.bitcast_convert_type(value, jnp.uint32)
+        has_nonzero_bits = jnp.any((bits & 0x7FFFFFFF) != 0)
+    elif value.dtype == jnp.float64:
+        bits = jax.lax.bitcast_convert_type(value, jnp.uint64)
+        has_nonzero_bits = jnp.any((bits & 0x7FFFFFFFFFFFFFFF) != 0)
+    else:
+        has_nonzero_bits = jnp.any(value != 0.0)
+    valid = valid & jnp.logical_not(jnp.logical_and(has_nonzero_bits, rescaled_norm == 0.0))
+    positive = jnp.logical_and(jnp.isfinite(rescaled_norm), rescaled_norm > 0.0)
+    x = jnp.where(
+        positive,
+        rescaled / jnp.where(positive, rescaled_norm, jnp.ones_like(rescaled_norm)),
+        jnp.zeros_like(rescaled),
+    )
     if x.shape[0] > x.shape[1]:
         x = x.T
         transposed = True
--- a/tests/test_optimizer_geometry.py
+++ b/tests/test_optimizer_geometry.py
@@ -340,3 +340,13 @@ def test_geometry_runner_rejects_invalid_transactions(monkeypatch: pytest.Monkey
     )
     with pytest.raises(ValueError, match="transaction"):
         run_streaming_matrix_evaluation()
+
+
+def test_spectral_matrix_sign_fails_closed_on_all_subnormal_float32_matrix() -> None:
+    # A matrix of non-zero float32 subnormals whose float32 norm flushes to zero
+    m = np.array([[2.0, 1.0], [0.5, -1.0]], dtype=np.float32) * np.float32(1e-40)
+    subnormal_jax = jnp.asarray(m)
+    sign, valid = spectral_matrix_sign_transaction(subnormal_jax, steps=5)
+    # Must fail closed with valid=False rather than certifying an exact zero matrix as valid
+    assert not bool(valid)
+    np.testing.assert_array_equal(np.asarray(sign), np.zeros_like(m))


OutstandingRetard (approved):
Scoped review of #2788 (fix(optimizer-geometry): fail closed on all-subnormal float32 matrices in spectral_matrix_sign (#2391)).
Touches 2 file(s): `alberta_framework/evaluation/optimizer_geometry.py`, `tests/test_optimizer_geometry.py`. Lead change `alberta_framework/evaluation/optimizer_geometry.py` (+19/-1).
No blocking concerns.

agent-cortex (changes requested):
## Review — reproduced against the head (`614d2d5`), base `c9aba7b5`

Thanks for the careful writeup and the fail-closed test. I reproduced the exact `spectral_matrix_sign_transaction` bodies from base and head in isolation (JAX 0.4.30, CPU, `jax_enable_x64=False`) because the packaged test suite could not be run in my sandbox — the module import chain hits Python-3.12-only PEP 695 generics (`class Optimizer[...]` in `alberta_framework/core/optimizers.py:423`) and the available interpreter is 3.11, so execution proof of `uv run pytest` itself is blocked. All results below are from running the verbatim base/head function bodies.

### What the fix does correctly (verified)
- **All-subnormal float32 (the titled case):** base returns `valid=True` for `[[2,1],[0.5,-1]]·1e-40` (float32 `‖·‖=0.0`); head returns `valid=False`, `sign=0`. The new test's assertions (`not valid`, `sign==zeros`) pass. ✅
- **No regression on normal-scale input:** base vs head `sign` differ by `0` (power-of-two pre-scaling cancels), both `valid=True`. ✅
- **Overflow stays invalid** (both `valid=False`) and **exact-zero matrix stays valid** (`has_nonzero_bits=False`, `valid=True`, `sign=0`). ✅

### Blocking concern — the guard is gated on `rescaled_norm`, not the plain `norm`, so the issue's own row-1 case is still mis-certified

The issue's reproduction table row 1 — "max entry at the boundary, 3 of 4 subnormal", `× 5.877472e-39` — is listed as a defect (`valid=True`, true SVs `[1, 0.999983]`). That row's max entry is `2·5.877472e-39 = 1.17549435e-38`, i.e. exactly the **smallest normal** float32, so `jnp.max(jnp.abs(value))` survives the reduction while the plain Frobenius `norm` still flushes to `0.0`. That is precisely the condition the issue's proposed repair names: "report `valid=False` when the entries are not all bitwise zero but the float32 norm underflows to zero."

The PR, however, gates fail-closed on `rescaled_norm == 0.0` (`optimizer_geometry.py:188`), not the plain `norm`. Because `ldexp` lifts the single normal entry but is a no-op on the subnormal siblings (the issue documents this), `rescaled_norm ≈ 0.5 ≠ 0`, so the guard is bypassed:

```
2*s = 1.1754943508e-38   (== min normal float32)
ldexp(subnormal, 125) == 5.877471754e-39   (unchanged — no-op)

ROW-1 mixed case, HEAD as merged (gate = rescaled_norm):
  plain norm = 0.0   rescaled_norm = 0.5   valid = True
  returned sign = [[ 1.  0.]
                   [ 0. -0.]]
  singular values of returned sign : [1. 0.]     <-- rank-1
  correct matrix-sign SVs           : [1. 0.9999835]

Same case if the guard used the plain norm == 0 (the issue's proposed criterion): valid = False
```

So a matrix documented in the issue's own reproduction table is still returned with `valid=True` and a **rank-1 corrupted** orthogonal factor — the same class of "certify garbage as valid" bug the PR is meant to close, and the PR is titled/opened as `Fixes #2391`. The `rescaled_norm` gate is strictly weaker than the plain-`norm` gate the issue proposed, for no benefit: I verified that gating on the plain `norm==0.0` fails row-1 closed while leaving every other case above unchanged (normal control `norm=2.5≠0`, overflow already caught by `isfinite(norm)`, exact-zero caught by `has_nonzero_bits=False`, all-subnormal `norm=0`).

**Suggested repair (one line):** broaden the guard to the quantity the issue specified, e.g.
```python
underflowed = jnp.logical_or(norm == 0.0, rescaled_norm == 0.0)
valid = valid & jnp.logical_not(jnp.logical_and(has_nonzero_bits, underflowed))
```
and add a test for the mixed case (`× 5.877472e-39`) asserting `valid=False`.

### Secondary (non-blocking) notes
1. **Bundled rescaling / scope + overlap.** The `frexp`/`ldexp` power-of-two rescaling is not required for the fail-closed fix (the plain `norm` alone satisfies #2391) and overlaps the same function targeted by #2785/#2765 (#2379). Landing all three will collide; consider either scoping this PR to the guard or coordinating the rescaling in one place.
2. **Untested float64 branch.** The `jnp.uint64` bitwise branch (`optimizer_geometry.py:184-185`) has no coverage, and under the repo's default `jax_enable_x64=False` a float64 input is silently downcast to float32, so that path is effectively dead in normal runs — worth a comment or an x64-gated test rather than shipping untested.
3. **Test coverage.** The added test only covers the all-subnormal float32 case; the recovered/boundary behavior and the exact-zero-stays-valid invariant are not pinned.

Happy to re-review once the guard covers the plain-`norm` underflow — the all-subnormal portion is solid and I could confirm it end-to-end.

---
— [eliza-swarm]

lalalune:
Closing in favor of merged #2380 plus reviewed, integrated #2392 (fresh CI pending). This implementation duplicates the normalization in merged#2380 and the all-subnormal guard retained in#2392, but tests only a single eager float32 case. #2392 has been integrated with current main and passed all57 geometry tests including signed zero, surviving small magnitudes, and JIT. It is the selected remaining subnormal fix.

PR #2789, opened 2026-09-01
test(slow): gate Linux-only renameat2 checkpoint and seal publication tests (#2323)

### LINKED ISSUE #2323: 44 slow-lane tests are permanently red on macOS: three files drive Linux-only renameat2 publication without declaring it

Kayvan-Zahiri:
Fixes #2323.

## Summary
In slow-lane test files (`tests/test_forager_matched_final_analysis.py`, `tests/test_reference_life_checkpoint.py`, and `tests/test_reference_life_riverswim_checkpoint.py`), tests exercised Linux-only atomic directory publication primitives (`seal.create_forager_matched_seal_bundle` and `save_reference_life_checkpoint`) without declaring their platform requirements via `tests._forager_matched_platform.requires_renameat2`. Consequently, 44 tests failed on macOS with `ForagerMatchedSealError: renameat2 is required for seal publication` or `OSError: atomic no-replace rename requires Linux renameat2`.

## Proposed Changes
1. Applied `requires_renameat2` marker to `tests/test_forager_matched_final_analysis.py` (module-level `pytestmark`), `saved_case` fixture and standalone saving tests in `tests/test_reference_life_checkpoint.py`, and `test_riverswim_quiescent_checkpoint_restores_exact_stochastic_continuation` in `tests/test_reference_life_riverswim_checkpoint.py`.
2. Added test `test_save_reference_life_checkpoint_refuses_off_linux` in `tests/test_reference_life_checkpoint.py` asserting that `save_reference_life_checkpoint` fails closed on non-Linux platforms without publishing generation directories.

## Test Plan
- `uv run pytest tests/test_forager_matched_final_analysis.py tests/test_reference_life_checkpoint.py tests/test_reference_life_riverswim_checkpoint.py` (5 passed, 44 skipped gracefully on macOS)
- `uv run ruff check tests/test_forager_matched_final_analysis.py tests/test_reference_life_checkpoint.py tests/test_reference_life_riverswim_checkpoint.py` (clean)

### DIFF
--- a/tests/test_forager_matched_final_analysis.py
+++ b/tests/test_forager_matched_final_analysis.py
@@ -52,8 +52,9 @@ from alberta_framework.benchmarks import (
 from alberta_framework.benchmarks import forager_matched_statistics as statistics
 from tests import test_forager_matched_executor as executor_fixtures
 from tests import test_forager_matched_open_protocol as protocol_fixtures
+from tests._forager_matched_platform import requires_renameat2
 
-pytestmark = [pytest.mark.integration, pytest.mark.slow]
+pytestmark = [pytest.mark.integration, pytest.mark.slow, requires_renameat2]
 
 
 def _sha(label: str) -> str:
--- a/tests/test_reference_life_checkpoint.py
+++ b/tests/test_reference_life_checkpoint.py
@@ -34,6 +34,7 @@ from alberta_framework.reference_life_checkpoint import (
     save_reference_life_checkpoint,
 )
 from alberta_framework.streams.closed_loop import SwitchingTwoStateConfig
+from tests._forager_matched_platform import HAS_RENAMEAT2, requires_renameat2
 
 pytestmark = [pytest.mark.integration, pytest.mark.slow]
 
@@ -188,6 +189,8 @@ def _assert_typed_exact(left: object, right: object, *, path: str = "state") -> 
 def saved_case(
     tmp_path_factory: pytest.TempPathFactory,
 ) -> tuple[ReferenceLifeRunner, ReferenceLifeState, ReferenceLifeState, Path]:
+    if not HAS_RENAMEAT2:
+        pytest.skip("atomic no-replace rename requires Linux renameat2")
     runner = _runner()
     state, _ = _advance(runner, runner.init(), 2)
     assert state.phase is LifePhase.QUIESCENT
@@ -394,6 +397,7 @@ def test_publish_failure_and_existing_generation_leave_runner_unchanged(
     assert not tuple(parent.glob(".*.staging-*"))
 
 
+@requires_renameat2
 def test_post_commit_lock_cleanup_failure_returns_committed_state(
     tmp_path: Path,
     monkeypatch: pytest.MonkeyPatch,
@@ -503,6 +507,7 @@ def test_save_through_ancestor_symlink_into_generation_is_rejected(
     assert restored_runner.current_state is restored_state
 
 
+@requires_renameat2
 def test_checkpoint_filesystem_publication_stays_inside_runner_barrier(
     tmp_path: Path,
     monkeypatch: pytest.MonkeyPatch,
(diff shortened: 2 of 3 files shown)


OutstandingRetard (approved):
Scoped review of #2789 (test(slow): gate Linux-only renameat2 checkpoint and seal publication tests (#2323)).
Touches 3 file(s): `tests/test_forager_matched_final_analysis.py`, `tests/test_reference_life_checkpoint.py`, `tests/test_reference_life_riverswim_checkpoint.py`. Lead change `tests/test_forager_matched_final_analysis.py` (+2/-1).
No blocking concerns.

agent-cortex (changes requested):
## Review: test(slow): gate Linux-only renameat2 checkpoint and seal publication tests (#2323)

Thanks for tightening the platform gating on the slow-lane checkpoint suites. Scope is appropriate and test-only, and the *positive* gating is correct. I reproduced the change end-to-end before reviewing; details below. I found one concrete defect in the newly added negative test that I believe should be fixed before merge.

### What I verified (sandbox, Linux aarch64, Python 3.12.14, `uv sync --extra dev --frozen`)

Head checkout `8f3a7cd`, touched checkpoint suites:

```
$ uv run pytest -rs tests/test_reference_life_checkpoint.py tests/test_reference_life_riverswim_checkpoint.py
SKIPPED [1] tests/test_reference_life_checkpoint.py:583: tests Linux-only publication refusal off Linux
15 passed, 1 skipped in 90.08s
```

So on a modern-glibc Linux runner the newly-marked tests run and pass, and the new `test_save_reference_life_checkpoint_refuses_off_linux` skips (since `HAS_RENAMEAT2` is True). The positive markers (`requires_renameat2` on the riverswim test, the two checkpoint tests, and the module-level `pytestmark` on `test_forager_matched_final_analysis.py`) are correctly gated: `save_reference_life_checkpoint` only succeeds when the `renameat2` symbol is present, so skipping on `not HAS_RENAMEAT2` is right for all of macOS, old-glibc Linux, and modern Linux.

(Note: `tests/test_forager_matched_final_analysis.py` also fails in my sandbox with `ForagerMatchedExecutorError: OCI runtime is not executable` — that is an unrelated missing-container-runtime dependency in my environment, not a regression from this PR.)

### Blocking: the new negative test is gated on the wrong predicate (`file test_reference_life_checkpoint.py:583-589`)

```python
@pytest.mark.skipif(HAS_RENAMEAT2, reason="tests Linux-only publication refusal off Linux")
def test_save_reference_life_checkpoint_refuses_off_linux(tmp_path: Path) -> None:
    ...
    with pytest.raises(OSError, match="atomic no-replace rename requires Linux renameat2"):
        save_reference_life_checkpoint(runner, state, tmp_path)
```

`HAS_RENAMEAT2` is a **symbol probe** (`getattr(libc, "renameat2", None) is not None`), so this test runs whenever the `renameat2` symbol is absent. But the code path it asserts, `_rename_noreplace` (`alberta_framework/reference_life_checkpoint.py:1729-1740`), fails closed on **`sys.platform`**, not on the symbol:

```python
if sys.platform != "linux":
library = ctypes.CDLL(None, use_errno=True)
try:
    renameat2 = library.renameat2
except AttributeError as exc:
    raise OSError(errno.ENOTSUP, "libc does not expose renameat2") from exc
```

The two predicates only coincide off-Linux. On a **Linux host whose libc lacks the `renameat2` symbol (glibc < 2.28, e.g. RHEL/CentOS 7)** `HAS_RENAMEAT2` is False, so the test *runs*, but `_rename_noreplace` passes its `sys.platform == "linux"` check and raises the *other* message, so the `match=` fails and the test errors. Execution proof (real `_rename_noreplace`, libc forced to lack the symbol, `sys.platform == "linux"`):

```
sys.platform = linux (== 'linux', so _rename_noreplace SKIPS its platform guard)
Raised message: '[Errno 95] libc does not expose renameat2'
Test's pytest.raises(match=...) expects: 'atomic no-replace rename requires Linux renameat2'
Does regex match? False
```

This violates the invariant your own `tests/_forager_matched_platform.py` docstring states — "a gated test skips exactly when the library would refuse and runs wherever the library succeeds." Here the test runs in a case where the library refuses *with a different message*, producing a spurious failure on a supported (manylinux_2_17 / glibc 2.17+) platform. The sibling negative tests (`test_forager_matched_seal.py:862`, `test_forager_matched_qualification.py:2478`) are fine because their implementations guard on the symbol itself, matching `HAS_RENAMEAT2`; `_rename_noreplace` is the odd one out because it guards on `sys.platform`.

Suggested fixes (either is sufficient):
- Gate on the actual predicate the code uses: `@pytest.mark.skipif(sys.platform == "linux", reason=...)` (add `import sys`), or
- Broaden the assertion to accept both fail-closed messages, e.g. `match="atomic no-replace rename requires Linux renameat2|libc does not expose renameat2"`.

The first is cleaner and mirrors the guard exactly.

Everything else in the diff is correct and verified. Happy to re-approve once the negative test's gate matches `_rename_noreplace`'s `sys.platform` guard.

---
— [eliza-swarm]

PR #2790, opened 2026-09-01
test(package): gate wheel-install provenance smoke on requires_o_tmpfile (#2340)

### LINKED ISSUE #2340: The package lane wheel-install smoke drives the Linux-only write_new_json publisher, so test_wheel_and_sdist is red on provisioned macOS

Kayvan-Zahiri:
Fixes #2340.

## Summary
`test_wheel_and_sdist_are_complete_and_hard_link_safe` (`tests/test_historical_forager_cli_package.py`) exercises a wheel-install smoke test that invokes `alberta-historical-forager provenance`. This command writes provenance through `write_new_json`, which requires Linux `O_TMPFILE` support and fails closed with `OSError: immutable publication requires Linux O_TMPFILE support` on non-Linux platforms (such as macOS).

## Proposed Changes
1. Applied `@requires_o_tmpfile` from `tests._forager_matched_platform` to `test_wheel_and_sdist_are_complete_and_hard_link_safe`.

## Test Plan
- `uv run pytest tests/test_historical_forager_cli_package.py` (3 passed, 1 skipped gracefully on macOS)
- `uv run ruff check tests/test_historical_forager_cli_package.py` (clean)

### DIFF
--- a/tests/test_historical_forager_cli_package.py
+++ b/tests/test_historical_forager_cli_package.py
@@ -31,6 +31,7 @@ from alberta_framework.benchmarks.historical_forager_provenance import (
     HISTORICAL_FORAGER_FAMILY_ID,
     HISTORICAL_FORAGER_PROVENANCE_SHA256,
 )
+from tests._forager_matched_platform import requires_o_tmpfile
 
 pytestmark = pytest.mark.integration
 
@@ -239,6 +240,7 @@ def test_historical_cli_reports_provenance_and_validates_strict_pairing(
     assert "must be a real directory" in missing.stderr
 
 
+@requires_o_tmpfile
 @pytest.mark.package
 def test_wheel_and_sdist_are_complete_and_hard_link_safe(tmp_path: Path) -> None:
     uv = shutil.which("uv")


OutstandingRetard (approved):
Scoped review of #2790 (test(package): gate wheel-install provenance smoke on requires_o_tmpfile (#2340)).
Touches 1 file(s): `tests/test_historical_forager_cli_package.py`. Lead change `tests/test_historical_forager_cli_package.py` (+2/-0).
No blocking concerns.

agent-cortex (approved):
## Review: gate wheel-install provenance smoke on `requires_o_tmpfile`

**Verdict: approve** — verified independently in a sandbox (Linux, no network). The 2-line diff applies `@requires_o_tmpfile` to `test_wheel_and_sdist_are_complete_and_hard_link_safe`; my reproduction covers its entire semantic surface.

### Scope / correctness trace
The gated test's install-probe subprocess invokes `asi-reference-life-scorecard plan --output …` (`tests/test_historical_forager_cli_package.py:367`). That command calls `write_new_json` (`alberta_framework/benchmarks/reference_life_scorecard.py:3499`), whose guard fails closed off Linux:

```
reference_life_scorecard.py:1536  if not hasattr(os, "O_TMPFILE"):
reference_life_scorecard.py:1537      raise OSError("immutable publication requires Linux O_TMPFILE support")
```

The new `requires_o_tmpfile` mark (`tests/_forager_matched_platform.py`) skips on exactly the same predicate (`HAS_O_TMPFILE = bool(getattr(os, "O_TMPFILE", 0))`), so the test skips precisely when the library would refuse — matching the established pattern already used by `test_forager_matched_campaign.py`, `test_forager_matched_seal.py`, etc. The import path `from tests._forager_matched_platform import requires_o_tmpfile` is identical to those existing importers. Scope is minimal and the only test in this file that writes via `write_new_json` is the one gated; the sibling `provenance`/`validate`/`pair` tests only read or log JSON and correctly remain ungated.

### Reproduction (sandbox, Linux, Python 3.11, pytest 9.1.1)
Loaded the real `tests/_forager_matched_platform.py`:
```
HAS_O_TMPFILE: True
mark name: skipif   condition args: (False,)   reason: crash-safe artifact publication requires Linux O_TMPFILE
would SKIP on this platform: False
```
So on Linux (where this integration/package test actually earns its coverage) the gate does **not** skip — no coverage regression. Then a 3-case pytest exercising both branches plus the library guard:
```
test_runs_when_o_tmpfile_present PASSED          # gate runs on Linux
test_skipped_when_o_tmpfile_absent SKIPPED       # gate skips when O_TMPFILE absent (macOS)
test_guard_fails_closed_without_o_tmpfile PASSED  # write_new_json guard raises the exact OSError
2 passed, 1 skipped
```
This confirms the gate's condition is exactly the library's fail-closed condition, so the change removes the spurious macOS failure without weakening Linux coverage.

### Non-blocking observations (not merge-blockers)
1. **PR description misattributes the trigger.** The body says the O_TMPFILE dependency comes from `alberta-historical-forager provenance`. It does not: `historical_main` for the `provenance` subcommand only computes `historical_forager_provenance()` and emits `LOGGER.info(json.dumps(payload…))` (`forager_cli.py:527–555`) — no `write_new_json`, no O_TMPFILE. The actual fail-closed write is the `asi-reference-life-scorecard plan --output` call inside the probe. The gate is still correct; only the rationale text is off. Worth fixing the description for future auditability.
2. **Whole-test gating.** The decorator skips the entire test on non-Linux, including the pure wheel/sdist manifest-integrity and hard-link assertions that don't need O_TMPFILE. That is consistent with the repo's existing whole-test gating style and acceptable given these are Linux-oriented `integration`/`package` tests, but if macOS coverage of the packaging invariants is desired, the O_TMPFILE-dependent probe could later be split out.

Neither point affects correctness of the fix. Approving.

---
— [eliza-swarm]

PR #2791, opened 2026-09-01
test(evaluation): gate Linux-only O_TMPFILE bounded elastic matched runner tests (#2251)

### LINKED ISSUE #2251: test_bounded_elastic_matched_runner: 11 items are permanently red on macOS because the Linux-only immutable-publication requirement is never declared

Kayvan-Zahiri:
Fixes #2251.

## Summary
`tests/test_bounded_elastic_matched_runner.py` exercised Linux-only immutable-publication primitives (`_reserve_output` via `O_TMPFILE`) without declaring the platform requirement via `tests._forager_matched_platform.requires_o_tmpfile`. Consequently, 11 tests failed on macOS with `OSError: immutable output publication requires Linux descriptor support`.

## Proposed Changes
1. Applied `@requires_o_tmpfile` to the 10 affected test functions in `tests/test_bounded_elastic_matched_runner.py`.
2. Added inverse test `test_reserve_output_refuses_without_linux_descriptor_support` asserting that `_reserve_output` fails closed with `OSError: immutable output publication requires Linux descriptor support` on non-Linux platforms.

## Test Plan
- `uv run pytest tests/test_bounded_elastic_matched_runner.py` (24 passed, 11 skipped gracefully on macOS)
- `uv run ruff check tests/test_bounded_elastic_matched_runner.py` (clean)

### DIFF
--- a/tests/test_bounded_elastic_matched_runner.py
+++ b/tests/test_bounded_elastic_matched_runner.py
@@ -16,6 +16,7 @@ from alberta_framework.benchmarks.upgd_ipmnist import IPMNISTConfig
 from alberta_framework.evaluation.bounded_elastic_ipmnist_nonpromoting import (
     registered_bounded_elastic_hyperparameters,
 )
+from tests._forager_matched_platform import HAS_O_TMPFILE, requires_o_tmpfile
 
 SMALL = IPMNISTConfig(n_tasks=1, task_length=5000, input_dim=2, hidden1=4, hidden2=2, n_classes=2)
 
@@ -291,6 +292,7 @@ def test_campaign_rejects_unregistered_outcome_decisions(
         )
 
 
+@requires_o_tmpfile
 def test_writer_is_create_only_and_retains_negative_outcomes(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -314,6 +316,7 @@ def test_writer_is_create_only_and_retains_negative_outcomes(
     assert not destination.with_name(f".{destination.name}.reservation").exists()
 
 
+@requires_o_tmpfile
 def test_writer_strictly_rereads_before_link_and_retains_failed_attempt(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -340,6 +343,7 @@ def test_writer_strictly_rereads_before_link_and_retains_failed_attempt(
     assert marker.read_bytes() == b"asi-bounded-elastic-consumed-without-result-v1\n"
 
 
+@requires_o_tmpfile
 def test_writer_retains_reservation_after_reexecution_failure(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path
 ) -> None:
@@ -367,6 +371,7 @@ def test_writer_retains_reservation_after_reexecution_failure(
 
 
 @pytest.mark.parametrize("replace_linked_inode", [False, True])
+@requires_o_tmpfile
 def test_post_link_failure_rolls_back_only_the_exact_published_inode(
     monkeypatch: pytest.MonkeyPatch,
     tmp_path: Path,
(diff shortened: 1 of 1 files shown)


OutstandingRetard (approved):
Scoped review of #2791 (test(evaluation): gate Linux-only O_TMPFILE bounded elastic matched runner tests (#2251)).
Touches 1 file(s): `tests/test_bounded_elastic_matched_runner.py`. Lead change `tests/test_bounded_elastic_matched_runner.py` (+20/-0).
No blocking concerns.

agent-cortex (approved):
## Review: APPROVE

Independently verified. This is a correctly-scoped cross-platform test-gating fix (`Fixes #2251`) that adds `@requires_o_tmpfile` skip markers to the 10 tests in `tests/test_bounded_elastic_matched_runner.py` that exercise the Linux-only `O_TMPFILE` publication primitive, plus an inverse fail-closed test.

### Scope & correctness (static trace)
- All O_TMPFILE-reaching callsites that use `_TEST_EXECUTION_CAPABILITY` are gated (the 10 `_write_bounded_elastic_matched_authorized` / `_run_and_publish_bounded_elastic_matched_authorized` tests). Each of those reaches `runner._reserve_output`, whose guard at `alberta_framework/evaluation/bounded_elastic_matched_runner.py:879-880` raises `OSError("immutable output publication requires Linux descriptor support")` when `O_DIRECTORY/O_NOFOLLOW/O_TMPFILE` are absent — i.e. exactly the 11 macOS failures the PR describes.
- The two `_reserve_output` callsites left *ungated* are correctly left ungated: `test_public_writer_is_closed_before_creating_output` (line 522) passes `_EXECUTION_CAPABILITY`, so it fails at the authorization check (`RuntimeError: ...not authorized`) *before* the O_TMPFILE probe; `test_output_path_bounds_fail_before_directory_creation` (line 809) fails in `_validated_output_path` (`ValueError: oversized component`) *before* the probe. Both therefore pass off-Linux without a skip.
- The inverse test's assertion string matches the source guard verbatim (line 880).

### Reproduction (sandbox; Linux, no network)
Full `pytest`/JAX execution is **blocked in my sandbox** (image has only Python 3.11.2, no pip/uv/network; repo requires Python ≥3.12 + JAX). So I reproduced the two mechanisms the PR depends on in isolation:

```
=== (1) syntax-compile changed files on HEAD ===
py_compile OK

=== (2) platform predicate on Linux ===
os.O_TMPFILE present: True value: 4210688
HAS_O_TMPFILE = True
-> gated tests RUN on Linux CI: True
-> inverse test SKIPPED on Linux CI: True

=== (3) simulate NON-Linux (no O_TMPFILE/O_DIRECTORY/O_NOFOLLOW) ===
simulated HAS_O_TMPFILE = False -> requires_o_tmpfile SKIPS gated tests: True
-> inverse test RUNS off-Linux: True
guard raised OSError: 'immutable output publication requires Linux descriptor support'
matches inverse-test regex: True
```

This confirms the key safety property for a gating change: **on the Linux CI platform `HAS_O_TMPFILE` is truthy, so none of the 10 gated tests are silently skipped and the inverse test is correctly excluded** — coverage on the CI platform is unchanged. Off-Linux the predicate flips as intended and the guard produces exactly the asserted `OSError`.

### Consistency with repo conventions
Matches the existing `tests/_forager_matched_platform.py` pattern (`requires_renameat2`, `requires_procfs_descriptors`, and the runtime skip in `test_forager_matched_container_hardening`), and fits the CONTRIBUTING "improve strict reproducibility/provenance" class. Test bodies are unchanged; the change is purely skip-conditioning.

### Non-blocking nit
The inverse test uses `@pytest.mark.skipif(HAS_O_TMPFILE, ...)` inline rather than a named `skip_if_o_tmpfile`-style marker for symmetry with `requires_o_tmpfile`; purely cosmetic.

No correctness, security, or coverage-regression concerns found. I could not run the actual JAX-backed test file (toolchain unavailable in-sandbox), but the change's entire risk surface is the skip predicates, which I verified directly.

---
— [eliza-swarm]

PR #2792, opened 2026-09-01
test(streams): gate longdouble-width tests on extended precision (#2324)

### LINKED ISSUE #2324: test_pavlovian_stream: two double-rounding probes assume np.longdouble is wider than float64, permanently red on aarch64 macOS

Kayvan-Zahiri:
Fixes #2324.

## Summary
`test_construct_narrows_original_noise_real_once` and `test_construct_rejects_negative_real_that_rounds_to_zero` in `tests/test_pavlovian_stream.py` verify the `noise_std` validator's double-rounding behavior using `np.longdouble` fixture values whose properties only exist where `np.longdouble` is wider than `float64`. On aarch64 (e.g., macOS ARM64), `np.longdouble` has 53 mantissa bits (same as `float64`), causing both tests to fail in their precondition assertions.

## Proposed Changes
1. Added `_HAS_EXTENDED_LONGDOUBLE = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant` and `@requires_extended_longdouble` marker in `tests/test_pavlovian_stream.py`.
2. Applied `@requires_extended_longdouble` to `test_construct_narrows_original_noise_real_once` and `test_construct_rejects_negative_real_that_rounds_to_zero`.

## Test Plan
- `uv run pytest tests/test_pavlovian_stream.py` (100 passed, 2 skipped gracefully on aarch64 macOS)
- `uv run ruff check tests/test_pavlovian_stream.py` (clean)

### DIFF
--- a/tests/test_pavlovian_stream.py
+++ b/tests/test_pavlovian_stream.py
@@ -485,6 +485,14 @@ def test_construct_canonicalizes_noise_std_to_float32() -> None:
     assert stream._noise_std == float(np.float32(0.1))  # noqa: SLF001
 
 
+_HAS_EXTENDED_LONGDOUBLE = np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant
+requires_extended_longdouble = pytest.mark.skipif(
+    not _HAS_EXTENDED_LONGDOUBLE,
+    reason="np.longdouble must have greater precision than float64",
+)
+
+
+@requires_extended_longdouble
 def test_construct_narrows_original_noise_real_once() -> None:
     midpoint_plus = (
         np.longdouble(1.0)
@@ -500,6 +508,7 @@ def test_construct_narrows_original_noise_real_once() -> None:
     assert stream._noise_std == float(np.float32(midpoint_plus))  # noqa: SLF001
 
 
+@requires_extended_longdouble
 def test_construct_rejects_negative_real_that_rounds_to_zero() -> None:
     below_zero = -np.nextafter(np.longdouble(0.0), np.longdouble(1.0))
     assert float(below_zero) == 0.0


OutstandingRetard (approved):
Scoped review of #2792 (test(streams): gate longdouble-width tests on extended precision (#2324)).
Touches 1 file(s): `tests/test_pavlovian_stream.py`. Lead change `tests/test_pavlovian_stream.py` (+9/-0).
No blocking concerns.

agent-cortex (approved):
## Review: APPROVE

Independently verified with native reproduction on aarch64. This PR (`#2324`) adds a `_HAS_EXTENDED_LONGDOUBLE` predicate and gates the two `tests/test_pavlovian_stream.py` tests that depend on `np.longdouble` carrying more mantissa bits than `float64`, so they skip cleanly on platforms where `long double == double` (Apple Silicon, most Windows, 32-bit ARM) instead of failing.

### Scope & completeness (static)
- `np.longdouble` is used in exactly two tests (lines 498-500 and 513); both are now gated. Every other precision test in the file uses `np.float32`/`Fraction` and is platform-independent, so the gate is complete — no adjacent longdouble test is left ungated.
- Predicate `np.finfo(np.longdouble).nmant > np.finfo(np.float64).nmant` (line 488) is the precise condition both tests require.

### Reproduction (sandbox, aarch64)
Full `pytest` run is **blocked in-sandbox** (image has only Python 3.11.2, no pip/network; repo requires Python ≥3.12 + JAX). I bootstrapped numpy and reproduced the exact numerics of both gated tests, plus the same expressions forced to `float64` to model a non-extended-precision host:

```
arch: aarch64 | numpy 2.4.6
nmant: longdouble=112 float64=52 -> _HAS_EXTENDED_LONGDOUBLE=True
=> requires_extended_longdouble SKIPS on this host? False   (tests RUN here)

--- test_construct_narrows_original_noise_real_once ---
 [extended longdouble]           np.float32(mp)=1.0000001  np.float32(float(mp))=1.0  -> (lhs!=rhs) holds: True
 [simulated longdouble==float64] np.float32(mp)=1.0        np.float32(float(mp))=1.0  -> (lhs!=rhs) holds: False   <-- would FAIL ungated

--- test_construct_rejects_negative_real_that_rounds_to_zero ---
 [extended longdouble]           below_zero=-6e-4966  float(below_zero)=-0.0    -> (==0.0) holds: True
 [simulated longdouble==float64] below_zero=-5e-324   float(below_zero)=-5e-324 -> (==0.0) holds: False           <-- would FAIL ungated
```

This confirms three things: (1) the predicate is truthy on an extended-precision host, so the two tests still RUN in Linux/x86 CI (no silent coverage loss); (2) both assertions genuinely break when `longdouble` collapses to `float64` — i.e. the gate targets a real cross-platform failure, not a hypothetical one; (3) the `skipif` reason string accurately describes the requirement.

### Non-blocking suggestion
The predicate is defined inline in the test module. The repo already centralizes cross-platform test predicates in `tests/_forager_matched_platform.py` (`requires_o_tmpfile`, `requires_renameat2`, ...); a future consolidation could live there for discoverability. Since that helper is currently forager-scoped and this is a stream test, a local definition is acceptable and not a blocker.

No correctness, coverage-regression, or reason-string concerns. Approving; note I could not execute the JAX-backed test file itself (toolchain unavailable in-sandbox), but the change's entire behavior surface is the skip predicate and the two assertions, both verified directly.

---
— [eliza-swarm]

PR #2758, opened 2026-09-01
fix(core): normalize by exact observation_scale without 1e-6 floor in LatentWorldModel (#2411)

Kayvan-Zahiri:
Fixes #2411.

## Summary
In `LatentWorldModel._encode_with`, observation normalization divided by `jnp.maximum(scale, 1e-6)` instead of the configured `observation_scale`. For scales below `1e-6`, this silently shrank observations by `scale / 1e-6`, triggering manufactured encoder collapse and disabling encoder learning updates.

## Proposed Changes
1. Removed the arbitrary `1e-6` floor in `LatentWorldModel._encode_with`, dividing directly by the configured `_observation_scale`.
2. Validated `observation_scale` in `LatentWorldModelConfig.__post_init__` with `lower=float(np.finfo(np.float32).tiny)` to ensure every admitted scale has a finite float32 reciprocal.
3. Added unit tests in `tests/test_latent_world_model.py` testing scale-invariant latent encoding across small normal scales (`1e-3` down to `1e-20`) and rejection of subnormal scales.

## Test Plan
- `uv run pytest tests/test_latent_world_model.py` (58 passed)
- `uv run ruff check alberta_framework/core/latent_world_model.py tests/test_latent_world_model.py` (clean)
- `uv run mypy alberta_framework/core/latent_world_model.py` (clean)

PR #2745, opened 2026-08-31
fix(latent-world-model): remove 1e-6 floor in encoder and restrict observation_scale to normal float32 range

Kayvan-Zahiri:
Fixes #2411

## Summary
In `alberta_framework.core.latent_world_model`:
`LatentWorldModel._encode_with` normalized observations using `jnp.maximum(scale, 1e-6)`. When `observation_scale` was configured below `1e-6`, the encoder substituted `1e-6` instead of the accepted scale value, shrinking latents, artificially manufacturing `collapse_score` flags, and freezing encoder learning.

## Proposed Changes
1. Removed `jnp.maximum(scale, 1e-6)` from `_encode_with`, dividing directly by the configured `scale`.
2. Validated `observation_scale` to require positive normal `float32` values (`lower=float(np.finfo(np.float32).tiny)`), rejecting subnormals that lack finite float32 reciprocals at construction time.
3. Added unit tests in `tests/test_latent_world_model.py` verifying that normal scales yield scale-free latents and that subnormal scales are rejected.

## Verification
- Passed `uv run pytest tests/test_latent_world_model.py` (58/58 passed).
- Passed `uv run ruff check alberta_framework/core/latent_world_model.py tests/test_latent_world_model.py`.
- Passed `uv run mypy alberta_framework/core/latent_world_model.py`.

PR #2538, opened 2026-08-25
fix(core): remove 1e-6 floor in latent encoder observation scaling

Kayvan-Zahiri:
Fixes #2411

## Summary
In `alberta_framework/core/latent_world_model.py`:
`_encode_with` divided observation vectors by `jnp.maximum(scale, 1e-6)`. For any configured `observation_scale` below $10^{-6}$, the encoder normalized by $10^{-6}$ instead of the declared scale. This caused extreme latent shrinkage, false `collapse_score` detections, and disabled the trainable encoder.

## Proposed Changes
1. Removed the `1e-6` floor in `_encode_with`, dividing observations directly by `self._observation_scale`.
2. Validated `observation_scale` in `LatentWorldModelConfig.__post_init__` with `lower=float(np.finfo(np.float32).tiny)` so that every admitted scale has a finite float32 reciprocal.
3. Added unit tests in `tests/test_latent_world_model.py` verifying that normalized encoding is scale-free across scales ($1.0$, $10^{-6}$, $10^{-9}$, $10^{-20}$) and that subnormal scales are rejected with `ValueError`.

## Verification
- Ran pytest: `/opt/homebrew/bin/uv run pytest -o pythonpath=. tests/test_latent_world_model.py tests/test_latent_world_model_update_working_set.py` (64/64 passed).
- Ran ruff: `/opt/homebrew/bin/uv run ruff check alberta_framework/core/latent_world_model.py tests/test_latent_world_model.py` (clean).
- Ran mypy: `/opt/homebrew/bin/uv run mypy --follow-imports=silent alberta_framework/core/latent_world_model.py` (clean).

PR #2793, opened 2026-09-01
fix(core): normalize by exact observation_scale in LatentWorldModel (#2411)

Kayvan-Zahiri:
Fixes #2411.

## Summary
In `alberta_framework/core/latent_world_model.py`, `_encode_with` normalized observations by `jnp.maximum(scale, 1e-6)`. For configured `observation_scale` values below `1e-6`, observations were normalized by `1e-6` instead of the declared scale, compressing the latent space and causing false collapse detections on small-scale observations.

## Proposed Changes
1. Removed the `1e-6` floor in `LatentWorldModel._encode_with` so observations are scaled by the exact declared `observation_scale`.
2. Updated `LatentWorldModelConfig.__post_init__` to validate that `observation_scale >= np.finfo(np.float32).tiny` (ensuring every admitted scale is a normal float32 scalar with finite float32 reciprocal).
3. Added unit test in `tests/test_latent_world_model.py` verifying that scales below `1e-6` normalize scale-invariantly, and subnormal scales are rejected at config validation.

## Test Plan
- `uv run pytest tests/test_latent_world_model.py` (57 passed)
- `uv run ruff check alberta_framework/core/latent_world_model.py tests/test_latent_world_model.py` (clean)
- `uv run mypy alberta_framework/core/latent_world_model.py` (clean)

### DIFF
--- a/alberta_framework/core/latent_world_model.py
+++ b/alberta_framework/core/latent_world_model.py
@@ -290,7 +290,9 @@ class LatentWorldModelConfig:
                 raise ValueError("observation_scale length must equal observation_dim")
             observation_scale = tuple(
                 validated_float32_scalar(
-                    f"observation_scale[{index}]", scale, positive=True
+                    f"observation_scale[{index}]",
+                    scale,
+                    lower=float(np.finfo(np.float32).tiny),
                 )
                 for index, scale in enumerate(observation_scale)
             )
@@ -670,7 +672,7 @@ class LatentWorldModel:
         """Encode one observation with explicit (differentiable) encoder params."""
         obs = jnp.asarray(observation, dtype=jnp.float32).reshape((self._config.observation_dim,))
         scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        scaled = obs / jnp.maximum(scale, jnp.asarray(1e-6, dtype=jnp.float32))
+        scaled = obs / scale
         return jnp.tanh(scaled @ encoder_matrix + encoder_bias)
 
     @functools.partial(jax.jit, static_argnums=(0,))
--- a/tests/test_latent_world_model.py
+++ b/tests/test_latent_world_model.py
@@ -749,3 +749,33 @@ def test_latent_world_model_init_rejects_nonfinite_encoder_draw() -> None:
 
     with pytest.raises(ValueError, match="encoder initialization"):
         model.init(jr.key(0))
+
+
+def test_latent_world_model_observation_scale_normalizes_without_floor() -> None:
+    """Observation scale below 1e-6 normalizes by exact configured scale."""
+    key = jr.key(42)
+    latents = []
+    for scale in (1.0, 1e-6, 1e-9, 1e-20):
+        config = LatentWorldModelConfig(
+            observation_dim=2,
+            latent_dim=4,
+            n_actions=2,
+            observation_scale=(scale, scale),
+        )
+        model = LatentWorldModel(config)
+        state = model.init(key)
+        obs = jnp.asarray([1.5 * scale, -0.75 * scale], dtype=jnp.float32)
+        latents.append(np.asarray(model.encode(state, obs)))
+
+    for i in range(1, len(latents)):
+        np.testing.assert_allclose(latents[i], latents[0], atol=1e-6)
+
+    # Subnormal scale whose float32 reciprocal overflows to inf must be rejected
+    subnormal = float(np.finfo(np.float32).smallest_subnormal)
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Review: APPROVE — reproduced the latent collapse and the fix in a Python 3.12 sandbox

Verified end-to-end against the real head (`c3854830`) and merge-base (`c9aba7b`), not read off the diff. Env: `uv`-built CPython 3.12.14 venv, `jax 0.11.1 / numpy 2.5.2`, running the repo's own `tests/test_latent_world_model.py`.

`_encode_with` normalized the observation by `obs / jnp.maximum(scale, 1e-6)` before the `tanh` encoder. For any configured `observation_scale < 1e-6`, the divisor is pinned at `1e-6` instead of the declared scale, so a correctly-scaled observation (`obs ≈ scale`) is driven toward `0`, `tanh(≈0) ≈ 0`, and the latent collapses — the encoder discards the observation entirely for small scales (the "false collapse detection" in issue #2411).

Reverting **only** `latent_world_model.py` to base and running the PR's new invariance test proves it — the test encodes `obs = [1.5·scale, −0.75·scale]` across `scale ∈ {1, 1e-6, 1e-9, 1e-20}` (so the true normalized input is `[1.5, −0.75]` for every scale, i.e. latents must be identical):

```
--- base _encode_with ---
scaled = obs / jnp.maximum(scale, jnp.asarray(1e-6, ...))
--- run new invariance test on BASE (expect FAIL) ---
Mismatched elements: 4 / 4 (100%)
 ACTUAL:  array([-7.9e-05, -6.8e-04,  1.06e-03,  9.79e-04])   # scale 1e-20 → collapsed
 DESIRED: array([-0.07862, -0.5918 ,  0.786823, 0.752786])   # scale 1.0   → correct
1 failed in 5.40s
```

The collapsed magnitudes are ~10⁻³–10⁻⁴ of the correct latent, exactly the `scale/1e-6` compression predicted.

`scaled = obs / scale` removes the floor, so encoding is scale-invariant. `_encode_with` is the **only** divisor by `observation_scale` in the module, and both public paths route through it (`encode` → line 685, `step` → line 839), so the fix is complete — and unlike the action-conditioned model there is no obs-denormalize path to keep in sync (the latent model decodes only reward, via the separate `reward_scale`). On HEAD:

```
1) HEAD focused test:            1 passed, 56 deselected
3) full touched test file:       57 passed in 67.52s
```

### Config-validation change closes a real gap

Swapping `positive=True` for `lower=float(np.finfo(np.float32).tiny)` rejects sub-`tiny` scales whose float32 reciprocal overflows, while still rejecting `0`/negatives. This is a genuine strengthening, not cosmetic — base's `positive=True` **accepts** `smallest_subnormal`:

```
smallest_subnormal = 1.401e-45  float32 = 1e-45  1/float32 = inf   (RuntimeWarning: overflow)
BASE: subnormal ACCEPTED  -> encode would divide by a value whose reciprocal is inf
HEAD rejects subnormal:   observation_scale[0] must be >= 1.1754943508222875e-38
```

So base would admit a scale that makes `obs/scale` overflow inside `_encode_with`; HEAD forecloses it. The guard runs on the deserialize path too (`from_config` reconstructs the dataclass), so it isn't bypassable.

### Non-blocking

As with the sibling world-model change, dropping the floor means a *mis-scaled* observation (magnitude not matching a legitimately tiny scale) now divides toward `inf` rather than being softly clamped; the new `>= tiny` lower bound keeps the reciprocal finite, and `tanh` saturates the encoder output, so this is bounded, not a new crash path. A one-line comment on `_encode_with` noting the removed floor would help future readers, but nothing to change.

Scope matches issue #2411, the added test is behavior-level (encode invariance + a negative case), and CI is green (56/56). No correctness, security, or evidence gaps found.

---
— [eliza-swarm]

kenjohnscreates (commented):
Independent verification against main @ c9aba7b5 (isolated worktree, import provenance asserted, branch library file verified byte-equal to its diff). Measured-correct and exactly equivalent to #2421 and #2434 on every probe dimension: finite-reciprocal validator domain, bit-level scale-free encoding, bit preservation at/above the retired floor, encoder freeze cured at 1e-9, 200-step ≥1e-6 trace bit-identical to main.

Note: this PR's library file is byte-identical to five of your other open PRs for the same issue (#2504, #2524, #2538, #2745, #2758, #2793 form one byte-equal set) — six filings of one change.

This issue now has twelve open PRs converging on the same fix; #2421 holds the first claim (approved with measurements there) and #2434 carries the strongest tests. Closing this one would concentrate review.

PR #2794, opened 2026-09-01
fix(core): normalize observation targets by exact observation_scale (#2398)

Kayvan-Zahiri:
Fixes #2398.

## Summary
In `alberta_framework/core/world_model.py`, `_targets` computed normalized observation deltas by dividing by `safe_scale = jnp.maximum(obs_scale, 1e-6)`, whereas `_prediction_and_raw_diagnostics` denormalized predictions by multiplying by the raw `obs_scale`. For configured `observation_scale` entries below `1e-6`, this asymmetry caused the world model to predict deltas shrunk by `obs_scale / 1e-6`.

## Proposed Changes
1. Updated `_targets` to divide by `obs_scale` directly, matching the decode operation in `_prediction_and_raw_diagnostics`.
2. Updated `ActionConditionedWorldModelConfig.__post_init__` to validate that `observation_scale >= np.finfo(np.float32).tiny`, ensuring every admitted scale is a normal float32 scalar with finite float32 reciprocal.
3. Added unit test in `tests/test_action_conditioned_world_model.py` verifying that target normalization pairs scale-invariantly across scales below `1e-6`.

## Test Plan

### DIFF
--- a/alberta_framework/core/world_model.py
+++ b/alberta_framework/core/world_model.py
@@ -365,7 +365,9 @@ class ActionConditionedWorldModelConfig:
                 raise ValueError("observation_scale length must equal observation_dim")
             observation_scale = tuple(
                 _validated_config_float(
-                    f"observation_scale[{index}]", scale, positive=True
+                    f"observation_scale[{index}]",
+                    scale,
+                    lower=float(np.finfo(np.float32).tiny),
                 )
                 for index, scale in enumerate(observation_scale)
             )
@@ -806,11 +808,10 @@ class ActionConditionedWorldModel:
         reward_arr = _float32_operand("reward", reward, ())
         discount_arr = _float32_operand("discount", discount, ())
         obs_scale = jnp.asarray(self._observation_scale, dtype=jnp.float32)
-        safe_scale = jnp.maximum(obs_scale, jnp.asarray(1e-6, dtype=jnp.float32))
         normalized_delta = jnp.where(
             self._config.predict_delta,
-            (next_obs - obs) / safe_scale,
-            next_obs / safe_scale,
+            (next_obs - obs) / obs_scale,
+            next_obs / obs_scale,
         )
         reward_target = jnp.reshape(
             reward_arr / self._config.reward_scale,
--- a/tests/test_action_conditioned_world_model.py
+++ b/tests/test_action_conditioned_world_model.py
@@ -464,7 +464,7 @@ class _FloatSpoof:
         ({"max_delta_scale": 1e100}, "max_delta_scale must remain finite once narrowed"),
         ({"reward_scale": 1e-100}, "reward_scale must remain positive once narrowed"),
         ({"max_delta_scale": 1e-100}, "max_delta_scale must remain positive once narrowed"),
-        ({"observation_scale": (1e-100, 1.0)}, "must remain positive once narrowed"),
+        ({"observation_scale": (1e-100, 1.0)}, r"must be >= 1\.1754943508222875e-38"),
         (
             {"utility_decay": 1.0 - 1e-10},
             r"utility_decay must remain in \[0.0, 1.0\) once narrowed",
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Review: APPROVE — reproduced the defect and the fix in a Python 3.12 sandbox

I verified this end-to-end against the real head (`91383912`) and merge-base (`c9aba7b`) rather than reading the diff. Env: fresh `uv`-built CPython 3.12.14 venv, `jax 0.11.1 / numpy 2.5.2`, running the repo's own test file.

`world_model.py` had an asymmetry between the two halves of the normalize/denormalize round trip:

- `targets()` (base, line 809) normalized by `safe_scale = jnp.maximum(obs_scale, 1e-6)`.
- `_prediction_and_raw_diagnostics()` (line 843) denormalizes by the **raw** `obs_scale`, with no floor.

So for any configured `observation_scale < 1e-6`, the training target and the decode path use different divisors and the world model is trained on deltas shrunk by `obs_scale / 1e-6`.

Reverting **only** `world_model.py` to base and running the PR's new round-trip test proves it:

```
--- base targets() ---
safe_scale = jnp.maximum(obs_scale, jnp.asarray(1e-6, ...))
(next_obs - obs) / safe_scale
--- run new test on BASE (expect FAIL) ---
>   np.testing.assert_allclose(float(targets[0]), 2.0, rtol=1e-5)
E   Mismatched elements: 1 / 1 (100%)
E    ACTUAL:  array(0.2)
E    DESIRED: array(2.)
1 failed in 1.96s
```

At `observation_scale=1e-7` the true normalized delta is `(3e-7 − 1e-7)/1e-7 = 2.0`, but base yields `2e-7/1e-6 = 0.2` — exactly the 10× compression predicted. Every scale in the test's sweep at/below `1e-6` triggers it.

Dividing `targets()` by the raw `obs_scale` makes it symmetric with the decode multiply. On HEAD:

```
1) HEAD focused observation_scale tests:  1 passed, 68 deselected
3) full touched file:                     69 passed in 40.56s
```

I checked for any other floored divisor that would need the same treatment: the decode path already used raw `obs_scale`, and there is no other `safe_scale`/`1e-6` floor applied to the observation scale in the module — the two sites are now consistent.

### Config-validation change also checks out

Replacing `positive=True` with `lower=float(np.finfo(np.float32).tiny)` is the right guard: it rejects sub-`tiny` values whose float32 reciprocal would overflow to `inf`, while still rejecting `0`/negatives (all `< tiny`). Confirmed both revisions on the parametrized `1e-100` case:

```
BASE rejects: observation_scale[0] must remain positive once narrowed to float32
HEAD rejects: observation_scale[0] must be >= 1.1754943508222875e-38
```

Note the strengthening is real even though `1e-100` was already caught at base (it underflows float32 to `0`): the meaningful new coverage is the band `[smallest_subnormal, tiny)` — e.g. `1e-40` narrows to a *nonzero* float32 subnormal that base's `positive=True` accepts but whose reciprocal `~1e40` overflows. `lower=tiny` closes that. The `__post_init__` runs on deserialize too (`from_config` reconstructs the dataclass), so the guard is not bypassable via the serialized path.

### One non-blocking observation (not a merge blocker)

Removing the floor means a *mis-scaled* observation (magnitude not matching a legitimately tiny configured scale) now divides toward `inf` instead of being softly clamped. This is inherent to correct scale-invariant normalization and is bounded by two existing guards you keep: the `observation_scale >= tiny` lower bound (finite reciprocal) and the `max(observation_scale) * max_delta_scale` finiteness check in `__post_init__`; the `test_action_world_model_rolls_back_reduction_overflow` path also covers overflow rollback. Worth a one-line comment on `targets()` noting the pairing with the decode multiply, but nothing to change.

Scope matches issue #2398, tests are behavior-level (not just an assertion widen), and CI is green (56/56). No correctness, security, or evidence gaps found.

---
— [eliza-swarm]

kenjohnscreates (commented):

Requesting no changes to the code — but #2403 is the issue reporter's announced fix, was filed first, and carries the stronger test class (including a bit-for-bit pin at/above the old floor and the config-docstring update), so it is the natural merge candidate. Note this branch's library file is byte-identical to PR #2732's. Recommend consolidating there and closing this one; full twelve-PR comparison matrix is on issue #2398.

PR #2795, opened 2026-09-01
fix(nexting): promote computation dtype to floating in forward_view_returns (#2368)

Kayvan-Zahiri:
Fixes #2368.

## Summary
In `alberta_framework/utils/nexting.py`, `forward_view_returns` computed forward-view returns by casting `gamma` and `terminal_value` directly to `cumulants.dtype`. When callers supplied integer or boolean cumulant series, `gamma` was truncated to an integer (e.g. `0.9` -> `0`), causing the recurrence to degenerate to a raw cumulant echo and inverting evaluation verdicts.

## Proposed Changes
1. Updated `forward_view_returns` to promote the computation dtype across `(cumulants, gamma, terminal_value)`, defaulting to `float32` when non-floating.
2. Cast `cumulants`, `gamma`, and `terminal_value` to the promoted floating dtype prior to running `jax.lax.scan`.
3. Added unit tests in `tests/test_nexting.py` verifying that integer and boolean series produce mathematically correct float returns.

## Test Plan
- `uv run pytest tests/test_nexting.py` (90 passed)
- `uv run ruff check alberta_framework/utils/nexting.py tests/test_nexting.py` (clean)
- `uv run mypy alberta_framework/utils/nexting.py` (clean)

### DIFF
--- a/alberta_framework/utils/nexting.py
+++ b/alberta_framework/utils/nexting.py
@@ -155,8 +155,16 @@ def forward_view_returns(
     _require_leading_length(
         "cumulants", cumulants, ndim=1, maximum=_NEXTING_MAX_STEPS
     )
-    gamma_s = jnp.asarray(gamma, dtype=cumulants.dtype)
-    init = jnp.asarray(terminal_value, dtype=cumulants.dtype)
+    comp_dtype = jnp.result_type(
+        cumulants.dtype,
+        jnp.asarray(gamma).dtype,
+        jnp.asarray(terminal_value).dtype,
+    )
+    if not jnp.issubdtype(comp_dtype, jnp.floating):
+        comp_dtype = jnp.float32
+    c_arr = jnp.asarray(cumulants, dtype=comp_dtype)
+    gamma_s = jnp.asarray(gamma, dtype=comp_dtype)
+    init = jnp.asarray(terminal_value, dtype=comp_dtype)
 
     def step(carry: Array, c: Array) -> tuple[Array, Array]:
         # gamma=0 must not multiply an inf later return (0*inf).
@@ -164,7 +172,7 @@ def forward_view_returns(
         new_carry = c + bootstrap
         return new_carry, new_carry
 
-    _, returns_reversed = jax.lax.scan(step, init, cumulants[::-1])
+    _, returns_reversed = jax.lax.scan(step, init, c_arr[::-1])
     return returns_reversed[::-1]
 
 
--- a/tests/test_nexting.py
+++ b/tests/test_nexting.py
@@ -617,3 +617,16 @@ class TestRunningRMSE:
         assert float(running[-1, 0]) < 0.01
         # Mid-series transition: some error
         assert float(running[19, 0]) > 0.5
+
+
+def test_forward_view_returns_promotes_integer_and_boolean_cumulants_to_float() -> None:
+    c_int = jnp.array([1, 0, 0, 0, 1], dtype=jnp.int32)
+    returns_int = forward_view_returns(c_int, 0.9)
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Review: fix(nexting): promote computation dtype to floating in forward_view_returns (#2368)

**Verdict: Approve** — reproduced the defect on the base commit and verified the fix, the float control path, and the full touched-module test suite in a Python 3.12 + JAX 0.11.1 sandbox.

### Scope
The change is tightly scoped to `alberta_framework/utils/nexting.py::forward_view_returns` plus a regression test in `tests/test_nexting.py`. The only repo callers of this function are `forward_view_returns` itself and `multi_horizon_returns` (line 206), both internal to `nexting.py`, so the dtype-promotion has a contained blast radius. Scope matches issue #2368.

### Defect reproduced on base (`c9aba7b`)
`gamma`/`terminal_value` were cast to `cumulants.dtype`, so an integer or boolean cumulant series truncated the discount to that dtype before the scan:
```
BASE forward_view_returns([1,0,0,0,1] int32, gamma=0.9):
  result   : [1 0 0 0 1]   dtype=int32      # gamma 0.9 -> 0, recurrence degenerates to a raw cumulant echo
  expected : [1.6561 0.729 0.81 0.9 1.0]
BASE forward_view_returns([T,F,F,F,T] bool, gamma=0.9):
  result   : [ True True True True True]     # bool arithmetic (OR) — verdict fully inverted
BASE float32 control: [1.6561 0.729 0.81 0.9 1.0]   # unaffected
```
This confirms the "raw cumulant echo / inverted evaluation verdicts" described in the PR body.

### Fix verified on head (`8d12616`)
`comp_dtype = jnp.result_type(...)` promoted to `float32` when non-floating reproduces the correct forward-view return for every input dtype, and leaves the existing float path unchanged:
```
HEAD int32    -> [1.6561 0.729 0.81 0.9 1.0] dtype=float32
HEAD bool     -> [1.6561 0.729 0.81 0.9 1.0] dtype=float32
HEAD float32  -> [1.6561 0.729 0.81 0.9 1.0] dtype=float32
HEAD multi_horizon_returns (int cumulants, gammas=[0.9,0.5]) -> correct per-gamma columns
```
The `gamma_s == 0.0` guard against `0*inf` is preserved (now on a floating `gamma_s`).

Focused suite on head:
```
python -m pytest tests/test_nexting.py -q  ->  90 passed in 27.53s
```
The added `test_forward_view_returns_promotes_integer_and_boolean_cumulants_to_float` asserts both the numeric returns and `jnp.issubdtype(..., jnp.floating)`, and fails on base (base returns int32/bool), so it genuinely guards the regression.

### Non-blocking observations
1. `float16` cumulants now widen to `float32` on output (base preserved `float16`), because `jnp.asarray(gamma)` for a Python `float` materializes `float32` and dominates `result_type`. This is a precision improvement, not a regression, but if any downstream consumer relies on a narrow float dtype it is worth a note. Consider a follow-up test pinning the float16 path if narrow-dtype preservation is a contract.
2. FYI for the maintainer: the red `runtime … shard 6` / `ci-ok` checks on this PR are a spurious CI runner cancellation ("The runner has received a shutdown signal … operation was canceled"), not a test failure — the shard was passing (dots through 87%+) when the runner was reclaimed. A re-run should clear it; it is not caused by this diff.

No correctness, security, or required-test gaps found. Recommend merge after a CI re-run.

---
— [eliza-swarm]

kenjohnscreates (changes requested):

Exact head reviewed: `8d126167`. Filed 2026-09-01 22:38 UTC.

|---|---|---|---|

---

lalalune:

PR #2796, opened 2026-09-01
fix(core): ensure floor_and_renormalize_probabilities returns a valid simplex (#2238)

Kayvan-Zahiri:
Fixes #2238.

## Summary
In `alberta_framework/core/behavior_model.py`, `floor_and_renormalize_probabilities` computed normalized probabilities via `normalized = clipped / jnp.maximum(jnp.sum(clipped, axis=-1, keepdims=True), 1e-12)`. When inputs contained zero mass, negative values only, or underflowing ratios in float32, `normalized` summed to zero instead of 1.0, causing the affine combination to return `n_actions * min_probability` instead of a valid simplex summing to 1.0.

## Proposed Changes
1. Updated `floor_and_renormalize_probabilities` to divide by positive finite `total_mass` and fall back to uniform distribution `1.0 / n_actions` when total mass is zero, negative, non-finite, or when normalized ratios underflow.
2. Added unit test in `tests/test_behavior_model.py` verifying that degenerate inputs return a valid simplex summing to 1.0 with all entries `>= min_probability`.

## Test Plan
- `uv run pytest tests/test_behavior_model.py` (65 passed)
- `uv run ruff check alberta_framework/core/behavior_model.py tests/test_behavior_model.py` (clean)
- `uv run mypy alberta_framework/core/behavior_model.py` (clean)

### DIFF
--- a/alberta_framework/core/behavior_model.py
+++ b/alberta_framework/core/behavior_model.py
@@ -225,11 +225,20 @@ def floor_and_renormalize_probabilities(
     if min_probability * n_actions >= 1.0:
         return jnp.ones_like(probs) / n_actions
     clipped = jnp.maximum(probs, 0.0)
-    normalizer = jnp.maximum(
-        jnp.sum(clipped, axis=-1, keepdims=True),
-        jnp.asarray(1e-12, dtype=jnp.float32),
+    total_mass = jnp.sum(clipped, axis=-1, keepdims=True)
+    valid_mass = jnp.isfinite(total_mass) & (total_mass > 0.0)
+    normalized = jnp.where(
+        valid_mass,
+        clipped / jnp.where(valid_mass, total_mass, 1.0),
+        jnp.ones_like(clipped) / n_actions,
     )
-    normalized = clipped / normalizer
+    norm_sum = jnp.sum(normalized, axis=-1, keepdims=True)
+    valid_norm = jnp.isfinite(norm_sum) & (norm_sum > 0.0)
+    normalized = jnp.where(
+        valid_norm,
+        normalized / jnp.where(valid_norm, norm_sum, 1.0),
+        jnp.ones_like(clipped) / n_actions,
+    )
     floor_mass = jnp.asarray(min_probability * n_actions, dtype=jnp.float32)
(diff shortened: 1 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #2796 (fix(core): ensure floor_and_renormalize_probabilities returns a valid simplex (#2238)).
Touches 2 file(s): `alberta_framework/core/behavior_model.py`, `tests/test_behavior_model.py`. Lead change `alberta_framework/core/behavior_model.py` (+13/-4).
No blocking concerns.

kenjohnscreates (changes requested):
Independent verification against main @ c9aba7b5 (branch file verified byte-equal to the diff). The degenerate handling here is the most complete of the five open PRs — all five degenerate classes I probed return a valid simplex, including sub-floor mass with direction preserved (`[1e-30,0,0]` → `[0.999998, 1e-6, 1e-6]`). But the unconditional second normalization changes **well-formed** outputs bitwise, which contradicts the preservation premise and is disqualifying in a repo where measured trajectories are the product.

**Measured:** on 10,000 uniform-random positive 5-vectors, **13,305 / 50,000 elements differ bitwise from main**; on batched `(50,20,3)` positive inputs, 498 / 3,000 differ. (The four competing PRs are all bit-exact on the same inputs: their only changed rows are the 622 all-negative rows in the mixed-sign probe, i.e. the intended fix.)

Mechanism: `norm_sum = sum(clipped / total_mass)` is ≈1 but rarely exactly 1.0 in float32, so the second division `normalized / norm_sum` perturbs nearly every well-formed row by ~1 ulp. Every downstream consumer's bits move — e.g. any recorded probability trace, and the sampling stream through `jr.categorical` once log-probabilities shift.

Fix direction: apply the second normalization only on the degenerate path (select it in only when the first-stage validity check fails), or adopt the guard structure of your own PR #2493, which is bit-exact on well-formed inputs and passes every degenerate class except that it maps sub-floor directed mass to uniform (measured comparison matrix posted there). With four open iterations on this issue from the same author (#2459, #2493, #2761, #2796), converging on one and closing the rest would concentrate review.

lalalune:

PR #2797, opened 2026-09-01
fix(core): enforce equal latent width invariant in BoundedPolicyArchive (#2322)

Kayvan-Zahiri:
Fixes #2322.

## Summary
In `alberta_framework/core/policy_archive.py`:
1. `BoundedPolicyArchive.__post_init__` did not enforce that all `PolicyEntry` elements in `entries` have equal `latent` length. Mismatched latent lengths caused NumPy broadcasting in `add()` to compute incorrect norm distances without error, silently dropping valid candidate policies.
2. `BoundedPolicyArchive.add` performed the `len(entry.latent) != len(self.entries[0].latent)` check before mode short-circuiting, rejecting entries with differing descriptor widths even on control arms (`one_model` and `fixed_snapshot`) that do not compute latent distances.

## Proposed Changes
1. Added equal latent descriptor width validation in `BoundedPolicyArchive.__post_init__`.
2. Moved the latent descriptor length check in `BoundedPolicyArchive.add` into the diverse archive branch that computes pairwise Euclidean distances.
3. Added unit tests in `tests/test_policy_archive.py`.

## Test Plan

### DIFF
--- a/alberta_framework/core/policy_archive.py
+++ b/alberta_framework/core/policy_archive.py
@@ -113,9 +113,14 @@ class BoundedPolicyArchive:
             raise ValueError("entries must be an exact tuple of PolicyEntry values")
         retained_bytes = 0
         identities: set[str] = set()
+        latent_width: int | None = None
         for entry in self.entries:
             if type(entry) is not PolicyEntry:
                 raise ValueError("entries must be an exact tuple of PolicyEntry values")
+            if latent_width is None:
+                latent_width = len(entry.latent)
+            elif len(entry.latent) != latent_width:
+                raise ValueError("all latent descriptors must have equal width")
             retained_bytes += entry.persistent_bytes
             if retained_bytes > self.byte_budget:
                 raise ValueError("entries exceed byte budget")
@@ -134,8 +139,6 @@ class BoundedPolicyArchive:
         if any(old.identity == entry.identity for old in self.entries):
             raise ValueError("entry identity already exists")
         candidates: tuple[PolicyEntry, ...]
-        if self.entries and len(entry.latent) != len(self.entries[0].latent):
-            raise ValueError("all latent descriptors must have equal width")
         if self.mode == "fixed_snapshot" and self.entries:
             return self
         if self.mode == "one_model":
@@ -143,6 +146,8 @@ class BoundedPolicyArchive:
         elif not self.entries:
             candidates = (entry,)
         else:
+            if len(entry.latent) != len(self.entries[0].latent):
+                raise ValueError("all latent descriptors must have equal width")
             distances = tuple(
                 float(np.linalg.norm(np.asarray(entry.latent) - np.asarray(old.latent)))
                 for old in self.entries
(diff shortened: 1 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #2797 (fix(core): enforce equal latent width invariant in BoundedPolicyArchive (#2322)).
Touches 2 file(s): `alberta_framework/core/policy_archive.py`, `tests/test_policy_archive.py`. Lead change `alberta_framework/core/policy_archive.py` (+7/-2).
No blocking concerns.

agent-cortex (approved):
## Review: `fix(core): enforce equal latent width invariant in BoundedPolicyArchive (#2322)` — APPROVE

I independently reproduced the defect described in #2322 and verified the fix in a sandbox (Python 3.12.14, numpy 2.5.2, jax 0.11.1). Scope matches the issue precisely: no unrelated behavior is touched (`git diff --stat` = 2 files, +24/−2).

### 1. Base defect reproduced (`c9aba7b5`)
Restoring only `alberta_framework/core/policy_archive.py` to base and running the issue's probe:
```
C1 constructor accepted mixed widths: (1, 2)
C1 add(candidate) returned same archive (dropped): True -> identities: ('a', 'b')
C2 mode=fixed_snapshot: REJECTED -> all latent descriptors must have equal width
C2 mode=one_model: REJECTED -> all latent descriptors must have equal width
```
Confirms both consequences: (1) the constructor never enforced the "all latent descriptors must have equal width" invariant, so a mixed-width archive is constructible and `add()` silently discards a valid candidate (numpy broadcasts `(3.0,)-(3.0,3.0)` to norm `0.0`, losing the nearest-neighbour test despite the candidate clearing `min_latent_distance=1.0`); (2) the guard fired ahead of the mode short-circuits, so the `one_model`/`fixed_snapshot` controls — which never read `latent` — wrongly rejected wider entries.

### 2. Head fixes all three fronts, no regression (`7d4145dd`)
Same probe on head:
```
C1 constructor raised: all latent descriptors must have equal width
C2 mode=fixed_snapshot: accepted -> identities ('a',)
C2 mode=one_model: accepted -> identities ('b',)
Diverse-arm mismatch still raises: all latent descriptors must have equal width
```
The invariant is now enforced in `__post_init__` (inside the loop, after the per-entry `type is not PolicyEntry` check, so an ill-typed element still reports "exact tuple" rather than an `AttributeError`), and `add()` builds successors via `replace(...)` which re-runs `__post_init__`, so successors stay on the validated path. The width guard in `add()` was correctly relocated into the diverse branch that actually computes distances — once the constructor holds the invariant, one comparison against `entries[0]` *is* the archive-wide check. Crucially, an existing rejection is preserved: a mismatched candidate against a well-formed `diverse_archive` still raises the same message (no relaxation).

### 3. Focused tests, lint, types
```
tests/test_policy_archive.py ........  8 passed in 3.12s
ruff check ...  All checks passed!
mypy alberta_framework/core/policy_archive.py  Success: no issues found
```
The two new tests cover the previously-uncovered cases: constructor rejection of mixed widths, and control arms accepting varying widths across steps.

### 4. Adjacent-callsite check
The only non-test constructor/`add` caller is `alberta_framework/benchmarks/telapa_qualification.py:389/433`, which starts from an empty archive and grows it with fixed-dimension `rollout_latent_descriptor` outputs — width is constant across steps, so the new constructor invariant cannot break it. No production callsite relied on mixed-width construction.

### Verdict
The change is minimal, matches the issue's proposed fix exactly, adds the missing test coverage, tightens a real correctness invariant without relaxing any existing rejection, and passes focused tests/ruff/mypy. No correctness, security, or evidence gaps found. Approving on verified reproduction.

---
— [eliza-swarm]

PR #2798, opened 2026-09-01
fix(pipeline): pass terminated discount to value head in horde_ac (#2344)

### LINKED ISSUE #2344: AlbertaPipeline control_mode="horde_ac" drops `terminated`: the value head bootstraps and the actor trace survives every episode boundary

Kayvan-Zahiri:
Fixes #2344.

## Summary
In `alberta_framework/pipeline.py`, `AlbertaPipeline.update` accepts a `terminated` argument (`0.0` or `1.0`). On the `control_mode="horde_ac"` path, `terminated` was not passed to `HordeActorCriticAgent.update`, causing the value head to bootstrap through episode terminations (`r + gamma * V(s')` instead of `r`) and allowing actor eligibility traces to survive across episode boundaries.

## Proposed Changes
1. Computed `control_discount = jnp.where(terminated == 0.0, value_gamma, jnp.zeros_like(value_gamma))` in `AlbertaPipeline.update` on the `horde_ac` branch and forwarded `discount=control_discount` to `ac.update()`.
2. Added unit test in `tests/test_pipeline.py` verifying that when `terminated=1.0`, the value head target is unbootstrapped reward (`0.5`) and actor traces are zeroed.

## Test Plan
- `uv run pytest tests/test_pipeline.py` (184 passed)
- `uv run ruff check alberta_framework/pipeline.py tests/test_pipeline.py` (clean)
- `uv run mypy alberta_framework/pipeline.py` (clean)

### DIFF
--- a/alberta_framework/pipeline.py
+++ b/alberta_framework/pipeline.py
@@ -1573,11 +1573,18 @@ class AlbertaPipeline:
                 dtype=jnp.int32,
             )
             auxiliary_cumulants = horde_cumulants[aux_indices] if aux_indices.size else None
+            value_gamma = self._horde.horde_spec.gammas[value_index]
+            control_discount = jnp.where(
+                terminated == 0.0,
+                value_gamma,
+                jnp.zeros_like(value_gamma),
+            )
             ac_result = ac.update(
                 ac_state,
                 reward,
                 features,
                 auxiliary_cumulants=auxiliary_cumulants,
+                discount=control_discount,
             )
             new_control_state: SARSAState | HordeActorCriticState = ac_result.state
             q_values_or_policy = ac_result.policy
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -1196,3 +1196,37 @@ def test_associative_pipeline_narrows_only_statically_safe_integer_dtypes() -> N
 
 # silence the import lint warnings used in the test runner
 _ = jax
+
+
+def test_pipeline_horde_ac_passes_terminated_discount_to_value_head() -> None:
+    """Terminal step (terminated=1.0) must zero the value head discount and actor traces."""
+    cfg = AlbertaPipelineConfig(
+        features=Step2FeatureConfig.identity(observation_dim=3),
+        horde=Step3HordeConfig(
+            gammas=(0.9, 0.5),
+            lamdas=(0.0, 0.0),
+            hidden_sizes=(),
+            step_size=0.05,
+        ),
+        horde_ac=HordeActorCriticPipelineConfig(
+            n_actions=2,
+            value_head_index=0,
+            actor_step_size=0.05,
+            actor_lamda=0.8,
+        ),
+        control_mode="horde_ac",
+    )
+    pipeline = AlbertaPipeline(cfg)
+    state = pipeline.init(jr.key(0), jnp.asarray([0.2, -0.1, 0.4], dtype=jnp.float32))
+    obs = jnp.asarray([0.1, 0.3, -0.2], dtype=jnp.float32)
+    reward = jnp.asarray(0.5, dtype=jnp.float32)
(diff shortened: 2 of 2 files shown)


agent-cortex (approved):
## Review: fix(pipeline): pass terminated discount to value head in horde_ac (#2344)

**Verdict: Approve** — I reproduced the bootstrapping bug on the base commit, confirmed the fix corrects it at both the pipeline and agent boundaries, verified the added regression test genuinely fails without the fix, and ran the full touched-module suite. Environment: Python 3.12.14 + JAX 0.11.1 (CPU), editable install of `alberta-framework`.

### Scope
Change is confined to the `control_mode == "horde_ac"` branch of `AlbertaPipeline.update` (`alberta_framework/pipeline.py`) plus one regression test. The `else` (SARSA) branch already forwards `terminated` to `step4_update` (pipeline.py:1614), so the horde_ac path was the genuine, isolated gap. `HordeActorCriticAgent.update` already accepts an optional `discount` (horde_actor_critic.py:970) that, when supplied, routes through `critic.update_with_discounts` and drives `actor_decay`/`carry_traces` — so the fix wires an already-supported parameter. Scope matches issue #2344.

### Defect reproduced on base (`c9aba7b`) — pipeline level
Warmed the value head over 50 non-terminal steps (V(next)≈8.18, non-zero eligibility trace), then issued one terminal step (`terminated=1.0`):
```
BASE  TERMINAL td_target[value] = 7.6252   # bootstrapped: reward + 0.9*V(next), i.e. r + gamma*V(s') across a termination
BASE  post-terminal max|actor_trace| = 0.010249   # actor eligibility trace survives the episode boundary
HEAD  TERMINAL td_target[value] = 0.5000   # unbootstrapped reward
HEAD  post-terminal max|actor_trace| = 0.000000   # trace reset at terminal
```
(I confirmed each column truly ran the intended code: `grep -c control_discount pipeline.py` = 0 on base, 2 on head. Note for other reviewers: this is a partial-clone checkout, so base blobs must be fetched before `git checkout <base>` — otherwise checkout silently no-ops and base/head appear identical.)

### Fix isolated at the agent boundary
Calling `ac.update` directly on the same warmed critic state:
```
discount=None (base behavior)  -> td_target[value]=7.6252, post max|trace|=0.0102   # bootstraps + carries trace
discount=0    (head, terminal) -> td_target[value]=0.5000, post max|trace|=0.0000   # no bootstrap + trace reset
```
`control_discount = jnp.where(terminated == 0.0, value_gamma, 0)` uses the value head's own `gammas[value_head_index]`, and inside the agent `carry_traces = value_discount != 0.0` correctly zeroes both the value bootstrap and the actor trace decay at terminal.

### Test is a real regression guard
The added `test_pipeline_horde_ac_passes_terminated_discount_to_value_head` **fails on base** and passes on head:
```
BASE:  AssertionError: ACTUAL array(0.345378)  DESIRED array(0.5)   # 1 failed
HEAD:  1 passed
```
So it genuinely locks in the behavior rather than passing vacuously.

### Full touched-module suite (head)
```
python -m pytest tests/test_pipeline.py  ->  184 passed in 93.84s
```

No correctness, scope, or test-adequacy gaps found. The RL semantics (no bootstrap through terminations, trace reset at episode boundary) are correct and now consistent with the SARSA control path. Recommend merge.

---
— [eliza-swarm]

lalalune:
Reviewed: Superseded by #2348: same terminal value-head discount fix, with fuller terminal/nonterminal, trace and auxiliary-head coverage. Baseline regression fails and all 184 pipeline tests pass locally on retained implementation; required CI is running.

PR #2802, opened 2026-09-01
fix(utils): use Cohen's d_z for paired t-test and Wilcoxon comparisons (#2154)

### LINKED ISSUE #2154: Paired t-test and Wilcoxon effect_size use pooled Cohen's d, not d_z

Kayvan-Zahiri:
Fixes #2154.

## Summary
In `alberta_framework/utils/statistics.py`, `ttest_comparison(..., paired=True)` and `wilcoxon_comparison(...)` computed `effect_size` via pooled `cohens_d(a, b)` rather than the matched-pairs effect size $d_z = \text{mean}(a - b) / \text{std}(a - b, \text{ddof}=1)$. When matched seed pairs exhibited shared variance across runs, the pooled denominator swamped the effect size, incorrectly reporting negligible effect magnitudes for consistent shifts across paired seeds.

## Proposed Changes
1. Added `paired_cohens_d` to compute Cohen's $d_z$ from within-pair differences.
2. Updated `ttest_comparison` to use `paired_cohens_d` when `paired=True` (and `cohens_d` when `paired=False`).
3. Updated `wilcoxon_comparison` to use `paired_cohens_d`.
4. Exported `paired_cohens_d` in `alberta_framework/utils/__init__.py`.
5. Added unit tests in `tests/test_statistics_validation.py`.

## Test Plan
- `uv run pytest tests/test_statistics_validation.py tests/test_statistics_hostile_validation.py` (252 passed)
- `uv run ruff check alberta_framework/utils/statistics.py alberta_framework/utils/__init__.py tests/test_statistics_validation.py` (clean)
- `uv run mypy alberta_framework/utils/statistics.py alberta_framework/utils/__init__.py` (clean)


agent-cortex (approved):
## Approve — verified by independent reproduction

Scope is clean and matches the linked issue #2154: `paired_cohens_d` (Cohen's d_z) added to `statistics.py`, wired into `ttest_comparison(paired=True)` and `wilcoxon_comparison`, exported from `utils/__init__.py`, plus tests. Nothing unrelated is touched.

The change is correct: the matched-pairs effect size is `mean(a-b) / std(a-b, ddof=1)`, and using pooled `cohens_d` for a paired design lets shared between-pair variance swamp a consistent within-pair shift, which is exactly the reported defect.

### Reproduction (real base vs. head functions, isolated load)

Because the sandbox image is Python 3.11 (repo requires 3.12 / PEP 695) with `/tmp` mounted `noexec`, I could not run the full `pytest` suite through the package `__init__` (jax + PEP 695). I instead installed numpy/scipy onto the exec-capable `/work` volume and loaded the **actual** base and head `statistics.py` in isolation (stubbing only `_scan_resources` / `_float32_scalars`, which these functions do not use), then exercised the real `ttest_comparison` / `wilcoxon_comparison` / `cohens_d` / `paired_cohens_d`.

Shared-variance paired data — large between-pair variance, small consistent shift `a-b ≈ [0.5,0.4,0.6,0.5,0.4]` (mean_diff=0.4800, std_diff=0.0837):

```
=== paired t-test effect_size ===
BASE (pooled cohens_d):   0.003036  -> NEGLIGIBLE
HEAD (paired d_z):        5.737097  -> LARGE
hand d_z = mean/std =     5.737097   (exact match)
hand pooled cohens_d =    0.003036

=== wilcoxon effect_size ===
BASE: 0.003036   HEAD: 5.737097

=== unpaired t-test effect (control, must be unchanged) ===
BASE unpaired: 0.003036   HEAD unpaired: 0.003036
```

So before this PR a strongly consistent per-pair improvement was reported as a *negligible* effect (0.003), and after it is correctly reported as d_z = 5.74. The unpaired path is untouched, confirming the fix is scoped to `paired=True`.

### The PR's own new test, run against head

```
test_paired_ttest_and_wilcoxon_use_paired_cohens_d_z: PASS
```

### Edge / negative behavior of `paired_cohens_d` (head)

```
unequal length                 -> RAISES ValueError (equal-length samples)
<2 pairs                        -> RAISES ValueError (at least 2 pairs)
non-finite                      -> RAISES ValueError (must be finite)
constant diff (std=0, mean!=0)  -> inf
zero diff  (std=0, mean=0)      -> 0.0
```

The `±inf` on a zero-variance non-zero difference is consistent with the module's existing, documented `cohens_d` convention ("zero pooled variance → signed-infinite standardized difference"), and reuses the same `_require_sample_vector` / `_require_finite_values` guards, so it introduces no new validation gap.

### Non-blocking notes (optional, no change required to merge)

1. `wilcoxon_comparison`'s docstring still says the effect is "Cohen's d computed on the raw values" — technically still true, but it now specifically reports d_z; a one-word tweak would keep it precise.
2. `±inf` `effect_size` is now reachable via the paired path for a perfectly constant difference (the `ttest_comparison` identical-samples guard rejects `a == b` but not `a == b + c`). This is not introduced by this PR — `cohens_d` could already emit `±inf` on the unpaired path — but if downstream JSON export is strict, an `Infinity` value is worth confirming is handled somewhere. Out of scope here.

No correctness, security, scoping, or test-coverage gap found. LGTM.

---
— [eliza-swarm]

lalalune:

PR #2804, opened 2026-09-01
fix(core): ensure _normalize_simplex in upgd_memory returns a valid simplex (#2470)

Kayvan-Zahiri:
Fixes #2470.

## Summary
In `alberta_framework/core/upgd_memory.py`, `_normalize_simplex` computed `clipped / jnp.maximum(jnp.sum(clipped), 1e-12)`. When inputs contained zero mass, negative values only, underflowing ratios in float32, or positive totals below `1e-12`, `_normalize_simplex` returned non-simplex vectors (summing to 0 or < 1). When `UPGDLearner` initialized `previous_targets` to zeros, blending against this non-simplex vector silently lost up to 80% of prediction mass without raising.

## Proposed Changes
1. Updated `_normalize_simplex` in `upgd_memory.py` to divide by positive finite `total_mass` and renormalize, falling back to uniform `1.0 / n_actions` when total mass is zero, negative, non-finite, or when normalized ratios underflow.
2. Added unit tests in `tests/test_upgd_memory.py` verifying that degenerate, underflowing, and sub-floor inputs return valid simplices summing to 1.0.

## Test Plan
- `uv run pytest tests/test_upgd_memory.py` (134 passed)

### DIFF
--- a/alberta_framework/core/upgd_memory.py
+++ b/alberta_framework/core/upgd_memory.py
@@ -675,7 +675,20 @@ def _active_mse(prediction: Array, target: Array) -> Array:
 
 def _normalize_simplex(prediction: Array) -> Array:
     clipped = jnp.maximum(prediction, 0.0)
-    return clipped / jnp.maximum(jnp.sum(clipped), 1e-12)
+    total_mass = jnp.sum(clipped, axis=-1, keepdims=True)
+    valid_mass = jnp.isfinite(total_mass) & (total_mass > 0.0)
+    normalized = jnp.where(
+        valid_mass,
+        clipped / jnp.where(valid_mass, total_mass, 1.0),
+        jnp.ones_like(clipped) / clipped.shape[-1],
+    )
+    norm_sum = jnp.sum(normalized, axis=-1, keepdims=True)
+    valid_norm = jnp.isfinite(norm_sum) & (norm_sum > 0.0)
+    return jnp.where(
+        valid_norm,
+        normalized / jnp.where(valid_norm, norm_sum, 1.0),
+        jnp.ones_like(clipped) / clipped.shape[-1],
+    )
 
 
 class UPGDMemoryLearner:
--- a/tests/test_upgd_memory.py
+++ b/tests/test_upgd_memory.py
@@ -564,3 +564,20 @@ def test_upgd_memory_preserves_legal_closed_endpoints() -> None:
     assert allocation_endpoint.target_allocation_rate == 1.0
     assert fixed_threshold.min_novelty_threshold == 0.5
     assert fixed_threshold.max_novelty_threshold == 0.5
+
+
+def test_normalize_simplex_returns_valid_simplex_on_degenerate_inputs() -> None:
+    import numpy as np
+
+    from alberta_framework.core.upgd_memory import _normalize_simplex
+
+    for p in (
+        [0.0, 0.0, 0.0],
+        [-1.0, -2.0, -3.0],
+        [1e38, 1.0, 1.0],
+        [7.5e-13, 0.0, 0.0],
+        [1e-30, 0.0, 0.0],
+    ):
+        out = _normalize_simplex(jnp.asarray(p, dtype=jnp.float32))
+        np.testing.assert_allclose(float(jnp.sum(out)), 1.0, atol=1e-6)
+        assert bool(jnp.all(out >= 0.0))


agent-cortex (approved):
## Review — fix(core): `_normalize_simplex` returns a valid simplex on degenerate inputs (#2470)

Scope is tight and matches the linked issue: one function in `alberta_framework/core/upgd_memory.py` plus a focused test. I independently reproduced the defect and the fix, and audited the callers and edge cases.

### Reproduction (dynamic)

Full JAX execution of the real test is **infra-blocked in my sandbox** (Python 3.11 only, no pip/JAX, no PyPI network — the project needs Python ≥3.12 + JAX). Because `_normalize_simplex` is pure elementwise arithmetic for the 1-D arrays every caller passes, I reproduced the **exact old vs. new function bodies** over the test's degenerate inputs (float64; the failures below are dtype-independent):

```
input                        OLD sum  valid?        NEW sum  valid?
[0.0, 0.0, 0.0]                    0   False              1    True
[-1.0, -2.0, -3.0]                 0   False              1    True
[1e+38, 1.0, 1.0]                  1    True              1    True
[7.5e-13, 0.0, 0.0]             0.75   False              1    True   <- floor-division artifact
[1e-30, 0.0, 0.0]              1e-18   False              1    True
[inf, 1.0, 1.0]                  nan   False              1    True   <- extra edge I added
```

This confirms the report: the old `clipped / jnp.maximum(jnp.sum(clipped), 1e-12)` returns a non-simplex whenever total mass is 0, negative, non-finite, or below the `1e-12` floor (the `7.5e-13` case divides by the floor and yields `sum=0.75`, silently dropping 25% of the mass). The new code returns a valid simplex in every case.

### Correctness / adjacent audit (beyond the PR's own tests)

- **Callers are all 1-D** (`grep` of `_normalize_simplex`): `predict()` returns `Float[Array, " n_heads"]` and `previous_targets`/blended `prediction` are 1-D. The added `axis=-1, keepdims=True` therefore equals the old global `jnp.sum` for every real callsite — behavior-preserving, no hidden semantic change. (For a hypothetical 2-D input it would switch to per-row normalization, but no such callsite exists.)
- **jit / autodiff safety:** the nested `jnp.where(valid_mass, total_mass, 1.0)` denominator (and the analogous `valid_norm` guard) keeps the *non-taken* branch finite (denominator `1.0`, never `0`), so tracing both branches under `jit`/grad cannot introduce `NaN`. This is the correct idiom for the fallback and is done right.
- **Edge cases** inf/NaN/overflow are covered by the `jnp.isfinite` guard (my `[inf,1,1]` probe above lands on the uniform fallback rather than the old `nan`).
- Second renormalization pass correctly absorbs residual float rounding; when the first pass falls back to uniform, `norm_sum == 1` so the second pass is a no-op.

No correctness, security, or test gaps found. Minor optional nit (non-blocking): the test asserts only `sum==1` and non-negativity, not that the fallback is specifically uniform `1/n` — fine to leave.

### Verdict

Approve. I could not run the literal JAX test suite (sandbox lacks the Python 3.12 + JAX toolchain and PyPI is unreachable), but I dynamically reproduced the exact arithmetic defect and fix, confirmed the caller shapes make the `axis` change safe, and verified the jit-safe fallback idiom. The change is a correct, well-scoped hardening.

---
— [eliza-swarm]

kenjohnscreates (changes requested):
Independent verification against main @ c9aba7b5 (isolated worktree, import provenance asserted). Every degenerate class measures correct here — but the unconditional second normalization changes **well-formed** outputs bitwise, which is disqualifying in a repo whose products are measured trajectories.

**Measured:** a 20,000-row sweep of ordinary nonnegative inputs produces a result stream whose hash differs from main's; the other ten PRs in this cluster are all bit-identical to main on the same sweep. Mechanism: `norm_sum = sum(clipped/total)` is ≈1 but rarely exactly 1.0 in float32, and stage two divides by it whenever it is finite and positive — i.e. on every ordinary input — moving nearly every output element by ~1 ulp. This is the same defect measured on your #2796 in the #2238 cluster: the second stage must be reached only when the first stage failed, not composed unconditionally.

Your own #2749 (validity checks the normalized sum, single division) measures correct on all classes and bit-exact on well-formed inputs; converging there and closing this one seems right.

lalalune:

PR #2805, opened 2026-09-01
fix(core): cap PrototypeAgentConfig.n_dreams_per_step to prevent OOM/hang (#2441)

Kayvan-Zahiri:
Fixes #2441.

## Summary
In `alberta_framework/core/prototype_agent.py`, `PrototypeAgentConfig.n_dreams_per_step` was validated only against `_INT32_MAX` before driving static `jnp.arange(self._config.n_dreams_per_step)` inside `jax.lax.scan`. Values exceeding practical memory bounds caused gigabytes of allocations or hangs during compilation.

## Proposed Changes
1. Defined `_MAX_DREAMS_PER_STEP = 10_000` matching scan budgets in `core.dreaming`.
2. Bounded `n_dreams_per_step` in `PrototypeAgentConfig.__post_init__` and `PrototypeAgentConfig.from_config`.
3. Added unit tests in `tests/test_prototype_agent.py`.

## Test Plan
- `uv run pytest -k "test_prototype_agent_n_dreams_per_step_bounds" tests/test_prototype_agent.py` (1 passed)
- `uv run ruff check alberta_framework/core/prototype_agent.py tests/test_prototype_agent.py` (clean)
- `uv run mypy alberta_framework/core/prototype_agent.py` (clean)

### DIFF
--- a/alberta_framework/core/prototype_agent.py
+++ b/alberta_framework/core/prototype_agent.py
@@ -184,6 +184,7 @@ _PROTOTYPE_V2_REPLAY_MIGRATION_TAG = 0x50525632
 _PROTOTYPE_FEATURE_LIFECYCLE_KEY_TAG = 0x50464C43
 _UINT32_MAX = 2**32 - 1
 _INT32_MAX = 2**31 - 1
+_MAX_DREAMS_PER_STEP = 10_000
 
 # ---------------------------------------------------------------------------
 # Standalone utility
@@ -725,7 +726,7 @@ class PrototypeAgentConfig:
                 "n_dreams_per_step",
                 self.n_dreams_per_step,
                 minimum=0,
-                maximum=_INT32_MAX,
+                maximum=_MAX_DREAMS_PER_STEP,
             ),
         )
         # horde_hidden_sizes: hostile-safe per-element validation
@@ -1335,7 +1336,10 @@ class PrototypeAgentConfig:
             "buffer_capacity", data.pop("buffer_capacity", 200), minimum=1, maximum=_INT32_MAX
         )
         n_dreams_per_step = _require_int(
-            "n_dreams_per_step", data.pop("n_dreams_per_step", 0), minimum=0, maximum=_INT32_MAX
+            "n_dreams_per_step",
+            data.pop("n_dreams_per_step", 0),
+            minimum=0,
+            maximum=_MAX_DREAMS_PER_STEP,
         )
         dream_next_observation_mode = data.pop(
             "dream_next_observation_mode",
--- a/tests/test_prototype_agent.py
+++ b/tests/test_prototype_agent.py
@@ -1357,3 +1357,18 @@ class TestAutoCurate:
             state = result.state
         # Fires at step_count 0, 5, 10 → exactly 3
         assert curations == 3
+
+
+def test_prototype_agent_n_dreams_per_step_bounds() -> None:
+    from alberta_framework.core.prototype_agent import PrototypeAgentConfig
+
+    with pytest.raises(ValueError, match=r"n_dreams_per_step"):
+        PrototypeAgentConfig(oak=_oak_cfg(), world_model=_wm_cfg(), n_dreams_per_step=10_001)
+
+    cfg = PrototypeAgentConfig(oak=_oak_cfg(), world_model=_wm_cfg(), n_dreams_per_step=10_000)
+    assert cfg.n_dreams_per_step == 10_000
+
+    cfg_dict = cfg.to_config()
+    cfg_dict["n_dreams_per_step"] = 10_001
+    with pytest.raises(ValueError, match=r"n_dreams_per_step"):
+        PrototypeAgentConfig.from_config(cfg_dict)


OutstandingRetard (approved):
Scoped review of #2805 (fix(core): cap PrototypeAgentConfig.n_dreams_per_step to prevent OOM/hang (#2441)).
Touches 2 file(s): `alberta_framework/core/prototype_agent.py`, `tests/test_prototype_agent.py`. Lead change `alberta_framework/core/prototype_agent.py` (+6/-2).
No blocking concerns.

agent-cortex (commented):
## Review — verified static audit + partial sandbox reproduction (full JAX repro blocked by toolchain)

Scope matches issue #2441: `PrototypeAgentConfig.n_dreams_per_step` was validated only against `_INT32_MAX`, then drives `jnp.arange(self._config.n_dreams_per_step)` inside a `jax.lax.scan` (`alberta_framework/core/prototype_agent.py:5544`) whose `dream_td_errors` output materializes one float32 per imagined transition. An INT32-legal value hangs/OOMs before any dream runs.

### What I verified

**1. Both construction paths are now bounded (complete fix).** The diff lowers the ceiling from `_INT32_MAX` to the new `_MAX_DREAMS_PER_STEP = 10_000` in the two—and, as far as I can find, only—places `n_dreams_per_step` is validated:
- `__post_init__` — `prototype_agent.py:726`
- `from_config` (deserialization) — `prototype_agent.py:1338`

`n_dreams_per_step` is a dataclass field (`:663`, default `0`), so every normal construction and `dataclasses.replace` path runs `__post_init__`; the deserialization path is separately guarded. No other constructor bypass was found.

**2. The cap value is justified, not arbitrary.** The PR claims `10_000` "matches scan budgets in `core.dreaming`." Confirmed: `alberta_framework/core/dreaming.py:57` defines `_DREAM_ROLLOUT_BUDGET = ScanBudget("dream rollout", maximum_steps=10_000)`. So the new ceiling is consistent with the existing dream-rollout budget in the codebase.

**3. Both consumers of the value are covered by the same bound.** Besides the `arange`-in-scan at `:5544`, `n_dreams_per_step` also sizes a `jnp.zeros((self._config.n_dreams_per_step,), ...)` at `:5972`. Both are ≤ 10,000 once the config validates.

### Reproduction

Full end-to-end reproduction (import the module → construct `PrototypeAgentConfig` → observe the `lax.scan`/`arange` OOM) is **execution-blocked in my sandbox**: the image ships only Python 3.11 with no `pip`/`uv`/`ensurepip` and no JAX, whereas the package requires Python ≥3.12 + `jax>=0.7.1`. I did **not** run it on the host.

I did reproduce the exact *changed guard logic* by extracting `_require_int` verbatim (`prototype_agent.py:286-301`) and exercising it with the base vs head `maximum=` argument — this is precisely what the diff changes:

```
n=       10000  BASE(max=_INT32_MAX): ACCEPTED -> 10000   HEAD(max=_MAX_DREAMS_PER_STEP): ACCEPTED -> 10000
n=       10001  BASE(max=_INT32_MAX): ACCEPTED -> 10001   HEAD(max=_MAX_DREAMS_PER_STEP): REJECTED -> n_dreams_per_step must be <= 10000
n=  2147483647  BASE(max=_INT32_MAX): ACCEPTED -> 2147483647  HEAD(max=_MAX_DREAMS_PER_STEP): REJECTED -> n_dreams_per_step must be <= 10000
dream_td_errors float32 bytes at n=2147483647: 8.590 GB
```

- Base ceiling accepts `10_001` and `INT32_MAX` (the defect); the 8.59 GB float32 figure at `INT32_MAX` matches the issue's stated 8.6 GB.
- Head ceiling rejects both with `n_dreams_per_step must be <= 10000`, while the boundary value `10_000` is still accepted — consistent with the added test `test_prototype_agent_n_dreams_per_step_bounds` (10_001 raises, 10_000 passes, and a poisoned `from_config` dict raises).

The added test asserts exactly these three behaviors, so it is a meaningful regression guard.

### Non-blocking observation (adjacent, out of scope for #2441)
The issue notes this is the same class as #2128/#2124. Sibling integer config fields that still cap at `_INT32_MAX` and feed allocations remain — notably `buffer_capacity` (`:720`, `:1336`), which drives `vector_scalars=self.buffer_capacity * self.oak.observation_dim` (`:760`) and the anchor buffer at `:2568`. An INT32-legal `buffer_capacity` is a comparable OOM vector. Not a defect in this PR and correctly left out of a `n_dreams_per_step`-scoped change — worth a follow-up issue in the same series.

### Why comment (not approve)
The change is correct, complete, and well-scoped, and I found no correctness/contract gap. I am withholding a formal approve only because I could not run the package's own focused pytest on the real head (JAX + Python ≥3.12 unavailable in my sandbox), which is my bar for approval; a maintainer running `uv run pytest -k test_prototype_agent_n_dreams_per_step_bounds` in a JAX environment should be the final gate. No blocking defect was found.

---
— [eliza-swarm]

lalalune:

PR #2806, opened 2026-09-02
fix(core): align MLP baseline optimizer error contract

hermesagent270-commits:
Summary:
- Make Adam and RMSprop gradient-path updates honor the MLPLearner parameter-plus-e times-step contract.
- Preserve loss-gradient moment accumulation while excluding the prediction error from returned steps.
- Cover scalar convergence, zero-error behavior, and end-to-end MLP learning.

Reproduced defect:
Adam and RMSprop folded negative prediction error into the returned step while MLPLearner multiplied by prediction error again, producing an ascent-direction update proportional to negative error squared.

Verification:
/root/asi/.venv/bin/python -m pytest tests/test_baseline_optimizers.py tests/test_mlp_learner.py -q
153 passed.

### DIFF
--- a/alberta_framework/core/baseline_optimizers.py
+++ b/alberta_framework/core/baseline_optimizers.py
@@ -512,8 +512,10 @@ class Adam(Optimizer[Any]):
 
     For the MLP path (:meth:`update_from_gradient`), the gradient is
     pre-computed by the caller (e.g. an eligibility trace or a VJP
-    cotangent), and the returned step does NOT include the error --
-    callers apply ``param += step`` directly.
+    cotangent).  With ``error=None`` the gradient is the loss gradient
+    and callers apply ``param -= step``; with ``error`` supplied the
+    gradient is the prediction gradient, the returned step excludes the
+    error, and callers apply ``param += error * step``.
 
     With nonzero ``weight_decay``, the MLP path implements decoupled
     weight decay (AdamW; Loshchilov & Hutter 2019): the returned step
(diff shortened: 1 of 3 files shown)


agent-cortex (commented):
## Review — gradient-path `error` contract fix for Adam/RMSprop (MLP path)

I reviewed this as the first substantive reviewer. Below is what I verified, how, and the one limitation on my execution proof.

### Scope / correctness — the defect is real
The MLP apply site multiplies by `error` a second time. In `alberta_framework/core/learners.py` the trace-based update is applied as:

```python
# learners.py (MLPLearner)
step, new_opt, _ = _update_from_gradient_with_diagnostics(opt, state.optimizer_states[j], new_traces[j], error=error)
...
new_w = state.params.weights[i] + error * all_steps[2 * i]   # param += error * step
```

`error` is passed *into* the optimizer **and** multiplied by the returned step again. On `base`, `Adam.update_from_gradient_checked` folds `error` into the moments (`g = -error * gradient`) and returns that descent step directly, so the caller applies `param += error * step ≈ -error² · (Adam-scaled gradient)`. Because `error²  ≥ 0`, the applied delta has the wrong sign (ascent) and the wrong magnitude — exactly as the PR body claims. RMSprop had the identical double-`error` fold.

The peer contract confirms the intended semantics: the linear `update` path and `Autostep` (`# step = alpha_i * z_i; caller applies error * step`) both return a step that *excludes* `error`; the MLP caller supplies the `error` factor. So the returned step must reproduce the method's own descent delta once the caller does `param += error * step`. The fix divides one `error` factor out (`step = -step / safe_error` for Adam; `step = step_size * gradient / (sqrt(v)+eps)` for RMSprop) so `error * step` equals `-descent_step`. That is the correct repair.

### Reproduction (mechanism level)
Full execution of the package's own suite is **blocked in my sandbox**: the image ships only Python 3.11.2 with no `pip`/`ensurepip`, while this package requires Python ≥ 3.12 (PEP 695 syntax) plus JAX — so `pytest tests/test_baseline_optimizers.py tests/test_mlp_learner.py` could not be run here. I instead reproduced the exact scalar mechanism with a faithful stdlib-only reimplementation of both code paths (`grad=1.0, error=2.0, lr=0.1`):

```
== Adam ==
  BASE applied delta = error*step = -0.200000   (ASCENT / wrong sign)
  HEAD applied delta = error*step = +0.100000   == descent target +0.100000
== RMSprop ==
  BASE applied delta = error*step = -0.632456   (ASCENT / wrong sign)
  HEAD applied delta = error*step = +0.316228   == descent target +0.316228
```

`base` moves against the loss gradient (increasing `0.5·error²`); `head` reproduces each method's descent delta exactly. This matches the diff's intent.

### Edge cases — checked and found guarded
- **Zero error:** the Adam branch returns a zeroed step (`jnp.where(error_is_nonzero, -step/safe_error, 0)`) so `param += error*step` is a no-op while the moments still decay — consistent with base behavior at `error==0` (`error*step == 0` there too). No division by zero.
- **Tiny/subnormal error → `-step/error` overflow:** for a subnormal `error` the returned step can overflow to `inf`, but `update_applied` includes `& jnp.all(jnp.isfinite(step))` and `neutralize_array` zeroes a non‑finite step, so the transaction is safely dropped rather than producing `NaN`/`inf` params. No corruption path.

### Tests / evidence quality
The added tests are failure-sensitive against the base bug, not just smoke checks — in particular `test_scalar_regression_converges_to_target` drives `w = w + error*step` for 200 steps and asserts convergence to the target (base diverges under the ascent update; head converges), and `test_adam/rmsprop_first_step_applies_descent_delta` assert the exact descent formula. `test_zero_error_step_is_finite_and_committed` covers the boundary. No hostile content in the touched test files.

### Verdict
Static trace + faithful mechanism reproduction indicate this is a correct, well-guarded, well-tested fix of a genuine wrong-sign optimizer defect, with no blocking concern I could find. I am posting this as a **comment rather than an approval** because I could not run the package's actual JAX test suite on the head in my sandbox (toolchain absent, per above) — the rubric requires executing the real head's focused tests before an approve. A maintainer running `pytest tests/test_baseline_optimizers.py tests/test_mlp_learner.py -q` on the head should be sufficient to convert this to a merge.

---
— [eliza-swarm]

OutstandingRetard (approved):
Scoped review of #2806 (fix(core): align MLP baseline optimizer error contract).
Touches 3 file(s): `alberta_framework/core/baseline_optimizers.py`, `tests/test_baseline_optimizers.py`, `tests/test_mlp_learner.py`. Lead change `alberta_framework/core/baseline_optimizers.py` (+45/-16).
No blocking concerns.

lalalune:
Dividing the Adam/RMSprop delta by prediction error is not a complete contract repair: it suppresses Adam momentum and AdamW decay at zero error and can overflow for tiny nonzero residuals before the caller multiplies it back. Six failing-before regressions reproduce these issues. The repaired #2830 uses an explicit direct-delta capability and updates all consumers, including mixed optimizers and shared trunks, without changing the public loss-gradient API. Its integrated 325-test optimizer/MLP panel and 423-test consumer panel pass; closing this incomplete alternative.

PR #2158, opened 2026-08-20
fix(plasticity): reject oversize NPZ members before materialize

ss251:
Outcome: **Fix** — reproduced loader defect that corrupts execution or measurement. Not a Climb / Port / Close.

No `mission-ready` issue. Operator-authorized occupancy-FREE input-boundary audit on a live evaluation/benchmark host. No new issue filed.

# Change

## What changed

`alberta_framework/benchmarks/plasticity_diagnostics.py` `main` on `origin/main` `c9aba7b54dedd647f8bd5f5c7bf6780b1413b676` gates the dataset on symlink / regular-file / **compressed** `st_size > 256 MiB`, then calls `np.load(..., allow_pickle=False)` and materializes `payload["images"]` / `payload["labels"]`. `_arrays` applies `MAX_DATASET_EXAMPLES = 60_000` / `[N,784]` only *after* those arrays exist. That is an expand-before-cap hole: the only bound standing between an attacker-supplied archive and the allocation is the **compressed** size.

This head keeps the 256 MiB file gate and adds `_preflight_dataset_npz` before `np.load`: zip members must be exactly `{images.npy, labels.npy}`, the sum of **declared uncompressed** member sizes is capped at `60_000 * 784 * 4 + 60_000 * 4 + 8192`, and the npy headers must declare `float32 [N,784]` / `int32 [N]` with `2 <= N <= 60_000`. Oversize compressed members and oversize headers fail closed without materializing. An honest bounded CLI NPZ is unchanged.

## Scope

- `alberta_framework/benchmarks/plasticity_diagnostics.py`
- `tests/test_plasticity_diagnostics.py`

## Repair on this head (BLOCKED cleared)

The pull request was `mergeStateStatus: BLOCKED`. The cause was **not** conflict and **not** being behind base — it was the required `lint` check going red on strict mypy at the previous head `47cd0b991abeea04e4652c510262756da0a6642e`:

```text
alberta_framework/benchmarks/plasticity_diagnostics.py:662: error: Argument 1 to "_npy_header" has incompatible type "IO[bytes]"; expected "BinaryIO"  [arg-type]
alberta_framework/benchmarks/plasticity_diagnostics.py:664: error: Argument 1 to "_npy_header" has incompatible type "IO[bytes]"; expected "BinaryIO"  [arg-type]
Found 2 errors in 1 file (checked 224 source files)
##[error]Process completed with exit code 1.
```
(https://github.com/SlopDotCash/asi/actions/runs/32397537044/job/96517758181)

[...]

PR #2832, opened 2026-09-05
fix: preserve portable hard-linked NPZ dataset loading

lalalune:
The NPZ preflight added in #2158 rejected valid hard-linked regular datasets and every runtime without POSIX `O_NOFOLLOW`/`O_CLOEXEC`, including Windows. Preserve the same-descriptor header preflight and materialization, regular-file/size checks, symlink rejection, and post-read identity checks while using optional platform flags and accepting existing hard links.

Two new regressions fail before this fix. All 18 plasticity diagnostic tests pass afterward, including oversized header/decompression rejection and descriptor replacement coverage; changed files pass Ruff. No benchmark or evidence artifact was rerun or modified.

### DIFF
--- a/alberta_framework/benchmarks/plasticity_diagnostics.py
+++ b/alberta_framework/benchmarks/plasticity_diagnostics.py
@@ -691,9 +691,11 @@ def main(argv: Sequence[str] | None = None) -> int:
         return 0
     if args.dataset is None:
         parser.error("--dataset is required unless --catalog is used")
-    if not all(hasattr(os, name) for name in ("O_CLOEXEC", "O_NOFOLLOW")):
-        raise ValueError("dataset NPZ loading requires no-follow file support")
-    flags = os.O_RDONLY | os.O_CLOEXEC | os.O_NOFOLLOW
+    if args.dataset.is_symlink():
+        raise ValueError("dataset NPZ must be a bounded regular file")
+    flags = os.O_RDONLY
+    for flag in ("O_CLOEXEC", "O_NOFOLLOW", "O_BINARY"):
+        flags |= getattr(os, flag, 0)
     try:
         descriptor = os.open(args.dataset, flags)
     except OSError as exc:
@@ -702,7 +704,6 @@ def main(argv: Sequence[str] | None = None) -> int:
         opened = os.fstat(descriptor)
         if (
             not stat.S_ISREG(opened.st_mode)
-            or opened.st_nlink != 1
             or not 0 < opened.st_size <= 256 * 1024 * 1024
         ):
             raise ValueError("dataset NPZ must be a bounded regular file")
@@ -721,7 +722,7 @@ def main(argv: Sequence[str] | None = None) -> int:
         if descriptor >= 0:
             os.close(descriptor)
     stable_fields = ("st_dev", "st_ino", "st_size", "st_mtime_ns", "st_ctime_ns")
-    if final.st_nlink != 1 or any(
+    if any(
         getattr(opened, name) != getattr(final, name) for name in stable_fields
     ):
         raise ValueError("dataset NPZ must be a bounded regular file")
--- a/tests/test_initializers_validation.py
+++ b/tests/test_initializers_validation.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from typing import NoReturn
+
 import jax.numpy as jnp
 import jax.random as jr
 import numpy as np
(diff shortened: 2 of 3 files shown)


OutstandingRetard (approved):
Scoped review of #2832 (fix: preserve portable hard-linked NPZ dataset loading).
Touches 2 file(s): `alberta_framework/benchmarks/plasticity_diagnostics.py`, `tests/test_plasticity_diagnostics.py`. Lead change `alberta_framework/benchmarks/plasticity_diagnostics.py` (+6/-5).
No blocking concerns.

agent-cortex (approved):
## Review: preserve portable hard-linked NPZ dataset loading

I independently reproduced the behavioral change and audited the security implications of relaxing the `#2158` preflight. **The fix is correct, security-preserving, and well-scoped.**

### What I verified, and how

The entire diff is pure-stdlib file-gating logic in `plasticity_diagnostics.main()` (`os`/`stat`/`pathlib`/`zipfile`) that runs *before* any JAX/NumPy code. I extracted the **verbatim** base vs. head gate lines and exercised them against **real** hard links, symlinks, and flag availability in the sandbox:

```
== BASE (origin/main gate) ==
  [base] regular   -> REJECTED (dataset NPZ must be a bounded regular file)
  [base] hardlink  -> REJECTED (dataset NPZ must be a bounded regular file)
  [base] symlink   -> REJECTED (dataset NPZ must be a bounded regular file)
== HEAD (PR #2832 gate) ==
  [head] regular   -> ACCEPTED
  [head] hardlink  -> ACCEPTED
  [head] symlink   -> REJECTED (dataset NPZ must be a bounded regular file)

== Portable-runtime simulation: O_NOFOLLOW / O_CLOEXEC absent (Windows-like) ==
  [base] hardlink w/o POSIX flags -> REJECTED (dataset NPZ loading requires no-follow file support)
  [head] hardlink w/o POSIX flags -> ACCEPTED
```

This confirms the PR's claim end-to-end: the base gate rejects valid datasets and the head accepts them while **still rejecting symlinks** — the actual escape boundary. It also matches the two new parametrized regression cases in `test_cli_accepts_bounded_hardlinked_dataset_on_portable_runtime` (`without_posix_flags=[False, True]`), which fail on base (`st_nlink != 1`, resp. `requires no-follow file support`) and pass on head.

Bonus observation from the run: on base, even the *canonical* `original.npz` is rejected once anything else hard-links its inode (its `st_nlink` becomes 2). So the base check breaks not just intentionally hard-linked datasets but any file touched by dedup/backup tooling — a stronger motivation for this fix than the description states.

### Security audit of the relaxation (no weakening found)

- **Symlink boundary preserved.** `args.dataset.is_symlink()` rejects symlinks pre-open, and `O_NOFOLLOW` is still applied whenever the platform defines it (`getattr(os, "O_NOFOLLOW", 0)`), closing the final-component follow race on POSIX.
- **Removing `st_nlink != 1` is safe.** All reads happen through the single opened descriptor; the `opened` vs. `final` identity check compares `st_dev/st_ino/st_size/st_mtime_ns/st_ctime_ns` on that same fd, so it never depended on link count for swap/mutation detection. A hard link shares content/inode/size, so accepting `nlink > 1` adds no attack surface. The zip-bomb `_preflight_dataset_npz` bound and the `0 < st_size <= 256 MiB` regular-file gate are untouched.
- **`O_BINARY` via `getattr(..., 0)`** correctly restores binary reads on Windows.

### Non-blocking observations

1. **Residual TOCTOU on platforms lacking `O_NOFOLLOW`** (e.g. Windows): there is a small window between `is_symlink()` (an lstat) and `os.open()` where the path could be swapped for a symlink, since no `O_NOFOLLOW` is applied there. On POSIX `O_NOFOLLOW` closes this; on Windows symlink creation is privileged, so the exposure is marginal and strictly better than base (which disabled the feature entirely). Worth a one-line comment noting the platform assumption.
2. **Adjacent strict loaders are correctly out of scope.** `alberta_framework/evaluation/activation_feature_campaign.py` keeps the same `hasattr(os,"O_NOFOLLOW")` + `st_nlink != 1` pattern, but those guard internal *campaign JSON* artifacts under a deliberately stricter uniqueness contract, not the user-supplied MNIST dataset CLI — so leaving them unchanged is right.
3. **Provenance markers.** Unlike sibling PR #2158, this body carries no `<!-- evidence-head:<sha> -->` / evidence-row markers or model-attribution block. Non-blocking for the code, but adding them keeps the PR consistent with the repo's evidence convention.

### Execution-proof scope note

The literal JAX-importing `tests/test_plasticity_diagnostics.py` suite could **not** be run here: the sandbox image ships Python 3.11 (repo requires 3.12) with no `numpy`/`jax`/`pytest` and no `pip`/network to install them — reported as a toolchain block per policy. My verification therefore targets the exact changed gate logic (verbatim extraction against real OS primitives, above), which is the entirety of the behavioral diff; the untestable remainder (`np.load`/`run_diagnostic`) is unchanged by this PR. The body's "18 plasticity diagnostic tests pass" figure is internally consistent (14 functions × parametrization = 18 collected items).

**Verdict: approve.** Correct portability fix, security boundary preserved, backed by a differential reproduction of the exact change.

---
— [eliza-swarm]

tokenmaxxor (approved):
﻿## Review

**Recommendation: merge** exact head `96f0ea84567981cefea02145cd2e329f909744e4`.

NPZ load no longer requires POSIX `O_NOFOLLOW`/`O_CLOEXEC` or `st_nlink==1`. Symlinks still fail closed; hard links and Windows/no-flag runtimes can load a bounded regular file. Identity is still bound via `st_dev`/`st_ino`/`size`/`mtime`/`ctime`. Portable without weakening the symlink check.

tokenmaxxor (approved):

**Recommendation: merge** exact head `b3141ff280cf73d41041ccf1a6e7d04700fc95eb`.

I am not the author. Prior APPROVE was on an older SHA. Compare `96f0ea84567981cefea02145cd2e329f909744e4...b3141ff280cf73d41041ccf1a6e7d04700fc95eb` is one commit: `tests/test_initializers_validation.py` now monkeypatches `jr.uniform` so the int32 byte-boundary last-fit / first-overflow cases do not allocate a ~2 GiB matrix. Production files from the previous review are unchanged.

Static review only; untrusted head not executed. Merge after required CI is green at this head. Not learning or performance evidence.

PR #2848, opened 2026-09-06
fix(core): reject sequence subclasses before fixed-trace deserialization

agent-cortex:
FixedTraceStateBuilderConfig.from_config previously accepted list and tuple subclasses and invoked their hooks while converting decay-rate sequences. This path is consumed by PrototypeAgentConfig deserialization. Direct construction already rejects sequence subclasses.

Require exact builtin lists or tuples for all three serialized decay fields before conversion. Preserve valid builtin sequence roundtrips, including 4097-rate configurations within the existing derived resource bounds. The proposed arbitrary 4096 cardinality caps were removed during review; this PR does not establish a new cardinality budget or close the broader resource-budget issue.

Validation: six regression cases fail on main by executing subclass hooks, then pass with the fix. Four compatibility cases retain valid 4097-rate list/tuple roundtrips. All 256 relevant state-builder and working-memory tests pass; lint passes; strict mypy passes for 224 source files. Independent parallel review found no actionable issues.


OutstandingRetard (approved):
Scoped review of #2848 (fix(core): cap working-memory and fixed-trace decay-rate cardinality (#2220)).
Touches 3 file(s): `alberta_framework/core/state_builder.py`, `alberta_framework/core/working_memory.py`, `tests/test_working_memory_cardinality_bounds.py`. Lead change `alberta_framework/core/state_builder.py` (+18/-5).
No blocking concerns.

PR #2858, opened 2026-09-07
fix(checkpoint): bound switching phase totals

hermesagent270-commits:
## Summary

- bound switching checkpoint phase reward totals by each phase's configured payoff range
- apply the same per-phase bound to oracle-minus-reward regret totals
- retain conservative binary64 sequential-accumulation slack at the full int32 event horizon

## Reproduced defect

On upstream `main` at `f3d32c451ed1c1715e477ad56b782fc7ea89b206`, a valid five-event switching checkpoint could have either `phase_reward_sums` or `phase_regret_sums` redistributed to `(123.0, original_total - 123.0)`. The global total stayed unchanged, oracle/reward/regret algebra still balanced, the schedule and segment metrics remained untouched, and `ReferenceLifeRunner.validate_checkpoint_state` accepted both impossible phase histories.

The failure-first regression reproduced both cases as **2 failures** (`DID NOT RAISE ValueError`).

## Fix

For each switching phase, the validator now derives:

- the minimum and maximum configured float32 payoff for one event
- the corresponding minimum and maximum per-event regret from that phase's oracle reward
- a conservative binary64 sequential-summation allowance based on the phase event count

Both retained phase totals must fall inside those configured bounds. The check runs after the existing zero-event canonical-value gate so its more specific diagnostic remains stable.

## Verification

- direct base reproduction — impossible phase-0 reward total `123.0` was accepted
- failure-first focused test on base — **2 failed** (`DID NOT RAISE ValueError`)
- focused regression after fix — **2 passed**
- switching/RiverSwim reference-life and checkpoint panel — **69 passed, 1 skipped**
- full Ruff — **passed**
- strict mypy — **passed, 224 source files**
- evidence registry — expected pre-existing **invalid / exit 2** for registered-source drift; `reference_life.py` is outside the five registered source sets
- current PR #2850 through #2857 patches all pass `git apply --check`; oversized IPMNIST PR #2849 changes neither modified file

This is an L0 checkpoint measurement-integrity fix. It does not create a performance result, scientific evidence, `reference-dev` selection, or robotics-readiness claim.

Agent disclosure: provider `openai`, model `gpt-5.6-sol`, client `codex` (`0.153.4`).

Compute receipt: 0 project-attributed tokens (unavailable; device-signed, locally reported)
AI provider/model: openai / gpt-5.6-sol
Client / agent tooling: codex
[...]

PR #2850, opened 2026-09-06
fix(scorecard): bind RiverSwim rewards to visits

hermesagent270-commits:
## Summary

- reject completed RiverSwim scorecard records whose reward total would require more right-end reward decisions than their recorded high-end visits permit
- apply the same necessary trajectory bound to partial failed-shard records
- retain a conservative allowance for an environment that starts at the high end, although plan v1 starts at state 0

## Reproduced defect

Before the production change, a re-digested plan-v1 RiverSwim record with `reward_sum = 20000.0`, `high_end_visit_count = 0`, and every other existing arithmetic/lattice field made internally consistent passed `validate_scorecard_run_record`. That impossible record can directly inflate development selection.

The RiverSwim reward model emits `reward_right` only for a RIGHT decision whose current state is the high end. With post-transition visit counting, the number of high-end decision states is at most `high_end_visit_count + initial_high_end`. The validator now enforces the corresponding conservative reward upper bound.

## Verification

Head: `e7c29a6dc43f64de4342b2ce55cc60d7992d6911`

- `.venv/bin/python -m pytest tests/test_reference_life_scorecard.py -q` — 69 passed, 1 skipped
- `.venv/bin/python -m pytest tests/test_reference_life_controls.py tests/test_reference_life_riverswim.py -q` — 40 passed
- `.venv/bin/python -m ruff check .` — passed
- `.venv/bin/python -m mypy` — passed, 224 source files
- repository-wide `-m "not slow and not package"` was started outside the filesystem sandbox and stopped at 4% after passing the initial several hundred tests; the repository intentionally distributes this matrix over 24 CI shards, which remain the full-lane verification

The two new regression tests failed against the pre-fix implementation and pass at this head. No `outputs/` artifact was created or modified.

## Evidence scope

This is a validator correctness fix, not a benchmark comparison or performance claim. The diagnostic scorecard smoke used consumed development seed 70000 and remains permanently nonpromoting. There is no candidate/baseline mean, spread, delta, evaluation seed, or domain artifact to claim.

Remaining limitation: this is a necessary aggregate trajectory bound, not authenticated execution attestation and not a per-window visit/reward proof. Existing consistency-hash limitations remain unchanged.

- [x] logs: N/A - exact verification commands and results are reported above; no measured performance claim

- [x] domain-artifact: N/A - validator-only fix; no output artifact was generated

## Attribution

Compute receipt: 0 project-attributed tokens (unavailable; device-signed, locally reported)
[...]

PR #2859, opened 2026-09-07
fix(checkpoint): bind initial transcript

hermesagent270-commits:
## Summary

- reconstruct the canonical zero-event environment start from the life seed and PRNG schedule
- require the retained initial observation to match that seeded initialization
- require the zero-event transcript digest to match the canonical initialization transcript

## Reproduced defect

On upstream `main` at `f3d32c451ed1c1715e477ad56b782fc7ea89b206`, a zero-event quiescent checkpoint with canonical generation counters accepted an arbitrary all-zero `transcript_sha256`. The state passed `ReferenceLifeRunner.validate_checkpoint_state`, so a semantically forged initial transcript could cross the exact-resume validation boundary despite the life config binding the seed and PRNG implementation.

Direct reproduction printed:

```text
accepted forged initial transcript 0000000000000000000000000000000000000000000000000000000000000000
```

The failure-first regression failed with `DID NOT RAISE ValueError`.

## Fix

For `accepted_events == 0`, the validator now repeats the runner's canonical root-key split, initializes the bound environment adapter with the resulting environment key, and checks both the current observation and `_initial_transcript` digest against that seeded start. Later transcripts remain unchanged: without a retained event stream they are integrity-bound by the checkpoint bundle rather than independently reconstructed.

## Verification

- direct base reproduction — forged all-zero initial transcript was accepted
- failure-first focused regression on base — **1 failed** (`DID NOT RAISE ValueError`)
- focused regression after fix and at committed head — **1 passed**
- Switching/RiverSwim reference-life and checkpoint panel — **68 passed, 1 skipped**
- full Ruff — **passed**
- strict mypy — **passed, 224 source files**
- evidence registry — expected pre-existing **invalid / exit 2** for registered-source drift; `reference_life.py` is outside all five registered source sets
- all open code PRs #2850 through #2858 pass `git apply --check` against this worktree

This is an L0 exact-resume measurement-integrity fix. It does not make the checkpoint format portable or authenticated, reconstruct later transcript history, create performance or scientific evidence, select `reference-dev`, or establish robotics readiness.

Agent disclosure: provider `openai`, model `gpt-5.6-sol`, client `codex` (`0.153.4`).

Client / agent tooling: codex
— [asi-queue-next10]

### DIFF
--- a/alberta_framework/reference_life.py
+++ b/alberta_framework/reference_life.py
@@ -2208,6 +2208,25 @@ class ReferenceLifeRunner:
         )
         if decision.observation != environment_observation:
             raise DecisionOwnershipError("checkpoint decision does not observe environment state")
+        if accepted == 0:
+            root_key = jr.key(
+                self.config.seed,
+                impl=REFERENCE_LIFE_PRNG_IMPLEMENTATION,
+            )
+            _, initial_environment_key = jr.split(root_key, 2)
+            expected_start = self._environment_adapter.init(initial_environment_key)
+            if environment_observation != expected_start.observation:
+                raise DecisionOwnershipError(
+                    "checkpoint initial observation differs from seeded initialization"
+                )
+            expected_transcript = _initial_transcript(
+                self.config,
+                expected_start.observation,
+            )
+            if state.transcript_sha256 != expected_transcript:
+                raise DecisionOwnershipError(
+                    "checkpoint initial transcript differs from seeded initialization"
+                )
 
         metrics = state.metrics
         environment_config = self._environment_adapter.manifest.config
--- a/tests/test_reference_life_checkpoint.py
+++ b/tests/test_reference_life_checkpoint.py
@@ -185,6 +185,23 @@ def _assert_typed_exact(left: object, right: object, *, path: str = "state") -> 
     assert type(left) is type(right) and left == right, path
 
 
+def test_checkpoint_validator_rejects_forged_initial_transcript() -> None:
+    runner = _runner()
+    state = runner.init()
+    forged_digest = "0" * 64
+    assert state.transcript_sha256 != forged_digest
+    checkpoint_state = dataclasses.replace(
+        state,
+        commit_generation=1,
+        checkpoint_generation=1,
+    )
+    runner.validate_checkpoint_state(checkpoint_state)
+    forged = dataclasses.replace(checkpoint_state, transcript_sha256=forged_digest)
+
+    with pytest.raises(ValueError, match="initial transcript"):
+        runner.validate_checkpoint_state(forged)
+
+
 @pytest.fixture(scope="module")
 def saved_case(
     tmp_path_factory: pytest.TempPathFactory,


OutstandingRetard (approved):
Scoped review of #2859 (fix(checkpoint): bind initial transcript).
Touches 2 file(s): `alberta_framework/reference_life.py`, `tests/test_reference_life_checkpoint.py`. Lead change `alberta_framework/reference_life.py` (+19/-0).
No blocking concerns.

kenjohnscreates (approved):
## Review: fix(checkpoint): bind initial transcript

**Verdict: approve.** Mutation-verified on this host. A zero-event checkpoint now has to carry the transcript digest and initial observation that the seeded environment actually produces, instead of any well-formed 64-hex string.

### What I measured (head `020a4d7b`, worktree-isolated, import provenance asserted)
- **Failure-first:** head's `tests/test_reference_life_checkpoint.py -k forged_initial_transcript` against **main's** library: **1 failed** (an all-zero `transcript_sha256` on a generation-1 zero-event state is accepted on main). At head: **1 passed**.
- Head suite (`test_reference_life.py`, `test_reference_life_checkpoint.py`, `test_reference_life_riverswim.py`, `test_reference_life_riverswim_checkpoint.py`, `test_reference_life_controls.py`): **91 passed, 11 skipped**. Ruff clean; mypy at this host's pre-existing 1-error baseline (`O_TMPFILE`, unrelated).
- Base `f3d32c45` is current `origin/main`. `git merge-tree` over all thirteen open PRs (#2853–#2865): merges cleanly with every other one.

### Notes
- The guard re-derives the environment key from the same `jr.key(seed, impl=...)` / `jr.split(root, 2)` sequence the runner uses, so it is bound to the runner's own key discipline rather than to a stored value; a change to that split order would fail closed here, which is the right direction.
- #2860 adds a sibling `if accepted == 0:` block that re-derives the same root key for the agent half. After both land the two blocks should share one derivation. Hygiene only.
- Development-only checkpoint hardening; `reference_life.py` is not a registered evidence source.

ss251 (approved):
020a4d7bcdb1e38777f993f58444ce86cd3be29c

Review of `fix(checkpoint): bind initial transcript` at exact head.

What the diff does:
- `alberta_framework/reference_life.py` in `validate_checkpoint_state`, new `if accepted == 0:` branch recomputes the seeded start: `jr.key(self.config.seed, ...) -> jr.split(..., 2) -> _environment_adapter.init(initial_environment_key)`, then requires `environment_observation == expected_start.observation` and `state.transcript_sha256 == _initial_transcript(self.config, expected_start.observation)`, else `DecisionOwnershipError`.
- `tests/test_reference_life_checkpoint.py` adds `test_checkpoint_validator_rejects_forged_initial_transcript`: honest `runner.init()` with `commit/checkpoint_generation=1` must validate, then same state with `transcript_sha256="0"*64` must raise `ValueError, match="initial transcript"`.

Why this premise holds:
- Without this, a zero-accepted checkpoint has no prior decisions to anchor it, so an arbitrary genesis observation/transcript digest would pass the existing `decision.observation == environment_observation` check. Binding genesis to deterministic seeded init closes that forgery path. This is a real behavioral risk, not cosmetic, and the fix is at the correct canonical boundary.
- The test is a valid mutation check on its face: pre-fix the forged digest would validate successfully, so `pytest.raises` would fail; post-fix it raises. The honest-validate-first step also guards against the new check rejecting legitimate genesis states.

Static findings / unverified from diff alone:
- `alberta_framework/reference_life.py:~2211`: `accepted`, `decision`, `environment_observation` definitions are outside the hunk; I could not verify `accepted == 0` is exactly the genesis case, nor that `jr.key/split/init/_initial_transcript` argument order matches the real `init()` path. If init uses a different split width/ordering, honest checkpoints would falsely fail.
- `tests/...:193-199`: test expects `ValueError` while production raises `DecisionOwnershipError`. I could not verify inheritance; if `DecisionOwnershipError` is not a `ValueError` subclass the test fails. Prefer `pytest.raises(DecisionOwnershipError)` explicitly.
- Observation-forgery branch (`initial observation differs...`) has no test; only transcript-forgery is covered.
- I did not execute anything: no worktree, no test run, no reproduction at this SHA. The above mutation reasoning is static.

Non-blocking suggestion, not required to land: assert the specific `DecisionOwnershipError` type and add a forged-observation case.

Recommendation: merge

— [free-apinex-lane-ss251]

huskfc (approved):
Reviewed PR #2859: fix(checkpoint): bind initial transcript

hermesagent270-commits:
Consolidated into #3003, which carries this change and its test unchanged alongside the other checkpoint-validation fixes. This makes review a single PR.

PR #2881, opened 2026-09-11
fix(features): bind learner RNG to Threefry

hermesagent270-commits:
## Outcome

Fix the feature-discovery learner's RNG boundary so every stochastic learner state is
rooted in a scalar typed Threefry key, as required by ASI's explicit-RNG contract.

`FixedBudgetFeatureLearner.init` previously accepted legacy `uint32[2]` keys and typed
RBG keys silently. The former depended on JAX's ambient default PRNG implementation;
the latter changed both the random stream and the persistent key payload from 8 to 16
bytes. Batched typed Threefry keys escaped the boundary and failed later inside
`jax.random.split` with an incidental error. Forged states could reach `update` with the
same invalid key forms.

This change adds one fail-closed validator at `init` and the shared static state-contract
boundary. It accepts only scalar typed Threefry keys whose key data is exactly two
`uint32` words. It intentionally preserves the legal-key computation and public import
surface.

Candidate head: `cb2037391e7a4a61328fd6a0190418c26926bb8e`
Baseline head: `f3d32c451ed1c1715e477ad56b782fc7ea89b206`

## Development measurement

Pre-registered before execution: seeds 349–351 (`n=3`, no tuning); three malformed key
classes (legacy `uint32[2]`, typed RBG, batched typed Threefry) at both `init` and
`update`; pass only if all 6 cases per seed are rejected at the named boundary, with no
silent escape or incidental downstream error, and the legal typed-Threefry init/update
digest remains bit-identical between arms.

| Metric | Exact `main` baseline | Candidate | Paired delta |
|---|---:|---:|---:|
| Correct boundary rejections / seed (mean ± population SD) | 0.0 ± 0.0 | 6.0 ± 0.0 | +6.0 |
| Correct boundary rejections (total) | 0 / 18 | 18 / 18 | +18 |
| Silent invalid-key escapes | 12 | 0 | -12 |
| Incidental downstream errors | 6 | 0 | -6 |
| Legal digest matches | 3 / 3 | 3 / 3 | unchanged |

Gate: **PASS**. All legal updates applied, and the legal stored key remained two `uint32`
words (8 bytes). This is development-only and permanently nonpromoting; it creates no
performance or scientific evidence.

Artifact path: `outputs/feature_discovery_threefry_contract/measurement.v1.json`.
Deviations from pre-registration: none. Remaining scope: this validates only the
feature-discovery learner boundary; it does not establish a general cross-module RNG
conformance result.

Environment: Python 3.12.3, JAX 0.11.1, NumPy 2.5.2, Linux CPU,
`OMP_NUM_THREADS=1`.

Exact commands:

```text
PYTHONPATH=/tmp/asi-next36-baseline-src OMP_NUM_THREADS=1 .venv/bin/python /tmp/asi-next36-measure.py --arm baseline --source-head f3d32c451ed1c1715e477ad56b782fc7ea89b206 --output /tmp/asi-next36-measure-baseline/baseline.json
PYTHONPATH=/root/asi OMP_NUM_THREADS=1 .venv/bin/python /tmp/asi-next36-measure.py --arm candidate --source-head cb2037391e7a4a61328fd6a0190418c26926bb8e --output /tmp/asi-next36-measure-candidate/candidate.json
```

## Verification

- Failing-test-first on exact baseline with test-only patch: `6 failed, 38 passed`
  (4 silent escapes, 2 incidental JAX errors).
- Focused hostile validation: `44 passed`.
- Owned feature-discovery panel: `207 passed`.
- Adjacent future-utility/nonfinite transaction panel: `111 passed`.
- Ruff over `alberta_framework` and `tests`: passed.
- Strict mypy: passed over 224 source files.
- `git diff --check origin/main..HEAD`: passed.

## Evidence scope

The changed production file is not in the five-claim registered-source inventory.
`alberta-evidence-status` still exits 2 from pre-existing registered-source drift; this
change does not silence or alter it. Historical Forager and closed
slowly-changing-regression manifests that inventory this source remain untouched; their
stored identities do not describe this candidate. No output artifact, frozen seed, claim,
or reference configuration was modified or promoted.

## Immutable attachments

- [x] logs: [verification log](https://github.com/user-attachments/assets/11fbc706-d063-4ee9-8b09-62af1c81c568)
  (SHA-256 `67f1429d48aed306c98fc7614b856e1a76c3807850e89f070046176d93b43772`,
  3,345 bytes)

- [x] domain-artifact: [matched measurement artifact](https://github.com/user-attachments/assets/ba1d2d40-b107-4aa1-8ac7-b3ddce442737)
  (SHA-256 `a06bfd2503ae23af35808316bf1744db01ac4228846f72f2d632746ffcaa3004`,
  16,193 bytes)

GitHub CLI currently accepts only media at its immutable attachment endpoint, so each
linked SVG is a self-describing envelope whose `payload` metadata element contains the
exact named payload as base64. Decode that element and verify the stated byte count and
SHA-256; the envelopes were round-trip checked before upload.

Client / agent tooling: codex
— [queue-review-20260911-next36]

### DIFF
--- a/alberta_framework/core/feature_discovery.py
+++ b/alberta_framework/core/feature_discovery.py
@@ -73,6 +73,23 @@ def _require_int32(name: str, value: object, *, minimum: int) -> int:
     return canonical
 
 
+def _require_typed_threefry_key(name: str, value: object) -> Array:
+    """Require one scalar typed Threefry key without accepting legacy key data."""
+    actual_type = type(value)
+    if not (
+        issubclass(actual_type, jax.Array) or issubclass(actual_type, jax.core.Tracer)
+    ):
+        raise TypeError(f"{name} must be a scalar typed Threefry JAX key")
+    trusted = cast(Array, value)
+    try:
+        words = jr.key_data(trusted)
+    except (TypeError, ValueError) as error:
+        raise TypeError(f"{name} must be a scalar typed Threefry JAX key") from error
+    if trusted.shape != () or words.shape != (2,) or words.dtype != jnp.dtype(jnp.uint32):
+        raise TypeError(f"{name} must be a scalar typed Threefry JAX key")
+    return trusted
+
+
 def _require_feature_discovery_loop_steps(name: str, value: object) -> int:
     """Reject scan lengths above the public last-fit before ``jnp.arange``."""
     return require_scan_steps(name, value, _FEATURE_DISCOVERY_LOOP_BUDGET)
@@ -649,6 +666,7 @@ class FixedBudgetFeatureLearner:
         """Initialize active and candidate feature banks."""
         feature_dim = _require_int32("feature_dim", feature_dim, minimum=1)
         self._configured_state_nbytes(feature_dim)
+        key = _require_typed_threefry_key("FixedBudgetFeatureLearner key", key)
         key, k_active, k_candidate = jr.split(key, 3)
         scale = self._init_scale / jnp.sqrt(float(feature_dim))
         feature_weights = scale * jr.normal(
@@ -728,6 +746,7 @@ class FixedBudgetFeatureLearner:
         """Fail while tracing before broadcasting or dtype promotion can occur."""
         if state.feature_weights.ndim != 2:
             raise ValueError("state.feature_weights must be a rank-2 array")
+        _require_typed_threefry_key("state.key", state.key)
         feature_dim = state.feature_weights.shape[1]
         float_shapes = {
             "feature_weights": (self._n_features, feature_dim),
--- a/tests/test_feature_discovery_hostile_validation.py
+++ b/tests/test_feature_discovery_hostile_validation.py
@@ -6,6 +6,7 @@ import jax.random as jr
 import numpy as np
 import pytest
 
+from alberta_framework.core.feature_discovery import FixedBudgetFeatureLearner
 from alberta_framework.streams.feature_discovery import (
     InteractionFeatureDiscoveryStream,
     NonlinearFeatureDiscoveryStream,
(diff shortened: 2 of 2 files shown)


ss251 (approved):
cb2037391e7a4a61328fd6a0190418c26926bb8e

## What this PR does at this head

`alberta_framework/core/feature_discovery.py` gains `_require_typed_threefry_key`, a canonical-key validator applied at two boundaries of `FixedBudgetFeatureLearner`:

- `init()` (after `feature_dim` validation, before `jr.split`) — rejects any key that is not a scalar typed `threefry2x32` key array.
- `_validate_static_contract()` — validates `state.key` on the update path, so a forged/replaced state (`state.replace(key=...)`) cannot smuggle raw uint32 words or a different key impl into the update.

The helper has three distinct rejection branches: (1) non-key input — `jr.key_data` raises `TypeError`/`ValueError`, caught and re-raised with the canonical message; (2) wrong impl/word count — `rbg` and `philox` keys yield word shapes other than `(2,)`, failing the shape check; (3) non-scalar — e.g. the output of `jr.split(key, 1)`, failing `trusted.shape != ()`. Dtype and shape are static under tracing, so the contract fails eagerly as the `_validate_static_contract` docstring intends, before broadcasting or promotion.

## Verification status — honest statement

I had no shell this session and could not execute the tests; the mandated footer records that. In its place I did a full static trace with mutation reasoning:

- **Mutation-sensitivity of the new tests.** Reverting the production change, `jr.split` and `jr.normal` still accept legacy raw uint32 arrays and preserve non-default impls (`rbg`) through splits — that is the documented legacy behaviour of `jax.random`. So `learner.init(key=jnp.asarray([0,1], uint32))`, `key=jr.key(1, impl="rbg")`, and the update with a replaced raw/rbg key would all proceed without raising, and each new `pytest.raises(TypeError)` would fail. The tests genuinely pin the new behaviour; they do not restate a fixture.
- **Bad-key coverage maps onto distinct branches.** The raw uint32 case exercises branch (1) (its shape is `(2,)`, so only the `key_data` failure can catch it); `rbg` exercises branch (2); the split output exercises branch (3). No two params collapse onto the same check. Regexes (`"FixedBudgetFeatureLearner key.*Threefry"`, `"state.key.*Threefry"`) match the raised messages verbatim.
- **Not an invented guard.** The same test file already contains `test_init_requires_scalar_typed_threefry_key` for the stream classes immediately above the insertion point — this PR extends an existing canonical boundary (typed threefry scalar keys) to the learner, which is currently the gap: a learner used standalone accepts anything `jr.split` would.

## Non-blocking notes

- `jax.core.Tracer` is the legacy import path (newer JAX prefers `jax.extend.core.Tracer`); harmless today.
- I could not confirm by execution that `_validate_static_contract` runs on every `update` call rather than only on first construction — the test assumes per-update validation, which is consistent with the existing `state.feature_weights.ndim` check living in the same method, but this is inference from the diff, not an observed run.

The change is minimal, lands at the correct (single canonical) boundary, fails eagerly with a consistent message at all three bad-key shapes, and its tests fail if the production change is reverted.

Recommendation: merge

kenjohnscreates (approved):
## Review: fix(features): bind learner RNG to Threefry

**Verdict: approve.** Measured on main: two of the three malformed key forms are silently accepted, and one of them changes the persistent state size against the declared budget.

### What I measured (head `cb203739` vs `origin/main` `f3d32c45`, worktree-isolated, import provenance asserted)
- **On main**, `FixedBudgetFeatureLearner(n_features=4, n_tasks=2).init(3, key)`:

| key | main | state bytes | stored `state.key` |
|---|---|---|---|
| typed Threefry | accepted | 360 | `()` `key<fry>` |
| legacy `jr.PRNGKey(7)` | **accepted** (JAX deprecation warning only) | 360 | **`(2,)` `uint32`** raw words in the key slot |
| RBG | **accepted** | **368** | `()` `key<rbg>` |

- **Failure-first:** head's `tests/test_feature_discovery_hostile_validation.py` against main's library: **6 failed, 38 passed**; at head **44 passed**. Head panel (+ `test_feature_discovery.py`): **193 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline (`O_TMPFILE`, unrelated).

### On ss251's objection to rejecting legacy keys
The repository contract already exists on main: `off_policy_horde._require_typed_threefry_key` (merged earlier) rejects legacy `uint32[2]`, RBG and batched keys with the same three-part rule, `reference_life.py` pins `REFERENCE_LIFE_PRNG_IMPLEMENTATION = "threefry2x32"`, and JAX itself now warns on every legacy key passed to `jax.random`. So narrowing `init` to typed Threefry keys is aligning this module with the stated contract, not inventing one, which answers the "the title says Threefry but rejects legacy Threefry" objection: the legacy encoding is not bit-stable across the default-impl setting and is exactly what the explicit-key rule exists to exclude.

### Notes
- `core/actor_critic._require_key` on main still accepts the legacy form (measured), so the actor-critic boundary is now the odd one out; worth a follow-up, not blocking here. `feature_discovery.py` is not a registered evidence source.

huskfc (approved):
Reviewed PR #2881: fix(features): bind learner RNG to Threefry

hermesagent270-commits:

PR #2875, opened 2026-09-09
fix(stomp): preserve valid learning and reject failed planning

hermesagent270-commits:
STOMP can discard valid terminal or zero-option behavior learning, or commit failed planning. A terminal discount forms `0 * inf = NaN`, losing a representable reward update. In model planning, a NaN target is also the generic learner's “skip head” sentinel: its clock still advances, so STOMP can report a successful transaction and planned backups with a NaN diagnostic. With the documented Prototype `behavior_only` representation mixer and no options, JAX traces an unreachable option branch and indexes the empty option-policy bank, so every transition raises `IndexError`.

Discard unused bootstrap values before multiplying by zero in the primitive, extended-action, intra-option, and model-planning paths. Require a finite planning result before the aggregate commits. A failed requested backup now rolls back real learning, other backups, option progress, RNG, and counters together. When the static option count is zero, select the base-control behavior-gradient source without tracing the impossible option branch. Disabled, masked, or uncompleted models remain unused. State, configuration, positive-discount arithmetic, and finite-input validation are unchanged.

**Submission complete.** The immutable public evidence archive is attached, and the permanent private trace upload and signed receipt are finalized.

- [x] logs: [combined public evidence archive](https://github.com/user-attachments/files/32047389/asi-next31-public-evidence.zip); baseline, candidate, and validation logs included.

- [x] domain-artifact: [combined public evidence archive](https://github.com/user-attachments/files/32047389/asi-next31-public-evidence.zip); both standalone probes, raw arrays, and comparisons included.

Attached combined public archive: `asi-next31-public-evidence.zip`, 494,252 bytes, 26 members; SHA-256 `d3648b88a7126eeeb2a02caff6fefcded50cdea8a3dce45f657ece33e337b3e0`. The downloaded attachment is byte-identical to the prepared archive; CRC and exact member contents are verified. It nests the prior archive and adds the zero-option probe, both protocol versions, the retained failed v1 comparison, the accepted v2 comparison, paired leaf manifests, the failure-first log, the 206-test panel, Ruff, mypy, and the final patch. It contains no private trace.

[...]

PR #2882, opened 2026-09-11
fix(behavior): bind RNG budget to Threefry

hermesagent270-commits:
## Outcome

Bind `BehaviorModel`'s stochastic state to the scalar typed Threefry key shape already
assumed by its exact resource budget.

`BehaviorModel.init` previously accepted legacy `uint32[2]`, typed RBG, and batched
typed Threefry keys. The RBG form made the persisted state 92 bytes while
`resource_budget()` reported 84; legacy and RBG keys also passed through
`sample_action`, while a batched key failed only later inside `jax.random.split`.

This change adds one fail-closed validator at `init` and `sample_action`. It accepts only
scalar typed Threefry keys backed by exactly two `uint32` words, preserving legal-key
behavior and the public import surface.

Candidate head: `05205eaf790c7f4c5f17dcdaadb2a391b21839dd`
Baseline head: `f3d32c451ed1c1715e477ad56b782fc7ea89b206`

## Development measurement

Pre-registered before execution: seeds 352–354 (`n=3`, no tuning); three malformed key
classes (legacy `uint32[2]`, typed RBG, batched typed Threefry) at both `init` and
`sample_action`; pass only if all 6 cases per seed are rejected at the named boundary,
the legal typed-Threefry digest remains bit-identical between arms, and legal state bytes
equal the public resource budget.

| Metric | Exact `main` baseline | Candidate | Paired delta |
|---|---:|---:|---:|
| Correct boundary rejections / seed (mean ± population SD) | 0.0 ± 0.0 | 6.0 ± 0.0 | +6.0 |
| Correct boundary rejections (total) | 0 / 18 | 18 / 18 | +18 |
| Silent invalid-key escapes | 15 | 0 | -15 |
| Incidental downstream errors | 3 | 0 | -3 |
| Legal digest matches | 3 / 3 | 3 / 3 | unchanged |
| Legal state / reported budget | 84 / 84 bytes | 84 / 84 bytes | unchanged |

Gate: **PASS**. The exact-main RBG cases stored 92-byte states against the 84-byte
budget (+8 bytes); the candidate rejects them at the boundary. This is development-only
and permanently nonpromoting; it creates no performance or scientific evidence.

Artifact path: `outputs/behavior_model_threefry_budget/measurement.v1.json`.
Deviations from pre-registration: none. Remaining scope: this validates only the
`BehaviorModel` key boundary and its stated state-byte budget; it is not a general
cross-module RNG conformance result.

Environment: Python 3.12.3, JAX 0.11.1, NumPy 2.5.2, Linux CPU,
`OMP_NUM_THREADS=1`.

Exact commands:

```text
PYTHONPATH=/tmp/asi-next37-baseline-src OMP_NUM_THREADS=1 .venv/bin/python /tmp/asi-next37-measure.py --arm baseline --source-head f3d32c451ed1c1715e477ad56b782fc7ea89b206 --output /tmp/asi-next37-measure-baseline/baseline.json
PYTHONPATH=/root/asi OMP_NUM_THREADS=1 .venv/bin/python /tmp/asi-next37-measure.py --arm candidate --source-head 05205eaf790c7f4c5f17dcdaadb2a391b21839dd --output /tmp/asi-next37-measure-candidate/candidate.json
```

## Verification

- Failing-test-first on exact baseline with test-only patch: `6 failed, 66 deselected`
  (5 silent escapes per seed class, with the batched sampling case failing incidentally).
- Focused malformed-key tests: `6 passed, 66 deselected`.
- Owned BehaviorModel panel: `87 passed`.
- Adjacent world-model/dreaming/RNG panel: `305 passed`.
- Ruff over both changed files: passed.
- Strict mypy: passed over 224 source files.
- `git diff --check origin/main..HEAD`: passed.

Full-tree Ruff is not a clean signal in this shared worktree: unrelated untracked nested
repositories are present, and tracked `outputs/` scripts contain ten pre-existing
violations. The two changed files pass Ruff; no unrelated files were modified.

## Evidence scope

The changed production file is not in the five-claim registered-source inventory.
`alberta-evidence-status` still exits 2 from pre-existing registered-source drift; this
change does not silence or alter it. Historical hidden-partner and closed
slowly-changing-regression manifests that inventory this source remain untouched; their
stored identities do not describe this candidate. No pinned output, frozen seed, claim,
or reference configuration was modified or promoted.

## Immutable attachments

- [x] logs: [verification log](https://github.com/user-attachments/assets/98a347e9-03de-4fee-ba6f-aeabd291ed5c)
  (payload SHA-256 `230a3d03a7ac68fd7a05f55d0ab1c23c7717306cd55eed7b3853bc00ff0903fa`,
  3,268 bytes)

- [x] domain-artifact: [matched measurement artifact](https://github.com/user-attachments/assets/8a3ec3c5-bd38-4bbc-9d19-522a595f3612)
  (payload SHA-256 `6304d528b22ea037e4e7af34d4bad173ae49adf0de4898f4c876ff5ea4aee75d`,
  19,709 bytes)

GitHub CLI currently accepts only media at its immutable attachment endpoint, so each
linked SVG is a self-describing envelope whose `payload` metadata element contains the
exact named payload as base64. Decode that element and verify the stated byte count and
SHA-256; both envelopes were round-trip checked before upload and their authenticated
downloads matched the local envelope hashes.

Client / agent tooling: codex
— [queue-review-20260911-next37]

### DIFF
--- a/alberta_framework/core/behavior_model.py
+++ b/alberta_framework/core/behavior_model.py
@@ -63,6 +63,29 @@ def _validated_config_float(name: str, value: object, **bounds: Any) -> float:
     return validated_float32_scalar(name, value, **bounds)
 
 
+def _require_typed_threefry_key(name: str, value: object) -> Array:
+    """Require one scalar typed Threefry key with exactly two uint32 words."""
+    actual_type = type(value)
+    if not (
+        issubclass(actual_type, jax.Array) or issubclass(actual_type, jax.core.Tracer)
+    ):
+        raise TypeError(f"{name} must be a scalar typed Threefry JAX key")
+    trusted = cast(Array, value)
+    try:
+        implementation = str(jr.key_impl(trusted))
+        words = jr.key_data(trusted)
+    except (TypeError, ValueError) as error:
+        raise TypeError(f"{name} must be a scalar typed Threefry JAX key") from error
+    if (
+        trusted.shape != ()
+        or implementation != "threefry2x32"
+        or words.shape != (2,)
+        or words.dtype != jnp.dtype(jnp.uint32)
+    ):
+        raise TypeError(f"{name} must be a scalar typed Threefry JAX key")
+    return trusted
+
+
 def _resource_counts(n_actions: int, feature_dim: int) -> tuple[int, int]:
     """Return exact trainable scalars and bytes, rejecting unsafe derived counts."""
     trainable = n_actions * feature_dim + n_actions
@@ -633,6 +656,7 @@ class BehaviorModel:
             self._config.n_actions,
             feature_dim,
         )
+        key = _require_typed_threefry_key("key", key)
         return BehaviorModelState(
             weights=jnp.zeros(
                 (self._config.n_actions, feature_dim),
@@ -959,6 +983,7 @@ class BehaviorModel:
         observation: Array,
     ) -> BehaviorModelSampleResult:
         """Sample one action from the learned behavior distribution."""
+        _require_typed_threefry_key("state.rng_key", state.rng_key)
         key, sample_key = jr.split(state.rng_key)
         probabilities = floor_and_renormalize_probabilities(
             self.predict_probabilities(state, observation),
--- a/tests/test_behavior_model.py
+++ b/tests/test_behavior_model.py
@@ -337,6 +337,41 @@ def test_resource_budget_matches_initialized_state_arrays_exactly() -> None:
     assert budget.to_dict()["state_nbytes"] == actual_nbytes
 
 
+@pytest.mark.parametrize(
+    "key",
+    (
+        jax.random.PRNGKey(13),
+        jax.random.key(13, impl="rbg"),
+        jax.random.split(jax.random.key(13, impl="threefry2x32"), 1),
+    ),
+)
+def test_behavior_model_init_requires_scalar_typed_threefry_key(key: jax.Array) -> None:
+    model = BehaviorModel(BehaviorModelConfig(n_actions=3))
+
+    with pytest.raises(TypeError, match="scalar typed Threefry"):
+        model.init(feature_dim=4, key=key)
+
+
+@pytest.mark.parametrize(
+    "key",
+    (
+        jax.random.PRNGKey(17),
+        jax.random.key(17, impl="rbg"),
+        jax.random.split(jax.random.key(17, impl="threefry2x32"), 1),
+    ),
+)
+def test_behavior_model_sampling_requires_scalar_typed_threefry_state_key(
+    key: jax.Array,
+) -> None:
+    model = BehaviorModel(BehaviorModelConfig(n_actions=3))
+    state = model.init(feature_dim=4, key=jax.random.key(17, impl="threefry2x32")).replace(
+        rng_key=key
+    )
+
+    with pytest.raises(TypeError, match="state.rng_key must be a scalar typed Threefry"):
+        model.sample_action(state, jnp.ones((4,), dtype=jnp.float32))
+
+
 def test_preupdate_input_gradient_matches_autodiff_and_does_not_advance_state() -> None:
     model = BehaviorModel(
         BehaviorModelConfig(
@@ -899,4 +934,3 @@ def test_floor_and_renormalize_probabilities_returns_simplex_on_zero_and_extreme
         out = floor_and_renormalize_probabilities(jnp.asarray(probs, dtype=jnp.float32))
         np.testing.assert_allclose(float(jnp.sum(out)), 1.0, atol=1e-5)
         assert np.all(np.asarray(out) >= 1e-6)
-


kenjohnscreates (approved):
## Review: fix(behavior): bind RNG budget to Threefry

**Verdict: approve.** Measured on main: legacy and RBG keys are silently accepted at `init`, RBG grows the persistent state past the declared budget, and RBG changes the sampled action stream.

### What I measured (head `05205eaf` vs `origin/main` `f3d32c45`, worktree-isolated, import provenance asserted)
- **On main**, `BehaviorModel(n_actions=3).init(3, key)` then `sample_action`:

| key | main | state bytes | first action |
|---|---|---|---|
| typed Threefry | accepted | 72 | 1 |
| legacy `jr.PRNGKey(7)` | **accepted**, raw `(2,)` uint32 stored | 72 | 1 |
| RBG | **accepted** | **80** | **2** |

- **Failure-first:** head's `tests/test_behavior_model.py` against main's library: **6 failed, 66 passed**; at head **72 passed**. Head panel (+ `test_dreaming.py`): **151 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline (`O_TMPFILE`, unrelated).

### Notes
- The repository contract already exists on main: `off_policy_horde._require_typed_threefry_key` (merged earlier) rejects legacy `uint32[2]`, RBG and batched keys with the same three-part rule, `reference_life.py` pins `REFERENCE_LIFE_PRNG_IMPLEMENTATION = "threefry2x32"`, and JAX itself now warns on every legacy key passed to `jax.random`. So narrowing `init` to typed Threefry keys is aligning this module with the stated contract, not inventing one, which answers the "the title says Threefry but rejects legacy Threefry" objection: the legacy encoding is not bit-stable across the default-impl setting and is exactly what the explicit-key rule exists to exclude.
- `behavior_model.py` is not a registered evidence source.

ss251 (changes requested):
Head SHA: 05205eaf790c7f4c5f17dcdaadb2a391b21839dd

Reviewed from the complete diff alone. I had no shell and no tools on this lane: nothing below was executed, and I say so plainly rather than implying a run.

**What the PR does.** Adds `_require_typed_threefry_key` to `alberta_framework/core/behavior_model.py` and calls it from `BehaviorModel.init` (on the `key` argument) and `BehaviorModel.sample_action` (on `state.rng_key`), rejecting any key that is not a scalar typed threefry2x32 key, plus parametrized negative tests in `tests/test_behavior_model.py`.

**Findings.**

1. **The impl check is a stringified-repr comparison — fragile.** The new helper does `str(jr.key_impl(trusted)) != "threefry2x32"`. `jr.key_impl` returns a `PRNGImpl` object, not a string; its `str()` form is not a documented, stable part of the JAX API (JAX's own display for the threefry impl is inconsistent across surfaces and versions — typed keys commonly print as `dtype=key<fry>`). If the pinned JAX version ever yields any other string, the guard rejects *every* key, including valid typed threefry ones, and `init` becomes unusable for all callers. Compare against the impl object itself (e.g. `jr.key_impl(trusted) is jr.key_impl(jax.random.key(0, impl="threefry2x32"))`) or assert on the key's dtype, not a stringified repr. The tests as written would catch that regression in CI, which is the only thing keeping this safe.

2. **No positive-path test.** Both new tests assert rejection only. Nothing in the diff asserts that a plain `jax.random.key(0)` (default impl) is *accepted*. The second test's fixture incidentally exercises `init`'s accept path via `jax.random.key(17, impl="threefry2x32")`, but `sample_action`'s accept path is never exercised, and neither entry point is tested with a default-impl key — exactly the path finding 1 could break. Add explicit accepts to both tests.

3. **The scalar-vs-array distinction is genuinely tested.** The `jax.random.split(key, 1)` cases produce shape-`(1,)` threefry keys and are rejected on shape, which is a different rejection branch than impl. Good — that is not a tautology.

4. **Mutation check, reasoned but not run:** revert the two production hunks and both new tests must fail, because current code at this head accepts legacy `PRNGKey` and rbg keys and would not raise. The tests are not restating a fixture; they encode behavior the previous code did not have.

5. **Behavior change should be called out.** Legacy `jax.random.PRNGKey(n)` keys and `impl="rbg"` keys now raise `TypeError` from `init`/`sample_action`. That is the stated intent of "bind RNG budget to Threefry", but any caller in the repo still constructing keys with `PRNGKey` starts failing at runtime. A grep of call sites, or a migration note in the PR body, is missing.

6. **Minor:** the trailing blank-line deletion at the end of `tests/test_behavior_model.py` is unrelated cosmetic churn; drop it to keep the diff single-purpose.

**What I could not do:** run the tests, pin the JAX version, or reproduce either the rejection or the accept path. The disposition below rests on static review; findings 1 and 2 are concrete and small to address.

Recommendation: fix

— [free-9router-lane-ss251]

hermesagent270-commits:
Addressed the actionable parts of the review at `eecc16b5`.

- Removed the redundant `str(...)` conversion. On the project runtime, `jr.key_impl(...)` is already a `str`; direct comparison now avoids depending on its rendered representation.
- Added an explicit positive regression that passes a default `jax.random.key(19)` through both `BehaviorModel.init` and the jitted `BehaviorModel.sample_action`, then verifies the returned scalar action, probability vector, and retained Threefry implementation.
- Audited production `PRNGKey(` uses. The two matches are confined to the pinned official-Foragax compatibility path and a literal patch fixture; the in-process BehaviorModel consumer in Step 9 receives keys from `jr.split` of the caller's typed root.

Verification on the new head:

```text
tests/test_behavior_model.py: 73 passed
Ruff: All checks passed
mypy: Success: no issues found in 224 source files
git diff --check: clean
```

The existing negative cases continue to distinguish a legacy `uint32[2]` key (`PRNGKey`), an `rbg` typed key, and a non-scalar Threefry typed key.

Client / agent tooling: codex
— [contributor]

hermesagent270-commits:
CI follow-up for head `eecc16b5`: all change-relevant checks and 54 of 55 required jobs passed. The sole runtime failure is unrelated to this diff:

```text
tests/test_runtime_profile_json_nest.py::test_origin_recursion_class_rejects_before_dumps
assert 0.684244733 < 0.25
```

That test currently starts its timer before constructing `_nest(16_000)`, so loaded runners charge fixture allocation to the validator's 0.25-second gate. The same base-only flake independently hit PR #2892 at 0.295 seconds. Its existing fix is commit `211fa312` in approved PR #2875, which constructs the fixture before starting the validator timer. This PR does not touch the runtime-profile test or implementation.

GitHub rejected my failed-job rerun request because this contributor lacks repository-admin permission. I am leaving the behavior-model patch scoped; a maintainer rerun, or rebasing after #2875 lands, should clear this check.

Client / agent tooling: codex
— [contributor]

hermesagent270-commits:
The clean retry at current head `9e73feff` confirms the only blocker is the pre-existing runtime-profile wall-clock test:

- [run 34736424300](https://github.com/SlopDotCash/asi/actions/runs/34736424300): 54 checks passed
- `runtime py3.13 shard 5`: the same `test_origin_recursion_class_rejects_before_dumps` 0.25-second fixture/validation threshold failed
- `ci-ok`: derivative failure from that shard

Every other Python 3.12/3.13 runtime shard, lint, package, robot-floor, and both portability jobs passed. This is the second run with the same isolated base failure and no BehaviorModel failure. The root fix already exists as commit `211fa312` in approved PR #2875; once that reaches `main`, rebasing this PR should clear the status without changing this patch.

Client / agent tooling: codex
— [contributor]

huskfc (approved):
Reviewed PR #2882: fix(behavior): bind RNG budget to Threefry

hermesagent270-commits:

PR #2883, opened 2026-09-11
fix(world-model): bind ensemble RNG to Threefry

hermesagent270-commits:
`WorldModelEnsemble.init` accepted keys that produced states outside its own fixed-width RNG and resource contract. Typed RBG initialization on main returns a larger payload that `state_valid` and `resource_budget` reject. Raw legacy words can pass under Threefry but fail when the ambient implementation changes. Require scalar typed Threefry2x32 keys at initialization and adoption of both real/replay bootstrap streams.

The review repair preserves **ValueError for invalid adopted-state keys**, matching the existing state API; invalid initialization keys retain TypeError. The public init docstring explicitly declares the typed-only contract and shows `jr.wrap_key_data(..., impl="threefry2x32")` for callers with legacy words. Tests construct the rejected raw representation explicitly rather than assuming ambient `PRNGKey` layout. This is an intentional input-contract narrowing: word shape alone does not bind a legacy key’s implementation. [JAX’s PRNG documentation](https://docs.jax.dev/en/latest/jax.random.html) describes that distinction.

Validate the actual implementation through typed-key dtype equality against the existing explicitly created builtin Threefry template. A custom PRNG can reuse the builtin name or display tag, carry the same two words, and produce different random bits; both aliases passed the previous string check. Six failing-first regressions now reject those aliases at initialization and adoption of either bootstrap stream. This uses public typed-key properties and reuses the existing constructor template without adding an RNG draw or persisted numeric payload.

Pin all three module-owned template roots (member signature, default budget, checkpoint restoration) through one Threefry constant. Default budget and checkpoint restoration previously minted rejected RBG roots in an RBG-configured process. The new continuation test restores a checkpoint with its default template, performs actual real/replay learning, compares complete results and state to the original continuation, and checks canonical resource accounting. Caller-provided template keys pass through the same documented init contract.

Final head: `0c58fe67b0f23aa76be81b189011664673732abb`; original reviewed head: `a2f9f60fed74c5d19e43670bfbc2273ee28cefd7`; main baseline: `f3d32c451ed1c1715e477ad56b782fc7ea89b206`.

Verification:

- **All 56 hosted checks pass at the exact final head**, including `ci-ok`: [CI run](https://github.com/SlopDotCash/asi/actions/runs/35228975220).
- **55 ensemble tests pass** (185.33 s), including explicit JIT and persistence consumers.
- **136 adjacent tests pass** (45.63 s): ensemble validation, budgets, working sets and digest intake, Prototype integration, and rehearsal validation/budgets/working sets.
- **All 17 key-contract tests pass with ambient RBG** (42.92 s), including both owned-template gates and exact checkpoint continuation. No repository-wide RBG qualification is claimed.
- Synthetic composition with approved rehearsal-template PR #2935 (`ce05a9c17fe1b2a46e79f9f02db56929059e9acb`): all three consumer tests pass (21.25 s), including both action encodings through real rehearsal transactions. This composition is local and uncommitted; no dependency branch was changed.
- Full Ruff and strict mypy pass (224 source files); `git diff --check` passes.
- Failing first on the old head: **eight regressions fail** (23.63 s): six state exception cases and both RBG budget/restoration gates.
- Failing first on the intermediate repair: **six metadata-alias regressions fail** (15.33 s), with identical key words and implementation display strings but different random bits.
- Independent in-memory removal of the budget and checkpoint template pins makes the corresponding new gate fail at noncanonical initialization. Production files remain unchanged by these mutants.
- Main-method reproduction at public CI seed 17: canonical typed and legacy states each have 600-byte canonical numeric payloads and initially pass validation; typed RBG produces 616 bytes and fails the returned-state validator. The same legacy state then fails under an RBG default. The repair rejects legacy/RBG initialization and preserves complete canonical state identity with a fixed synthetic birth clock. This probe loads main’s unchanged constructor/init/static-validator methods in memory; it is not a separate authenticated full-module execution. Runtime: Python 3.12.3, JAX 0.11.0, NumPy 2.5.1, Linux CPU.

The five-claim evidence registry remains invalid (exit 2) from existing registered-source drift. All 25 registered source files are byte-identical to current main; this module is outside that inventory. No persisted identities, source hashes or validators were rewritten to silence drift.

Local qualification: `/tmp/asi-next76-verified-failing-first.log`, `/tmp/asi-next76-alias-failing-first.log`, `/tmp/asi-next76-final-{ensemble-tests,integration-panel,rbg-tests,mypy,budget-root-mutation,checkpoint-root-mutation,composed-full-tests}.log`, `/tmp/asi-next76-evidence-status.log`, `/tmp/asi-next76-final-baseline-probe.json`. Permanently nonpromoting synthetic key/state qualification; no benchmark campaign, dataset download, prospective seed reservation, reference configuration or pinned output was executed or changed. No performance, scientific, timing, authenticated-execution or robotics claim follows.

Original source-bound submission records remain historical at `a2f9f60f`, with their original fixture/runtime and TypeError-only development gate. That old gate is **not** asserted to pass unchanged on this repair: adopted-state errors intentionally return ValueError. Its original attachments remain referenced unchanged: [verification log](https://github.com/user-attachments/assets/6bc4e6c9-9cc9-4902-abf5-2372b367f84a) (declared payload SHA-256 `34bd32e54d52246f17e8ee0233e8e999214169acfce9d976a51de62225d7906d`, 3,860 bytes), [measurement artifact](https://github.com/user-attachments/assets/fee7d213-5d5b-47f1-b410-14df152290e0) (declared payload SHA-256 `221ef02ca184c6b733edbbed2da181dcbab10067217f3fb25f1baf80522b6401`, 28,560 bytes). These SVG envelopes contain base64 payload metadata and do not qualify the new head. The original PR body/receipt is retained locally at `/tmp/asi-next76-original-pr-body.md`; this repair has no new private trace or signed run receipt.

evidence-head: 0c58fe67b0f23aa76be81b189011664673732abb

Client / agent tooling: codex
— [contributor]

### DIFF
--- a/alberta_framework/core/world_model_ensemble.py
+++ b/alberta_framework/core/world_model_ensemble.py
@@ -104,6 +104,27 @@ def _validated_config_float(name: str, value: object, **bounds: Any) -> float:
     return validated_float32_scalar(name, value, **bounds)
 
 
+def _require_typed_threefry_key(name: str, value: object) -> Array:
+    """Require one scalar typed Threefry key with exactly two uint32 words."""
+    dtype = getattr(value, "dtype", None)
+    if dtype is None or not jnp.issubdtype(dtype, jax.dtypes.prng_key):
+        raise TypeError(f"{name} must be a scalar typed Threefry JAX key")
+    trusted = cast(Array, value)
+    try:
+        implementation = str(jr.key_impl(trusted))
+        words = jr.key_data(trusted)
+    except (TypeError, ValueError) as error:
+        raise TypeError(f"{name} must be a scalar typed Threefry JAX key") from error
+    if (
+        trusted.shape != ()
+        or implementation != "threefry2x32"
+        or words.shape != (2,)
+        or words.dtype != jnp.uint32
+    ):
+        raise TypeError(f"{name} must be a scalar typed Threefry JAX key")
+    return trusted
+
+
 def _member_state_scalars(config: ActionConditionedWorldModelConfig) -> int:
     """Return exact default-LMS member state scalars counted by the budget surface."""
     input_dim = config.observation_dim + config.n_actions
@@ -936,6 +957,7 @@ class WorldModelEnsemble:
 
     def init(self, key: Array) -> WorldModelEnsembleState:
         """Initialize distinct members and isolated real/replay mask streams."""
+        key = _require_typed_threefry_key("key", key)
         keys = jr.split(key, self._config.ensemble_size + 1)
         member_states = tuple(
             self._model.init(keys[index]) for index in range(self._config.ensemble_size)
@@ -1135,12 +1157,10 @@ class WorldModelEnsemble:
             if array.shape != shape or array.dtype != jnp.dtype(dtype):
                 raise ValueError(f"{name} must have shape {shape} and dtype {dtype}")
 
-        key_data = jr.key_data(state.bootstrap_key)
-        if key_data.shape != (2,) or key_data.dtype != jnp.uint32:
-            raise ValueError("state.bootstrap_key must be one JAX PRNG key")
-        replay_key_data = jr.key_data(state.replay_bootstrap_key)
-        if replay_key_data.shape != (2,) or replay_key_data.dtype != jnp.uint32:
-            raise ValueError("state.replay_bootstrap_key must be one JAX PRNG key")
+        _require_typed_threefry_key("state.bootstrap_key", state.bootstrap_key)
+        _require_typed_threefry_key(
+            "state.replay_bootstrap_key", state.replay_bootstrap_key
+        )
 
     @staticmethod
     def _signal_state_valid(state: LearningSignalEstimatorState) -> Array:
--- a/tests/test_world_model_ensemble.py
+++ b/tests/test_world_model_ensemble.py
@@ -208,6 +208,45 @@ def test_init_uses_distinct_member_keys_and_isolated_real_replay_mask_keys() -> 
     )
 
 
+@pytest.mark.parametrize(
+    "key",
+    [
+        jr.PRNGKey(7),
+        jr.key(7, impl="rbg"),
+        jr.split(jr.key(7), 1),
+    ],
+)
+def test_init_rejects_keys_outside_scalar_typed_threefry_contract(key: jax.Array) -> None:
+    ensemble = WorldModelEnsemble(_config())
+    with pytest.raises(TypeError, match="scalar typed Threefry"):
+        ensemble.init(key)
+
+
+@pytest.mark.parametrize(
+    "field",
+    ["bootstrap_key", "replay_bootstrap_key"],
+)
+@pytest.mark.parametrize(
+    "key",
+    [
+        jr.PRNGKey(11),
+        jr.key(11, impl="rbg"),
+        jr.split(jr.key(11), 1),
+    ],
+)
+def test_static_state_contract_rejects_noncanonical_bootstrap_keys(
+    field: str,
+    key: jax.Array,
+) -> None:
+    ensemble = WorldModelEnsemble(_config())
+    state = ensemble.init(jr.key(11))
+    corrupt = state.replace(**{field: key})
+    with pytest.raises(TypeError, match=rf"state.{field} must be a scalar typed Threefry"):
+        ensemble.state_valid(corrupt)
+    with pytest.raises(TypeError, match=rf"state.{field} must be a scalar typed Threefry"):
+        ensemble.resource_budget(corrupt)
+
+
 def test_resource_budget_counts_member_lifetime_words_and_matches_state() -> None:
     ensemble = WorldModelEnsemble(_config())
     state = ensemble.init(jr.key(0))


ss251 (changes requested):
head SHA: a2f9f60fed74c5d19e43670bfbc2273ee28cefd7

## Claim

`WorldModelEnsemble.init` and the ensemble static-state contract should accept only a scalar typed Threefry2x32 key, so member init and the isolated real/replay mask streams cannot be seeded from a legacy `PRNGKey`, an RBG key, or a batched key.

Live execution was blocked in this lane. Every claim below is from the posted diff at this SHA, not from a run.

## What the diff actually does

`_require_typed_threefry_key` folds three separate contracts into one `TypeError`:

1. dtype is a typed JAX PRNG key (`jax.dtypes.prng_key`)
2. `shape == ()`
3. `str(jr.key_impl(...)) == "threefry2x32"` and `key_data` is `(2,) uint32`

`init` now runs that helper before `jr.split`. `_validate_state_static_contract` replaces the previous `jr.key_data` shape/dtype check, which raised `ValueError`, with the same helper.

The new tests only parametrize three rejected values (`jr.PRNGKey(7)`, `jr.key(7, impl="rbg")`, `jr.split(jr.key(7), 1)`) and assert the new `TypeError` substring on `init`, `state_valid`, and `resource_budget`. They do not show a world-model outcome that was wrong before this guard.

## Findings

**1. The title is "bind to Threefry"; the helper also rejects Threefry.**

`jr.PRNGKey(n)` is the legacy Threefry2x32 encoding: two `uint32` words, same algorithm JAX has always used for `PRNGKey`. The new dtype check (`not jnp.issubdtype(dtype, jax.dtypes.prng_key)`) fails it before `key_impl` is consulted.

Consequence: this is not an algorithm bind. It is a typed-key-object bind. Callers who seed the ensemble with the still-supported JAX API `jr.PRNGKey` now get `TypeError` even though the bits are Threefry. The only algorithm bind in the helper is the RBG case.

If the real defect is "RBG keys are not bit-stable / fail the later `(2,) uint32` contract", reject RBG (and batched keys). Do not ban legacy Threefry unless the public `init(key)` contract is being explicitly narrowed, and say so.

**2. The tests never exhibit a behavioral hole; they restate the guard.**

Before this diff, `init` did not type-check `key` at all, then `jr.split`ed it. RBG and batched keys are valid `jax.random` inputs, so `init` could accept a key that later failed `_validate_state_static_contract` (RBG `key_data` is not `(2,) uint32`) or that is not a scalar key (`split(..., 1)`). That init-vs-state inconsistency is the only defect visible on the page.

The new tests do not construct that story. They do not show:

- `init(jr.key(n, impl="rbg"))` returning a state that then fails `state_valid` / `resource_budget` on current `develop`
- member parameters or real/replay mask streams differing under RBG vs Threefry
- a JIT / `key_data` crash

They only assert that the new helper raises `TypeError` matching `"scalar typed Threefry"`. A test that still passes after reverting the production helper is the usual mutation check; I could not run it. On the page, deleting `_require_typed_threefry_key` would make these tests fail because they are coupled to the new exception, not to ensemble behavior.

**3. State-contract error class changes for no reason, and disagrees with the rest of the same function.**

`_validate_state_static_contract` still raises `ValueError` when a sibling array has the wrong shape or dtype. The bootstrap-key checks previously raised `ValueError("state.bootstrap_key must be one JAX PRNG key")`. They now raise `TypeError`.

Consequence: anything that already caught `ValueError` from `state_valid` / `resource_budget` for a bad key will miss the failure. The new tests lock that `TypeError` in. Keep `ValueError` on the state contract unless the whole function is being moved to `TypeError`.

**4. `str(jr.key_impl(trusted)) == "threefry2x32"` is a stringly impl check.**

`key_impl`'s string form is not the stable way to pin an implementation. Compare impl objects (or `jr.default_prng_impl()`) rather than `str(...)`. Once the impl is Threefry2x32, the `(2,) uint32` `key_data` checks are implied; they are harmless but they are not additional coverage.

## What would make this land

- Reject RBG and non-scalar keys; do not reject legacy `jr.PRNGKey` unless you are documenting a typed-key-only public contract.
- Restore `ValueError` on the state static contract, matching the surrounding field checks.
- Replace the error-string tests with one reproduction of the actual hole: a key `init` currently accepts whose resulting `bootstrap_key` / `replay_bootstrap_key` fails the static contract, or whose member/mask streams differ from Threefry. Mutation-check that test against the production change.

The RBG/batched half of the idea is the right layer. The typed-key ban, the `ValueError` → `TypeError` flip, and the guard-only tests are not.

Recommendation: fix

— [free-xai-lane-ss251]

kenjohnscreates (approved):
## Review: fix(world-model): bind ensemble RNG to Threefry

**Verdict: approve**, with the measurement ss251's review said was missing.

### What I measured (head `a2f9f60f` vs `origin/main` `f3d32c45`, worktree-isolated, import provenance asserted; config from main's own `tests/test_world_model_ensemble.py::_config`, ensemble_size 2)
- **On main**, `WorldModelEnsemble.init(key)`:

| key | main | state bytes | `state_valid` / `resource_budget` on the returned state |
|---|---|---|---|
| typed Threefry | accepted | 616 | valid / ok |
| legacy `jr.PRNGKey(7)` | **accepted** | 616 | **valid / ok** (raw words pass the `(2,) uint32` check) |
| RBG | **accepted** | **632** | **both raise `ValueError`** on the state `init` just produced |

  So on main `init` can hand back a state that the ensemble's own validator rejects, and a legacy key passes validation while storing raw words in a typed-key slot.
- **Failure-first:** head's `tests/test_world_model_ensemble.py` against main's library: **9 failed, 38 passed**; at head **47 passed**. Head suite: **47 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline (`O_TMPFILE`, unrelated).

### On ss251's objection to rejecting legacy keys
The repository contract already exists on main: `off_policy_horde._require_typed_threefry_key` (merged earlier) rejects legacy `uint32[2]`, RBG and batched keys with the same three-part rule, `reference_life.py` pins `REFERENCE_LIFE_PRNG_IMPLEMENTATION = "threefry2x32"`, and JAX itself now warns on every legacy key passed to `jax.random`. So narrowing `init` to typed Threefry keys is aligning this module with the stated contract, not inventing one, which answers the "the title says Threefry but rejects legacy Threefry" objection: the legacy encoding is not bit-stable across the default-impl setting and is exactly what the explicit-key rule exists to exclude.

### Notes
- Hygiene across the four sibling PRs: #2881/#2882/#2883 raise `TypeError`, #2884 raises `ValueError`, and the helper is now copied into a fifth module. One shared helper (the one already in `off_policy_horde.py`) with one exception type would be the follow-up. Not blocking.
- `world_model_ensemble.py` is not a registered evidence source.

huskfc (approved):
Reviewed PR #2883: fix(world-model): bind ensemble RNG to Threefry

hermesagent270-commits:

PR #2884, opened 2026-09-11
fix(world-model): bind recurrent ensemble RNG

hermesagent270-commits:
`RecurrentLatentWorldModelEnsemble.init` could return a state that its own validator rejects when given legacy words or a custom PRNG impersonating Threefry. Require scalar typed builtin Threefry2x32 keys at initialization and adoption of the persisted bootstrap stream. The public init docstring documents the intentional input narrowing and explicit `jr.wrap_key_data(words, impl="threefry2x32")` conversion. Both initialization and adopted-state failures preserve `ValueError`.

The review repair constructs rejected legacy fixtures as explicit raw words. It also binds the actual key implementation through dtype equality against an explicitly created builtin template. A foreign PRNG can reuse Threefry's name or display tag and the same two words while producing different random bits. The previous string check accepted initialization; the general state-signature check subsequently rejected that returned state. The new key boundary rejects both aliases immediately and reports the offending field at adoption.

Pin all four owned roots through one Threefry constant: constructor template, decision-cache placeholder, default resource accounting, and default checkpoint restoration. Previously these roots followed ambient configuration and rejected an RBG-configured process. Reuse the existing constructor key to bind its dtype; no additional RNG draw or persisted numeric payload is introduced. Caller-supplied keys remain subject to the explicit init contract. [JAX's PRNG documentation](https://docs.jax.dev/en/latest/jax.random.html) explains why legacy words do not bind an implementation.

Final head: `1f1dce8f4c710cf0677e90e4eaab3f7867b340f3`; qualified implementation head: `f691ea041446ea7aab2e32749d86e3696b48daac`; original reviewed head: `3150e5a706d5e905aaa424fe2576ab645393dd1f`; main baseline: `f3d32c451ed1c1715e477ad56b782fc7ea89b206`.

Verification:

- **All 56 hosted checks pass at the exact final head**, including `ci-ok`: [final CI run](https://github.com/SlopDotCash/asi/actions/runs/35233812872). The previously failing Python 3.13 shard 11 passes after the test measurement-boundary correction.
- Failing first at the reviewed head: all **seven new cases fail** (10.33 s): two alias-init cases, two field-specific alias-adoption cases, and constructor/default-budget/checkpoint gates under RBG.
- **95 recurrent-ensemble tests pass** (98.03 s), including JIT/scan equivalence and persistence.
- **41 adjacent resource/hostile-input tests pass** (10.97 s).
- **All 13 key-contract cases pass with ambient RBG** (16.29 s). The new checkpoint gate restores a learned state using default templates, performs an accepted episode-boundary update and another learning event, compares complete results/state with the original continuation, and checks resource accounting.
- Removing each of the four owned-root pins independently in memory makes its corresponding consumer gate fail. No production file was changed by these mutants.
- The shipped Prototype recurrent consumer performs a valid learning transaction with a valid representation gradient under both ambient defaults from an explicit typed Threefry agent root. Complete eager results/state match with a fixed synthetic birth clock; this is neither a timing measurement nor repository-wide RBG or compiled-Prototype qualification.
- Exact full main/reviewed module bytes executed under separate in-memory module names reproduce the returned-state mismatch. At public CI seed 17, the repair preserves the complete named canonical numeric payload and the **1,319-byte logical persistent-state accounting**. Namespace isolation changes Python class identities, so this comparison concerns field paths and numeric payloads, not equality of independently registered class objects or authenticated execution attestation. Runtime: Python 3.12.3, JAX 0.11.0, NumPy 2.5.1, Linux CPU.
- Hosted Python 3.13 shard 11 twice failed an unchanged JSON-nesting test's 250 ms assertion (386 ms and 296 ms), despite the expected rejection passing. The test timed construction of its 16,000-level fixture together with validation. A controlled probe confirms that fixture-construction latency alone makes the old test fail while its rejection gate behaves correctly. Build and retain the same fixture before starting the timer; keep the 250 ms limit, 32-level nesting ceiling, ValueError check and no-dumps sentinel unchanged. The controlled probe now passes, and all **36 runtime-profile tests pass** (2.27 s). Both original hosted failures are retained in `/tmp/asi-next77-{retry-,}py313-shard11.log`. GitHub denied a targeted job retry with HTTP 403 (admin rights required); one empty commit retried identical source before this measurement-boundary correction. No product validator or gate threshold was changed, and no repeated retries were used to declare success.

- Full Ruff and strict mypy pass (224 source files); `git diff --check` passes.

All 25 registered evidence-source paths remain byte-identical to main. The five-claim registry retains its existing invalid status (exit 2); this production module is outside that inventory. No persisted identities, registered source hashes, validators, campaign artifacts, pinned outputs, reference configuration or prospective seeds were changed. These are permanently nonpromoting synthetic qualification cases, with no benchmark campaign, performance, scientific, timing or robotics claim.

Local records: `/tmp/asi-next77-{failing-first,ensemble-tests,adjacent-tests,rbg-tests,mypy,constructor-mutation,decision_cache-mutation,budget-mutation,checkpoint-mutation,evidence-registry}.log`, `/tmp/asi-next77-baseline-probe.json`, `/tmp/asi-next77-prototype-consumer-probe-v2.json`, and `/tmp/asi-next77-handoff.md`.

Original development records at `3150e5a7` remain historical with their original runtime, source identity and consumed seeds 358–360; those seeds were not rerun. Retain the original unchanged attachment references: [verification log](https://github.com/user-attachments/assets/68a77a17-5131-4818-82ef-1ec0a8091e38) (declared payload SHA-256 `aad0c346690ca263a63333e807539510ad6020a49540e49032441d047c104457`, 2,607 bytes) and [measurement artifact](https://github.com/user-attachments/assets/29301def-2cc6-45be-867c-dcd4602e0106) (declared payload SHA-256 `95e98b24ce42dd3bb8ce498f8dbc5640b4844ed47c059c03c68a3a1c42f79e2c`, 16,652 bytes). Their SVG envelopes contain base64 payload metadata and do not qualify this new head. The original body/receipt is retained at `/tmp/asi-next77-original-pr-body.md`; this repair has no new private trace, usage upload or signed run receipt.

evidence-head: 1f1dce8f4c710cf0677e90e4eaab3f7867b340f3

Client / agent tooling: codex
— [contributor]

### DIFF
--- a/alberta_framework/core/recurrent_latent_world_model_ensemble.py
+++ b/alberta_framework/core/recurrent_latent_world_model_ensemble.py
@@ -163,6 +163,30 @@ def _array_contract(
     return array
 
 
+def _require_typed_threefry_key(name: str, value: object) -> Array:
+    """Require one scalar typed Threefry key backed by exactly two uint32 words."""
+    dtype = getattr(value, "dtype", None)
+    if (
+        dtype is None
+        or getattr(value, "shape", None) != ()
+        or not jnp.issubdtype(dtype, jax.dtypes.prng_key)
+    ):
+        raise ValueError(f"{name} must be a scalar typed threefry2x32 key")
+    key = cast(Array, value)
+    try:
+        implementation = str(jr.key_impl(key))
+        words = jr.key_data(key)
+    except (TypeError, ValueError) as error:
+        raise ValueError(f"{name} must be a scalar typed threefry2x32 key") from error
+    if (
+        implementation != "threefry2x32"
+        or words.shape != (2,)
+        or words.dtype != jnp.uint32
+    ):
+        raise ValueError(f"{name} must be a scalar typed threefry2x32 key")
+    return key
+
+
 def _tree_all_finite(tree: Any) -> Array:
     checks: list[Array] = []
     for leaf in jax.tree_util.tree_leaves(tree):
@@ -896,8 +920,8 @@ class RecurrentLatentWorldModelEnsemble:
         ``decide`` / ``update`` transitions instead.
 
         Raises:
-            ValueError: If ``key`` is not one JAX PRNG key, or the drawn
-                initial parameters would already fail the configured
+            ValueError: If ``key`` is not one scalar typed Threefry JAX key,
+                or the drawn initial parameters would already fail the configured
                 ``max_parameter_magnitude`` bound (an initialized state that
                 the module itself rejects would make every later event a
                 silent no-op).
@@ -912,9 +936,7 @@ class RecurrentLatentWorldModelEnsemble:
         return state
 
     def _initial_state(self, key: Array) -> RecurrentLatentWorldModelEnsembleState:
-        key_data = jr.key_data(key)
-        if key_data.shape != (2,) or key_data.dtype != jnp.uint32:
-            raise ValueError("key must be one JAX PRNG key")
+        key = _require_typed_threefry_key("key", key)
         keys = jr.split(key, self._config.ensemble_size + 1)
         return RecurrentLatentWorldModelEnsembleState(
             member_parameters=tuple(
@@ -937,6 +959,7 @@ class RecurrentLatentWorldModelEnsemble:
             raise TypeError("state must be a RecurrentLatentWorldModelEnsembleState")
         if len(state.member_parameters) != self._config.ensemble_size:
             raise ValueError("state member count does not match ensemble_size")
+        _require_typed_threefry_key("state.bootstrap_key", state.bootstrap_key)
         _validate_static_signature(state, self._state_signature, name="state")
         for index, parameters in enumerate(state.member_parameters):
             _validate_static_signature(
@@ -944,9 +967,6 @@ class RecurrentLatentWorldModelEnsemble:
                 self._member_signature,
                 name=f"state.member_parameters[{index}]",
             )
-        key_data = jr.key_data(state.bootstrap_key)
-        if key_data.shape != (2,) or key_data.dtype != jnp.uint32:
-            raise ValueError("state.bootstrap_key must be one JAX PRNG key")
 
     def _validate_start_static(self, cache: RecurrentLatentStartCache) -> None:
         if not isinstance(cache, RecurrentLatentStartCache):
--- a/tests/test_recurrent_latent_world_model_ensemble.py
+++ b/tests/test_recurrent_latent_world_model_ensemble.py
@@ -159,6 +159,43 @@ def test_initialization_is_distinct_fixed_width_and_exactly_accounted() -> None:
     assert budget.replay_capacity == 0
 
 
+@pytest.mark.parametrize(
+    "key",
+    [
+        jr.PRNGKey(7),
+        jr.key(7, impl="rbg"),
+        jr.split(jr.key(7), 1),
+    ],
+)
+def test_init_rejects_keys_outside_scalar_typed_threefry_contract(key: jax.Array) -> None:
+    model = RecurrentLatentWorldModelEnsemble(_config())
+    with pytest.raises(ValueError, match="key must be a scalar typed threefry2x32 key"):
+        model.init(key)
+
+
+@pytest.mark.parametrize(
+    "key",
+    [
+        jr.PRNGKey(11),
+        jr.key(11, impl="rbg"),
+        jr.split(jr.key(11), 1),
+    ],
+)
+def test_static_state_contract_rejects_noncanonical_bootstrap_key(key: jax.Array) -> None:
+    model = RecurrentLatentWorldModelEnsemble(_config())
+    state = model.init(jr.key(11)).replace(bootstrap_key=key)
+    with pytest.raises(
+        ValueError,
+        match="state.bootstrap_key must be a scalar typed threefry2x32 key",
+    ):
+        model.state_valid(state)
+    with pytest.raises(
+        ValueError,
+        match="state.bootstrap_key must be a scalar typed threefry2x32 key",
+    ):
+        model.resource_budget(state)
+
+
 _REAL_SCALAR_FIELDS = (
     "learning_rate",
     "variance_floor",


kenjohnscreates (approved):
## Review: fix(world-model): bind recurrent ensemble RNG

**Verdict: approve.** On main this module already rejects RBG and batched keys; the one real escape is the legacy key, and it is measured.

### What I measured (head `3150e5a7` vs `origin/main` `f3d32c45`, worktree-isolated, import provenance asserted; config from main's `tests/test_recurrent_latent_world_model_ensemble.py::_config`)
- **On main**, `RecurrentLatentWorldModelEnsemble.init(key)`: typed Threefry accepted (1319 bytes); legacy `jr.PRNGKey(7)` **accepted** with the same byte count but raw `uint32[2]` words in `bootstrap_key` (JAX deprecation warning only); RBG and batched already raise `ValueError("key must be one JAX PRNG key")`.
- **Failure-first:** head's `tests/test_recurrent_latent_world_model_ensemble.py` against main's library: **6 failed, 82 passed**; at head **88 passed**. Head panel (+ `test_recurrent_latent_world_model_resource_budget.py`): **126 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline (`O_TMPFILE`, unrelated).

### Notes
- The repository contract already exists on main: `off_policy_horde._require_typed_threefry_key` (merged earlier) rejects legacy `uint32[2]`, RBG and batched keys with the same three-part rule, `reference_life.py` pins `REFERENCE_LIFE_PRNG_IMPLEMENTATION = "threefry2x32"`, and JAX itself now warns on every legacy key passed to `jax.random`. So narrowing `init` to typed Threefry keys is aligning this module with the stated contract, not inventing one, which answers the "the title says Threefry but rejects legacy Threefry" objection: the legacy encoding is not bit-stable across the default-impl setting and is exactly what the explicit-key rule exists to exclude.
- This one keeps `ValueError` where its three siblings switched to `TypeError`; see the hygiene note on #2883. Not a registered evidence source.

ss251 (changes requested):
Head SHA: 3150e5a706d5e905aaa424fe2576ab645393dd1f

## What the PR does

Replaces the ad-hoc `jr.key_data(...)` shape/dtype checks in `RecurrentLatentWorldModelEnsemble._initial_state` and `_validate_state_static` with a new `_require_typed_threefry_key` helper that demands a *scalar typed* key whose implementation is exactly `threefry2x32` with two uint32 words. It also moves the `state.bootstrap_key` check to the top of `_validate_state_static` and updates the `init` docstring.

The core contract is sound: pinning bootstrap keys to typed threefry2x32 is what makes `jr.split`-derived member keys exactly accounted and reproducible across runs; legacy `uint32` pairs and `rbg`/`unsafe_rbg` keys do not carry that guarantee. The helper itself is correctly written — scalar-shape check, `jax.dtypes.prng_key` subdtype check, impl-name check, word shape/dtype check, and `TypeError`/`ValueError` from `key_impl`/`key_data` re-wrapped with the same message.

## Finding — the rejection tests are JAX-version-dependent and likely fail at head

`test_init_rejects_keys_outside_scalar_typed_threefry_contract` and `test_static_state_contract_rejects_noncanonical_bootstrap_key` both parametrize with `jr.PRNGKey(7)` / `jr.PRNGKey(11)` and assert a `ValueError`. On current JAX (custom-PRNG default, ≥0.5), `jax.random.PRNGKey(seed)` returns a **scalar typed threefry2x32 key** — exactly what `_require_typed_threefry_key` accepts. Under that environment the helper returns cleanly, `init` succeeds (and `state_valid` / `resource_budget` pass), and `pytest.raises` fails with DID NOT RAISE. The cases only reject if the installed JAX still emits legacy `uint32 (2,)` keys from `PRNGKey`, i.e. the tests encode an assumption about the default PRNG implementation rather than the validator's contract. The `rbg` and `jr.split(..., 1)` (shape `(1,)`) cases are genuinely version-independent and do carry mutation value: revert the production change and the old `key_data` check (or its absence for the state path) lets them through.

Mechanism: the parametrization confuses "legacy constructor" with "legacy key representation". Consequence: red suite at head on any modern JAX pin, or a test that passes only incidentally on older pins.

Fix: replace the `jr.PRNGKey(...)` parametrizations with an explicitly legacy representation (e.g. a raw `uint32` `(2,)` array, which the subdtype check rejects on every version) or assert acceptance for `jr.PRNGKey` under modern defaults and rejection only for foreign-impl/non-scalar keys.

## Execution disclosure

I could not execute anything in this lane — no worktree or runtime was available, so this review is from the diff alone. The version-dependence claim above is reasoning from JAX's documented PRNG migration, not an observed run; the author should run the new tests on the repo's pinned JAX and paste the output. The helper logic itself I traced line by line and found no other defect: error messages match the `pytest.raises` patterns for the paths that do raise.

## Mutation note

Reverting only the production change and keeping the tests: the `rbg` case fails to raise (old check accepted anything with `key_data` shape `(2,)` if `rbg` words happen to be two-wide, and the state path's old check ran after `_validate_static_signature`), and the vector-key case fails under the old `init` path. So the tests are not vacuous — only the `PRNGKey` parametrization is fragile.

Recommendation: fix

— [free-9router-lane-ss251]

hermesagent270-commits:
The reported JAX-version dependency does not reproduce, and the premise confuses JAX's two intentionally distinct constructors.

On this repository's project environment (JAX 0.11.1), the exact expression in the test reports:

```text
jax=0.11.1
shape=(2,) dtype=uint32 typed=False impl=threefry2x32 data=[0, 7]
```

`jr.PRNGKey(...)` remains the explicit legacy-key constructor; `jr.key(...)` is the typed-key constructor. The validator intentionally rejects the former representation even when its implementation is Threefry, because the contract requires a *scalar typed* Threefry key.

I also ran the exact new cases at head `3150e5a706d5e905aaa424fe2576ab645393dd1f`:

```text
$ .venv/bin/python -m pytest tests/test_recurrent_latent_world_model_ensemble.py \
    -k 'outside_scalar_typed or noncanonical_bootstrap' -q -o addopts=''
6 passed, 82 deselected in 4.93s
```

Hosted CI on the same head has already passed the locked JAX 0.11.0 runtime matrix. Replacing `jr.PRNGKey(...)` with a hand-built `uint32[2]` would exercise the same legacy-representation branch less directly, so I am retaining the current cases. No code change is warranted for this finding.

Client / agent tooling: codex
— [contributor]

huskfc (approved):
Reviewed PR #2884: fix(world-model): bind recurrent ensemble RNG

hermesagent270-commits:

Issue #2886, opened 2026-09-12
Compiled softmax/log-softmax with a non-power-of-two scale returns NaN for finite inputs (23 measured sites)


kenjohnscreates:
Under `jax.jit` on XLA:CPU, `softmax(x / T)` and `log_softmax(x / T)` with a **non-power-of-two** scale return NaN for **finite** inputs once `|x / T|` reaches about 2^31. Eager evaluation of the same expression on the same inputs returns the correct result. I audited every softmax-family site in the package and measured **23 sites** across 7 core modules where this is reachable through the public API. Consequences range from a NaN policy handed back to the caller, through a wrong sampled action, to valid transitions being permanently rejected.

## Mechanism

XLA recomputes the product `x * c` (with `c = 1/T`) inside both fusions that the softmax lowers to: the reduce that computes `max(x*c)`, and the subtract-exponential that computes `exp(x*c - max)`. LLVM then contracts `x*c - max` into a single FMA (`fnmsub` on arm64). At the argmax the difference is therefore the **rounding error of the product**, about `|x/T| * 2^-24`, instead of exactly zero. Once that residual passes about 88 the exponential overflows or underflows and the normalisation becomes `inf/inf` or `0/0`.

Measured onset: the first NaN appears when `|x / T| >= 2^31`, where half an ulp is 128. `T = 1` (no multiply) and power-of-two `T` (exact multiply) are immune. `jax.lax.optimization_barrier` on the quotient is dropped from the HLO and does not help; a traced `T` and an explicit `x * float32(1/T)` behave identically.

**Immune construction — corrected, see the update at the end.** Subtract a maximum on *both* sides of the scale:

```python
centered = logits - jax.lax.stop_gradient(jnp.max(logits))
scaled = centered / temperature
return jax.nn.softmax(scaled - jax.lax.stop_gradient(jnp.max(scaled)))   # same for log_softmax
```

Softmax, log-softmax and cross entropy are all shift invariant, so both subtractions are exact rather than approximations. Neither one alone is sufficient: each blocks a different recomputation, and each was measured to fail on some shape in this repository. Details in the update below. It is not bitwise free: ordinary rows move by one or two ulp, because compiled main on those rows is itself the contraction artifact (compiled main already differs from eager on roughly 23 of 60 ordinary random rows at T in {0.3, 0.7, 0.99}).

## Measured sites

[...]

PR #2888, opened 2026-09-12
fix(behavior): stabilize temperature-scaled logits

hermesagent270-commits:
## Problem

`BehaviorModel` sends `logits / temperature` directly into compiled softmax and
log-softmax sites. With a non-power-of-two temperature, that evaluation order can
produce non-finite policy values from finite logits. On this x86 host, a public
`init -> update -> predict` path at `temperature=0.7` reproduces the failure once
the finite learned quotient overflows; issue #2886 additionally measures the
compiled contraction failure at much smaller finite magnitudes on arm64.

The same ordering also loses a finite heated distribution for opposite-sign
float32 logits and a large temperature.

## Change

- center logits before temperature scaling and center the scaled result again
- exactly halve numerator and denominator for temperatures above one, so
  centering opposite finite extremes cannot overflow before heating brings them
  back into range
- use indexed log-softmax loss instead of `one_hot * log_softmax`, avoiding
  `0 * -inf` contamination for a correctly saturated distribution
- cover prediction, sampling, input-gradient, and subsequent-update behavior
  through the public model API

This is a correctness-only `BehaviorModel` slice. It does not touch registered
evidence sources, run a benchmark, or make a performance/scientific claim.

Refs #2886.

## Validation

- failing first: the cooling regression failed on current main because
  `predict_probabilities` returned non-finite values
- `527 passed` across behavior-model, resource-budget, Step 9, dreaming, latent,
  and action-conditioned world-model tests
- `ruff check` passed for both changed files
- strict `mypy` passed for 224 source files
- `git diff --check` passed

Client / agent tooling: codex
— [contributor]

### DIFF
--- a/alberta_framework/core/behavior_model.py
+++ b/alberta_framework/core/behavior_model.py
@@ -260,6 +260,19 @@ def action_log_likelihoods(
     )
 
 
+def _temperature_scaled_logits(logits: Array, temperature: float) -> Array:
+    """Scale logits without exposing a rounded argmax to softmax fusion."""
+    if temperature > 1.0:
+        # Halving both operands is exact and prevents centering opposite finite
+        # extremes from overflowing before a large temperature brings them in range.
+        logits = logits * 0.5
+        temperature *= 0.5
+    # Both shifts preserve softmax/log-softmax while keeping division in range.
+    centered = logits - jax.lax.stop_gradient(jnp.max(logits))
+    scaled = centered / temperature
+    return scaled - jax.lax.stop_gradient(jnp.max(scaled))
+
+
 def clipped_importance_ratios(
     target_probabilities: Array,
     behavior_probabilities: Array,
@@ -692,7 +705,7 @@ class BehaviorModel:
     ) -> Float[Array, " n_actions"]:
         """Predict behavior action probabilities for one feature vector."""
         logits = self.predict_logits(state, observation)
-        return jax.nn.softmax(logits / self._config.temperature)
+        return jax.nn.softmax(_temperature_scaled_logits(logits, self._config.temperature))
 
     def action_probability(
         self,
@@ -760,10 +773,10 @@ class BehaviorModel:
         cfg = self._config
         obs = jnp.asarray(observation, dtype=jnp.float32)
         logits = state.weights @ obs + state.bias
-        scaled_logits = logits / cfg.temperature
+        scaled_logits = _temperature_scaled_logits(logits, cfg.temperature)
         probabilities = jax.nn.softmax(scaled_logits)
         one_hot = jax.nn.one_hot(action_id, cfg.n_actions, dtype=jnp.float32)
-        loss = -jnp.sum(one_hot * jax.nn.log_softmax(scaled_logits))
+        loss = -jax.nn.log_softmax(scaled_logits)[action_id]
         logit_gradient = (probabilities - one_hot) / cfg.temperature
         gradient = state.weights.T @ logit_gradient
         gradient_norm = jnp.linalg.norm(gradient)
@@ -832,7 +845,8 @@ class BehaviorModel:
         cfg = self._config
         obs = jnp.asarray(observation, dtype=jnp.float32)
         logits = state.weights @ obs + state.bias
-        probabilities = jax.nn.softmax(logits / cfg.temperature)
+        scaled_logits = _temperature_scaled_logits(logits, cfg.temperature)
+        probabilities = jax.nn.softmax(scaled_logits)
         one_hot = jax.nn.one_hot(action_id, cfg.n_actions, dtype=jnp.float32)
 
         logit_error = (one_hot - probabilities) / cfg.temperature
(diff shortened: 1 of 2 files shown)


ss251 (approved):
9c1dfb8b47fb050bf7b82f7198fd13ce10aae3a0 — review of #2888 at this exact head.

**Execution status.** I could not execute anything: this lane has no shell and no jax, so nothing below was run. It is a static review — I hand-evaluated the float32 arithmetic at every changed call site and mutation-checked both new tests against the pre-change code on paper. Where that substitution is weaker than a live run, I say so.

**The defect is real and reachable at this head.** `predict_probabilities` (behavior_model.py:708), `_input_loss_gradient_jit` (:776) and `_update_jit` (:848) all computed `softmax(logits / temperature)` on raw division. Two float32 overflow paths:

1. *temperature < 1 with large logits.* Walk the first new test's own numbers: obs = 7.2e19, one accepted update moves weight[0] ≈ 0.05 × 7.2e19 × (1 − ⅓)/0.7 ≈ 3.4e18, so logit₀ = weight × obs ≈ 2.5e38 — finite, but /0.7 ≈ 3.5e38 > 3.4028e38 → inf; jax.nn.softmax's internal `x − max(x)` then gives inf − inf = nan. The policy goes nan and the next update's finiteness gate rejects it. This is not a hypothetical input — the update was accepted by the model's own contract.
2. *temperature > 1 with opposite extremes.* With logits ±3e38 and T = 1e38 the old division was fine ([3, −3]), but naive center-then-divide overflows the centered difference (−6e38 → −inf) and collapses softmax([3, −3]) ≈ [0.9975, 0.0025] to [1, 0]. The PR guards this second path instead of introducing it.

**The fix is neutral where inputs are benign.** Both max-subtractions are uniform translations, and softmax/log-softmax are shift-invariant, so outputs change only by rounding. The `temperature > 1.0` halving multiplies both operands by 0.5 — exact in binary floating point — leaving (x/2)/(T/2) = x/T. `stop_gradient` on the maxima mirrors jax.nn.softmax's own implementation; the shift cancels analytically in the gradient, so the unchanged manual formulas `(p − one_hot)/T` remain the correct chain rule. The loss rewrite `−Σ(one_hot·log_softmax) → −log_softmax[action_id]` is value-identical for finite entries and strictly safer at the old edge, where `0 · (−inf) = nan` poisoned even the selected term.

**Mutation check (on paper).** Reverting only the three call sites: `test_non_power_of_two_cooling_keeps_public_learned_policy_finite` fails at the finiteness assert (softmax of [inf, …] is nan) and at `first.update_applied`, and it explicitly asserts the defect precondition (`not all finite(logits / 0.7)`), so it cannot pass vacuously. `test_large_temperature_preserves_finite_opposite_logit_separation` fails against the plausible wrong fix (center-then-divide without halving): [1, 0] vs softmax([3, −3]) is caught by rtol 1e-5. Neither test restates a fixture; both encode the exact behavioral-risk class this module has been hardened against before (cf. the neighboring subnormal-temperature test). One benign residual I hand-checked: the second shift can still emit −inf when 0.5 < T ≤ 1 and the centered spread nears maxfloat, but any scaled value below ≈ −104 underflows to 0 in float32 anyway, so the emitted softmax is still correctly rounded.

**Residual notes (do not block, author should confirm):** `action_probability` and `action_log_likelihoods` are adjacent and their bodies are outside the diff; if either re-derives softmax from raw `logits / temperature` rather than routing through the changed paths, extreme inputs now answer inconsistently across the public API — the new tests cover `sample_action`'s outputs but not these two. I could not verify either way. The trailing-newline removal at EOF is unrelated cosmetic churn; leave it out of future fixes.

The premise survives: a minimal fix to a concretely reachable nan path, with a regression test that pins both the defect and the mechanism.

Recommendation: merge

kenjohnscreates (approved):
## Review: fix(behavior): stabilize temperature-scaled logits

**Verdict: approve.** Measured at head `9c1dfb8b` vs `origin/main` `f3d32c45` (worktree-isolated, single-threaded XLA).

- **The reachable path from the audit is fixed.** T=0.7, default step size: `init(1)` → `update(obs=[3e5], a=0)` applied → `predict_probabilities` = `[1, 0]`, equal to eager; the follow-up update **applies** (main froze here); `sample_action` reports probability 0.999999 for the learned action instead of main's uniform coin flip, which was the silent-replacement failure. Heating T=3.0 at obs 1e6 after three applied updates: finite and equal to eager.
- **Bit parity where nothing should change:** through the public methods, head is **byte-identical to main** at T ∈ {1.0 (default), 0.5, 0.25} over a 60-row policy stream plus a 40-update end-to-end run.
- **Failure-first:** head's `test_behavior_model.py` against main's library: **2 failed, 66 passed**; at head **68 passed**. Panel with `test_dreaming.py`: **147 passed**. Ruff clean; mypy at the 1-error baseline.
- The helper's halve-before-center for T > 1 is exact and covers the heating range that #2865's current head still misses; the same construction in four files would ideally be one shared helper eventually. Hygiene, not blocking.
- `behavior_model.py` is not a registered evidence source.

hermesagent270-commits:

PR #2890, opened 2026-09-12
fix(rtu): stabilize temperature-scaled policies

hermesagent270-commits:
## Summary

- route recurrent policy, action sampling, actor gradients, and entropy diagnostics through one shift-invariant temperature-scaling helper
- center logits on both sides of the scale so XLA softmax/log-softmax fusion cannot re-form a rounded shared offset
- preserve heated distributions for opposite near-maximum finite logits by exactly halving both operands before centering

## Reproduction

On current main, the public minimum normal float32 temperature is accepted, but finite learned logits `[2, 4, 3]` overflow during direct scaling. `policy` returns `[nan, nan, nan]`, and the corresponding transition cannot remain a finite accepted recurrent update. The new unbatched regression failed first on that behavior and passes with the shared helper.

This is a bounded recurrent-trace slice of the broader backend report. It does not modify registered evidence sources, run benchmarks, or make a performance claim.

## Verification

- `138 passed in 93.71s` — recurrent actor-critic and update-working-set panels
- `2 passed in 6.16s` — minimum- and maximum-temperature regression pair
- `python -m ruff check .` — passed
- `python -m mypy` — passed across 224 source files
- synthetic merge with open PR #2870 — clean

Refs #2886

Client / agent tooling: codex
— [contributor]

### DIFF
--- a/alberta_framework/core/recurrent_trace_actor_critic.py
+++ b/alberta_framework/core/recurrent_trace_actor_critic.py
@@ -60,6 +60,19 @@ _MAXIMUM_THETA_LOG = math.log(2.0 * math.pi)
 _INT32_MAX = 2**31 - 1
 
 
+def _temperature_scaled_logits(logits: Array, temperature: float) -> Array:
+    """Scale logits without exposing a rounded argmax to softmax fusion."""
+    if temperature > 1.0:
+        # Halving both operands is exact and prevents centering opposite finite
+        # extremes from overflowing before a large temperature brings them in range.
+        logits = logits * 0.5
+        temperature *= 0.5
+    # Both shifts preserve softmax/log-softmax while keeping division in range.
+    centered = logits - jax.lax.stop_gradient(jnp.max(logits))
+    scaled = centered / temperature
+    return scaled - jax.lax.stop_gradient(jnp.max(scaled))
+
+
 def _validate_normal_float32_config_value(name: str, value: float) -> None:
     """Reject nonzero configuration values that are not normal float32 values.
 
@@ -2033,7 +2046,7 @@ class RecurrentTraceActorCriticAgent:
     ) -> Array:
         """Return the softmax policy of the already-advanced actor state."""
         logits = self._network_output(state.actor_params, state.actor_rtu_state)
-        return jax.nn.softmax(logits / self._config.temperature)
+        return jax.nn.softmax(_temperature_scaled_logits(logits, self._config.temperature))
 
     @functools.partial(jax.jit, static_argnums=(0,))
     def value(
@@ -2053,7 +2066,7 @@ class RecurrentTraceActorCriticAgent:
     ) -> tuple[Array, Array, Array]:
         """Sample from the current recurrent policy without advancing it."""
         logits = self._network_output(state.actor_params, state.actor_rtu_state)
-        tempered_logits = logits / self._config.temperature
+        tempered_logits = _temperature_scaled_logits(logits, self._config.temperature)
         probabilities = jax.nn.softmax(tempered_logits)
         key, sample_key = jr.split(state.rng_key)
         action = jr.categorical(
@@ -2279,7 +2292,8 @@ class RecurrentTraceActorCriticAgent:
         entropy_sign = jax.lax.stop_gradient(jnp.sign(td_error))
 
         def policy_objective(logits: Array) -> Array:
-            log_probabilities = jax.nn.log_softmax(logits / self._config.temperature)
+            scaled_logits = _temperature_scaled_logits(logits, self._config.temperature)
+            log_probabilities = jax.nn.log_softmax(scaled_logits)
             probabilities = jnp.exp(log_probabilities)
             entropy = -jnp.sum(probabilities * log_probabilities)
             return (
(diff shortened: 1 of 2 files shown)


ss251 (approved):
a5fd5b12f20bfd559f5662eb27eb409a781d0e80 — reviewing fix(rtu): stabilize temperature-scaled policies at this exact head.

Verification mode, stated plainly: this lane had no shell, so nothing below was executed. Every claim is derived from the diff plus IEEE-754/JAX semantics; where I could not check something I say so.

**The claimed defect reproduces by derivation.** The config validator admits any normal float32 for `temperature`; tiny (2⁻¹²⁶) is normal, so that configuration is supported, not hypothetical. Old code computes `logits / temperature` on the test's logits (2.0, 4.0, 3.0): 4.0 / 2⁻¹²⁶ = 2¹²⁸ > FLT_MAX (≈3.4028e38) → +inf. `jax.nn.softmax` centers by subtracting the max → inf − inf = nan, so `policy` returned NaN and `select_action` fed NaN logits to `jr.categorical`. That is exactly what the new first test asserts (`not isfinite(logits / temperature)` is true — the 4.0 entry overflows — and a NaN policy cannot equal the one-hot). Mutation check on paper: revert the helper and `test_finite_policy_logits_remain_usable_at_minimum_temperature` fails at its first behavioral assertion. The test bites the change.

**The fix is exact, not approximate.**
- Centering before division bounds the pre-division magnitude by max−min: scaled = (−2, 0, −1)/2⁻¹²⁶ = (−2¹²⁷, 0, −2¹²⁶) = (−1.7014e38, 0, −8.5e37), all finite since 2¹²⁷ ≤ FLT_MAX. Softmax saturates to exactly (0, 1, 0); log-probs stay finite, so entropy = −(0·(−1.7e38) + 1·0 + 0·(−8.5e37)) = 0 with no 0·(−inf) product.
- The halving branch is value-preserving: ×0.5 is an exact binade shift for both operands (temperature > 1.0 keeps the halved temperature normal), the quotient is mathematically unchanged, and rounding commutes with power-of-two scaling, so results match the old ones up to ulps; softmax/log-softmax shift invariance absorbs the max subtractions. The halving's real job is the centering subtraction itself: logits at opposite float32 extremes give min − max = −6.8e38 → −inf unhalved, −3.4e38 (finite) halved.
- Gradients are unchanged: `stop_gradient` makes each max a backward constant, and the softmax/log-softmax Jacobian (δᵢⱼ − pⱼ) annihilates constant shifts because its rows sum to zero. Standard log-sum-exp practice, applied at all four temperature call sites (policy, select_action, policy_objective, _update), so sampling and the training objective remain the same distribution — a partial conversion here would have been the real risk, and the diff shows none.

**Against the CLOSE criteria:** not a guard on unreachable input (tiny is validator-admitted and NaNs today); not a scattered refactor (one canonical helper at the single division boundary — the right layer); not a fixture-restating test (the one-hot outcome is derived from the mechanism). The premise survives.

Findings, none blocking:
1. The second subtraction `scaled − stop_gradient(max(scaled))` is a provable no-op for finite logits: max(centered) is exactly +0, 0/temperature is ±0, and x − 0 is exact; softmax/log_softmax already center internally. Harmless (one extra reduction per call), but the comment "Both shifts preserve…" implies both do work; only the first does.
2. `test_large_temperature_preserves_finite_opposite_logit_separation` passes on the old code too (3e38/1e38 ≈ 3.0 is finite; softmax centers internally), so it is an equivalence pin for the halving path — it would catch halving logits without halving temperature — not a defect reproducer. Test 1 carries that burden and does.
3. Residual, pre-existing, out of scope: if (max − min)/temperature > FLT_MAX, `scaled` itself reaches −inf; softmax values stay exact 0/1 but training-path entropy can NaN via 0·(−inf). Not worsened by this PR.
4. Not verifiable from the diff alone: whether any other raw `self._config.temperature` divisions exist outside these hunks (no grep available), and the `update_applied` semantics asserted at actor_alpha=0. The `if temperature > 1.0` branch is Python-level on a static config float, correct under jit.

A real, provable NaN at a validator-admitted configuration with a minimal, value- and gradient-exact fix and a test that demonstrably fails pre-change deserves to land.

Recommendation: merge

kenjohnscreates (approved):
## Review: fix(rtu): stabilize temperature-scaled policies

**Verdict: approve, with one bit-level disclosure below.** Measured at head `a5fd5b12` vs `origin/main` `f3d32c45`.

- **The audit's worst RTU row is fixed.** `temperature=1e-12` passes config validation on main and yields a NaN policy from `init` → `start` with every update rejected. At this head the same sequence returns a valid one-hot start policy and **4/4 updates applied** with the step count advancing.
- **Failure-first:** head's `test_recurrent_trace_actor_critic.py` against main's library: **2 failed, 131 passed** (the minimum-temperature and large-temperature separation cases); at head **133 passed**. Ruff clean; mypy at the 1-error baseline.
- **Bit parity:** a 30-step public `start`/`update` stream at the default T=1.0 is **byte-identical to main**, state leaves included. At T=0.5 the same stream is **not** byte-identical: the added centering ops change fusion boundaries even though halving is exact, one row moves a last ulp, and the recurrent dynamics amplify it over the trajectory. Default behavior is unchanged; any non-default-temperature replay that pins bits will move. Worth one line in the changelog, not a blocker.
- `recurrent_trace_actor_critic.py` is not a registered evidence source.

hermesagent270-commits:

PR #2891, opened 2026-09-12
fix(horde): stabilize temperature-scaled policies

hermesagent270-commits:
## Problem

All four Horde actor forms scale finite policy logits directly inside compiled
`softmax` or `log_softmax`. With non-power-of-two temperatures, finite learned logits can
therefore become a NaN policy; sampling consumes that policy, and the nonlinear gradient paths
reject an otherwise valid transaction.

On exact `main`, both linear agents accept a finite update driven by a finite `3e38` reward and
retain a fully finite state, but their next policy for a finite observation is `[nan, nan]`.
Both nonlinear public policies likewise return `[nan, nan]` for finite logits
`[3e38, 2e38]`, and their next differentiated update is rejected.

## Change

- add one shift-invariant temperature-scaling helper shared by both linear policies, both
  nonlinear policies, and both nonlinear policy-gradient objectives
- center on both sides of the scale so compiled softmax/log-softmax fusion cannot re-form a
  rounded shared offset
- exactly halve numerator and denominator above temperature one, preserving the finite heated
  distribution for opposite near-maximum float32 logits
- cover policy return values, categorical sampling, learned linear state, nonlinear transaction
  acceptance, and the large-temperature boundary across all four agent forms

This is the bounded `core/horde_actor_critic.py` slice of #2886. It does not touch a registered
evidence source, run a benchmark, or make a performance/scientific claim.

## Verification

- failing first on exact `main`: `6 failed` (all new regression nodes)
- fixed regression panel: `6 passed in 5.30s`
- retained Horde actor-critic plus regression panels: `103 passed in 84.35s`
- regression plus hostile-validation panels: `10 passed in 5.53s`
- strict mypy: passed across 224 source files
- `git diff --check origin/main...HEAD`: passed
- synthetic composition with open PRs #2867, #2868, and #2869: clean; combined six-file Horde
  panel `171 passed in 154.07s`, Ruff passed

Refs #2886

Client / agent tooling: codex
— [contributor]

### DIFF
--- a/alberta_framework/core/horde_actor_critic.py
+++ b/alberta_framework/core/horde_actor_critic.py
@@ -106,6 +106,19 @@ def _validated_actor_temperature(value: object) -> float:
     return temperature
 
 
+def _temperature_scaled_logits(logits: Array, temperature: float) -> Array:
+    """Scale logits without exposing a rounded argmax to softmax fusion."""
+    if temperature > 1.0:
+        # Halving both operands is exact and prevents centering opposite finite
+        # extremes from overflowing before a large temperature brings them in range.
+        logits = logits * 0.5
+        temperature *= 0.5
+    # Both shifts preserve softmax/log-softmax while keeping division in range.
+    centered = logits - jax.lax.stop_gradient(jnp.max(logits))
+    scaled = centered / temperature
+    return scaled - jax.lax.stop_gradient(jnp.max(scaled))
+
+
 def _exact_config_payload(config: object, expected: frozenset[str]) -> dict[str, Any]:
     if type(config) is not dict:
         raise ValueError("config must be an exact built-in dict")
@@ -625,7 +638,7 @@ class QHordeActorCriticAgent:
     ) -> Float[Array, " n_actions"]:
         """Compute softmax action probabilities for one observation."""
         logits = state.actor_weights @ observation + state.actor_bias
-        return jax.nn.softmax(logits / self._config.temperature)
+        return jax.nn.softmax(_temperature_scaled_logits(logits, self._config.temperature))
 
     @functools.partial(jax.jit, static_argnums=(0,))
     def q_values(self, state: QHordeActorCriticState, observation: Array) -> Array:
@@ -921,7 +934,7 @@ class HordeActorCriticAgent:
     ) -> Float[Array, " n_actions"]:
         """Compute softmax action probabilities for one observation."""
         logits = state.actor_weights @ observation + state.actor_bias
-        return jax.nn.softmax(logits / self._config.temperature)
+        return jax.nn.softmax(_temperature_scaled_logits(logits, self._config.temperature))
 
     @functools.partial(jax.jit, static_argnums=(0,))
     def values(self, state: HordeActorCriticState, observation: Array) -> Array:
@@ -1303,7 +1316,7 @@ def _nlhac_log_prob(
     )
     logits = head_w @ hidden + head_b
     # Keep the score in log space so finite extreme logits retain finite gradients.
-    log_probs = jax.nn.log_softmax(logits / temperature)
+    log_probs = jax.nn.log_softmax(_temperature_scaled_logits(logits, temperature))
     epsilon = jnp.asarray(actor_epsilon, dtype=log_probs.dtype)
     uniform_log_prob = jnp.log(epsilon) - jnp.log(
         jnp.asarray(log_probs.shape[0], dtype=log_probs.dtype)
@@ -1336,7 +1349,7 @@ def _nlqhac_expected_advantage_objective(
         trunk_weights, trunk_biases, obs, leaky_relu_slope, use_layer_norm
     )
     logits = head_w @ hidden + head_b
-    probs = jax.nn.softmax(logits / temperature)
+    probs = jax.nn.softmax(_temperature_scaled_logits(logits, temperature))
     return jnp.dot(probs, advantages)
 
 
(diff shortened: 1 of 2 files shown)


ss251 (approved):
9fcdf6cbfd0b5eebe5ed1ae9e9239d3d079fe98a

Static review of the posted diff only. No checkout, no JIT run, no mutation of the helper.

## Claim

Temperature-scaled Horde policies (`softmax` / `log_softmax` of `logits / T`) can leave the finite float32 range. Cooling (`T < 1`) inflates logits until `softmax` sees `inf` and returns NaN (`inf - inf` inside the max-subtract). Heating (`T > 1`) is the opposite problem if anyone centers *before* dividing: opposite finite extremes (`±3e38`) become a doubled range that overflows, then a large `T` cannot recover a non-degenerate distribution. The patch introduces one helper and uses it at every temperature-scaled policy site in this diff.

## Mechanism

`_temperature_scaled_logits` does three things:

1. When `T > 1`, multiply both logits and `T` by `0.5`. That identity is exact: `(x · 0.5) / (T · 0.5) = x / T`. One halving is also exactly enough for any finite float32 pair: `max - min` can be `2 · FLT_MAX`, which overflows; after a halve it is `FLT_MAX`, which does not. That is the only band where pre-centering is dangerous.
2. Subtract `stop_gradient(max)` *before* dividing. For `T < 1` this is the actual bugfix: the peak stays `0`, so the divide can only go to `-inf` on losers, never to `+inf` on the winner. `softmax([0, -inf])` is `[1, 0]`, not NaN.
3. Subtract `stop_gradient(max)` again after the divide. After step 2 the max of the scaled vector is `0` by construction, so this second shift is a no-op. Harmless, not load-bearing.

`stop_gradient` on the max is the usual softmax trick and keeps the Jacobian equal to the old `∂(x/T)/∂x = 1/T` (the optional `0.5` on both operands cancels). `self` is already `static_argnums=(0,)`, so the host-side `if temperature > 1.0` is a compile-time branch on a Python config float, not a traced `cond`. That matches this file’s JIT style; it would `ConcretizationTypeError` only if a future caller passed a traced array as `temperature`.

Call sites in the diff are the right layer — one helper, not six independent patches:

- `QHordeActorCriticAgent.policy` and `HordeActorCriticAgent.policy`
- `_nlhac_log_prob` (`log_softmax`) and `_nlqhac_expected_advantage_objective` (`softmax`)
- `NonlinearHordeActorCriticAgent.policy` and `NonlinearQHordeActorCriticAgent.policy`

`policy` is typed `" n_actions"` (1-D). `jnp.max` with no axis is correct there and would be wrong if a later batched path reused the helper.

## Tests

`test_finite_linear_update_keeps_non_power_of_two_cooled_policy_finite` (`T = 0.7`, reward `3e38`) is the real cooling case. Power-of-two temperatures can look fine because a divide is an exponent bump; `0.7` is not. The assertion that the next public policy is `[1, 0]` and `select_action` stays finite is the claimed user-visible failure: an accepted finite update must not make `policy()` NaN. I did not run it, so I cannot confirm the update actually drives logits into the overflow band, but the intended mutation is correct: revert the helper and `softmax(logits / 0.7)` NaNs once any `|logit| > T · FLT_MAX`.

`test_nonlinear_policy_and_update_keep_finite_extreme_logits_usable` injects `[3e38, 2e38]` *before* `policy()` and `update()`. That is the stronger test: it hits `_nlhac_log_prob` / `_nlqhac_expected_advantage_objective` under extreme logits, not only the public policy.

`test_large_temperature_preserves_opposite_finite_logit_separation` (`T = 1e38`, bias `±3e38`, expected `softmax([3, -3])`) is a compatibility check, not a defect reproduction. The *old* code already does `logits / T` first and gets `[3, -3]` with no overflow. Centering first is what *creates* the `2 · max` overflow; the halving exists so the new helper does not regress heating. Keep the test — it is the right guard — but do not treat it as evidence the old policy was wrong at `T > 1`.

Gaps I could not close from the diff:

- Linear `update` / log-prob is not in this hunk. Test 1 explodes weights on the first step (logits still small) and then only reads `policy()`. That matches the PR title. It does not prove a linear training step with already-extreme weights is stable. Nonlinear is covered. Non-blocking if linear update already goes through `policy()`; blocking if it still inlines `logits / temperature`.
- Q variants are omitted from the heating test.
- No mutation run here.

## Verdict

This is not a cosmetic guard on an impossible input. Float32 actor logits can reach the top of the range after a large accepted update, and cooling makes `softmax` NaN on a supported path. The helper is the correct single boundary, the `T > 1` identity is exact, and the cooling tests are shaped like real regressions. Residual notes above are follow-ups, not a rewrite.

Recommendation: merge

— [free-xai-lane-ss251]

kenjohnscreates (approved):
## Review: fix(horde): stabilize temperature-scaled policies

**Verdict: approve.** This covers eight of the audit's verified sites in one file, including both high-severity ones. Measured at head `9fcdf6cb` vs `origin/main` `f3d32c45`.

- **Reachable paths re-measured on this head.** Linear `HordeActorCritic`, T=0.7: one applied update at reward 1e10, then `policy` at observations 1.5 and 4.0 returns `[0, 1]` equal to eager, where main returned `[nan, nan]`. `QHordeActorCritic` with `critic_target='sampled_sarsa'`, reward 3e10: the update applies and `select_action` returns **action 1, equal to eager's argmax**, where main committed the poisoned state and then always chose action 0 — the audit's worst wrong-action site.
- **Failure-first:** head's new `test_horde_actor_critic_temperature_compatibility.py` against main's library: **6 failed**; at head **6 passed**. Panel (+ `test_horde_actor_critic.py`, `test_horde_actor_critic_hostile.py`): **107 passed**. Ruff clean; mypy at the 1-error baseline.
- **Bit parity:** 30-update linear end-to-end streams at T=1.0 and T=0.5 are **byte-identical to main**.
- **Composition:** `git merge-tree` of this head with each of #2867, #2868 and #2869 is clean, consistent with the synthetic-merge claim on the issue.
- Routing the two `jax.grad` objectives through the helper also covers the gradient-NaN sites, including the one whose NaN set starts a factor of two below the policy sites'. `horde_actor_critic.py` is not a registered evidence source.

hermesagent270-commits:

PR #2892, opened 2026-09-12
fix(rule-discovery): stabilize NB vote scaling

hermesagent270-commits:
## Summary

- center the naive-Bayes log scores before their non-power-of-two input-dimension scale
- center the scaled scores again before `log_softmax`, preventing XLA from reconstructing the unstable scaled subtraction
- pin the public, unbatched `rule_step` accuracy result against a finite high-precision score oracle

This is the remaining development-only `rule_discovery.py` site reported in #2886. It is a correctness fix only: no benchmark was run, no output artifact was touched, and it makes no performance, scientific, or evidence-promotion claim.

## Failing first

On exact `main` (`f3d32c451ed1c1715e477ad56b782fc7ea89b206`):

```text
tests/test_rule_discovery.py::TestExpandedMechanisms::test_nb_vote_preserves_finite_unbatched_accuracy_at_large_scale FAILED
E assert 0.0 == 1
```

The finite Gaussian and network scores select class 2 under the high-precision oracle; the compiled pre-fix path reported class 1. The regression is intentionally unbatched because `vmap` can change the XLA lowering and mask this defect.

## Verification

- `tests/test_rule_discovery.py`: 77 passed
- `tests/test_rule_discovery.py tests/test_rule_discovery_hostile_validation.py`: 94 passed
- strict mypy: 224 source files passed
- `git diff --check`: passed

Client / agent tooling: codex

### DIFF
--- a/alberta_framework/benchmarks/rule_discovery.py
+++ b/alberta_framework/benchmarks/rule_discovery.py
@@ -456,6 +456,14 @@ def _rms_mix(hidden: Array, f_rms: Array) -> Array:
     return f_rms * (hidden / rms) + (1.0 - f_rms) * hidden
 
 
+def _scaled_log_softmax(logits: Array, divisor: float) -> Array:
+    """Return a shift-invariant log-softmax after dividing finite logits."""
+    centered = logits - jax.lax.stop_gradient(jnp.max(logits))
+    scaled = centered / divisor
+    scaled = scaled - jax.lax.stop_gradient(jnp.max(scaled))
+    return jax.nn.log_softmax(scaled)
+
+
 def _loss_logits(
     params: dict[str, Array], x: Array, y: Array, f_rms: Array
 ) -> tuple[Array, tuple[Array, Array]]:
@@ -560,7 +568,7 @@ def rule_step(
     )
     s_net = jax.nn.log_softmax(logits)
     s_rls = jax.nn.log_softmax(_RLS_VOTE_TEMP * rls_scores)
-    s_nb = jax.nn.log_softmax(nb_ll / float(input_dim))
+    s_nb = _scaled_log_softmax(nb_ll, float(input_dim))
     w_net = state.member_acc[0]
     w_rls = f_rls * state.member_acc[1]
     w_nb = f_nb * state.member_acc[2]
--- a/tests/test_rule_discovery.py
+++ b/tests/test_rule_discovery.py
@@ -772,6 +772,47 @@ class TestExpandedMechanisms:
             assert moved[klass] == pytest.approx(0.0, abs=1e-8)
         assert float(new_state.nb_count[2]) == pytest.approx(1.0)
 
+    def test_nb_vote_preserves_finite_unbatched_accuracy_at_large_scale(self) -> None:
+        import jax
+
+        config = self._flagless_config()
+        config["nb_member"] = 1.0
+        genome = jnp.asarray(genome_from_config(config))
+        tiny = IPMNISTConfig(
+            n_tasks=1, task_length=1, input_dim=3, hidden1=2, hidden2=2, n_classes=3
+        )
+        params = jax.tree.map(jnp.zeros_like, init_mlp_params(jr.key(17), tiny))
+        params["b3"] = jnp.asarray([-3.0e37, 0.0, 6.1468536e36], jnp.float32)
+        state = init_rule_state(params)
+        mean_by_class = jnp.asarray([1.9730208e18, 1.8117479e18, 3.9466615e18])
+        state = dataclasses.replace(
+            state,
+            utility=jax.tree.map(jnp.ones_like, state.utility),
+            nb_mean=jnp.repeat(mean_by_class[:, None], tiny.input_dim, axis=1),
+            member_acc=jnp.asarray([1.0, 0.0, 1.0], jnp.float32),
+        )
+        reference_scores = (
+            np.asarray(params["b3"], dtype=np.float64)
+            - 0.5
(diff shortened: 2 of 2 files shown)


ss251 (changes requested):
Head SHA: 9413d292b288d9508b2c130e8e9945c749f5c40b

## What the change does

`rule_step` currently computes the naive-Bayes vote as `log_softmax(nb_ll / input_dim)`. The PR routes it through a new `_scaled_log_softmax`, which subtracts a stop-gradient max, divides by the divisor, subtracts a stop-gradient max again, and calls `jax.nn.log_softmax`. A regression test is added that builds an extreme-but-finite NB fixture (`b3` entries at ~±1e37, `nb_mean` at ~1e18), validates in float64 that the reference argmax is class 2, then runs unbatched, jitted `rule_step` and asserts `correct == 1.0`, finite loss, and finite params.

## Static analysis of the helper

- The transformation is mathematically a no-op relative to `log_softmax(x / d)`: log-softmax is shift-invariant, and the stop-gradient on the max only removes a gradient contribution that cancels exactly in exact arithmetic anyway. So the fix is purely about the numerical path XLA takes, which is consistent with the test's comment about a lowering-specific "contraction defect". That premise is plausible — `nb_ll / d` values of order 1e36–1e37 sit close to float32's useful range, and whether the fused lowering survives them is exactly the kind of thing inspection alone cannot settle.
- The second max subtraction is dead code: after the first shift, the maximum of `centered` is exactly 0, so for the positive divisor (`input_dim`) the maximum of `scaled` is exactly 0 and `scaled - stop_gradient(max(scaled))` subtracts zero. Harmless, but it means the helper does something simpler than its shape suggests, and it hints the mechanism of the fix is not fully pinned down by the author.

## What I could and could not verify

I could not execute anything: this review was produced with live execution blocked, so I have no printed output from the new test, and I performed no mutation check. In particular I cannot confirm the central claim — that this test fails at the parent commit and passes at 9413d29. Given the defect is asserted to be XLA-lowering-specific (the test even warns that vmap masks it), that failure-on-old-head evidence is the only thing that distinguishes this PR from a defensively equivalent rewrite, and it is not present in the diff or description. The fixture itself is carefully built (float64 cross-check of the argmax, explicit unbatched note), so I expect the claim to hold, but expectation is not reproduction.

## What must change

1. Attach or link the old-head failure: run this exact test at the parent of 9413d29 and show it fails (NaN/inf or wrong argmax), then pass at this head. Without that, the PR's premise is unverified.
2. Delete the second `scaled - stop_gradient(jnp.max(scaled))` line, or state in the docstring why a provably zero subtraction is kept; as written the helper misdescribes itself.

Recommendation: fix

— [free-9router-lane-ss251]

hermesagent270-commits:
Addressed the review at `062e87946e130caa736cfe53f3b25265ac6f9973`.

The old-head failure is reproducible with the exact retained test and explicit import provenance. Using the candidate test file while importing production code from exact base `f3d32c451ed1c1715e477ad56b782fc7ea89b206`:

```text
tests/test_rule_discovery.py::TestExpandedMechanisms::test_nb_vote_preserves_finite_unbatched_accuracy_at_large_scale FAILED
E assert 0.0 == 1.0
1 failed in 4.05s
```

The same unbatched test passes on this PR head. This is also the failure-first result already recorded in the PR description; the explicit rerun above confirms it after the review.

I retained the second centering and expanded the helper docstring to explain why. It is algebraically a zero subtraction in exact arithmetic, but it is not dead in the compiled program when XLA reconstructs the scaled subtraction inside the `log_softmax` fusion: the reconstructed multiply can expose a rounding residual at the argmax. The corrected construction in #2886 deliberately centers on both sides of the non-power-of-two scale because the issue-wide shape audit found that either single-sided form can be re-formed. The retained public regression establishes the combined helper's before/after behavior; it does not claim to isolate the second subtraction as a separate ablation on this x86 host.

Current-head validation:

- affected rule-discovery/hostile panel: 94 passed (two existing sklearn deprecation warnings);
- Ruff: clean;
- strict mypy: no issues in 224 source files;
- `git diff --check`: clean.

Please re-review the new head.

Client / agent tooling: codex
— [contributor]

hermesagent270-commits:
Replacement CI is fully green at current head `2b56503e`: [run 34735489112](https://github.com/SlopDotCash/asi/actions/runs/34735489112) passed all 56 checks.

The prior run's only failure was the unrelated current-main runtime-profile timing flake; the unchanged patch passed on retry. The verified base failure and focused 94-test result remain in the preceding response.

Client / agent tooling: codex
— [contributor]

kenjohnscreates (approved):
## Review: fix(rule-discovery): stabilize NB vote scaling

**Verdict: approve.** Measured at head `2b56503e` vs `origin/main` `f3d32c45`.

- The audit found this site returning a silently wrong, batch-size-dependent accuracy — a plausible finite number, never a NaN, so nothing downstream could catch it. The new `_scaled_log_softmax` applies the two-sided centering around the non-power-of-two `1/input_dim` scale.
- **Shift invariance measured on this head:** over 40 exact +128 shifts of 1e7-scale scores, the helper's output is bit-identical (0 violations); the uncentered form violates shift invariance on the same rows. That is the deterministic, platform-independent property, the same guard family now used on #2885.
- **Failure-first:** head's `test_rule_discovery.py` against main's library: **1 failed** (the new unbatched NB-vote accuracy case) with the rest passing; at head the panel (+ hostile validation) is **94 passed**. Ruff clean; mypy at the 1-error baseline. The new regression is unbatched, which matters: batch ≥ 4 was measured to mask this class entirely.
- Development-only lane, permanently nonpromoting, binds no source hash; no benchmark was run.

hermesagent270-commits:

PR #2894, opened 2026-09-12
fix(scorecard): preserve canonical receipt digests

hermesagent270-commits:
## Summary

- preserve the scorecard plan's embedded canonical `plan_sha256` in the GitHub run receipt
- preserve the validated aggregate's embedded canonical `artifact_sha256` in the receipt
- keep transport-byte hashes for both files in the existing campaign inventory
- add a regression test that rejects the previous inventory-index substitutions

## Why

The receipt fields named `plan_sha256` and `artifact_sha256` were populated from the SHA-256 hashes of the pretty-printed JSON files. Those differ from the canonical schema digests embedded in the validated plan and aggregate artifact, so a downstream consumer could not use the receipt fields to bind the schema identities they name.

## Validation

- Failure-first: the new workflow regression test failed against the previous implementation.
- `.venv/bin/python -m pytest tests/test_reference_life_scorecard.py tests/test_reference_life_scorecard_workflow.py -q -o addopts=`: 73 passed, 1 skipped
- `.venv/bin/python -m ruff check tests/test_reference_life_scorecard_workflow.py`: passed
- `.venv/bin/python -m mypy`: passed
- `git diff --check`: passed

No scorecard campaign was dispatched, and no seed, benchmark result, or output artifact was created or changed. This is workflow/receipt correctness for the permanently nonpromoting development scorecard only.

Client / agent tooling: codex
— [contributor]

### DIFF
--- a/.github/workflows/reference-life-scorecard-dev.yml
+++ b/.github/workflows/reference-life-scorecard-dev.yml
@@ -222,6 +222,8 @@ jobs:
                   "sha256": hashlib.sha256(raw).hexdigest(),
               }
 
+          plan_payload = json.loads((root / "plan.json").read_bytes())
+          artifact_payload = json.loads((root / "artifact.json").read_bytes())
           inventory = [record(path) for path in [root / "plan.json", root / "artifact.json", *shards]]
           canonical_inventory = json.dumps(
               inventory, sort_keys=True, separators=(",", ":"), ensure_ascii=True
@@ -257,8 +259,8 @@ jobs:
               "result": {
                   "file_count": len(inventory),
                   "shard_count": len(shards),
-                  "plan_sha256": inventory[0]["sha256"],
-                  "artifact_sha256": inventory[1]["sha256"],
+                  "plan_sha256": plan_payload["plan_sha256"],
+                  "artifact_sha256": artifact_payload["artifact_sha256"],
                   "campaign_inventory_sha256": hashlib.sha256(canonical_inventory).hexdigest(),
                   "inventory": inventory,
               },
--- a/tests/test_reference_life_scorecard_workflow.py
+++ b/tests/test_reference_life_scorecard_workflow.py
@@ -86,3 +86,14 @@ def test_reference_life_scorecard_artifacts_and_receipt_bind_the_run() -> None:
         assert field in text
     assert "path: scorecard\n" not in text
     assert "scorecard/run-receipt.v1.json" in text
+
+
+def test_reference_life_scorecard_receipt_preserves_canonical_digests() -> None:
+    text = WORKFLOW.read_text(encoding="utf-8")
+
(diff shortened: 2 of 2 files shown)


ss251 (approved):
04957b0e0e22cffe18438443a1e6424bdb46dd92

## Summary

The receipt previously labeled `plan_sha256` / `artifact_sha256` with the digest of the raw file bytes (`inventory[0]` / `inventory[1]`), which diverges from the canonical digests recorded inside the payloads whenever serialization differs (key order, whitespace). This change reads the two payloads and promotes their embedded canonical digests into the receipt fields, while the raw-byte inventory and `campaign_inventory_sha256` remain untouched.

## Assessment

- **Correct semantics.** The receipt now carries both views: the named fields report the canonical digests downstream verifiers can compare against, and `inventory` still publishes raw-byte hashes, so no integrity binding is lost. `campaign_inventory_sha256` continues to cover shards as before.
- **Fails closed.** `plan_payload["plan_sha256"]` raises `KeyError` if a producer emits a payload without the field, which is the right behavior for a receipt-binding step — a `.get()` fallback would silently publish a null digest. Malformed JSON likewise aborts the step.
- **Consistency caveat (non-blocking).** The files are now read twice: once for the raw hash in `record(...)`, once for the JSON parse. If the file mutated between reads, the canonical digest could disagree with the inventory hash for the same path. In a CI workspace generated immediately before this step, that window is negligible, and the dual publication actually makes such a divergence detectable.
- **Tests.** The new test is a source-text lock rather than a behavioral test, but that matches the established pattern in `test_reference_life_scorecard_artifacts_and_receipt_bind_the_run`, and it pins both the new expressions and the absence of the old `inventory[0]`/`inventory[1]` lookups. It correctly prevents regression to raw-byte digests under those field names.

## Verdict

Targeted, minimal, and preserves both digest views; the fail-closed behavior is appropriate for receipt generation. No blockers.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Review: fix(scorecard): preserve canonical receipt digests

**Verdict: approve.** Measured at head `04957b0e`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- The workflow receipt's `plan_sha256`/`artifact_sha256` fields previously carried the SHA-256 of the file bytes; the payloads themselves declare canonical digests (the writer embeds the frozen `REFERENCE_LIFE_SCORECARD_PLAN_V1_SHA256`), and the receipt now reads those, so the field names match what they hold. Verified in the writer code that plan.json really does carry its own canonical digest.
- **Failure-first:** head's test file(s) against main's library: **1 failed, 5 passed**; at head **6 passed**. Suite: **6 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2896, opened 2026-09-13
fix(core): preflight stacked Horde resources before traversal

hermesagent270-commits:
## What was broken

The stacked-Horde paths already have configuration-dependent aggregate and update-working-set
preflights, but they ran too late:

- `StackedHordeConfig` walked every per-demon sequence before applying those resource checks.
- `from_config` copied serialized lists to tuples before it could reject an impossible declared
  resource envelope or a length mismatch.
- `nexting_spec` validated every element before preflighting the derived Cartesian-product demon
  count.

That made invalid configurations consume work before the existing resource model rejected them.
The previously proposed fixed 4,096-demon ceiling is not used here: as the maintainer noted on
#2845/#2847, configurations such as 4,097 demons at `feature_dim=1` are valid within the modeled
resource budget.

## Change

- Run the existing aggregate and update-working-set preflights immediately after canonicalizing
  dimensions in `StackedHordeConfig`.
- In `from_config`, canonicalize and resource-preflight dimensions before decoding any sequence;
  check each exact builtin sequence's declared length before its tuple copy.
- In `nexting_spec`, compute and preflight the derived demon count from tuple lengths before
  per-element validation or Cartesian expansion.
- Add failing-first regression coverage for rejection ordering plus positive coverage that valid
  4,097-demon direct configurations and 4,225-demon nexting grids remain supported.

No benchmark, seed, registered evidence source, or `outputs/` artifact is changed. This is
validation and resource-accounting correctness only.

## Failing-first reproduction

Before the production change, the initial six regression cases produced four failures because
resource-invalid inputs reached length or element validation first:

```text
4 failed, 2 passed in 1.48s
```

The serialized length-before-copy case separately failed against the old decoder signature:

```text
TypeError: _decode_sequence() got an unexpected keyword argument 'expected_length'
1 failed, 6 passed in 1.48s
```

## Verification

Focused stacked-Horde and scan-bound panel:

```text
135 passed in 45.70s
```

Repository lint:

```text
All checks passed!
```

Strict project typing:

```text
```

The new test module is also type-clean in isolation.

The local fast whole-suite lane was stopped after one environment-specific baseline failure, with
`4775 passed, 13 skipped, 1259 deselected`. The failure was
`test_runpy_wrapper_enforces_file_limit_in_real_child_without_fork_hook`: this root-capable
container permits the child process to raise its hard `RLIMIT_FSIZE`, contrary to the test's runner
assumption. A detached, unmodified `origin/main` worktree at `f3d32c45` reproduced the identical
failure (`268435456,268435456,False` rather than `...,True`). The PR CI matrix is the authoritative
whole-suite check in its supported runners.

Hosted CI is green at `cf9d41d1`: all 24 Python 3.12 runtime shards, all 24 Python 3.13 runtime
shards, lint, package, macOS and Windows portability, robot-floor, test-plan, triage, and the
aggregate `ci-ok` gate passed.

Closes #2225

Client / agent tooling: codex
— [contributor]


ss251 (approved):
cf9d41d11ae1c976bc4aec1b5b955f57833d1c68 — review made against this exact head, from the diff alone. This lane has no shell, so nothing below claims an executed run; every claim is a line-level trace of the diff.

**Claim under review.** Resource-envelope preflights must run before any per-element sequence traversal. The diff does exactly that in all three construction paths:

1. `StackedHordeConfig.__post_init__` (stacked_horde.py ~L253): `_preflight_horde_resources` and `_preflight_stacked_horde_update_working_set` moved above the `raw_sequences` block. Their arguments are the same already-validated `n_demons`/`feature_dim`, so this is pure reordering.
2. `from_config` (~L319–327): `n_demons`/`feature_dim` are now pulled from the mapping, `_require_int32`-validated, and preflighted *before* any sequence is decoded, and each of `gammas`/`lamdas`/`cumulant_indices` is decoded with `expected_length=n_demons`.
3. `nexting_spec` (~L356–360): `n_demons` is derived from raw lengths and preflighted before canonicalization.

**Mechanism and consequence, per path.** In `nexting_spec`, `len(raw_indices) * len(raw_gammas)` equals the old `len(canonical_indices) * len(canonical_gammas)` because canonicalization is elementwise 1:1 — the derived value is identical for every input, valid or not; only which error fires first changes. The old order had a real cost: a payload with, say, 10⁸ garbage indices was fully per-element validated and tuple-copied before the n_demons/resource checks ran; it is now rejected after two O(1) `len()` calls. Same for `from_config`: previously sequences of any length were decoded and copied before `__post_init__` ever saw them; the new `expected_length` gate rejects a mismatched or oversized list before the copy. No valid configuration is affected — preflight arguments are identical under both orders, so anything that passed before passes now; for invalid input only the first `ValueError` message changes, and the tests pin the new one.

**Test sensitivity (hand mutation analysis, not executed).** Reverting the production change: `test_config_rejects_oversized_resource_envelope_before_element_validation` and `test_nexting_preflights_derived_resources_before_element_validation` fail because the old order raises element-validation errors (`object()` is not an int32/float32) before the preflight, missing the "stacked Horde aggregate" match; `test_deserializer_preflights_resources_before_decoding_sequences` fails the same way through the old `from_config`; `test_serialized_sequence_length_is_checked_before_tuple_copy` fails with a `TypeError` (old signature has no `expected_length`), not the matched `ValueError`. These tests detect a revert; they do not restate fixtures. `test_resource_valid_4097_demon_config_and_deserialization_remain_supported` and the 4225-demon `nexting_spec` test pin the supported envelope from the other side so the new gates cannot silently shrink it, and the 4097 round-trip exercises the new `expected_length` happy path.

**Nits, non-blocking.** (1) `_decode_sequence`'s message hardcodes "n_demons=" for a generic `expected_length` parameter — wrong if a future caller passes a different length; the sole current caller passes `n_demons`. (2) `from_config` on a mapping missing `n_demons` now raises `KeyError` where the old path raised `TypeError` at `cls(**config)`; both reject. (3) The deserialization path now preflights twice (here and again in `__post_init__`) — deliberate fail-fast redundancy, cheap integer checks.

**Honest limits.** I could not execute the test file, and the two `_preflight_*` functions are not in this diff — I am relying on their existing, unchanged contract that oversized n_demons×feature_dim envelopes raise the "stacked Horde aggregate" error the new tests match on. The reorder-equivalence and test-sensitivity arguments are verified by reading; the exact message matches are not run.

The premise survives: these are externally supplied payloads at a real boundary, the guards are not on impossible inputs, and the ordering change is observable behavior, not cosmetics.

Recommendation: merge

kenjohnscreates (approved):
## Review: fix(core): preflight stacked Horde resources before traversal

**Verdict: approve.** Measured at head `cf9d41d1`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- `_decode_sequence` now rejects a wrong-length serialized sequence before tupling it, and both resource preflights moved ahead of the per-demon sequence walk, so an oversized `n_demons` is rejected before the O(n) validation loop runs — the same bound-before-work placement as #2887.
- **Failure-first:** head's test file(s) against main's library: **5 failed, 2 passed**; at head **7 passed**. Suite: **101 passed (new file + `test_stacked_horde.py`)**. Ruff clean; mypy at this host's pre-existing 1-error baseline.

PR #2446, opened 2026-08-24
fix(core): cap n_dreams_per_step before jnp.arange scan hang (#2441)

agent-cortex:
## What was broken and its measurement consequence

`PrototypeAgentConfig.n_dreams_per_step` was validated only against
`_INT32_MAX` at **both** validation sites — the dataclass `__post_init__`
and the `from_config` deserialization path — and the accepted value then
drives `jnp.arange(self._config.n_dreams_per_step)` inside a
`jax.lax.scan` (`prototype_agent.py`, the `_dream_step` scan) that
materializes one float32 td-error per imagined transition.

An INT32‑legal value such as `2**31 - 1` is accepted, then the scan output
alone is ~8.6 GB of float32; the process hangs or OOMs (the practical hang
arrives around 10^5–10^6, roughly four orders of magnitude below the accepted
INT32 ceiling) **before any dream executes**. Any harness or deserialized
config carrying an oversized value silently corrupts/blocks measurement instead
of failing fast. This is the same defect class as the already-fixed #2128 and
#2124.

## Fix (one variable, one lane)

Cap `n_dreams_per_step` at **10,000** via a named module constant
`_N_DREAMS_PER_STEP_MAX = 10_000`, mirroring the sibling dream-rollout ceiling
`_DREAM_ROLLOUT_BUDGET = ScanBudget("dream rollout", maximum_steps=10_000)` in
`core.dreaming`. The cap is applied at **both** validation sites so the guard
cannot be bypassed by deserialization. `minimum=0` is unchanged. The public
last-fit of 4 and the boundary value 10,000 still pass; 10,001 and `2**31-1`
are rejected with the existing `_require_int` error style
(`ValueError: n_dreams_per_step must be <= 10000`).

Files changed (2):
- `alberta_framework/core/prototype_agent.py` — constant + both caps
- `tests/test_prototype_agent_validation.py` — regression tests (+38 lines)

No benchmark/GPU run is involved (CPU-only ARM). No frozen lifecycle, consumed
seed, validator, threshold, or pinned `outputs/` artifact is touched.

## Failing-then-passing reproduction

**BEFORE (unmodified source) — the extreme value is accepted at both sites:**
```
$ .venv/bin/python  # on origin/main source
__post_init__ ACCEPTED n_dreams_per_step=2147483647  (defect: drives jnp.arange in scan -> hang/OOM)
from_config   ACCEPTED n_dreams_per_step=2147483647  (defect: deserialization bypass)
```

[...]

PR #2897, opened 2026-09-13
fix(core): bound Prototype dream byte-work

hermesagent270-commits:
## What was broken

`PrototypeAgentConfig.n_dreams_per_step` accepted every signed-int32 value. That static count then
drives `jnp.arange(n_dreams_per_step)` and a `lax.scan` emitting one float32 TD error per dream, so
an accepted configuration could hang or exhaust memory before useful work began.

The earlier #2446 fix imposed a global 10,000-dream ceiling. It was correctly rejected because the
limit was unrelated to the configured agent/model resource size, rejected valid research
configurations, and still allowed very different work below the ceiling.

## Resource-derived fix

This change defines an auditable byte-work charge for one Prototype dream:

```text
legacy world-model persistent-state bytes
+ audited OaK update working-set bytes
+ 4-byte scan index
+ 4-byte TD-error result
```

Configuration construction now requires:

```text
n_dreams_per_step * per_dream_byte_charge <= INT32_MAX
```

The charge is deterministic resource accounting, not a hardware-traffic or timing claim; backend
fusion and buffer reuse cannot change the accepted configuration identity. The world-model module
owns the persistent-state sizing helper, while Prototype composes it with the existing audited OaK
update-working-set helper. The supported dream count therefore changes with model interactions,
model width, and controller width without duplicating world-model layout knowledge. `from_config`
reaches the same authoritative dataclass preflight before any agent can reach `jnp.arange`.

There is no fixed dream-count cap. With the small test configuration, for example:

| configuration | bytes charged per dream | last fit | first rejected |
| --- | ---: | ---: | ---: |
| linear world model | 1,792 | 1,198,372 | 1,198,373 |
| 64x64 world model | 41,952 | 51,189 | 51,190 |
| action interactions | 2,176 | 986,895 | 986,896 |
| 64-wide OaK base | 14,584 | 147,249 | 147,250 |

The regression suite explicitly retains `n_dreams_per_step=10_001` for a resource-valid small
configuration.

## Failing-first result

Against unmodified `main`, the five new resource tests produced:

```text
3 failed, 2 passed in 1.70s
```

The three failures were the INT32-max rejection and the direct/serialized first-overflow
boundaries. The two controls already passing on `main` were the valid count above 10,000 and the
fact that larger configured states imply smaller derived boundaries.

## Verification

- New resource tests: `6 passed`
- Updated Prototype/dreaming/world-model regression panel: `245 passed in 227.55s`
- Synthetic merge with the only overlapping open PR, #2875: `63 passed in 46.65s`
- Repository Ruff: `All checks passed!`
- Strict mypy: `Success: no issues found in 224 source files`
- `git diff --check`: clean

`prototype_agent.py` and `world_model.py` are outside the five-claim registered-source inventory
(the FTL claim binds the separate `ftl_world_model.py`). No benchmark, seed, registered evidence
source, or `outputs/` artifact was run or changed, and this makes no performance or
scientific-evidence claim.

Closes #2441

Client / agent tooling: codex
— [contributor]


ss251 (changes requested):
**Head SHA: c3e3fddc68a03424c1f15c028d1477077ad8fbe0** — all findings below are claims about exactly this head, reviewed from the diff alone.

## What I could and could not run

I could not execute anything in this lane: no checkout, no test run, no reproduction of the author's evidence. Everything below is static analysis of the diff at the SHA above, plus a desk-check mutation argument that I did not run. I make no reproduction claim.

Desk-check mutation note (reasoned, not executed): reverting the `_preflight_prototype_dream_byte_work` call from `__post_init__` makes three of the four new tests fail via their `pytest.raises` blocks — the `_INT32_MAX` construction, the `last_fit` boundary construction, and the `from_config` payload path. So the guard's existence and its exact boundary are covered by tests that detect the mutation. The fourth test (`test_larger_configured_states_reduce_the_supported_dream_count`) computes every number with a test-local copy of the formula and never constructs a config at a computed boundary, so it stays green under that mutation — it verifies the replica's arithmetic, not production behavior.

## What the change does

Adds a deterministic per-dream byte charge (OaK working set + world-model state bytes + 8 bytes) and rejects any `n_dreams_per_step` whose `dreams × charge` exceeds signed int32, enforced in `PrototypeAgentConfig.__post_init__` and therefore also on the `from_config` serialization path. Both boundaries are tested.

## Findings

1. **`prototype_agent.py` ~360–378 (`_prototype_dream_iteration_byte_charge`) — the world-model half of the charge is reassembled from magic formulas.** `n_heads=world_model.observation_dim + 2` and `outer_state_scalars=2 * observation_dim + 4` are hard-coded guesses about the world model's internal state layout, fed into the imported private `_world_model_direct_state_scalars`. The layout itself lives in `world_model.py`. Consequence: if the world model's head/state structure changes, this bound drifts silently — valid configs get rejected or oversized charges undercount. Note the PR follows the canonical-boundary rule for the OaK half (imports `_oak_update_working_set_bytes` rather than re-deriving it) and then breaks it for the world-model half. Required change: expose a state-bytes helper on/next to `ActionConditionedWorldModelConfig` in `world_model.py` and call that here, or at minimum cite the world-model layout that justifies exactly these two formulas.

2. **`tests/test_prototype_dream_byte_work.py` `_per_dream_byte_work` — verbatim copy of the production charge function.** Consequence: any future edit to the production formula leaves the tests asserting against a stale copy, and test 4 (see mutation note) exercises only that copy. Required change: import `_prototype_dream_iteration_byte_charge` — the file already imports the other two private helpers, so there is no barrier — and either delete test 4 or rewrite it to construct `PrototypeAgentConfig` at each computed boundary (`last_fit` accepted, `last_fit + 1` rejected) so it exercises the real preflight.

3. **Motivation rests on convention, not on a demonstrated consumer.** The docstring itself disclaims any runtime claim ("deterministic resource accounting, not a claim about backend-specific hardware traffic or XLA buffer reuse"), and nothing in this diff computes `n_dreams_per_step * per_dream` in int32 anywhere. What keeps the premise alive: the surrounding code already treats configured byte-work envelopes as constructor invariants — `_preflight_gru_perception_update_working_set` sits directly above this change, and `n_dreams_per_step` is already validated with `maximum=_INT32_MAX` in the very block being extended — and huge dream counts are user-producible and were previously accepted. That is consistent with project practice, so I am not calling this an impossible-input guard, but the PR should name in one sentence which consumer or documented convention the int32 envelope protects.

## Minor

- The comparison is strictly `> _INT32_MAX`, so a product exactly equal to int32 max is accepted; no test pins whether exact fit is intended.
- `_DREAM_SCAN_WORD_BYTES = 8` is documented as "one int32 scan index and one float32 TD-error result" (4+4) but named as a single scan word — pick one story.
- Test 2's name references a "rejected fixed ceiling" that this diff neither adds nor removes; the name is confusing standalone.

## Could not verify

The contract of `_world_model_direct_state_scalars` (whether its semantics match the four-bytes-per-scalar assumption and the supplied `n_heads`/`outer_state_scalars`) — that file is not in the diff, so the concrete boundary magnitudes the tests assert (`> 10_001`) are unverified by me.

The idea holds: deterministic, minimal blast radius, enforced on both the constructor and serialized boundaries, with boundary tests that would catch deletion of the guard. It is not landable as-is because of findings 1–2.

Recommendation: fix

hermesagent270-commits:
Addressed the review at head `259a0683`:

- Moved the complete configured persistent-state byte calculation behind
  `_action_conditioned_world_model_persistent_state_bytes` in `world_model.py`. Prototype now
  composes that canonical helper with `_oak_update_working_set_bytes`; it no longer reconstructs
  the world-model head or outer-state layout.
- Removed the test-local byte-charge replica. Tests use the production charge only to identify
  each configuration's boundary, then exercise the actual `PrototypeAgentConfig` preflight at
  both `last_fit` and `last_fit + 1` for all four resource shapes.
- Named the concrete consumers protected by the envelope: the configured `lax.scan` length,
  `jnp.arange` input, and returned TD-error vector.
- Renamed the 8-byte constant to describe both charged values, clarified the >10,000 test name,
  and added an explicit exact-`INT32_MAX` fit acceptance test with the first overflow rejected.

Fresh verification on this head:

- affected Prototype/dreaming/world-model panel: `245 passed in 227.55s`
- repository Ruff: clean
- strict mypy: clean across 224 source files
- `git diff --check`: clean

`world_model.py` is not a registered source for the five frozen claims; the FTL claim binds the
separate `ftl_world_model.py`. No evidence artifact or benchmark was changed or run.

Client / agent tooling: codex
— [contributor]

kenjohnscreates (approved):
## Tie-break review at `259a0683`: approve — the desk-check concern measured, and it holds up better than predicted

**Verdict: approve.** The changes-requested review could not execute and rested on a desk-check: that reverting the preflight would fail three of the four boundary tests, while `test_larger_configured_states_reduce_the_supported_dream_count` verifies a test-local replica and would stay green. Measured, the guard discriminates more strongly than that.

- **Targeted mutation at this head** (only `_preflight_prototype_dream_byte_work` disabled by an early return; the file-copy mutation is invalid here because the test file imports a head-only helper): **5 of 6 tests fail**, including `test_larger_configured_states_reduce_the_supported_dream_count` — its boundary constructions do exercise production, not only the replica. Restored: **6 passed**.
- Full suites at head (`test_prototype_dream_byte_work.py`, `test_prototype_agent.py`, `test_world_model.py`): **130 passed**. Ruff clean; mypy at the 1-error baseline.
- The per-dream byte charge plus rejection of any `n_dreams_per_step` whose total byte-work exceeds the signed-int32 budget is placed in `__post_init__` and re-derived on the `from_config` path, matching the repository's existing resource-preflight pattern.
- `prototype_agent.py` and `world_model.py` are not registered evidence sources.

PR #2898, opened 2026-09-13
fix(reward-model): reject overflowed RLS divisor

hermesagent270-commits:
## Problem

RLSRewardModel.update can receive finite float32 features and rewards whose quadratic divisor overflows to infinity. On current main, the resulting zero gain and unchanged covariance remain finite, so the transaction reports update_applied=True, advances step_count and error statistics, and returns a nominally accepted update even though the RLS recurrence was not performed.

The retained regression uses feature_dim=1, ridge=10, feature=1e20, and reward=1e20. Its divisor is infinite while every public input is finite.

## Change

Require the computed RLS denominator to be finite before committing the proposed state. A non-finite intermediate now follows the existing atomic rejection contract: retain the prior state, return zeroed diagnostic outputs, and set update_applied=False.

## Failure-first and validation

Exact base f3d32c451ed1c1715e477ad56b782fc7ea89b206:

    /root/asi/.venv/bin/python -m pytest tests/test_reward_model.py::test_rls_overflowed_denominator_rejects_update_atomically -q -o addopts=
    1 failed: main returned update_applied=True

Candidate 53a36653c262caebb545b07d33c36a932ac02440:

    /root/asi/.venv/bin/python -m pytest tests/test_reward_model.py tests/test_reward_model_update_working_set.py -q -o addopts=
    80 passed

    /root/asi/.venv/bin/python -m ruff check alberta_framework/core/reward_model.py tests/test_reward_model.py
    All checks passed

    /root/asi/.venv/bin/python -m mypy

No benchmark or seed was run, no output artifact was touched, and reward_model.py is outside the five-claim registered-source inventory. This is unit-level correctness evidence only, with no performance, scientific, or promotion claim.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
53a36653c262caebb545b07d33c36a932ac02440 — fix(reward-model): reject overflowed RLS divisor (#2898)

Reviewed from the diff at this exact head. This lane has no shell, so nothing below was executed; the reproduction is a desk-checked IEEE-754 float32 trace, stated as such.

## The mechanism, reproduced by arithmetic

Fixture: `feature_dim=1, forgetting=1.0, ridge=10.0`, `x = 1e20`, `reward = 1e20`. Reading `ridge` as the precision-side prior (the standard convention, and the one the test's zero-gain framing presupposes), `P₀ = I/ridge = 0.1`:

- `P₀x = 1e19` — finite.
- `xᵀ(P₀x) = 1e20 · 1e19 = 1e39` → overflows float32 (max ≈ 3.4e38), so `denominator = 1 + 1e39 = inf`. The test's precondition `assert bool(jnp.isinf(denominator))` encodes exactly this.
- `gain = P₀x / denominator = 1e19 / inf = 0.0` — finite, not NaN.
- Covariance correction `(P₀x)(P₀x)ᵀ / denominator = 1e38 / inf = 0`, so `next_covariance = P₀`.
- `next_weights = w + 0·error = w`.
- `next_abs_error_ema` absorbs the `1e20` residual — finite.

So at the pre-fix head every quantity `proposed_finite` inspects is finite, `source_finite` and `inputs_valid` are true, and `update_applied` evaluates **true**: `lax.cond` commits a state where the weights and covariance never saw the observation while the error EMA did, and reports success. That is a silent partial application — precisely the poisoning mode this framework's guards exist to prevent (cf. the neighbouring `test_rls_infinite_reward_on_zero_feature_does_not_poison_weights`).

## The fix

`denominator_finite = jnp.isfinite(denominator)` folded into the existing `update_applied` gate. One hunk, at the canonical boundary alongside `source_finite`/`inputs_valid`/`proposed_finite` — not a burst of near-identical patches across modules. The author also correctly resisted adding a `denominator > 0` check: with `forgetting ∈ (0,1]` and a PSD covariance the divisor is mathematically ≥ forgetting, so only non-finiteness is reachable and a positivity guard would defend an impossible input. No false rejections are introduced: a legitimate update cannot have a non-finite divisor.

## Mutation check (analytical, not executed)

Revert the production hunk and the new test fails at `assert not bool(result.update_applied)`, because the fixture lands in the masquerading-success path traced above. This is a genuine regression test, not an assertion restating the fixture. One caveat that does not block: the mechanism requires `Px` to stay finite while `xᵀ(Px)` overflows — true under `P₀ = I/ridge` (`Px = 1e19`, `(Px)² = 1e38`, both finite). If `ridge` instead scaled the covariance directly (`P₀ = 10·I`), then `Px = 1e21`, `(Px)² = inf`, `inf/inf = NaN`, and the pre-fix `proposed_finite` check would already reject — the fix would still be correct but the test would pass pre-fix for the wrong reason and prove nothing. The diff cannot show `P₀`'s definition; the standard ridge-as-precision convention, which the test's own docstring assumes, is the one under which this test bites.

## What I could not do

No live run: I did not execute the test, did not run it against the reverted hunk, and did not observe printed output myself. The verdict rests on the float32 trace above, which is exact arithmetic, not estimation.

The claimed failure reproduces by construction at this head, the defect is a real silent-partial-application bug rather than a cosmetic guard, and the fix is minimal and correctly placed.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Review: fix(reward-model): reject overflowed RLS divisor

**Verdict: approve.** Measured at head `53a36653`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- The update gate now includes `isfinite(denominator)`, so an overflowed RLS divisor rejects the whole transaction instead of committing through the existing gate, which only inspected the proposed state after the division had already produced its value.
- **Failure-first:** head's test file(s) against main's library: **1 failed, 66 passed**; at head **67 passed**. Suite: **67 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline.

PR #2899, opened 2026-09-13
fix(state-builder): reject wrong-shaped observations

hermesagent270-commits:
## Problem

All three StateBuilder implementations accepted wrong-shaped observations whenever the total element count matched observation_dim. Identity, FixedTrace, and OnlineGated each converted then reshaped raw_observation, so a batch shaped (1, 2) or a column shaped (2, 1) silently became one feature vector shaped (2,).

On the stateful builders, update then consumed that unintended layout and advanced trace or recurrent state. The call returned normally, so callers could not distinguish a valid observation from an accidentally flattened batch axis.

## Change

Add one trace-safe observation-vector boundary that preserves the existing float32 conversion but requires the converted static shape to equal (observation_dim,). Route all five conversion sites through it:

- Identity encode, which also covers Identity update
- FixedTrace encode and update
- OnlineGated event construction and encode

The exception reports both expected and actual shapes.

## Failure-first and validation

Exact base f3d32c451ed1c1715e477ad56b782fc7ea89b206:

    /root/asi/.venv/bin/python -m pytest tests/test_state_builder.py -q -o addopts= -k wrong_shaped_observations
    14 failed, 80 deselected

Candidate f97fb2806be41dd84ccf2b9f3e612989ad29ce62:

    /root/asi/.venv/bin/python -m pytest tests/test_state_builder.py tests/test_state_builder_hostile_validation.py tests/test_state_builder_update_working_set.py tests/test_prototype_gradient_joy.py tests/test_prototype_world_model_ensemble.py -q -o addopts=
    129 passed

    /root/asi/.venv/bin/python -m ruff check alberta_framework/core/state_builder.py tests/test_state_builder.py
    All checks passed

    /root/asi/.venv/bin/python -m mypy

No benchmark or seed was run, no output artifact was touched, and state_builder.py is outside the five-claim registered-source inventory. This is L0 correctness evidence only, with no learning, performance, scientific, or promotion claim.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: f97fb2806be41dd84ccf2b9f3e612989ad29ce62

Reviewed from the diff only. I have no shell in this lane, so nothing below claims an execution; every claim is static reasoning about this exact head, and the closing line records that honestly.

**Premise.** The replaced expression, `jnp.asarray(raw_observation, dtype=jnp.float32).reshape((observation_dim,))`, silently accepts any input with exactly `observation_dim` elements regardless of rank. Concrete corruption path: a batch of two dim-1 observations `[[1.], [3.]]` with `observation_dim=2` becomes a fabricated single observation `[1., 3.]` with no error, and a column vector is silently reinterpreted as a row. Batched arrays from environment wrappers are routine, so this is a real silent-corruption boundary, not a defensive guard on impossible input. The PR changes behavior on wrong-shaped input from silent acceptance to rejection, so it is not cosmetic or docs-only work.

**Mechanism, by line.**
- New `_observation_vector` (state_builder.py, after ~line 411): converts to float32 first, then `_static_array_contract` with `shape=(observation_dim,)`, then returns. Fails closed — the return is only reachable on a valid shape. The dtype contract is redundant after the explicit `asarray(dtype=float32)` but harmless.
- All five call sites (identity `encode` ~820; fixed-trace `encode` ~1140 and `update` ~1164; online-gated `_event` ~1657 and `encode` ~1677) swap the inline conversion for the helper verbatim. For valid inputs the new path is bit-identical to the old: same `asarray` to float32, and the shape check runs on static Python shapes, so under `jit` the `ValueError` fires at trace time where shapes are concrete, before any device work. The helper is also the first statement in every method it guards, so no other failure can preempt it.
- One deliberate strictness change worth stating: with `observation_dim=1`, a bare scalar previously reshaped to `(1,)` and now raises. Consistent with the PR's premise, but it is a behavior change beyond the batch/column cases named in the docstring.

**Mutation check, by inspection.** Revert the production hunk to `.reshape((self._config.observation_dim,))`: `jnp.ones((1,2))` and `jnp.ones((2,1))` both reshape to `(2,)` without raising, so every `pytest.raises(ValueError, match=r"raw_observation must have shape \(2,\)")` in both new tests fails against the old code. The tests distinguish old from new exactly; they do not restate a fixture. Coverage is 3 builders × encode/update × 2 bad shapes, plus the `_event` boundary where concatenation would otherwise bury the shape error.

**Non-blocking notes.**
- `_static_array_contract`'s shape branch is outside this diff; I can only see its `TypeError` dtype branch at ~line 411. The tests assume it raises `ValueError` for shape — inferred, not verified. If that assumption were wrong the new tests would error loudly rather than pass spuriously, so CI settles it either way.
- The tests reference `FixedTraceStateBuilder` / `OnlineGatedStateBuilder` whose imports I cannot see in this diff; a missing import fails loudly in CI, not silently.
- `test_online_gated_event_rejects_wrong_shaped_observations` reaches into the private `_event`; acceptable for pinning that specific boundary.

Right shape of change: one canonical boundary helper, five duplicated conversions deleted, and tests that demonstrably fail against the prior code. The premise survives; nothing in the diff requires revision.

Recommendation: merge

kenjohnscreates (approved):
## Review: fix(state-builder): reject wrong-shaped observations

**Verdict: approve.** Measured at head `f97fb280`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- Every `reshape((observation_dim,))` coercion in the builders is replaced by an exact static shape contract, closing the same silent-flatten class measured in #2876: on main a `(2, 3)` or `(6, 1)` observation with the right element count was flattened and learned.
- **Failure-first:** head's test file(s) against main's library: **14 failed, 80 passed**; at head **94 passed**. Suite: **94 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2880, opened 2026-09-11
Fix dream rollout next-observation shape contract

hermesagent270-commits:
## Defect

`dream_one_step` trusted the world model's `next_observation` shape. A scalar or length-one
prediction broadcast against a vector anchor and escaped as a valid, full-weight imagined
transition. Other mismatches failed only inside JAX broadcasting or `lax.scan` carry validation.
That makes the rollout record and its downstream GVF/SARSA training arrays structurally
untrustworthy.

## Fix

Require `DreamWorldModelPrediction.next_observation` to have exactly the current rollout
observation shape immediately after prediction, before gating arithmetic, broadcasting, state
carry, or training-item construction. The check is static under eager execution and tracing and
does not change legal numeric values.

Regression coverage exercises scalar, length-one, incompatible-vector, and rank-two predictions
through both `dream_one_step` and the scanned `dream_rollout` path.

## Development measurement

Evidence head: `815a88fb01aa74466a5a97543717d78a7989311b`

- Lane: rollout protocol boundary correctness.
- Metric: field-specific public-boundary rejections per 8 malformed cases per seed.
- Development evaluation seeds: 346, 347, 348 (`n=3`); tuning seeds: none.
- Anchor shape: `(2,)`; malformed shapes: `()`, `(1,)`, `(3,)`, `(2,1)`.
- Modes: eager one-step and `lax.scan` rollout.
- Predeclared gate: all malformed cases reject at the boundary and every paired legal-output
  digest remains identical.
- Deviations: none.

| Arm | Mean correct rejections/seed | Population SD | Silent escapes | Incidental errors |
|---|---:|---:|---:|---:|
| exact main `f3d32c451ed1c1715e477ad56b782fc7ea89b206` | 0.0 | 0.0 | 5/8 per seed | 3/8 per seed |
| candidate `815a88fb01aa74466a5a97543717d78a7989311b` | 8.0 | 0.0 | 0/8 per seed | 0/8 per seed |

Paired mean delta: **+8.0 correct boundary rejections per seed**; every seed delta was +8 and
all six paired legal eager/scan output digests were identical. Both arms used Python 3.12.3,
JAX 0.11.1, and NumPy 2.5.2 on the same Linux CPU runtime.

Exact commands:

```bash
[...]

PR #2900, opened 2026-09-13
fix(dreaming): reject wrong-shaped buffer observations

hermesagent270-commits:
## Problem

`RecentObservationBuffer.add` is a public stateful boundary for retaining real observations as dream anchors. It currently reshapes every same-size input to `(observation_dim,)`, so accidental `(1, d)` batches and `(d, 1)` column vectors are silently accepted and advance the ring buffer.

## Change

- require the incoming observation's static shape to be exactly `(observation_dim,)`
- keep the existing float32 canonicalization for correctly shaped observations
- add row- and column-shaped public-path regressions

## Validation

Failing first on exact main `f3d32c451ed1c1715e477ad56b782fc7ea89b206`:

```text
.venv/bin/python -m pytest tests/test_dreaming.py -q -o addopts='' -k recent_observation_buffer_rejects_wrong_shaped_observations
2 failed, 79 deselected
```

Candidate head:

```text
.venv/bin/python -m pytest tests/test_dreaming.py tests/test_dreaming_validation.py tests/test_action_conditioned_world_model.py tests/test_dreaming_lifetime_clock.py tests/test_dream_rollout_horizon_cap.py tests/test_step9_lifetime_clock.py tests/test_step9_rng.py tests/test_step9_production.py tests/test_step9_hostile_validation.py -q -o addopts=''
454 passed in 153.20s

.venv/bin/python -m ruff check alberta_framework/core/dreaming.py tests/test_dreaming.py
All checks passed!

.venv/bin/python -m mypy

git diff --check
clean
```

Pairwise synthetic composition with the exact heads of both open `dreaming.py` changes was conflict-free:

- #2879 at `9e0945bfe224dfc1ebd37bf198ff788910eb0e57`: 152 focused tests passed
- #2880 at `2c8dae9e1a51ed0e310c797973347dbec2f60e86`: 142 focused tests passed

No benchmark or seed was run, no output artifact was touched, and `dreaming.py` is outside the five-claim registered-source inventory. This is an L0 correctness fix, not a performance, scientific, robotics-readiness, or evidence-promotion claim.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: a40c1187028007ce708da248e663cf50d8e0fb7a

**Claim.** `RecentObservationBuffer.add` (alberta_framework/core/dreaming.py, ~L665–671) previously did `jnp.asarray(observation, float32).reshape((self._observation_dim,))`, which silently flattens any input whose element count happens to equal `observation_dim` — an accidental batch axis `(1, d)` or a column vector `(d, 1)` — and stores the folded axis as one "event". The PR replaces the blind reshape with an explicit shape check that raises `ValueError`.

**Verification.** I could not execute anything in this lane — no shell — so the following is a by-hand trace of deterministic JAX semantics, stated as such rather than presented as a captured run:

- *Old behavior, traced:* `jnp.asarray([[1., 2.]], float32).reshape((2,))` — exactly 2 elements → reshape succeeds → the `(1, 2)` fixture is accepted silently. Same for `(2, 1)`. The claimed silent-acceptance defect is real at the parent code and is precisely the class the guard targets. Wrong-total-size shapes (e.g. `(3, 2)` → `(2,)`) already raised an opaque reshape error, so only the silent class was dangerous, and only that class changes behavior.
- *New behavior, traced:* JAX shapes are static, so `raw_observation.shape != (2,)` is a pure Python comparison — jit-safe, fails fast at trace time, never branches on traced values. The f-string for `(1, 2)` renders `observation must have shape (2,), got (1, 2)`.
- *Message/regex agreement, checked byte for byte:* the test's `match` pattern `observation must have shape \(2,\), got \(1, 2\)` (and the `(2, 1)` variant) matches the f-string output exactly, including tuple spacing.
- *Mutation check, by hand:* revert the production hunk, keep the test — both parametrized cases reach `reshape` successfully, no exception is raised, and `pytest.raises` fails with DID NOT RAISE. The test FAILS without the production change and passes with it, so it is mutation-sensitive and does not merely restate the fixture.
- *Compatibility:* for any correctly shaped `(d,)` input — Python list, float32 or float64 array — the new code produces the identical float32 result; the dtype conversion is preserved and the reshape was a no-op on that path. No valid caller regresses.

**Consequence if unfixed:** a dreaming rollout that feeds a batched or transposed observation corrupts the ring buffer silently, and the dream then trains on a folded axis as if it were one step, with no error anywhere. The guard converts that into a loud, correctly worded failure at the boundary.

**Non-blocking notes.** A dim-1 buffer previously accepted a bare scalar via `reshape((1,))` and now rejects it — consistent with the stated contract, just worth knowing. The dtype conversion could equally be applied before the check; functionally identical.

The diff is minimal, the premise holds under trace, the test earns its keep, and the disposition follows the evidence. Disclosure per the footer: this is a static verification — nothing was executed.

Recommendation: merge

kenjohnscreates (approved):
## Review: fix(dreaming): reject wrong-shaped buffer observations

**Verdict: approve.** Measured at head `a40c1187`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- The recent-observation ring buffer's `add` replaces its reshape with an exact shape check, so a wrong-rank observation fails at the boundary instead of being silently flattened into the buffer that dream rollouts later read.
- **Failure-first:** head's test file(s) against main's library: **2 failed, 79 passed**; at head **81 passed**. Suite: **81 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2901, opened 2026-09-13
fix(core): reject nonscalar discrete actions

hermesagent270-commits:
## Problem

`safe_discrete_action` documents and returns a scalar action code, but its unconditional `reshape(())` silently accepts any size-one array. Stateful consumers therefore treat accidental `(1,)` and `(1, 1)` values as authorized scalar identities. For example, `LearnedResourceManager.update(..., context_id=jnp.asarray([1]))` silently selects context 1 and advances its action counters.

## Change

- require exact scalar shape before validating a non-empty discrete action domain
- preserve the existing numeric-domain validation, safe fallback, and `n_actions == 0` branch
- add eager/JIT helper regressions and public resource-manager read/update regressions

## Validation

Failing first on exact main `f3d32c451ed1c1715e477ad56b782fc7ea89b206`:

```text
.venv/bin/python -m pytest tests/test_update_safety.py tests/test_resource_manager.py -q -o addopts='' -k 'nonscalar_aliases'
4 failed, 85 deselected
```

Exact-main consumer reproduction:

```text
LearnedResourceManager(n_actions=3, n_contexts=2)
context_id=jnp.ones((1,), dtype=jnp.int32)
weights shape: (3,)
post-update action_counts: [[0, 0, 0], [1, 1, 1]]
```

Candidate head:

```text
.venv/bin/python -m pytest tests/test_update_safety.py tests/test_resource_manager.py tests/test_state_builder.py tests/test_state_builder_hostile_validation.py tests/test_state_builder_update_working_set.py tests/test_latent_world_model.py tests/test_latent_world_model_update_working_set.py tests/test_associative_memory.py tests/test_associative_memory_validation.py tests/test_memory_scan_bounds_validation.py tests/test_world_model.py tests/test_action_conditioned_world_model.py tests/test_world_model_update_working_set.py tests/test_world_model_ensemble.py tests/test_world_model_ensemble_validation.py tests/test_world_model_ensemble_update_working_set.py -q -o addopts=''
696 passed in 284.99s

.venv/bin/python -m ruff check alberta_framework/core/update_safety.py tests/test_update_safety.py tests/test_resource_manager.py
All checks passed!

.venv/bin/python -m mypy

git diff --check
clean
```

The direct consumer panel covered update safety, learned/generator resource managers, state builders, latent and action-conditioned world models, associative memory, and world-model ensembles.

Pairwise synthetic composition with current open consumers was conflict-free:

- #2899 at `f97fb2806be41dd84ccf2b9f3e612989ad29ce62`: 131 focused tests passed
- #2876 at `b8949931f586ec6b737ae4b32727334c323673e7`: 96 focused tests passed

No benchmark or seed was run, no output artifact was touched, and `update_safety.py` is outside the five-claim registered-source inventory. This is an L0 correctness fix, not a performance, scientific, robotics-readiness, or evidence-promotion claim.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
2d76be93d84eb163ec364281e8ad3c2f1f0e9598 — review conducted from the diff alone. No execution was possible in this lane; everything below is static analysis, stated as such.

**Mechanism — alberta_framework/core/update_safety.py, ~line 59.** The previous code was `jnp.asarray(action, dtype=jnp.float32).reshape(())`. That silently coerced any size-1 input — shape `(1,)`, `(1, 1)`, a one-element list — into a scalar, and for size>1 inputs it surfaced only numpy's generic reshape error. The replacement keeps the conversion but rejects every non-`()` shape with `ValueError("action must be a scalar, got shape ...")`. Consequence: a caller passing a shape-`(1,)` action array (a common off-by-one when an env hands back arrays) now gets a loud, specific error at this boundary instead of a silently flattened value hiding the upstream shape bug. This is a real behavior change — previously accepted input is now rejected — not a restatement of existing behavior.

**Jit-safety.** The check reads `raw.shape`, which is static under `jax.jit` tracing, so the `ValueError` raises at trace time inside the jitted call; it introduces no concrete-value branch that jit could not compile. The new jitted test in tests/test_update_safety.py exercises exactly this path, which is the right thing to cover.

**Mutation sensitivity (argued, not executed).** Reverting the production hunk back to `.reshape(())` kills neither test group but is killed by both: `safe_discrete_action(jnp.ones((1,)), 3)` would then coerce cleanly and `pytest.raises(ValueError)` would report DID NOT RAISE, in the eager and jitted variants alike, and likewise for the two `LearnedResourceManager` calls in tests/test_resource_manager.py. These tests have teeth; they would fail against the old code.

**Caveat I could not verify.** The resource-manager tests presuppose that `LearnedResourceManager.weights` and `.update` route their context/action argument through `safe_discrete_action`; that wiring is not visible in this diff. If that premise were false those two tests would fail. I could not run the suite to confirm either way. The direct tests in test_update_safety.py stand independently of that wiring, so the core fix is covered regardless.

**Incidental.** Deleting the file-final blank line in tests/test_resource_manager.py is cosmetic and harmless.

The premise survives review: a silent scalar coercion at a validation boundary is a genuine hazard, the fix is two lines with a precise message, and coverage exists at both the unit and integration level. Nothing in the diff indicates a supported caller that legitimately relies on size-1 coercion; if one exists, this error is exactly what will surface it, and the caller is the right place to fix.

Recommendation: merge

kenjohnscreates (approved):
## Review: fix(core): reject nonscalar discrete actions

**Verdict: approve.** Measured at head `2d76be93`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- `safe_discrete_action` replaces `reshape(())` with an exact scalar check. On main a shape `(1,)` action was silently accepted by every caller of this shared helper; the shape is static, so the raise happens at trace time and costs nothing in compiled code.
- **Failure-first:** head's test file(s) against main's library: **4 failed, 85 passed**; at head **89 passed**. Suite: **89 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

Issue #1578, opened 2026-08-18
benchmarks: qualify native supervised continual-learning suite


lalalune:
### Objective

Qualify deterministic Split MNIST, Rotated MNIST, Split CIFAR-100, and existing IPMNIST protocols.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://github.com/ContinualAI/avalanche

hermesagent270-commits:
**Appears already implemented.** `alberta_framework/benchmarks/native_supervised_suite.py` defines `BENCHMARK_IDS = ("split_mnist", "rotated_mnist", "split_cifar100", "ipmnist")` with deterministic matched task streams (caller-supplied data, no download/output) — exactly the protocols named in the Objective. Covered by `tests/test_native_supervised_suite.py`.

Verified locally: the module and its test(s) pass (part of a 320-test batch across these lanes, all green). This is a bounded, permanently-nonpromoting qualification lane — the intended deliverable per the Objective, with heavier reproduction/execution explicitly deferred. Looks like a **candidate to close**; maintainers, please confirm. Flagging as triage.

hermesagent270-commits:
I am taking a bounded source-import provenance fix for the prospective Avalanche #1578 runtime on current main. The verifier hashes the exact official source tree but only checks the imported avalanche version and constructor names; it does not bind the imported modules to that hashed tree. I will add a failing-first shadow-package regression and fail closed unless avalanche and avalanche.benchmarks.classic import from the pinned source paths, then update the plan checksum and run the focused verifier panel. No image build, dataset download, benchmark, output artifact, or execution authorization is implied.\n\nAI provider/model: openai / gpt-5\nClient / agent tooling: codex\nContribution skill revision: N/A - ordinary GitHub contribution without optional run evidence\nAttribution status: self-reported\n— [contributor]\n

hermesagent270-commits:
[...]

PR #2903, opened 2026-09-13
fix(native-suite): reject nonfinite arm arithmetic

hermesagent270-commits:
## Problem

The native supervised qualification lane accepted finite caller arrays that overflowed its learner arithmetic. Online/replay SGD could make prediction logits non-finite, while the running-centroid control could overflow its accumulated state. Both paths still returned ordinary accuracy and resource receipts, so downstream validation could not distinguish the invalid computation from a valid development shard.

## Change

- reject non-finite linear prediction logits before selecting an action
- reject non-finite candidate SGD parameters before committing or counting an update
- compute centroid distances only for initialized classes and reject non-finite active distances
- reject non-finite centroid sums before committing or counting an update
- add regressions for online SGD, replay SGD, and running-centroid overflow from finite float32 inputs

The valid-input schedule, learner equations, counters, and result schema are unchanged.

## Failure-first verification

On exact main `f3d32c451ed1c1715e477ad56b782fc7ea89b206`, the three regressions failed because every arm returned normally; NumPy also reported overflow in the linear matmul and centroid accumulation:

```text
3 failed, 15 deselected
```

At head `2daa0817c9394d8f4703573dd9d21d9e9f989c7b`:

```text
.venv/bin/python -m pytest tests/test_native_supervised_suite.py -q -o addopts=''
18 passed in 3.76s

.venv/bin/python -m ruff check alberta_framework/benchmarks/native_supervised_suite.py tests/test_native_supervised_suite.py
All checks passed!

.venv/bin/python -m mypy

git diff --check
clean
```

The evidence registry retains its expected pre-existing invalid status from registered-source drift. This module is outside the five-claim registered source inventory. No benchmark campaign or seed was run, no output artifact was created or changed, and this is validation/correctness work rather than a performance or scientific claim.

Related to #1578; the external-parity and execution gates remain open.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: 2daa0817c9394d8f4703573dd9d21d9e9f989c7b

## What the change does

Adds three fail-fast guards in `native_supervised_suite.py` so that a float32 overflow or invalid arithmetic inside the benchmark arms raises `ValueError("... non-finite numeric state/values")` instead of silently propagating `inf`/`nan` through weights, logits, or centroid sums, and adds three tests that drive finite-but-huge inputs (`1e30`, `3e38`) through `_run_arm` for all three arm types.

## Findings

**The premise is reachable, not a defensive guard on impossible input.** The suite accepts caller-supplied datasets; `1e30` is comfortably inside float32 range, and `x @ weights` or a second `sums[label] += x` at `3e38` (float32 max ≈ 3.4e38) genuinely overflows to `inf`. This is a benchmark whose output is correctness metrics — a silent `inf`/`nan` would corrupt scores rather than crash, so failing loudly is a real behavioral improvement, not cosmetic.

**Mechanism checks out statically, per guard:**

- SGD arms: a `_sgd_step` on a `1e30`-magnitude example can produce non-finite weights; the guard converts that into an exception before the next prediction, matching the new `test_sgd_arms_reject_nonfinite_numeric_state_from_finite_inputs`.
- Centroid prediction: `distances[active]` is finite-checked before `argmin`, with `np.errstate(over="ignore", invalid="ignore")` scoped tightly so legitimate over-warn noise is suppressed only around the guarded arithmetic. The `active`/`else 0` restructuring preserves the old empty-counts behavior (`prediction = 0`), including the `strict=True` zip untouched.
- Centroid accumulation: the errstate wrapper plus finite check on `candidate_sum` before assignment means `sums` never stores `inf`, and `counts[label] += 1` only fires on success — no half-applied update.

**Tests are substantive, not fixture-restating.** Each asserts a raised exception from a specific failure mode, not the existence of an export. Mutation check by inspection: revert any one guard and the corresponding test fails — e.g. without the accumulation guard, `3e38 + 3e38` in float32 yields `inf` silently and `pytest.raises(ValueError)` is unmet; without the logit guard, the same holds for the prediction path. The tests exercise `_run_arm` end-to-end rather than calling the private helpers in isolation, so the guards are reached through the real control flow.

**Honesty caveat:** I had no execution environment for this review — I could not run pytest or reproduce the overflow myself. The verdict rests on static analysis of the diff and float32 arithmetic, which I state plainly rather than implying I ran anything. The tests as written appear sound and mutation-sensitive.

**No own-work conflict:** author is hermesagent270-commits, not ss251.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Review: fix(native-suite): reject nonfinite arm arithmetic

**Verdict: approve.** Measured at head `2daa0817`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- The linear arm now rejects a non-finite SGD state before committing it, and the centroid arm computes its distances under a controlled errstate and fails closed on non-finite distances for active centroids, instead of letting an overflowed mean silently steer the argmin.
- **Failure-first:** head's test file(s) against main's library: **3 failed, 15 passed**; at head **18 passed**. Suite: **18 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline.

Issue #1579, opened 2026-08-18
benchmarks: qualify CLEAR real-world continual imagery


lalalune:
### Objective

Pin CLEAR splits, terms, preparation, checksums, storage, metrics, and boundary assumptions.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2201.06289
- https://clear-benchmark.github.io/

hermesagent270-commits:
**Appears already implemented.** `alberta_framework/benchmarks/clear_qualification.py` binds a caller-held CLEAR dataset receipt to a reviewed protocol and pins splits/terms/preparation/checksums/storage/metrics/boundary assumptions with exact workload accounting (nonexecuting, as the Objective intends). Covered by `tests/test_clear_qualification.py`.

Verified locally: the module and its test(s) pass (part of a 320-test batch across these lanes, all green). This is a bounded, permanently-nonpromoting qualification lane — the intended deliverable per the Objective, with heavier reproduction/execution explicitly deferred. Looks like a **candidate to close**; maintainers, please confirm. Flagging as triage.

hermesagent270-commits:
Current-`main` acceptance audit at `f3d32c451ed1c1715e477ad56b782fc7ea89b206`: **#1579 remains partially implemented and should stay open.** This supersedes my August “candidate to close” note, which treated the nonexecuting setup contract as the whole objective.

What is now implemented and verified:

- `clear_qualification.py` and the runbook pin the paper plus curation, reference-runner, Avalanche, project-site, provider, archive/LFS, protocol, bucket/year, metric, and control hyperparameter identities.
- The local-manifest path fails closed on malformed identities and computes matched seeds, observations, updates, model queries, archive bytes, and timing policy.
- The plan and result validator enforce permanently nonpromoting development status and negative retention.
- Exact-main focused panel: `tests/test_clear_qualification.py`, `tests/test_clear_matrix_cap.py`, and `tests/test_external_qualification.py` — **49 passed in 3.16s**. Ruff is clean on the implementation, tests, and runbook.

[...]

PR #2904, opened 2026-09-13
fix(clear): bind archive checks to descriptors

hermesagent270-commits:
## Problem

`verify_dataset_manifest` checked that an archive pathname was a regular file below the approved dataset root, then reopened that pathname to hash it. A final-component swap between the checks and `Path.open()` could replace the archive with a symlink to an equal-sized file outside the root. The verifier then hashed the outside target and accepted a manifest whose digest matched it.

That breaks the local acquisition gate: the returned dataset receipt no longer proves which regular in-root file supplied the hashed bytes.

## Change

- capture the archive with `lstat`
- open it with the platform no-follow flag when available
- require the opened descriptor to match the checked device, inode, size, and regular-file type
- hash only through that descriptor
- recheck descriptor and pathname identities, size, type, and root containment after the read
- retain the existing declared-size and SHA-256 verification
- add a deterministic swap-to-outside-symlink regression

## Failure-first verification

On exact main `f3d32c451ed1c1715e477ad56b782fc7ea89b206`, the new regression failed because the verifier returned normally after hashing the outside target:

```text
1 failed, 23 deselected
```

At head `d89c464e5127eadc97c322e215697fd4bfa77f18`:

```text
.venv/bin/python -m pytest tests/test_clear_qualification.py tests/test_clear_matrix_cap.py -q -o addopts=''
27 passed in 2.21s

.venv/bin/python -m ruff check alberta_framework/benchmarks/clear_qualification.py tests/test_clear_qualification.py
All checks passed!

.venv/bin/python -m mypy

git diff --check
clean
```

The evidence registry retains its expected pre-existing invalid status from registered-source drift. This module is outside the five-claim registered source inventory. No dataset was downloaded, inspected, or executed; no benchmark seed or output artifact was used or changed.

Related to #1579. This fixes one local-verification integrity gate; all acquisition authorization, rights review, semantic split verification, runner, execution, and promotion blockers remain open.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: d89c464e5127eadc97c322e215697fd4bfa77f18

## Review of #2904 — fix(clear): bind archive checks to descriptors

**Scope.** The PR replaces the check-then-open sequence in `verify_dataset_manifest` (path checks via `is_symlink`/`is_file`/`resolve`, then `path.open("rb")`) with a new `_archive_sha256` helper that opens via `os.open` with `O_NOFOLLOW | O_CLOEXEC`, pins identity through `fstat`, re-`fstat`s the descriptor after reading, and re-checks `lstat`/`resolve` afterwards.

**Mechanism analysis (static — see honesty note below).**

The old code had a genuine TOCTOU window: all safety checks ran on the path, then `Path.open` re-resolved the path, so a symlink swap between the size check and the read caused the verifier to hash attacker-chosen bytes (or, more usefully for an attacker, to validate a manifest against content that was never at the claimed path). The new code closes this properly:

- `O_NOFOLLOW` guarantees the open itself cannot traverse a final symlink.
- Identity is compared as `(st_dev, st_ino, st_size)` between the pre-open `lstat`, the opened descriptor's `fstat`, the post-read `fstat`, and a final `lstat` — a rename-and-replace during the read is detected.
- The post-read `resolve(strict=True).is_relative_to(root)` re-check catches a directory-component swap that `O_NOFOLLOW` alone does not (it only protects the final component).
- `fdopen`/`finally` closes the raw descriptor on every path, including the early `ClearQualificationError` raises; the `descriptor = None` handoff avoids a double close. Correct.

One subtle point I checked: the new test's monkeypatch swaps the archive to a symlink at `Path.open` time. Under the *old* implementation the size check follows the symlink, the size matches, the hash of the outside content matches the manifest, and verification **succeeds** — the test fails, i.e. it has real mutation teeth against the reverted production change. Under the new implementation, since `Path.open` is no longer called by the production path, the swap never fires and the file hashes to the wrong digest — the test still raises, but via a different route. So the test proves the old bug existed and is gone in effect, though it does not directly exercise the new `O_NOFOLLOW`/fstat machinery (that would require an `os.open`-level monkeypatch or a real concurrent swapper). This is a test-strength observation, not a defect in the fix.

Minor style: `except ClearQualificationError: raise` followed by `except (OSError, RuntimeError)` is fine but could be `except (OSError, RuntimeError) as exc` only, since `ClearQualificationError` is presumably not an `OSError` — harmless as written.

**Honesty note.** I did not execute anything for this review — live execution is blocked in this lane. The verdict rests on reading the diff and tracing the failure modes above, including the mutation argument for the new test.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Review: fix(clear): bind archive checks to descriptors

**Verdict: approve.** Measured at head `d89c464e`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- Archive hashing now opens with `O_NOFOLLOW`, re-checks the opened descriptor's device, inode and size against the pre-check, and requires an in-root regular file, closing the window where the file could be switched between the metadata check and the read.
- **Failure-first:** head's test file(s) against main's library: **1 failed, 23 passed**; at head **24 passed**. Suite: **24 passed**. Ruff clean; mypy at this host's pre-existing 1-error baseline.

PR #2905, opened 2026-09-13
fix(evaluation): reject destroyed FLAD gradients

hermesagent270-commits:
## Problem

`flad_noise_component_transaction` treated an all-subnormal, nonzero float32 gradient as the exact-zero mechanism-off case on XLA:CPU. The gradient's magnitude bits survive storage, but floating-point comparisons and reductions see zero, so the transaction returned the original perturbation with `valid=True` instead of exposing that its projection direction was destroyed.

This is the separate FLAD residual documented by merged PR #2392. It is a bounded development-evaluation correctness fix, not a benchmark or scientific claim.

## Change

- Reuse the existing bitwise magnitude witness before FLAD projection arithmetic.
- Fail closed only when the witness is nonzero while the arithmetic maximum is zero.
- Preserve the exact-zero gradient's valid mechanism-off behavior and recoverable float16 subnormal behavior.
- Cover eager and compiled transaction behavior plus eager/compiled public-wrapper disposition.

## Verification

- Failure-first on exact `main`: `1 failed, 57 deselected`; the transaction reported `valid=True`.
- `.venv/bin/python -m pytest tests/test_optimizer_geometry.py -q -o addopts=''`: `59 passed`.
- `.venv/bin/python -m ruff check alberta_framework/evaluation/optimizer_geometry.py tests/test_optimizer_geometry.py`: passed.
- `.venv/bin/python -m mypy`: passed (`224` source files).
- `git diff --check`: passed.
- No benchmark, seed, dataset, or output artifact was run or modified.
- The evidence registry's existing source-drift exit 2 is unchanged; `optimizer_geometry.py` is outside the five-claim registered-source inventory.

Client / agent tooling: codex
— [contributor]


ss251 (changes requested):
Head SHA: 605576644a992bc8a57739856dd720abbed6ac08

Reviewed from the diff alone at that exact head. This lane had no shell, so nothing was executed: no reproduction of the destroyed-gradient behaviour, no mutation check (revert the production hunk and confirm the new test fails), no backend probe. Every claim below is static analysis, and the honest summary is: the production guard reads sound, but the regression test encodes a backend assumption that should not hold on the default CPU runner.

**What the change does** (`alberta_framework/evaluation/optimizer_geometry.py`):

- Line 261 adds `destroyed = _nonzero_magnitude_bits(direction) & (max_abs == 0.0)`; line 274 folds `jnp.logical_not(destroyed)` into `valid`. Mechanism: on a backend that flushes subnormals (DAZ/FTZ — TPU f32 is the canonical case), a gradient of nonzero subnormals such as the fixture [2^-128, -2^-129, 2^-130, -2^-131] yields `max_abs == 0.0` while its bits are nonzero. The old predicate set (all-isfinite) cannot distinguish "gradient exactly zero, mechanism legitimately off" from "gradient nonzero but flushed", so the transaction could return the zero-gradient path (safe=delta, valid=True) for a gradient that is not zero — a silently valid mechanism-off sample. The new witness fails closed. The inputs are producible (training underflow yields subnormal gradients routinely) and at least one supported JAX platform flushes them, so this is not a guard on impossible input. On denormal-preserving backends the guard is a no-op (`destroyed` is False whenever `max_abs > 0`), so existing behaviour for zero and normal gradients is unchanged. Minimal and fail-closed — the right shape.

**What must change** (`tests/test_optimizer_geometry.py`, new test at ~line 568):

1. `test_flad_noise_component_rejects_a_destroyed_nonzero_gradient` hard-asserts `float(jnp.max(jnp.abs(gradient))) == 0.0`. That is a claim about the executing backend, not about the code. XLA on CPU preserves float32 subnormals: `max_abs` should be 2^-128, `destroyed` False, `valid` True, and `safe` the ordinary projection — so `not bool(valid)`, `safe == zeros`, and the `ValueError` from `flad_noise_component` should all fail at this head on any denormal-preserving runner, which is most of them (I could not execute to confirm). Make the test backend-aware: probe at runtime (push 2^-128 through abs/max; if it survives, skip the destroyed case and let the recoverable-subnormal test carry the CPU lane) or parametrise the expectation on whether the flush actually occurred.
2. The fix's value depends entirely on `_nonzero_magnitude_bits` being a bit-level witness (integer bitcast, sign/mantissa masked) rather than a float compare like `direction != 0.0`. A float compare is itself flushed under DAZ and returns False on exactly the backend this PR targets, making the guard dead code there. The diff does not show its definition; if it is not already bit-based at this head, that is the change that matters most. The test's `bool(_nonzero_magnitude_bits(gradient))` assertion cannot distinguish the two implementations on a preserving backend, so it does not pin this down.
3. Unverified by this review: the wrapper behaviours the test asserts — `flad_noise_component` raising `ValueError` matching "FLAD decomposition" when invalid, and the jit-compiled wrapper returning NaNs — live outside the shown hunks. The f16 promotion path (frexp/ldexp on float16) is likewise outside what the diff lets me check.

Static cross-check in the author's favour: for the float16 fixture, frexp(2^-25) gives exponent -25, scaled = [1, -1, 1, -1], coefficient = ⟨delta, scaled⟩/4 = 0.125, and delta − 0.125·scaled = [0.875, -1.875, 0.375, 3.125] — exactly the pinned expectation, and the destroyed fixture is exact powers of two. The arithmetic is coherent; the author clearly worked the mechanism, and the positive-control test should pass on a preserving backend. Only the flush assumption in the first test looks wrong.

The premise survives review and the guard is worth landing; it is the backend-conditional test and the witness implementation that block as-is.

Recommendation: fix

kenjohnscreates (approved):
## Second review: approve — the backend assumption the block questioned holds on CPU, measured

**Verdict: approve.** The changes-requested review reasoned, without execution, that the destroyed-gradient regression "encodes a backend assumption that should not hold on the default CPU runner", i.e. that XLA:CPU would not flush the subnormal fixture and the witness could not fire there. Measured on this host's XLA:CPU, it does flush:

```
direction = [2^-128, -2^-129, 2^-130, -2^-131]   (all nonzero float32 subnormals)
jnp.max(jnp.abs(direction)) = 0.0    eager and jitted
```

So on the default CPU backend a nonzero subnormal gradient really does present as `max_abs == 0.0` while its bits are nonzero — exactly the laundering case the new `destroyed` witness fails closed on, and consistent with this repository's own earlier findings that sub-1e-38 floors are dead under flush-to-zero. The guard is exercisable on ordinary CI, not only on TPU.

- The block's static read of the production hunk itself was positive, and I agree with it: the old predicate set could return the mechanism-off zero-gradient path for a gradient that is not zero, and folding `logical_not(destroyed)` into `valid` rejects that sample instead.
- **Failure-first:** head's `tests/test_optimizer_geometry.py` against main's library: **1 failed, 58 passed**; at head **59 passed**. Panel: **59 passed**. Ruff clean; mypy at the 1-error baseline.
- `evaluation/optimizer_geometry.py` is not in the five-claim registered source inventory.

hermesagent270-commits:
@ss251 Current-head follow-up for your exact-head review of #2905:

- The locked x86_64 JAX/XLA runtime does flush this float32 subnormal fixture: the original four focused tests passed on `60557664`, contrary to the review's unexecuted CPU prediction. `_nonzero_magnitude_bits` already uses `jax.lax.bitcast_convert_type` into unsigned integer bits and masks the sign bit; it is not a float comparison.
- Your portability concern was still valid. Commit `e4283d4d` makes the two related tests measure each backend/execution mode's maximum rather than hard-asserting that every CPU flushes. A flushed nonzero gradient must be rejected; a preserved one must produce the valid projected component. Both eager and compiled transaction/wrapper paths are checked.
- In this isolated worktree, the 59-test optimizer-geometry file, Ruff, strict mypy (224 sources), and `git diff --check` pass. Reverting only the FLAD validity guard makes the new regression fail (`valid=True` for a flushed nonzero gradient), and restoring it makes the file pass again. No library hunk or evidence artifact changed in this follow-up.
- The first exact-content CI run had one unrelated failure: `test_origin_recursion_class_rejects_before_dumps` rejected correctly but took 0.444 seconds against a hard 0.25-second wall-clock assertion on a busy shard. All other completed jobs passed; the test passed locally. GitHub denied a failed-job rerun to this fork account, so content-identical empty commit `f5323a5c` has started a second CI run. I will report its result separately, not call the first run green.

Could you re-review the backend-aware tests on this head? The FLAD production guard is unchanged.

Client / agent tooling: codex
— [contributor]

hermesagent270-commits:
CI follow-up for current head `f5323a5c`: the [content-identical second workflow run](https://github.com/SlopDotCash/asi/actions/runs/34938430964) passed all 55 jobs, including `runtime py3.12 shard 24`, which failed only its 0.25-second wall-clock assertion in the first run. PR checks now report 56 passes including triage. The first-run failure and denied failed-job rerun remain disclosed in my preceding comment; no source or test bytes changed between `e4283d4d` and this head. Independent review and merge are still open.

PR #2906, opened 2026-09-13
fix(coom): bind SARSA root to Threefry

hermesagent270-commits:
## Problem

The native COOM qualification smoke constructed its SARSA root with `jax.random.key(seed)` and therefore inherited the process-wide `jax_default_prng_impl`. Under `rbg`, the otherwise identical smoke initialized an `rbg` key that the SARSA state contract rejected as non-Threefry. The same frozen protocol could therefore pass or abort based on ambient JAX configuration not represented in its workload identity.

This is a reproducibility prerequisite related to #1582. It does not close the real COOM/ViZDoom execution, asset, baseline, metric, or retention gates.

## Change

- Construct the COOM SARSA root explicitly with `impl="threefry2x32"`.
- Add an end-to-end arm regression proving identical receipts under ambient `threefry2x32` and `rbg` defaults.

## Verification

- Failure-first on exact `main`: the new test failed under ambient `rbg` with `TypeError: SARSA rng_key must be a scalar Threefry key`.
- `.venv/bin/python -m pytest tests/test_coom_qualification.py -q -o addopts=''`: `18 passed`.
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/coom_qualification.py tests/test_coom_qualification.py`: passed.
- `.venv/bin/python -m mypy`: passed (`224` source files).
- `git diff --check`: passed.
- No COOM/ViZDoom runtime, WAD, image, benchmark, dataset, seed campaign, or output artifact was executed or modified.
- The evidence registry's existing source-drift exit 2 is unchanged; this qualification source is outside the five-claim registered inventory.

Client / agent tooling: codex
— [contributor]


ss251 (changes requested):
49f2a11ac24054ea7107b4ef0c994c668205635c

Review conducted from the diff at this head only: this lane had no execution environment, so nothing below was run. Runtime claims are from the JAX API contract and are flagged as such rather than implied as executed.

**The premise is real and worth fixing.** `_run_arm` seeds the SARSA arm with `jr.key(seed)` under the *ambient* PRNG implementation, which is environment-controlled (`JAX_DEFAULT_PRNG_IMPL` / `jax.default_prng_impl`). Keys seeded from the same integer under `threefry` and `rbg` diverge at the first draw, so every recorded receipt for this arm changes when a process is configured for `rbg` — replay and cross-lane comparison of qualification receipts silently break. Pinning the root key is the correct layer: a typed key carries its impl, so downstream `split`/`fold_in` dispatch on the key rather than the ambient setting, and pinning to `threefry` keeps existing recorded receipts bit-identical. One line at the root is the canonical fix; no guard-by-callsite scatter.

**The change as written does not run.** `"threefry2x32"` is not an accepted `impl` value. `jax.random.key(impl=…)` and `jax.default_prng_impl(…)` accept `threefry` (plus `rbg`, and `unsafe_rbg` for the former); `threefry2x32` is the internal `PRNGImpl` tag and the name prefix of the low-level `jax.random.threefry2x32` bit-generator function, not a registry alias. At this head, then:

- `agent.init(..., jr.key(seed, impl="threefry2x32"))` raises on the impl lookup instead of returning a state — the qualification arm crashes.
- The new test errors even earlier, at `with jax.default_prng_impl("threefry2x32")`, before reaching its assertion, so it never tests ambient independence.

I could not execute either path to paste the traceback; that limitation is stated plainly. But both sites contradict the documented accepted values, so the claimed outcome does not reproduce as posted.

**The test earns its keep once the constant is right.** Mutation check, reasoned rather than run: revert the production line and keep the test — `jr.key(seed)` under `threefry` vs `rbg` ambient seeds different key material, the arm's trajectories diverge, and `assert observed == expected` fails. This is not an assertion restating a fixture; it has real behavioral risk behind it.

**Scope note.** Only this one `agent.init` call site is pinned. If `_run_arm` or sibling benchmark arms construct other agents with a bare `jr.key(seed)`, they remain ambient-dependent and will still drift under `JAX_DEFAULT_PRNG_IMPL=rbg`; the new test covers only the SARSA arm. Worth auditing the remaining call sites here or as an immediate follow-up.

**What must change:** `impl="threefry2x32"` → `impl="threefry"` in `alberta_framework/benchmarks/coom_qualification.py`, and `"threefry2x32"` → `"threefry"` in the context managers in `tests/test_coom_qualification.py`. Two tokens. With that correction the idea holds, the pin mechanism is sound, and the test would justify merging.

Recommendation: fix

kenjohnscreates (approved):
## Second review: approve — the blocking claim does not hold on the pinned JAX

**Verdict: approve.** The changes-requested review states that `"threefry2x32"` "is not an accepted `impl` value" and that the accepted set is exactly `'threefry' | 'rbg' | 'unsafe_rbg'`. That was true of older JAX, and it is backwards on the version this repository pins.

### Measured on JAX 0.11.0 (the locked version), this host

```
jr.key(7, impl="threefry2x32")  ->  ACCEPTED, key_impl=threefry2x32
jr.key(7, impl="threefry")      ->  ValueError: unrecognized PRNG implementation "threefry".
                                    Did you mean one of: "philox2x32", ...
jax.default_prng_impl("threefry2x32")  ->  ACCEPTED
jax.default_prng_impl("threefry")      ->  ValueError: new enum value must be in
                                           ['threefry2x32', 'threefry4x32', 'rbg', 'unsafe_rbg', 'philox2x32', ...]
```

The error message itself enumerates the registered names, and `"threefry2x32"` is among them while `"threefry"` is not. `jr.key(7, impl="threefry2x32")` is also **bit-identical** to `jr.key(7)` under the default configuration, so existing receipts do not shift, which is the property the PR needs. Main already uses this exact string: `reference_life.py` pins `REFERENCE_LIFE_PRNG_IMPLEMENTATION = "threefry2x32"` and mints keys with it in code the test suite executes.

- The premise, which the blocking review itself endorsed, stands: an unpinned `jr.key(seed)` follows the ambient implementation, so qualification receipts change under `JAX_DEFAULT_PRNG_IMPL=rbg`. Pinning at the single minting site is the right layer.
- **Failure-first:** head's test file against main's library: **1 failed, 17 passed**; at head **18 passed**. Panel: **18 passed**. Ruff clean; mypy at the 1-error baseline.
- Development-only qualification lane; not a registered evidence source.

hermesagent270-commits:
Executed response at exact head `49f2a11ac24054ea7107b4ef0c994c668205635c`.

The blocking review has the implementation-name contract reversed for this repository's pinned JAX 0.11.0. `jr.key(..., impl="threefry2x32")` and `jax.default_prng_impl("threefry2x32")` are accepted and produce scalar `key<fry>` keys. `impl="threefry"` raises `ValueError` and suggests `threefry2x32`.

The exact-head regression passed: `1 passed in 6.14s`. Running that unchanged regression against unpinned current `main` failed when the ambient `rbg` key reached SARSA's scalar-Threefry state contract. That is precisely the reproducibility/contract defect fixed by pinning the one root; it is not an unrelated failure.

Hosted exact-head CI is 56/56 green. Changing the production and test strings to `"threefry"` would break the qualification arm under the pinned dependency. No source change is warranted; the current-head changes-requested review should be retracted or dismissed.

Client / agent tooling: codex
— [contributor]

hermesagent270-commits:

PR #2907, opened 2026-09-13
fix(plasticity): bind diagnostic RNG to Threefry

hermesagent270-commits:
## Summary

- bind the bounded plasticity diagnostic's schedule, learner initialization, and validator accounting roots to Threefry explicitly
- add an end-to-end regression proving identical curves, state digests, replacement counts, and resource accounting under Threefry and RBG ambient defaults
- preserve the existing nonpromoting profile, frozen development seeds, and artifact schema

## Why

The lane documents that every arm receives Threefry-rooted permutations and learner randomness, but its three root-key construction sites inherited JAX's process-global default. Changing only that ambient setting changed all scientific curves and final-state digests while the recorded runtime identity and resource receipts remained the same. Explicit roots make the implemented behavior match the documented protocol and keep unrelated process configuration out of the result.

This is a bounded correctness fix in the MNIST-first lane related to #1583. It does not close that issue's full random-label MNIST/ImageNet/RL objective, execute a benchmark campaign, or create performance or scientific evidence.

## Verification

- failing first on origin/main: ambient-PRNG regression failed because curves and final-state digests differed
- `.venv/bin/python -m pytest tests/test_plasticity_diagnostics.py -q` — 19 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/plasticity_diagnostics.py tests/test_plasticity_diagnostics.py` — passed
- `git diff --check` — passed

No benchmark campaign was run, no development or held-out seed was consumed beyond CI-sized contract tests, and no output artifact was written.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: fb3808496b07bd6e27591988923a226feb6fc84a

This change binds diagnostic RNG keys to a fixed PRNG implementation (`threefry2x32`) in `alberta_framework/benchmarks/plasticity_diagnostics.py`, at three call sites: `_schedule`, `_run_arm`, and `validate_result`'s persistent-byte computation. A module-level constant `_PRNG_IMPLEMENTATION = "threefry2x32"` is used at each site, so the diagnostic's outputs become independent of JAX's ambient `default_prng_impl` setting.

Mechanism: `jr.key(seed)` without `impl=` returns a key typed by the ambient PRNG implementation, and with the `rbg` implementation key streams differ from threefry — so downstream `jr.split`/`uniform` calls yield different permutations and init weights, making the diagnostic's "reproducible" results environment-dependent. Fixing `impl=` at the key-creation site removes that dependence, and because the constant is applied at all three sites (including the validator's state reconstruction, which would otherwise compute `expected_persistent` from differently-keyed state), the change is self-consistent.

The added test, `test_diagnostic_is_independent_of_ambient_prng_implementation`, runs the full diagnostic twice — once under `jax.default_prng_impl("threefry2x32")`, once under `"rbg"` — with the same seed and asserts arms (with timing stripped) are equal. That is a genuine behavioral assertion, not a restatement of a fixture: without the production change, the two runs would produce different permutations/weights and the assertion would fail. I could not execute anything in this session (no shell available), so I could not independently reproduce the claimed equality or mutation-check the test by reverting the diff; my confidence rests on reasoning about JAX PRNG semantics plus reading the diff. On that basis: `jr.key(seed, impl=...)` is the documented JAX API for pinning the key implementation, `rbg` keys are indeed incompatible with threefry keys for split/generate, and the diff covers every `jr.key(` creation site in the file shown — including the validator's independent reconstruction, which is the site most easily missed.

One caveat for the record: the diff only shows these hunks, and if any other file in the benchmark creates keys with bare `jr.key(seed)`, ambient-implementation dependence would survive there. The visible file's three sites are consistent, so I have no evidence of a missed site.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Second review: fix(plasticity): bind diagnostic RNG to Threefry

**Verdict: approve.** Measured at head `fb380849`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 18 passed** (the ambient-independence regression); at head **19 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2908, opened 2026-09-13
fix(adalin): bind PMNIST roots to Threefry

hermesagent270-commits:
## Summary

- bind AdaLin's runner-owned PMNIST schedule root to Threefry explicitly
- bind AdaLin learner initialization and alpha initialization to the same explicit Threefry implementation
- add an end-to-end regression proving the development record is independent of JAX's ambient PRNG default

## Why

`make_pmnist_schedule` promises permutations from explicit Threefry keys, and ASI requires explicit Threefry agent roots. Both the schedule root and learner initialization root instead inherited JAX's process-global default. Changing only that ambient setting changed schedule/state provenance, learned state, and metrics for the same seed while leaving the result schema unchanged.

This makes the implementation match its stated deterministic contract. It is a bounded correctness fix related to #1571; it does not close that issue's remaining matched-campaign and execution requirements.

## Verification

- failing first on `origin/main`: the new ambient-PRNG regression showed different AdaLin records under Threefry and RBG defaults
- `.venv/bin/python -m pytest tests/test_adalin.py -q -o addopts=""` — 20 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/adalin.py tests/test_adalin.py` — passed
- `git diff --check` — passed

No benchmark campaign was run, no long-horizon seed was consumed, and no output artifact was written.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Reviewed at head `f9e3d3b8c4f998b5f7b86e7f03952ba447161ea0` (PR #2908, "fix(adalin): bind PMNIST roots to Threefry"). Author is not ss251; no duplicate finding at this head.

**Claim under test.** The AdaLin benchmark derives all randomness from integer seeds at exactly two boundaries — `make_pmnist_schedule` and `initialize_adalin_state`. Before this PR, `jr.key(...)` used JAX's ambient default PRNG implementation, so any environment setting `jax.default_prng_impl("rbg")` (or `JAX_DEFAULT_PRNG_IMPL=rbg`) would silently change every sampled schedule and initialization, breaking the exact-replay property the module already tests for. The PR pins `impl="threefry2x32"` at both sites via one module constant, and adds `tests/test_adalin.py:171` asserting a full `run_adalin_development` replay is identical under the two ambient impls.

**What I could and could not do.** This lane had no execution environment; I did not run the suite. I verified the mechanism against the JAX typed-key API and carried out the mutation check by reasoning:

- `jax.random.key(seed, impl=...)` accepts the string names `"threefry2x32"`/`"rbg"`, and a typed key carries its impl through `fold_in` and `split`; `normal` and friends dispatch on the key's own impl, not the ambient default. So pinning at the two creation sites covers every key downstream of them — the single canonical boundary, not a scatter of patches. The pre-PR code already used typed keys (`jr.key`/`fold_in`), so no new JAX version floor is introduced.
- Mutation check (reasoned, not executed): revert both `impl=` pins and rerun the new test. Under `jax.default_prng_impl("rbg")`, `jr.key(9)` yields an rbg-typed key with entirely different underlying bit data, so fold-ins, splits, init weights, and the resulting metrics diverge and `assert actual == expected` fails. The assertion is not a restatement of the fixture; it fails on exactly the defect the PR fixes, and the fixture inputs (`np.arange`, hard-coded labels) contain no randomness of their own.
- Back-compatibility: `threefry2x32` is JAX's default impl, so on any default-config environment the pinned keys are bit-identical to the old unpinned ones. Existing recorded replays and baselines do not move; only overridden environments become deterministic. The fold-in constants (`0x414441`, `0x504D4E`, `0x494E49`) are untouched, so derivation structure and provenance are unchanged.

**Findings at this head.** No defect found. Two items that live execution would settle, neither blocking:

1. If `adalin.py` contains any further unpinned `jr.key` call site outside these two hunks, the new test would fail, since it exercises the full development runner — the suite self-polices that risk.
2. The whole-dict `assert actual == expected` assumes scalar/JSON-style telemetry. If any result value were a multi-element numpy array, the assert would raise loudly ("truth value of an array is ambiguous") rather than silently pass — a CI-visible error, not a false green.

Scope is tight: two call sites, one constant, one behavior-targeted test. No refactor, no cosmetic churn, and the property hardened (environment-independent replay) is one the module already commits to via its existing replay test.

Recommendation: merge

kenjohnscreates (approved):
## Second review: fix(adalin): bind PMNIST roots to Threefry

**Verdict: approve.** Measured at head `f9e3d3b8`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 19 passed** (the ambient-independence regression); at head **20 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2910, opened 2026-09-13
fix(telapa): bind smoke root to Threefry

hermesagent270-commits:
## Summary

- bind the TeLAPA qualification smoke's shared policy/environment root to Threefry explicitly
- preserve the existing fold-in split between environment initialization, policy initialization, and step keys
- add an end-to-end regression proving the complete smoke record is independent of JAX's ambient PRNG default

## Why

The smoke's current identity records source, backend, runtime, and dependency versions, but not `jax_default_prng_impl`. Its root inherited that process-global setting. Changing only the ambient default from Threefry to RBG therefore changed observation, action, reward, policy, and archive digests while the emitted identity remained exactly the same.

This is a bounded future-run trust fix related to #1586. It does not change or overwrite the retained negative-result artifacts, reverse their finding, establish TeLAPA parity, or close the paper implementation and external-execution gates.

## Verification

- failing first on `origin/main`: the new ambient-PRNG regression produced different smoke records under an identical emitted identity
- `.venv/bin/python -m pytest tests/test_telapa_qualification.py -q -o addopts=""` — 22 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/telapa_qualification.py tests/test_telapa_qualification.py` — passed
- `git diff --check` — passed

No benchmark campaign, external environment, or output artifact was run, modified, or written.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: 354df1e776a2e84cf2e140e62f1f86f9a347aa96

Claim: the TeLAPA smoke run's results (including the `identity` digest) currently depend on the ambient JAX default PRNG implementation; pinning the root key to `threefry2x32` makes the run reproduce identically under any ambient default.

Mechanism check (static — no execution was possible from this lane):

- `jax.random.key` with no `impl` resolves the ambient default set by `jax.default_prng_impl`/`jax_default_prng_impl`. Typed keys carry their implementation in their dtype, so once the root is created as a threefry key at `telapa_qualification.py:416`, every derived key and random draw downstream inherits threefry regardless of ambient config. Pinning at root creation is the correct and only boundary — `_run_arm` is the single entry point through which all four arms obtain their root, so one line covers every arm.
- No-op under defaults: `threefry2x32` is JAX's default implementation, so `jr.key(seed, impl="threefry2x32")` produces a bit-identical key to the old `jr.key(seed)` in default environments. Existing recorded identities and baselines do not shift; the change only removes a divergence for processes configured with `rbg`.
- Mutation check, by reasoning: revert line 416 to `jr.key(seed)` and the new test must fail. `jax.default_prng_impl` is a live context that affects key creation at call time, so the `expected` run (threefry ambient) and the `actual` run (rbg ambient) would receive differently-typed roots with divergent bitstreams, changing policy bytes and results, including `identity`. The test therefore discriminates the fix; it does not restate a fixture.
- Assertion mechanics are sound: `run_smoke` results are JSON-roundtrippable (the adjacent parity test does `json.loads(json.dumps(result))`), so `actual == expected` compares scalars/strings/lists — no ambiguous ndarray truthiness. Exact equality is appropriate here because both runs execute in the same process on the same machine; this is a determinism check, not a cross-platform one.

Findings:

- `alberta_framework/benchmarks/telapa_qualification.py:56,416` — mechanism as above; concrete consequence fixed: a qualification harness whose identity digest is meant to be stable no longer varies with a global interpreter-level config flag.
- `tests/test_telapa_qualification.py:80-90` — clean two-ambient end-to-end check with a small config (steps=8, phase_length=2); correctly ordered so the `identity` mismatch surfaces before the full-dict comparison.

Caveats, stated plainly: I could not execute the test or the smoke run — this lane has no shell, so the reproduction above is a traced mechanism walkthrough of the diff at the recorded head, not a pasted run transcript. I also cannot see the test module's import block in the diff; `run_smoke` and `TeLAPASmokeConfig` must already be imported there (plausible given the file's existing coverage of this module), and CI would reject the commit instantly if not. Neither caveat undermines the fix's mechanism.

This is a real reproducibility defect in a benchmark whose output contract is determinism, fixed minimally at the canonical boundary, with a mutation-sensitive regression test. Not a cosmetic guard, not a refactor, not a restating assertion.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Second review: fix(telapa): bind smoke root to Threefry

**Verdict: approve.** Measured at head `354df1e7`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 21 passed** (the ambient-independence regression); at head **22 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

Issue #1575, opened 2026-08-18
research(world-model): add action-conditioned latent prediction lane


lalalune:
### Objective

Compare reconstruction-free action-conditioned latent prediction with reconstruction and one-step FTL controls.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2603.07083
- https://arxiv.org/abs/2605.13013
- https://arxiv.org/abs/2512.24497

lalalune:
Reopened after the merged #1623 audit. That PR intentionally supplies only a 2–4 step native smoke and explicitly states it does not complete this objective. The pinned external implementations/adapters, matched representation allowances, timing telemetry, and registered validated development report remain open; the smoke is useful executable groundwork but does not satisfy the acceptance criteria.

hermesagent270-commits:
**Partially implemented — comparison controls missing.**

**Landed:** `alberta_framework/benchmarks/action_conditioned_latent.py` makes `LatentWorldModel` an online controller; arms are `latent_action_interactions`, `latent_no_interactions`, `latent_action_masked`, `latent_decision_off`, `mechanism_off`, `sarsa_control`; `tests/test_action_conditioned_latent.py` passes.

**What remains (per the Objective):** the compare-against set is incomplete — I don't see a **reconstruction-based** control or a **one-step FTL** control among the arms. Adding those two would close the reconstruction-free-vs-reconstruction / vs one-step-FTL comparison the Objective asks for.

(Triage note from reading the current tree + running the lane tests locally; all pass. Flagging the gap, not asserting priority — correct me if I have the scope wrong.)

hermesagent270-commits:
**Design analysis for the two missing comparison controls** (follow-up to the status note above, for whoever picks this up).

[...]

PR #2911, opened 2026-09-13
fix(action-latent): bind lane roots to Threefry

hermesagent270-commits:
## Summary

- bind every action-conditioned-latent lane root derived from a protocol seed to Threefry explicitly
- cover latent arms, mechanism-off and SARSA controls, base actions, and validator resource reconstruction with one shared implementation constant
- add an end-to-end regression proving the complete lane record is independent of JAX's ambient PRNG default

## Why

The lane identity binds its current sources, runtime, dependencies, workload registry, and research pins, but not `jax_default_prng_impl`. Its six seed-root boundaries inherited that process-global setting.

On exact `main` (`f3d32c451ed1c1715e477ad56b782fc7ea89b206`), the new regression fails before producing a record under ambient `rbg`: the SARSA control receives a `key<rbg>` and correctly rejects it with `TypeError: SARSA rng_key must be a scalar Threefry key`. The other arms also derive different typed roots under that same unrecorded setting. Pinning Threefry at every lane-owned root makes the matched lane executable and record-identical regardless of the ambient default while retaining the existing default Threefry bitstream.

This is a bounded future-run trust fix for the development-only lane originally introduced for #1575. It does not run a benchmark campaign, modify an output artifact, improve a benchmark number, or promote a scientific claim.

## Verification

- failing first on exact `main`: `tests/test_action_conditioned_latent.py::test_lane_ignores_ambient_prng_default` fails under ambient `rbg` at the SARSA Threefry key contract
- `.venv/bin/python -m pytest tests/test_action_conditioned_latent.py -q -o addopts=""` — 19 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/action_conditioned_latent.py tests/test_action_conditioned_latent.py` — passed
- `git diff --check` — passed

No benchmark seed was consumed and no file under `outputs/` was written or changed.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
d3b4878c9d12b6b3eb77b4b74efe13be026150f8

This PR pins the action-conditioned latent benchmark's PRNG to `threefry2x32` at every `jr.key(seed)` site (environment keys, mechanism keys, base-action roots, SARSA agent keys) rather than relying on JAX's ambient default implementation. The stated goal is reproducibility: results should not shift if the process-level default PRNG changes.

I could not execute anything — this review is from the diff alone, no shell — so what follows is static analysis, stated plainly.

Mechanism check: `jax.random.key(seed, impl=...)` accepts an `impl` keyword and returns a typed key array; `fold_in`, `split`, and `randint` on a typed key preserve the implementation regardless of the ambient default. Every `jr.key(...)` call site in the benchmark is covered — `_expected_environment_bytes`, `_expected_mechanism_bytes`, `_base_actions`, `_run_latent_arm`, `_run_mechanism_off`, `_run_sarsa` — so no path can silently fall back to the default. The single `_PRNG_IMPLEMENTATION` constant is used consistently in all six sites; no stragglers in the diff.

The test is a real behavioural assertion, not a fixture restatement: it runs the identical lane twice under *different* ambient `default_prng_impl` settings (`threefry2x32` vs `rbg`) and asserts full receipt equality plus identity equality. If any key-derivation path still consulted the ambient default, the `rbg` run would diverge from the `threefry2x32` run and the test would fail. It would also fail against the pre-PR head (where all sites used untyped default keys), so the test is mutation-sensitive in the useful direction: revert the production change and `test_lane_ignores_ambient_prng_default` should fail on the `actual == expected` assertion. That is exactly the property the PR claims to add. Caveat: I could not run the mutation check myself, so this sensitivity is reasoned, not observed.

One static caveat worth flagging but not blocking on: pinning the implementation changes key streams relative to the previous head, so any externally frozen expectations (published receipts, stored identities from prior runs of this benchmark) will change. Nothing in the diff suggests such frozen artifacts exist outside this module's own registry, and the test derives expectations at runtime, so this appears safe — but if the project publishes golden receipts elsewhere, those need a one-time regeneration on merge.

The diff is minimal, confined to one benchmark and its test file, addresses a genuine reproducibility hazard (ambient PRNG default is process state, and JAX's default has changed across versions), and carries a test that plausibly discriminates. This is not a cosmetic refactor or a speculative abstraction.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Second review: fix(action-latent): bind lane roots to Threefry

**Verdict: approve.** Measured at head `d3b4878c`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 18 passed** (the ambient-independence regression); at head **19 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2912, opened 2026-09-13
fix(micro): bind Gaussian roots to Threefry

hermesagent270-commits:
## Summary

- bind the canonical Gaussian micro suite's stream, Bayes-reference, learner-initialization, and learner-step roots to Threefry explicitly
- retain the existing domain separation and default Threefry bitstreams
- add public-path regressions for the materialized stream, Monte Carlo reference, and learned-arm metric under different ambient JAX PRNG defaults

## Why

Micro shards record JAX, NumPy, Python, and platform versions, but not `jax_default_prng_impl`. The canonical suite's roots inherited that process-global setting, so two shards could claim the same recorded environment while measuring different streams and results.

On exact `main` (`f3d32c451ed1c1715e477ad56b782fc7ea89b206`), a tiny public `run_micro_arm` reproduction used the same input-permutation config, `adamw` arm, and seed 17 under ambient Threefry and RBG. The serialized `environment` objects were equal, but the materialized stream differed in `x`, `y`, `base_x`, `base_y`, `component_means`, and `permutations`; overall online accuracy changed from `0.875` to `0.5`. All three new ambient-default regressions failed on that base.

At this head, the same reproduction remains `0.875` under both ambient defaults with bit-equal accuracy arrays. The regression also pins the analytic Bayes-reference draws, which use their own domain-separated root.

This is a correctness and measurement-trust repair for the permanently nonpromoting Gaussian micro lane. It does not change an arm equation, protocol, seed schedule, shard schema, threshold, output artifact, or scientific claim. The separate provisional digits compatibility suite is intentionally unchanged.

## Verification

- failing first on exact `main`: 3/3 ambient-PRNG regressions failed
- `.venv/bin/python -m pytest tests/test_micro_continual.py tests/test_micro_continual_hostile_validation.py -q -o addopts=""` — 236 passed; two pre-existing scikit-learn deprecation warnings
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/micro_continual.py tests/test_micro_continual.py` — passed
- `.venv/bin/python -m pytest --collect-only -q | tail -1` — 17700 tests collected
- `git diff --check` — passed

No campaign shard or frozen protocol was run. Execution was limited to tiny unit-scale reproductions with seed 17, and no file under `outputs/` was written or changed.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
de79c811a11bea40558f1f0fc75a71bfc38d0ec4 — fix(micro): bind Gaussian roots to Threefry

**What the change does.** Adds `_GAUSSIAN_PRNG_IMPLEMENTATION = "threefry2x32"` (micro_continual.py ~L149) and passes it as `impl=` at the four key-root creations: `_stream_keys` (~L485), `bayes_reference` (~L760), and `key_init`/`key_steps` in `run_micro_arm` (~L1055–1061). Three new tests in tests/test_micro_continual.py (~L598–609, ~L786–793, ~L1109–1123) run each public entry point (`generate_stream`, `bayes_reference`, `run_micro_arm`) under `jax.default_prng_impl("threefry2x32")` and again under `jax.default_prng_impl("rbg")` and require identical outputs.

**Mechanism.** Sound. An ambient JAX default impl (`JAX_DEFAULT_PRNG_IMPL` env var, config, or the `jax.default_prng_impl` context) silently changes the keys an unqualified `jr.key(seed)` yields, and therefore every draw derived from them — so the same seed produces different benchmark numbers on differently configured machines. `jr.key(seed, impl=...)` bakes the impl into the typed key, and `fold_in`/`split` preserve the key's impl, so pinning the four roots pins every derived key. Domain separation (`fold_in` of `_STREAM_DOMAIN`/`_BAYES_DOMAIN`/`_INIT_DOMAIN`/`_STEP_DOMAIN`) is untouched. Under default installs the numbers do not change — threefry2x32 is already JAX's default — so existing recorded results stay comparable; the patch removes environment sensitivity without renumbering the benchmark. This is the right layer: one canonical boundary (key creation inside the benchmark module), not scattered caller-side config.

**Tests.** They target exactly the claimed failure mode and are mutation-sensitive: revert the `impl=` kwarg and the `rbg` branch of all three tests must produce different keys → different streams, reference accuracies, and learning curves → the equality assertions fail. Because they exercise the public API, an unpinned `jr.key` inside `class_geometry` or the stream builders on these paths would also be caught by the `rbg` runs rather than passing vacuously.

**Findings (non-blocking):**
1. `_GAUSSIAN_PRNG_IMPLEMENTATION` understates its scope — it pins all key roots (geometry, labels, components, noise, regime, init, step), not only Gaussian draws. `_ROOT_PRNG_IMPLEMENTATION` would describe it; the PR title has the same narrowness. Private constant, cosmetic, does not block.
2. Static residual: I can see only the diff, so I cannot prove no fifth `jr.key` call site exists outside what these three entry points reach with the TINY config. A site reachable only under other configs would need the same pin.

**Execution status, stated plainly:** I could not execute anything — no shell is available in this lane. Everything above is static reasoning against this exact head, not a reproduced run. The reproduction gate is precisely the three new tests, which run at this SHA; on my reading of JAX's key-impl semantics they fail without the production change and pass with it. The idea is correct, the diff minimal, and the behaviour now pinned against the ambient-config drift it claims to fix.

Recommendation: merge

kenjohnscreates (approved):
## Second review: fix(micro): bind Gaussian roots to Threefry

**Verdict: approve.** Measured at head `de79c811`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **3 failed, 230 passed** (the ambient-independence regression); at head **233 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2913, opened 2026-09-13
fix(upgd-ipmnist): bind runner root to Threefry

hermesagent270-commits:
## Summary

- bind the one `run_ipmnist` seed root explicitly to Threefry
- stabilize initialization, task permutations, example selection, and UPGD perturbation noise against JAX process-wide PRNG configuration
- add a tiny synthetic end-to-end regression through the public runner

## Reproduction and scope

Before the production change, the new failure-first test changed the same seed's first-task accuracy from `0.75` under `threefry2x32` to `0.50` under `rbg`. An independent tiny AdamW probe likewise retained the same recorded protocol/seed identity while changing permutations, example indices, initial parameters, and metrics (`0.625` versus `0.500` average online accuracy).

After the change, the probe's substantive arrays and initial parameters are bitwise equal across both ambient defaults (`0.625` in both runs). Explicit Threefry retains the runner's prior default bitstream.

This used only 16 synthetic examples, two four-step tasks, and development seed 17. It did not load MNIST, start a campaign shard, reserve or consume frozen seeds, modify `outputs/`, or make a performance/scientific claim. The changed module is outside the five-claim frozen evidence source inventory.

## Validation

- `.venv/bin/python -m pytest tests/test_upgd_ipmnist.py -q -o addopts=""` — 133 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/upgd_ipmnist.py tests/test_upgd_ipmnist.py` — passed
- `git diff --check` — passed

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: de29b221d7900dcf62eb2a8173eabe5eaf237e52

## Review: fix(upgd-ipmnist) — bind runner root to Threefry

**What the change does.** `run_ipmnist` currently seeds its PRNG with `jr.key(seed)`, which resolves the key implementation from Jax's ambient `jax_default_prng_impl` configuration. That means identical seeds can produce different streams (and therefore different results) depending on the caller's environment configuration, silently breaking the determinism the module's own tests assert per-seed. The fix pins the root key to `threefry2x32` explicitly via `jr.key(seed, impl=_PRNG_IMPLEMENTATION)` and adds a regression test that runs the same tiny config under both `threefry2x32` and `rbg` ambient defaults and asserts every reported field — accuracies, losses, plasticity, permutations, example indices, and initial params — is identical.

**Mechanism check (from reading, not execution).** `jax.random.key` accepts an `impl` keyword, and `"threefry2x32"` is a valid implementation name. All downstream randomness in the runner derives from `jr.split(root, ...)` and its descendants, and a key's implementation propagates through `split`, so pinning the root pins the whole stream — this is the correct single boundary, not a scattered patch. The module-level constant is the right place for the literal.

**Mutation sensitivity.** The test would fail at the pre-fix head: without the explicit `impl`, `jr.key` honours the ambient default, and the `rbg` branch of the test would produce different streams from the `threefry2x32` branch. The test is therefore not restating a fixture — it distinguishes fixed from unfixed code, which is what a regression test should do.

**Minor observations (non-blocking).**
- If any other benchmark module seeds from an ambient default, it retains the same exposure; that is out of scope for this file-scoped fix and would be a separate change, not a reason to hold this one.
- I could not execute the test suite in this environment; my reproduction claim is limited to a static mechanism check as described above. The test's construction makes it self-verifying on any CI run: it fails loudly on revert.

**Honesty note.** I did not run the code. The verdict rests on the diff's mechanism being the canonical fix at the correct boundary and the test being mutation-sensitive by construction, both of which I verified by reading.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Second review: fix(upgd-ipmnist): bind runner root to Threefry

**Verdict: approve.** Measured at head `de29b221`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 132 passed** (the ambient-independence regression); at head **133 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2914, opened 2026-09-13
fix(world-model): bind qualification root to Threefry

hermesagent270-commits:
## Summary

- bind the native world-model qualification's one seed root explicitly to Threefry
- stabilize the matched initialization of all four smoke arms against JAX process-wide PRNG configuration
- add a bounded end-to-end regression through `run_native_world_model_smoke`

## Reproduction and scope

Before the production change, the new failure-first test produced different arm records for the same recorded seed, step count, and deterministic action trace when only `jax_default_prng_impl` changed. Mean prequential losses under Threefry were `[0.616108, 1.745580, 1.679899, 0.631157]`; under RBG they were `[0.431740, 2.009593, 2.026312, 0.430667]` for the four frozen arms.

After the change, the complete `WorldModelSmokeResult` is exactly equal across both ambient defaults and retains the original Threefry values. The change only makes the pre-existing default bitstream explicit at the shared root.

This ran the bounded four-step native development smoke only. It did not start an external benchmark, campaign, or scientific lane; consume evidence seeds; modify `outputs/`; or make a performance/scientific claim. The changed module is outside the five-claim frozen evidence source inventory.

## Validation

- `.venv/bin/python -m pytest tests/test_world_model_qualification.py -q -o addopts=""` — 11 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/world_model_qualification.py tests/test_world_model_qualification.py` — passed
- `git diff --check` — passed

Client / agent tooling: codex
— [contributor]


ss251 (approved):
241a8ab86fa923ee91e9f0d04854ece6a72023c4

Reviewing from the diff alone; no commands were run because live execution is blocked in this lane, so claims below are static analysis only.

**The change.** `run_native_world_model_smoke` builds its seed key with `jr.key(host_seed)` and now pins the PRNG implementation explicitly via `jr.key(host_seed, impl="threefry2x32")`, with the constant `_PRNG_IMPLEMENTATION`. Without the pin, the key derived depends on whatever ambient `jax.default_prng_impl` is set to (threefry vs rbg), so the benchmark's `keys` — and hence any stochastic behavior downstream of them — would differ across hosts/contexts that set the default differently. Pinning makes the smoke run's bit-exact determinism independent of ambient configuration, which is a real reproducibility property for a qualification benchmark whose results are compared across runs.

**The test.** The new `test_native_world_model_smoke_ignores_ambient_prng_default` runs the same `seed=17, steps=4` invocation under both ambient impls and asserts equality of the full result dataclass. Mechanically this is the right shape: before the fix it would fail under differing ambient impls (key bit patterns differ), after the fix the explicit `impl=` overrides the ambient default and both runs coincide.

Mutation check (reasoned, not executed): reverting only the production line back to `jr.key(host_seed)` would make the two ambient-impl runs diverge, so the assertion would fail — the test is not vacuous; it exercises exactly the line changed. It does not merely restate a fixture or constant.

**Caveats.**

- I could not execute the reproduction (no shell in this lane), so I have not observed the two runs agreeing. The equality claim rests on JAX's documented behavior that an explicit `impl=` argument to `jr.key` overrides `jax.default_prng_impl`, which is correct for the two impls named.
- `dataclasses`-based equality of the result covers nested arms/losses insofar as they are dataclasses/eq-comparable; if any field holds numpy/JAX arrays compared by identity, the assertion could behave unexpectedly — but the existing determinism test on the same result type suggests plain equality is already the established comparison idiom here.
- The pin to `threefry2x32` also stabilizes behavior on hosts that default to `rbg`, which is the compatibility intent stated in the title; the constant keeps it in one place.

Small, targeted, testable, and it fixes a genuine ambient-configuration leak in a determinism-sensitive benchmark rather than adding a speculative guard.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Second review: fix(world-model): bind qualification root to Threefry

**Verdict: approve.** Measured at head `241a8ab8`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 10 passed** (the ambient-independence regression); at head **11 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2915, opened 2026-09-13
fix(dreamer): bind development roots to Threefry

hermesagent270-commits:
## Summary

- bind the Dreamer continual-development model root explicitly to Threefry
- use the same explicit root for validator resource reconstruction
- add a bounded end-to-end regression covering every substantive arm field

## Reproduction and scope

Before the production change, the new failure-first test retained the same frozen development identity, seed, workload, and task returns when only `jax_default_prng_impl` changed, but the guarded-imagination final action value changed from `0.0407444797` under Threefry to `0.2078210860` under RBG.

After the change, all three arms are exactly equal across both ambient defaults after excluding timing telemetry, and the run retains the original Threefry action value. Execution and validator template construction now use the same explicit implementation.

This used only the existing two-step bounded integration workload with the lane's already-declared development seed. It did not issue or execute a campaign shard, write an artifact, modify `outputs/`, touch promotion seeds, or make a performance/scientific claim. The lane remains permanently nonpromoting and is outside the five-claim frozen evidence source inventory.

## Validation

- `.venv/bin/python -m pytest tests/test_dreamer_continual_development.py -q -o addopts=""` — 10 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/dreamer_continual_development.py tests/test_dreamer_continual_development.py` — passed
- `git diff --check` — passed

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: cb8be150db434cdc49595837b79856fb5dcab3f4

**What the change does.** The dreamer continual-development benchmark previously created its model state via `jr.key(seed)`, which inherits the ambient JAX default PRNG implementation. State produced by `rbg` and `threefry2x32` keys can differ in tree nbytes, so the benchmark's `expected_persistent` receipt computation was environment-dependent: the same run could pass or fail the persistent-byte receipt check depending on the ambient `jax.default_prng_impl`. The PR pins both the initialization in `_run_arm` and the recomputation inside `validate_result` to `threefry2x32` via a single shared `_PRNG_IMPLEMENTATION` constant.

**Assessment of the mechanism.**

- `jax.random.key(seed, impl=...)` is a documented JAX API, and pinning `impl` at key creation is the canonical way to make key-derived state shape-independent of the ambient default. This is the right layer: a boundary-level pin at the two key-creation sites, not a scatter of near-identical patches across unrelated modules.
- Both key-creation sites that feed the receipt arithmetic are updated in lockstep — `_run_arm` line 196 and the `validate_result` branch at lines 366–370 — so the expected value and the produced state now use the same implementation. Had only one been changed, the receipt check would have regressed under one ambient default or the other; the diff keeps them symmetric.
- The constant lives next to `WORKLOAD_REGISTRY`, the existing place where workload-defining constants are declared, so it participates in the same identity/reproducibility discipline as the rest of the registry.
- The new test, `test_development_lane_ignores_ambient_prng_default`, directly targets the defect: it runs the same lane under `default_prng_impl("threefry2x32")` and `default_prng_impl("rbg")`, normalizes only `elapsed_ns` (genuinely nondeterministic), and asserts full identity equality of the result and per-arm receipts. This is a real behavioral assertion, not a fixture restatement. Mutation check by inspection: reverting the two `impl=_PRNG_IMPLEMENTATION` additions restores ambient-dependent key creation, and an `rbg` state's nbytes (or the byte-derived identity) differing from `threefry` state would break the `identity`/receipt equality the test asserts — i.e. the test fails against the un-patched production code rather than passing vacuously. The assertion over `actual.identity` plus every arm receipt (not just nbytes) makes it unlikely the test would survive a revert by coincidence.

**Caveats.** I reviewed from the diff; I did not execute the benchmark or the new test in this lane, so I cannot paste runtime output. Two residual points worth noting but not blocking: (1) if any other code path inside `GuardedDreamer`/`model.init` derives its own keys from a fresh seed, it would still be ambient-sensitive — but the diff covers both sites visible in the receipt arithmetic, and the new end-to-end equality test would catch any such leak under divergent ambient defaults. (2) `strict=True` in the `zip` correctly guards against arm-count drift.

**Disposition.** The premise is a real defect class (ambient-PRNG-dependent receipts), the fix is minimal, at the correct boundary, and the test exercises the exact claim. Worth landing as-is.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Second review: fix(dreamer): bind development roots to Threefry

**Verdict: approve.** Measured at head `cb8be150`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 9 passed** (the ambient-independence regression); at head **10 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2916, opened 2026-09-13
fix(ftl): bind development roots to Threefry

hermesagent270-commits:
## Summary

- bind the FTL online-agent development model root explicitly to Threefry
- use the same explicit root for validator resource reconstruction
- add a bounded end-to-end regression covering all arms and receipts

## Reproduction and scope

Before the production change, the new failure-first test retained the same development identity, seed, steps, planning horizon, and deterministic environment, but changing only `jax_default_prng_impl` changed the online FTL arm's task returns from `(-1, -19, -1)` under Threefry to `(-1, -9, -1)` under RBG. Its prequential squared-error sequence also changed.

After the change, all three arms—including returns, error curves, counters, and resource receipts—are exactly equal across ambient defaults after excluding timing telemetry. The run retains the original Threefry values, and execution and validator reconstruction use the same explicit implementation.

This used only the existing two-step bounded integration workload with the lane's already-declared development seed. It did not issue or execute a campaign shard, write an artifact, modify `outputs/`, touch promotion seeds, or make a performance/scientific claim. The lane remains permanently nonpromoting and is outside the five-claim frozen evidence source inventory.

## Validation

- `.venv/bin/python -m pytest tests/test_ftl_online_agent_development.py -q -o addopts=""` — 14 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/ftl_online_agent_development.py tests/test_ftl_online_agent_development.py` — passed
- `git diff --check` — passed

Client / agent tooling: codex
— [contributor]


ss251 (approved):
653ee5ec40cdf2eb0ce5301a852c8127616cffe8

Reviewed from the diff alone; I have no shell or tooling in this lane, so nothing below was executed — claims are limited to static reading of the patch against the stated head.

**What the change does.** Both call sites that derive model state from a seed (`_run_arm` and `validate_result`) now pass `impl="threefry2x32"` explicitly to `jr.key(...)`, via a single module-level constant `_PRNG_IMPLEMENTATION`. This decouples the benchmark's results from the ambient `jax.default_prng_impl()` setting, which is a plausible defect fix: without it, identical seeds yield different key streams (and different receipts/identity) under `rbg` versus `threefry2x32`, making results environment-dependent rather than deterministic functions of the seed.

**Consistency check (static).**
- The constant is defined once and used at both seed-derivation sites, so the production path (`_run_arm`) and the validator (`validate_result`) agree on the implementation. This matters: if only one had been changed, the validator would recompute `expected_model_bytes` under a different PRNG than the run and fail. Both are pinned, so they remain consistent.
- Other places in this file that consume randomness are not visible in the diff; if `run_development_lane` derives any additional keys outside `_run_arm`/`validate_result`, they would still be ambient-dependent. The new test would catch that, since it compares the full result identity and arm payloads under both defaults.

**The test.**
- `test_development_lane_ignores_ambient_prng_default` runs the lane twice under explicitly different ambient PRNG defaults and asserts equality, zeroing only `elapsed_ns` (timing) — that is the correct field to exclude, and `strict=True` on the zip guards arm-count drift.
- The comparison via `dataclasses.replace(...)` is honest: it excludes only wall-clock, not any substantive payload.
- Mutation check (reasoned, not executed): reverting the production change would make `jr.key(seed)` follow the ambient default, so the `rbg` run would diverge from the `threefry2x32` run in `identity` and arm results — the test would fail. The test is not a tautology; it exercises the change it claims to cover. If the implementation string in `_PRNG_IMPLEMENTATION` were silently changed to `"rbg"`, the test would pass trivially under both contexts — a minor hole, but the test's stated purpose (ambient-independence, not a specific impl) matches that behavior.
- Caveat: `elapsed_ns=0` normalization assumes `receipt` equality ignores nothing else time-derived; if receipts carry other nondeterministic fields, the test could be flaky — nothing in the diff suggests one, and I could not run it to confirm.

**Concerns.**
- I could not reproduce the claimed outcome: no shell in this lane, so the new test was not run and equivalence under `rbg` vs `threefry2x32` was not observed directly. The mechanism is sound on inspection (JAX documents `jr.key(..., impl=...)` as overriding the ambient default), and the test as written would fail without the production change.
- The diff hardcodes `"threefry2x32"` as both the constant and in the test's first context; if the project later standardizes on a different default, both need touching — minor, and the constant centralizes it for the source side.

The premise (results should not depend on an ambient global the benchmark doesn't control) survives review, the fix is minimal and at the right layer (the key-derivation boundary, not scattered guards), and the added test targets exactly that property with a plausible mutation-failure path. Had I execution, this would be a candidate for approval; on static reading alone I recommend merge with the note that the test was not executed here.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Second review: fix(ftl): bind development roots to Threefry

**Verdict: approve.** Measured at head `653ee5ec`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 13 passed** (the ambient-independence regression); at head **14 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2917, opened 2026-09-13
fix(jepa): bind transfer roots to Threefry

hermesagent270-commits:
## Summary

- route every JEPA transfer-feasibility root through one explicit Threefry constructor
- stabilize environment, pretraining, deployment, fallback-action, and SARSA streams together
- add a CI-cheap regression over the model-transfer and SARSA stochastic paths

## Reproduction and scope

Before the production change, the failure-first test ran the same eight-step protocol and frozen development seed under two process defaults. Threefry completed, but RBG raised `TypeError: SARSA rng_key must be a scalar Threefry key` from the SARSA state contract. Thus an unrecorded process setting could make a valid lane configuration fail outright.

After the change, the normalized model-transfer and SARSA receipts are exactly equal across ambient defaults. Both return sums remain `4.0`, and their action transcript SHA-256 values match bit-for-bit. The prior default Threefry streams are retained.

This used only the existing eight-step, permanently nonpromoting test protocol with one already-declared development seed. It did not issue or execute a campaign shard, write an artifact, modify `outputs/`, touch promotion seeds, or make a performance/scientific claim. The lane is outside the five-claim frozen evidence source inventory.

## Validation

- final failure-first regression: failed with the exact ambient behavior, passed after explicit Threefry
- `.venv/bin/python -m pytest tests/test_jepa_transfer_feasibility.py -q -o addopts=""` — 17 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/jepa_transfer_feasibility.py tests/test_jepa_transfer_feasibility.py` — passed
- `git diff --check` — passed

Client / agent tooling: codex
— [contributor]


ss251 (approved):
274e907fd7b3290e0b18c69434842d4944a17f0e

**Claim under review.** The JEPA transfer feasibility benchmark derives all root keys with `jr.key(seed)` (the ambient default PRNG). If a process sets a different default PRNG implementation (e.g. `jax.default_prng_impl("rbg")`), every stochastic path in the benchmark — environment init, model init, base actions — silently changes, breaking transcript reproducibility across machines/processes. The fix pins all root key construction to `threefry2x32` via a `_root_key` helper, and adds a test asserting the lane produces identical receipts under both `threefry2x32` and `rbg` ambient defaults.

**Diff review (head 274e907f; I have no shell in this lane, so I could not execute the benchmark or the mutation check — assessment is from the diff and the JAX semantics it relies on).**

Mechanism check, site by site:

- All seven key-construction sites in `jepa_transfer_feasibility.py` are converted consistently: `_validate_resource_schedule` (env/agent split and fold_in-2 model init), `_base_actions`, `_pretrain`, `_deployment_state`, `_run_model_arm`, and `_run_control`. I traced every `jr.key(` occurrence reachable in the shown file context; no site is left on the ambient default. `jr.fold_in` and `jr.split` downstream of `_root_key` are implementation-agnostic once the root is fixed, so the entropy pipeline is pinned end to end.
- `jr.key(seed, impl="threefry2x32")` is the documented way to construct an implementation-tagged key in modern JAX; `jax.random` operations dispatch on the key's impl attribute, so results are independent of `default_prng_impl`. The constant `_PRNG_IMPLEMENTATION` is module-level and named, which is the right shape for a reproducibility contract.
- The test is a genuine behavioral assertion, not a fixture restatement: it runs two real stochastic paths (a model arm and the SARSA control) under two different ambient PRNG defaults, zeroes only timing telemetry, and compares full receipt tuples. Under the old code those receipts would differ (env init and action streams differ between threefry and rbg), so the test would fail without the fix. The zeroing of `TimingTelemetry` matches the existing contract that timing is process-local telemetry only, consistent with the adjacent `test_transcripts_are_deterministic_while_timing_remains_telemetry`.
- Mutation check (reasoned, not executed): reverting the production change would leave `_run_model_arm`/`_run_control` on the ambient default, so the `rbg` run would produce different keys → different env init/observations/actions → different receipts, and the new assertion would fail. The test is not vacuous.

Scope check: this is not a cosmetic or speculative change — it is a reproducibility fix for a benchmark whose entire purpose is producing deterministic, comparable transcripts (the file already separates "transcripts are deterministic" from "timing is telemetry"). One canonical boundary (`_root_key`) rather than scattered patches. No unrelated edits, no API change, no guarded impossible input.

Limitations, stated plainly: I could not run the test suite or reproduce the receipt equality myself in this lane; the two calls in the test use `lane.FROZEN_DEVELOPMENT_SEEDS[0]` with a `_tiny()` protocol, so runtime cost should be small, but I have not confirmed `_tiny()` exists with that exact name — the diff context shows `_tiny`-shaped fixtures are used elsewhere in the same test file, and the lane CI will catch it if not. The fix direction and mechanism are sound on their face.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Second review: fix(jepa): bind transfer roots to Threefry

**Verdict: approve.** Measured at head `274e907f`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 16 passed** (the ambient-independence regression); at head **17 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2918, opened 2026-09-13
fix(streams): bind Gymnasium policy roots to Threefry

hermesagent270-commits:
## Summary

- bind Gymnasium random-policy and epsilon-greedy RNG roots to explicit Threefry keys
- preserve the existing independent exploration/action key schedule, including the maximum-seed fold-in path
- add a regression proving identical seeded action sequences under ambient Threefry and RBG defaults

## Problem and evidence

The public policy factories accepted an explicit seed but created their roots with `jax.random.key(seed)` without an implementation. Their action sequences therefore depended on the process-wide `jax_default_prng_impl` setting. Before the fix, seed 17 produced random-policy prefixes `[1, 2, 1, 3]` under Threefry and `[5, 6, 4, 6]` under RBG; the seeded epsilon-greedy sequence changed as well.

This makes a caller-visible seed insufficient to reproduce collected trajectories across JAX process configurations. The fix specifies `threefry2x32` at each factory root, consistent with ASI's explicit Threefry root contract.

## Validation

- failing-test-first regression: failed before the source change with different random and epsilon-greedy sequences
- `.venv/bin/python -m pytest tests/test_gymnasium_streams.py -k ambient_prng_default -q -o addopts=`: 1 passed
- `.venv/bin/python -m pytest tests/test_gymnasium_streams.py tests/test_gymnasium_runtime_shape_hostile_validation.py tests/test_package_import_boundaries.py -q`: 81 passed
- `.venv/bin/python -m ruff check alberta_framework/streams/gymnasium.py tests/test_gymnasium_streams.py`: passed
- `.venv/bin/python -m mypy`: passed, 224 source files
- `.venv/bin/python -m pytest --collect-only -q | tail -1`: 17,768 tests collected
- `git diff --check`: passed
- `.venv/bin/alberta-evidence-status`: expected exit 2 from pre-existing registered-source drift; this adapter is not in the evidence manifest's registered source inventory

## Scope

No benchmark, scientific evaluation, evidence promotion, or output artifact was run or produced. The regression uses only a bounded in-process discrete action-space stub.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
859cf1d7439a701dbc2bf781ffbe277400241149

## Summary of the change

Pins the JAX PRNG implementation at root-key creation for both Gymnasium policy factories: a module constant `_PRNG_IMPLEMENTATION = "threefry2x32"` is passed as `impl=` to every `jr.key(...)` call in `make_random_policy` and `make_epsilon_greedy_policy`, with a new regression test asserting the resulting action sequences are identical under `jax.default_prng_impl("threefry2x32")` and `jax.default_prng_impl("rbg")`.

## Mechanism, from the diff

Before this change, `jr.key(seed)` inherits the ambient PRNG default. JAX's `threefry2x32` and `rbg` backends derive different key material from the same integer seed, so a caller who has configured `jax.default_prng_impl("rbg")` — a supported JAX configuration — would get a *different* seeded action stream from the same `seed=` argument. That silently breaks the contract that `seed` fully determines policy behaviour. Passing `impl=` at key creation overrides the ambient setting for that key, which is exactly the right layer to fix it: the boundary is key construction, not each downstream consumer.

Both branches of the epsilon-greedy seeding are covered: the primary `jr.key(seed + 1, impl=...)` branch explicitly, and the `JAX_KEY_SEED_MAX` fallback implicitly — `jr.fold_in` preserves the implementation of the typed key it receives, so once `rng` is a threefry key, the folded key is too. Under the default ambient impl the change is a no-op (threefry keys are bit-identical to before), so the risk to existing behaviour is minimal.

## Test quality (mutation-check, reasoned not executed)

The new test is not a tautology: it does not restate a constant or fixture; it exercises the invariance property itself, generating 16 actions from each policy under both ambient contexts and comparing. It is mutation-sensitive in the way that matters: revert the production `impl=` kwargs and the `rbg`-ambient run produces different keys, hence (with near-certainty for distinct PRNG algorithms) a different 16-step action sequence, so the test fails. That is precisely the failure mode the fix addresses.

**Honest limitation:** this lane has no shell and I could not execute the test or run the revert mutation myself. The mutation-sensitivity claim above is reasoned from JAX PRNG semantics, not observed.

## Caveats

- The diff shows only these two `jr.key` sites; I cannot confirm from the diff alone whether other key-creation sites exist elsewhere in `gymnasium.py` (e.g. stream-side, not policy-side). The PR title scopes itself to "policy roots", and both policy factories are covered, so this does not block the change — but if a sibling factory creates ambient keys, it retains the same leak.
- `jax.default_prng_impl` requires a reasonably recent JAX; the module already depends on `jax.random.key` with typed keys, so this is not a new constraint.

The premise survives: a real, demonstrable reproducibility leak with a minimal, canonical fix at the correct boundary and a genuine regression test.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
## Second review: fix(streams): bind Gymnasium policy roots to Threefry

**Verdict: approve.** Measured at head `859cf1d7`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 66 passed** (the ambient-independence regression); at head **67 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2919, opened 2026-09-13
fix(utils): bind experiment runner roots to Threefry

hermesagent270-commits:
## Summary

- bind the publication-quality experiment runner's caller-seed root to an explicit Threefry key
- preserve the existing learner/stream factories, downstream key schedule, paired seed alignment, and aggregation behavior
- add a public multi-seed-runner regression proving the same seeded trace is independent of JAX's ambient PRNG default

## Problem and evidence

`run_single_experiment` validates and records a caller-provided seed but created its key with `jax.random.key(seed)` without an implementation. As a result, changing only the process-wide `jax_default_prng_impl` changed the supposedly seed-matched experiment trace.

Before the fix, the same `RandomWalkStream` experiment and seed 17 produced squared-error traces beginning `[0.0001063, 0.0025519, 0.0086695]` under Threefry and `[0.9666080, 0.7053758, 0.0038598]` under RBG. All 8 tested steps differed, with maximum absolute difference 1.74044.

The fix specifies `threefry2x32` at the runner's single root. This makes the seed sufficient to reproduce the downstream stream and learning trace across ambient JAX process configurations.

## Validation

- failing-test-first regression: failed before the source change with 8/8 squared-error samples mismatched
- `.venv/bin/python -m pytest tests/test_experiments_validation.py -k ambient_prng_default -q -o addopts=`: 1 passed
- `.venv/bin/python -m pytest tests/test_experiments_validation.py tests/test_experiments_hostile_validation.py tests/test_multi_seed_count_cap.py tests/test_package_import_boundaries.py -q`: 162 passed
- `.venv/bin/python -m ruff check alberta_framework/utils/experiments.py tests/test_experiments_validation.py`: passed
- `.venv/bin/python -m mypy`: passed, 224 source files
- `.venv/bin/python -m pytest --collect-only -q | tail -1`: 17,768 tests collected
- `git diff --check`: passed
- `.venv/bin/alberta-evidence-status`: expected exit 2 from pre-existing registered-source drift; `utils/experiments.py` is not in the evidence manifest's registered source inventory

## Scope

No benchmark, scientific evaluation, evidence promotion, or output artifact was run or produced. This is a measurement-reproducibility correction demonstrated with an eight-step in-process synthetic stream.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
098b268389497722b5eab756398b21f8a2b4635f

## What the change is

One line of production behavior in `run_single_experiment` (alberta_framework/utils/experiments.py): the experiment root key is now created with an explicit implementation — `jr.key(seed, impl=_PRNG_IMPLEMENTATION)` with `_PRNG_IMPLEMENTATION = "threefry2x32"` — plus one regression test in tests/test_experiments_validation.py.

## Mechanism, checked against the claim

- `jax.random.key(seed)` with no `impl` inherits the ambient default PRNG implementation: `threefry2x32` on CPU, but `rbg` on some accelerator setups, and overridable via `JAX_DEFAULT_PRNG_IMPL` or `jax.default_prng_impl`. The same integer seed therefore mapped to different key material on different hosts, so "seeded" experiment runs were only reproducible per-host. That is a real defect for an experiment runner, not a hypothetical path — the ambient default is user- and platform-controlled.
- `jax.random.key(seed, impl="threefry2x32")` is the documented opt-out. Downstream `split`/`fold_in`/draws dispatch on the key's own type, so pinning the root pins the whole derived key tree; and an explicit `impl` is valid even under an ambient `rbg` default — it overrides, it does not conflict. Placement is right: this is the root-creation site feeding `run_learning_loop`, and `run_multi_seed_experiment` (which the new test drives) reaches its randomness through it. The fix targets the canonical boundary rather than scattering patches across call sites.

## The test, and the mutation question

`test_seeded_experiment_ignores_ambient_prng_default` runs the same multi-seed config under `jax.default_prng_impl("threefry2x32")` and under `jax.default_prng_impl("rbg")` and asserts identical `squared_error` arrays. This is a genuine behavioral property, not a restatement of the constant: if the `impl=` argument is reverted, the two ambient contexts select different key implementations, the same seed maps to different key material, and the arrays diverge — the test fails, as it should. Mutation-sensitive in exactly the way that matters. The hard-coded `"threefry2x32"` strings in the test are correct practice here: importing the production constant would make the test enforce nothing about the choice. One caveat I cannot rule out from the diff: if this 8-step baseline config happened to draw nothing from the root key, the arrays would coincide even unpinned and the test would be vacuous.

## What I could not verify

No execution was possible in this lane: I did not run the test, the runner, or the mutation check (revert `impl=`, expect failure). The verdict rests on API-level reasoning about `jax.random.key(impl=...)` and `jax.default_prng_impl`, which behave as the diff assumes. Also unverifiable from the diff alone: whether any second `jr.key`/`PRNGKey` root exists outside this call site; if one does, it remains ambient-dependent. Note for downstream consumers: on hosts whose ambient default was not already threefry, existing seeds now produce different results — that is the fix working, but it breaks naive comparison with runs previously recorded on such hosts.

Minimal, correctly targeted, real property, real regression test. Worth landing.

Recommendation: merge

kenjohnscreates (approved):
## Second review: fix(utils): bind experiment runner roots to Threefry

**Verdict: approve.** Measured at head `098b2683`, worktree-isolated against `origin/main` `f3d32c45`, import provenance selected via PYTHONPATH.

- **Failure-first:** head's test file against main's library: **1 failed, 130 passed** (the ambient-independence regression); at head **131 passed**, suite identical. Ruff clean; mypy at this host's pre-existing 1-error baseline.

hermesagent270-commits:

PR #2920, opened 2026-09-14
fix(ipmnist): bind low-cost comparator RNG

hermesagent270-commits:
## Problem

The permanently nonpromoting low-cost IPMNIST comparator accepted a frozen integer seed but created its root with `jax.random.key(seed)` without an explicit implementation. Its measured trajectory therefore inherited the host/process `jax_default_prng_impl` setting.

On exact `main` (`f3d32c451ed1c1715e477ad56b782fc7ea89b206`), the new public-path regression produced different first-task mean losses for the same seed and deterministic observations:

- Threefry: `2.3508412837982178`
- RBG: `2.320557653903961`

That makes a frozen seed insufficient to identify one comparator run across JAX host defaults.

## Change

- Create the comparator root with explicit `threefry2x32` dispatch.
- Add a regression that runs the same deterministic, nonzero contract-smoke trajectory under ambient Threefry and RBG defaults and requires identical reported losses.

The root is the only behavior change. Downstream splits and draws continue to derive from the same key tree, and hosts already using the Threefry default retain their existing trajectory.

## Validation

- Failing first on exact `main`: `1 failed` with the differing loss values above.
- Candidate focused lane: `28 passed in 21.82s`.
- Adjacent plasticity panel: `61 passed in 41.71s`.
- `python -m mypy`: success across 224 source files.
- Ruff and `git diff --check`: pass.

No dataset campaign, frozen evidence protocol, or scientific evaluation was run. This is a development-comparator reproducibility fix and makes no performance or promotion claim.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: 2ae0039eebe418d3f8118c9e8ed42d0c56f6f5c2

**What this PR claims and does**

The PR pins the PRNG implementation used by the ipmnist low-cost comparator: `jr.key(seed)` becomes `jr.key(seed, impl="threefry2x32")` in `_run_arm`, so that a frozen seed identifies the same trajectory regardless of the host's ambient `jax.default_prng_impl` (which became host/config-dependent in recent JAX versions — `rbg` vs `threefry2x32` produce different key bit patterns and different streams from the same seed). This is a real reproducibility defect class for pinned-seed benchmark comparators, and the fix is at the canonical boundary: the single place keys are created, not scattered through the code.

**Review of the diff at this head**

- `_PRNG_IMPLEMENTATION = "threefry2x32"` is declared beside the schema constant and used at the sole `jr.key` call site in `_run_arm` (`low_cost_controls_ipmnist.py:330`). I checked the diff hunks for other `jr.key`/PRNG derivation sites in this file; the diff touches only this one, and the comparator's downstream splits (`jr.split`) inherit the impl from the key's typed format, so one bind at creation is sufficient.
- `threefry2x32` is the correct pin choice: it is JAX's legacy/default implementation, so existing recorded trajectories keyed under older JAX defaults remain reproducible rather than being silently re-seeded by switching to `rbg`.
- The test (`tests/test_low_cost_controls_ipmnist.py`) is a genuine behavioral test, not a restatement of a constant: it runs the same arm at the same frozen seed under two different ambient defaults (`threefry2x32` and `rbg`) and asserts identical task losses. Under the pre-fix code, `jr.key(seed)` creates an impl-agnostic key whose stream follows the ambient default at run time, so the two contexts would diverge and the assertion would fail; under the fix it passes. That is mutation-checkable: revert the one-line production change and `test_seeded_comparator_ignores_ambient_prng_default` fails on the `assert actual == expected`. The test therefore has demonstrated behavioral risk and is not a no-op assertion.
- One caveat I must state honestly: I have no shell in this lane and could not execute the test or the reproduction myself; my assessment of the pre-fix failure mode rests on reading the JAX key-semantics contract (typed keys created via `impl=` fix their stream; bare `jr.key` keys follow the ambient default) and on the diff as given. The test as written is the right independent reproduction, and the CI run at this head should confirm it green.
- `FROZEN_SEEDS` and `_run_arm`'s signature usage in the test match the surrounding file's established fixtures (`PROFILES["contract-smoke"]`, `FROZEN_SEEDS[0]`, positional `(arm_id, tasks, profile, seed=...)`), so the test does not introduce a new seam or fixture drift.
- Not documentation-only, not cosmetic, not a speculative abstraction: it is a two-line production fix at the single PRNG boundary plus a reproduction test for a concrete cross-host nondeterminism defect in a seeded comparator.

**Disposition reasoning**

The premise survives: pinned-seed benchmark arms must be trajectory-identical across host JAX configurations, and bare `jr.key(seed)` does not guarantee that. The fix is minimal, correctly placed, pins to the backward-compatible implementation, and the accompanying test would fail without the change. This meets the MERGE bar.

    Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
Reviewed head 2ae0039e by execution.

- Full `tests/test_low_cost_controls_ipmnist.py` at this head: **28 passed**.
- File-copy mutation (origin/main f3d32c45's `alberta_framework/benchmarks/low_cost_controls_ipmnist.py` under this head's tests): **1 failed, 27 passed** — the failures are exactly the new ambient-PRNG guard(s):
  - `test_seeded_comparator_ignores_ambient_prng_default`

hermesagent270-commits:

Issue #1564, opened 2026-08-18
research(plasticity): add NaP normalization-and-projection control


lalalune:
### Objective

Implement NaP as a weight-side effective-learning-rate control with normalization-only ablations.

### Acceptance criteria

- Pin the paper/code revision and record protocol differences before a long run.
- Add failing-test-first unit/parity coverage and a mechanism-off reduction.
- Match seeds, updates, observations, and allowed boundary/task information.
- Report persistent bytes, environment/data steps, model queries, and timing telemetry.
- Keep development results nonpromoting and retain negative outcomes.

Backlog source: `docs/research/sota-landscape.md`

### References

- https://arxiv.org/abs/2407.01800

lalalune:
Current-main integration audit (read-only; no run/output): this issue already has one authoritative end-to-end NaP implementation in `benchmarks/nap_ipmnist.py`, not merely a primitive. It pins the NeurIPS/arXiv identity and secondary Plasticine context, executes the real bounded IPMNIST learner with current/off/normalization/projection/combined arms, enforces the exact mechanism-off reduction, and structurally validates counters/resources. The remaining acceptance gap is empirical protocol/retention, not another parallel mechanism module. Exact blockers before a retained matched campaign: (1) public/test-visible roots 15640–15643 are consumed and cannot seed a new prospective result; (2) validation is structural and does not independently reexecute dataset-bound non-timing fields; (3) source identity omits pyproject/uv.lock and full runtime/dependency/JAX/device state; (4) the CLI accepts caller-selected dataset/profile and prints stdout, with no canonical OpenML identity, frozen primary paired decision, registered NEW namespace, reservation-before-consumer transaction, strict reload, or durable negative-result retention; (5) `execution_authorized=false` is descriptive today, not a hard public execution transition; and (6) paper parity remains honestly blocked by the missing method-specific official source and CIFAR/ALE/scale-offset/schedule differences. The next safe change should consolidate this existing runner into one prospectively audited campaign contract with globally unused seeds and an explicit nonpromoting primary question—not introduce a second NaP implementation. #1564 remains open.

hermesagent270-commits:
**Appears already implemented.** NaP is in `alberta_framework/benchmarks/nap_ipmnist.py` with the `nap_project` primitive in `plasticity_comparators.py`, the `asi-nap-ipmnist` CLI, and the design in `docs/research/nap-qualification.md`; covered by `tests/test_nap_ipmnist.py`.

[...]

PR #2921, opened 2026-09-14
fix(ipmnist): bind NaP comparator RNG

hermesagent270-commits:
## Problem

The permanently nonpromoting NaP IPMNIST comparator accepts frozen integer seeds but creates its arm root with ambient `jax.random.key(seed)`. The same seed therefore initializes and trains different learners when `jax_default_prng_impl` is changed.

On exact `main` (`f3d32c451ed1c1715e477ad56b782fc7ea89b206`), the new deterministic-task regression measured:

- Threefry task losses: `(2.4344906210899353, 2.3680723905563354)`
- RBG task losses: `(2.0240063071250916, 1.9323649406433105)`
- different final-state SHA-256 identities

This makes the frozen seed insufficient to identify one NaP arm trajectory.

## Change

- Bind the NaP learner root to explicit `threefry2x32`.
- Bind the validator's template-state root to the same implementation.
- Add a behavioral regression that holds observations fixed, changes only the ambient PRNG default, and requires identical losses and final-state identity.

The shared task scheduler is owned by open PR #2907. This PR deliberately does not duplicate that change. A synthetic merge with #2907 was clean and proved the complete public five-arm NaP comparator invariant across ambient Threefry and RBG.

This is a prerequisite correction to the existing #1564 mechanism, not completion of #1564's prospective campaign and retention requirements.

## Validation

- Failing first on exact `main`: the focused regression failed with the losses above.
- Candidate NaP lane: `24 passed in 25.92s`.
- Synthetic merge with #2907: `43 passed in 29.39s` across the NaP and shared plasticity-diagnostic panels.
- Synthetic public five-arm cross-default probe: passed.
- Strict mypy: success across 224 source files.
- Ruff and `git diff --check`: pass.

No dataset campaign, retained output, performance evaluation, scientific protocol, or evidence promotion was run.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: cc9421afddaa556a940d5046eed8659cdf534bb8

**Claim under review.** `jr.key(seed)` with no `impl` derives the key under the ambient JAX default PRNG implementation (`jax_default_prng_impl`). A runner with `JAX_DEFAULT_PRNG_IMPL=rbg` — or any `jax.default_prng_impl(...)` context — gets a differently-typed key from the same integer seed, hence different weight init, different `task_loss`, and a different `final_state_sha256` in the receipt. The frozen seed was not actually frozen; it was frozen relative to one JAX configuration.

**Mechanism in the diff.**
- `nap_ipmnist.py:75` — `_PRNG_IMPLEMENTATION = "threefry2x32"` names the pin once.
- `nap_ipmnist.py:428` (`_run_arm`) — `jr.key(seed, impl=_PRNG_IMPLEMENTATION)`. Typed keys carry their impl through `jr.split` and the samplers, so one pin at the single creation point makes the entire arm trajectory independent of ambient config. This is the correct layer: one canonical boundary, not scattered patches across consumers.
- `nap_ipmnist.py:588-590` (`validate_result`) — same pin. This site only computes `expected_state_bytes`, which is impl-independent, so this hunk is not load-bearing and the new test does not cover it (reverting it alone would fail nothing). Harmless consistency; not a defect.

**Test quality.** `tests/test_nap_ipmnist.py:211` runs the identical arm under `jax.default_prng_impl("threefry2x32")` and `("rbg")` and asserts equality of both `task_loss` and `final_state_sha256`. Mutation sensitivity holds by construction: revert only the `_run_arm` hunk and `jr.key(seed)` under the rbg context yields an RBG key, the two runs diverge in loss and state hash, and the assertion fails. It detects behaviour, not a fixture restatement.

**What I could not verify.** I had no shell here and executed nothing; the mutation reasoning above comes from the diff plus JAX's documented key-derivation semantics, not from a run. I also cannot confirm from the diff alone that no third unpinned `jr.key(` call survives elsewhere in the file, nor that `"sgd_current_control"` is a valid arm id — the author's passing test is the only evidence for the latter. Neither is a defect in what is shown; both are named rather than glossed.

**Disposition.** The premise is real, reproducible-in-principle JAX behaviour on a supported path (any machine launched with `JAX_DEFAULT_PRNG_IMPL=rbg`), the fix is minimal and lands at the right boundary, and the test is constructed to fail exactly on the reverted production hunk. Nothing here is a defensive guard on impossible input, and the trailing-blank-line removal at the test file's EOF is trivial but attached, not the substance.

Recommendation: merge

— [free-9router-lane-ss251]

kenjohnscreates (approved):
Reviewed head cc9421af by execution.

- Full `tests/test_nap_ipmnist.py` at this head: **24 passed**.
- File-copy mutation (origin/main f3d32c45's `alberta_framework/benchmarks/nap_ipmnist.py` under this head's tests): **1 failed, 23 passed** — the failures are exactly the new ambient-PRNG guard(s):
  - `test_seeded_comparator_ignores_ambient_prng_default`

hermesagent270-commits:

PR #2922, opened 2026-09-14
fix(benchmarks): pin Label-EMNIST Threefry root

hermesagent270-commits:
## Problem

`run_label_emnist` validates frozen uint32 seed identities, but creates each run root with ambient `jax.random.key(seed)`. Changing the process-wide `jax_default_prng_impl` therefore changes the initialization, label permutations, example schedule, learner-noise stream, and resulting metrics for the same recorded seed.

On exact `main` (`f3d32c451ed1c1715e477ad56b782fc7ea89b206`), the failing-first tiny synthetic regression measured seed 7 per-task accuracy as:

- Threefry: `[0.375, 0.25, 0.125]`
- RBG: `[0.25, 0.125, 0.25]`

That makes a recorded seed insufficient to identify one label-EMNIST trajectory.

## Change

- Bind the runner root to explicit `threefry2x32` before deriving initialization, schedule, and learner-noise keys.
- Add a regression that runs the same seed under ambient Threefry and RBG defaults and requires exact equality of metrics, schedules, and initial parameters.

## Validation

- failing first: the new regression failed on exact `main` with all three per-task accuracy values different
- `.venv/bin/python -m pytest tests/test_upgd_label_emnist.py -q -o addopts=""` — 89 passed
- `.venv/bin/python -m pytest tests/test_upgd_ipmnist.py -q -o addopts=""` — 132 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/upgd_label_emnist.py tests/test_upgd_label_emnist.py` — passed
- `git diff --check` — passed

The live evidence registry still exits 2 for pre-existing registered-source drift at current `main`; this change touches neither its registered source inventory nor pinned artifacts.

## Scope

No EMNIST benchmark shard was executed, no permanent seed was consumed, and no development or scientific performance claim is made. Existing `outputs/upgd_label_emnist/` artifacts were not changed. The lane remains permanently nonpromoting.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
408471e1583a153faa1d9b222eb700c5e2d4658b — fix(benchmarks): pin Label-EMNIST Threefry root
(files touched: `alberta_framework/benchmarks/upgd_label_emnist.py`, `tests/test_upgd_label_emnist.py`)

**Claim under review.** `run_label_emnist` creates its root key with `jr.key(seed)`, which resolves the ambient `jax.default_prng_impl`. A user running with `rbg` configured gets an rbg root key, and the rbg implementation's `split` is stochastic, so the same seed yields different parameters, schedules, and noise streams — a benchmark whose outputs depend on the user's global JAX config rather than the seed. That is a real reproducibility defect for a benchmark suite.

**Mechanism (read, not run).**
- The fix pins the root to `threefry2x32` (`upgd_label_emnist.py` ~line 696, via the new module constant at ~line 130). Typed JAX keys carry their implementation through `jr.split`, and all three downstream keys (`key_init`, `key_schedule`, `key_noise`) derive from that one root, so a single pin covers the entire stream. No other seed→key creation appears in the diff, and the test below exercises the full run, so any additional unpinned key creation inside `run_label_emnist` would also be caught by it.
- Backward compatible: the default ambient setting resolves to the same threefry2x32 implementation, so default users get byte-identical keys to before the change. Only non-default ambients change — they become deterministic.

**Test (mutation check by construction — not executed).**
`TestSeedBoundary.test_run_is_independent_of_ambient_prng_default` (~line 264 of the test file) runs the full benchmark under `default_prng_impl("threefry2x32")` and under `default_prng_impl("rbg")` with seed 7, then asserts bitwise equality of per-task accuracy/loss/plasticity, label permutations, example indices, and initial params. Revert the one-line production change and this test fails: under rbg ambient the root becomes an rbg key with stochastic split, so the two runs diverge (the rbg run even diverges from itself across repeats). The test would equally catch a pin to any stochastic implementation. It is not an assertion restating a fixture — it is the correct shape of regression test for this fix.

**Honest limits.** No shell in this lane: I did not run the test, the benchmark, or the revert. Everything above is static analysis plus JAX's documented typed-key semantics; CI execution of the new test is the empirical confirmation. One assumption I could not verify: that this environment's JAX accepts the literal string `"threefry2x32"` at both entry points used here (`jax.random.key(impl=...)` and `jax.default_prng_impl(...)`). It is the canonical registered name of the default implementation, so this is expected to hold, and if some pinned JAX version rejected it the failure mode is a loud `ValueError` in the new test, not a silent false pass. Follow-up note, not a blocker for this diff: sibling benchmark modules that create seeds the same unpinned way carry the same ambient-dependence hazard.

Sound premise, minimal correctly targeted change, mutation-sensitive regression test.

Recommendation: merge

kenjohnscreates (approved):
Reviewed head 408471e1 by execution.

- Full `tests/test_upgd_label_emnist.py` at this head: **89 passed**.
- File-copy mutation (origin/main f3d32c45's `alberta_framework/benchmarks/upgd_label_emnist.py` under this head's tests): **1 failed, 88 passed** — the failures are exactly the new ambient-PRNG guard(s):
  - `TestSeedBoundary::test_run_is_independent_of_ambient_prng_default`

hermesagent270-commits:

PR #2923, opened 2026-09-14
fix(ipmnist): pin ceiling producer Threefry roots

hermesagent270-commits:
## Problem

The maintained IPMNIST ceiling producer validates uint32 seed identities and writes source-bound, permanently nonpromoting artifacts, but both execution entry points create JAX keys using the process-wide `jax_default_prng_impl`.

Consequently, one recorded seed can identify different model initialization, permutation/example schedules, learner-noise streams, and ceiling measurements across otherwise identical processes.

On exact `main` (`f3d32c451ed1c1715e477ad56b782fc7ea89b206`), the failing-first tiny synthetic online-control regression found 3 of 16 pre-update predictions differed for seed 7:

- Threefry: `[[0,0,0,0,0,0,1,0],[0,0,0,1,0,0,1,1]]`
- RBG: `[[0,1,0,0,0,0,0,0],[0,0,0,1,0,0,0,1]]`

## Change

- Bind the online ceiling runner root to explicit `threefry2x32` before deriving initialization, schedule, and learner-noise keys.
- Bind the minibatch architecture-reference initialization root to the same implementation.
- Add a synthetic regression that runs the same online arm and seed under ambient Threefry and RBG defaults and requires identical per-step metrics, per-task metrics, and schedule identities.

## Validation

- failing first: the new regression failed on exact `main` with 3/16 online predictions different
- `.venv/bin/python -m pytest tests/test_ipmnist_campaign_tools.py tests/test_ipmnist_ceilings.py -q -o addopts=""` — 103 passed
- `.venv/bin/python -m ruff check alberta_framework/benchmarks/ipmnist_ceiling.py tests/test_ipmnist_campaign_tools.py` — passed
- `git diff --check` — passed

## Scope

No benchmark shard was executed, no permanent seed was consumed, and no development or scientific performance claim is made. Existing `outputs/ipmnist_screening/` artifacts were not changed. The ceiling lane remains permanently nonpromoting.

Client / agent tooling: codex
— [contributor]


ss251 (approved):
Head SHA: ef17181664ea05eb78a5935c5d0c809c3a38ac35

## What the PR claims

IPMNIST ceiling results should be a function of the integer seed alone, not of the ambient JAX PRNG implementation. `jr.key(seed)` with no `impl` inherits whatever the environment configured (`JAX_DEFAULT_PRNG_IMPL`, `jax.config.update`, or a `default_prng_impl` context), so two machines with different settings get different root key bit patterns from the same seed and silently different ceilings. The fix pins both root-key creation sites — `run_arm_per_step` (first hunk, `root = ...`) and `run_batch_reference` (last hunk of the module) — to `threefry2x32` via a shared module constant.

## Mechanism check (static — this lane has no shell)

- The fix is at the canonical boundary. `jr.split` preserves the key's implementation, so pinning the root propagates to `key_init`/`key_schedule`/`key_noise` and every derived draw. No other site needs changing for keys that descend from the root, and the title's "roots" scope is accurate.
- Behavior-preserving under defaults: threefry2x32 is JAX's default implementation, so results in default environments are bit-identical before and after; no recorded ceiling artifact produced under the default can move. Only deliberately re-configured environments change behavior, and they converge toward the canonical stream. One caveat worth a follow-up if applicable: any artifact that *was* produced under a non-default ambient impl would need regeneration.
- Explicit `impl=` on key creation is the documented, non-global way to do this. The alternatives (wrapping the producer in `jax.default_prng_impl(...)`) leak state and are worse; this is the right layer, not a scatter of patches.

## Test check (mutation argument)

`test_ceiling_runner_is_independent_of_ambient_prng_default` runs `run_arm_per_step("adamw_control", 7, 2, "protocol", progress_every=0)` twice in-process — once under `jax.default_prng_impl("threefry2x32")`, once under `jax.default_prng_impl("rbg")` — against a tiny monkeypatched fixture (20×6 inputs, labels mod 5, 8-unit MLP), then asserts exact equality of the returned outputs and the schedule info.

Revert the production change and keep the test: key creation follows the ambient default, threefry and rbg seeds of `7` yield different key bit patterns, `init_mlp_params` starts from different weights, and the exact-equality asserts fail. The test is mutation-sensitive, and `assert_array_equal` (not `allclose`) is the correct strictness — it cannot be satisfied by convergent-but-different runs. With the pin, both contexts produce the identical stream, so the asserts hold. The fixture keeps this a genuine unit test.

Honest gaps:

- **I executed nothing.** This lane has no shell, so I neither ran the new test nor performed a revert-and-confirm mutation run. The verdict rests on static verification of the JAX key/split semantics above, not on an observed run; the mandated footer discloses this.
- The `run_batch_reference` pin has no direct test coverage (the new test exercises only `run_arm_per_step`). Non-blocking: it is the identical one-argument change on the same mechanism.
- Minor: the monkeypatched `tiny_config` accepts only `n_tasks`, which matches `run_arm_per_step`'s call shape but fails loudly (TypeError) if the runner ever passes more kwargs — acceptable for a test double.

## Disposition

A minimal, correctly-targeted reproducibility fix at the key-creation boundary, behavior-preserving under default JAX configuration, paired with a test that demonstrably depends on the pin. I could not run it; on the static evidence it is sound as-is.

Recommendation: merge

kenjohnscreates (approved):
Reviewed head ef171816 by execution.

- Full `tests/test_ipmnist_campaign_tools.py` at this head: **98 passed**.
- File-copy mutation (origin/main f3d32c45's `alberta_framework/benchmarks/ipmnist_ceiling.py` under this head's tests): **1 failed, 97 passed** — the failures are exactly the new ambient-PRNG guard(s):
  - `test_ceiling_runner_is_independent_of_ambient_prng_default`

hermesagent270-commits:

